PR #4194, opened 2020-06-04
Update yarn cache version

garrettqmartin8:
This should hopefully invalidate the yarn and node_modules cache in order to fix the issue in [this thread](https://ustcdp3.slack.com/archives/CP4U2NKRT/p1591293011090900).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -376,10 +376,10 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - v2-cache-yarn-v2-{{ checksum "yarn.lock" }}
+            - v3-cache-yarn-v3-{{ checksum "yarn.lock" }}
       - restore_cache:
           keys:
-            - v2-mymove-node-modules-{{ checksum "yarn.lock" }}
+            - v3-mymove-node-modules-{{ checksum "yarn.lock" }}
       - run:
           name: Install YARN dependencies
           # setting network concurrency to 1 because using the default failed with errors trying to extract package tar files saying they were corrupt.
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -376,25 +376,25 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - v2-cache-yarn-v2-{{ checksum "yarn.lock" }}
+            - v3-cache-yarn-v3-{{ checksum "yarn.lock" }}
```
gidjin, line 379: not saying we should change this, but curious as to why v2/v3 is in this twice.

.circleci/config.yml
```diff
@@ -376,25 +376,25 @@ jobs:
...
       - run: scripts/check-generated-code yarn.lock
-      # `v2-cache-yarn-v2-{{ checksum "yarn.lock" }}` is used to cache yarn sources
+      # `v3-cache-yarn-v3-{{ checksum "yarn.lock" }}` is used to cache yarn sources
       - save_cache:
-          key: v2-cache-yarn-v2-{{ checksum "yarn.lock" }}
+          key: v3-cache-yarn-v3-{{ checksum "yarn.lock" }}
           paths:
-            - ~/.cache/yarn/v2
+            - ~/.cache/yarn/v3
       # `v2-mymove-node-modules-{{ checksum "yarn.lock" }}` is used to cache installed node modules
```
gidjin, line 395: Small nit, modify this comment to keep it in line

.circleci/config.yml
garrettqmartin8: I'm not sure why.

chrisgilmerproj: Honestly there was probably a `v1` in `~/.cache/yarn/v2` at some point. Each being a different thing. If you consolidate it probably won't hurt anything. Just keep `v3-` as the prefix.

PR #4196, opened 2020-06-05
MB-2334 Allowances table component

donaldthai:
## Description

As a TOO
When I navigate to the Move Details page
I am able to see Allowances information

## Reviewer Notes

- Creates allowances table component for move details page
- Adds allowances table to storybook
- Hooks up the allowances table to the move details page

## Setup

```sh
make server_run
make office_client_run
```

Locally test:
1. Log in to office app
2. Navigate to `http://officelocal:3000/too/customer-moves/`
3. Click on any moves
4. New allowances table should be on page

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2334) for this change

## Screenshots

image


src/pages/TOO/moveDetails.jsx
```diff
@@ -78,6 +80,18 @@ class MoveDetails extends Component {
...
+              branch: customer.agency,
+              rank: moveOrder.grade,
+              weightAllowance: `${allowances.totalWeight} lbs`,
+              authorizedWeight: `${allowances.authorizedWeight} lbs`,
+              progear: `${allowances.proGearWeight} lbs`,
+              spouseProgear: `${allowances.proGearWeightSpouse} lbs`,
+              storageInTransit: `${allowances.storageInTransit} days`,
+              dependents: allowances.dependentsAuthorized ? 'Authorized' : 'Unauthorized',
+            }}
+          />
```
donaldthai, line 94: `weightAllowance`, wasn't sure if this was the total weight or not. Didn't see anything similar in the database or go models.

src/components/Office/AllowancesTable.jsx
```diff
@@ -0,0 +1,65 @@
...
+            <th scope="row">Branch, rank</th>
+            <td data-cy="branchRank">{`${info.branch}, ${info.rank}`}</td>
+          </tr>
+          <tr>
+            <th scope="row">Weight allowance</th>
+            <td data-cy="weightAllowance">{info.weightAllowance}</td>
+          </tr>
+          <tr>
+            <th scope="row">Authorized weight</th>
+            <td data-cy="authorizedWeight">{info.authorizedWeight}</td>
```
duncan-truss, line 28: I'm surprised all these weights are whole numbers and we'll never have to account for decimals or rounding.  Unless that happens at some other stage, I searched the database schema and see almost all integer values.  Are there ever conversions applied from metric to imperial?

kimallen: Was looking at that too- The allowances are pulled from a map of hardcoded (integer) allotments based on the service member/ customer rank.  See `models/entitlements.go`.

src/components/Office/AllowancesTable.jsx
```diff
@@ -0,0 +1,65 @@
...
+            <th scope="row">Dependents</th>
+            <td data-cy="dependents">{info.dependents}</td>
+          </tr>
+        </tbody>
+      </table>
+    </div>
+  );
+};
+
+AllowancesTable.propTypes = {
```
kimallen, line 52: Not sure if it matters (more a curiosity question)- I'm wondering why we pass these in all as strings rather than their original type and then converting them and labeling them within the component

src/pages/TOO/moveDetails.jsx
kimallen: I couldn't figure it out either by looking-And do we want to also include the "plus dependent" total weight in the chart in addition to total weight for self?  Probably worth finding a definitive answer on what this language refers to.

src/components/Office/AllowancesTable.jsx
donaldthai: I didn't know what data type I was working with before hooking it up to the page. I think you bring up a good point and will refactor the expected prop types. Thanks!

src/components/Office/AllowancesTable.jsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import * as PropTypes from 'prop-types';
+
+const AllowancesTable = ({ info }) => {
+  return (
+    <div>
+      <div className="stackedtable-header">
+        <div>
+          <h4>Allowances</h4>
```
duncan-truss, line 9: Would it be semantically better to add this as part of the `<caption>`? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption

donaldthai: Moved the header into a `<caption>` tag and the differences were very minor! Pushed up refactored code to reflect this.

src/components/Office/AllowancesTable.jsx
```diff
@@ -0,0 +1,65 @@
...
+    branch: PropTypes.string,
+    rank: PropTypes.string,
+    weightAllowance: PropTypes.number,
+    authorizedWeight: PropTypes.number,
+    progear: PropTypes.number,
+    spouseProgear: PropTypes.number,
+    storageInTransit: PropTypes.number,
+    dependents: PropTypes.bool,
+  }).isRequired,
+};
```
suzubara, line 63: mostly just musing, but I'm wondering why we don't pass in each of these as individual props, instead of all together as one object? it looks like there are a few similar components that follow this pattern, which maybe could be refactored into a more generic table UI component that reduces some of the markup duplication. not a blocker though, just thinking about the future :)

src/pages/TOO/moveDetails.jsx
donaldthai: Quick update, looks like `weightAllowance` is the `totalWeight` from entitlements according to Alexi.

src/components/Office/AllowancesTable.jsx
donaldthai: I think that was just pure laziness on my end to avoid writing out the variable names again in the destructuring statement that the airbnb eslint complained about. The linter wanted to destructure all the props.

donaldthai: Definitely a rule we can relax if that makes sense to do so. I'm sure @gidjin has some opinions on this.

gidjin: I'm all for any refactoring reusable portions. I do feel this kind of defeats the reason that airbnb requires destructuring the props, but I'm totally open to adjusting our lint rules. My main goal with the lint rules is code consistency, clarity, and avoiding common problems.

PR #4212, opened 2020-06-08
couple small README tweaks

sojeri:
## Description

This makes a couple small changes to fix issues I ran into while setting up my local env:
- dd907ab adds some missing dashes to CLI instructions
- bbae8e0 adds a note encouraging folks to use .envrc.local install instructions b/c the chamber route is a little broken right now & TBH I didn't want to troubleshoot when the other was already working >_>
- f3b1ce8 adds some troubleshooting notes about the new global node requirement for pre-commit

## Reviewer Notes

I am not attached to bbae8e0.

Also my pre commit hooks are not yet working, so there may be another commit or two if CI catches some pre commit opinions.

This is a docs only change w/ no attached story.

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,20 +378,20 @@ This background job is built as a separate binary which can be built using `make
 When creating new features, it is helpful to have sample data for the feature to interact with. The TSP Award Queue is an example of that--it matches shipments to TSPs, and it's hard to tell if it's working without some shipments and TSPs in the database!
 
 * `make bin/generate-test-data` will build the fake data generator binary
-* `bin/generate-test-data -named-scenario="e2e_basic"` will populate the database with a handful of users in various stages of progress along the flow. The emails are named accordingly (see [`e2ebasic.go`](https://github.com/transcom/mymove/blob/master/pkg/testdatagen/scenario/e2ebasic.go)). Alternatively, run `make db_dev_e2e_populate` to reset your db and populate it with e2e user flow cases.
-* `bin/generate-test-data` will run binary and create a preconfigured set of test data. To determine the data scenario you'd like to use, check out scenarios in the `testdatagen` package. Each scenario contains a description of what data will be created when the scenario is run. Pass the scenario in as a flag to the generate-test-data function. A sample command: `./bin/generate-test-data -scenario=2`.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -266,7 +266,7 @@ Run `direnv allow` to load up the `.envrc` file. It should complain that you hav
 You can add a `.envrc.local` file. One way to do this is using the [chamber tool](https://github.com/segmentio/chamber) to read secrets from AWS vault.
 Then run `chamber env app-devlocal >> .envrc.local`. If you don't have access to chamber you can also `touch .envrc.local` and add any values that the output from direnv asks you to define. Instructions are in the error messages.
 
-If you wish to not maintain a `.envrc.local` you can alternatively run `cp .envrc.chamber.template .envrc.chamber` to enable getting secret values from `chamber`. **Note** that this method does not work for users of the `fish` shell unless you replace `direnv allow` with `direnv export fish | source`.
+If you wish to not maintain a `.envrc.local` you can alternatively run `cp .envrc.chamber.template .envrc.chamber` to enable getting secret values from `chamber`. **Note** that this method does not work for users of the `fish` shell unless you replace `direnv allow` with `direnv export fish | source`. **Note also** that as of 2020/6 this method appears to be somewhat broken. If you do not wish to troubleshoot it, the best way forward is to follow the above `.envrc.local` instructions.
```
chrisgilmerproj, line 269: We should fix this instead of moving forward. Can you help me understand what's broken with chamber for you? Also, what is your `$AWS_PROFILE` set to?

chrisgilmerproj: @sojeri - the only thing that's holding me up here is figuring out what's wrong with `chamber` for you. I would really like to NOT add this note because `chamber` is a critical function of development and should always be working. I've worked really hard to ensure folks aren't creating their own `.envrc.local` with secrets in it because those files get stale and also create a possible security leak for developer secrets.

sojeri: so using chamber in .envrc.local works well for my machine and my internet connection. I am moving in a few weeks & may not need to use this pattern when I have better ISPs available.

I've seen a couple specific things:
- when trying to setup the .envrc.chamber flow, sometimes I have to direnv allow twice in order to get all the vars correctly pulled. I suspect this is related to poor internet connectivity, not a flaw in the chamber pattern.
- if I am using a terminal pattern that opens several tabs at once in the mymove directory, the processes seem to get stuck during direnv allow. terminal is much more likely to get confused here if I haven't recently opened the mymove directory and dropped an AWS token to unblock pulling chamber vars.

would you be more comfortable if I instead added a note to the effect that the .envrc.local pattern is not recommended but may be the only solution for folks with poor internet connections?

chrisgilmerproj: Yeah, that sounds acceptable. We want to encourage folks to use the chamber version as much as possible and to fix it if its broken. I'm sorry for your situation and if you need to pair sometime I or anyone in InfraSec would be happy to pair.

PR #4221, opened 2020-06-10
MB-2790 Fix date format inconsistency for createdAt and updatedAt

LeDeep:
## Description

We're currently using several different date formats in our yaml files for `createdAt` and `updatedAt`. These include `date`, `datetime`, and `date-time`. The swagger docs provide `date` and `date-time` as supported formats.

This PR fixes the inconsistency and uses `date-time` as the format. 

## Reviewer Notes

I plan to write up a separate ticket to see if we should be using `date-time` for all date fields. If that is the case then would be good to have an ADR that makes this clear.

## Setup

```sh
make server_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2790) for this change

### DIFF
--- a/data.json
+++ b/data.json
@@ -1,0 +1,7 @@
+{
+  "mtoShipmentID": "475579d5-aaa4-4755-8c43-c510381ff9b5",
+  "ifMatch": "MjAyMC0wNi0xMFQyMDoxNjowMS4xMDQ4Mjda",
+  "body": {
+    "scheduledPickupDate": "2018-03-19"
+  }
+}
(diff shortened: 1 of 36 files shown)


data.json
```diff
@@ -0,0 +1,7 @@
+{
```
abbyoung, line 1: Did you accidentally commit the test json?

pkg/services/mto_shipment/mto_shipment_updater.go
```diff
@@ -67,10 +67,10 @@ func setNewShipmentFields(planner route.Planner, db *pop.Connection, oldShipment
 		oldShipment.ActualPickupDate = updatedShipment.ActualPickupDate
 	}
 
-	scheduledPickupTime := *oldShipment.ScheduledPickupDate
+	//scheduledPickupTime := *oldShipment.ScheduledPickupDate
```
abbyoung, line 70: Looks like you made a fix relevant to my bugfix for mb-2811 😄 i will wait for this to merge and then retest on my end + add in the unit test.

PR #4222, opened 2020-06-11
[MB-2436] Fix malformed errors sent to prime api client

shimonacarvalho:
## Description

When using prime api client we used to get a lot of unmarshalling errors due to the response not containing valid json. This PR is meant to cleanup some of these cases. In the course of searching for these I found that if we have a panic we also return a null, when we could return a well formed error with trace id, to help debug the error with logs. 

So this ticket does 2 things
- Changes recoveryLogger to include a well formed response with traceID (this only gets returned on a panic)
- Changes prime and support InternalServerError format to match the prescribed title/detail/instance.

**AC from JIRA**
Find every instance in prime and support handlers where we are returning a string instead of json, and return message in json. Involves looking at every single handler and every error we return.

## Setup

It's a bit hard to test since you need an internal server error to happen. One known one right now is if you try to update a shipment that has no current scheduledPickupDate, you should get a server error. You will want to see a response like this...
```
{
  "Payload": {
    "detail": "An unexpected server error has occurred.",
    "instance": "68d8784f-9f85-4443-9fe1-f6f2fd90c413",
    "title": "Internal Server Error"
  }
}
```

## References

* [MB-2436 Fix unmarshalling errors in Prime API client - Jira](https://dp3.atlassian.net/browse/MB-2436) for this change
* [handle backend errors · transcom/mymove Wiki](https://github.com/transcom/mymove/wiki/handle-backend-errors) explains more about the approach used.

[MB-2436]: https://dp3.atlassian.net/browse/MB-2436

### DIFF
--- a/pkg/middleware/recovery.go
+++ b/pkg/middleware/recovery.go
@@ -19,7 +19,6 @@ func Recovery(logger Logger) func(inner http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			defer func() {
 				if obj := recover(); obj != nil {
-					fmt.Println("\n\nwe got to recovery")
 
 					// Log the error and optionally the stacktrace
 					fields := []zap.Field{
--- a/pkg/middleware/recovery.go
+++ b/pkg/middleware/recovery.go
@@ -1,11 +1,15 @@
 package middleware
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"runtime/debug"
 
 	"go.uber.org/zap"
+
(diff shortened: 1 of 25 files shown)


pkg/middleware/recovery.go
```diff
@@ -1,11 +1,15 @@
 package middleware
```
garrettqmartin8, line 1: So this is basically recovering from any panic and returning a nicely formatted error with the trace id? Awesome!

shimonacarvalho: Yes it used to send a blank internal server error, but now we fill in the payload including traceID.

PR #4239, opened 2020-06-12
MB-299 Migrate orders to using entities

amitch23:
## Description

Migrates redux pattern to using entities for **orders** and **uploads** for the PPM service member signup flow. 
This ticket was originally intended to only migrate the createOrders API func to entities, but as I started the work, I found that related objects needed to be loaded onto entities, i.e. documents and uploads. So, I added the modules funcs to documents to create and delete uploads.

Note:
* There is now a separate Uploader, called "OrdersUploader" to use with the swaggerRequest call & payload for adding and deleting uploads to integrate with filepond. This could potentially be combined back into the Uploader, since most of the code is repeated, but it may be more desirable to have a clearer separation between the components that populate entities with normalized data vs. not (i.e. the "old" model).
* The API endpoints that fetch orders for a logged in user are erroring in master due to a couple most-likely related bugs. These should be fixed with this PR for the SM flow purposes, but since these API changes are used elsewhere they should be tested with other apps i.e. Office, TOO, etc.. (see https://dp3.atlassian.net/browse/MB-2951 and https://dp3.atlassian.net/browse/MB-3007)
* The swaggerRequest has been adapted to allow for deleting arrays from the state store, in addition to just removing a single entity.

## Setup

```sh
# populate with test users
make db_dev_reset (if necessary)
make db_dev_migrate (if necessary)
bin/generate-test-data --named-scenario="e2e_basic"

make server_run
make client_run
```

Since this affects the entire SM and PPM sign up flow, the following cases should be tested:

Note: You should expect to see the redux state populate orders, uploads, documents, and moves @ `entities` rather than in their respective, current areas. I.e. when uploads are existent, they should appear @ `entities.uploads` rather than `orders.currentOrders.uploads`. (Note that `orders.CurrentOrders` is continuing to be populated via the getLoggedInUser endpoint as well, but only used and referenced in the entities.)

1. Sign up as a new service member in the local sign in flow and go through to the Review page. Click on Edit Orders and interact with the uploader and Uploadtable. 
2. Sign in as a service member and enter the PPM flow at various points. Log out, log back in, refresh, etc.
3. Sign in as a PPM office user, find the move and cancel it. Sign back in as a Service Member - You should see "start new move" or "your move has been canceled"

 have been satisfied.
* [ x] Request review from a member of a different team.
* [x ] Have the Jira acceptance criteria been met for this change?

## References
Jira Stories:
* https://dp3.atlassian.net/browse/MB-298
* https://dp3.atlassian.net/browse/MB-299
* https://dp3.atlassian.net/browse/MB-300
* https://dp3.atlassian.net/browse/MB-3007
* https://dp3.atlassian.net/browse/MB-2951


README.md
```diff
@@ -379,7 +379,7 @@ This background job is built as a separate binary which can be built using `make
 When creating new features, it is helpful to have sample data for the feature to interact with. The TSP Award Queue is an example of that--it matches shipments to TSPs, and it's hard to tell if it's working without some shipments and TSPs in the database!
 
 * `make bin/generate-test-data` will build the fake data generator binary
-* `bin/generate-test-data -named-scenario="e2e_basic"` will populate the database with a handful of users in various stages of progress along the flow. The emails are named accordingly (see [`e2ebasic.go`](https://github.com/transcom/mymove/blob/master/pkg/testdatagen/scenario/e2ebasic.go)). Alternatively, run `make db_dev_e2e_populate` to reset your db and populate it with e2e user flow cases.
+* `bin/generate-test-data --named-scenario="e2e_basic"` will populate the database with a handful of users in various stages of progress along the flow. The emails are named accordingly (see [`e2ebasic.go`](https://github.com/transcom/mymove/blob/master/pkg/testdatagen/scenario/e2ebasic.go)). Alternatively, run `make db_dev_e2e_populate` to reset your db and populate it with e2e user flow cases.
```
amitch23, line 382: testing comments

src/shared/Entities/modules/documents.js
```diff
@@ -1,8 +1,12 @@
-import { documentModel } from '../schema';
+import { documents } from '../schema';
 import { ADD_ENTITIES } from '../actions';
 import { denormalize } from 'normalizr';
+import { getClient } from 'shared/Swagger/api';
+import { swaggerRequest } from 'shared/Swagger/request';
 
 export const STATE_KEY = 'documents';
+export const createUploadLabel = 'documents.createUpload';
+export const deleteUploadLabel = 'documents.deleteUpload';
```
amitch23, line 10: add for deleteuploads

src/scenes/Review/EditOrders.jsx
```diff
@@ -142,38 +139,38 @@ class EditOrders extends Component {
...
-        if (!this.props.hasSubmitError) {
-          this.props.editSuccessful();
-          this.props.history.goBack();
-          if (this.props.isPpm) {
-            this.props.checkEntitlement(this.props.match.params.moveId);
-          }
-        } else {
-          scrollToTop();
+    if (this.props.currentOrders && !isEmpty(this.state.deleteQueue)) {
+      this.props.deleteUploads(this.state.deleteQueue);
```
amitch23, line 149: @suzubara - there's an issue with this deleteUploads call, where if you delete more than 1 upload from the UploadsTable, it successfully removes them from the entities, but if you delete only 1, it disappears, then reappears. I suspect it has to do with the orders being loaded on the Review page before the swaggerRequest has completed, but I'm not sure. For both cases, the uploads are successfully deleted in the DB. It might be related to how options.deleteIds are being handled and normalized in the swaggerRequest, OR, it's related to this.props.deleteUploads not being set as a condition to the Promise below (which I tried to do but got lots of wild errors)

src/shared/Entities/modules/orders.js
```diff
@@ -49,12 +66,54 @@ export function selectOrdersForMove(state, moveId) {
...
+    return {};
+  }
+  let activeOrders = fetchActive(selectOrdersForServiceMemberId(state, serviceMember.id));
+  if (isEmpty(activeOrders)) {
+    activeOrders = fetchActive(get(state, 'user.userInfo.service_member.orders', {}));
+  }
+  return activeOrders || {};
+}
+
+export function selectActiveOrLatestOrders(state) {
```
amitch23, line 106: Check if we can use just this func

src/shared/Swagger/request.js
```diff
@@ -134,9 +134,19 @@ export function swaggerRequest(getClient, operationPath, params, options = {}) {
         if (options.deleteId) {
           // eslint-disable-next-line security/detect-object-injection
           var oldEntity = state.entities[schemaKey][options.deleteId];
-          action.entities = normalizePayload([oldEntity], payloadSchema).entities;
+          action.entities = normalize([oldEntity], payloadSchema).entities;
+        } else if (options.deleteIds) {
+          var oldEntities = state.entities[String(schemaKey)];
```
amitch23, line 139: Use filter on oldEntites (make array from keys)

src/scenes/Orders/Orders.jsx
```diff
@@ -112,8 +116,9 @@ function mapStateToProps(state) {
 function mapDispatchToProps(dispatch) {
   return bindActionCreators(
     {
-      updateOrders,
-      createOrders,
+      fetchLatestOrders,
+      update: updateOrders,
```
suzubara, line 120: nitpick: what was the reason for changing the names of these 2 props? personally I preferred `updateOrders, createOrders` just to avoid being too generic. `this.props.update, this.props.create` could be more difficult to keep track of later down the road, i.e., in a search.

src/scenes/Orders/TransitionToMove.jsx
```diff
@@ -39,15 +42,18 @@ export class TransitionToMove extends Component {
 }
 
 function mapStateToProps(state) {
-  const move = get(state, 'moves.currentMove');
+  const serviceMemberId = get(state, 'serviceMember.currentServiceMember.id');
+  const move = selectActiveOrLatestMove(state);
+
   const props = {
+    serviceMemberId: serviceMemberId,
```
suzubara, line 49: minor syntax sugar: if the key name and the variable are the same, you can use shorthand like so
```
const props = {
  serviceMemberId,
  ...
```

src/shared/Uploader/OrdersUploader.jsx
```diff
@@ -0,0 +1,200 @@
+// eslint-disable-next-line no-unused-vars
```
suzubara, line 1: if it's not too much of a burden, could we add this file under `src/components/OrdersUploader`? so that it is clear it's a "new" version and also it will be linted under the stricter set of rules.

src/scenes/Review/EditOrders.jsx
amitch23: Thanks for looking into this - I ended up changing the uploadsTable to behave like the UploadOrders page and actually delete each upload when clicking "delete". This is simpler, more consistent, reduces complexity and risk - so I just removed related handling for batch deleting in the swagger request.

src/scenes/Orders/Orders.jsx
amitch23: I adapted it to match how it was when I found it, but agree that it makes it harder to track/search. Changed to your suggestion and to match established pattern for action names

src/scenes/Orders/TransitionToMove.jsx
amitch23: Changed to the shorthand

amitch23:
@kimallen responses inline:

"in Redux, under requests.req_id (or maybe it was requests.lastErrors?, I expect to see a key req_# but see Orders.showServiceMemberOrders` as the key.
requests_key_name"

I'm not able to recreate this - maybe something that I changed fixed this? This is what I see for showServiceMemberOrders request:
Screen Shot 2020-07-06 at 1 47 42 PM

"Perhaps something about error creation with showing orders is off? I think the way to recreate it is to start a new SM and check when you get to the first orders page.
what is uploaded_orders on the entities orders? I see a key there even before an order is uploaded. It also doesn't ever change as uploads are added."

Uploaded_orders is the "user_uploads" reference, AKA the document that is created when the orders are created, and associated to the orders. When the user adds uploads, the uploads are associated to this "user uploads" document. It's curious that when the orders are created, not only is the move created, but also the associated document before the orders uploads are uploaded, but this is how it was when I started this ticket.

"Should the status of the orders change (in redux entities) after the Office approves basics? I see the status of move and ppm change, but orders still has status DRAFT (in the office app). ( don't know current behavior)"

I don't believe so. If it should, I think it should be handled in a separate ticket, because I don't believe this work touches how an orders status changes when then office approves basics. I would assume that this is current behavior now.

src/shared/Uploader/OrdersUploader.jsx
amitch23: @suzubara I'd like to try and get this merged hopefully by EOD, since it's carried over from a few sprints ago. When I moved this into `src/components/OrdersUploader` there are a ton of errors that lead me to believe that this won't be an easy lift, and I also haven't worked with css modules before (which it looks like the components there use). Would you be OK with this being a separate ticket from this task? Or, if you have time/bandwidth to pair on this today and understand it to be more straightforward, that would work too!

src/scenes/Review/EditOrders.jsx
suzubara: @amitch23 agreed, good idea!!

src/shared/Uploader/OrdersUploader.jsx
amitch23: resolved through pairing on ESLint errors

PR #4247, opened 2020-06-15
[MB-2875] Fetch All MTOs service object

Ronolibert:
## Description

Extract querying of all MTOs into service object method for better reuse

## Reviewer Notes
- Wasn't sure of a way only using the eager query once but conditionally adding the Where query if isAvailableToPrime was true, if there even is a way

## Setup
`make server_run`
Use fetch all MTO updates primeapi endpoint

## References
* [Jira story](https://dp3.atlassian.net/browse/MB-2492) for this change

### DIFF
--- a/pkg/handlers/primeapi/move_task_order.go
+++ b/pkg/handlers/primeapi/move_task_order.go
@@ -1,50 +1,28 @@
 package primeapi
 
 import (
-	"time"
-
 	"github.com/transcom/mymove/pkg/services"
 
 	"github.com/go-openapi/runtime/middleware"
 	"github.com/gofrs/uuid"
 	"go.uber.org/zap"
 
-	"github.com/transcom/mymove/pkg/handlers/primeapi/internal/payloads"
-	"github.com/transcom/mymove/pkg/models"
-
 	movetaskorderops "github.com/transcom/mymove/pkg/gen/primeapi/primeoperations/move_task_order"
 	"github.com/transcom/mymove/pkg/handlers"
+	"github.com/transcom/mymove/pkg/handlers/primeapi/internal/payloads"
 )
 
 // FetchMTOUpdatesHandler lists move task orders with the option to filter since a particular date
 type FetchMTOUpdatesHandler struct {
 	handlers.HandlerContext
+	services.MoveTaskOrderFetcher
 }
 
 // Handle fetches all move task orders with the option to filter since a particular date
 func (h FetchMTOUpdatesHandler) Handle(params movetaskorderops.FetchMTOUpdatesParams) middleware.Responder {
 	logger := h.LoggerFromRequest(params.HTTPRequest)
 
-	var mtos models.MoveTaskOrders
-
-	query := h.DB().Where("available_to_prime_at IS NOT NULL").Eager(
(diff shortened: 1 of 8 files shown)


pkg/services/move_task_order/move_task_order_fetcher.go
```diff
@@ -28,6 +29,60 @@ func (f moveTaskOrderFetcher) ListMoveTaskOrders(moveOrderID uuid.UUID) ([]model
...
+			"MTOShipments.PickupAddress",
+			"MTOShipments.SecondaryDeliveryAddress",
+			"MTOShipments.SecondaryPickupAddress",
+			"MTOShipments.MTOAgents",
+			"MoveOrder.Customer",
+			"MoveOrder.Entitlement")
+
+		if since != nil {
+			since := time.Unix(*since, 0)
+			query = query.Where("updated_at > ?", since)
```
shimonacarvalho, line 52: This looks like a conditionally added where query? Could something similar work for the availableToPrimeAt date?

Ronolibert: Yeah I'm not sure why it doesn't work. If I try:
```
query := f.db.Eager(
		"PaymentRequests.PaymentServiceItems.PaymentServiceItemParams.ServiceItemParamKey",
		"MTOServiceItems.ReService",
		"MTOServiceItems.Dimensions",
		"MTOServiceItems.CustomerContacts",
		"MTOShipments.DestinationAddress",
		"MTOShipments.PickupAddress",
		"MTOShipments.SecondaryDeliveryAddress",
		"MTOShipments.SecondaryPickupAddress",
		"MTOShipments.MTOAgents",
		"MoveOrder.Customer",
		"MoveOrder.Entitlement")
	
	if isAvailableToPrime {
		query = query.Where("available_to_prime_at IS NOT NULL")

		if since != nil {
			since := time.Unix(*since, 0)
			query = query.Where("updated_at > ?", since)
		}

		err = query.All(&moveTaskOrders)
	}
```
Then I get error: `Cannot use 'query.Where("available_to_prime_at IS NOT NULL")' (type *Query) as type *Connection in assignment `

pkg/handlers/primeapi/move_task_order.go
```diff
@@ -1,50 +1,28 @@
...
-		"MTOShipments.MTOAgents",
-		"MoveOrder.Customer",
-		"MoveOrder.Entitlement")
-	if params.Since != nil {
-		since := time.Unix(*params.Since, 0)
-		query = query.Where("updated_at > ?", since)
-	}
-
-	err := query.All(&mtos)
+	mtos, err := h.MoveTaskOrderFetcher.ListAllMoveTaskOrders(true, params.Since)
```
Ronolibert, line 25: Also not really sure why this is throwing a nil pointer dereference

pkg/services/move_task_order/move_task_order_fetcher.go
Ronolibert: @shimonacarvalho Fixed it! Had to convert the eager call to a query using `Q()` for it to work that way

shimonacarvalho: Oh great!

PR #4249, opened 2020-06-15
add a GEX task: MB-104

lynzt:
## Description

Allow a GEX command to be run/scheduled via milmove-task

## Reviewer Notes

If you run this, it throws a `x509: certificate signed by unknown authority`, but we think that's due to a cert error from the GEX side which is being looked into.

This is also quick check to see if we can run this in prod and hit GEX to rule out IP issues w/ the connection.

I am ignoring the `transaction-name` as we don't need it for this test, but leaving it in the code as we will use it in the future.

## Setup

to run: `go run ./cmd/milmove-tasks post-file-to-gex --edi ./cmd/send_to_gex/helloworld.json --transaction-name 'transactionName' --gex-url 'gexURL'`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-104) for this change

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -762,6 +762,21 @@ tasks_send_payment_reminder: tasks_build_linux_docker ## Run send-payment-remind
 		--rm \
 		$(TASKS_DOCKER_CONTAINER):latest \
 		milmove-tasks send-payment-reminder
+
+tasks_post_file_to_gex: tasks_build_linux_docker ## Run post-file-to-gex from inside docker container
+	@echo "sending payment reminder with docker command..."
+	DB_NAME=$(DB_NAME_DEV) DB_DOCKER_CONTAINER=$(DB_DOCKER_CONTAINER_DEV) scripts/wait-for-db-docker
+	docker run \
+		-t \
+		-e DB_HOST="database" \
+		-e DB_NAME \
+		-e DB_PORT \
+		-e DB_USER \
+		-e DB_PASSWORD \
+		--link="$(DB_DOCKER_CONTAINER_DEV):database" \
+		--rm \
+		$(TASKS_DOCKER_CONTAINER):latest \
+		milmove-tasks post-file-to-gex
 #
 # ----- END SCHEDULED TASK TARGETS -----
 #
--- a/pkg/cli/gex.go
+++ b/pkg/cli/gex.go
@@ -2,7 +2,6 @@ package cli
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -762,6 +762,21 @@ tasks_send_payment_reminder: tasks_build_linux_docker ## Run send-payment-remind
 		--rm \
 		$(TASKS_DOCKER_CONTAINER):latest \
 		milmove-tasks send-payment-reminder
+
+tasks_post_file_to_gex: tasks_build_linux_docker ## Run post-file-to-gex from inside docker container
+	@echo "sending payment reminder with docker command..."
+	DB_NAME=$(DB_NAME_DEV) DB_DOCKER_CONTAINER=$(DB_DOCKER_CONTAINER_DEV) scripts/wait-for-db-docker
+	docker run \
+		-t \
+		-e DB_HOST="database" \
```
esacteksab, line 771: Is all this database stuff boilerplate or copy & paste? I'm wondering why all the DB connection stuff is necessary to do an HTTP post to an external endpoint.

pkg/cli/gex.go
```diff
@@ -29,9 +30,9 @@ func InitGEXFlags(flag *pflag.FlagSet) {
 // CheckGEX validates GEX command line flags
 func CheckGEX(v *viper.Viper) error {
 	gexURL := v.GetString(GEXURLFlag)
-	if len(gexURL) > 0 && gexURL != "https://gexweba.daas.dla.mil/msg_data/submit/" {
+	if len(gexURL) > 0 && strings.HasPrefix(gexURL, "https://gexweba.daas.dla.mil/msg_data/submit") {
```
esacteksab, line 33: I'm left wondering if this URL could be an environment variable, as to 1) not hard code an address in the source code and 2) to change this URL should it change w/o having to do a code deploy?

lynzt: this is just a url check, and I've now disabled this check as it was failing the `acceptance_tests`... will go back to see why.
Wondering if we even need this check or if `len(gexURL) > 0` is enough of a check...

pkg/cli/gex.go
```diff
@@ -30,9 +30,9 @@ func InitGEXFlags(flag *pflag.FlagSet) {
 // CheckGEX validates GEX command line flags
 func CheckGEX(v *viper.Viper) error {
 	gexURL := v.GetString(GEXURLFlag)
-	if len(gexURL) > 0 && strings.HasPrefix(gexURL, "https://gexweba.daas.dla.mil/msg_data/submit/") {
+	if len(gexURL) > 0 && strings.HasPrefix(gexURL, "https://gexweba.daas.dla.mil/msg_data/submit") {
```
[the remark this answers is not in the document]
lynzt: this is just a check to see that the base url matches what we enter... and I've commented this check out as i don't think we need it.

Makefile
lynzt: I left this in as we will eventually be using a DB connection to extract data and send to GEX...

pkg/cli/gex.go
esacteksab: `len(gexURL) > 0` would simply confirm if a string (URL) existed. I think the intent here is to verify that we are in fact going to the URL we intend to go to. At least, that's my guess. I could very well be wrong here.

pkg/cli/gex.go
```diff
@@ -29,10 +29,10 @@ func InitGEXFlags(flag *pflag.FlagSet) {
...
 func CheckGEX(v *viper.Viper) error {
 	gexURL := v.GetString(GEXURLFlag)
-	if len(gexURL) > 0 && gexURL != "https://gexweba.daas.dla.mil/msg_data/submit/" {
-		return fmt.Errorf("invalid gexUrl %s, expecting "+
-			"https://gexweba.daas.dla.mil/msg_data/submit/ or an empty string", gexURL)
-	}
+	// if len(gexURL) > 0 && strings.HasPrefix(gexURL, "https://gexweba.daas.dla.mil/msg_data/submit") {
+	// 	return fmt.Errorf("invalid gexUrl %s, expecting "+
+	// 		"https://gexweba.daas.dla.mil/msg_data/submit or an empty string", gexURL)
+	// }
```
chrisgilmerproj, line 35: Could we have updated this to allow any of a set of different URLs instead of commenting this out?

pkg/cli/gex.go
chrisgilmerproj: Yep, the intent is to ensure this tool can only be used with the known URL and not any others. Generally when we have multiple correct inputs we make a list of them and ensure the input exists in the list. The `pkg/cli/` code has several examples of this.

PR #4234, opened 2020-06-12
MB-1583 Implement Service Area Origin Lookup

noelledusahel:
## Description
Add functionality to lookup the ServiceAreaOrigin Service Item Param when a  payment request is made. 

## Reviewer Notes

Is there anything you would like reviewers to give additional scrutiny?

## Setup
- First run `make db_dev_e2e_populate` to ensure your database is setup with the E2E data.
- Next, save this example JSON payload to a file called `pr_payload.json`

```json
{
  "body": {
    "isFinal": false,
    "moveTaskOrderID": "9c7b255c-2981-4bf8-839f-61c7458e2b4d",
    "serviceItems": [
      {
        "id": "76055c99-0990-410c-a7c9-69373b0b53eb"
      }
    ]
  }
}
```

-  Ensure your server is running: `make server_run`

- Create a payment request using that payload: `go run ./cmd/prime-api-client --insecure create-payment-request --filename pr_payload.json | jq`

- Look at the DB and verify that the appropriate records appear in the payment_service_item_params table.

## Code Review Verification Steps

* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/wiki/backend#logging)
* [ ] The requirements listed in
 [Querying the Database Safely](https://github.com/transcom/mymove/wiki/backend#querying-the-database-safely)
 have been satisfied.
* Any new migrations/schema changes:
  * [ ] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://github.com/transcom/mymove/wiki/migrate-the-database#zero-downtime-migrations))
  * [ ] Have been communicated to #g-database
  * [ ] Secure migrations have been tested following the instructions in our [docs](https://github.com/transcom/mymove/wiki/migrate-the-database#secure-migrations)
* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](tbd) for this change
* [this article](tbd) explains more about the approach used.

## Screenshots

If this PR makes visible UI changes, an image of the finished UI can help reviewers and casual
observers understand the context of the changes. A before image is optional and
[...]

PR #4250, opened 2020-06-15
[MB-486] Implement management and counseling pricing

reggieriser:
## Description

This PR adds support for pricing of management (MS) and counseling (CS) service items.

## Reviewer Notes

This is the first pricing PR we've done since accepting payment requests/parameters/etc.  There are a handful of pricers we wrote late last year that pre-date that, but I expect we'll be refactoring those as we go forward.  Because this is the first new pricer, this PR has some additional plumbing work to support this pricer and others yet to be implemented.

I added `ContractCode` as a required parameter for MS/CS service codes.  I anticipate most of the other service codes will need this too, but I will leave that decision for those pricing stories.  I also added a placeholder contract code lookup for now that just expects the contract code to be `TRUSS_TEST`.

You'll notice that the service item pricer service object is wired into the create payment request endpoint, but that creates an interesting problem given the use of a transaction within the payment request creator.  I added a `UsingDB` method to the service item pricer to allow the transactional context to be set, but it feels a little strange.  Open to ideas to improve this.

In the design doc, I posed a question about additional "parameters" that may result from the rate engine lookup.  There's not a rate in the pricing of MS and CS -- it's a flat price given a contract year -- but we may need to think more about storing these with future pricers (e.g., a domestic linehaul may have a base rate and an escalation factor we may want to capture).

## Setup

- `make db_dev_e2e_populate`
- ~~Since we don't have any pricing data already in our migrations, you'll need to add some locally to test this:~~
    - ~~`ghc-pricing-parser --filename pkg/parser/pricing/fixtures/pricing_template_2019-09-19_fake-data.xlsx --display --contract-code=TRUSS_TEST`~~
    - ~~To save time for future runs, consider using the recently introduced `pricing-import` script to create a migration from this pricing data.  If you've already done this, you can do a `psql-dev < tmp/20200609194054_import_pricing_data_fake.up.sql` (replace the filename with your migration name).~~
- Now that PR #4255 has been merged, we have fake pricing data in our migrations.
- `make server_run`
- Make a `pr_payload.json` file containing the following (which references E2E data):
    ```
    {
      "body": {
        "isFinal": false,
        "serviceItems": [
          {
            "id": "76055c99-0990-410c-a7c9-69373b0b53eb"
          },
          {
            "id": "05fbfad0-731f-4342-87e9-abe55566bb63"
          }
        ]
      }
    }
    ```
- `go run ./cmd/prime-api-client --insecure create-payment-request --filename pr_payload.json | jq`
- Verify parameters and that the service items have been priced.  Compare accuracy with pricing data found in the pricing template.

* [ ] The requirements listed in [Querying the Database Safely](https://github.com/transcom/mymove/wiki/backend#querying-the-database-safely) have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-486) for this change
* [Pricing service items design](https://docs.google.com/document/d/1KBQWC1bSUC6S69w-1fcdDpPId4tEiTZDQZfTTBNieVU/edit); note that some implementation details may have changed a bit, but the general idea is the same


pkg/models/re_service.go
```diff
@@ -9,9 +9,6 @@ import (
 	"github.com/gofrs/uuid"
 )
 
-// ReServiceName is a type of service name
-type ReServiceName string
```
reggieriser, line 13: This seems to be a leftover.  No usages of this anywhere, so I removed it as it was confusing in auto-complete.

pkg/services/errors.go
```diff
@@ -176,3 +176,19 @@ func NewConflictError(id uuid.UUID, message string) ConflictError {
 		message: message,
 	}
 }
+
+// NotImplementedError is returned when some functionality is not yet implemented
+type NotImplementedError struct {
```
garrettqmartin8, line 181: Nice addition.

pkg/services/ghc_rate_engine.go
```diff
@@ -3,11 +3,42 @@ package services
...
 	"github.com/transcom/mymove/pkg/route"
 
 	"github.com/transcom/mymove/pkg/unit"
 )
 
+// ServiceItemPricer prices a generic payment service item for a GHC move
+//go:generate mockery -name ServiceItemPricer
+type ServiceItemPricer interface {
+	PriceServiceItem(item models.PaymentServiceItem) (unit.Cents, error)
+	UsingDB(db *pop.Connection) ServiceItemPricer
```
garrettqmartin8, line 18: Is `UsingDB` the best name for this? I think what it's actually doing is `UsingTransaction` or something like that.

pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go
```diff
@@ -64,6 +57,7 @@ func ServiceParamLookupInitialize(
 	s.lookups[models.ServiceItemParamNameZipPickupAddress.String()] = ZipPickupAddressLookup{}
 	s.lookups[models.ServiceItemParamNameZipDestAddress.String()] = ZipDestAddressLookup{}
 	s.lookups[models.ServiceItemParamNameMTOAvailableToPrimeAt.String()] = MTOAvailableToPrimeAtLookup{}
+	s.lookups[models.ServiceItemParamNameContractCode.String()] = ContractCodeLookup{}
```
garrettqmartin8, line 60: Similar to my [comment](https://github.com/transcom/mymove/pull/4234#discussion_r440303142) on @noelledusahel's PR, I think we can get away with avoiding structs here and just pass functions. Doesn't need to be addressed in this PR, but I may look into this myself soon.

pkg/services/ghc_rate_engine.go
reggieriser: I initially had it as `UsingTransaction`, but then I thought that may be too specific because the DB connection passed in didn't necessarily have to be one with a transaction.  Thinking about it more, I probably should have named it `UsingConnection` instead.  I think I'll change it to that, but I'm willing to consider `UsingTransaction` if you still feel strongly about it.

pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go
reggieriser: It's an interesting point.  I think the original motivation was that the lookups could have their own (perhaps different) set of struct fields but still share a common `lookup` interface.  If that's not needed, we could change this to be a static map of service codes -> function type.  Interested in @jacquelineIO 's take as I believe she authored the PR that introduced this.

pkg/services/ghcrateengine/task_order_services_helper.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"time"
+
+	"github.com/gobuffalo/pop"
+
+	"github.com/transcom/mymove/pkg/models"
+)
+
+func fetchTaskOrderFee(db *pop.Connection, contractCode string, serviceCode models.ReServiceCode, mtoAvailableToPrimeAt time.Time) (models.ReTaskOrderFee, error) {
```
gidjin, line 11: A bit on the nitpicky side, but I like having unit tests for everything. But since this is tested elsewhere by being used I'll leave it up to you if you want to add one or two tests for this helper.

reggieriser: I go back and forth with whether to test unexported functions like this, particularly if they're exercised by tests of other exported functions/methods as these are.  But it was pretty easy to add a couple of tests here, so I went ahead and did that.

pkg/services/ghc_rate_engine.go
garrettqmartin8: `UsingConnection` makes sense to me, and it's actually describing the type of the parameter.

reggieriser:
Changed contract code to `TRUSS_TEST` to match what's in PR #4255 (which needs to get merged before this PR).

PR #4251, opened 2020-06-15
Change availableToPrime to availableToPrimeAt

Ronolibert:
## Description

Change availableToPrime to be availableToPrimeAt

### DIFF
--- a/pkg/services/support/move_task_order/move_task_order_creator.go
+++ b/pkg/services/support/move_task_order/move_task_order_creator.go
@@ -268,8 +268,9 @@ func MoveTaskOrderModel(mtoPayload *supportmessages.MoveTaskOrder) *models.MoveT
 		ContractorID:       uuid.FromStringOrNil(mtoPayload.ContractorID.String()),
 	}
 
-	if mtoPayload.IsAvailableToPrime != nil {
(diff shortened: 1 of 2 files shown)


pkg/services/support/move_task_order/move_task_order_creator.go
```diff
@@ -268,8 +268,9 @@ func MoveTaskOrderModel(mtoPayload *supportmessages.MoveTaskOrder) *models.MoveT
 		ContractorID:       uuid.FromStringOrNil(mtoPayload.ContractorID.String()),
 	}
 
-	if mtoPayload.IsAvailableToPrime != nil {
-		model.IsAvailableToPrime = *mtoPayload.IsAvailableToPrime
+	if mtoPayload.AvailableToPrimeAt != nil {
+		availableToPrimeAt := time.Time(*mtoPayload.AvailableToPrimeAt)
+		model.AvailableToPrimeAt = &availableToPrimeAt
 	}
```
gidjin, line 274: Are there not tests for this because it's a `support` endpoint?

Ronolibert: I think I just forgot to write tests for the service object because there's tests for the handler, I can add tests for this service as a chore later today

PR #4114, opened 2020-05-19
MB-2262 create new migration for pricing imports

noelledusahel:
## Description
- We need a way to get GHC pricing data in to production
- Write a script to Initiate a pg_dump of re_* tables (excluding re_services) after the pricing parser is complete
- create a secure migration that will make data available in staging/experimental/production

## Reviewer Notes

Creating a secure migration for pricing data is going to be handled by a developer for now. 
Are the instructions clear for you to follow?

## Setup
**Step 1:  Run the Pricing Parser to populate data to `re_` tables**
```
1. make db_dev_reset db_dev_migrate
2. If your rate engine tables currently have records we want to clear them.
      Open psql-dev and run the following query: 
      TRUNCATE re_contracts, re_contract_years CASCADE;
3. rm -f bin/ghc-pricing-parser && make bin/ghc-pricing-parser
4. ghc-pricing-parser --filename pkg/parser/pricing/fixtures/pricing_template_2019-09-19_fake-data.xlsx --display --contract-code=12346
```

**Step 2: Run the pricing-import script**
Run the script and follow the prompts: 
```
/bin/sh pricing-import
``` 

## Code Review Verification Steps

* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/tree/master/docs/backend.md#logging)
* [ ] The requirements listed in
 [Querying the Database Safely](https://github.com/transcom/mymove/tree/master/docs/backend.md#querying-the-database-safely)
 have been satisfied.
* Any new migrations/schema changes:
  * [ ] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://github.com/transcom/mymove/tree/master/docs/database.md#zero-downtime-migrations))
  * [ ] Have been communicated to #dp3-engineering
  * [ ] Secure migrations have been tested following the instructions in our [docs](https://github.com/transcom/mymove/blob/master/docs/database/migrate-the-database.md#secure-migrations)
* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

[...]

PR #4255, opened 2020-06-15
MB-2799 Import fake pricing data to staging/experimental

noelledusahel:
## Description

Follows steps outlined in https://dp3.atlassian.net/browse/MB-2262 to import and upload fake pricing data via secure migration

- WIll not be deployed to production

- Deployed to Staging & Experimental only

 Steps for deploying to experimental:
 - https://github.com/transcom/mymove/wiki/deploy-to-experimental

- https://github.com/transcom/mymove/wiki/migrate-the-database#secure-migrations-for-one-environment

## Reviewer Notes
- follow instructions in https://github.com/transcom/mymove/pull/4114
- `aws-vault login $AWS_PROFILE`
- open `S3`

Look for `20200615192021_import_pricing_data_june_2020.up.sql` in the following folders
- `transcom-ppp-app-experimental-us-west-2/secure-migrations`
- `transcom-ppp-app-staging-us-west-2/secure-migrations`
- `transcom-ppp-app-prod-us-west-2/secure-migrations` (this file should only contain a comment, no sql)

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2799)

## Screenshots


noelledusahel:
> It looks like the contract code is set to `Pricing`. Is that what we want for the test data? I'm thinking we may need to make it more obvious this is test data rather than real pricing. Also, my PR #4250 assumed it was named `TRUSS` (which I just made up), so I'll need to change that accordingly to whatever the name is here.
> 
> Note that if we do decide to change the code, we'll need to restore the experimental DB to make the migration run again there (infra can help us with that -- we've done it several times before).

OK i definitely missed that the code 'Pricing' stuck, i meant for it to be 'june_2020' @jacquelineIO  do you have any guidance on what Contract Code we want attached to this fake data?

reggieriser:
I went ahead and had infra roll back the experimental database so we can run this migration against experimental again when we're ready (let's make it the last thing we do after reviewer approval but before merging).

See #4279 for a fix that will need to be merged into this PR after #4279 is approved.

Also, I think we should consider changing the name of the migration since `import_pricing_data_june_2020` could imply this is real data.  I'm thinking more like `import_fake_pricing_data`.  We still need to get confirmation of the contract code we want to use.

With the fix from #4279, I checked all the `re_*` tables imported from this PR against a psql-loaded migration and they seemed to match. 🎉

noelledusahel:
> I went ahead and had infra roll back the experimental database so we can run this migration against experimental again when we're ready (let's make it the last thing we do after reviewer approval but before merging).
> 
> See #4279 for a fix that will need to be merged into this PR after #4279 is approved.
> 
> Also, I think we should consider changing the name of the migration since `import_pricing_data_june_2020` could imply this is real data. I'm thinking more like `import_fake_pricing_data`. We still need to get confirmation of the contract code we want to use.
> 
> With the fix from #4279, I checked all the `re_*` tables imported from this PR against a psql-loaded migration and they seemed to match. 🎉

to change the name of the migraiton to `import_fake_pricing_data_` i would need to write code in the script that differentiates between an upload of fake data vs real data. i guess it would look like an extra step in the script that asks the user whether this is real or fake data. would it be better to just give the fake data a contract code that indicates that its fake? like (Fake Truss Data)

reggieriser:
> to change the name of the migraiton to `import_fake_pricing_data_` i would need to write code in the script that differentiates between an upload of fake data vs real data. i guess it would look like an extra step in the script that asks the user whether this is real or fake data. would it be better to just give the fake data a contract code that indicates that its fake? like (Fake Truss Data)

I don't think the script has to change.  You should be able to change the migration filename after the `pricing-import` script creates it -- just remember to also change the `migrations_manifest.txt` file accordingly.  I'm also fine with `import_pricing_data_fake` (if you want to keep the same prefix) or whatever you'd like that makes it clear that it's not real data.  That reminds me -- we will also need to clean up S3 to remove the old migration in each environment and add the new one when it's ready.

noelledusahel:
> > to change the name of the migraiton to `import_fake_pricing_data_` i would need to write code in the script that differentiates between an upload of fake data vs real data. i guess it would look like an extra step in the script that asks the user whether this is real or fake data. would it be better to just give the fake data a contract code that indicates that its fake? like (Fake Truss Data)
> 
> I don't think the script has to change. You should be able to change the migration filename after the `pricing-import` script creates it -- just remember to also change the `migrations_manifest.txt` file accordingly. I'm also fine with `import_pricing_data_fake` (if you want to keep the same prefix) or whatever you'd like that makes it clear that it's not real data. That reminds me -- we will also need to clean up S3 to remove the old migration in each environment and add the new one when it's ready.

ok so you're talking about manually changing it right?

reggieriser:
> ok so you're talking about manually changing it right?

Right.  But if you wanted to change the pricing parser, I'd probably just give the user the option/flexibility to enter a migration name rather than asking them if it's fake data or not.  But I think a manual change is fine for this story.

PR #4263, opened 2020-06-16
[MB-2503] redesign TabNav component, add move header with router

duncan-truss:
## Description

This PR repurposes the TabNav component to act as a navigation component for the Office Users moving between Move Details, Move Task Order, Payment requests, and History pages.

The previous implementation used a 3rd party plugin react-tabs that made it too inflexible to use with a router and was designed more for simple content tab switching.  I also tried to follow the USWDS nav implementation that comes from their header component.

The nav greedily handles all URLs prefixed with `/moves/:id` because this is ultimately going to be the format ([URLs for MilMove](https://docs.google.com/spreadsheets/d/1VWtagZWTcmgqfWaLdQlUHKBJJVtqBXlEJ_KuMMOSZMs/edit?ts=5eb04fcc#gid=0)) when a move-code locator becomes the standard.  There are still others that take precedence and are unaffected by these changes such as `/orders` and `/documents` routes.

For now the url parameter is only going to be the move order id for the details page.  Whereas the mto component expects a move order id and the payment requests component expects a payment request id.  I think it's outside of the scope of this change to decide how to fetch all of those associated ids.

## Reviewer Notes

You should still be able to navigate directly to the `/mto` and `/payment-request` routes if the id in the url is valid for those items, but the nav will not populate the proper ids for the other pages at this time.

It took a lot of work to make this play nicely with uswds as well as not break other components such as LeftNav so there may be some visual regressions.  Let me know if you see any significant styling changes to fix but hopefully Loki will catch others.  The TabNav module scss can likely be consolidated but it is fragile.

This may be part of Suz's upcoming refactor of the Office scene but I saw that render is preferred on a route over using component.

## Setup

View the move nav in the local office app:
1. `make server_run`
2. `make client_run`
3. Navigate to officelocal:3000 using the local login.
4. Find a ppm move order in the queue and change the url to `/moves/:id/details`
5. Verify the nav menu appears and selects the right tab, hover states, etc ...

View the storybook component for TabNav
1. `make storybook`
2. visit http://localhost:6006/?path=/story/components-tabnav--default

```sh
echo "Code goes here"
```

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/secure/RapidBoard.jspa?rapidView=27&modal=detail&selectedIssue=MB-2503) for this change

## Screenshots

Screen Shot 2020-06-23 at 11 22 26 AM (2)


src/components/TabNav/index.module.scss
```diff
@@ -1,49 +1,102 @@
...
 
-.tab-title {
+.tab-item :global .usa-nav__link {
+  display: inline-block;
+  &:hover {
+    text-decoration: none;
+  }
+}
+
+.tab-item :global .usa-nav__link .tab-title {
```
duncan-truss, line 41: Can someone explain the :global use to me, not sure I fully understand does it just work like an include of a global selector?

suzubara: By default, CSS classes defined in a module file will be hashed in the compile - so for example `.tab-item` might become `.TabNav__tab-item__123abc`. However sometimes you need to select non-hashed (global) CSS classes from within a module file. `:global` will switch the scope so that the `usa-nav__link` classname exported by this file will _not_ be hashed. More info at https://github.com/css-modules/css-modules#exceptions

duncan-truss: Thank you, yes I had to brush on my css modules knowledge during this pr and tried to eliminate those global classes getting namespaced.

src/components/Office/TXOMoveInfo.jsx
```diff
@@ -0,0 +1,26 @@
+import React, { lazy } from 'react';
```
suzubara, line 1: Since this component maps directly to a route, I would place it under `src/pages/Office`. I think it would be appropriate to render the TabNav directly in here also, I don't see why not. This page can represent a TXO user viewing a move.

src/components/TabNav/index.jsx
```diff
@@ -1,50 +1,26 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Tag } from '@trussworks/react-uswds';
-import { Tab, Tabs, TabList } from 'react-tabs';
 import classNames from 'classnames/bind';
 import styles from './index.module.scss';
 
 const cx = classNames.bind(styles);
```
suzubara, line 6: i am personally not a fan of the `classnames.bind` syntax, because it becomes less apparent which classes are scoped to a module vs. global. I would recommend switching to the default syntax, ie: `<nav className={classnames('usa-nav', styles.tabNav)}>`

This works better if the classnames are camelCased so I would also recommend using that naming convention for CSS module classnames. More info at https://github.com/JedWatson/classnames

src/components/TabNav/index.jsx
```diff
@@ -1,50 +1,26 @@
...
+    <div className="usa-nav__inner">
+      <ul className={cx('usa-nav__primary', 'tab-list')}>
+        {items.map((item, index) => (
+          <li key={index.toString()} className={cx('usa-nav__primary-item', 'tab-item')} role="tab">
+            {item}
+          </li>
+        ))}
+      </ul>
+    </div>
+  </nav>
```
suzubara, line 19: this is a big improvement for this component IMO! 🎉

src/components/TabNav/index.module.scss
```diff
@@ -1,49 +1,102 @@
...
+// overrides default to float the nav in a collapsed menu
+@media only screen and (min-width: $mobile) {
+  nav.tab-nav {
+    border: none;
+    width: auto;
+    display: block;
+    float: none;
+    position: relative;
+  }
+}
```
suzubara, line 14: I don't see float being applied to this element anywhere? However it looks like on mobile `usa-nav` is assumed to be the primary header navigation, and becomes positioned absolutely. I think we probably want to avoid that classname and roll our own for these styles.

suzubara: Looking further at this file, yeah the USWDS header navigation CSS is _not_ very modular, and I think will just require us to unset more properties than we'd need to set if we were styling from scratch. So that would be my recommendation.

src/pages/TIO/moveHistory.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const MoveHistory = () => {
```
suzubara, line 3: IMO we should scope page components by the app, not necessarily user type, especially since some of the pages can be viewed by multiple user roles. Can this go into `/src/Pages/Office/MoveHistory/`?

src/shared/Header/Office/index.jsx
```diff
@@ -28,4 +32,44 @@ function QueueHeader() {
   );
 }
 
-export default QueueHeader;
+function MoveTabNav(props) {
```
suzubara, line 35: I don't see why this component needs to be defined here. This seems like it represents the tabbed navigation for the Office view move page (`/moves/:moveId`) and could be rendered directly in the new page you added.

src/components/Office/TXOMoveInfo.jsx
duncan-truss: Yea I could see why it might be better categorized as a page eventhough it's more of a container wrapper.  I was trying not to get it confused for standalone content.

It's true the MoveTabNavWithRouter is not doing much now but we'll need to pass more state into it eventually but maybe it's premature to separate them.

src/components/TabNav/index.jsx
duncan-truss: Yes that pattern is what I wanted!

src/components/TabNav/index.module.scss
```diff
@@ -1,49 +1,102 @@
...
+
+.tab-list .tab-item {
+  border: none;
   @include u-padding('2');
   @include u-display('inline-block');
   &:focus {
     @include u-outline('05', 'blue-40v');
-  //  @include u-padding-top('05'); adjusting padding on focus will make our menu items scoot, we don't want that!!
-  //  @include u-padding-bottom('05');
+    //  @include u-padding-top('05'); adjusting padding on focus will make our menu items scoot, we don't want that!!
```
Ryan-Koch, line 78: Do we feel we can remove these commented lines or do we need to keep them?

duncan-truss: They are probably safe to remove now

duncan-truss:
> One thing I ran into while testing this manually is that coming from `/too/customer-moves` takes me to a page that is rendering only the tabnav component. I suspect that's because the link is going to `/moves/id` instead of `/moves/id/details` (it's good that you've corrected that to match the spec). Would you be adverse to including an update to the link to address this? I think you'd need to adjust the `handleCustomerInfoClick` function in `scenes/Office/TOO/too.jsx` like so:
> 
> ```
>  handleCustomerInfoClick = (moveOrderId) => {
>     this.props.history.push(`/moves/${moveOrderId}/details`);
>   };
> ```

Yea absolutely, that's good info I didn't have about how others would be using/testing.

duncan-truss:
> tested in edge, firefox, and chrome. looks good and is functioning as expected.
> 
> Worth noting, there is a [documented bug](https://dp3.atlassian.net/browse/MB-1906) in regards to keyboard focus with the TabNav and it might be time to address it. In Firefox, the TabNav does not get keyboard focus at all when I tab through the page, and In Edge and Chrome the check boxes get focus when I start to tab through the page. I have to `Shift + Tab` to go backwards in the focus order in order for the TabNav items to receive focus.
> 
> Approving because PR still meets JIRA acceptance requirements, but felt it was worth flagging.

We troubleshooted because the same issue cropped up with the LeftNav on Firefox.  Turns out on the Mac there is a system setting in the keyboard that needs enabled to focus on non-form elements.  Followed by adding a config flag to firefox for `accessibility.tabfocus`.  More details here: https://stackoverflow.com/questions/11704828/how-to-allow-keyboard-focus-of-links-in-firefox

src/components/TabNav/index.jsx
Ryan-Koch: At a future standup would either of you be willing to talk about this briefly? It would be good to socialize this to keep things consistent.

duncan-truss: Okay one issue with changing to not bind all styles is that I have overrides to uswds in the module.scss which I don't think get incorporated anymore.

src/components/TabNav/index.module.scss
duncan-truss: I think it's really only under $mobile (365px) that it goes to fixed.  That's probably not the only mobile work that needs done but I think it's as good a default as any at this point.  The guidance I saw said that there should only be one `role="navigation"` per page so I made that a prop that can optionally be added.

src/shared/Header/Office/index.jsx
suzubara: re-iterating I think this should go into its own component in the new file structure.

src/pages/TXOMoveInfo.jsx
```diff
@@ -0,0 +1,162 @@
...
+  } else if (matchPath(location.pathname, { path: '/moves/:id/history', exact: true })) {
+    matchHistory = true;
+  }
+
+  /* eslint-disable react/jsx-props-no-spreading */
+  return (
+    <>
+      <header className="nav-header">
+        <div className="grid-container-desktop-lg">
+          <TabNav
```
suzubara, line 44: This is a minor thing I noticed which may fall outside the scope of this story. Since not all of these tabs are accessible by both TOO & TIO users, if you log in as one you will view all of the tabs, but clicking one you don't have access to brings you back to the home page. It might be a better UX if inaccessible tabs are not shown to those user roles at all.

src/pages/TXOMoveInfo.jsx
```diff
@@ -0,0 +1,162 @@
...
+
+const TXOMoveInfo = ({ too, tio, tag, match, location }) => {
+  const { moveOrderId } = match.params;
+
+  let matchDetails = false;
+  let matchMTO = false;
+  let matchPaymentRequests = false;
+  let matchHistory = false;
+
+  // Used to set aria-expanded attribute to selected tab
```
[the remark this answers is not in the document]
duncan-truss: Yea I see what you're saying, they use that attribute for the submenus.  NavLink sets `aria-current="page"` so that is likely enough.

src/pages/TXOMoveInfo.jsx
duncan-truss: Yep I was tempted to dive into disabling or hiding but wanted to not expand the scope further.  There seems to be a few tickets that deal with unauthorized and page not found behavior out there where some of this can be tackled.

PR #4264, opened 2020-06-16
Add ADR about F/E file organization

suzubara:
## Description

I've heard a few times that there is some confusion around the new F/E file structure, so I wanted to add an ADR that describes why we're changing file structures and how to approach it, as well as some additional guidelines to help us stay organized. The file structure itself is described in the [wiki](https://github.com/transcom/mymove/wiki/frontend#file-layout--naming).

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,8 @@
     "prettier": "prettier --loglevel warn 'src/**/*.{js,jsx}'",
     "lint": "eslint --ext .js --ext .jsx src",
     "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook"
+    "build-storybook": "build-storybook",
+    "spellcheck": "mdspell --ignore-numbers --ignore-acronyms --en-us --no-suggestions"
   },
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -86,7 +86,8 @@
     "prettier": "prettier --loglevel warn 'src/**/*.{js,jsx}'",
     "lint": "eslint --ext .js --ext .jsx src",
     "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook"
+    "build-storybook": "build-storybook",
+    "spellcheck": "mdspell --ignore-numbers --ignore-acronyms --en-us --no-suggestions"
```
suzubara, line 90: added this because I prefer not to use yarn global

suzubara:
> I like this! I wonder if we could add some danger rules to warn people when we touch files in the scenes and shared folders? (To recommend seeing if touched files could be updated.)

I'm fine with that! I would like to address it in a separate PR though to keep the ADR one simple

PR #4278, opened 2020-06-18
MB-2700 Make moveTaskOrderID required for service items

abbyoung:
## Description

* This PR updates `prime.yaml` and affected code to make `moveTaskOrderID` required for `CreateMTOServiceItem`. 

* The parent ids for `CreateMTOShipment` and `CreateMTOServiceItem` were previously removed from the path and added to the body payload for each. [MB-2700](https://dp3.atlassian.net/browse/MB-2700) is to make sure these fields are marked as required for the payload.
* `CreateMTOShipment` already has `moveTaskOrderID` marked as required and returns the proper validation error if omitted, so no action was needed for that AC.

## Reviewer Notes

> One question that came up in conversations with @shimonacarvalho and @sandy-wright was, "Do 

> we want these fields to always be required for the `MTOServiceItem` definition? Is this definition used elsewhere?"
> 
> Besides the `CreateMTOServiceItem` endpoint, the `MTOServiceItem` definition is referenced in `CreateShipmentPayload`, which is used in `CreateMTOShipment`. Here is where I need help:
> 
> * **CreateMTOShipment** - create a new MTO shipment, which requires a `moveTaskOrderID`, and returns an `mtoShipmentID`
> * **CreateMTOServiceItem** - create a new service item, which requires an `mtoShipmentID`, and that shipment's `moveTaskOrderID`.
> 
> Is it correct to assume you cannot create a service item unless you have an existing mtoShipment to link? Or should the client be able to pass in mto service items during the creation of an mtoShipment?

> Our [current documentation](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/transcom/mymove/master/swagger/prime.yaml#operation/createMTOShipment) makes it seem like the latter, stating the mtoShipment can be passed "An array of optional accessorial service item codes."

> Let's discuss and clarify the expected flow! 🙏 

### Update to above questions
* Service items can be created without an associated shipment, e.g. Counseling Fee, so `mtoShipmentID` should not be required on the definition.
* But `mtoShipmentID` is required for the Prime using the createMTOServiceItem endpoint, because they will be using this endpoint to create accessorial service items which will have both `mtoShipmentID` and `moveTaskOrderID`.
  * This means it's ok to call out they are required values in the prime.yaml description of the endpoint; just not in the definition.
* `moveTaskOrderID` *can* be required for creating service items.
  * Currently in the support-create-mto endpoint, it appears as though you can create an mto with service items, but this is incorrect and will be fixed in a future story.

### Test Endpoint
`make db_dev_e2e_populate && make server_run`
* Test CreateMTOServiceItem endpoint and verify validation error if no `moveTaskOrderID` is passed.

### Check Documentation
In a separate tab, run:
`redoc-cli server swagger/prime.yaml`

* Verify in documentation for createMTOServiceItem that `moveTaskOrderID` is marked as required.

### Run Tests
`make db_test_e2e_populate`
`go test ./pkg/handlers/primeapi/ -testify.m TestCreateMTOServiceItemHandler`

## References
* [Jira story](https://dp3.atlassian.net/browse/MB-2700) for this change

### DIFF
--- a/pkg/handlers/primeapi/internal/payloads/model_to_payload.go
+++ b/pkg/handlers/primeapi/internal/payloads/model_to_payload.go
@@ -445,14 +445,10 @@ func MTOServiceItem(mtoServiceItem *models.MTOServiceItem) primemessages.MTOServ
 	}
 
 	// set all relevant fields that apply to all service items
-	var shipmentIDStr string
-	if mtoServiceItem.MTOShipmentID != nil {
-		shipmentIDStr = mtoServiceItem.MTOShipmentID.String()
-	}
 
 	payload.SetID(strfmt.UUID(mtoServiceItem.ID.String()))
-	payload.SetMoveTaskOrderID(strfmt.UUID(mtoServiceItem.MoveTaskOrderID.String()))
-	payload.SetMtoShipmentID(strfmt.UUID(shipmentIDStr))
+	payload.SetMoveTaskOrderID(handlers.FmtUUID(mtoServiceItem.MoveTaskOrderID))
+	payload.SetMtoShipmentID(handlers.FmtUUIDPtr(mtoServiceItem.MTOShipmentID))
 	payload.SetReServiceID(strfmt.UUID(mtoServiceItem.ReServiceID.String()))
 	payload.SetReServiceName(mtoServiceItem.ReService.Name)
 	payload.SetStatus(primemessages.MTOServiceItemStatus(mtoServiceItem.Status))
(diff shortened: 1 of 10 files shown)


pkg/handlers/primeapi/internal/payloads/model_to_payload.go
```diff
@@ -445,14 +445,10 @@ func MTOServiceItem(mtoServiceItem *models.MTOServiceItem) primemessages.MTOServ
 	}
 
 	// set all relevant fields that apply to all service items
-	var shipmentIDStr string
```
abbyoung, line 448: If the field becomes required, do we still need this code block?

PR #4293, opened 2020-06-19
MB-1813 Document Viewer Components

Shkeating:
## Description

Adds the Document Viewer Menu and the Document Viewer Content Area to Storybook.

## Reviewer Notes

Here are some things that need to be adjusted, input welcomed and appreciated:

- [x] too much space at the top of the menu
- [x] Menu Items don't get keyboard focus, when a `tabindex=0` is applied, eslint argues that a tabindex has been given to a "non-interactive" element and blocks the commit.
- [ ] need to add labels or alt text to icon-only buttons such as the close on the menu and the arrows to navigate between documents
- [ ] weird scrolling background behavior on menu when placed in context
- [ ] need to configure the menu to expand and collapse correctly

## Setup

```sh
yarn install
yarn storybook
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-1813?atlOrigin=eyJpIjoiNDJlNTBjOTI3NjA3NDI2ZmE1YThmMjdkN2IzNGYwODkiLCJwIjoiaiJ9) for this change
* [Abstract Spec](https://share.goabstract.com/604f7f29-c8bc-4c91-af36-92cd130b5b3f?collectionLayerId=ee8c3b38-8397-4a02-99a2-b8051afe431f&mode=build)

## Screenshots
image
image


src/components/DocumentViewer/content.module.scss
```diff
@@ -0,0 +1,81 @@
+@import '../../shared/styles/basics';
+@import '../../shared/styles/mixins';
+@import '../../shared/styles/colors';
+
+$menuWidth: 308px;
+$contentWidth: 80%;
+
+svg path {
```
duncan-truss, line 8: Seems like you scoped this on the menu component under a class name, did you want to do the same thing here?

src/components/DocumentViewer/menu.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+import styles from './menu.module.scss';
+
+import { ReactComponent as XLightIcon } from 'shared/icon/x-light.svg';
+import { ReactComponent as DocMenu } from 'shared/icon/doc-menu.svg';
+
+const testThumbnail = (
+  <a className={styles.thumbnailItem} href="#">
+    <p>ASamplePrettyLongDocumentTitle.png</p>
+    <div className={styles.thumbnailImage} />
```
duncan-truss, line 12: I know this is a just a testThumbnail but we probably want this to contain an `` element instead of specifying the background-url.

src/components/DocumentViewer/menu.module.scss
```diff
@@ -0,0 +1,132 @@
...
+    @include u-text('no-underline');
+    p {
+      @include u-text('primary-vivid');
+    }
+    .thumbnailImage {
+      @include border-inset(2px, $link);
+    }
+  }
+  &:active,
+  &.active {
```
duncan-truss, line 108: Do you want to show this off in your stories to show one of the items with the active class?

src/components/DocumentViewer/content.module.scss
suzubara, line 8: This will style _all_ `svg path` elements globally since it's not scoped with a CSS module class. If this isn't what you want then this should probably be scoped to a class.

src/stories/docviewer.stories.jsx
```diff
@@ -0,0 +1,17 @@
...
+import DocViewerMenu from '../components/DocumentViewer/menu';
+import DocViewerContent from '../components/DocumentViewer/content';
+
+storiesOf('Components|Document Viewer', module)
+  .add('menu', () => (
+    <div>
+      <DocViewerMenu />
+    </div>
+  ))
+  .add('content area', () => (
```
suzubara, line 13: I think these two should actually be structured as stories nested within `Components|Document Viewer` (ie, `Components|Document Viewer|Menu` &  `Components|Document Viewer|Content Area`) since they represent individual components, not stories (test cases)

src/components/DocumentViewer/content.module.scss
Shkeating: I did this this way because I was having a really hard time getting all the icons to take the blue fill in their default state. I agree that leaving it at the top of the file like that was a little weird though, I moved it furthr down neareer to other sass styling icon fills, and left a comment explaing it's purpose

Shkeating: I realized that you actually meant that I needed to scope this to a class or it would be applied to svg icons outside of the component, and fixed it

src/components/DocumentViewer/menu.jsx
Shkeating: yeah i wasn't sure what would be best, I just switched it to be an `` that's clipped by a `.thumbnailImage` `div`. Thanks!

src/stories/docviewer.stories.jsx
Shkeating: so do you think it should be one of those lil folders in the side bar?

suzubara: @Shkeating yup!

PR #4294, opened 2020-06-19
MB-3001 upgrade HERE URLs

garrettqmartin8:
## Description

We're currently using [HERE](https://www.here.com/)'s CIT (Customer Integration and Testing) environments to do distance calculations and routing. On June 30th, CIT will be deprecated. This PR moves us to using the production URLs that are supported by HERE. The API is under maintenance/on long-term support, but we'll be moving to another system in the next few months.

## Reviewer Notes

According to HERE support, the only change we needed to make was to the URLs - our current API keys still work.

## Setup

- `make db_dev_e2e_populate`
- `make server_run`
- `make client_run`
- Navigate to `http://milmovelocal:3000/devlocal-auth/login` and create a new Milmove user
- Fill out move information until you hit the weight entitlement screen. You should be able to move the slider and see your entitlement estimate.
- To test that this would break without the proper HERE urls, stop the server, open `.envrc` and edit the `HERE_MAPS_*` variables to bogus urls. Start the server again, and on the weight entitlement screen you'll see this:
  Screen Shot 2020-06-19 at 10 46 34
- You should also see an error in the server log complaining about a bad host.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3001) for this change


garrettqmartin8:
> This worked for me locally as described, but I still see a couple of references to the old URL in our CircleCI `.config` file:
> 
> https://github.com/transcom/mymove/blob/22d0965f204f1c8141025e4e2fb290bd29e69162/.circleci/config.yml#L326-L327

@reggieriser argh, good find. Need to tweak `rg` to make sure it looks for references in Circle config.

PR #4295, opened 2020-06-19
[MB-2873] Refactor Office index

suzubara:
## Description

Some minor clean up & changes to the Office entry point, to allow for easier testing and more cleanup down the road.
- Moved the Office `ConnectedRouter` from `Office/index` to `App/index`. This will let us write tests for how the Office index component handles URLs (and since the router & Redux set up usually go hand-in-hand, it makes sense for the Router & Provider to be rendered in the same place).
- Moved `scenes/Office/index` to `pages/Office/index` and made ESLint fixes
- Added some tests to `Office/index` (I'm not super happy with how the routes are tested but was really coming up against lazy loaded components not rendering... see the comment I left)
- Added a `MockProviders` test component to more easily render components inside configurable Provider/Routers for testing specific state & URLs
- Removed over-abstracted conditional wrappers and cruft from the Office render fn.
  - Page content is always rendered in a `<main>` tag, this is fine for IE11.
  - Only 2 exceptions don't render the QueueHeader so I made that more explicit
  - Removed extraneous Suspense & Switch wrappers

## Reviewer Notes

There should be no discernible changes to the user, so code review & testing should focus on making sure there are no regressions.

## Setup

1. Run the app locally
    - `make server_run`
    - `make client_run`
1. Visit `officelocal:3000` and log in using the local login for each office user type (PPM, TOO, TIO)
1. Verify you are directed to the correct home screen based on your user role, and can navigate to expected pages
1. Log in as an office user with multiple roles (you may need to add roles to a user using the admin) and verify you can switch roles using the "Change user role" link in the header

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2873) for this change


src/pages/Office/index.jsx
```diff
@@ -0,0 +1,205 @@
...
+  loadInternalSchema: PropTypes.func.isRequired,
+  loadPublicSchema: PropTypes.func.isRequired,
+  getCurrentUserInfo: PropTypes.func.isRequired,
+  context: PropTypes.shape({
+    flags: PropTypes.shape({
+      too: PropTypes.bool,
+      tio: PropTypes.bool,
+    }),
+  }),
+  location: PropTypes.shape({
```
duncan-truss, line 172: I think I added this prop type to `src/types/router.js` next to the match shape so we don't have to keep redefining it.

src/pages/Office/index.test.jsx
```diff
@@ -0,0 +1,333 @@
...
+        );
+
+        const renderedRoute = app.find('Route');
+        expect(renderedRoute).toHaveLength(1);
+        expect(renderedRoute.prop('path')).toEqual('/moves/:moveId/documents/:moveDocumentId?');
+
+        // DocumentViewer does NOT render the header
+        expect(app.find('QueueHeader')).toHaveLength(0);
+      });
+    });
```
duncan-truss, line 157: These tests are great, unfortunate there's that lazy load limitation.

src/pages/Office/index.test.jsx
```diff
@@ -0,0 +1,333 @@
...
+          );
+
+          const renderedRoute = app.find('Route');
+          expect(renderedRoute).toHaveLength(1);
+          expect(renderedRoute.prop('path')).toEqual('/moves/:moveOrderId');
+        });
+
+        it('handles the CustomerDetails URL', () => {
+          const app = mount(
+            <MockProviders initialState={loggedInTOOState} initialEntries={['/too/123/customer/abc']}>
```
duncan-truss, line 222: It might be too much work for this story but I wonder if we can make these values match the e2e test data.

src/shared/User/PrivateRoute.jsx
```diff
@@ -37,10 +37,10 @@ const PrivateRouteContainer = (props) => {
     return displaySelectApplication ? (
       <>
         <Link to="/select-application">Change user role</Link>
-        <Route {...routeProps} />
+        <Route path={path} {...routeProps} />
       </>
     ) : (
-      <Route {...routeProps} />
+      <Route path={path} {...routeProps} />
```
duncan-truss, line 43: The react-router docs seem to prefer `render` or `children` over the `component` prop.  Do you think that's a change worth making https://reacttraining.com/react-router/web/api/Route/component ?

src/pages/Office/index.jsx
suzubara: oh yeah! I will swap that out

src/shared/User/PrivateRoute.jsx
suzubara: actually `component` takes precedence over `render`... `render` is more performant if you are passing in an inline function (as opposed to a component reference) and `children` is for if you want to handle the `match` value yourself.

duncan-truss: Ahh okay, I think I read it wrong meaning component would always render a new component regardless of if you had an inline function or not.

src/pages/Office/index.jsx
```diff
@@ -0,0 +1,208 @@
...
+      }) ||
+        matchPath(pathname, {
+          path: '/moves/:moveId/orders',
+          exact: true,
+        }));
+
+    return (
+      <div className="site">
+        <FOUOHeader />
+        {!hideHeader && <QueueHeader />}
```
duncan-truss, line 104: I noticed that the change user role link has now moved below the header.  On staging it is above, if you didn't mean to change the placement.

src/scenes/Office/index.jsx
```diff
@@ -1,281 +0,0 @@
...
-                    requiredRoles={[roleTypes.TOO, roleTypes.TIO]}
-                  />
-                )}
-                {!isProduction && (
-                  <PrivateRoute
-                    path="/playground"
-                    component={(props) => (
-                      <Suspense fallback={<LoadingPlaceholder />}>
-                        <RenderWithOrWithoutHeader
-                          component={ScratchPad}
```
duncan-truss, line 211: Seems like this component can be deleted now if `/playground` is no longer being used.

src/pages/Office/index.jsx
```diff
@@ -0,0 +1,208 @@
...
+                    component={TXOMoveInfo}
+                    requiredRoles={[roleTypes.TOO, roleTypes.TIO]}
+                  />
+                )}
+                {too && <Route path="/verification-in-progress" component={TOOVerificationInProgress} />}
+                {tio && <PrivateRoute path="/invoicing/queue" component={TIO} requiredRoles={[roleTypes.TIO]} />}
+                {tio && (
+                  <PrivateRoute
+                    path="/payment_requests"
+                    component={PaymentRequestIndex}
```
duncan-truss, line 152: I don't think this is anything you changed but is related to the routes refactor where the link on the PaymentRequestIndex page should go to `/moves/:id/payment-requests` instead of `/payment_requests/:id` 

@Ryan-Koch is that your understanding?

src/pages/Office/index.jsx
suzubara: Ahh good catch, I didn't even see that. if it's not a big deal for now I think I'd like to address that as part of MB-3015

src/pages/Office/index.jsx
suzubara: Yeah this PR shouldn't change any URLs, I think this can be changed as part of MB-3056

PR #4299, opened 2020-06-19
MB-3030 - Remove cgilmer CAC access

chrisgilmerproj:
## Description

Removes CAC access. The record of who had access for auditing is in both the git repo and the s3 bucket history. While disabling all permissions feels like an ok route to take here its just as easy for someone to re-add those permissions. And the CAC itself is going to be destroyed so removing seems more prudent.

This doesn't have to be a secure migration because we're only identifying the sha256 digest in dev/experimental/staging/prod. And if it doesn't exist in those environments this is a noop.

## Reviewer Notes

**NOTE:** This needs to be deployed manually to experimental to apply in that environment.

## Setup

```sh
make db_dev_run db_dev_migrate
```

### DIFF
--- a/migrations/app/migrations_manifest.txt
+++ b/migrations/app/migrations_manifest.txt
@@ -515,3 +515,4 @@
 20200527145249_add_mto_available_timestamp.up.sql
 20200529171253_update_task_order_services_params.up.sql
 20200602144311_remove_is_available_to_prime.up.sql
+20200619120000_remove_cgilmer_cac.up.sql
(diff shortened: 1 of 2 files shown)


chrisgilmerproj:
@rdhariwal @mr337 - there's a problem with the deploy pipeline here and its not clear why. Here's the output from the [circleci workflow](https://app.circleci.com/pipelines/github/transcom/milmove_orders/567/workflows/49a50bf1-dfcc-4bbf-9b4f-4b56d110557b/jobs/4352).

```

PR #4300, opened 2020-06-19
Kim mb 2137 orders type dropdown

kimallen:
## Description

This adds two more options to the orders_type field in the Orders Wizard form page.

## Reviewer Notes

I spent too much time trying to figure out how to check the fields in the client test, and ended up opting for checking in the cypress test instead since we will have to build that out anyway as we update the form.

## Setup

Use local user `needs@orde.rs`

Go to the orders page.  Check the drop down and see the three options.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2137) for this change
* [this convo](https://ustcdp3.slack.com/archives/CTNTFJSBA/p1592498834001600) for copy instructions.

## Screenshots

image

### DIFF
--- a/src/scenes/Orders/Orders.jsx
+++ b/src/scenes/Orders/Orders.jsx
@@ -60,7 +60,7 @@ export class Orders extends Component {
         serverError={error}
       >
         <h1 className="sm-heading">Tell us about your move orders</h1>
-        <SwaggerField data-cy="orders_type" fieldName="orders_type" swagger={this.props.schema} required />
+        <SwaggerField fieldName="orders_type" swagger={this.props.schema} required />
         <SwaggerField fieldName="issue_date" swagger={this.props.schema} required />
         <div style={{ marginTop: '0.25rem' }}>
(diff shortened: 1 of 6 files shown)


src/scenes/Orders/Orders.jsx
```diff
@@ -60,7 +60,7 @@ export class Orders extends Component {
         serverError={error}
       >
         <h1 className="sm-heading">Tell us about your move orders</h1>
-        <SwaggerField fieldName="orders_type" swagger={this.props.schema} required />
+        <SwaggerField data-cy="orders_type" fieldName="orders_type" swagger={this.props.schema} required />
```
amitch23, line 63: I gather that `data-cy` is an attribute for the cypress test to aid in selection, but I'm not seeing how it's being used in the test adapted above. Did you intend to use this for the orders integration test, will it be used in the future, or maybe am I missing something about how it's being used?

kimallen: oh, right.  I added that when I was attempting to write the client test, but it's no longer necessary.

PR #4307, opened 2020-06-22
MB-2509 Clean up service items in Support API

abbyoung:
## Description

* Only `status` and `rejectionReason` can be updated via `UpdateMTOServiceItemStatus`. This PR makes all other fields `readOnly` so it's less confusing to the user and the fields do not appear  in the docs request payload.

* Many date-time fields for `MTOServiceItem` are not used or stored in the database. There is future work planned around these, so for now consensus was to remove them from `support.yaml`.

* Same hold applies to `createdAt` and `updatedAt` -- unsure if we want to include these in the payload. Pop automatically creates these in the db, so an easy revert if we decide to use them in the API response.

* Added title, error, instance patter for returning `404` from `UpdateMTOServiceItemStatus`. 

## Setup

```sh
```

### readOnly fields
```
redoc-cli serve swagger/support.yaml
```

* Observe `updateMTOServiceItemStatus` section and confirm only `status` and `rejectionReason` appear in the request body schema.
* Test `support/v1/service-items/:mtoServiceItemID/status` and confirm it's working and returns the expected payload.

### 404 Testing
* Submit a request with a faulty uuid to `support/v1/service-items/:mtoServiceItemID/status` and observe the 404 error returned. It should look similar to the following: 

```
{
    "detail": "id: 00000000-0000-0000-0000-000000000000 not found MTOServiceItemID",
    "instance": "17330d5c-3d36-4272-af8a-a41417169f9b",
    "title": "Not Found Error"
}

```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2509) for this change

### DIFF
--- a/pkg/handlers/supportapi/mto_service_item.go
+++ b/pkg/handlers/supportapi/mto_service_item.go
@@ -38,7 +38,8 @@ func (h UpdateMTOServiceItemStatusHandler) Handle(params mtoserviceitemops.Updat
 
 		switch e := err.(type) {
 		case services.NotFoundError:
-			return mtoserviceitemops.NewUpdateMTOServiceItemStatusNotFound()
+			payload := payloads.ClientError(handlers.NotFoundMessage, err.Error(), h.GetTraceID())
+			return mtoserviceitemops.NewUpdateMTOServiceItemStatusNotFound().WithPayload(payload)
 		case services.InvalidInputError:
 			payload := payloads.ValidationError("The information you provided is invalid", h.GetTraceID(), e.ValidationErrors)
 			return mtoserviceitemops.NewUpdateMTOServiceItemStatusUnprocessableEntity().WithPayload(payload)
--- a/pkg/gen/supportapi/embedded_spec.go
+++ b/pkg/gen/supportapi/embedded_spec.go
@@ -1352,21 +1352,26 @@ func init() {
       "properties": {
         "approvedAt": {
           "type": "string",
-          "format": "date"
+          "format": "date",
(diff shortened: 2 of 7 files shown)


pkg/handlers/supportapi/mto_service_item.go
```diff
@@ -38,7 +38,8 @@ func (h UpdateMTOServiceItemStatusHandler) Handle(params mtoserviceitemops.Updat
 
 		switch e := err.(type) {
 		case services.NotFoundError:
-			return mtoserviceitemops.NewUpdateMTOServiceItemStatusNotFound()
+			payload := payloads.ClientError(handlers.NotFoundMessage, err.Error(), h.GetTraceID())
```
shimonacarvalho, line 41: The error you pasted in the PR description shows 0000 id not found. If you had passed in a valid uuid that didn't exist, would the error message properly populate the uuid?

abbyoung: great catch -- checking now 👍

sandy-wright: I also just tested this - it won't 😄 . This line (38) in `services/mto_service_item/mto_service_item_updater.go` needs to change:

```
if err != nil {
    return nil, services.NewNotFoundError(mtoServiceItem.ID, "MTOServiceItemID")
}
```

to use the ID that was passed in and not the one associated with the empty model.

abbyoung: just updated!

PR #4308, opened 2020-06-22
[MB-2566] ADR about optimistic locking when updating children in an endpoint

shimonacarvalho:
## Description

Currently in some endpoints like updateMTOShipment, we pass in the etag of the shipment and then update the address. However, this means we are not checking the etag of the address. And there’s only one spot in the header for If-Match.

This ADR decides in this case, it's best to create a separate endpoint for the child, so it can be properly updated with its own etag. 

Slack convo: https://ustcdp3.slack.com/archives/CP6PTUPQF/p1592849994343000

### DIFF
--- a/docs/adr/0049-etag-for-child-updates.md
+++ b/docs/adr/0049-etag-for-child-updates.md
@@ -1,0 +1,42 @@
+# Do not update child records using parent's etag
+
+**User Story:** [https://dp3.atlassian.net/browse/MB-2566]
+
+When we have an endpoint that updates a record in the db, it's sometimes desirable to update a child record as well. 
+
+Generally, to update a record, the caller must provide an etag, passed in the header `If-Match` that matches that of the record in the db. 
+
+However the parent and child have two different etags, and the etag is passed in a sole parameter in the header.
+
+Therefore, it's not possible to pass in the child and parent etag cleanly. 
+
+## Considered Alternatives
+
+* Make a new endpoint for child updates so they can be updated separately with the correct etag.
+* Pass a second etag in the body if the child is to be updated.
+* Bubble up a child's updated_at value to the parent, so that the child and parent will have one etag.
+
+## Decision Outcome
+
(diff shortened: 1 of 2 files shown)


docs/adr/0049-etag-for-child-updates.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Considered Alternatives
+
+* Make a new endpoint for child updates so they can be updated separately with the correct E-tag.
+* Pass a second E-tag in the body if the child is to be updated.
+* Bubble up a child's updated_at value to the parent, so that the child and parent will have one E-tag.
+
+## Decision Outcome
+
+We will make a new endpoint for child updates so they can be updated separated with the correct E-tag.
```
LeDeep, line 21: typo - separated should be separately

shimonacarvalho: oh funny, i saw that before, thought I'd fixed it.

docs/adr/0049-etag-for-child-updates.md
```diff
@@ -0,0 +1,43 @@
+# Do not update child records using parent's E-tag
+
+**User Story:** [Jira Story](https://dp3.atlassian.net/browse/MB-2566)
+
+When we have an endpoint that updates a record in the db, it's sometimes desirable to update a child record as well.
+
+Generally, to update a record, the caller must provide an E-tag, passed in the header `If-Match` that matches that of the record in the db.
+
+However the parent and child have two different E-tags, and the E-tag is passed in a sole parameter in the header.
```
lynzt, line 9: Might be clearer:  "...and only parent E-tag is passed in the header"
I assume it's the parent e-tag...

shimonacarvalho: Yes. the parent e-tag. 

And unless I'm misunderstanding your questions, nothing will be inconsistent, because the child update will not update the parent's etag in anyway. Or vice versa.

PR #4316, opened 2020-06-23
MB-2969 add fouo banner to ssw pdf

sojeri:
## Description

Adds the finest FOUO all caps text to the SSW PDF -- now with green banner. :)

## Reviewer Notes

~~Green bar aside, do you have an opinion on whether it's better to do this programmatically or by updating the backing png we use to generate the PDFs? (I am always more inclined toward raw text because it is more accessible.)~~

## Setup

1. Pull changes and generate yourself a shiny new SSW.
2. Bask in the reality of the new text.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2969) for this change

## Screenshots

image


sojeri:
@kimallen-- Nope, we can't use the react component. The PDF generator is part of the go BE server and places all text by unit-based measurements. A lot of what you see in the final PDF is actually a png we're drawing over. If we want an accessible PDF -- that is, one screen readers can actually read -- we will need to do this differently. IDK if feasible but we could revisit starting from react components at that time. :)

PR #4337, opened 2020-06-26
[MB-3061] Consolidate service_members and customers tables

reggieriser:
## Description

There is almost 100% duplication between the `service_members` table and the `customers` table at the moment.  The new onboarding process does not seem to suggest any reason why these should need to remain separated – the customer profile data collected appears to be the same regardless of the type of move you're making.  So, this PR consolidates that data into a single table going forward.

To minimize the breadth of the changes (and risk of breaking production), this first pass attempts to migrate the data and change just the database interactions on the `customers` side to make them use the `service_members` table instead.  No APIs or payloads have been changed yet.  Separate stories will handle API rework and potentially changing the `service_members` naming to `customers`.

## Reviewer Notes

I'm marking this as a WIP for the moment.  How far do we want to go with renaming of symbols and such?  What other cleanups are appropriate?

## Setup

- `make server_test`
- `make e2e_test`
- Try out various functionality that deals with profile information -- onboarding, converting PPM to GHC, endpoints with customer data, etc.

* [ ] The requirements listed in [Querying the Database Safely](https://github.com/transcom/mymove/wiki/backend#querying-the-database-safely) have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3061) for this change
* [MilMove Database Notes](https://docs.google.com/document/d/1-iNZoTLFOwQcPBqQbAjXB87YxxFYHiXWySz4HSbS_Zs/edit#) has some artifacts from my research into the duplication


migrations/app/schema/20200623195126_customers_to_service_members.up.sql
```diff
@@ -0,0 +1,61 @@
...
+       first_name,
+       last_name,
+       agency,
+       email,
+       phone,
+       current_address_id,
+       false -- required field, so set to false (matching what's in dev/exp/prod)
+       -- ignoring destination_address_id from customers as it's never set except from testdatagen
+FROM customers
+ON CONFLICT (user_id) DO UPDATE SET old_customer_id = EXCLUDED.id;
```
garrettqmartin8, line 37: TIL about `ON CONFLICT`!

swagger/ghc.yaml
```diff
@@ -1313,6 +1313,7 @@ definitions:
       current_address:
         x-nullable: true
         $ref: '#/definitions/Address'
+      # TODO: No more destination address, so could be removed from API payload
```
garrettqmartin8, line 1316: Should these TODOs to be addressed now, or in a future PR?

monfresh:
Thanks for the comments, @lynzt. Here are the answers:

1. This is only an issue in non-prod environments, where I've been told the customer data was initially generated by copying it from service_members. I asked multiple times, and the consensus is that on conflict, we want to drop the customer data and use the service_member data instead.

2. I didn't think about that. I don't think there should be an issue. The only table that referenced customers was move_orders, so that's the one I updated to point to the new service_member id.

3. I don't think that's an issue according to Reggie's comment in the migration that It was only set by testdatagen. I see that it is being referenced by some React views, but I just tested it, and it's not causing any errors. I believe Reggie wanted to make the least amount of changes in this PR, and then come back and clean up the API.

4. Locking the service_members table as well is probably a good idea.

5. I had the same question when I was reviewing this PR, and looking through other migrations where we lock tables, it doesn't look like we need to do anything special to unlock it.

6. That's a good idea.

swagger/ghc.yaml
monfresh: I think the plan is to clean up the API in a future PR once everything is working as expected.

PR #4345, opened 2020-06-29
[MB-2664] Consistent client server errors for Prime API

shimonacarvalho:
## Description

Trying to make the errors consistent. This is a tedious process so broke the tickets up. In this ticket I did createMTOShipment, updateMTOShipment, createMTOServiceItem and updateMTOPostCounseling.
Separate ticket created for support api and for paymentrequest related items, because we probably need to talk to a-team first. 

Went with this strategy... 

For correctness, we want server errors only to be returned when it is really an unexpected error on the server and not an error that the client can correct. 

If we can read and understand the request, it’s likely the error is one of the following
- Unprocessable Entity - these are validation errors, missing values, invalid values
- Conflict Error - we cannot process the request due to the current state of the server (for e.g.. we cannot change this value once this date is past)
- Not Found error - we cannot find an object related to completing the request.

The above are ALL 4xx errors. Note that Bad request isn’t one of them. Typically we should never need to return Bad Request because it implies we were unable to read the request. Since the request is first parsed by swagger, swagger will return bad request. And the handler never gets called unless the request was able to be correctly read and parsed.

If we can read and understand the request, AND there’s nothing wrong with the request BUT it still fails, we can return an Internal Server Error. In the service object, we can generate a queryError which then returns an InternalServer error response. 
- The logger should print the pq error
- The response should NOT contain the pq error, but should not be a nil payload. It should have a default internal server error title/message/instance

## Reviewer Notes

Try to generate errors with the endpoints that were fixed. Ideally we should get useful responses not 500 errors. 

## References

* [MB-2664 Make server errors consistent for the Prime APIs - Jira](https://dp3.atlassian.net/browse/MB-2664)

[MB-2664]: https://dp3.atlassian.net/browse/MB-2664

### DIFF
--- a/pkg/handlers/primeapi/mto_service_item_test.go
+++ b/pkg/handlers/primeapi/mto_service_item_test.go
@@ -92,7 +92,7 @@ func (suite *HandlerSuite) TestCreateMTOServiceItemHandler() {
 
 	})
 
-	suite.T().Run("POST failure - 412", func(t *testing.T) {
+	suite.T().Run("POST failure - 422 Unprocessable Entity Error", func(t *testing.T) {
 		mockCreator := mocks.MTOServiceItemCreator{}
 		handler := CreateMTOServiceItemHandler{
 			handlers.NewHandlerContext(suite.DB(), suite.TestLogger()),
@@ -107,6 +107,23 @@ func (suite *HandlerSuite) TestCreateMTOServiceItemHandler() {
 
 		response := handler.Handle(params)
 		suite.IsType(&mtoserviceitemops.CreateMTOServiceItemUnprocessableEntity{}, response)
+	})
+
+	suite.T().Run("POST failure - 409 Conflict Error", func(t *testing.T) {
+		mockCreator := mocks.MTOServiceItemCreator{}
+		handler := CreateMTOServiceItemHandler{
+			handlers.NewHandlerContext(suite.DB(), suite.TestLogger()),
+			&mockCreator,
+		}
+		// ConflictError should generate a Conflict response
+		err := services.ConflictError{}
+
+		mockCreator.On("CreateMTOServiceItem",
+			mock.Anything,
+		).Return(nil, nil, err)
+
+		response := handler.Handle(params)
+		suite.IsType(&mtoserviceitemops.CreateMTOServiceItemConflict{}, response)
 	})
 
 	suite.T().Run("POST failure - 404", func(t *testing.T) {
--- a/pkg/handlers/primeapi/mto_service_item_test.go
+++ b/pkg/handlers/primeapi/mto_service_item_test.go
@@ -92,12 +92,13 @@ func (suite *HandlerSuite) TestCreateMTOServiceItemHandler() {
 
 	})
 
-	suite.T().Run("POST failure - 400", func(t *testing.T) {
+	suite.T().Run("POST failure - 412", func(t *testing.T) {
 		mockCreator := mocks.MTOServiceItemCreator{}
 		handler := CreateMTOServiceItemHandler{
 			handlers.NewHandlerContext(suite.DB(), suite.TestLogger()),
 			&mockCreator,
 		}
+		// InvalidInputError should generate an UnprocessableEntity response
 		err := services.InvalidInputError{}
 
 		mockCreator.On("CreateMTOServiceItem",
@@ -105,7 +106,7 @@ func (suite *HandlerSuite) TestCreateMTOServiceItemHandler() {
 		).Return(nil, nil, err)
 
 		response := handler.Handle(params)
-		suite.IsType(&mtoserviceitemops.CreateMTOServiceItemBadRequest{}, response)
+		suite.IsType(&mtoserviceitemops.CreateMTOServiceItemUnprocessableEntity{}, response)
 	})
 
 	suite.T().Run("POST failure - 404", func(t *testing.T) {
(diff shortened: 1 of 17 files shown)


pkg/services/move_task_order/move_task_order_fetcher.go
```diff
@@ -58,12 +58,7 @@ func (f moveTaskOrderFetcher) ListAllMoveTaskOrders(isAvailableToPrime bool, sin
 	err = query.All(&moveTaskOrders)
 
 	if err != nil {
-		switch err {
-		case sql.ErrNoRows:
-			return models.MoveTaskOrders{}, services.NotFoundError{}
-		default:
-			return models.MoveTaskOrders{}, err
-		}
+		return models.MoveTaskOrders{}, services.NewQueryError("MoveTaskOrder", err, "Unexpected error while querying db.")
```
shimonacarvalho, line 61: This is not a notFound error because if no moveTaskOrders are found, there is no error from the query command. 
In this case, an empty array is (reasonably) returned.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -68,7 +68,7 @@ func (o *moveTaskOrderUpdater) UpdatePostCounselingInfo(moveTaskOrderID uuid.UUI
 	err := o.builder.FetchOne(&moveTaskOrder, queryFilters)
 
 	if err != nil {
-		return nil, services.NewNotFoundError(moveTaskOrder.ID, "")
+		return nil, services.NewNotFoundError(moveTaskOrderID, "while looking for moveTaskOrder.")
```
shimonacarvalho, line 71: Horrible syntax, filed ticket for fix the messaging on all NotFound errors.

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -116,20 +112,14 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
 			}
 		}
 
-		// create MTOServiceItems List
```
shimonacarvalho, line 119: deleted because this is dead code. The service items are never transferred from payload to model so this never runs. Filed ticket to fix this properly.

pkg/handlers/primeapi/mto_service_item_test.go
```diff
@@ -92,20 +92,21 @@ func (suite *HandlerSuite) TestCreateMTOServiceItemHandler() {
 
 	})
 
-	suite.T().Run("POST failure - 400", func(t *testing.T) {
+	suite.T().Run("POST failure - 412", func(t *testing.T) {
```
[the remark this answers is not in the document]
chrisrcoles: We did add another error `ConflictError` - should we add an integration test for capturing that error too?

pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -112,8 +111,10 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
 		return nil
 	})
 
-	if verrs != nil || err != nil {
-		return nil, verrs, err
+	if verrs != nil && verrs.HasAny() {
```
chrisrcoles, line 114: nice! we should really be doing this level of error checking everywhere! this is great

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -116,20 +112,14 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
...
-					return fmt.Errorf("%#v %e", verrs, error)
-				}
-				serviceItems[i] = *serviceItem
-			}
-		}
 		return nil
 	})
 
-	return shipment, err
+	if verrs != nil && verrs.HasAny() {
```
chrisrcoles, line 118: I wonder if we should make this a function since we will re-use it

chrisrcoles:
I also noticed some `server_test(s)` weren't passing - can you also take a look at the failing tests:
```
TestAuthSuite - authentication
Failed
2020/06/30 02:57:44 package auth/authentication is attempting to connect to database test_db
2020/06/30 02:57:44 attempting to truncate the database test_db
attempting to clone database test_db to test_db_auth_authentication_pejrr6... success
2020/06/30 02:57:45 package auth/authentication is attempting to connect to database test_db_auth_authentication_pejrr6
TestAuthSuite/TestAuthorizeUnknownUserAdminLogsIn - authentication
Failed
    TestAuthSuite/TestAuthorizeUnknownUserAdminLogsIn: auth_test.go:928: 
```

pkg/handlers/primeapi/mto_service_item_test.go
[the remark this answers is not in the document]
shimonacarvalho: You're right! It's 422 and I will add a test for ConflictError

pkg/services/mto_shipment/mto_shipment_creator.go
shimonacarvalho: Hmm not sure what you mean Chris? which part as a function? The if statement returning the error? That I think not because it's already hard enough to reason about buried error handling, I'd rather keep the errors close to/visible in the code that generated them. But not sure if that's what you meant.

chrisrcoles: sounds good!

PR #4349, opened 2020-06-30
Kim/mb 3131 convert to ghc flag

kimallen:
## Description

This sets up a feature flag for when we convert old tables to new GHC tables.

## Reviewer Notes

I don't think this can be tested until the next story is written?

I chose to name this `FeatureFlagConvertProfileOrdersToGHC`.  I'm sure there's better. . .feel free to make suggestions if you have a preference for something different.

I plan to add instructions in the Wiki, though wanted to get this in first so that it doesn't block.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3133) for this change


pkg/cli/featureflag.go
```diff
@@ -12,13 +12,16 @@ const (
 	FeatureFlagRoleBasedAuth string = "feature-flag-role-based-auth"
 	// FeatureFlagConvertPPMsToGHC is the convert-ppms-to-ghc feature flag
 	FeatureFlagConvertPPMsToGHC string = "feature-flag-convert-ppms-to-ghc"
+	// FeatureFlagConvertProfileOrdersToGHC is the convert-ppms-to-ghc feature flag
```
sojeri, line 15: why profile orders to GHC and not something like PPMsToGHCv2 with a clarifying comment?

kimallen: To me, it seemed to distinguish better between the two similar fns without having to remember which flag belonged where.  The conversion converts the info entered from the profile and the orders.  `ConvertFromPPMToGHC` converts the info from those plus the PPM.  At the point this is going to be used, there isn't actually any PPM yet anyway, so its not actually related to a PPM.  If you have a strong preference for v2, I can change it.  As Lindsay said, we will be deleting it as soon as the db reconciliation work is done, so I don't feel strongly.

PR #4358, opened 2020-07-02
MB-3156 add UI for SelectMoveType component (behind feature flag)

SirenaBorracha:
## Description

Adds a page in the SM move signup flow to select a move type - HHG or PPM. This page appears after the transition to move page, has a disabled radio button for HHG (since this is the first page for the HHG feature), and continues to the PPM sign up flow. The select move type page is behind the hhgFlow feature flag, which is checked at the getWorkflowRoutes level, in the context that is now being accessed in the index and Landing pages.

Also:
* Added a constant for the PPM move type to `shared/constants.js` and updated accordingly.
* Cleanup work for the migrate orders to entities work that was merged recently (https://github.com/transcom/mymove/pull/4239) - namely, the uploads need to now be passed into funcs in getWorkflowRoutes and checked in entities and also on the old orders.

## Reviewer Notes

Since I (Andrea) shouldn't be able to accept my own PR (:)), I added myself as an Assignee. 

## Setup

```sh
make server_run
make client_run
```

Create a new SM user and go through the flow. You should see the select move type page appear after the "transition to move" page. You should not be able to select "Have Professionals pack and move it all", and hitting "Next" should take you to the "dates and location" PPM page. 

To verify this is behind a feature flag, you can change the value to "false" for the 'development' key in "environmentFlags" in `src/shared/featureFlags.js`. If hhgFlow is set to false, you should not see the select move page and be routed to the PPM Date and Location page after Transition to Move.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3156) for this change

## Screenshots

Screen Shot 2020-07-09 at 2 07 37 PM


src/scenes/Moves/Hhg/SelectMoveType.jsx
```diff
@@ -0,0 +1,65 @@
+import React, { Component } from 'react';
+import RadioButton from 'shared/RadioButton';
```
kimallen, line 2: I believe we want to be using the react-USWDS storybook components here for form elements.
https://trussworks.github.io/react-uswds/?path=/story/forms-form--text-input-form
https://ustcdp3.slack.com/archives/CTNTFJSBA/p1594073491139900

src/shared/constants.js
```diff
@@ -47,6 +47,7 @@ export const UPLOAD_SCAN_STATUS = {
 
 export const SHIPMENT_TYPE = {
```
kimallen, line 48: I'm wondering if we should call this `SHIPMENT_OPTIONS` since a PPM isn't technically a shipment type.  When we go to handle that value, it may be more clear that they don't all direct map to MTOShipment types.

src/scenes/Moves/Hhg/SelectMoveType.jsx
amitch23: Good catch! I brought this up at the frontend checkin today and got this feedback, so updated it just now

src/shared/RadioButton/index.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { string, func, bool } from 'prop-types';
 import { uniqueId } from 'lodash';
 
-const RadioButton = ({ name, label, onChange, value, checked, inputClassName, labelClassName, testId }) => {
+const RadioButton = ({ name, label, onChange, value, checked, disabled, inputClassName, labelClassName, testId }) => {
```
kimallen, line 5: Doesn't hurt to have this here, but don't necessarily need it now that you are using the react-uswds buttons.

src/shared/constants.js
amitch23: Good point - changed to OPTIONS

src/pages/MyMove/SelectMoveType.jsx
```diff
@@ -0,0 +1,36 @@
...
+      </div>
+      <div className="grid-row grid-gap">
+        <div className="grid-col-9 desktop:grid-col-12">
+          <Radio
+            id={SHIPMENT_OPTIONS.HHG}
+            label="Have professionals pack and move it all"
+            value={SHIPMENT_OPTIONS.HHG}
+            name="moveType"
+            disabled
+          />
```
suzubara, line 29: You should be able to render the two `<Radio />` components one right after the other instead of each in separate grid `div`s. That will help make sure vertical spacing is consistent.

src/pages/MyMove/SelectMoveType.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { Radio } from '@trussworks/react-uswds';
+
+import { SHIPMENT_OPTIONS } from 'shared/constants';
+
+export const SelectMoveType = () => (
+  <div className="grid-container usa-prose">
```
suzubara, line 7: It looks like WizardPage content is already rendered inside a `<div className="grid-container usa-prose">` so this one can be removed.

src/pages/MyMove/SelectMoveType.jsx
```diff
@@ -0,0 +1,36 @@
...
+    <div className="usa-grid">
+      <div className="grid-row grid-gap">
```
suzubara, line 9: Since there is only one column in this grid layout, `grid-gap` is not necessary.

src/pages/MyMove/SelectMoveType.jsx
```diff
@@ -0,0 +1,36 @@
...
+        <h1 className="sm-heading">How do you want to move your belongings?</h1>
```
suzubara, line 10: I believe all child elements of a `grid-row` should be contained within a `grid-col` (so this header can move into the `grid-col` div)

src/pages/MyMove/SelectMoveType.jsx
```diff
@@ -0,0 +1,36 @@
...
+        <div className="grid-col-9 desktop:grid-col-12">
```
suzubara, line 11: Is this column specifically supposed to be fewer columns at sizes below desktop? The grid column sizes themselves are responsive. I suspect this can just be `<div className="grid-col">`

src/pages/MyMove/SelectMoveType.test.jsx
```diff
@@ -0,0 +1,20 @@
...
+    const wrapper = shallow(<SelectMoveType />);
+    expect(wrapper.find(Radio).length).toBe(2);
+
+    // PPM button should be checked on page load
+    expect(wrapper.find(Radio).at(0).dive().text()).toContain('Arrange it all yourself');
+    expect(wrapper.find(Radio).at(0).dive().find('.usa-radio__input').html()).toContain('checked');
+
+    // HHG button should be disabled
+    expect(wrapper.find(Radio).at(1).dive().text()).toContain('Have professionals pack and move it all');
+    expect(wrapper.find(Radio).at(1).dive().find('.usa-radio__input').html()).toContain('disabled');
```
suzubara, line 18: Instead of using `dive()` in every expect statement, you can use `mount` instead of `shallow` which will fully mount nested components, and then you can remove the `.dive()` calls

src/scenes/Landing/index.jsx
```diff
@@ -149,6 +163,14 @@ export class Landing extends Component {
   }
 }
 
+Landing.propTypes = {
+  context: PropTypes.shape({
+    flags: PropTypes.shape({
+      hhgFlow: PropTypes.bool,
```
suzubara, line 169: Since this is a new feature flag it may make sense for this component to have default props set to `false` instead of requiring them

src/scenes/MyMove/getWorkflowRoutes.jsx
```diff
@@ -154,8 +158,19 @@ const pages = {
       );
     },
   },
+  '/moves/:moveId/select-type': {
+    isInFlow: inHhgFlow,
+    isComplete: always,
+    render: (key, pages) => () => {
+      return (
+        <WizardPage handleSubmit={no_op} pageList={pages} pageKey={key}>
```
suzubara, line 166: You may want to render `<WizardPage>` inside of `SelectMoveType` so that you have control over the whether the Next button is enabled (i.e., only if the form is valid)

src/shared/constants.js
```diff
@@ -45,8 +45,9 @@ export const UPLOAD_SCAN_STATUS = {
   PROCESSING: 'PROCESSING',
 };
 
-export const SHIPMENT_TYPE = {
+export const SHIPMENT_OPTIONS = {
```
suzubara, line 48: just curious why the rename?

amitch23: https://github.com/transcom/mymove/pull/4358#discussion_r451804129

src/pages/MyMove/SelectMoveType.test.jsx
amitch23: Oh man, this is what I was looking for yesterday and didn't know how to get nested components! thanks!

src/scenes/MyMove/getWorkflowRoutes.jsx
amitch23: I originally had `SelectMoveType` rendering the WizardPage inside of itself, but when I added the Radio buttons and didn't need to handle state, the linter wanted me to change the component structure. Since this ticket doesn't hook up the component to the API, there was some work to do with formik and the wizard, and there's only one selectable option right now (i.e. PPM), I figured this could be done in a separate task when we actually want to update the move with the selected type.

PR #4368, opened 2020-07-07
MB-3157 Handle distance out of bounds for pricing DLH and DSH

gidjin:
## Description

We discovered that `Zip5TransitDistance` and `Zip3TransitDistance` had checks in them for distance minimums but those checks do not apply to all places we need to call `Zip5TransitDistance` and `Zip3TransitDistance` from. This PR removes the check from `Zip3TransitDistance` since it hasn't been used yet, and introduces a rename to `Zip5TransitDistance` since it has a check for line haul distance minimum and adds a `Zip5TransitDistance` that doesn't do the check.

This PR didn't move the check out of the original `Zip5TransitDistance` as it's called in a lot of places and it wasn't always clear if the error should be checked and returned there or not. I would discourage further use of the new `Zip5TransitDistanceLineHaul` in favor of not putting the checks in the planner code. 

## Reviewer Notes

Is the updated wiki page mentioned below clear?

## Setup

To test locally

```sh
make db_dev_e2e_populate
make server_run
make client_run
```

Follow the steps in the [acceptance testing payment requests wiki page](https://github.com/transcom/mymove/wiki/Acceptance-Testing-Payment-Requests#creating-a-new-mto-with-shipments) to setup a brand new MTO with a DLH and DSH.

* Once the MTO is sent to the prime, and the shipments are approved
* Update the actual weight on both shipments (no need to update estimated weight it will already be set)
* Create a payment request for both DLH and DSH services.
* Verify the price is correct

* [x] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/wiki/backend#logging)
* [x] The requirements listed in
 have been satisfied.

## References

* [MB-3157](https://dp3.atlassian.net/browse/MB-3157) for this change
* [this wiki page](https://github.com/transcom/mymove/wiki/Acceptance-Testing-Payment-Requests) explains more about the approach used.


garrettqmartin8:
@gidjin started going through the [instructions](https://github.com/transcom/mymove/wiki/Acceptance-Testing-Payment-Requests) but I ran into an error that may or may not have to do with the changes you've made:

```
route/here_planner.go:178	Got non-200 response from HERE routing.	{"git_branch": "MB-3157_handle_distance_out_of_bounds_for_pricing_dlh_and_dsh", "git_commit": "74cc41e877519dcb423c6eec065cd4cfb1ba51cf", "http_status": 400, "here_error": "{\"_type\":\"ns2:RoutingServiceErrorType\",\"type\":\"ApplicationError\",\"subtype\":\"NoRouteFound\",\"details\":\"Error is NGEO_ERROR_ROUTE_NO_START_POINT\",\"additionalData\":[{\"key\":\"error_code\",\"value\":\"NGEO_ERROR_ROUTE_NO_START_POINT\"}],\"metaInfo\":{\"timestamp\":\"2020-07-08T20:08:17Z\",\"mapVersion\":\"8.30.110.152\",\"moduleVersion\":\"7.2.202027-7399\",\"interfaceVersion\":\"2.6.76\",\"availableMapVersion\":[\"8.30.110.152\"]}}\n"}
```

It doesn't prevent me from submitting the move, but the estimated incentive doesn't get calculated. From the TOO side, I can't see the move that I created.

At first I thought maybe it was an issue with upgrading the HERE API URLs, but it works fine when I create a new move in staging.

pearl-truss:
@gidjin I'm looking at your PR now. I'm getting an internal server error when I try using the prime client api on the create payment step https://github.com/transcom/mymove/wiki/Acceptance-Testing-Payment-Requests#create-a-payment-request. Is it right that I should replace the mtoTaskOrderID and serviceItems ids with the ids from the Customer Details Page (what I see after sending the approved service items to the Prime in the UI)?

gidjin:
> @gidjin started going through the [instructions](https://github.com/transcom/mymove/wiki/Acceptance-Testing-Payment-Requests) but I ran into an error that may or may not have to do with the changes you've made:
> 
> ```
> route/here_planner.go:178	Got non-200 response from HERE routing.	{"git_branch": "MB-3157_handle_distance_out_of_bounds_for_pricing_dlh_and_dsh", "git_commit": "74cc41e877519dcb423c6eec065cd4cfb1ba51cf", "http_status": 400, "here_error": "{\"_type\":\"ns2:RoutingServiceErrorType\",\"type\":\"ApplicationError\",\"subtype\":\"NoRouteFound\",\"details\":\"Error is NGEO_ERROR_ROUTE_NO_START_POINT\",\"additionalData\":[{\"key\":\"error_code\",\"value\":\"NGEO_ERROR_ROUTE_NO_START_POINT\"}],\"metaInfo\":{\"timestamp\":\"2020-07-08T20:08:17Z\",\"mapVersion\":\"8.30.110.152\",\"moduleVersion\":\"7.2.202027-7399\",\"interfaceVersion\":\"2.6.76\",\"availableMapVersion\":[\"8.30.110.152\"]}}\n"}
> ```
> 
> It doesn't prevent me from submitting the move, but the estimated incentive doesn't get calculated. From the TOO side, I can't see the move that I created.
> 
> At first I thought maybe it was an issue with upgrading the HERE API URLs, but it works fine when I create a new move in staging.

@garrettqmartin8 I haven't seen that, maybe there is an outage with HERE when you tried it? I'm happy to pair and see if we can find what's going on.

gidjin:
> @gidjin I'm looking at your PR now. I'm getting an internal server error when I try using the prime client api on the create payment step https://github.com/transcom/mymove/wiki/Acceptance-Testing-Payment-Requests#create-a-payment-request. Is it right that I should replace the mtoTaskOrderID and serviceItems ids with the ids from the Customer Details Page (what I see after sending the approved service items to the Prime in the UI)?

Yes taking the MTO ID and the Service Item ids from the details page should work. I would check that you set the actual weight on both shipments, and if that is true look at the server log to see what caused the 500. I'm happy to pair if you need help.

pkg/route/planner.go
```diff
@@ -46,6 +46,22 @@ func urlencodeAddress(address *models.Address) string {
 }
 
 func zip5TransitDistanceHelper(planner Planner, source string, destination string) (int, error) {
```
jacquelineIO, line 48: Can you add some comments to each of these functions to say what they are for.

jacquelineIO, line 48: Can you add some comments to each of these functions to say what they are for.

pkg/route/planner.go
```diff
@@ -46,6 +46,22 @@ func urlencodeAddress(address *models.Address) string {
...
+		return 0, err
+	}
+	distance, err := planner.LatLongTransitDistance(sLL, dLL)
+	if err != nil {
+		return 0, err
+	}
+	return distance, err
+}
+
+func zip5TransitDistanceLineHaulHelper(planner Planner, source string, destination string) (int, error) {
```
jacquelineIO, line 64: So this function is NOT for short hual, just for line haul.. is this used by PPM? (I know we talked about this, but I'm not remembering right now)

pkg/route/planner.go
```diff
@@ -77,16 +93,15 @@ func zip3TransitDistanceHelper(planner Planner, source string, destination strin
 	if err != nil {
 		return 0, err
 	}
-	if distance < 50 {
```
jacquelineIO, line 80: So now this function is intended for use by linehaul HHG (is PPM using this one?)

pkg/route/planner.go
gidjin: Yes, this function represents what is in master now, and is used by the PPM flow. It is setup to error out if the distance calculated is less than 50. However, in HHG we need to use zip5 calculations for short haul (ie less than 50) so it would always fail for that. So this one is not for short haul. I hope that one day we can refactor out this function, but that work was too big to tackle in this PR.

pkg/route/planner.go
gidjin: Correct, this function is intended for HHG line haul. PPM flow does not use it as it was added as part of the new GHC rate engine work. However, when it was added this check was copied over and I think checking for this business logic here is not the best place.

PR #4375, opened 2020-07-08
Create second convert function but without shipment creation

kimallen:
## Description

This creates a (temp) function for converting the old tables for orders and moves to `move_orders` and `move_task_orders` to use until the database reconciliation is finished. 

Because there is a similar function run at the end of the PPM submission, and we don't want a customer to have more than one active `move_order`, I also put a check in the `ConvertFromPPMToGHC` fn so that it uses any existing move order instead of creating a new one.

## Setup

`make server_test` or `go test .pkg/services/converthelper`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3132) for this change


pkg/services/converthelper/convert.go
```diff
@@ -36,36 +36,47 @@ func ConvertFromPPMToGHC(db *pop.Connection, moveID uuid.UUID) (uuid.UUID, error
...
-	mo.DateIssued = &orders.IssueDate
-	mo.ReportByDate = &orders.ReportByDate
-	mo.LinesOfAccounting = orders.TAC
-
-	if err := db.Save(&mo); err != nil {
-		return uuid.Nil, fmt.Errorf("Could not save move order, %w", err)
+	var moveOrders []models.MoveOrder
+	err := db.Where("customer_id = $1", sm.ID).All(&moveOrders)
+	if err != nil {
+		return uuid.Nil, fmt.Errorf("Could not fetch move order, %w", err)
```
amitch23, line 43: If there are any moves in live envs that haven't yet submitted their PPMs (i.e. there are no move_orders and move_task_orders for the respective orders and move), then this will error. Do we know if we need to handle these cases?

pkg/services/converthelper/convert.go
```diff
@@ -36,36 +36,47 @@ func ConvertFromPPMToGHC(db *pop.Connection, moveID uuid.UUID) (uuid.UUID, error
...
+	if len(moveOrders) == 0 {
+		orders := move.Orders
+		mo.CreatedAt = orders.CreatedAt
+		mo.UpdatedAt = orders.UpdatedAt
+		mo.Customer = &sm
+		mo.CustomerID = &sm.ID
+		mo.DestinationDutyStation = &orders.NewDutyStation
+		mo.DestinationDutyStationID = &orders.NewDutyStationID
+
+		orderType := "GHC"
```
amitch23, line 55: I know this was hardcoded in here already, but I'm wondering if it would be better to use the enum value in the ghc messages (if applicable) - https://github.com/transcom/mymove/blob/master/pkg/gen/ghcmessages/move_order.go#L231

pkg/services/converthelper/convert.go
```diff
@@ -36,36 +36,47 @@ func ConvertFromPPMToGHC(db *pop.Connection, moveID uuid.UUID) (uuid.UUID, error
...
+		mo.OrderNumber = orders.OrdersNumber
+		mo.OrderType = &orderType
+		orderTypeDetail := "TBD"
```
amitch23, line 58: I'm a little confused about this value, "TBD". In the ghc.yaml swagger, ordertype is set to a string, whereas orderTypeDetail is set to an enum of "GHC" or "NTS" (https://github.com/transcom/mymove/blob/master/swagger/ghc.yaml#L1461). Maybe I'm looking at the wrong reference, but do you know the difference between Order type and order type detail here?

pkg/services/converthelper/convert.go
kimallen: Does `.All` return an error if there are no records?  I was thinking it returned an empty struct.  If it returns an error, you're right, that's not what I want to do.  I want to check if the service member has any existing `move_orders`, and if they don't then I want to create one as I do below.

pkg/services/converthelper/convert.go
kimallen: I think there's some alignment that needs to happen between GHC and internal.yaml for `ordersType` and `ordersTypeDetail`.  I was assuming that since these values had previous approval, that they worked for the purpose of their creation.  This is something that I think needs discussion, but  could just set it to GHC or NTS so that it fits both type parameters.

pkg/services/converthelper/convert.go
gidjin: I created a PPM move locally with an empty database and no errors showed up. Seems like it's working, so I'm thinking it returns an empty list for no data found.

kimallen: This isn't actually called yet -it's just the function, so it wouldn't impact the app yet, but I did write a test for it which passed.

amitch23: Ah, I didn't realize that about `.All`, my mistake!

pkg/services/converthelper/convert.go
kimallen: Probably that's better practice (though I think the past pattern we've used is to set the enum constants in the model and use those).  Since there seems to be inconsistency/question about what those enum values should be for `OrderTypeDetail`, and this fn is only going to be around for a couple weeks, I went ahead and hardcoded some more realistic values, but am inclined to wait for certainty on the enum values to use them.

pkg/services/converthelper/convert.go
kimallen: Looks like our yaml files need alignment, and what is in `internal.yaml` is correct and not `ghc.yaml`, but I think that should be done in another PR (perhaps the endpoint one). Are you okay with that? https://ustcdp3.slack.com/archives/CP74RNQ9L/p1594246968055100

pkg/services/converthelper/convert.go
amitch23: Ok - sound good!

kimallen: dope!  It actually requires me to use the existing enums.  Changing `orderTypeDetail` values back for now.

PR #4379, opened 2020-07-09
[MB-3128] Service Object and method to check if MTO is available to prime

Ronolibert:
## Description

Adds a service object and method for checking if an MTO is available to prime so we can provide an error if we are trying to edit a record that is not available to prime

## Reviewers Notes
- I ended up keeping the MTOAvailableToPrime on the MTOShipment service object but renamed it to `MTOShipmentsMTOAvailableToPrime` because judging by the tests, there is a situation where they try to update the MTO Shipment without providing the MTO id. Since the MTOChecker's method relies on the MTO ID only then this wouldn't work in this scenario

## Setup
`make server_test`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3128) for this change


pkg/services/move_task_order/move_task_order_checker.go
```diff
@@ -0,0 +1,41 @@
...
+		switch err {
+		case sql.ErrNoRows:
+			return false, services.NewNotFoundError(moveTaskOrderID, "")
+		default:
+			return false, err
+		}
+	}
+
+	if mto.AvailableToPrimeAt == nil {
+		return false, services.NewInvalidInputError(mto.ID, nil, nil, "MTO is not available to prime")
```
sandy-wright, line 37: We decided that we would return a `Not Found Error` when the prime tries to update a non-available object so that we don't reveal too much info about our db -- should we change the error here or let the handler check and set the `Not Found Error` even when this function returns an `Invalid Input Error`?

Ronolibert: Updated this to be a not found error instead!

PR #4380, opened 2020-07-09
Refactor createPaymentRequestHandler to use switch statement

kimallen:
## Description

Reggie pointed out in a PR awhile back that it would be cleaner to use a switch in `CreatePaymentRequestHandler` and it bothers me every time I look at it that I didn't refactor, so finally just did it.  This puts the possible error types into a switch statement.

## Setup

Tests should still pass:
`go test ./pkg/handlers/primeapi`

## References

* No Jira story for this change- this is a rogue fix.

### DIFF
--- a/pkg/handlers/primeapi/payment_request.go
+++ b/pkg/handlers/primeapi/payment_request.go
@@ -79,33 +79,33 @@ func (h CreatePaymentRequestHandler) Handle(params paymentrequestop.CreatePaymen
 	createdPaymentRequest, err := h.PaymentRequestCreator.CreatePaymentRequest(&paymentRequest)
 	if err != nil {
 		logger.Error("Error creating payment request", zap.Error(err))
-		if typedErr, ok := err.(services.InvalidCreateInputError); ok {
-			verrs := typedErr.ValidationErrors
+		switch e := err.(type) {
+		case services.InvalidCreateInputError:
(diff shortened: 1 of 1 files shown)


pkg/handlers/primeapi/payment_request.go
```diff
@@ -79,33 +79,33 @@ func (h CreatePaymentRequestHandler) Handle(params paymentrequestop.CreatePaymen
...
 				zap.Any("payload", payload))
 			return paymentrequestop.NewCreatePaymentRequestNotFound().WithPayload(payload)
-		}
-		if _, ok := err.(*services.BadDataError); ok {
+		case *services.BadDataError:
 			payload := payloads.ClientError(handlers.SQLErrMessage, err.Error(), h.GetTraceID())
 
 			logger.Error("Payment Request",
 				zap.Any("payload", payload))
 			return paymentrequestop.NewCreatePaymentRequestBadRequest().WithPayload(payload)
```
shimonacarvalho, line 103: Bad request generally means we couldn't parse the request, which would be unlikely to realize in the handler. However I think that the internals of payment request are using this bad data error for a variety of things and it would be a big change to update. Oh and we have a separate ticket https://dp3.atlassian.net/browse/MB-3130

So it's probably ok not to tackle that here.

PR #4387, opened 2020-07-10
Kim/mb 3133 apply convert to handler

kimallen:
## Description

Adds ConvertProfileOrdersToGHC fn to the create orders handler.  In the SM flow, this means when they submit their initial orders info (the next page is "upload orders" page), the function gets run and creates a corresponding `move_order` for a `order` and a `move_task_order` for a `move`.

## Setup

- Log in locally to SM app with `needs@orde.rs`
- Check the db for number of `move_orders` and `move_task_orders` records.
- Move through the app until you get to the upload orders page.
- You should see a log that says `ProfileOrders->GHC conversion successful. New MoveOrder and MoveTaskOrder created.`. Check the database after getting to that page and see that the number of records increases by 1 for each of those tables.
- Then continue through the flow until you've submitted a PPM.  Check the number of records on each table again and see 1 more `move_task_order` but it should be on the same `move_order`, so there will be no additional `move_orders` records.

If you want to check that the flag works, go the `.envrc` and set `export FEATURE_FLAG_CONVERT_PROFILE_ORDERS_TO_GHC` to false.  Then go through a SM until the order uploads page.  In the logs you should see `ProfileOrders->GHC conversion skipped (env var not set)`. (Remember to change the flag back to true)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3133) for this change


pkg/handlers/internalapi/orders.go
```diff
@@ -117,6 +122,17 @@ func (h CreateOrdersHandler) Handle(params ordersop.CreateOrdersParams) middlewa
 	}
 	newOrder.Moves = append(newOrder.Moves, *newMove)
 
+	// temp create corresponding move order and move task order until db is reconciled
+	if h.HandlerContext.GetFeatureFlag(cli.FeatureFlagConvertProfileOrdersToGHC) {
+		if moID, convertErr := converthelper.ConvertProfileOrdersToGHC(h.DB(), newMove.ID); convertErr != nil {
+			logger.Error("ProfileOrders->GHC conversion error", zap.Error(convertErr))
+			// don't return an error here because we don't want this to break the endpoint
```
lynzt, line 129: don't return an error why?

kimallen: If the conversion doesn't work, we don't want the app user to get blocked, is my understanding.  This is copied from the first conversion error.

kimallen:
> I reset my db and was able to replicate the behavior in the description! :). After I hit the convert ProfileOrders -> GHC handler, I checked my db and found what I expected: a newly-created move_orders and move_task_order associated to my servicemember/customer. However, when I submit the PPM, the convert PPM->GHC handler creates another move task order, like you call out in your description. This means I now have 2 moves for my orders, which seems wrong - shouldn't I only have one? I'm wondering if the PPM->GHC handler should also check for existing move_task_orders (like you did for the orders-> move_orders) instead of creating one regardless, i.e. here: https://github.com/transcom/mymove/blob/master/pkg/services/converthelper/convert.go#L83

I talked with @gidjin a bit about this.  I don't think it's a problem, though I'll check with him.  It will be a moot point soon since the db reconciliation will be done in a week or two, probably.

pkg/handlers/internalapi/orders.go
gidjin: Also the feature flag is only on in staging, experimental, and development as I understand. So users of the converter should be able to find logs if the conversion fails.

kimallen:
Fixed this so that only one MO and one MTO are created, even when both fns are run.  (Thanks @amitch23 for your attention to detail!)

PR #4390, opened 2020-07-11
MB-3113 Consolidate move_orders and orders tables

monfresh:
## Description

This is part of the ongoing effort to reduce duplication between
certain tables, which is causing confusion and pain in implementing
some features.

Because the move_orders table is not used in production, we opted to
consolidate move_orders into orders. In staging and experimental, the
move_orders (and related tables) were populated, and because the
orders table requires a document to be associated with an order, if we
wanted to to preserve the data in staging, that would mean creating
a document for each move_order. In addition, the orders_type column
in move_orders was of type enum, versus varchar in the orders table.
Rather than spend time preserving and converting data, we opted to
delete all the data in staging by using the Postgres TRUNCATE and
CASCADE commands, which clears out the following tables:
move_orders, move_task_orders, mto_service_items, payment_requests,
mto_shipments, payment_service_items, mto_service_item_dimensions,
mto_service_item_customer_contacts, proof_of_service_docs, mto_agents,
and payment_service_item_params.

For the difference in the orders_type column type, we opted to leave
the orders column as a varchar for now, and add the enum values
defined in the move_orders table ("GHC" and "NTS") into the enum
values defined for the orders table in the swagger `internal.yaml`
config. The reasoning behind that is that converting a varchar column
into an enum requires 5 separate deployments for a zero-downtime
migration. We thought it was better to get the tables consolidated
first with the least amount of changes, and then go back and make
changes if desired.

Similarly, when adding columns into orders that previously only
existed in the move_orders table, we preserved the type, even though
it resulted in inconsistencies within the orders table. For example,
the move_orders table used `*DutyStation` for the `OriginDutyStation`
whereas the orders table uses `DutyStation` for the `NewDutyStation`.

Note that this PR affects the Support API:

- When creating a move task order, the associated move order is
expected to be associated with an existing document, so the document
must be created first (tied to the same service member). Other fields
that were required by the orders model and that are now also required
are: status, issue date, and orders type.

## Reviewer Notes

## Setup

```sh
make db_dev_reset && make db_dev_migrate
make clean && make build
```

 have been satisfied.

## References

## Screenshots


pkg/services/converthelper/convert_test.go
```diff
@@ -13,7 +14,30 @@ import (
...
+	suite.MustSave(&entitlement)
+	moveOrder.Entitlement = &entitlement
+	moveOrder.EntitlementID = &entitlement.ID
+	moveOrder.OriginDutyStation = &sm.DutyStation
+	moveOrder.OriginDutyStationID = &sm.DutyStation.ID
+	orderTypeDetail := internalmessages.OrdersTypeDetailHHGPERMITTED
+	moveOrder.OrdersTypeDetail = &orderTypeDetail
+	moveOrder.Grade = (*string)(sm.Rank)
+	suite.MustSave(&moveOrder)
+
```
monfresh, line 40: @kimallen With your recent changes to `convert.go`, if there is an existing order, it doesn't get converted, but because we now only have the orders table, this test assumes that the order has certain fields that previously only existed on the move_orders table. The easiest way to make this test pass without breaking other things was to set those fields as part of the test setup.

Given that we are now consolidating move_orders and orders, does it still make sense to have the order -> move_order conversion? Does the fact that I have to set all these fields in the test expose a bug in the conversion code?

kimallen: I'm not sure I fully understand.  I only am reading this after merging the use of the convert fn(https://github.com/transcom/mymove/pull/4387), so there may be additional changes.  I was expecting that as the db reconciliations happened, we could incrementally remove the tables from conversion in these functions.  Once db-reconciliation is finished, and we have the ability to create an mto_shipment, we should no longer need the conversions (though @gidjin could confirm that for his use of shorthaul mto_shipment)

gidjin: The conversion code is a temporary hack until we have a way to create an MTO and new MTO shipments and walk through the whole flow. I don't think this reveals a bug, but I'm wondering if there is a `testdatagen` function we could use to generate the moveOrder?

pkg/services/converthelper/convert_test.go
```diff
@@ -40,17 +64,17 @@ func (suite *ConvertSuite) TestConvertFromPPMToGHC() {
 	moID, conversionErr := converthelper.ConvertFromPPMToGHC(suite.DB(), move.ID)
 	suite.FatalNoError(conversionErr)
 
-	var mo models.MoveOrder
-	suite.FatalNoError(suite.DB().Eager("Customer", "Entitlement").Find(&mo, moID))
+	var mo models.Order
+	suite.FatalNoError(suite.DB().Eager("ServiceMember", "Entitlement").Find(&mo, moID))
```
gidjin, line 68: Is there a plan to rename `ServiceMember` on orders to `Customer`?

pkg/services/converthelper/convert_test.go
amitch23: I went through the Customer sign up flow, and the Create Orders handler now creates 2 orders for the customer, because of the profileOrders - > GHC converter. (In master, 1 orders AND 1 move orders is created, which is current desired behavior). Since with the elimination of `move_orders` in this PR, I believe desired behavior for the create orders endpoint would be to result in the creation of ONE orders (i.e. removing the creation of another orders in the profile -> GHC conversion - we'll still need an MTO for the move that's created, pending more db reconciliation work for move -> mto). In addition, the create orders handler would now need to set additional values for orders field from the service_member, i.e. `orders.grade` ( `service_member.rank`), `orders.origin_duty_station_id` (`service_member.duty_station_id`), and potentially `orders.entitlement_id` (though this may be created at the submit PPM step).

pkg/services/converthelper/convert_test.go
monfresh: Yes, eventually.

monfresh:
@amitch23 Thanks for the note about the e2e tests. It reminds me to add a note that folks will need to run `make clean && make build` to get things working again. When you make a change to certain files, like the ones in testdatagen, it requires a rebuild. You will also need to run `make db_dev_reset && make db_dev_migrate` before running something like `generate-test-data --named-scenario e2e_basic`.

Ryan-Koch:
> A side effect I'm seeing on the office app side is that the MoveOrder fetcher service object in `services/move_orders/move_order_fetcher.go` is getting a zero value it doesn't like for the association load (using `db.load()`) in `ListMoveOrders()` on line 39 where it's pulling in the origin duty station address. Going to look at it a bit closer and add another comment if I find something here soon. The effect though unfortunately is a 500 on the TOO queue (in local dev).

I had a chance to look at this a little closer. It looks like we have some addresses that are either nil or just have a zero id in our test data for the e2e. That said, I noticed that if I switch that service object function to use `eager()` instead of the `load()` function that things work. Looking through some of pop's source, I'm starting to suspect that there is some variance in how the reflect type calls are being handled between those two paths.

I can commit and push that small change up if you'd like, is essentially doing this:
```
	err := f.db.Eager("ServiceMember","ConfirmationNumber", "NewDutyStation", "OriginDutyStation", "Entitlement").All(&moveOrders)
```

Instead of this:

```
	err := f.db.All(&moveOrders)
....
....
	for i := range moveOrders {
		f.db.Load(&moveOrders[i], "ServiceMember")
		f.db.Load(&moveOrders[i], "ConfirmationNumber")
		f.db.Load(&moveOrders[i], "NewDutyStation")
		f.db.Load(&moveOrders[i].NewDutyStation, "Address")
		f.db.Load(&moveOrders[i], "OriginDutyStation")
		f.db.Load(moveOrders[i].OriginDutyStation, "Address")
		f.db.Load(&moveOrders[i], "Entitlement")
	}
...
...
return moveOrders, nil
```

This happens in both `ListMoveOrders()` and `FetchMoveOrders()` in the aforementioned service object.

migrations/app/schema/20200707153605_consolidate_move_orders_into_orders.up.sql
```diff
@@ -0,0 +1,48 @@
+-- Lock the move_orders and orders tables so we don't have changes while
+-- doing this migration
+LOCK TABLE orders IN SHARE MODE;
```
reggieriser, line 3: You mention locking `move_orders` in the comment, but we're not doing that here because we're not copying over any data.

migrations/app/schema/20200707153605_consolidate_move_orders_into_orders.up.sql
```diff
@@ -0,0 +1,48 @@
...
+-- mto_service_item_customer_contacts, proof_of_service_docs, mto_agents, and
+-- payment_service_item_params.
+TRUNCATE move_orders CASCADE;
+
+-- Add columns that only existed in move_orders to orders.
+ALTER TABLE orders ADD COLUMN confirmation_number text;
+ALTER TABLE orders ADD COLUMN grade text;
+ALTER TABLE orders ADD COLUMN entitlement_id uuid REFERENCES entitlements;
+CREATE INDEX ON orders (entitlement_id);
+ALTER TABLE orders ADD COLUMN origin_duty_station_id uuid REFERENCES duty_stations;
```
reggieriser, line 22: Nit: we can add all the columns in one ALTER TABLE statement.

migrations/app/schema/20200707153605_consolidate_move_orders_into_orders.up.sql
```diff
@@ -0,0 +1,48 @@
...
+-- a document), and because move_orders doesn't have that column, copying
+-- move_orders into orders poses a problem. Luckily, the move_orders table is
+-- empty in prod, so the simplest way forward is to truncate the move_orders
+-- table in staging and experimental, and all references that cascade down from
+-- it, which are all of these tables:
+-- move_task_orders, mto_service_items, payment_requests, mto_shipments,
+-- payment_service_items, mto_service_item_dimensions,
+-- mto_service_item_customer_contacts, proof_of_service_docs, mto_agents, and
+-- payment_service_item_params.
+TRUNCATE move_orders CASCADE;
```
reggieriser, line 15: TRUNCATE with a CASCADE makes me a little nervous.  That list of tables seems OK to me, but do we want to be explicit so we don't somehow truncate a table we don't intend to?

migrations/app/schema/20200707153605_consolidate_move_orders_into_orders.up.sql
```diff
@@ -0,0 +1,48 @@
...
+ALTER TABLE orders ADD COLUMN entitlement_id uuid REFERENCES entitlements;
+CREATE INDEX ON orders (entitlement_id);
+ALTER TABLE orders ADD COLUMN origin_duty_station_id uuid REFERENCES duty_stations;
+CREATE INDEX ON orders (origin_duty_station_id);
+
+-- The order_type column in move_orders was an enum versus a varchar in the
+-- orders table. If we prefer enum to varchar, then we'll need to update the
+-- order_types to add the enum values used by the orders table, as defined in
+-- the swagger internal.yaml. To make this a zero-downtime migration, we need
+-- to perform 6 steps:
```
reggieriser, line 29: Do we have consensus to not use an enum for now?  To avoid additional complexity in this PR, I think I'm OK with dropping the enum since the orders table never had it.  We can change to an enum later if desired.

migrations/app/schema/20200707153605_consolidate_move_orders_into_orders.up.sql
```diff
@@ -0,0 +1,48 @@
...
+-- ALTER TYPE order_types ADD VALUE 'SEPARATION';
+-- ALTER TABLE orders ADD COLUMN order_type order_types;
+
+ALTER TABLE move_task_orders
+    DROP CONSTRAINT move_task_orders_move_order_id_fkey;
+
+ALTER TABLE move_task_orders
+    ADD CONSTRAINT move_task_orders_move_order_id_fkey FOREIGN KEY (move_order_id) REFERENCES orders (id);
+
+DROP TABLE move_orders;
```
reggieriser, line 48: We also can get rid of the `order_types` enum too since it was only used by the `move_orders` table.

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -14,32 +14,25 @@ type moveOrderFetcher struct {
...
-	var moveOrders []models.MoveOrder
-	err := f.db.All(&moveOrders)
+func (f moveOrderFetcher) ListMoveOrders() ([]models.Order, error) {
+	var moveOrders []models.Order
+	err := f.db.Eager(
+		"ServiceMember",
+		"ConfirmationNumber",
+		"NewDutyStation",
+		"OriginDutyStation",
+		"Entitlement",
```
reggieriser, line 24: A couple of things here:

- ConfirmationNumber isn't an association, so nothing to eager load there (sort of surprised Pop doesn't error on that).
- The previous `Load` calls appear to be bringing in `DestinationDutyStation.Address` and `OriginDutyStation.Address`.  We're not eager fetching those here.  Was that an oversight or are they no longer needed?

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -49,25 +42,24 @@ func NewMoveOrderFetcher(db *pop.Connection) services.MoveOrderFetcher {
...
-	err := f.db.Find(moveOrder, moveOrderID)
+func (f moveOrderFetcher) FetchMoveOrder(moveOrderID uuid.UUID) (*models.Order, error) {
+	moveOrder := &models.Order{}
+	err := f.db.Eager(
+		"ServiceMember",
+		"ConfirmationNumber",
+		"NewDutyStation",
+		"OriginDutyStation",
+		"Entitlement",
+	).Find(moveOrder, moveOrderID)
```
reggieriser, line 53: Same association concerns here as with the list fetcher.

monfresh:
@kim That sounds like a bug, and perhaps a missing test. Do we not have an end-to-end test that goes through the whole flow?

reggieriser:
> I'm looking at the `Orders` created when I walk through the flow. I don't see a value for orders.OrdersNumber, orders.DepartmentIndicator, or orders.confirmationNumber. Should those have values?

@kimallen ... I think this is OK, at least in terms of what we're trying to accomplish in this PR.  Both `OrdersNumber` and `DepartmentIndicator` appear to be set later in the office PPM UI after the move has been submitted (this is the case even without using the converter).  The `ConfirmationNumber` field doesn't appear to have had a value yet during onboarding (even before this reconciliation), so we haven't lost anything there.

kimallen:
> > I'm looking at the `Orders` created when I walk through the flow. I don't see a value for orders.OrdersNumber, orders.DepartmentIndicator, or orders.confirmationNumber. Should those have values?
> 
> @kimallen ... I think this is OK, at least in terms of what we're trying to accomplish in this PR. Both `OrdersNumber` and `DepartmentIndicator` appear to be set later in the office PPM UI after the move has been submitted (this is the case even without using the converter). The `ConfirmationNumber` field doesn't appear to have had a value yet during onboarding (even before this reconciliation), so we haven't lost anything there.

Got it- that makes sense.  Sounds good to me!

PR #4396, opened 2020-07-13
[MB-2622] TIO navigate between payment request service items

duncan-truss:
## Description

After a payment request is available from the prime, the TIO can review (approve or reject) all service items from the associated shipments (hhg, nts) and misc service items (counseling fee, shipment management fee).

This component is the sidebar to the document viewer that will display the uploaded payment request pdf or images.  In this PR we order the service items by createdAt and add the footer UI to cycle between the service items and add a placeholder for the sum total of approved amounts.

## Reviewer Notes

I implemented a sorting of the service items by shipment group, where it takes the earliest createdAt date of each service item in each unique shipment and uses that to order the groups.  I think this is the behavior we want because it would be confusing to the user if you were approving or rejecting service items randomly outside of their shipment group.

We could easily adjust this to always put the Basic Service Items with no shipment (Counseling Services and Shipping Management fees) to always come first like it looks like later mocks may want (or stacked on the same screen).

This is something to consider once the data loading for the payment requests index page (moves/:id/payment-requests) becomes clearer.

I noticed too that we should start getting consistent around labels for service items so we aren't calling them different things on the approve MTO form, in the database, and then in the review.

Here are the current database codes and labels:
Screen Shot 2020-07-14 at 5 32 40 PM

## Setup

```shell
make server_run
make office_client_run
```

1. Login as a TIO user to the office app
2. Navigate to the Payment request review URL `/moves/1/payment-requests/a2c34dba-015f-4f96-a38b-0c0b9272e208`
3. The Previous button should be disabled on the first service item for review
4. Use the next button to cycle through the service items
5. At the last item the Next button is set to disabled
6. Clicking previous will let you navigate in reverse order back to the beginning

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2622) for this change
* [Storybook component](http://localhost:6006/?path=/story/too-tio-components-review-service-items)
* [Abstract Mock](https://share.goabstract.com/3cff868b-be81-401c-82fc-3634b1e26b52?collectionLayerId=80f03f6d-4eb8-46ef-908c-ff46af5a525c&mode=design)

## Screenshots

review-service-items-navigate


src/components/Office/ReviewServiceItems/ReviewServiceItems.jsx
```diff
@@ -8,28 +8,92 @@ import { ServiceItemCardsShape } from 'types/serviceItemCard';
 import { ReactComponent as XLightIcon } from 'shared/icon/x-light.svg';
 import ServiceItemCard from 'components/Office/ReviewServiceItems/ServiceItemCard';
 
+// Sort ascending by objects with string iso timestamps
+const dateCreatedSort = (a, b) => {
+  return Date.parse(a.createdAt) - Date.parse(b.createdAt);
+};
+
+const sortByGroup = (serviceItemCards) => {
```
suzubara, line 16: since `sortedCards` never changes in state, this kind of sorting/filtering is probably best done by the parent component. I don't think this component should be responsible for it.

src/components/Office/ReviewServiceItems/ReviewServiceItems.jsx
```diff
@@ -8,28 +8,92 @@ import { ServiceItemCardsShape } from 'types/serviceItemCard';
...
+  // Make a copy so we're not mutating the props
+  const cards = [...serviceItemCards];
+  // Will populate with earliest service item of each shipment id
+  const shipmentOrder = [];
+  // Contains sorted service items keyed by shipment id or undefined for basic items
+  const shipmentServiceItems = {};
+
+  cards.sort(dateCreatedSort);
+
+  cards.map((serviceItem) => {
```
suzubara, line 26: `Array.map` is used to return a new array created from the existing one. If you want to loop through each item of the array and do something contained within the callback, it would be better to use `Array.forEach` (in which the callback is not expected to return anything)

suzubara: I think this whole block can be simplified:
```
cards.forEach((serviceItem) => {
  const { shipmentId } = serviceItem;
  if (!shipmentServiceItems[`${shipmentId}`]) {
    // group all cards by shipment
    shipmentServiceItems[`${shipmentId}`] = cards.filter((item) => item.shipmentId === shipmentId);
    shipmentOrder.push(serviceItem)
  }
})
```

src/components/Office/ReviewServiceItems/ReviewServiceItems.jsx
```diff
@@ -8,28 +8,92 @@ import { ServiceItemCardsShape } from 'types/serviceItemCard';
...
+
+  shipmentOrder.sort(dateCreatedSort);
+
+  const sortedCards = [];
+  shipmentOrder.map((shipment) => {
+    sortedCards.push(...shipmentServiceItems[`${shipment.shipmentId}`]);
+    return true;
+  });
+
+  return sortedCards;
```
suzubara, line 46: just to make sure I'm reading this right, `sortedCards` is meant to be an array of _all_ serviceItems, grouped by shipment, where the shipments are sorted by the earliest created date of its earliest created serviceItem?

suzubara: what should happen if a service item has no `shipmentId`? I see that's the case in some of the test data below..

src/components/Office/ReviewServiceItems/ReviewServiceItems.jsx
duncan-truss: Whoops, yep I moved this from somewhere else and didn't fix it.

src/components/Office/ReviewServiceItems/ReviewServiceItems.jsx
duncan-truss: The service items without a shipment id (move management and counseling) are considered basic service items.  They all get lumped under the same undefined shipment id and appear together.  I considered making it more explicit but it looks like in follow up stories they may appear on the same screen even instead of individually. So they may end up needing to be a special case.

src/components/Office/ReviewServiceItems/ReviewServiceItems.jsx
duncan-truss: Generalized a little and move to shared utils

src/shared/constants.js
```diff
@@ -53,6 +53,23 @@ export const SHIPMENT_OPTIONS = {
   NTS: 'NTS',
 };
 
+export const SERVICE_ITEM_STATUS = {
+  SUBMITTED: 'SUBMITTED',
+  APPROVED: 'APPROVED',
+  REJECTED: 'REJECTED',
+};
+
+export const SERVICE_ITEM_CODES = {
```
Ryan-Koch, line 62: Nice add, should help with consistency

src/components/Office/ReviewServiceItems/ReviewServiceItems.jsx
```diff
@@ -2,34 +2,62 @@ import React, { useState } from 'react';
...
 
-  // debugging
-  // console.log(curServiceItemCard);
+  const handleClick = (index) => {
+    setCardIndex(index);
+  };
 
   return (
-    <div data-testid="ReviewServiceItems" className={styles.ReviewServiceItems}>
+    <div data-cy="ReviewServiceItems" className={styles.ReviewServiceItems}>
```
suzubara, line 26: just a note that it sounds like we will be moving to `data-testid` throughout the codebase

src/shared/utils/serviceItems.js
```diff
@@ -0,0 +1,32 @@
+import { dateSort } from 'shared/utils';
```
suzubara, line 1: do you mind adding this file to a new `src/utils` directory? I think we should avoid adding new code under `/shared` as much as possible, since the lint rules on that directory are much less strict

duncan-truss: That makes sense, I forgot about the folder specific rules.

PR #4408, opened 2020-07-15
[MB-3111] Add loadUser saga to Office App

suzubara:
## Description

This PR represents adding [redux-saga](https://redux-saga.js.org/) to the codebase, and adding one saga to illustrate usage (which replaces the Office App's `getCurrentUserInfo` thunk). The primary purpose of this PR is to introduce sagas as a tool we can continue to use for managing and testing complex data flow patterns.
- The `fetchUser` saga is scoped to the Office App only, by calling the new `LOAD_USER` action instead of the existing `getCurrentUserInfo` thunk
- Set up a `rootSaga` entry point, where any new sagas can be added

## Reviewer Notes

- There should be no effective user-facing changes
- No changes are expected in the customer app
- Verify when you go to the Office App, you are shown the login page, and you are able to successfully log in
- After logging into the Office App, verify you can navigate around, refresh the page, and view the correct pages for your user role
- **This feature should also be explicitly tested in IE11**

If you have redux devtools extension installed, you can use that to verify the saga action is being called on the Office App. In the Office App, `LOAD_USER` is called to kick off the `fetchUser` saga (and the existing `GET_LOGGED_IN_USER_START, etc.` actions are still dispatched afterwards):

image

In the customer app, only `GET_LOGGED_IN_USER_START, etc.` actions are called: 

image

## Setup

Since this PR adds a new library to `package.json`, you will have to run `yarn install` (which should already happen when you run `make client_run`).

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3111) for this change


src/sagas/auth.test.js
```diff
@@ -0,0 +1,101 @@
...
+    it('dispatches the GET_LOGGED_IN_USER_START action', () => {
+      expect(generator.next().value).toEqual(put(getLoggedInActions.start()));
+    });
+
+    it('makes the GetIsLoggedIn API call', () => {
+      expect(generator.next().value).toEqual(call(GetIsLoggedIn));
+    });
+
+    it('dispatches the User is not logged in error action', () => {
+      expect(generator.next(false).value).toEqual(put(getLoggedInActions.error('User is not logged in')));
```
duncan-truss, line 35: Here passing the value false in `generator.next(false).value` resumes like the last `yield call(GetIsLoggedIn)` returned false?

src/sagas/auth.test.js
```diff
@@ -0,0 +1,101 @@
...
+    });
+
+    it('stores the user data in Redux', () => {
+      expect(generator.next().value).toEqual(put(getLoggedInActions.success(testUser)));
+    });
+
+    it('is done', () => {
+      expect(generator.next().done).toEqual(true);
+    });
+  });
```
duncan-truss, line 100: Is it worth adding in an example of using `throw()` since we'll be using try catches now?

src/sagas/auth.test.js
suzubara: yup exactly, when you call `generator.next()` you can pass in whatever value would be returned by `yield`!

src/sagas/auth.test.js
suzubara: yeah that is a good point, it looks like there's been some advancement in saga testing tools so I'm gonna look into those and maybe that'll make it easier to test thrown errors

PR #4414, opened 2020-07-16
[MB-3158] frontend for HHG move setup

kimallen:
## Description

This creates the UI for the HHG setup form in which data can be entered.  Creates an `AddressField`, `ContactField`, and `HHGDetailsForm` component, each with a storybook version and client tests.
It does not include styling and a couple subtitles, so does not match the wireframe to the T.

## Reviewer Notes

This story is a blocker for other stories, so if there are any visual change request that are more than minimal, that could be put in a different PR so that at least the form is available for the blocked story.

The linter was complaining about using the same name as the default export of `HHGDetailsForm` in the `HHGMoveSetup` page, so We are importing as `HHGDetailsFormComponent`, but wondered if there was a way not to have to change the name?

**UPDATE 6/17/20**: Routing is complicated with both client tests and e2e tests, which check that we are getting the right routes in the flow, while not actually having a way to identify if a move's selected move type is an HHG.  The temp solution is to have the tests pass, and while you are testing locally, you need to change the `inFlow` value for the page to just read `inHhgFlow` (in `getWorkflowRoutes.jsx`) to view the UI.

## Setup

run storybook:
`make storybook`
Check out:
 Component --> HHGDetailsForm
Component --> AddressFields
Component --> ContactInfoFields

locally, run the client and server. Log in as a service member with a move.
Navigate to `moves/[moveId]/hhg-start` and try out the form

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3158) for this change

## Screenshots

hhgsetup


src/components/form/ContactInfoFields/ContactInfoFields.test.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
+import { ContactInfoFields } from './ContactInfoFields';
+
+const wrapper = mount(<ContactInfoFields />);
+
+describe('ContactInfoFields component', () => {
+  it('renders', () => {
+    expect(wrapper.find('ContactInfoFields').length).toBe(1);
```
Ronolibert, line 10: I don't think we need this test since we have something similair in `src/components/Customer/HHGDetailsForm.test.jsx`. Instead, we could test for the right amount of inputs in it instead

src/components/form/AddressFields/AddressFields.test.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
+import { AddressFields } from './AddressFields';
+
+const wrapper = mount(<AddressFields />);
+
+describe('AddressFields component', () => {
+  it('renders', () => {
+    expect(wrapper.find('AddressFields').length).toBe(1);
```
Ronolibert, line 10: I don't think we need this test since we have something similair in `src/components/Customer/HHGDetailsForm.test.jsx`. Instead, we could test for the right amount of inputs in it instead

src/components/Customer/HHGDetailsForm.jsx
```diff
@@ -0,0 +1,144 @@
...
+import { Fieldset, Radio, Label } from '@trussworks/react-uswds';
+
+import { Form } from '../form/Form';
+import { DatePickerInput, TextInput } from '../form/fields';
+import { AddressFields } from '../form/AddressFields/AddressFields';
+import { ContactInfoFields } from '../form/ContactInfoFields/ContactInfoFields';
+import { WizardPage } from '../../shared/WizardPage';
+
+class HHGDetailsForm extends Component {
+  constructor(props) {
```
Ronolibert, line 13: [nit] Instead of defining the state in the constructor, we could instead define it as a property of the class to reduce the boilerplate like so:
```
class HHGDetailsForm extends Component {
  state = {
    hasDeliveryAddress: false,
  }
}
```

Ronolibert:
I can't seem to get the same lint error when I rename `HHGDetailsFormComponent` to `HHGDetailsForm`. There should be nothing wrong with importing the same name as the default

src/components/Customer/HHGDetailsForm.jsx
kimallen: I had done that, but the linter complained, but I'll try again to make sure.

src/pages/MyMove/HHGMoveSetup.jsx
```diff
@@ -0,0 +1,24 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+import HHGDetailsFormComponent from 'components/Customer/HHGDetailsForm';
+
+// eslint-disable-next-line react/prefer-stateless-function
+export class HHGMoveSetup extends Component {
```
Ronolibert, line 7: Could we just have one component instead of an extra layer in between the route and the component with all the fields in it? I'd suggest moving everything in `HHGDetailsForm` into this component instead so we reduce passing down props 1 layer

src/components/form/ContactInfoFields/ContactInfoFields.test.jsx
kimallen: Good point

src/components/Customer/HHGDetailsForm.jsx
kimallen: I tried again, and eslint failed with:
` 13:3  error  State initialization should be in a constructor  react/state-in-constructor`

kimallen:
> I can't seem to get the same lint error when I rename `HHGDetailsFormComponent` to `HHGDetailsForm`. There should be nothing wrong with importing the same name as the default

I checked, and it wasn't a lint error, but the UI was erroring (failed to compile)
image

Moot point if we combine route page with the form component, but baffling. . .

src/pages/MyMove/HHGMoveSetup.jsx
kimallen: Yeah, seems unnecessary to have an extra component layer just for the route.  That said, the reason we put the form into its own component was so that we could isolate it in storybook.  Maybe it's okay to put a route page into storybook?  I'm interested in hearing an opinion from @suzubara (who suggested we put the route page under the pages directory.

src/components/Customer/HHGDetailsForm.test.jsx
```diff
@@ -0,0 +1,20 @@
...
+import HHGDetailsForm from './HHGDetailsForm';
+
+describe('HHGDetailsForm component', () => {
+  it('renders expected form components', () => {
+    const wrapper = mount(<HHGDetailsForm pageList={['page1', 'anotherPage/:foo/:bar']} pageKey="page1" />);
+    expect(wrapper.find('HHGDetailsForm').length).toBe(1);
+    expect(wrapper.find('DatePickerInput').length).toBe(2);
+    expect(wrapper.find('AddressFields').length).toBe(1);
+    expect(wrapper.find('ContactInfoFields').length).toBe(2);
+    expect(wrapper.find('TextInput').length).toBe(1);
```
suzubara, line 13: minor nit, I think it would be a little more meaningful to check that the form renders a `TextInput` with a specific `name` value

src/components/form/AddressFields/AddressFields.test.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
+import { AddressFields } from './AddressFields';
+
+const wrapper = mount(<AddressFields />);
+
+describe('AddressFields component', () => {
+  it('renders', () => {
+    expect(wrapper.find('input[data-cy="mailingAddress1"]').length).toBe(1);
```
suzubara, line 10: it sounds like we'll be changing over to `data-testid`

src/pages/MyMove/HHGMoveSetup.jsx
suzubara: I think it's valuable to have the form be its own component rendered inside a route component. I know this one isn't doing a whole lot now, but eventually it will likely be responsible for things like possibly hooking in data from Redux, making an API call on mount, handling access control, etc.

src/components/Customer/HHGDetailsForm.jsx
suzubara: I think this rule is primarily to enforce a consistent style, I don't have a super strong preference for one or the other.

suzubara:
> > I can't seem to get the same lint error when I rename `HHGDetailsFormComponent` to `HHGDetailsForm`. There should be nothing wrong with importing the same name as the default
> 
> I checked, and it wasn't a lint error, but the UI was erroring (failed to compile)
> image
> 
> Moot point if we combine route page with the form component, but baffling. . .

I am also not seeing this lint error, but I can provide some background here since this is a subtle but pretty important lint rule. When you import the default export from a module (i.e., `import MyForm from './MyForm'`), there is nothing to enforce what you call it, as you're essentially assigning it a name when you import it. So you could just as soon do `import NonsenseWhatever from './MyForm'` and it would be the same thing.

However for **named** exports, you _have_ to import by the name it is exported by (i.e., `import { MyForm } from './MyForm'`). Default exports don't have a name (hence why you have to give them one when you import them). So this rule is to guard against a scenario where if you have a module with both a default export and named exports...
```
export class MyForm { ...

export default MyForm
```

...it will not allow you to name the _default_ export the same as one of the _named_ exports...
```
// this would cause a lint error, because there's a potential naming conflict between the default export & the MyName export
import MyForm from './MyForm
```

It ensures that you could theoretically always import all exports from a given file, without naming conflicts. I also think it provides value in keeping it clear _what_ you're importing (since if you did import a default and give it the same name as something else, it might be confusing which one you mean).

A common pattern is to export a component with HOCs (such as connecting to Redux) as the default export, and also export the component on its own as a named export (such as for testing purposes if you want to test the component in isolation):

```
export class MyComponent { ... }

export default connect(MyComponent)
```

In this case I find it helpful to be descriptive when importing the default, and call it something like `ConnectedMyComponent` so that it is clear I am importing the component + its connectors.

src/components/Customer/HHGDetailsForm.jsx
Ronolibert: Ah ok! don't have any strong feelings one way or the other. We can default to the linter

src/pages/MyMove/HHGMoveSetup.jsx
SirenaBorracha: @Ronolibert we're going to stick with what's in there based on Suz's comment :)

src/components/form/AddressFields/AddressFields.test.jsx
SirenaBorracha: Updated to search for `name="remarks"`

src/components/Customer/HHGDetailsForm.stories.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { array, text } from '@storybook/addon-knobs';
```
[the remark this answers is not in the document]
SirenaBorracha: Removed

kimallen:
> > > I can't seem to get the same lint error when I rename `HHGDetailsFormComponent` to `HHGDetailsForm`. There should be nothing wrong with importing the same name as the default
> > 
> > 
> > I checked, and it wasn't a lint error, but the UI was erroring (failed to compile)
> > image
> > Moot point if we combine route page with the form component, but baffling. . .
> 
> I am also not seeing this lint error, but I can provide some background here since this is a subtle but pretty important lint rule. When you import the default export from a module (i.e., `import MyForm from './MyForm'`), there is nothing to enforce what you call it, as you're essentially assigning it a name when you import it. So you could just as soon do `import NonsenseWhatever from './MyForm'` and it would be the same thing.
> 
> However for **named** exports, you _have_ to import by the name it is exported by (i.e., `import { MyForm } from './MyForm'`). Default exports don't have a name (hence why you have to give them one when you import them). So this rule is to guard against a scenario where if you have a module with both a default export and named exports...
> 
> ```
> export class MyForm { ...
> 
> export default MyForm
> ```
> 
> ...it will not allow you to name the _default_ export the same as one of the _named_ exports...
> 
> ```
> // this would cause a lint error, because there's a potential naming conflict between the default export & the MyName export
> import MyForm from './MyForm
> ```
> 
> It ensures that you could theoretically always import all exports from a given file, without naming conflicts. I also think it provides value in keeping it clear _what_ you're importing (since if you did import a default and give it the same name as something else, it might be confusing which one you mean).
> 
> A common pattern is to export a component with HOCs (such as connecting to Redux) as the default export, and also export the component on its own as a named export (such as for testing purposes if you want to test the component in isolation):
> 
> ```
> export class MyComponent { ... }
> 
> export default connect(MyComponent)
> ```
> 
> In this case I find it helpful to be descriptive when importing the default, and call it something like `ConnectedMyComponent` so that it is clear I am importing the component + its connectors.

This is really helpful and makes sense.  Thanks for the explanation!

SirenaBorracha:
> This is looking really good! making separate components for the address & contact info fieldsets is a great idea :D One thing that's left to fully hook the form inputs up with Formik is to pass in the Formik `values` and `onChange` handler to all inputs. Then Formik will become aware of all of the input values and keep them in sync at the form level. If you do this, then for example you wouldn't actually need the `hasDeliveryAddress` state at all anymore because you can look at `values['has-delivery-address']` instead.
> 
> Formik provides a few different ways to do this, but probably the easiest one to do based on what you have already is to wrap all of its children like so (full docs at https://formik.org/docs/api/formik#children-reactreactnode--props-formikpropsvalues--reactnode):
> 
> ```
> <Formik initialValues={initialValues}>
> {({ handleChange, values, ... }) => (
>   <WizardPage ...>
>   // everything else in the form...
>   </WizardPage>
> )}
> </Formik>
> ```
> 
> _then_ you have access to `values, handleChange` (and any other Formik context props or methods you need) inside the form itself. Each input will need to get `onChange={handleChange}` added, as well as `value={values.<input name>}`. Then the inputs will become "controlled" by Formik, and it'll become easier to add branches and validation later on. let me know if you have any questions and I would be happy to help and/or pair on this!
> 
> A few other minor things:
> 
> * I would make all input `name` values camelCase so it's easier to access them in JS. The `name` attributes should all match the keys in the form data model, so perhaps using the shape you created for `props.initialValues` would make sense here.
> * For input `id` values, it is usually a best practice to use a `uuid` generator. The requirements of an input `id` attribute are they have to match their `label for` attribute, and they have to be unique. Since they're hardcoded right now, there's more than one element with the same idea for example in the repeated Address & ContactInfo fieldsets. Usually I like to create an input ID using the name and a random uuid, so for example `mailingAddress1_a1b2c3`

@suzubara Kim and I believe we've made these requested changes

@Ronolibert please let us know if there's anything else you'd like us to do

src/components/Customer/HHGDetailsForm.jsx
```diff
@@ -0,0 +1,171 @@
...
+                  onChange={handleChange}
+                  value={values.requestedDeliveryDate}
+                />
+                <span className="usa-hint" id="deliveryDateHint">
+                  Your movers will confirm this date or one shortly before or after.
+                </span>
+              </Fieldset>
+              <Fieldset legend="Delivery location" className={fieldsetClasses.margin}>
+                <Label>Do you know your delivery address?</Label>
+                <Radio
```
suzubara, line 76: These `Radio` components need a `value` prop to tell the form what value they correspond to. Probably you want the "Yes" one to have `value={true}` and the "No" one to have `value={false}`

src/components/form/AddressFields/AddressFields.jsx
```diff
@@ -0,0 +1,83 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Fieldset, Label, TextInput } from '@trussworks/react-uswds';
+import { v4 as uuidv4 } from 'uuid';
+
+export const AddressFields = ({ legend, className, values, handleChange }) => {
+  const addressFieldsUUID = uuidv4();
```
suzubara, line 7: 👍

src/components/Customer/HHGDetailsForm.jsx
```diff
@@ -0,0 +1,171 @@
...
+                  label="Yes"
+                  name="hasDeliveryAddress"
+                  onChange={this.handleChangeHasDeliveryAddress}
+                />
+                <Radio
+                  className="display-inline-flex"
+                  id="no-delivery-address"
+                  label="No"
+                  name="hasDeliveryAddress"
+                  defaultChecked
```
suzubara, line 88: to determine which Radio input is checked by default should probably be informed by the `initialValues`, so you can replace this prop with `checked={initialValues.hasDeliveryAddress}`

src/components/Customer/HHGDetailsForm.jsx
```diff
@@ -0,0 +1,171 @@
...
+                </span>
+              </Fieldset>
+              <Fieldset legend="Delivery location" className={fieldsetClasses.margin}>
+                <Label>Do you know your delivery address?</Label>
+                <Radio
+                  className="display-inline"
+                  id="has-delivery-address"
+                  label="Yes"
+                  name="hasDeliveryAddress"
+                  onChange={this.handleChangeHasDeliveryAddress}
```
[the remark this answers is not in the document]
kimallen: These radios aren't actually part of the form.  They are the means to toggle between showing the delivery address inputs and showing the duty station zip.  We do not save the values "yes" and "no" anywhere.

src/components/Customer/HHGDetailsForm.jsx
```diff
@@ -0,0 +1,171 @@
...
+                />
+                <Radio
+                  className="display-inline-flex"
+                  id="no-delivery-address"
+                  label="No"
+                  name="hasDeliveryAddress"
+                  defaultChecked
+                  onChange={this.handleChangeHasDeliveryAddress}
+                />
+                {hasDeliveryAddress ? (
```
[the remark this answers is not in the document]
kimallen: again, `hasDeliveryAddress` is not a value we want to pass through the form (there is no db field for that), it's just a way to toggle viewing the delivery address.

src/components/Customer/HHGDetailsForm.jsx
kimallen: Same  as above about these Radios not being part of the form values.

src/components/Customer/HHGDetailsForm.jsx
```diff
@@ -0,0 +1,171 @@
...
+import PropTypes from 'prop-types';
+import { Formik } from 'formik';
+import { Fieldset, Radio, Label } from '@trussworks/react-uswds';
+
+import { Form } from '../form/Form';
+import { DatePickerInput, TextInput } from '../form/fields';
+import { AddressFields } from '../form/AddressFields/AddressFields';
+import { ContactInfoFields } from '../form/ContactInfoFields/ContactInfoFields';
+import { WizardPage } from '../../shared/WizardPage';
+
```
Shkeating, line 11: I would recommend adding a css module to this to fix a few issues that are coming through. Using the USWDS classes is fine, but sometimes they don't have everything we need to achieve exactly what we want.

there is a lot of extra space after the `legend` at the beginning of each section. It doesn't seem that it has a massive margin bottom assigned to it or anything. I was able to resolve this in the browser by giving `.usa-legend` a `margin-bottom` of `-30px` or so. some text types rely on this space so it may need to be added back in in some spots. We could also leave the `.usa-legend`s alone and remove the margin/padding top from the elements that do have it to decrease the space.

We could also give a little `margin-bottom` to our `fieldset`s but I did notice some helper text is currently not being held within its respective `fieldset`
comments

src/components/Customer/HHGDetailsForm.jsx
```diff
@@ -0,0 +1,171 @@
...
+                <Label>Do you know your delivery address?</Label>
+                <Radio
+                  className="display-inline"
+                  id="has-delivery-address"
+                  label="Yes"
+                  name="hasDeliveryAddress"
+                  onChange={this.handleChangeHasDeliveryAddress}
+                />
+                <Radio
+                  className="display-inline-flex"
```
Shkeating, line 84: the way this uswds class works, you need to apply this to the _container_ of the items you want to display inline, not the items themselves. You also don't want to mix `.display-inline-flex` and `.display-inline` I would personally recommend just using the flexbox based one. You want to wrap whatever you want to display inline with a wrapping `div` with that class applied to get the desired behavior!

I would do it like this:
```jsx
           <div className="display-flex margin-top-1">
                <Radio
                  id="has-delivery-address"
                  label="Yes"
                  name="hasDeliveryAddress"
                  onChange={this.handleChangeHasDeliveryAddress}
                />
                <Radio
                  id="no-delivery-address"
                  label="No"
                  name="hasDeliveryAddress"
                  defaultChecked
                  onChange={this.handleChangeHasDeliveryAddress}
                />
        </div>
```

Shkeating: this should also prevent the extra space you are seeing on the left of the first item, and should give the radio buttons a little space up top from their label
image

src/components/Customer/HHGDetailsForm.jsx
kimallen: Thanks Shauna- this is actually meant to be a functional form.  We are blocking other teams that need to be able to use the form to create an MTO shipment.  We will be styling in phase 2.  I only added a little styling so that it doesn't hurt the eyes SO much.  I'll add a note that the current styles are temporary.

src/components/Customer/HHGDetailsForm.jsx
kimallen: I went ahead and wrapped the Radios.  Not bothering with other spacing until phase 2.

src/components/Customer/HHGDetailsForm.jsx
```diff
@@ -0,0 +1,171 @@
...
+                  id="requestedPickupDate"
+                  onChange={handleChange}
+                  value={values.requestedPickupDate}
+                />
+              </Fieldset>
+              <span className="usa-hint" id="pickupDateHint">
+                Your movers will confirm this date or one shortly before or after.
+              </span>
+              <AddressFields
+                initialValues={initialValues.pickupLocation}
```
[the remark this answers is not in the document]
kimallen: Added name props and removed initialValues

src/components/Customer/HHGDetailsForm.jsx
```diff
@@ -0,0 +1,171 @@
...
+  handleChangeHasDeliveryAddress = () => {
+    this.setState((prevState) => {
+      return { hasDeliveryAddress: !prevState.hasDeliveryAddress };
+    });
+  };
+
+  render() {
+    const { initialValues, pageKey, pageList } = this.props;
+    const { hasDeliveryAddress } = this.state;
+    const fieldsetClasses = { margin: 'margin-top-2' };
```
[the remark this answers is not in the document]
kimallen: removed object

src/components/Customer/HHGDetailsForm.jsx
Shkeating: ah gotcha i missed that!

src/components/Customer/HHGDetailsForm.jsx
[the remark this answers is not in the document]
suzubara: So, the form doesn't have to 100% represent what will be stored in the database. I often find it's most helpful for the form data model to match the UI, and then modify the inputs that are sent in the API request at the point of submission. However storing this value in state is also a valid approach so this is not a blocking change. you just may find it easier to make all the inputs consistent in which onChange handlers they are using.

src/components/Customer/HHGDetailsForm.jsx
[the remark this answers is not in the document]
suzubara: (see comment above)

src/components/Customer/HHGDetailsForm.jsx
suzubara: Even if these inputs aren't being saved in the form, they still need a `value` prop set to distinguish the two options semantically. The only reason setting `hasDeliveryAddress` in state is working now is because it's defaulted to false (the onChange handler is just toggling the current value, instead of being attached to the actual input the user clicks).

kimallen: Ok, I'll put a a note somewhere to modify this- just want to get this in asap so that we can unblock the other stories.  Thanks for all the review!

PR #4432, opened 2020-07-21
MB-2892 [Prime API] CreateMTOShipment - Adds mtoServiceItems to response payload

chrisrcoles:
## Description

This PR adds some enhancements to the Prime API endpoint, `CreateMTOHandler`. `CreateMTOShipmentHandler` is responsible for creating the initial MTO shipment. Prior to this PR, the endpoint was not properly handling `mtoServiceItems`, a request parameter. This PR makes the following additions: 
1. Includes the `mtoServiceItems` that are processed and stored as apart of the response payload. 
2. Ensures a test case is added that captures this updated payload. 

## Reviewer Notes

In-line comments 

## Setup

1. Start and run `mymove` Prime API
```sh
mylaptop ~ %: cd /mymove
mylaptop ~ %: make server_run
```

2. Load end to end data for pre-existing MTO Shipment
```sh
mylaptop mymove  %: make db_dev_e2e_populate
```

3. Get example payload and save locally. 
```
mylaptop mymove  %: curl https://gist.githubusercontent.com/chrisrcoles/663f30c2805eb03ba163711096edab5e/raw/a5037e8c9ff49508dc272131acabc10b82662fdb/create_mto_shipment.json >> create_mto_shipment.json
```

4. Test endpoint
```sh
mylaptop mymove %: make clean && make build #ensures new prime yaml has been re-compiled successfully
mylaptop mymove %: prime-api-client --insecure create-mto-shipment --filename create_mto_shipment.json | jq
```

5. After sending request to endpoint, ensure `mtoServiceItems` is included in response

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2892) for this change.


swagger/prime.yaml
```diff
@@ -859,6 +859,10 @@ definitions:
         readOnly: true
       agents:
         $ref: '#/definitions/MTOAgents'
+      mtoServiceItems:
+        items:
+          $ref: '#/definitions/MTOServiceItem'
+        type: array
```
chrisrcoles, line 865: Ideally, this would look something like: 
```suggestion
        mtoServiceItems:
            $ref: '#/definitions/MTOServiceItem'
```
After trying this several times, it looks like `go-swagger` had problems handling the definition this way. Very strange, especially considering `MTOAgents` had [no issue being defined this way](https://github.com/transcom/mymove/blob/17d2b4ae8f49ce1bd36f9e6432b8f6902d62762a/swagger/prime.yaml#L860). I think there is a bug in swagger's validation library that is mishandling definitions with multiple types, like [MTOServiceItemModelType](https://github.com/transcom/mymove/blob/17d2b4ae8f49ce1bd36f9e6432b8f6902d62762a/swagger/prime.yaml#L963)

pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -67,6 +67,7 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
 	// We can have two service items that come in from a MTO approval that do not have an MTOShipmentID
 	// they are MTO level service items. This should capture that and create them accordingly, they are thankfully
 	// also rather basic.
+	// TODO is it okay for this to happen on createMTOShipment given that mtoShipmentID will be null
 	if serviceItem.MTOShipmentID == nil {
```
chrisrcoles, line 71: Is this okay for `serviceItem.MTOShipmentID` to be `nil` especially considering this is the `createMTOShipment` endpoint? I added persisting mto service items as apart of the larger mto shipment transaction but I'm not sure how we should handle service items with no mto shipment id. Some questions I had are: 
1. Is this okay to be `nil`, given this is being used in a `createMTOShipment` endpoint where that ID could be null? 
2. What are the circumstances under which an `MTOServiceItem MTOShipmentID` is `null`?

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -112,6 +116,20 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
 			}
 		}
 
+		// create MTOServiceItems List
+		if serviceItems != nil {
```
chrisrcoles, line 120: Includes service items as apart of the transaction, but what should happen if one fails for any reason?

chrisrcoles: The Jira ticket also asks to check that the `MTOServiceItem.MoveTaskOrderID === MTOShipment.MoveTaskOrderId`. Both of these considerations should be tackled in a new ticket.

pkg/handlers/primeapi/internal/payloads/payload_to_model.go
```diff
@@ -61,8 +61,8 @@ func MTOAgentsModel(mtoAgents *primemessages.MTOAgents) *models.MTOAgents {
 	return &agents
 }
 
-// MTOServiceItemList model
-func MTOServiceItemList(mtoShipment *primemessages.CreateShipmentPayload) (models.MTOServiceItems, *validate.Errors) {
+// MTOServiceItemModelListFromCreate model
```
chrisrcoles, line 64: renamed this function to something that made more sense

chrisrcoles:
Will be taking a look at these tests and resolving them!

LeDeep:
i'm seeing the response, but id in the response is all zeros. The db record is being created with a proper id, though.

Screen Shot 2020-07-23 at 1 53 31 PM

sandy-wright:
> i'm seeing the response, but id in the response is all zeros. The db record is being created with a proper id, though.
> 
> Screen Shot 2020-07-23 at 1 53 31 PM

I thought we didn't let the Prime add service items with type MTOServiceItemBasic? This is the error I get trying to add it with the `createMTOServiceItem` endpoint:

```MTOServiceItem modelType() not allowed: MTOServiceItemBasic ```

We would want this endpoint to also mimic that behavior right? I don't really know what the rules are with the Prime and service items yet.

File: pkg/services/mto_shipment/mto_shipment_creator.go (hunk 3 above)
Comment (sandy-wright) [reply pkg/services/mto_shipment/mto_shipment_creator.go:121]: My preference would be for the whole transaction to fail if any one service items fails. The Prime will have a limited feedback loop with us once this goes live, and partial successes are really likely to be missed, which could affect the system down the line. Open to other opinions about this though (we need to address it for MTO Agents as well actually).

And I think you're right that these can be tackled in different tickets!

File: pkg/services/mto_shipment/mto_shipment_creator.go [hunk 5]
```diff
@@ -39,6 +42,7 @@ func NewMTOShipmentCreator(db *pop.Connection, builder createMTOShipmentQueryBui
 		builder,
 		fetch.NewFetcher(builder),
 		createNewBuilder,
+		mtoserviceitem.NewMTOServiceItemCreator(builder),
```
LeDeep, line 45: prime can only create accessorial service items, so will need to add a check for that. If you check the MTOServiceItem handler you can see the list of allowed service items.

LeDeep: oops, was reviewing and didn't see @sandy-wright comment above. Sorry for the double up.

pkg/services/mto_shipment/mto_shipment_creator.go
sandy-wright: Actually, for the `moveTaskOrderID` you could just use whatever was passed in with the shipment and ignore whatever is on the service item.

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -113,6 +117,20 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
 			}
 		}
 
+		// create MTOServiceItems List
+		if serviceItems != nil {
+			for _, serviceItem := range serviceItems {
```
sandy-wright, line 122: You should set the `moveTaskOrderID` value and the `mtoShipmentID` value here (kinda like how it works for `agents`).

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -113,6 +117,20 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
...
+				_, validationErrs, serviceError := f.mtoServiceItemCreator.CreateMTOServiceItem(&serviceItem)
+				if validationErrs != nil && validationErrs.HasAny() {
+					return verrs
+				}
+				if serviceError != nil {
+					return serviceError
+				}
+				shipment.MTOServiceItems = serviceItems
```
sandy-wright, line 130: This should be set to what comes out of the create function, so the value returned first on line 123 I think, so that the client gets the new ID and the eTag back.

pkg/services/mto_shipment/mto_shipment_creator.go
chrisrcoles: Hey I resolved this - previously I was just setting `shipment.MTOServiceItems` to `serviceItems`, the model before it was persisted; thus why there was no `ID`. I've updated to make sure we set `shipment.MTOServiceItems` to the correct values - [the updated model with all the persisted values](https://github.com/transcom/mymove/blob/cc-MB-2892/primeapi-createMTOShipment-add-service-items/pkg/services/mto_shipment/mto_shipment_creator.go#L132).

pkg/services/mto_shipment/mto_shipment_creator.go
chrisrcoles: Hey I solved this the same way as I did [here](https://github.com/transcom/mymove/pull/4432#discussion_r459808465). Let me know if this works for you!

chrisrcoles:
Thanks a bunch @LeDeep and @sandy-wright for the thorough reviews. I've updated the PR with 
1. Ensuring that the `MTOServiceItem.ID` is being included in the response
2. Checking that the `MTOServiceItem.ModelType` is a valid type.

There are some unresolved issues around validating IDs. @sandy-wright and I will be speaking tomorrow to resolve this. This work will either be added to this PR or new tickets will be created for that work.

LeDeep:
> Thanks a bunch @LeDeep and @sandy-wright for the thorough reviews. I've updated the PR with
> 
> 1. Ensuring that the `MTOServiceItem.ID` is being included in the response
> 2. Checking that the `MTOServiceItem.ModelType` is a valid type.
> 
> There are some unresolved issues around validating IDs. @sandy-wright and I will be speaking tomorrow to resolve this. This work will either be added to this PR or new tickets will be created for that work

thanks for making the changes. they look good. I agree that separate tickets to not require shipment id and mto id, and just pass the mto id of the shipment to the service item, is a good idea.

pkg/services/mto_shipment/mto_shipment_creator.go
sandy-wright: So this is a different issue - that one was for display with the return payload (which is working well now, thanks!). This one is setting the IDs for _actually_ creating the service item. The problem now is that we're not creating the service items for the shipment that's being created, which is what we want with this endpoint.

LeDeep: i think what we'll ultimately want is to create the service items without needing to include the `moveTaskOrderID` and `mtoShipmentID` as part of the payload. 
So we might need a different service object  (i.e. `CreateMTOServiceItemForMT0Shipment`), since the `CreateMTOServiceItem` is used in multiple places, and you need those fields to create a service item via the `CreateMTOServiceItemHandler`.

i don't know if that should be a different story, or if it makes most sense to wrap it into this story since we want to make sure service items are tied to the correct shipment.

sandy-wright: Well the easiest way to handle is the way MTO Agents does it: 

```go
if shipment.MTOAgents != nil {
	for _, agent := range shipment.MTOAgents {
		agent.MTOShipmentID = shipment.ID
		verrs, err = txBuilder.CreateOne(&agent)
		if err != nil {
			return err
		}
	}
}
```

With the key part being to set `agent.MTOShipmentID = shipment.ID` before trying to create the agent. We could fairly easily do this for service items as well in this PR.

sandy-wright: All you need is:

```go
serviceItem.MTOShipmentID = shipment.ID
serviceItem.moveTaskOrderID = shipment.moveTaskOrderID
```

right here in the code (right after line 124)

LeDeep: i thought that would work too,  but i was getting a 404. maybe my payload was bad.

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -60,6 +64,40 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
...
+				if shipment.PrimeEstimatedWeight == nil {
+					return nil, services.NewConflictError(
+						serviceItem.ReService.ID,
+						"for creating a service item. MTOShipment associated with this service item must have a valid PrimeEstimatedWeight.",
+					)
+				}
+			}
+
+			println("🧸")
+			fmt.Printf("service item %v", serviceItem)
```
sandy-wright, line 95: Oops, I left these prints in, so they should be removed 😅 . Other than though, it seems to still be working though! Yay!

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -60,6 +64,38 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
 		return nil, services.NewNotFoundError(moveTaskOrderID, "for moveTaskOrder")
 	}
 
+	if serviceItems != nil {
```
LeDeep, line 67: it would be helpful, i think, to add a comment here (and below) on why this is here. Could be a little confusing to people seeing this for the first time, and a short comment could help explain the need for this.

PR #4437, opened 2020-07-21
[MB-3135] Create MTO Shipment Endpoint

Ronolibert:
## Description

Create MTO Shipment endpoint for the customer

## Reviewer Notes
- Contains only barebones information needed to create an MTO shipment corresponding with the fields on the HHG Setup screen
- Will be used for this [page](https://dp3.atlassian.net/browse/MB-3158) 
- opted to put endpoint in internal API instead of trying to use the GHC API to have a separation of concerns. The internal API version will be a lot more bare bones since it doesn't need as many fields as the TOO side provides
## Setup
`make server_run`
`make server_test`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3135) for this change


swagger/internal.yaml
```diff
@@ -2522,6 +2522,475 @@ definitions:
...
+    items:
+      $ref: '#/definitions/MTOAgent'
+    type: array
+  MTOShipmentType:
+    type: string
+    title: Shipment Type
+    example: HHG
+    enum:
+      - HHG
+      - INTERNATIONAL_HHG
```
kimallen, line 2581: Instead of International HHG and UB, we should use `HHG_LONGHAUL_DOMESTIC`, `HHG_SHORTHAUL_DOMESTIC`, `HHG_INTO_NTS_DOMESTIC` and `HHG_OUTOF_NTS_DOMESTIC`.  We will not be dealing with international until after Feb 1, but the others we will be developing for.

I've been trying to get answers regarding Motorhome,  and boat types, and the idea of longhaul and shorthaul being sub-categories of HHG, but I think until then we can use the shipment types I mentioned above.

swagger/internal.yaml
```diff
@@ -2522,6 +2522,475 @@ definitions:
...
+      HHG: HHG
+      INTERNATIONAL_HHG: International HHG
+      INTERNATIONAL_UB: International UB
+  MTOShipment:
+    properties:
+      moveTaskOrderID:
+        example: 1f2270c7-7166-40ae-981e-b200ebdf3054
+        format: uuid
+        type: string
+        readOnly: true
```
kimallen, line 2593: Wondering why this is readOnly?  Would this not need to be written when the MTO Shipment is created?

swagger/internal.yaml
```diff
@@ -2522,6 +2522,475 @@ definitions:
...
+        format: date-time
+        type: string
+        readOnly: true
+      updatedAt:
+        format: date-time
+        type: string
+        readOnly: true
+      scheduledPickupDate:
+        format: date
+        type: string
```
kimallen, line 2609: I may not be understanding when we should use readOnly, and when we don't.  I notice `scheduledPickupDate` and `actualPickupDate` are readOnly, while `requestedPickupDate` and `requiredDeliveryDate` are readyOnly.  Is that intentional?

swagger/internal.yaml
```diff
@@ -2522,6 +2522,475 @@ definitions:
...
+        $ref: '#/definitions/Address'
+      secondaryPickupAddress:
+        $ref: '#/definitions/Address'
+      secondaryDeliveryAddress:
+        $ref: '#/definitions/Address'
+  MTOShipments:
+    items:
+      $ref: '#/definitions/MTOShipment'
+    type: array
+  CreateShipmentPayload:
```
kimallen, line 2642: I think we are trying to get away from usuing the word `Payload` in the payload.  I'd just name this `CreateShipment`.

swagger/internal.yaml
```diff
@@ -4602,3 +5071,73 @@ paths:
...
+    post:
+      summary: createMTOShipment
+      description: |
+        Creates a MTO shipment for the specified Move Task Order.
+        Required fields include:
+        * Shipment Type
+        * Customer requested pick-up date
+        * Pick-up Address
+        * Delivery Address
+        * Releasing / Receiving agents
```
kimallen, line 5084: Remove this line since we have it listed as optional below?

swagger/internal.yaml
```diff
@@ -4602,3 +5071,73 @@ paths:
...
+
+        Optional fields include:
+        * Customer Remarks
+        * Releasing / Receiving agents
+        * An array of optional accessorial service item codes
```
kimallen, line 5089: Does the Customer need the service item codes? I _think_ just the prime and office need that info

swagger/internal.yaml
```diff
@@ -4602,3 +5071,73 @@ paths:
...
+          schema:
+            $ref: '#/definitions/CreateShipmentPayload'
+      responses:
+        '200':
+          description: Successfully created a MTO shipment.
+          schema:
+            $ref: '#/definitions/MTOShipment'
+        '400':
+          $ref: '#/responses/InvalidRequest'
+        '404':
```
kimallen, line 5109: Shall we also include a 401 and 403 error response?

swagger/internal.yaml
```diff
@@ -2522,6 +2522,475 @@ definitions:
...
+      pickupAddress:
+        $ref: '#/definitions/Address'
+      destinationAddress:
+        $ref: '#/definitions/Address'
+      agents:
+        $ref: '#/definitions/MTOAgents'
+      pointOfContact:
+        type: string
+        description: Email or id of a contact person for this update
+      mtoServiceItems:
```
kimallen, line 2670: I may not understand the whole flow and what needs to be created when.  Do we need serviceItems on the the create mto_shipment payload?  Would those be added later by the prime?

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -91,6 +91,7 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
 		if shipment.RequestedPickupDate == nil {
 			return services.NewInvalidInputError(uuid.Nil, nil, nil, "RequestedPickupDate is required to create MTO shipment")
 		}
+
 		//assign status to shipment submitted
 		shipment.Status = models.MTOShipmentStatusSubmitted
```
kimallen, line 96: I think it's okay to have the status as submitted for now since we are submitted at the same time as creating.  Later this will be we will have to have a separate submit handler that changes status to SUBMITTED, since the form will be in two Wizard pages, so the second will be an update.

I'm seeing that `SUBMITTED` is the earliest option.  I'm wondering if we will need a DRAFT status in case the Customer leaves the flow midway (when we have it split between wizard pages).

No change requested- just noting additional stories we will need.

pkg/models/mto_shipments.go
```diff
@@ -56,6 +56,7 @@ type MTOShipment struct {
 	MoveTaskOrderID                  uuid.UUID         `db:"move_task_order_id"`
 	ScheduledPickupDate              *time.Time        `db:"scheduled_pickup_date"`
 	RequestedPickupDate              *time.Time        `db:"requested_pickup_date"`
+	RequestedDeliveryDate            *time.Time        `db:"requested_delivery_date"`
```
kimallen, line 59: Good catch that this was missing!

pkg/handlers/internalapi/internal/payloads/model_to_payload.go
```diff
@@ -0,0 +1,96 @@
...
+// MTOShipment payload
+func MTOShipment(mtoShipment *models.MTOShipment) *internalmessages.MTOShipment {
+	payload := &internalmessages.MTOShipment{
+		ID:                       strfmt.UUID(mtoShipment.ID.String()),
+		MoveTaskOrderID:          strfmt.UUID(mtoShipment.MoveTaskOrderID.String()),
+		ShipmentType:             internalmessages.MTOShipmentType(mtoShipment.ShipmentType),
+		CustomerRemarks:          mtoShipment.CustomerRemarks,
+		PickupAddress:            Address(mtoShipment.PickupAddress),
+		DestinationAddress:       Address(mtoShipment.DestinationAddress),
+		SecondaryPickupAddress:   Address(mtoShipment.SecondaryPickupAddress),
```
kimallen, line 39: I don't see secondary addresses in the Customer flow. Are these added later by prime?  In which case, do we need them?  Or would we need them for getting the shipment in the review page?  (I'm particularly wondering because I remember LCLR said to nix a secondary address, but that may have been for the Customer profile)

pkg/handlers/internalapi/internal/payloads/payload_to_model.go
```diff
@@ -0,0 +1,146 @@
...
+
+	requestedPickupDate := time.Time(mtoShipment.RequestedPickupDate)
+	model := &models.MTOShipment{
+		MoveTaskOrderID:     uuid.FromStringOrNil(mtoShipment.MoveTaskOrderID.String()),
+		ShipmentType:        models.MTOShipmentType(mtoShipment.ShipmentType),
+		RequestedPickupDate: &requestedPickupDate,
+		CustomerRemarks:     mtoShipment.CustomerRemarks,
+	}
+
+	if mtoShipment.PickupAddress != nil {
```
kimallen, line 75: I believe the pickup address is required upon creation, so would it ever be nil?

pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -0,0 +1,58 @@
...
+// CreateMTOShipmentHandler is the handler to create MTO shipments
+type CreateMTOShipmentHandler struct {
+	handlers.HandlerContext
+	mtoShipmentCreator services.MTOShipmentCreator
+}
+
+// Handle creates the mto shipment
+func (h CreateMTOShipmentHandler) Handle(params mtoshipmentops.CreateMTOShipmentParams) middleware.Responder {
+
+	logger := h.LoggerFromRequest(params.HTTPRequest)
```
kimallen, line 24: Do we need to check session in order to check authorization here?  I don't see it happening in prime handlers, but do see it in some of the older handlers, so is authorization happening somewhere else?

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -0,0 +1,162 @@
...
+			PickupAddress: &internalmessages.Address{
+				City:           &pickupAddress.City,
+				Country:        pickupAddress.Country,
+				PostalCode:     &pickupAddress.PostalCode,
+				State:          &pickupAddress.State,
+				StreetAddress1: &pickupAddress.StreetAddress1,
+				StreetAddress2: pickupAddress.StreetAddress2,
+				StreetAddress3: pickupAddress.StreetAddress3,
+			},
+			PointOfContact:        "John Doe",
```
kimallen, line 65: I don't think we use a PointOfContact from the Customer, do we?  I think that's for the prime since they don't have individual users logged in.

swagger/internal.yaml
```diff
@@ -2522,6 +2522,475 @@ definitions:
...
+        type: string
+        example: handle with care
+        x-nullable: true
+      pickupAddress:
+        $ref: '#/definitions/Address'
+      destinationAddress:
+        $ref: '#/definitions/Address'
+      agents:
+        $ref: '#/definitions/MTOAgents'
+      pointOfContact:
```
kimallen, line 2667: I don't think we use `pointOfContact` for creation of an MTO_Shipment by the customer (just for prime to identify a user, since they don't have other individual identifiers?)

swagger/internal.yaml
Ronolibert: Yup! Added this

pkg/handlers/internalapi/mto_shipment_test.go
Ronolibert: We don't need it! It's an artifact of using the prime API as a template, removed this

pkg/handlers/internalapi/mto_shipment.go
Ronolibert: Yup! Added the auth check

pkg/handlers/internalapi/internal/payloads/payload_to_model.go
Ronolibert: You're right, this and the destination address would never be nil. I'm not sure why this check happens in the prime API but I removed it here and also since the field is required in the swagger file this check is unnecessary as well

pkg/handlers/internalapi/internal/payloads/model_to_payload.go
Ronolibert: Oh good call! I don't see any secondary fields on the UI either, I removed this

swagger/internal.yaml
Ronolibert: Removed this

pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -20,8 +20,11 @@ type CreateMTOShipmentHandler struct {
 
 // Handle creates the mto shipment
 func (h CreateMTOShipmentHandler) Handle(params mtoshipmentops.CreateMTOShipmentParams) middleware.Responder {
+	session, logger := h.SessionAndLoggerFromRequest(params.HTTPRequest)
 
-	logger := h.LoggerFromRequest(params.HTTPRequest)
+	if session == nil {
```
kimallen, line 25: Do we also need to check that it's a Milmove user?  I found an example in `models/order.go` where this is done:
```
if session.IsMilApp() && order.ServiceMember.ID != session.ServiceMemberID {
		return Order{}, ErrFetchForbidden
	}
	return order, nil
```
I see other places where the session is passed into the service and auth is checked there, but it seems to make sense to handle that first thing in the handler.  I don't see a lot of consistency in the code where we are doing auth.

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+
+		otherParams := mtoshipmentops.CreateMTOShipmentParams{
+			HTTPRequest: req,
+		}
+		response := handler.Handle(otherParams)
+
+		suite.IsType(&mtoshipmentops.CreateMTOShipmentBadRequest{}, response)
+	})
+
```
kimallen, line 135: Let's also add tests for 401 and 403 responses

swagger/internal.yaml
```diff
@@ -2522,6 +2522,470 @@ definitions:
...
+      - IOOUB
+      - IOPSIT
+      - IOSHUT
+      - IUBPK
+      - IUBUPK
+      - IUCRT
+      - MS
+      - NSTH
+      - NSTUB
+  MTOServiceItemBasic:
```
kimallen, line 2754: Do we need any of the `MTOServiceItem*` definitions now that you remove service item from the create payload?  I honestly don't know if the Customer will need this info or not.  If not, we could remove, but if we'll need it later, we could leave it.

pkg/handlers/internalapi/mto_shipment_test.go
Ronolibert: Added tests for them! Not really too clear what the difference is between 401 and 403 so I just added a test case for not being authorized

swagger/internal.yaml
Ronolibert: Removed the unnecessary fields and cleaned it up

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -0,0 +1,203 @@
...
+		badParams := params
+		badParams.Body.PickupAddress = nil
+
+		response := handler.Handle(badParams)
+
+		suite.IsType(&mtoshipmentops.CreateMTOShipmentUnprocessableEntity{}, response)
+	})
+
+	suite.T().Run("POST failure - 401 & 403 - permission denied", func(t *testing.T) {
```
kimallen, line 100: Sorry to nit- I think we need to separate these two out.  This test checks authentication, and should get a 401 since there is no user.  If someone is logged in, but not the right type of user (ex. prime user or office user), we want to get a 403).

PR #4439, opened 2020-07-21
MB-3401: Moving storage backend config out of mymove repo for experimental

cblkwell:
## Description

This change removes the STORAGE_BACKEND parameter from the configurations in the application environment configuration files in favor of placing this in an SSM parameter configured with Terraform. This PR is contingent on https://github.com/transcom/transcom-infrasec-com/pull/1122 being merged first; once that is merged in, *deploys to staging or prod will fail* until this PR is also merged in.

This has already been tested in experimental and worked successfully.

## References

* [MB-3401](https://dp3.atlassian.net/browse/MB-3401)


cblkwell:
> LGTM, but I'll let infra approve...
> What are we using if we're removing this env variable w/o a replacement?

I thought I explained this in the description, but we're replacing this with a SSM parameter set in Terraform (the same process we use for secrets). The plus side of this is that that change can be toggled at the infra level *without* requiring redeploy of the application -- we can just restart the app and it will pick up the new setting. If you look at the infrasec PR I linked up there, you can see it in the Terraform plans here:

```text
  # module.app.aws_ssm_parameter.storage_backend will be created
  + resource "aws_ssm_parameter" "storage_backend" {
      + arn         = (known after apply)
      + description = "Storage backend for application"
      + id          = (known after apply)
      + key_id      = (known after apply)
      + name        = "/app-staging/storage_backend"
      + tier        = "Standard"
      + type        = "String"
      + value       = (sensitive value)
      + version     = (known after apply)
    }
```

So that environment variable will still be set, it will just be set somewhere else.

lynzt:
you did mention it in the description, but as @kilbergr can attest, i'm bad at anything TF related and need things very verbose. thanks for the additional context 👍

PR #4447, opened 2020-07-22
MB-2618 TIO move level service items one card

donaldthai:
## Description

> Given that I am a TIO
 When I view the Move Management and Counseling service items
 Then they appear together on a single card
  And they can still be approved / denied separately

## Reviewer Notes

- In the code, you will see that multiple cards are displayed for the basic service items, ReviewServiceItems
- While clicking the next button, the card will scroll into view. This shows better when viewing the multiple basic service item cards

## Setup

```sh
make server_run
make office_client_run
```

To test:

1. Log in as TIO, office app
2. Navigate over to `http://officelocal:3000/moves/1/payment-requests` queue page
3. Click on any payment request
4. Click on the `next` button on the bottom right to navigate between service items
5. Each click should scroll to each card displayed and this should be the case for basic service items as well

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2618) for this change
* [Uses scrollIntoView()](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView)

## Screenshots

Jul-23-2020 11-49-09


src/components/Office/ReviewServiceItems/ReviewServiceItems.stories.jsx
```diff
@@ -70,12 +98,6 @@ export const MultipleShipmentsGroups = () => (
         amount: 0.01,
         createdAt: '2020-01-01T00:09:00.999Z',
       },
-      {
-        id: '2',
-        serviceItemName: 'Move management',
-        amount: 1234.0,
-        createdAt: '2020-01-01T00:06:00.999Z',
-      },
```
donaldthai, line 78: Deleting so we're left with just one basic service item for this one.

donaldthai:
> The code changes here look good! The only thing I wonder is whether the scroll effect is the right thing to do from a UX standpoint, because if the user's window is tall enough to display both items, nothing will effectively happen when they click prev/next and it could feel broken. But I'll leave that to @transcom/truss-design . Nice use of `useEffect` :D

Yea, I wonder if we need additional css changes like highlighting/focusing the current card for a better UX.

donaldthai:
> there is some weird stuff happening with keyboard focus where only one of the radio buttons from each fieldset is getting focus when keyboard navigating the page.
> 
> it seems to work fine on the [component page in storybook](https://storybook.move.mil/iframe.html?id=too-tio-components-reviewserviceitems--basic) so it might be a branch-specific issue
> 
> keyboardfocus

@Shkeating hmm so according to [W3 docs](https://www.w3.org/TR/wai-aria-practices/#radiobutton), tabbing `TAB` and `TAB + Shift` over a `radio button group` with one button focused is expected. Looks like we can still focus on another button in the group by using the `up` and `down` keys

Shkeating:
> > there is some weird stuff happening with keyboard focus where only one of the radio buttons from each fieldset is getting focus when keyboard navigating the page.
> > it seems to work fine on the [component page in storybook](https://storybook.move.mil/iframe.html?id=too-tio-components-reviewserviceitems--basic) so it might be a branch-specific issue
> > keyboardfocus
> 
> @Shkeating hmm so according to [W3 docs](https://www.w3.org/TR/wai-aria-practices/#radiobutton), tabbing `TAB` and `TAB + Shift` over a `radio button group` with one button focused is expected. Looks like we can still focus on another button in the group by using the `up` and `down` keys

if that's the case, I'll go ahead and approve!

PR #4459, opened 2020-07-23
MB-3097 Ignored IDs in Create MTO Shipment

sandy-wright:
## Description

This ticket was to mark `id` and `mtoShipmentID` as `readOnly` in the payload definitions for the `createMTOShipment` endpoint in the Prime API and to add tests to catch these fields when filled. These fields were already `readOnly`, so I did some extra clean-up as I was poking around in this file. Namely, I made `reServiceCode` required on all MTO Service Item model types and removed the `Payload` suffix from the `CreateShipmentPayload` and `CreatePaymentRequestPayload` definitions. 

## Setup

Run the server tests and try out some endpoints! In particular, the `createMTOShipment` needs to be tested with a variety of data, such as:

* values in any Agent's `id` field
* values in any Agent's `mtoShipmentID` field (maybe even try this in a _different_ agent)
* values in any `id` fields for the addresses
* values in `id` field for the MTO shipment itself
* a correct payload that actually gets updated

Also, the `createMTOServiceItem` and `createPaymentRequest` endpoints should be tested to make sure their functionality has not been altered after the YAML edits.

[Prime API Testing Documentation](https://github.com/transcom/prime_api_deliverable/wiki/API-Endpoints)

## References

* [MB-3097](https://dp3.atlassian.net/browse/MB-3097) for marking IDs as `readOnly` and adding tests
* [MB-3305](https://dp3.atlassian.net/browse/MB-3305) for removing `Payload` suffix


pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -123,3 +128,52 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
 
 	return shipment, transactionError
 }
+
+// checkShipmentIDFields checks that the client hasn't attempted to set ID fields that will be generated/auto-set
+func checkShipmentIDFields(shipment *models.MTOShipment) error {
+	verrs := validate.NewErrors()
+
+	if shipment.ID != uuid.Nil {
```
LeDeep, line 136: i'm able to pass a shipment id and still create the shipment. It ignored the id that's passed in and creates a unique shipment, but i think because the shipment hasn't been created yet, this line is always true. When I print the shipment.ID it is always uuid.Nil

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -123,3 +128,52 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
...
+			if agent.ID != uuid.Nil && !mtoAgentIDErr {
+				verrs.Add("agents:id", "cannot be set for new agents")
+				mtoAgentIDErr = true
+			}
+			if agent.MTOShipmentID != uuid.Nil && !mtoShipmentIDErr {
+				verrs.Add("agents:mtoShipmentID", "cannot be set for agents created with a shipment")
+				mtoShipmentIDErr = true
+			}
+
+			if mtoAgentIDErr && mtoShipmentIDErr {
```
LeDeep, line 154: is this check necessary? seems like we can get the check we want without them.

sandy-wright: This is so we don't get multiple of the same validation error multiple times with no new info, like if there were 3-4 service items with the same issue. It's not strictly necessary, just helps the output be a little tidier (and could potentially be more efficient if we might have loads of service items, although I doubt we ever will)

pkg/services/mto_shipment/mto_shipment_creator.go
sandy-wright: I noticed this too! I found out it's actually because the `CreateMTOShipment` definition doesn't even include the `id` field. So swagger lets us pass it in and doesn't check that it's not supposed to be in the payload, but then it doesn't even get to our handler because the go object for the payload doesn't include an `ID` field in the struct. But I wasn't sure if I should ignore this check because it does cover all our bases to include it? The only way to test it is with a go test though, so it's weird. I can remove it if it's confusing

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -123,3 +128,52 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
...
+				break // we've found all the errors we're gonna find here
+			}
+		}
+	}
+
+	addressMsg := "cannot be set for new addresses"
+	if shipment.PickupAddress != nil && shipment.PickupAddress.ID != uuid.Nil {
```
chrisrcoles, line 161: If I remove `lines 161-166` and remove `pickupAddress` and `destinationAddress`, I will still get the same error. Doesn't swagger validate these for us?

pkg/services/mto_shipment/mto_shipment_creator.go
LeDeep: mtoAgents is max 2 per shipment, right? In a situation where we can create more objects, something like this would be super useful. Since it's only 2, i think the code would be cleaner without this check. Not a blocker though, so i'll leave it up to you.

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -123,3 +128,52 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
...
+// checkShipmentIDFields checks that the client hasn't attempted to set ID fields that will be generated/auto-set
+func checkShipmentIDFields(shipment *models.MTOShipment) error {
+	verrs := validate.NewErrors()
+
+	if shipment.ID != uuid.Nil {
+		verrs.Add("id", "cannot be set for new shipments")
+	}
+
+	if len(shipment.MTOAgents) > 0 {
+		var mtoAgentIDErr = false
```
chrisrcoles, line 201: the payload response seems a bit off
```
 "agents": [
        {
            "agentType": "RELEASING_AGENT",
            "createdAt": "0001-01-01T00:00:00.000Z",
            "email": "jsemaj@example.com",
            "firstName": "James",
            "id": "00000000-0000-0000-0000-000000000000",
            "lastName": "Semaj",
            "mtoShipmentID": "00000000-0000-0000-0000-000000000000",
            "phone": "234-555-1234",
            "updatedAt": "0001-01-01T00:00:00.000Z"
        }
    ],
``` 
Should the `id` and `mtoShipmentID` values be `nil` like this? We might need to double check this.

pkg/services/mto_shipment/mto_shipment_creator.go
LeDeep: yeah, it's always going to be true, so i don't think there is much value in leaving it in. I would remove it.

pkg/services/mto_shipment/mto_shipment_creator.go
sandy-wright: sure thing, I'll clean it up!

pkg/services/mto_shipment/mto_shipment_creator.go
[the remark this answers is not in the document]
LeDeep: actually swagger does check for presence of pickup and destination address, so you only need to do the uuid.Nil check

pkg/services/mto_shipment/mto_shipment_creator.go
sandy-wright: Would you mind trying updating the branch and trying again? This was happening because in the chunk of code to add the agents starting at line 144, the data wasn't being passed back (like you did for service items). I just today added the bit that copies your code and should be passing it all back now (it's working for me at least!)

pkg/services/mto_shipment/mto_shipment_creator.go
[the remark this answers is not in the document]
sandy-wright: yeah, like deep said, this is a slightly different error check than the one swagger does

pkg/services/mto_shipment/mto_shipment_creator_test.go
```diff
@@ -121,9 +109,31 @@ func (suite *MTOShipmentServiceSuite) TestCreateMTOShipmentRequest() {
...
+	shipment.SecondaryDeliveryAddressID = nil
+	shipment.SecondaryDeliveryAddress = nil
+	shipment.ID = uuid.Nil
+	if len(shipment.MTOAgents) > 0 {
+		for _, agent := range shipment.MTOAgents {
+			agent.ID = uuid.Nil
+			agent.MTOShipmentID = uuid.Nil
+		}
+	}
+
```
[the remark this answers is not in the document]
sandy-wright: hmm, we don't _need_ to because we manually create the service items list in that test, and we don't set any IDs there (as opposed to having the data gen manually create a fake service item). If you want me to add them, I can though

pkg/services/mto_shipment/mto_shipment_creator.go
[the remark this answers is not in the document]
sandy-wright: oh I misunderstood! swagger does check for nil, but I added it here because you can accidentally send in nil data in the go tests, and I wanted to prevent tests from crashing completely because of that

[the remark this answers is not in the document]
sandy-wright: it's a result of a brief bit of pure TDD hahaha

pkg/services/mto_shipment/mto_shipment_creator.go
chrisrcoles: It's working great!

PR #4460, opened 2020-07-23
[MB-2621] TIO calculate total amount approved for service items

duncan-truss:
## Description

This PR keeps track of the total approved amount when a TIO reviews service items of a payment request.  The sum total of all approved service items for a given payment request is updated in real time when the status changes on any service item.  When a service item is approved the total will increase and it will decrease if a service items status switches from approved to reject or the clear selection button is pressed.

The component was modified to also take in service items with an existing rejection reason value and display it on load.

## Reviewer Notes

I tried to anticipate some of the needs of the final completion screen that displays the requested, rejected, and approved totals all at once.  The requested amount shouldn't change so it won't need to be part of the state.

Previously we used the built in handleChange event listener and also used a custom onClick listener to clear the form values.  I decided that the clear selection listener could use the same one as onChange from our radio inputs because we may be recomputing the totals if the item was previously approved.  When not using the presupplied formik handlers you need to call `setValues` to update the form values.

**One open question is if we should be using the SUBMITTED status instead of no status, assuming this is what the api will use?**  I don't think it will change the behavior of this component but may be worth aligning on in our test data.

## Setup

```sh
make server_run
make office_client_run
```

1. Log in as the TIO office user
2. Click on a payment request row in the queue [http://officelocal:3000/moves/1/payment-requests/a2c34dba-015f-4f96-a38b-0c0b9272e208](http://officelocal:3000/moves/1/payment-requests/a2c34dba-015f-4f96-a38b-0c0b9272e208)
3. Approve different service items and check that the amount is added to the total
4. Reject an approved service item and check that the amount is subtracted from the total
5. Clear an approved service item and check that the amount is subtracted from the total

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2621) for this change
* [Storybook](http://localhost:6006/iframe.html?id=too-tio-components-reviewserviceitems--with-status-and-reason)
* [Abstract Reference](https://share.goabstract.com/3cff868b-be81-401c-82fc-3634b1e26b52?collectionLayerId=fdf375ce-9025-4e93-8b33-093e8c27ebfc&mode=design)

## Screenshots

service-items-approved-amount


duncan-truss:
> this looks good! the alternative approach I mentioned where you could avoid using state would mean defining a function that takes the form `values`, handles the calculations and returns the approved/rejected amounts, and then using that in the render, i.e.:
> 
> ```
> <div data-testid="approvedAmount" className={styles.totalAmount}>
> {getApprovedTotal(values)}
> </div>
> ```
> 
> that way the total remains a purely computed value and you wouldn't need to worry about maintaining the state values. But since this component is pretty simple I don't think it's a big deal either way.
> 
> the main remaining change to make is just updating `data-cy` to `data-testid` since that change has been merged to master now (I'm happy to handle this since you'll be out!)

Yea it irked me to have to recompute those values and that the rejected totals would be rechecking the same items.  What you suggest feels pure-er because it's just a computed value.  I knew some of the events there wouldn't need to be a recomputation so that's what I liked about it.

suzubara:
Ok, after taking another look I decided to move the total calculation into the render function instead of handling it as state, bc I got nervous about having to handle the form input onChange manually.

Shkeating:
> It's so beautiful 😭 I'm still having trouble accessing staging so my feedback's based on eyeballing the screen capture (I've set aside time this afternoon to troubleshoot) - so apologies if any of this is inaccurate.
> 
> Some visual notes:
> 
>     * The color of the active radio buttons, `Clear selection` link, and `Next`/`Previous` buttons looks off. They should use the `link` color in Storybook.
> 
>     * There should be 8px padding between the `Accept` and `Reject` buttons. In the video it looks closer to 12?
> 
>     * Should be 16px between the `Reject` button and `Clear selection`
> 
>     * Should be 24px padding at the bottom of the service item container.
> 
> 
> Again, sorry if these are done and I'm just misinterpreting the screencap!

The colors have been adjusted since the sceencap, but I think the spacing adjustments could be made!

PR #4461, opened 2020-07-23
[MB-3248] Connect Create MTO Shipment to HHGMoveSetup form

Ronolibert:
## Description

Connect createMTOShipment endpoint to HHGMoveSetup form and add integration tests to HHG flow

## Reviewer Notes
- Just need to finish the integration tests for HHG move setup form and listener to the API endpoint being called

## Setup
`make server_run`
`make client_run`
Log in
Go through flow and try to set up an HHG move

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3248) for this change


src/components/Customer/HHGDetailsForm.jsx
```diff
@@ -25,22 +30,77 @@ class HHGDetailsForm extends Component {
...
+    deliveryLocation,
+    receivingAgent,
+    releasingAgent,
+    remarks,
+  }) => {
+    const { createMTOShipment } = this.props;
+    const { hasDeliveryAddress } = this.state;
+    const mtoShipment = {
+      // TODO: Use moveTaskOrderID when it is available
+      moveTaskOrderID: '5d4b25bb-eb04-4c03-9a81-ee0398cb779e',
```
kimallen, line 46: You should be able to get this ID from querying for MTOs on the Order.  You can get the orderId using the selector `selectActiveOrLatestOrders` (which comes from the `userInfo` in redux).  Probably need to build a fetch in modules for the latest MTO on the orders.  (In `Entities/modules/moveTaskOrders.js`?

src/components/Customer/HHGDetailsForm.jsx
```diff
@@ -1,13 +1,18 @@
 import React, { Component } from 'react';
-import PropTypes from 'prop-types';
+import { arrayOf, string, bool, shape, func } from 'prop-types';
```
kimallen, line 2: Aah, doing this is a nice change.  Makes the proptypes definitions much more readable- thanks!

src/components/Customer/HHGDetailsForm.jsx
```diff
@@ -133,43 +193,58 @@ class HHGDetailsForm extends Component {
...
   initialValues: {},
 };
 
-export default HHGDetailsForm;
+const mapDispatchToProps = {
+  createMTOShipment: createMTOShipmentAction,
+};
+
+export { HHGDetailsForm as HHGDetailsFormComponent };
+export default connect(null, mapDispatchToProps)(HHGDetailsForm);
```
kimallen, line 250: You may want to name this to indicate which of these is the connected component and which is not.  Something like `ConnectedHHGDetailsForm` (a previous suggestion by Suz) for the connected one.  And then you don't need to export the un-connected form for storybook with a different name.

swagger/internal.yaml
```diff
@@ -2610,19 +2610,11 @@ definitions:
         format: date-time
         type: string
         readOnly: true
-      scheduledPickupDate:
```
kimallen, line 2613: good catches!

src/components/Customer/HHGDetailsForm.jsx
Ronolibert: I think I did it this way because for most situations the connected component is the useful component, I just needed to export the unconnected component itself solely test purposes

src/components/Customer/HHGDetailsForm.jsx
Ronolibert: Ended up fixing this in this PR https://github.com/transcom/mymove/pull/4463

src/components/Customer/HHGDetailsForm.jsx
lynzt: I think the name is fine for now, but we might wanna address it when we split up the form into multiple pages...

Ronolibert:
@lynzt I haven't seen that behavior, but we might be able to narrow down that behavior when the flow is more complete

PR #4463, opened 2020-07-24
[MB-2999] Pickup Location Use Current Residence checkbox

Ronolibert:
## Description

Adds a checkbox in the pickup location section to automatically use current residence location

## Setup
`make server_run`
`make client_run`
Go to HHG flow
Click on checkbox, it should populate fields under it

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2999) for this change

## Screenshots

use-current-residence


src/shared/Entities/modules/user.js
```diff
@@ -0,0 +1,16 @@
+import { swaggerRequest } from 'shared/Swagger/request';
+import { getClient } from 'shared/Swagger/api';
+
+export const showLoggedInUserLabel = 'ServiceMember.showLoggedInUser';
+
+export function showLoggedInUser() {
```
Ronolibert, line 6: I had to add in this endpoint and usage of it because I found out that we don't have access to the user's residential info until we refresh the page. So there was a situation where if the customer goes through the flow in one sitting, those fields would be empty.

src/components/form/AddressFields/AddressFields.jsx
```diff
@@ -3,11 +3,12 @@ import PropTypes from 'prop-types';
 import { Fieldset, Label, TextInput } from '@trussworks/react-uswds';
 import { v4 as uuidv4 } from 'uuid';
 
-export const AddressFields = ({ legend, className, values, handleChange, name }) => {
+export const AddressFields = ({ legend, className, values, handleChange, name, renderCurrentResidenceCheckbox }) => {
```
kimallen, line 6: Since this is a generic address field component that could be used for any address, I'd give the checkbox a more generic name, like `renderExistingAddressCheckbox`.

src/shared/Entities/modules/user.js
kimallen: This makes it so that we now store user in Entities and outside of entities.  We want to avoid that.  @suzubara  had said that it was ok to have the user outside of entities, so I think it would be better to use the existing `GetLoggedInUser` fn in `getLoggedInUserInfo` OR if there's not too much setup, it looks like there's a new saga we can use in `sagas/auth.js`.  Might be worth checking with Suz on use of the saga.

Ronolibert: I did try both and I got weird behavior of the entire page flickering in the UI and also losing redux's state in the process which broke the route. I think once we have a way of saving selected move type then we should be able to simplify these calls

PR #4468, opened 2020-07-24
Redo: MB-2650 update save-fuel-price-data scheduled task to save-ghc-fuel-price-data

kilbergr:
## Description

CHANGE IN APPROACH: 
We are replacing `save-fuel-price-data` with `save-ghc-fuel-price-data` as a scheduled task to be run every night at 9:30am UTC. Since replacement might break prod if the timing isn't tight enough, I'm adding them both first and then removing the one that's no longer running on a schedule.

## Reviewer Notes

I believe the variables required per each environment are the same for `save-ghc-fuel-price-data` as for `save-fuel-price-data`. 

## Setup

This is going to be done in conjunction with this PR: (reference https://github.com/transcom/terraform-aws-app-environment/pull/25 and one other in the infrasec repo) 

* [X] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
* [] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2650) for this change


config/env/experimental.save-ghc-fuel-price-data.env
```diff
@@ -0,0 +1,6 @@
+DB_IAM=true
+DB_NAME=app
+DB_PORT=5432
+DB_SSL_MODE=verify-full
+DB_SSL_ROOT_CERT=/bin/rds-ca-2019-root.pem
```
esacteksab, line 5: Something to take note of, this won't work in GovCloud as `rds-ca-2019` doesn't exist. So we should create a ticket capturing this when we migrate these jobs to GovCloud.

cblkwell: Ugh. This is really annoying, and it's yet another reason why encoding this stuff in the application code is frustrating. If the environment in both commercial and GovCloud is called "experimental", this stuff will be conflicting, whereas if we encode it into the environment via SSM it can be set based on the infrastructure, not what the application thinks.

I'm more and more convinced that the application should have *no idea* of the concept of environments, and just accept parameters that infrastructure passes to it.

esacteksab: I 100% agree with you. I feel like I've spent the last 5+ years of my career telling dev-like folks, the app should give zero f's about its environment.

kilbergr: https://dp3.atlassian.net/browse/MB-3477

PR #4472, opened 2020-07-27
Add ADR and demo code for ReactQuery

suzubara:
## Description

This ADR proposes using [React Query](https://react-query.tanstack.com/) to handle API data management on the frontend, starting with the new Office app pages. I also included an example, used to populate the PaymentRequestIndex page to demonstrate the differences in implementation. We can decide whether or not the code changes should be included in this PR.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -603,6 +603,11 @@ redis-cli
 a11y
 npm
 userbase
+ApolloClient
+GraphQL
+SwaggerRequest
+SwaggerClient
+polyfill
 # Put all custom terms BEFORE this comment, lest 'pre-commit' and 'make spellcheck' yield different errors.
  - ./docs/backend/route-planner.md
  - docs/adr/0050-doc-viewer-fork.md
--- a/docs/adr/0053-use-react-query-office-app.md
+++ b/docs/adr/0053-use-react-query-office-app.md
@@ -1,0 +1,57 @@
+# Use React Query for Office App API interactions
+
+We are currently making heavy use of [Redux](https://redux.js.org/) in both the Office app & customer-facing app to manage API data in the browser. While there are other valid reasons to continue using Redux, it does put a lot of overhead on our code to handle fetching data, providing the correct data to the correct UI components, and deciding when to invalidate or re-fetch data that has become stale. [This article (_Why I Quit Redux_)](https://dev.to/g_abud/why-i-quit-redux-1knl) explains pretty well why this is less than ideal, and also proposes using a library called [React Query](https://react-query.tanstack.com/) as an alternative.
+
+After seeing that React Query follows many of the same conventions as ApolloClient (a client data-layer tool used with GraphQL), I decided to investigate further, since using ApolloClient on previous projects significantly reduced the amount of overhead work when hooking up the API to the frontend. After testing this out with a basic existing query, I believe if we take advantage of this tool, it will both reduce complexity of our frontend codebase and speed up our progress. Please see the [Pros and Cons](#pros-and-cons-of-the-alternatives) section for more technical details and comparison points.
+
+## Considered Alternatives
+
+- Continue using SwaggerRequest thunk as-is (do nothing)
+- Use [Redux-Saga](https://redux-saga.js.org/) for managing API interactions
+- Use React-Query
+
+## Decision Outcome
+
+- Chosen Alternative: React-Query for Office app (starting with TXO)
+- It's important to note that none of these options are incompatible with each other, meaning this is not an all-or-nothing decision. _However_ we want to avoid using too many patterns simultaneously, so I recommend scoping this change to an explicitly defined area of the codebase (in this case, the TXO pages).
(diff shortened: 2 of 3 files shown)


src/services/ghcApi.js
```diff
@@ -0,0 +1,64 @@
...
+    // eslint-disable-next-line no-console
+    console.log('Operation does not exist', operationPath);
+  }
+
+  let request;
+  try {
+    request = operation();
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.error('operation failed', e);
```
gidjin, line 51: I wonder if we should make an exception for `console.error` and maybe `console.debug`

src/services/ghcApi.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  return request
+    .then((response) => {
+      const payloadSchema = schema.paymentRequests;
+      return normalize(response.body, payloadSchema).entities;
+    })
+    .catch((response) => {
+      // eslint-disable-next-line no-console
+      console.log('response failed', response);
```
gidjin, line 61: `console.error` instead maybe?

src/scenes/Office/TIO/paymentRequestIndex.jsx
```diff
@@ -1,52 +1,50 @@
 import React from 'react';
```
gidjin, line 1: Should this file also move out of `src/scenes`? Though that may be more changes than this ADR should have.

src/scenes/Office/TIO/paymentRequestIndex.jsx
```diff
@@ -1,52 +1,50 @@
...
-import { connect } from 'react-redux';
-import { bindActionCreators } from 'redux';
-import { getPaymentRequestList, selectPaymentRequests } from 'shared/Entities/modules/paymentRequests';
-import { Link, withRouter } from 'react-router-dom';
+import { Link } from 'react-router-dom';
+import { useQuery } from 'react-query';
 
-class PaymentRequestIndex extends React.Component {
-  componentDidMount() {
-    // TODO - only get payment requests associated with the given move order ID
```
gidjin, line 9: Do you need this TODO, it didn't make it over into the new request.

src/services/ghcApi.js
tinyels: Comment about comment: There is a STIG about not showing stack traces, so it's good eslint warns about these in general.

suzubara:
I'm leaving the code changes in while this ADR is in review to serve as an example, but will revert them once the ADR is approved before merging.

src/services/ghcApi.js
```diff
@@ -0,0 +1,64 @@
...
+  let request;
+  try {
+    request = operation();
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.error('operation failed', e);
+  }
+
+  return request
+    .then((response) => {
```
duncan-truss, line 55: What happens on non 2XX responses that aren't rejected, do we need access to the response code/status not just the body?

suzubara: yeah but this is just an example, the actual implementation would include real error handling

duncan-truss: Okay, I saw you mentioned ReactQuery having retry logic so thought that might be included here.  That seems to be part of the component hook?

Do we know how it integrates with the If-Match logic used by the backend?

suzubara: yeah there is a default retry behavior we can customize (https://react-query.tanstack.com/docs/guides/query-retries). As far as If-Match, we will need to make sure to handle this in the code that makes HTTP calls and the logic for updating the cache (which I don't believe is any different from what we do now).

suzubara:
We will need to customize the retry behavior because [SwaggerClient](https://github.com/swagger-api/swagger-js/blob/master/docs/usage/http-client.md#status-error) actually rejects responses that are not 2xx (which differs from the `fetch` spec). So we'll probably want to modify it to only retry if the returned error is network-related (instead of on any API error).

suzubara:
I would like to move this ADR along since it will be impactful on the impending TXO work. Any files in this PR that are _not_ the ADR itself will be reverted before merging. I think the next step after this is accepted is to work on a story that establishes some baseline patterns for using React Query with complete examples (such as tests), and ensures parity with our other API interactions. I imagine this would include:

- Identify an existing TXO API interaction that can be swapped out for React Query and used as a first example
- Add a function that mirrors `SwaggerRequest` behavior to make actual API calls (but outside of the context of Redux)
- Ensure configuration (such as retries, error logging, etc.) meets our specific needs (documentation [here](https://react-query.tanstack.com/docs/api#reactqueryconfigprovider))
- Include tests and any utilities needed to test components that use React Query hooks (ie, [ReactQueryCacheProvider](https://react-query.tanstack.com/docs/api#reactquerycacheprovider))

If there are any other specific requirements or concerns about the above, please let me know!

PR #4494, opened 2020-07-27
[MB-2933] Create event package to trigger events

shimonacarvalho:
## Description

The event package is meant to handle triggering events from the rest of the codebase. It does not store these events but passes them on to handlers which may store the events. One handler is included as an example in this PR. It's a skeleton for the notification handler and it receives the info from the event pkg, assembles a payload and simply logs into the logger. 

Read more here https://docs.google.com/document/d/1tq2jWvuLXEv30Bab2S-nSj9V1bAzLrPk7DZo3xLX9-o/edit#heading=h.9y51v17t3a9c

## Reviewer Notes

One sample event trigger has been placed into the handler for the support endpoint `updatePaymentRequestStatus`. So if you approve a paymentRequest using this endpoint it should trigger the event and you should see the payload on the log. 

To test use prime-api-client or postman to approve the payment request.

## Useful links
* [MB-2933 Push Enable: Create event package to trigger event when paymentRequest is approved - Jira](https://dp3.atlassian.net/browse/MB-2933)
* [Documentation for updatePaymentRequestStatus](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/transcom/mymove/master/swagger/support.yaml#operation/updatePaymentRequestStatus)
* [Prime API Client usage](https://github.com/transcom/prime_api_deliverable/wiki/Prime-API-Client)


pkg/handlers/supportapi/payment_request.go
```diff
@@ -119,6 +121,19 @@ func (h UpdatePaymentRequestStatusHandler) Handle(params paymentrequestop.Update
 		}
 	}
 
+	_, err = event.TriggerEvent(event.Event{
```
garrettqmartin8, line 124: This feels like a nice API!

pkg/services/event/event.go
```diff
@@ -0,0 +1,150 @@
...
+		if !success {
+			err := services.NewEventError(fmt.Sprintf("Endpoint Key %s was not found in endpoints. Must use known endpoint key.", event.EndpointKey), nil)
+			return nil, err
+		}
+	}
+
+	event.clientCert = authentication.ClientCertFromRequestContext(event.Request)
+
+	// Get logger info
+	if event.clientCert != nil {
```
garrettqmartin8, line 129: Just to confirm, this conditional distinguishes an event in the Prime API using mTLS from one of our internal APIs?

pkg/services/event/notification.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/transcom/mymove/pkg/services"
+)
+
+// This file is an example notifications handler to exercise the events package
+// MB-2935 will flesh it out to create a DB table for the notifications payloads
+// These functions just log data instead.
+
+// PaymentRequestModelToPayload This is an example to log the data
+// Will need to move to payloads to model file.
+func paymentRequestModelToPayload(paymentRequest *models.PaymentRequest) *primemessages.PaymentRequest {
```
garrettqmartin8, line 21: Should this be exported like the other functions in the file?

pkg/services/event/event.go
```diff
@@ -0,0 +1,150 @@
...
+	if !success {
+		err := services.NewEventError(fmt.Sprintf("Event Key %s was not found in eventModels. Must use known event key.", e), nil)
+		return nil, err
+	}
+	return eventModel.ModelInstance, nil
+}
+
+// RegisteredEventHandlerFunc is a type of func that can be registered as an event handler
+// to be called by the eventing system
+type RegisteredEventHandlerFunc func(event *Event) error
```
garrettqmartin8, line 82: Ah cool, so we could have an event handler responsible for logging, one for saving to the db, one for auditing, etc.?

pkg/services/event/event.go
shimonacarvalho: Well right here it just gets the logger. Once in notification handler, we can also use it to figure out the source of the call and decide whether to send the notification.

pkg/services/event/notification.go
shimonacarvalho: Oops yeah. My next ticket is moving all those to a model_to_payload file and creating a record vs. logging but I'll update that here.

pkg/services/event/event.go
shimonacarvalho: Yup, we just pass the handler the event and they can do what they need. Any errors from one handler do not stop us running the next, but all errors will be consolidated into one error message and logged as an error level log.

shimonacarvalho: In the future we could even make functions for dynamic registration of new handlers but didn't seem useful right now.

pkg/handlers/supportapi/payment_request.go
shimonacarvalho: Thanks! Hope it ends up lasting. Futzed with it a lot, originally had 2 calls and then decided the simpler the better.

pkg/handlers/supportapi/payment_request.go
```diff
@@ -119,6 +121,19 @@ func (h UpdatePaymentRequestStatusHandler) Handle(params paymentrequestop.Update
...
+		EventKey:        event.PaymentRequestUpdateEventKey,
+		MtoID:           mtoID,
+		UpdatedObjectID: updatedPaymentRequest.ID,
+		Request:         params.HTTPRequest,
+		EndpointKey:     event.SupportUpdatePaymentRequestStatusEndpointKey,
+		DBConnection:    h.DB(),
+		HandlerContext:  h,
+	})
+	if err != nil {
+		logger.Error("primeapi.CreatePaymentRequestHanders could not generate the event")
```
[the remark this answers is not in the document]
shimonacarvalho: Oh yeah that is a relic thanks! Will fix.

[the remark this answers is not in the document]
shimonacarvalho: > I was wondering about the organization of the notification code a little bit - you include that file in `pkg/services/events` for now, but do you imagine it going somewhere else (such as in its own package, maybe) in the future? Or maybe a subfolder within the events package? I ask just because it seems like there will eventually be a lot of code in there, so one file might not be super maintainable.

I think it might have more files but I don't think it needs to be it's own pkg yet. So far I believe it will just need that file + a model_to_payload file. The other reason being that there is a notifications pkg elsewhere in the system (it sends emails I think) so I want this to clearly be associated with this event package as it should only be called by the event code. But if it gets unwieldy we can always split it out.

pkg/services/event/notification.go
```diff
@@ -0,0 +1,112 @@
...
+		zap.String("objectID", event.UpdatedObjectID.String()))
+
+	//Get the type of model which is stored in the eventType
+	modelBeingUpdated, err := GetModelFromEvent(event.EventKey)
+	if err != nil {
+		return err
+	}
+
+	// Based on which model was updated, construct the proper payload
+	switch modelBeingUpdated.(type) {
```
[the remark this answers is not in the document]
shimonacarvalho: Thanks! It took some searching to find good ways to do this. We still need custom code per event type but I think we can't get away from that as I know we will eventually need some custom handling. But at least we can probably make a function that encapsulates some of the common stuff so that its not a ton of copy paste.

pkg/handlers/supportapi/payment_request.go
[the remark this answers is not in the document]
sandy-wright: okay, that makes sense to me. Thanks for the explanation!

PR #4504, opened 2020-07-28
MB-3474 Remove references to move order in tests

monfresh:
MB-3474 Remove references to move order in tests

This is some small follow-up clean up work after consolidating the 
`orders` and `move_orders` tables. In the previous PR, we limited the
changes to only those necessary to merge the tables and nothing more,
to keep things simple and to avoid breaking APIs. This left references
to "move orders" even though the table and model are now gone.

In this PR, to keep things simple and safe, we are only removing
references to move orders in test data generation files and related
tests. We will clean up the rest of the app in future PRs.

**Reviewer Notes**

Previously, the test data generation for "move orders" created service 
members with random names and grades. That was not the case with the 
test data generation for orders, so we are no longer generating random 
data here unless a field is meant to be unique. We don't seem to 
generate random data in many other places in testdatagen, so I'm 
wondering what we may be losing by no longer doing this.

I would like a review from folks who depend on local data being available
in the TOO interface to make sure all the expected data is still being generated.

**Setup**
```
make clean && make build
make db_dev_reset && make db_dev_migrate
```


pkg/services/move_task_order/move_task_order_fetcher_test.go
```diff
@@ -10,9 +10,9 @@ import (
 )
 
 func (suite *MoveTaskOrderServiceSuite) TestMoveTaskOrderFetcher() {
-	expectedMoveOrder := testdatagen.MakeMoveOrder(suite.DB(), testdatagen.Assertions{})
+	expectedOrder := testdatagen.MakeOrder(suite.DB(), testdatagen.Assertions{})
```
reggieriser, line 13: Nit: This could call the simpler `MakeDefaultOrder` function (see several cases of this in the PR).

pkg/testdatagen/make_order.go
```diff
@@ -9,6 +10,39 @@ import (
...
+		"W_1",
+		"W_2",
+		"W_3",
+		"W_4",
+		"W_5",
+		"AVIATION_CADET",
+		"CIVILIAN_EMPLOYEE",
+		"ACADEMY_CADET",
+		"MIDSHIPMAN"}
+	return grades[rand.Intn(len(grades))]
```
reggieriser, line 43: I'm wondering if we need to preserve this randomness.  Is the lack of predictability when generating a record a concern?  I talked with @Ronolibert just now and he said the original motivation for this was to generating distinct-looking test data for the bootstrapped TOO interface, but that doesn't really apply anymore.  We don't seem to do random data (unless uniqueness calls for it) in other testdatagen makers.

pkg/services/move_task_order/move_task_order_fetcher_test.go
monfresh: Yeah, I noticed that as well. There are many more instances throughout the app, so I thought maybe we can make that a separate PR for the whole app.

pkg/testdatagen/scenario/e2ebasic.go
```diff
@@ -869,14 +869,14 @@ func (e e2eBasicScenario) Run(db *pop.Connection, userUploader *uploader.UserUpl
...
-		MoveOrder:     models.Order{ID: uuid.FromStringOrNil("6fca843a-a87e-4752-b454-0fac67aa4988")},
+	orders := testdatagen.MakeOrder(db, testdatagen.Assertions{
+		Order:         models.Order{ID: uuid.FromStringOrNil("6fca843a-a87e-4752-b454-0fac67aa4988")},
 		ServiceMember: customer,
 	})
 	mto := testdatagen.MakeMoveTaskOrder(db, testdatagen.Assertions{
 		MoveTaskOrder: models.MoveTaskOrder{
 			ID:                 uuid.FromStringOrNil("5d4b25bb-eb04-4c03-9a81-ee0398cb779e"),
-			MoveOrderID:        moveOrders.ID,
+			MoveOrderID:        orders.ID,
```
gidjin, line 879: Is there a plan to also update the MTO to have `OrderID` vs `MoveOrderID`?

monfresh: Yes, in a future PR. When we consolidated the tables, we didn't rename the column names to keep things simple.

monfresh:
@gidjin I fixed the error. After merging Reggie's changes, we had to change the way the service member was being passed in the assertions. Can you check again?

PR #4510, opened 2020-07-29
MB-2934 [PrimeAPI] Create subscriptions table

chrisrcoles:
## Description

This PR adds 2 table relations to our Postgres DB: `webhook_subscriptions` and `webhook_certs`. These tables are complete with attributes, types and restraints according to the [design document](https://docs.google.com/document/d/1tq2jWvuLXEv30Bab2S-nSj9V1bAzLrPk7DZo3xLX9-o/edit#). 

## Reviewer Notes

In-line comments

## Setup

Run mymove `server_test` command.
```
mylaptop ~ %: cd /mymove
mylaptop ~ %: make server_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2934)


pkg/models/webhook_cert.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"time"
+
+	"github.com/gobuffalo/pop"
+	"github.com/gobuffalo/validate"
+	"github.com/gobuffalo/validate/validators"
+	"github.com/gofrs/uuid"
+)
+
+// A WebhookCert represents a certificate that a notification's subscriber uses for authorizing with the notification app
```
shimonacarvalho, line 12: This comment I think should be the other way round. A WebhookCert represents a certificate Milmove will use to authenticate to the subscriber's server to send the notification.

migrations/app/schema/20200729030242_create_webhook_certs_table.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE webhook_certs
+(
+    id uuid PRIMARY KEY NOT NULL,
```
shimonacarvalho, line 3: I'd actually move the contractor_id here. It will be useful to have some context as to who the contractor is for the certs. Or else it's sort of orphaned.

pkg/models/webhook_subscription.go
```diff
@@ -0,0 +1,69 @@
...
+
+// WebhookSubscriptionStatus is a type representing the webhook subscription status type - string
+type WebhookSubscriptionStatus string
+
+// KeyType is a type representing the webhook subscription eventKey type - string
+type KeyType string
+
+const (
+	// WebhookPaymentRequestCreateEventKey is a key containing PaymentRequest.Create
+	WebhookPaymentRequestCreateEventKey KeyType = "PaymentRequest.Create"
```
shimonacarvalho, line 20: When we create/access the entries we should just use the event.KeyType and consts which are exported. I think it's perhaps not a good idea to have a duplication of the type, typename and const values here. Very easy to get out of sync and have typing issues.

pkg/models/webhook_subscription.go
```diff
@@ -0,0 +1,69 @@
...
+	// WebhookPaymentRequestUpdateEventKey is a key containing PaymentRequest.Update
+	WebhookPaymentRequestUpdateEventKey KeyType = "PaymentRequest.Update"
+)
+
+const (
```
shimonacarvalho, line 25: Since these are subscriptions and not notifications, I don't quite understand these statuses. I would expect Active and Disabled. Sorry if I said something else when we chatted! These are the statuses I'm using for notifications.

pkg/models/webhook_subscription.go
```diff
@@ -0,0 +1,69 @@
...
+	CallbackURL   string                    `db:"callback_url"`
+	Status        WebhookSubscriptionStatus `db:"status"`
+	CreatedAt     time.Time                 `db:"created_at"`
+	UpdatedAt     time.Time                 `db:"updated_at"`
+}
+
+// Validate gets run every time you call a "pop.Validate*" (pop.ValidateAndSave, pop.ValidateAndCreate, pop.ValidateAndUpdate) method.
+func (wS *WebhookSubscription) Validate(tx *pop.Connection) (*validate.Errors, error) {
+	return validate.Validate(
+		&validators.UUIDIsPresent{Field: wS.SubscriberID, Name: "SubscriberID"},
```
shimonacarvalho, line 51: Nice job on all the validation!

migrations/app/schema/20200729030242_create_webhook_certs_table.up.sql
```diff
@@ -0,0 +1,8 @@
...
+    key text NOT NULL,
```
shimonacarvalho, line 4: Do we know if it's secure enough to have the private key in the db table? I don't know the answer to this.

migrations/app/schema/20200729030242_create_webhook_certs_table.up.sql
chrisrcoles: Not sure if I understand this comment fully. Let's talk more when you're back on this piece!

pkg/models/webhook_subscription.go
chrisrcoles: I feel the same way! I wanted to keep these in just the [events package](https://github.com/transcom/mymove/blob/master/pkg/services/event/event.go), but unfortunately we'll need to duplicate those values if we want to avoid any circular import errors. Importing the `services/event/event.go` file here results in circular import errors. The alternative would be to define these values as constants in new place - neither the events package or the models package to prevent any circular imports.

pkg/models/webhook_subscription.go
chrisrcoles: I think I was confused after speaking with you on Tuesday! I'll remove these!

migrations/app/schema/20200729030242_create_webhook_certs_table.up.sql
chrisrcoles: As security is an issue, I think the best option would be to encrypt it and store the encrypted value in our database, similar to how a password would be stored. I'm currently working on getting more information from #prac-engineering ([see slack thread](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1596230922328100)).

pkg/models/webhook_subscription.go
shimonacarvalho: I think the easiest way is just don't define EventKeys in the model at all, and remove the validation. While it would be great to have full validation here, it would require moving too much code from events (where it makes sense and needs to be used by the devs) to the model. We can just be careful to validate the data ourselves before we run ValidateAndCreate. Since these subscriptions will be updated rarely, that should be fine.

migrations/app/schema/20200729030242_create_webhook_certs_table.up.sql
shimonacarvalho: Ah okay one thing - we would need to be able to decrypt to get the original cert so something like sha256 in `client_certs` can't work. That's a hash and will not return the original data.

chrisrcoles: Ok - as discussed in [this thread](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1596230922328100). We'll drop the table given we only have one subscriber right now - the Prime.

migrations/app/schema/20200803200704_create_subscriptions_table.up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE webhook_subscriptions
+(
+    id uuid PRIMARY KEY NOT NULL,
```
shimonacarvalho, line 3: Hey chris, this looks good except where did the status go? We still need an active/disabled status.

chrisrcoles: I've added it back [here](https://github.com/transcom/mymove/blob/61c9f28604f4becb95f457582f5b87b218173b87/migrations/app/schema/20200803200704_create_subscriptions_table.up.sql#L1)

PR #4514, opened 2020-07-29
Create flag to disable pages for demo

kimallen:
## Description

This removes the dummy landing page and the dummy move info page on staging.  It does not remove those pages for other environments

## Reviewer Notes

using 'disable' creates a double negative situation, so not sure it's too confusing.  but saying 'showForDemo" seemed strange since we are showing everything else for the demo, even without the flag. . .open to feedback.  

## Setup

Locally, the flag is set to false (we are not disabling the page locally).  To check locally, set flag with `?flag:disableForDemo=true`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3487) for this change

### DIFF
--- a/src/shared/featureFlags.js
+++ b/src/shared/featureFlags.js
@@ -17,16 +17,32 @@ const defaultFlags = {
   allOrdersTypes: false,
   hhgFlow: false,
   ghcFlow: false,
+  disableForDemo: false,
 };
 
 const environmentFlags = {
-  development: Object.assign({}, defaultFlags, { allOrdersTypes: true, hhgFlow: true, ghcFlow: true }),
+  development: Object.assign({}, defaultFlags, {
+    allOrdersTypes: true,
+    hhgFlow: true,
+    ghcFlow: true,
+    disableForDemo: false,
+  }),
 
   test: Object.assign({}, defaultFlags),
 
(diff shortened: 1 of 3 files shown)


src/shared/featureFlags.js
```diff
@@ -17,16 +17,32 @@ const defaultFlags = {
...
+  disableForDemo: false,
 };
 
 const environmentFlags = {
-  development: Object.assign({}, defaultFlags, { allOrdersTypes: true, hhgFlow: true, ghcFlow: true }),
+  development: Object.assign({}, defaultFlags, {
+    allOrdersTypes: true,
+    hhgFlow: true,
+    ghcFlow: true,
+    disableForDemo: false,
```
lynzt, line 28: delete `disableForDemo` here... i think we can rely on the default value here...

src/shared/featureFlags.js
```diff
@@ -17,16 +17,32 @@ const defaultFlags = {
...
+  }),
 
   test: Object.assign({}, defaultFlags),
 
-  experimental: Object.assign({}, defaultFlags, { allOrdersTypes: true, hhgFlow: true, ghcFlow: true }),
+  experimental: Object.assign({}, defaultFlags, {
+    allOrdersTypes: true,
+    hhgFlow: true,
+    ghcFlow: true,
+    disableForDemo: false,
```
lynzt, line 37: delete `disableForDemo` here... i think we can rely on the default value here...

PR #4515, opened 2020-07-29
MB-3398 Support API: Fix load testing task for createMoveTaskOrder endpoint

sandy-wright:
## Description

This PR has some changes that are complimentary to the load testing changes needed to make the `create_move_task_order` task function properly. Namely, I changed some of the properties in the `MoveOrder` definition to match their column name in the db, and I added a `maxItems` key to `MTOAgents`.

## Setup

The two endpoints to test to check that the changes are working properly are `createMoveTaskOrder` in Support API and `createMTOShipment` in Prime API (to check that there really is a max of 2 agents now). Follow the instructions [here](https://github.com/transcom/prime_api_deliverable/wiki/API-Endpoints) to test these endpoints.

You may also test them using [the load testing branch for MB-3398](https://github.com/transcom/milmove_load_testing/pull/26).

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3398) for this change.
* [Prime API Wiki](https://github.com/transcom/prime_api_deliverable/wiki) for more info on how to test.


swagger/prime.yaml
```diff
@@ -771,6 +771,7 @@ definitions:
     items:
       $ref: '#/definitions/MTOAgent'
     type: array
+    maxItems: 2
```
chrisrcoles, line 774: Nice! This validation is working for me! 
Screen Shot 2020-07-31 at 11 10 45 AM

swagger/support.yaml
```diff
@@ -768,6 +769,13 @@ definitions:
         type: string
         readOnly: true
     type: object
+    required:
```
chrisrcoles, line 772: thanks for adding this!

sandy-wright:
> LGTM! One thing that I'm thinking about is how can we ensure that as we make changes to the prime api we can also hold ourselves accountable to make the relevant changes in the load testing repo. It'd be great if we could eventually add load testing as an acceptance test that gets run as apart of the continuous integration flow we have w/ Github + CircleCI. That way, we would know once merging into mymove if our load testing passes what we expect.

@chrisrcoles I think this is a great idea! Shimona was floating around some thoughts like that too, so maybe we can work it into our flow somehow

PR #4491, opened 2020-07-27
Add ADR for consolidating moves and MTOs tables

monfresh:
This ADR proposes consolidating moves and move_task_orders into one
moves table based on the assumption that a move can only have one MTO
at a given time. We would like input from everyone to make sure our
assumptions are valid.

## Description

Explain a little about the changes at a high level.

## Reviewer Notes

Is there anything you would like reviewers to give additional scrutiny?

## Setup

Add any steps or code to run in this section to help others prepare to run your code:

```sh
echo "Code goes here"
```

## Code Review Verification Steps

* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/wiki/backend#logging)
* [ ] The requirements listed in
 [Querying the Database Safely](https://github.com/transcom/mymove/wiki/backend#querying-the-database-safely)
 have been satisfied.
* Any new migrations/schema changes:
  * [ ] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://github.com/transcom/mymove/wiki/migrate-the-database#zero-downtime-migrations))
  * [ ] Have been communicated to #g-database
  * [ ] Secure migrations have been tested following the instructions in our [docs](https://github.com/transcom/mymove/wiki/migrate-the-database#secure-migrations)
* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](tbd) for this change
* [this article](tbd) explains more about the approach used.

## Screenshots

If this PR makes visible UI changes, an image of the finished UI can help reviewers and casual
observers understand the context of the changes. A before image is optional and
can be included at the submitter's discretion.

Consider using an animated image to show an entire workflow instead of using multiple images. You may want to use GIPHY CAPTURE for this! 📸

_Please frame screenshots to show enough useful context but also highlight the affected regions._

PR #4533, opened 2020-08-02
MB-3114 Consolidate moves and MoveTaskOrders

monfresh:
## Description
In this PR, we are consolidating Moves and Move Task Orders into one Moves table per [this ADR](https://github.com/transcom/mymove/pull/4491). Following the same principles as the previous consolidations, we are trying to make the least amount of changes in this PR that will allow the tests to pass and to minimize breaking APIs.

Below are some notes and questions that came up while working on this:

- To deal with `is_canceled` not being ported over, I kept the reference
to `IsCanceled` in the APIs and added a new function to return the 
correct boolean depending on the `status` of the move.

- We ended up porting over `ReferenceID` because there's still an open question as to whether we need a separate Reference ID in addition to `Locator`. See [Slack discussion](https://ustcdp3.slack.com/archives/CP74RNQ9L/p1596152655036700).

- ContractorID is nullable, but I think because it's not a pointer in the `Move` model, Pop is not letting a Move be saved without a Contractor ID. Looking through the code, I noticed that when we first introduced the `contractor_id` field in the MTO table, we also ran a query to populate all existing MTOs with the same dummy ID. The only places where we currently set the contractor ID are in `convert.go` and via the support API. For now, I've hardcoded the contractor ID to the same dummy one we use in `convert.go` and in tests. Eventually, we will need to be able to identify the current active contractor and use that ID.

### API changes

- I added `Locator`  and `Status` to the support API but I'm not sure if these are attributes that we need the support API to include in the response after creating a move. If not, we can remove them. Note that `Locator` is not required when creating a move because it's automatically generated. It's meant to be a unique random number, so it's not something that would be passed as part of the creation payload. As for `Status`, it is being set with an initial status of `DRAFT`, but you can pass in a different status if desired.

## Setup
```
make clean build
```

## Reviewer notes: 

- Run through the app end to end and make sure everything looks good

- Where we previously had tests for reference id, make sure we have one for locator as well.


migrations/app/schema/20200729152728_move_task_orders_to_moves.up.sql
```diff
@@ -0,0 +1,53 @@
...
+-- NOTICE:  truncate cascades to table "prime_uploads"
+TRUNCATE move_task_orders CASCADE;
+
+-- Add columns that only existed in move_task_orders to moves.
+ALTER TABLE moves
+    ADD COLUMN contractor_id uuid REFERENCES contractors,
+    ADD COLUMN available_to_prime_at timestamp with time zone,
+    ADD COLUMN ppm_type varchar(10),
+    ADD COLUMN ppm_estimated_weight integer,
+    ADD COLUMN reference_id CHAR(9) UNIQUE;
```
reggieriser, line 23: It appears that the previous type of `reference_id` was `varchar(255)` (although looking at the migrations, it had been `char(9)` at some point in the past, but was later changed).  Was this an intentional change in type here?

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -284,3 +284,13 @@ func PaymentServiceItems(paymentServiceItems *models.PaymentServiceItems) *ghcme
 	}
 	return &payload
 }
+
+func isCanceled(move *models.Move) *bool {
+	truePointer := true
```
reggieriser, line 289: I see we have this same `isCanceled` function in three places.  Since it operates on a `models.Move`, could we not just make an `IsCanceled` method on a `models.Move` and then just call it with a `moveTaskOrder.IsCanceled()`?  Also, small thing -- instead of having to create the temp variables `truePointer` and `falsePointer`, you could use `swag.Bool(true)` and `swag.Bool(false)` directly I think.  Looks like we also have a `models.BoolPointer()` that's similar too (not sure why we use both).

pkg/handlers/primeapi/api_test.go
```diff
@@ -52,10 +52,10 @@ func TestHandlerSuite(t *testing.T) {
 }
 
 // MakeAvailableMoveTaskOrder returns a default MTO with AvailableToPrimeAt set. Shared throughout this test package.
-func MakeAvailableMoveTaskOrder(db *pop.Connection) models.MoveTaskOrder {
+func MakeAvailableMoveTaskOrder(db *pop.Connection) models.Move {
```
reggieriser, line 55: I know this was already in place and may be out of scope of this PR, but just noting for future cleanups that this probably should be in the testdatagen package with the other "makers".

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
monfresh: yeah, I discovered swag later on and used it in other places in this PR but forgot to update this one. I wrote a note in the PR description about the duplication here and whether it was worth it to extract it. Putting it in the model sounds good for now.

migrations/app/schema/20200729152728_move_task_orders_to_moves.up.sql
monfresh: Good catch. I probably looked up `reference_id` and found the initial migration, but didn't check to see the last column type. From what I can tell, the change to varchar(255) didn't have a specific reason. Also, the initial migration that altered the `reference_id` column didn't actually do anything because the MTOs table didn't exist. Either way, given there's still confusion about whether or not we need a separate reference ID and what format it should be in, I think `varchar(255)` is good for now.

pkg/handlers/primeapi/api_test.go
monfresh: Good point. I created a new JIRA story for this: https://dp3.atlassian.net/browse/MB-3556

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
monfresh: This is fixed now.

migrations/app/schema/20200729152728_move_task_orders_to_moves.up.sql
monfresh: This is fixed now.

pkg/handlers/supportapi/internal/payloads/payload_to_model.go
```diff
@@ -106,8 +106,8 @@ func MoveTaskOrderModel(mtoPayload *supportmessages.MoveTaskOrder) *models.MoveT
 		model.AvailableToPrimeAt = &availableToPrimeAt
 	}
 
-	if mtoPayload.IsCanceled != nil {
-		model.IsCanceled = *mtoPayload.IsCanceled
+	if mtoPayload.IsCanceled != nil && *mtoPayload.IsCanceled == true {
+		model.Status = models.MoveStatusCANCELED
```
reggieriser, line 110: So if the `IsCanceled` in the payload is `false`, what goes in `model.Status'?  Is it the empty string?  The status is not a nullable field in the database (nor a pointer in the move model), and it's in theory one of `DRAFT`, `SUBMITTED`, `APPROVED`, or `CANCELED` (although it doesn't look the model is validating that).

pkg/models/move.go
```diff
@@ -76,6 +78,14 @@ type Move struct {
 	SignedCertifications    SignedCertifications    `has_many:"signed_certifications" order_by:"created_at desc"`
 	CancelReason            *string                 `json:"cancel_reason" db:"cancel_reason"`
 	Show                    *bool                   `json:"show" db:"show"`
+	AvailableToPrimeAt      *time.Time              `db:"available_to_prime_at"`
+	ContractorID            uuid.UUID               `db:"contractor_id"`
```
reggieriser, line 82: I saw your note about `ContractorID` so this may be something already on your radar -- but if `contractor_id` is nullable in the database then it should be a pointer here (or vice versa, but they should be in sync).  This is the type of issue the `model-vet` tool checks for.  Also notice that we have the model validation commented out below -- should that return?

pkg/models/move.go
```diff
@@ -94,6 +104,7 @@ func (m *Move) Validate(tx *pop.Connection) (*validate.Errors, error) {
 		&validators.StringIsPresent{Field: m.Locator, Name: "Locator"},
```
reggieriser, line 104: If `ReferenceID` is required (for now), should we have a model validator for it like we do for `Locator`?

pkg/handlers/supportapi/internal/payloads/payload_to_model.go
monfresh: When a move gets created via the support API, it automatically gets a status of `DRAFT`. So, if `IsCanceled` is `false` or `nil` it doesn't change the status, but if it's `true`, it changes the status. I will add a test for this.

monfresh: I just pushed a commit that adds these 2 tests and they both pass.

pkg/services/converthelper/convert.go
```diff
@@ -63,11 +63,13 @@ func ConvertFromPPMToGHC(db *pop.Connection, moveID uuid.UUID) (uuid.UUID, error
 			return uuid.Nil, fmt.Errorf("Could not find contractor, %w", err)
 		}
 
-		mto = models.MoveTaskOrder{
-			MoveOrderID:  orders.ID,
+		mto = models.Move{
```
reggieriser, line 66: I'm having a hard time wrapping my head around what the converter is doing now.  Should we always be creating a new `moves` record when the converter kicks in so we don't mess with the PPM move?  I don't think this is doing that because the `moves` record looked up should be the one from the normal PPM process (since it would have the same `orders_id`), and it won't create another record.  But if we use the same `moves` record, won't that cause a problem if somebody just wants to test a normal PPM, for instance?

pkg/services/converthelper/convert.go
```diff
@@ -182,20 +184,6 @@ func ConvertProfileOrdersToGHC(db *pop.Connection, moveID uuid.UUID) (uuid.UUID,
...
-		MoveOrderID:  orders.ID,
-		CreatedAt:    time.Now(),
-		UpdatedAt:    time.Now(),
-		ContractorID: contractor.ID,
-	}
-
-	builder := query.NewQueryBuilder(db)
-	mtoCreator := movetaskorder.NewMoveTaskOrderCreator(builder, db)
-
-	if _, verrs, err := mtoCreator.CreateMoveTaskOrder(&mto); err != nil || (verrs != nil && verrs.HasAny()) {
```
reggieriser, line 196: Why did we remove this here but left the same type of thing above?  I think the original motivation when @kimallen added this was to have an MTO earlier in the onboarding process.  I may just be missing something.

monfresh: I removed this here to get the convert test to pass, but I made a note for myself to ask @kimallen  et al if we still need this file and what it's supposed to do, which I don't fully understand.

pkg/models/move.go
monfresh: This is something we need to figure out. I'm assuming we want to make it required like it used to be, which to me means setting the DB field to **not** be nullable, but that would mean backfilling moves in production. If we don't want to do that, then we can keep it nullable, and we can make it a pointer in the model, but if we add the validation, that would mean having a function that automatically creates the reference ID if it's missing when a move is updated, right?

monfresh: Reggie and I talked about this and decided we'll keep it nullable and not required for now until we decide if we really need this field.

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -16,14 +16,14 @@ type moveOrderFetcher struct {
...
-	// have an associated move_task_order.
+	// have an associated move.
 	var moveOrders []models.Order
 	err := f.db.Q().Eager(
 		"ServiceMember",
 		"NewDutyStation.Address",
 		"OriginDutyStation",
 		"Entitlement",
-	).InnerJoin("move_task_orders mto", "orders.id = mto.move_order_id").All(&moveOrders)
+	).InnerJoin("moves move", "orders.id = move.orders_id").All(&moveOrders)
```
reggieriser, line 26: Nit: Probably can drop the alias here and below since it's just shaving one character off (I know this was probably an automatic change based on what the prior code did here).

pkg/testdatagen/make_move.go
```diff
@@ -48,12 +64,21 @@ func MakeMoveWithoutMoveType(db *pop.Connection, assertions Assertions) models.M
...
-		OrdersID: orders.ID,
-		Status:   models.MoveStatusDRAFT,
-		Locator:  models.GenerateLocator(),
-		Show:     setShow(assertions.Move.Show),
+		Orders:       orders,
+		OrdersID:     orders.ID,
+		Status:       models.MoveStatusDRAFT,
+		Locator:      models.GenerateLocator(),
+		Show:         setShow(assertions.Move.Show),
+		ContractorID: contractorID,
```
reggieriser, line 81: Does this need to set ReferenceID too like the maker above?

pkg/testdatagen/shared.go
```diff
@@ -40,7 +40,7 @@ type Assertions struct {
 	Invoice                                  models.Invoice
 	Move                                     models.Move
 	MoveDocument                             models.MoveDocument
-	MoveTaskOrder                            models.MoveTaskOrder
+	MoveTaskOrder                            models.Move
```
reggieriser, line 43: Let's add a TODO here to get rid of this in a future cleanup like we did before.

pkg/testdatagen/make_move.go
monfresh: Might as well, for consistency. I will add it.

pkg/services/converthelper/convert.go
monfresh: Hopefully we should be able to get rid of this file entirely after this PR is merged: https://github.com/transcom/mymove/pull/4554

pkg/models/move.go
monfresh: Thanks for reminding me about the `model-vet` tool. I wonder why it's not included in the pre-commit hooks. I just ran it and it complained about `Show` being a pointer but it's not a nullable field. I opened a new JIRA story to fix that.

reggieriser: It's not in the pre-commit hooks because there are still some outstanding issues to solve in a few tables.  There's an [epic](https://dp3.atlassian.net/browse/MB-1880) for that work.  In fact, we already have a story in there for [handling the `Show` field](https://dp3.atlassian.net/browse/MB-1990) (it's a particularly tricky one to deal with).  I've marked yours as a duplicate.

shimonacarvalho:
We ran through load testing and manually tested a number of endpoints and nothing seems to be broken.

> I added Locator and Status to the support API but I'm not sure if these are attributes that we need the support API to include in the response after creating a move. If not, we can remove them. Note that Locator is not required when creating a move because it's automatically generated. It's meant to be a unique random number, so it's not something that would be passed as part of the creation payload. 

That is totally fine for us. We have never used locator but it's fine if it is returned. 

> As for Status, it is currently required to be passed in when creating a move, which is a breaking change. We can prevent that though by hardcoding the initial status to DRAFT, but still allow a different status to be passed in if desired. @shimonacarvalho I would love your thoughts on this please.

It is required at which level? DB or swagger? Your suggesting for initial status sounds good to me.

sandy-wright:
I have one comment about the behavior of the `createMoveTaskOrder` endpoint: When load testing creates a record in `moves`, it doesn't set the PPM_TYPE. However, instead of being defaulted to `<null>`, it's saving the empty string in that field. I don't know if this is desirable behavior or not, but we may want to note it for a future ticket if this is not what should be happening. Everything else was working well though.

monfresh:
Thanks for testing it, @shimonacarvalho. I ended up not making Status required in the payload by having it default to draft, but I forgot to update the PR description.

@sandy-wright That's a great question. Thanks for finding that. I will look into it.

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -16,14 +16,14 @@ type moveOrderFetcher struct {
...
-	// have an associated move_task_order.
+	// have an associated move.
 	var moveOrders []models.Order
 	err := f.db.Q().Eager(
 		"ServiceMember",
 		"NewDutyStation.Address",
 		"OriginDutyStation",
 		"Entitlement",
-	).InnerJoin("move_task_orders mto", "orders.id = mto.move_order_id").All(&moveOrders)
+	).InnerJoin("moves", "orders.id = moves.orders_id").All(&moveOrders)
```
reggieriser, line 26: This may not be needed anymore -- or at least in this form.  We put in the join to `move_task_orders` when we merged `move_orders` and `orders` in order to keep every move from appearing in the TOO interface.  Now that this is joining `moves` and `move_task_orders`, I think that every `orders` record has an associated `moves` record, doesn't it?  I think that'd mean we'd see every single move (including PPMs) on staging (all 700+ of them), just like the problem we had before.  Is there some other way we should distinguish moves that appear in the TOO list?  We should dig into this before merging.

monfresh: Good catch. We are addressing this by filtering by moves that have MTO shipments because those are the only ones a TOO would care about. See the latest commit.

pkg/models/move.go
```diff
@@ -488,6 +498,17 @@ func createNewMove(db *pop.Connection,
 		show = moveOptions.Show
 	}
 
+	var contractor Contractor
+	err := db.Where("contract_number = ?", "HTC111-11-1-1111").First(&contractor)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "Could not find contractor")
```
gidjin, line 504: `errors.Wrap` is discouraged. This should use the new `fmt.Errorf` with `%w` as noted in the [Backend Programming Guide](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#use-fmterrorf-and-%25w-verb-when-using-external-librariesl) See the section on errors.

pkg/models/move.go
```diff
@@ -488,6 +498,17 @@ func createNewMove(db *pop.Connection,
...
+	}
+
+	referenceID, err := GenerateReferenceID(db)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "Could not generate a unique ReferenceID")
```
gidjin, line 509: Same as above.

pkg/models/move_test.go
```diff
@@ -18,21 +22,22 @@ func (suite *ModelSuite) TestBasicMoveInstantiation() {
 		"locator":   {"Locator can not be blank."},
 		"orders_id": {"OrdersID can not be blank."},
 		"status":    {"Status can not be blank."},
+		//"contractor_id": {"ContractorID can not be blank."},
```
gidjin, line 25: Is there value in leaving this as a comment? Can it be uncommented or removed?

monfresh: Good catch. We can remove this.

pkg/models/move.go
monfresh: Thanks for this note. I copied and pasted this from existing code, but now is a good time to clean it up.

PR #4539, opened 2020-08-03
MB-2424 | Adding RDS-2017 to container

esacteksab:
RDS-2019 doesn't exist in GovCloud. RDS-2017 does. This makes it so.


esacteksab:
Got confirmation this passes locally also.

PR #4545, opened 2020-08-04
[MB-3521] Fix PPM document viewer route

suzubara:
## Description

This PR fixes the original bug reported, and also adds [redux-persist](https://github.com/rt2zz/redux-persist) to save user role selection across browser windows.

The [original bug](https://dp3.atlassian.net/browse/MB-3521) is that certain users, when logged into the Office app with the PPM role, could not access the PPM version of the document viewer and were shown the TXO interface instead. The root cause turned out to be a combination of things:

- This was only reproducible with users who have multiple roles, with at least one of the non-Office role types (Customer and/or Contracting Officer). This is because when logging in as a user with multiple roles, the Office app defaults to the first role in the array that is returned, and role types are returned by the API in alphabetical order. This resulted in the "active" role being Contracting Officer, which is not handled by the Office app interface. The UX in this case is that at login, the user is presented with a blank page and must select "Change user role" to switch to a PPM/TOO/TIO role in order to view anything.
- The other issue is that the "active" role selected in the "Change user role" page is saved in Redux, which only persists data within a single browser window. Since the document viewer opens in a new window, a user with multiple roles would default back to their "first" role upon opening it. This isn't a current issue for users with PPM & TOO/TIO roles when viewing the PPM document viewer, by virtue of PPM coming first alphabetically.

This PR fixes both of the above issues, by selecting a user's default role from Office roles only, and by persisting the "active" role value across browser windows.

_Additional note:_ this PR adds redux-persist, which can be used to store Redux data in browser local storage. This is a potential avenue for leaking secure data. This PR mitigates risk by both whitelisting what data is stored (currently limited to the `auth` store, which for now only includes the role type a user has), and by clearing that value when the user logs out. Going forward, precaution should be taken when saving more data in Redux Persist to ensure that no PII is stored, and everything is cleared out when a user's session ends.

## Reviewer Notes

To reproduce the bug (not on this branch):
- Log into the Office app as a user with multiple roles, _including non-Office roles_ such as "Customer".
- When you log in, the page will be blank below the milmove header.
- Change to a PPM role by clicking "Change user role" in the top left
- View an existing move, and click on the orders or any document link. It should open in a new window, but display the TXO interface instead of the PPM document viewer.

To reproduce the lack of browser persistence (not on this branch):
- Log into the Office app as a user with multiple _Office roles_ (PPM, TOO, TIO)
- Select a non-default role using the "Change user role" link
- Refresh the page, and note that you see the interface for the default Office role (not the one you had switched to)

## Setup

Test the bug fix:
- Log into the Office app as a user with multiple roles, _including non-Office roles_ such as "Customer".
- When you log in, you should be shown the Office interface for your first Office role (PPM, TOO, or TIO)
- If you select the PPM role using the "Change user role" link, and view a move orders/documents, you should see the expected page open in a new browser window.

Test the redux-persist functionality:
- Log into the Office app as a user with multiple _Office roles_ (PPM, TOO, TIO)
- Select a non-default role using the "Change user role" link
- Refresh the page and verify you still see the interface for the role you switched to.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3521) for this change

## Screenshots

### DIFF
--- a/src/shared/User/LoginButton.jsx
+++ b/src/shared/User/LoginButton.jsx
@@ -5,6 +5,7 @@ import { get } from 'lodash';
 import { selectCurrentUser } from 'shared/Data/users';
 import { isDevelopment } from 'shared/constants';
 import { LogoutUser } from 'shared/User/api.js';
+import { logOut } from 'store/auth/actions';
 
 const LoginButton = (props) => {
   if (!props.isLoggedIn) {
@@ -30,9 +31,14 @@ const LoginButton = (props) => {
       </React.Fragment>
     );
   } else {
+    const handleLogOut = () => {
+      props.logOut();
+      LogoutUser();
+    };
+
     return (
       <li className="usa-nav__primary-item">
-        <a className="usa-nav__link" href="#" onClick={LogoutUser}>
+        <a className="usa-nav__link" href="#" onClick={handleLogOut}>
           Sign Out
         </a>
       </li>
@@ -47,4 +53,9 @@ function mapStateToProps(state) {
     showDevlocalButton: get(state, 'isDevelopment', isDevelopment),
   };
 }
-export default connect(mapStateToProps)(LoginButton);
+
+const mapDispatchToProps = {
+  logOut,
+};
+
+export default connect(mapStateToProps, mapDispatchToProps)(LoginButton);
--- a/src/shared/store.js
+++ b/src/shared/store.js
@@ -1,4 +1,6 @@
 import { createStore, applyMiddleware } from 'redux';
+import { persistStore, persistReducer } from 'redux-persist';
+import storage from 'redux-persist/lib/storage';
 import { appReducer, adminAppReducer } from 'appReducer';
 import { createBrowserHistory } from 'history';
 import { routerMiddleware } from 'connected-react-router';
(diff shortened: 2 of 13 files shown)


src/store/auth/reducer.js
```diff
@@ -1,14 +1,33 @@
...
+      return initialState;
+    }
+
+    case 'GET_LOGGED_IN_USER_SUCCESS': {
+      if (state.activeRole) return state;
+
+      const {
+        payload: { roles = [] },
+      } = action;
+      const firstOfficeRole = roles?.find((r) => officeRoles.indexOf(r.roleType) > -1)?.roleType;
```
duncan-truss, line 21: Non office users will already have an activeRole set is that correct? For instance a user with a customer and contracting officer role.

suzubara: Right now the `activeRole` state is _only_ used in the Office app, so non-Office users would not have any value stored there. The concept of switching between roles doesn't exist in the customer/admin apps AFAIK so I don't know that it's needed.

src/shared/store.js
```diff
@@ -32,14 +34,24 @@ export const configureStore = (history, initialState = {}) => {
   }
 
   const composeEnhancers = composeWithDevTools({});
+
+  const persistConfig = {
+    key: 'root',
+    storage,
+    whitelist: ['auth'],
```
lynzt, line 41: after reading the docs, whitelist is a list of `reducers`, yes? so that would mean any state that is set in that reducer is saved? (just trying to wrap my head around what's actually being whitelisted and how that works)

src/shared/User/LoginButton.jsx
```diff
@@ -30,9 +31,14 @@ const LoginButton = (props) => {
       </React.Fragment>
     );
   } else {
+    const handleLogOut = () => {
+      props.logOut();
```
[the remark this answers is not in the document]
suzubara: `props.logOut` is the Redux action being dispatched, so it won't return anything. It will get hooked into the `auth` reducer, where I've added a case to look for the `LOG_OUT` action and clear Redux state accordingly.

src/shared/store.js
suzubara: yup that is correct, anything in the Redux `state.auth` will persist

PR #4549, opened 2020-08-05
Kim mb 3429 add draft to shipment

kimallen:
## Description

Adds a DRAFT option to the status enum (in db, model, and return payloads).
Changes status in create mtoshipment service from SUBMITTED to DRAFT.  The prime always creates with a SUBMITTED status, so applies that status in the prime mtoshipment create handler.

## Reviewer Notes

Since we don't have a means to change the status to SUBMITTED until we have an "update mtoshipment" api and apply it in the UI, we need to apply SUBMITTED status upon creation.  Therefore, I apply submitted in the handler (instead of service) for the customer's create mtoshipment handler.

The field Status on the MTOShipments table was set to an enum value.  Adding DRAFT required adding two migrations: one to add the default status to the  enum type.  The next to change the default status value, since DRAFT was not accessible in the same transaction in which it was created.  Since we are not in production, we do not have to worry about a gap in when the status is created and when the default is applied.

## Setup

`make server_test`
OR
`go test ./pkg/services/mto_shipment`
`go test ./pkg/handlers/primeapi`

 have been satisfied.
 (https://github.com/transcom/mymove/wiki/migrate-the-database#secure-migrations)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3429) for this change


pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -130,8 +130,8 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
 			return services.NewInvalidInputError(uuid.Nil, nil, nil, "RequestedPickupDate is required to create MTO shipment")
 		}
 
-		//assign status to shipment submitted
-		shipment.Status = models.MTOShipmentStatusSubmitted
+		//assign status to shipment draft
+		shipment.Status = models.MTOShipmentStatusDraft
```
sandy-wright, line 134: We might need to have a little more complicated logic here to make sure it can actually save the status passed in when appropriate (so when it's Prime or when you need to do it for your situation where you can update it to submitted yet)

pkg/handlers/primeapi/mto_shipment.go
```diff
@@ -93,7 +93,7 @@ func (h CreateMTOShipmentHandler) Handle(params mtoshipmentops.CreateMTOShipment
 			return mtoshipmentops.NewCreateMTOShipmentInternalServerError().WithPayload(payloads.InternalServerError(nil, h.GetTraceID()))
 		}
 	}
-
+	mtoShipment.Status = models.MTOShipmentStatusSubmitted
```
sandy-wright, line 96: This will return the payload saying the shipment has status "SUBMITTED" but it won't actually save this way - the shipment record in the DB will have status "DRAFT".

pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -56,7 +56,8 @@ func (h CreateMTOShipmentHandler) Handle(params mtoshipmentops.CreateMTOShipment
 			return mtoshipmentops.NewCreateMTOShipmentInternalServerError().WithPayload(payloads.InternalServerError(nil, h.GetTraceID()))
 		}
 	}
-
+	// TODO: remove this status change once the UpdateMTOShipment api is implemented and can update to Submitted
+	mtoShipment.Status = models.MTOShipmentStatusSubmitted
```
sandy-wright, line 60: Same comment as for the primeapi below (you can tell where my focus was haha)

pkg/services/mto_shipment/mto_shipment_creator.go
tinyels: If the status is passed in as `SUBMITTED` then we also need to do some validation that all the required fields are completed.

pkg/handlers/primeapi/mto_shipment.go
tinyels: Nice catch @sandy-wright

pkg/services/mto_shipment/mto_shipment_creator.go
kimallen: The status was already `SUBMITTED` before modified for `DRAFT`, so there are no changes in which fields are required.  The only difference in required fields from DRAFT state is that when a prime creates an mtoShipment, they have to have a delivery address, which they currently are handling in the yaml.

pkg/handlers/primeapi/mto_shipment.go
kimallen: Thanks for your eagle eyes-  changes are made

PR #4553, opened 2020-08-05
MB-3541 Add all GHC API Endpoints and Internal API Endpoints to map in events pkg

abbyoung:
## Description

In PR #4494, we added the event package to handle triggering events from the rest of the code base. It included endpoint keys and an endpoint map, populated with the info from the Support API.

This PR adds the same information for the GHC api, `ghc.yaml`, and internal api, `internal.yaml`. Data was generated using the same python script that generated the support endpoints info (thanks, @shimonacarvalho!) to maintain consistency. 

Each api has its own file containing the keys and map specific to it. In `endpoint.go`, we pull those maps together in another map, `apiEndpoints`.

## Setup

Run the tests:

```sh
go test ./pkg/services/event -testify.m Test_EventRecord -v
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3541) for this change


pkg/services/event/endpoint.go
```diff
@@ -82,11 +24,25 @@ func (e EndpointType) String() string {
 }
 
 // GetEndpointAPI returns the api name of the endpoint
-func GetEndpointAPI(key EndpointKeyType) string {
-	return endpoints[key].APIName
+func GetEndpointAPI(key EndpointKeyType) *string {
```
shimonacarvalho, line 27: Maybe a good idea to write one test that triggers a GHC endpoint and one that triggers an internal endpoint. New cases in event_test.go. Just to make sure all the logic works.

pkg/services/event/endpoint.go
```diff
@@ -82,11 +24,25 @@ func (e EndpointType) String() string {
...
 // GetEndpointOperationID retuns the operation ID of the endpoint
-func GetEndpointOperationID(key EndpointKeyType) string {
-	return endpoints[key].OperationID
+func GetEndpointOperationID(key EndpointKeyType) *string {
+	for _, endpointMap := range apiEndpoints {
+
+		if endpointInfo, ok := endpointMap[key]; ok {
+			var result = endpointInfo.OperationID
+			return &result
+		}
```
sandy-wright, line 45: Man, this go syntax still makes me uncomfortable, but good to see it being used more and more haha

pkg/services/event/endpoint.go
abbyoung: We can test a ghc endpoint with the current code, since it also has a PaymentRequest endpoint. For the internal api, I think we have to wait until there are more event keys created and added to the existing code. 

I went down the path of trying to manually add things in the test, but it got pretty messy and didn't end up working. If I'm missing something and there's an easier approach, lmk!

pkg/services/event/endpoint.go
shimonacarvalho: Haha yeah, it's very useful but (like a lot of go syntax) not so readable.

PR #4564, opened 2020-08-06
MB-3509 add get MTOShipments for Move ID to internal API

amitch23:
## Description

Adds get MTOShipments for Move ID to internal API. Copied almost directly from the GHC API. I ended up not adding the MTO service for getting the shipments because the service Fetchers already have abstracted DB calls and copied the pattern from GHC.

## Reviewer Notes

Do we currently need a getShipment by shipmentID endpoint as well?

## Setup

Run server unit tests or hit the endpoint with an existing Move Task Order UUID.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/secure/RapidBoard.jspa?rapidView=28&projectKey=MB&modal=detail&selectedIssue=MB-3509) for this change


swagger/ghc.yaml
```diff
@@ -474,7 +474,7 @@ paths:
       parameters: []
       responses:
         '200':
-          description: Successfully retrieved all line items for a move task order
+          description: Successfully retrieved all mto shipments for a move task order
```
kimallen, line 477: Good catch!

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -246,3 +247,87 @@ func (suite *HandlerSuite) TestCreateMTOShipmentHandler() {
...
+			mock.Anything,
+			mock.Anything,
+			mock.Anything,
+		).Return(internalServerErr)
+
+		response := handler.Handle(params)
+		suite.IsType(&mtoshipmentops.ListMTOShipmentsInternalServerError{}, response)
+	})
+
+	suite.T().Run("Failure list fetch - 404 Not Found - Move Task Order ID", func(t *testing.T) {
```
kimallen, line 314: We should probably also add an authorization and an authentication test (for 401 and a 403). Also for a 422 and 500, if those are also possible responses?

swagger/internal.yaml
```diff
@@ -4834,6 +4834,40 @@ paths:
...
+            $ref: '#/definitions/MTOShipments'
+        '400':
+          $ref: '#/responses/InvalidRequest'
+        '401':
+          $ref: '#/responses/PermissionDenied'
+        '403':
+          $ref: '#/responses/PermissionDenied'
+        '404':
+          $ref: '#/responses/NotFound'
+        '422':
```
kimallen, line 4867: What situation would result in a 422?  (not saying it couldn't, just can't think of how, and don't see in the handler that we have a 422 response)

swagger/internal.yaml
```diff
@@ -4834,6 +4834,40 @@ paths:
           $ref: '#/responses/UnprocessableEntity'
         '500':
           $ref: '#/responses/ServerError'
+  '/mto-shipments/{moveTaskOrderID}':
```
kimallen, line 4837: To be RESTful, if we are going to use an id in the path, we should do `moves/<moveTaskOrderID>/mto_shipments`

kimallen: or `move_task_orders/<moveTaskOrderID>/mto_shipments` -- I'm unclear on whether we should be using MTOs or moves

swagger/internal.yaml
amitch23: Removed it until we have a real use case for 422

pkg/handlers/internalapi/mto_shipment_test.go
amitch23: I added one for 401. 500 is already included, and I removed 422. 403 is not being returned

swagger/internal.yaml
amitch23: I changed to the first suggestion. And I agree on that question, and think once Moncef's PR goes in, it will mostly be a syntax choice and won't really "matter", since there are so many references to both `move` and `move_task_order`, even with his changes. I guess we can choose one and then try and stick with it? I left as moveTaskOrderID for now, since in the create MTO shipment, it's also referenced as MTOID, and not move.

swagger/internal.yaml
```diff
@@ -4842,7 +4842,7 @@ paths:
           $ref: '#/responses/UnprocessableEntity'
         '500':
           $ref: '#/responses/ServerError'
-  '/mto-shipments/{moveTaskOrderID}':
+  /moves/{moveTaskOrderID}/mto-shipment:
```
kimallen, line 4845: Can you change this to `mto_shipments` (plural and snake case)?  looks like all the other paths use snake case.

kimallen: That said, looks like the create mtoShipments path used kebab case, and should also use camelCase to be consistent.  If it's easy and you're willing to change that as well, that would be great.

pkg/handlers/internalapi/mto_shipment_test.go
kimallen: Sorry to be nitpicky- I think we should have a 403, and then differentiate between those errors.  Based on generated code, if the user is not logged in, it would be a 401 `ListMTOShipmentsUnauthorized`.  If the user is logged in, but is the wrong type of user, it would be a 403 `ListMTOShipmentsForbidden`.  I think it would just mean separating your check:

```
if session == nil  {
		return mtoshipmentops.NewListMTOShipmentsUnauthorized()
	}

if !session.IsMilApp() {
		return mtoshipmentops.NewListMTOShipmentsForbidden()
	}
```

not sure where `session.ServiceMemberID == uuid.Nil` fits. . .

PR #4566, opened 2020-08-06
MB-3591 Fix MTO shipment updater tests

monfresh:
These 2 integration tests that tested updating a shipment's status to
`approved` were using a shipment of type `HHG`, which isn't handled by
the code that performs the update in `mto_shipment_updater`. These
tests were bypassing the code they were attempting to test. In order
to test the updater, one of the supported shipment types must be used,
such as `HHG_LONGHAUL_DOM`. In turn, the ReServices table must be
populated with the appropriate codes for the shipment to be updated.

### DIFF
--- a/pkg/handlers/ghcapi/mto_shipment_test.go
+++ b/pkg/handlers/ghcapi/mto_shipment_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	routemocks "github.com/transcom/mymove/pkg/route/mocks"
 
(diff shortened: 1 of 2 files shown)


monfresh:
@Ryan-Koch Thinking about this some more, wouldn't it make sense to return an error if an unsupported shipment type was passed in to the updater, similarly to how we check for the proper status transitions?

Ryan-Koch:
> @Ryan-Koch Thinking about this some more, wouldn't it make sense to return an error if an unsupported shipment type was passed in to the updater, similarly to how we check for the proper status transitions?

Hm, yeah there would definitely be value to that. I imagine it'd be easier to debug in the future than seeing the failure to fetch message. Your call though on if you want to include it here, I'll give an approval though in the man time as what you've done to get the test working looks good.

PR #4579, opened 2020-08-07
MB-3576: Adding components to deploy to app-stg env

cblkwell:
## Description

This adds config files and CircleCI components to deploy the MilMove application into the GovCloud `stg` environment. This should be complete with the one caveat that it cannot deploy the mTLS app properly yet because of missing certs. For the moment, I've wired everything up so the images will get pushed to ECR, but the deploy will not happen to `stg` off master yet. Once we get certs, I will replace the placeholder branch name (`REPLACE_WHEN_CERTS_UPDATED`) with `master` and we should be off to the races.


config/env/stg.migrations.env
```diff
@@ -0,0 +1,6 @@
+DB_IAM=true
+DB_NAME=app
+DB_PORT=5432
+DB_RETRY_INTERVAL=5s
+DB_SSL_MODE=verify-full
+MIGRATION_MANIFEST=/migrate/migrations_manifest.txt
```
kilbergr, line 6: Do you not need a MIGRATION_PATH variable here? I see it's set in the makefile commands, so I think it's unnecessary but wanted to double check.

cmd/ecs-deploy/main.go
```diff
@@ -15,7 +15,7 @@ const (
 	environmentFlag  string = "environment"
 )
 
-var environments = []string{"prod", "staging", "experimental", "exp"}
+var environments = []string{"prod", "staging", "experimental", "exp", "stg", "prd"}
```
kilbergr, line 18: Did you add `prd` here in anticipation?

config/env/stg.app-client-tls.env
```diff
@@ -0,0 +1,27 @@
+AWS_CF_DOMAIN=assets.exp.move.mil
```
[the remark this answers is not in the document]
cblkwell: Woops. Not that this matters, since we aren't going to use Cloudfront in GovCloud, but..

config/env/stg.post-file-to-gex.env
```diff
@@ -0,0 +1,7 @@
+DB_IAM=true
```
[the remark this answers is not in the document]
cblkwell: That is now in SSM (via the `app-environment` module).

config/env/stg.migrations.env
cblkwell: I've also moved this to SSM because it has the bucket name in it, so now it can get set directly from the module where the bucket is created.

cmd/ecs-deploy/main.go
cblkwell: Yup.

config/env/stg.migrations.env
kilbergr: That is very nice.
I suspect we can take them out of make commands then.

cblkwell: Looking there, I think we'll actually have to fix those paths or more likely, find a better way to run those migrations.

PR #4580, opened 2020-08-07
Revert "MB-3271 cleanup"

amitch23:
Reverts transcom/mymove#4540


sandy-wright:
@sojeri oh, this can be closed! on Friday when master was going to heck, we tried to revert the latest changes to see if that would put the integration tests on track again. Turns out it was unrelated to the changes in master and was instead related to the absence of some data we pull periodically (I think this is the consensus now)

PR #4581, opened 2020-08-07
ADR for how to treat unfinished work 

amitch23:
## Description

Adds ADR for how to treat unfinished work 

## Reviewer Notes

Feedback pls :) 

Example:
yellowwip


esacteksab:
I think the only thing that came to mind, is how does the styling impact design where perhaps the existing styling is impacted by this? Is that a concern? Or is the expectation that we highlight and folks will dismiss it for the most part? This is before UAT in the sense that you want to get something out there quickly, but you're doing so with styling that may prohibit seeing early design patterns. 

Hopefully that all made sense.

amitch23:
@esacteksab I added a gif in the PR description as an example. The CSS class adds a background color of yellow to an element (`background-color: yellow;`, https://github.com/transcom/mymove/blob/master/src/scenes/Review/Review.css#L87 and https://github.com/transcom/mymove/blob/master/src/App/index.css#L35), a color which was chosen for its rarity as a legit chosen design, at least in this app ;). I don't believe there were design conflicts when this was implemented a couple years back, but for specific examples of how there might be a conflict, I would ask someone like @samadden or @tinyels.

samadden:
I don't think there will be any immediate conflict the use of bright yellow, though it does add some element of distraction, which can in turn make it harder for someone to evaluate the visual design. To that end, something moderately more subtle (`papayawhip`?) might be a substitute CSS color to consider.

My bigger question is what are the rules for when an element would be tagged with this class? Is it more meant to be tagged when the element/feature is marked as not being fully functional, yet, or because the styling is not complete? (My hope would be that this would rarely appear, since we're incrementally adding new features/functionality.)

esacteksab:
> @esacteksab I added a gif in the PR description as an example. The CSS class adds a background color of yellow to an element (`background-color: yellow;`, https://github.com/transcom/mymove/blob/master/src/scenes/Review/Review.css#L87 and https://github.com/transcom/mymove/blob/master/src/App/index.css#L35), a color which was chosen for its rarity as a legit chosen design, at least in this app ;). I don't believe there were design conflicts when this was implemented a couple years back, but for specific examples of how there might be a conflict, I would ask someone like @samadden or @tinyels.

Thank for the example. As I see it, I'm left wondering is the "page" or view what's a WIP or is the element that's highlighted a WIP?

amitch23:
> My bigger question is what are the rules for when an element would be tagged with this class? Is it more meant to be tagged when the element/feature is marked as not being fully functional, yet, or because the styling is not complete? (My hope would be that this would rarely appear, since we're incrementally adding new features/functionality.)

@samadden These are great questions! It seems to me that the answers could come from a conversation with the client + product, since one of the main goals is to decrease confusion during things like demos or interacting with staging. From our side, we can tag things that aren't fully functional (see the examples in ADR for "incomplete" features) or not working totally as expected yet, and in conversation with the client, deduce what is a cause for confusion or what will help them understand what shouldn't be considered "complete". I see one of the functions of this as being a tool for conversation starters with the client to help elucidate points of potential confusion.

kimallen:
I really like this idea for all the reasons Andrea mentioned.  I spent several hours implementing a way to hide pages that were in-progress in order to demo, and it felt like I could have been working on feature work to move those pages forward.  I agree we might have to more clearly define when we use it.  

Another use case might be that we have a new wizard page with a form, and since the wizard page wraps the form and shows the Next/Back buttons, we may have the form built out but the button doesn't actually submit yet. We might highlight a button on the page that isn't fully hooked up yet.  With one week sprints and the preference for having smaller PRs (they are easier to review), we need smaller stories, and this can mean we have something visual up before it's hooked up to the full functionality.

With the client wanting a more clear sense of what is finished and not finished, this will serve as a visual communication for what is finished and what is upcoming.  I think it would be more frustrating as the client to think something is supposed to function and then it doesn't, and then waste time writing a bug report or searching through JIRA to see if it's expected to work yet, when the reality is that it's just not yet built.  

I also think it's worth asking the client what they would prefer.

amy-meng:
This all makes sense and looks good to me. In response to some of the conversation up thread, I personally feel this is a feature whose value is **foremost for TRANSCOM**. 

It may wind up being shown to other audiences, or have incidental value to other audiences, but any implementation decisions around things like color, where the flag appears, etc. should be made primarily in relation to how it impacts TRANSCOM's understanding of what is or is not live.

PR #4088, opened 2020-05-13
Kim mb 2546 show requested incentive and weight

kimallen:
## Description

This displays the requested payment on the landing page if a payment has been requested.  It also re-organizes the display of estimates.
Changes weight section title in Review page from "Weight" to "Pre-move estimated weight" to be clear that the displayed weight and estimate are not based on the submitted weight and estimate.

## Setup

in SM app, log in as one of the requested payment users (ex. `ppm@requestingpayment.newflow`).  Check what is shown on the landing page before and after making a payment request.

## Code Review Verification Steps

* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2546) for this change

## Screenshots

image

PR #4582, opened 2020-08-07
Increase timeout for common Cypress failing tests

monfresh:
We are currently seeing failures in Circle CI on the master branch for
tests in `ppmCloseout.js`, which has happened several times in the
past. The failure has always been due to the test not finding the
element in time. By default, Cypress will only wait up to 4 seconds
total for a command and its assertions to execute.

To increase the timeout, we can pass in the timeout option to a
command, which is what I've done here for the most common failures.
I've increased it to 30 seconds to see if that helps.

I also removed a test that was looking for the word "unknown" in the
`ApprovedMoveSummary.jsx`. My guess is that the "unknown" was
referring to the estimated weight, which is no longer unknown,
possibly due to a recent change, or this one back in May:
https://github.com/transcom/mymove/pull/4088

## References

https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Applying-Timeouts

## Screenshots

Screen Shot 2020-08-07 at 3 02 52 PM

### DIFF
--- a/cypress/integration/mymove/ppmCloseout.js
+++ b/cypress/integration/mymove/ppmCloseout.js
@@ -57,7 +57,7 @@ describe('allows a SM to request a payment', function () {
     cy.location().should((loc) => {
       expect(loc.pathname).to.match(/^\/moves\/[^/]+\/ppm-payment-review/);
     });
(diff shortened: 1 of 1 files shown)


monfresh:
Closing in favor of #4588

PR #4588, opened 2020-08-10
run ppmCloseout tests, but skip tests that rely on BVS

lynzt:
We don't have new BVS scores in our system yet so a check on ppmCloseout fails because it can't calculate an estimate. Comment out this check till new BVS scores are inserted in the system.
This adds the ppmCloseout tests back after being skipped via this pr: https://github.com/transcom/mymove/pull/4583


cypress/integration/mymove/ppmCloseout.js
```diff
@@ -130,7 +130,7 @@ function serviceMemberSubmitsPaymentRequestWithMissingDocuments() {
   );
 
   cy.get('.title').contains('Next step: Contact the PPPO office');
-  cy.get('.missing-label').contains('Unknown');
+  // cy.get('.missing-label').contains('Unknown');
```
gidjin, line 133: Is this the part that is testing for BVS data?

lynzt: Well, it's not testing the BVS directly, but that value(or lack of value) influences what is shown here...

PR #4598, opened 2020-08-12
MB-3619 Add queries to MoveTaskOrder page

donaldthai:
## Description

> On the MoveTaskOrder page, replace the existing data fetch code with React Query, to fetch and populate MTO shipments and MTO service items.
> 
> Notes:
> 
> MoveTaskOrder should be converted from a class component to a functional component
> 
> The queries triggered by componentWillMount should be replaced with useQuery hooks, and Redux connect code should be deleted
> 
> The logic in existing mapStateToProps that maps API data to the format expected by the component may need to be replicated in the useQuery hook.
> 
> This may be a good time to also move the MoveTaskOrder.jsx file into the new file structure (ie., pages/Office/MoveTaskOrder/MoveTaskOrder.jsx)
> 
> AC:
> 
> MoveTaskOrder does not rely on Redux for any API data
> 
> MoveTaskOrder queries the API directly for any data needed (MTO shipments & MTO service items) based on the MTO ID in the URL params

## Reviewer Notes

## Setup

```sh
make db_dev_e2e_populate
make server_run
make office_client_run
```

To test:

1. Login in to Office app as TOO, local
2. Navigate to http://officelocal:3000/moves/5d4b25bb-eb04-4c03-9a81-ee0398cb779e/mto
3. mtoShipments and mtoServiceItems should be loaded

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3619) for this change
* [Article about react-query](https://react-query.tanstack.com/docs/api#usequery)


duncan-truss:
@donaldthai Can I suggest that we need to query the move task order itself?  I know it's needed for the duty station addresses at least.

src/pages/Office/MoveTaskOrder/moveTaskOrder.jsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
```
suzubara, line 1: nit: Can this file be capitalized so it is consistent with our other component files?

src/pages/Office/MoveTaskOrder/moveTaskOrder.jsx
```diff
@@ -0,0 +1,80 @@
...
+  const month = new Intl.DateTimeFormat('en', { month: 'short' }).format(dateObj);
+  const day = new Intl.DateTimeFormat('en', { day: '2-digit' }).format(dateObj);
+  return `${day} ${month} ${year}`;
+}
+
+export const MoveTaskOrder = ({ match }) => {
+  const { moveTaskOrderId } = match.params;
+
+  // TODO - Do something with isLoading and isError?
+  const { mtoShipments, mtoServiceItems } = useMoveTaskOrderQueries(moveTaskOrderId);
```
suzubara, line 31: for error/loading, if you want to use the same logic as PaymentRequestIndex, and return the loading screen/error screen that is probably good enough for now.

src/pages/Office/MoveTaskOrder/moveTaskOrder.jsx
```diff
@@ -0,0 +1,80 @@
...
+function formatShipmentDate(shipmentDateString) {
+  const dateObj = new Date(shipmentDateString);
+  const year = new Intl.DateTimeFormat('en', { year: 'numeric' }).format(dateObj);
+  const month = new Intl.DateTimeFormat('en', { month: 'short' }).format(dateObj);
+  const day = new Intl.DateTimeFormat('en', { day: '2-digit' }).format(dateObj);
+  return `${day} ${month} ${year}`;
+}
+
+export const MoveTaskOrder = ({ match }) => {
+  const { moveTaskOrderId } = match.params;
```
suzubara, line 28: the value of this parameter is actually `moveOrderId` -- it's named incorrectly in the Route definition. so I think the queries on this page will need to start with that (and will probably end up looking a lot like the data load on the MoveDetails page)

src/pages/Office/MoveTaskOrder/moveTaskOrder.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+  // TODO - Do something with isLoading and isError?
+  const { mtoShipments, mtoServiceItems } = useMoveTaskOrderQueries(moveTaskOrderId);
+
+  const serviceItems = map(mtoServiceItems, (item) => {
+    const detailText = { ZIP: item.pickupPostalCode, Reason: item.reason };
+    /* eslint-disable no-param-reassign */
+    item.serviceItem = item.reServiceName;
+    item.details = { text: detailText, imgURL: '' };
+    /* eslint-enable no-param-reassign */
```
suzubara, line 38: `no-param-reassign` is not a lint rule I think we should ever ignore, as it can have bad side effects. This fn should return a new object instead of modifying `item`.

src/pages/Office/MoveTaskOrder/moveTaskOrder.jsx
```diff
@@ -0,0 +1,80 @@
...
+    return item;
+  });
+
+  return (
+    <div style={{ display: 'flex' }}>
+      <div className="" style={{ width: '85%' }} data-testid="too-shipment-container">
+        {/* eslint-disable-next-line react/prop-types */}
```
suzubara, line 45: Is this eslint comment still needed?

src/pages/Office/MoveTaskOrder/moveTaskOrder.test.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
```
suzubara, line 1: same comment about capitalizing the file name

suzubara:
I think it would be good to get into the habit of adding e2e tests as new pages are added, so maybe we could add one to this PR that is just the TOO navigating to the MTO tab?

monfresh:
Since we no longer have Move Task Orders, should we rename any new instances of `MoveTaskOrder` to `Move`? For example, `src/pages/Office/MoveTaskOrder/moveTaskOrder.jsx` would become `src/pages/Office/Move/move.jsx`

src/shared/Entities/schema.js
```diff
@@ -146,11 +146,11 @@ export const mtoShipment = new schema.Entity('mtoShipments');
 export const mtoShipments = new schema.Array(mtoShipment);
 
 // Move Task Orders
-export const moveTaskOrder = new schema.Entity('moveTaskOrder');
+export const moveTaskOrder = new schema.Entity('moveTaskOrders');
 export const moveTaskOrders = new schema.Array(moveTaskOrder);
 
 // Move Orders
-export const moveOrder = new schema.Entity('moveOrder');
+export const moveOrder = new schema.Entity('moveOrders');
```
donaldthai, line 153: Wanted to keep these consistent with the other entity naming conventions.

cypress/integration/office/txo/tooFlows.js
```diff
@@ -17,6 +17,21 @@ describe('TOO user', () => {
     cy.signInAsUserPostRequest(TOOOfficeUserType, userId);
   });
 
+  it('is able to navigate to move task order page', () => {
```
suzubara, line 20: 🎉

suzubara:
> Hey @donaldthai for the moveOrders query the data object response seems to return keys for the duty station ids as keys and the entitlement object but none of the top level properties. I was really only expecting one moveOrder object is this right?
> 
> Screen Shot 2020-08-17 at 11 35 00 AM

ohhh good catch..! I think that is because of how `moveOrders` is defined in `Entities/schema.js` ... I've definitely been noticing some adjustments may be needed to make normalizr place more nicely with how we are using queries

src/services/ghcApi.js
```diff
@@ -43,6 +43,14 @@ export async function getPaymentRequest(key, paymentRequestID) {
   return makeGHCRequest('paymentRequests.getPaymentRequest', { paymentRequestID });
 }
 
+export async function getMoveOrder(key, moveOrderID) {
+  return makeGHCRequest('moveOrder.getMoveOrder', { moveOrderID }, { schemaKey: 'moveOrders' });
```
suzubara, line 47: I was investigating the issue @duncan-truss brought up and I believe if you change this `schemaKey` to `moveOrder` (singular) it normalizes the response correctly

donaldthai: Works! I went ahead and removed the explicit schemaKey option too which defaults to the singular one.

PR #4599, opened 2020-08-12
[MB-3511] create internalapi update mto shipment endpoint

sojeri:
## Description

Creates the update endpoint we need for the edit forms. :)

## Reviewer Notes

Any thoughts on consequences of using both the updater and status updater? (See [`pkg/handlers/internalapi/mto_shipment.go`](https://github.com/transcom/mymove/pull/4599/files#diff-68badd88373be4ee91ce0d5b2b6b19b5))

Eg, one thing not explicitly called out is we could now technically have a mixed status response if we wanted to implement that-- where we successfully update the mto shipment -- everything but the status. Then we fail on the status update. I went with this approach because I did not want to refactor these services and create potentially more unintended consequences. & I did not want to duplicate the status validation logic, either.

## Setup

I recommend playing with the tests. This command is the one I've been using:

```con
go test ./pkg/handlers/internalapi/ -testify.m Test[a-zA-Z]+MTOShipmentHandler
```

* [x] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/wiki/backend#logging)
* [x] The requirements listed in
 have been satisfied.

## References

- [jira story](https://dp3.atlassian.net/browse/MB-3511)
- [optimistic locking (etag adr)](https://github.com/transcom/mymove/blob/master/docs/adr/0042-optimistic-locking.md)


pkg/handlers/internalapi/api.go
```diff
@@ -128,6 +129,13 @@ func NewInternalAPI(context handlers.HandlerContext) *internalops.MymoveAPI {
 		mtoshipment.NewMTOShipmentCreator(context.DB(), builder, fetcher),
 	}
 
+	mtoServiceItemCreator := mtoserviceitem.NewMTOServiceItemCreator(builder)
```
sojeri, line 132: I've been thinking about this more & am leaning toward removing this & passing in a nil object -- because service member should only be able to have an mto shipment in draft or submitted status. this object is only used for approve/reject status changes.

pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -64,6 +68,103 @@ func (h CreateMTOShipmentHandler) Handle(params mtoshipmentops.CreateMTOShipment
...
+
+// Handle updates the mto shipment
+func (h UpdateMTOShipmentHandler) Handle(params mtoshipmentops.UpdateMTOShipmentParams) middleware.Responder {
+	session, logger := h.SessionAndLoggerFromRequest(params.HTTPRequest)
+
+	if session == nil {
+		return mtoshipmentops.NewUpdateMTOShipmentUnauthorized()
+	}
+
+	if !session.IsMilApp() && session.ServiceMemberID == uuid.Nil {
```
kimallen, line 90: Should we also check that the MTOShipment belongs to the right service member?

pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -64,6 +68,103 @@ func (h CreateMTOShipmentHandler) Handle(params mtoshipmentops.CreateMTOShipment
...
+			}
+			return mtoshipmentops.NewUpdateMTOShipmentInternalServerError().WithPayload(payloads.InternalServerError(nil, h.GetTraceID()))
+		default:
+			return mtoshipmentops.NewUpdateMTOShipmentInternalServerError().WithPayload(payloads.InternalServerError(nil, h.GetTraceID()))
+		}
+	}
+
+	returnPayload := payloads.MTOShipment(updatedMtoShipment)
+
+	if mtoShipment.Status != "" {
```
kimallen, line 130: I think if we are going to handle status in the updater, it should be in a transaction with updating the other fields

pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -64,6 +68,103 @@ func (h CreateMTOShipmentHandler) Handle(params mtoshipmentops.CreateMTOShipment
...
+		eTag := etag.GenerateEtag(updatedMtoShipment.UpdatedAt)
+		return h.handleShipmentStatus(logger, updatedMtoShipment, mtoShipment.Status, eTag)
+	}
+
+	return mtoshipmentops.NewUpdateMTOShipmentOK().WithPayload(returnPayload)
+}
+
+func (h UpdateMTOShipmentHandler) handleShipmentStatus(logger handlers.Logger, shipment *models.MTOShipment, newStatus models.MTOShipmentStatus, eTag string) middleware.Responder {
+	updatedMtoShipment, err := h.mtoShipmentStatusUpdater.UpdateMTOShipmentStatus(
```
kimallen, line 139: I think you could just call `UpdateMTOShipmentStatus` after you call `UpdateMTOShipment` and then handle the errors once and not necessarily have this function.  I'm curious to hear if there's other reasoning though to do it this way.

pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -64,6 +68,103 @@ func (h CreateMTOShipmentHandler) Handle(params mtoshipmentops.CreateMTOShipment
...
+		default:
+			return mtoshipmentops.NewUpdateMTOShipmentInternalServerError().WithPayload(payloads.InternalServerError(nil, h.GetTraceID()))
+		}
+	}
+
+	returnPayload := payloads.MTOShipment(updatedMtoShipment)
+
+	if mtoShipment.Status != "" {
+		eTag := etag.GenerateEtag(updatedMtoShipment.UpdatedAt)
+		return h.handleShipmentStatus(logger, updatedMtoShipment, mtoShipment.Status, eTag)
```
kimallen, line 132: cool that you implemented eTag!  If we don't end up combining this into ghc.yaml (Reggie and Moncef were going to do some research to see what it would take and if it's worth it), then we would want to use an etag for the entire update.  Does this one only apply to the status update?

pkg/handlers/internalapi/api.go
kimallen: Yeah, I don't think the Customer will ever create or update service items.  I think you can remove it.

swagger/internal.yaml
```diff
@@ -2658,6 +2660,7 @@ definitions:
         type: string
       shipmentType:
         $ref: '#/definitions/MTOShipmentType'
+        x-nullable: true
```
kimallen, line 2663: I don't think this should be nullable- it is required

swagger/internal.yaml
```diff
@@ -2670,14 +2673,44 @@ definitions:
         x-nullable: true
       pickupAddress:
         $ref: '#/definitions/Address'
+        x-nullable: true
```
kimallen, line 2676: I see in the model this is nullable (is a pointer), but this is a required field on the form.  We may need to dig a little to see why it's nullable in the model.

swagger/internal.yaml
```diff
@@ -2670,14 +2673,44 @@ definitions:
...
     required:
       - moveTaskOrderID
       - pickupAddress
       - shipmentType
+  UpdateShipment:
+    type: object
+    properties:
+      status:
+        $ref: '#/definitions/MTOShipmentStatus'
+        x-nullable: true
```
kimallen, line 2692: Cannot be null in the db.  But now since you put it nullable, I'm wondering if this is just saying that it can be null in the payload?  Is that how this works?

swagger/internal.yaml
```diff
@@ -2670,14 +2673,44 @@ definitions:
...
+      shipmentType:
+        $ref: '#/definitions/MTOShipmentType'
+        x-nullable: true
+      requestedPickupDate:
+        format: date
+        type: string
+      requestedDeliveryDate:
+        format: date
+        type: string
```
kimallen, line 2701: This is a nullable field

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -248,6 +209,271 @@ func (suite *HandlerSuite) TestCreateMTOShipmentHandler() {
...
+
+		oldShipment := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{})
+		params := suite.getUpdateMTOShipmentParams(oldShipment)
+		params.IfMatch = "intentionally-bad-if-match-header-value"
+
+		response := handler.Handle(params)
+
+		suite.IsType(&mtoshipmentops.UpdateMTOShipmentPreconditionFailed{}, response)
+	})
+
```
kimallen, line 443: There's a 422 response in the yaml- can you make a test for that?  And if that's not a response we handle, you could remove it from the yaml.

swagger/internal.yaml
kimallen: The user should not be able to set the value to null, but should be able to send a payload without this value.  I'm not clear on what adding `x-nullable` does.  Enlighten me if you know.

pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -38,8 +44,6 @@ func (h CreateMTOShipmentHandler) Handle(params mtoshipmentops.CreateMTOShipment
 	}
 
 	mtoShipment := payloads.MTOShipmentModelFromCreate(payload)
-	// TODO: remove this status change once the UpdateMTOShipment api is implemented and can update to Submitted
-	mtoShipment.Status = models.MTOShipmentStatusSubmitted
```
kimallen, line 42: If you remove `SUBMITTED` in this PR, you will need to also update the mtoShipment in the form submission handler on the frontend since the create happens at submission of the form until we split it into two pages.  You could also opt to leave this here, and we can write another story for removing it and updating the status on the form submission to Submitted.

pkg/handlers/internalapi/mto_shipment.go
sojeri: The pattern applies to both, though the first transaction simply uses the etag from the if-match header the client sends. We have to create a new one here because the first transaction will change the updated at time. With the code as is, attempting to reuse the same ifmatch should trigger a stale item error.

swagger/internal.yaml
sojeri: OK, maybe I got overzealous flagging things as nullable :)

swagger/internal.yaml
sojeri: I'll remove it for the next iteration :)

swagger/internal.yaml
sojeri: I'll remove it. You're right, x-nullable is allowing a null value to be set. What we really want is for this to be optional / not required, which is already true.

pkg/handlers/internalapi/mto_shipment.go
sojeri: largely that it felt cleaner to me to have two functions == two transactions

pkg/handlers/internalapi/mto_shipment.go
sojeri: do we have a standard way of checking this? the "closest" way I can find is to do a get to pull the real mto shipment record and then go `realMtoRecord.MoveTaskOrder.MoveOrder.ServiceMemberID` and check against `session.ServiceMemberID`.

pkg/handlers/internalapi/mto_shipment.go
sojeri: I am down to make this change but waiting on clarification from you. In [another comment](https://github.com/transcom/mymove/pull/4599#discussion_r469580216), you said I should include both the UpdateMTOShipment and UpdateMTOShipmentStatus calls in the same Handle method before returning relevant errors. If I use both services, that will maintain the two transaction pattern. Which pattern do you prefer? :)

pkg/handlers/internalapi/mto_shipment.go
sojeri: Wouldn't this be covered by the edit form story that Ron is working on?

pkg/handlers/internalapi/mto_shipment.go
kimallen: Yeah, I guess that's how I'd do it?  I forgot about that kind of check until Andrea mentioned it for another api call.  I don't know of other examples of how to do it.  I see in another handler that the check happens deep in some of the fetches, but don't know what an ideal pattern would be.

pkg/handlers/internalapi/mto_shipment.go
kimallen: No, it would only get covered when the split of the screen happens.  The way it works now is that when someone submits the form, the mto_shipment gets created, and we set the status to SUBMITTED because they have finishing filling out the form.   

When we split the form into 2 pages, the first page will create the shipment and set the status to DRAFT. The 2nd page will be an update, and we will set the status to SUBMITTED.

kimallen: I guess I'd suggest leaving this code for now, and we can make a story for removing it and getting the shipment into the right state through calls to handlers in the places needed.

swagger/internal.yaml
```diff
@@ -2707,10 +2701,8 @@ definitions:
         $ref: '#/definitions/Address'
       destinationAddress:
         $ref: '#/definitions/Address'
-        x-nullable: true
```
kimallen: I believe this one is ok to be x-nullable

swagger/internal.yaml
```diff
@@ -2707,10 +2701,8 @@ definitions:
...
       agents:
         $ref: '#/definitions/MTOAgents'
-        x-nullable: true
```
kimallen: This one can also be null

pkg/handlers/internalapi/mto_shipment.go
sojeri: this came up during our story alignment as well & this change will be reverted for now. I'll drop a ref also to mb 3428 in the comment.

pkg/handlers/internalapi/mto_shipment.go
amitch23: You could also do what the create PPM handler does and fetch the move (which fetches the orders and also checks that the SMID on the session matches the SM on the orders) and if that is successful, you have verification that the user is authorized - https://github.com/transcom/mymove/blob/master/pkg/handlers/internalapi/personally_procured_move.go#L104. Also FYI the MTO -> move PR has gone in, so now the MoveTaskOrder is really just a Move! So we don't need to do the workarounds to fetch the MTO etc.

pkg/handlers/internalapi/mto_shipment.go
kimallen: Resolved with removing UpdateShipmentStatus

pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -64,6 +68,79 @@ func (h CreateMTOShipmentHandler) Handle(params mtoshipmentops.CreateMTOShipment
...
+		logger.Error("Invalid mto shipment: params Body is nil")
+		return mtoshipmentops.NewUpdateMTOShipmentBadRequest().WithPayload(payloads.ClientError(handlers.BadRequestErrMessage,
+			"The MTO Shipment request body cannot be empty.", h.GetTraceID()))
+	}
+
+	mtoShipment := payloads.MTOShipmentModelFromUpdate(payload)
+	mtoShipment.ID = uuid.FromStringOrNil(params.MtoShipmentID.String())
+
+	status := mtoShipment.Status
+	if status != "" && status != models.MTOShipmentStatusDraft && status != models.MTOShipmentStatusSubmitted {
```
kimallen, line 104: Can we also check that the status can only go from DRAFT to SUBMITTED (and not vice versa?).  Here is what it looks like in the status updater:
```
if shipment.Status == models.MTOShipmentStatusDraft && status != models.MTOShipmentStatusSubmitted {
		return nil, ConflictStatusError{id: shipment.ID, transitionFromStatus: shipment.Status, transitionToStatus: models.MTOShipmentStatus(status)}
	}
```

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -248,6 +208,266 @@ func (suite *HandlerSuite) TestCreateMTOShipmentHandler() {
...
+		oldShipment := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{})
+		params := suite.getUpdateMTOShipmentParams(oldShipment)
+		params.MtoShipmentID = *uuidString
+
+		response := handler.Handle(params)
+
+		suite.IsType(&mtoshipmentops.UpdateMTOShipmentNotFound{}, response)
+	})
+
+	suite.T().Run("PATCH failure - 412 -- etag mismatch", func(t *testing.T) {
```
kimallen, line 402: Do we still need this test?

sojeri: we are still using etag, so yes :)

sojeri: otherwise this test should fail 👍

pkg/handlers/internalapi/mto_shipment.go
kimallen: Actually, I guess this above example doesn't test directionality, just that it doesn't skip to another status, which you already have covered.  Probably fine as is.  Disregard!

pkg/handlers/internalapi/mto_shipment.go
sojeri: as I mentioned during standup, I'll leave this authorization check out of this PR. [here's the story](https://dp3.atlassian.net/browse/MB-3698) I created to track this auth work.

pkg/handlers/internalapi/mto_shipment_test.go
kimallen: Oh!  I thought it was removed with the statusUpdater removed- I see now that it's in the mtoShipmentUpdater.  Great!

PR #4612, opened 2020-08-13
[MB-2109] Create MTO address component from DataPair/DataPoint tables

duncan-truss:
## Description

A TOO viewing the MTO page should be able to see the authorized address and the custom address table for each shipment of a move task order.  This PR modifies and reuses the DataPoint and DataPair components to make them suitable for display the address group with the arrow icon.

The only place the DataPoint component was being used was as part of the CustomerRemarksAgentDetails component.

## Reviewer Notes

It seems silly to have DataPoint and DataPair, so I essentially expanded DataPoint to be able to display multiple columns.  DataPair is now DataPointGroup that can have child DataPoints.

I had to remove some global custom style CSS that was limiting the width and causing margin issues with borders and such.

Thinking ahead it seems like we'll need to use these for an editable weight so some of the string props will need to accept nodes instead.

**Storybook Components to Review**

Updated:
[DataPoint ](http://localhost:6006/?path=/story/components-tables--standard-tables)(part of the Tables group) :
[DataPointGroup](http://localhost:6006/?path=/story/components-tables--standard-tables) (part of the Tables group):
[CustomerRemarksAgentDetails](http://localhost:6006/?path=/story/too-tio-components-customerremarksagentsdetails--with-details)
[ShipmentContainer - MTO Accessorial](http://localhost:6006/?path=/story/too-tio-components-shipmentcontainer--mto-accessorial)

## Setup

```
make db_dev_e2e_populate
make server_run
make office_client_run
```

1. Login is the TOO user
2. Find a move from the queue that has an approved MTO
3. Click on the move taking you to the details page
4. Change to the Move Task Order tab
5. View the Authorized Addresses table and Customer Addresses tables
6. Also the previous table with requested/pickup dates should use the same table formatting

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2109) for this change

## Screenshots

Screen Shot 2020-08-19 at 1 56 06 PM


src/components/DataPoint/index.jsx
```diff
@@ -1,28 +1,42 @@
 import React from 'react';
 import propTypes from 'prop-types';
```
suzubara, line 2: nit: we should standardize capitalizing `PropTypes` (the library) so as to distinguish from `component.propTypes` (the property)

src/components/DataPoint/index.jsx
```diff
@@ -1,28 +1,42 @@
...
-  <table className={`table--data-point ${custClass}`}>
+import styles from './index.module.scss';
+
+const DataPoint = ({ columnHeaders, dataRow, Icon, custClass }) => (
+  <table className={classnames(styles.dataPoint, 'table--data-point', custClass)}>
     <thead className="table--small">
       <tr>
-        <th>{header}</th>
+        {columnHeaders.map((header) => (
+          <th key={header}>{header}</th>
```
suzubara, line 12: since `header` is not necessarily a string, I don't think it should be used in the `key` value. Since this is a generic UI component, IMO it would be fine to use `index` for the key.

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -81,6 +90,14 @@ export const MoveTaskOrder = ({ match }) => {
                 requestedPickupDate={formatShipmentDate(mtoShipment.requestedPickupDate)}
                 scheduledPickupDate={formatShipmentDate(mtoShipment.scheduledPickupDate)}
               />
+              <ShipmentAddresses
+                pickupAddress={mtoShipment?.pickupAddress}
+                destinationAddress={mtoShipment?.destinationAddress}
+                // eslint-disable-next-line react/prop-types
```
suzubara, line 96: why disable prop types here? If we can put something in that would be preferable

src/pages/Office/TXOMoveInfo/TXOTab.module.scss
```diff
@@ -6,6 +6,16 @@
...
   flex-grow: 1;
+
+  .gridContainer {
+    .shipmentCard {
+      @include u-padding-left(4);
+    }
+
+    .shipmentCard > div {
+      @include u-margin-bottom(3);
+    }
```
suzubara, line 17: I think these styles should be applied to the `ShipmentContainer` components directly. That way this UI can be reproduced in Storybook instead of being dependent on the context of this page (which seems irrelevant)

src/stories/tables.stories.jsx
```diff
@@ -261,18 +267,40 @@ export const StandardTables = () => (
...
-        <DataPoint header="Receiving agent" body={dataPointBody} custClass="table--data-point--compact" />
+        <DataPoint
+          columnHeaders={['Receiving agent']}
+          dataRow={[dataPointBody]}
+          custClass="table--data-point--compact"
+        />
       </div>
       <div style={{ width: '40px' }} />
       <div>
         <h3>Data pair</h3>
```
suzubara, line 284: Seems like this header should be renamed to `Data point group`

src/components/DataPoint/index.jsx
duncan-truss: Yea i went back and forth about the key, you're right header won't be a string forever.

duncan-truss:
> I gotta be honest, I don't love this component being a `table`... in my mind I think a `dl` would be more appropriate (and also easier to style). But I know this PR is modifying an existing component so if we want to leave it as-is for now I'm ok with that.

Yea let's see where it goes as it becomes more widely used, I didn't love shoehorning the icons into the table and such.  It may give us some nice layout alignment for the time being.

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -23,21 +24,30 @@ function formatShipmentType(shipmentType) {
 
 function formatShipmentDate(shipmentDateString) {
   const dateObj = new Date(shipmentDateString);
+  const weekday = new Intl.DateTimeFormat('en', { weekday: 'long' }).format(dateObj);
   const year = new Intl.DateTimeFormat('en', { year: 'numeric' }).format(dateObj);
   const month = new Intl.DateTimeFormat('en', { month: 'short' }).format(dateObj);
   const day = new Intl.DateTimeFormat('en', { day: '2-digit' }).format(dateObj);
-  return `${day} ${month} ${year}`;
+  return `${weekday}, ${day} ${month} ${year}`;
 }
```
donaldthai, line 32: nit: Function would be cleaner if we used the [moment.js](https://momentjs.com/) library.

duncan-truss:
> The addresses should be top aligned.

@brooksr4 Thanks for pointing that out.  Should be fixed now.

PR #4613, opened 2020-08-13
TIO: Front-end styling for MB-2621

Shkeating:
## Description

- Adds 8px padding between the Accept and Reject buttons.  
- Adds 16px between the Reject button and Clear selection
- Adds 24px padding at the bottom of the service item container. Note: this is > 24 when the `X Clear Selection` button is there because it has a minimum touch target size of 44px, which adds space below it. We also cannot remove the padding from the bottom because the clear selection link isn't always there.
- Adjusts the way the `clear selection` button is implemented, adding an `X-Light` Icon in place of the capitalized letter X that was there before.
- resolves a bug in icon buttons where icons had too much margin/padding on the right and left

## Setup

```sh
yarn install
yarn storybook
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3455?atlOrigin=eyJpIjoiMGU1YmRkNjNhOTk0NGY3OTkzMmVmNmM0MjU3M2NkODgiLCJwIjoiaiJ9) for this change
* [Zeplin](https://zpl.io/2p6QyjJ)

## Screenshots

image


src/components/Office/ReviewServiceItems/ServiceItemCard.jsx
```diff
@@ -155,11 +155,15 @@ const ServiceItemCard = ({
                     <Button
                       type="button"
                       unstyled
+                      icon
```
duncan-truss, line 158: I'm seeing this prop is now deprecated in react-uswds https://github.com/trussworks/react-uswds/blob/0e0df9ce9443949ac56b2986dbe6850b91f48866/src/components/Button/Button.tsx#L55

src/components/Office/ReviewServiceItems/ServiceItemCard.jsx
```diff
@@ -3,16 +3,16 @@ import PropTypes from 'prop-types';
...
-
 import ShipmentContainer from 'components/Office/ShipmentContainer';
 import { mtoShipmentTypeToFriendlyDisplay, toDollarString } from 'shared/formatters';
 import { ShipmentOptionsOneOf } from 'types/shipment';
 import { PAYMENT_SERVICE_ITEM_STATUS } from 'shared/constants';
 import { ReactComponent as XHeavyIcon } from 'shared/icon/x-heavy.svg';
+import { ReactComponent as XLightIcon } from 'shared/icon/x-light.svg';
 import { ReactComponent as FormCheckmarkIcon } from 'shared/icon/form-checkmark.svg';
 
+import styles from './ServiceItemCard.module.scss';
```
suzubara, line 14: I think this violates one of our eslint rules... I'm surprised this didn't surface a lint error when you committed

src/components/Office/ReviewServiceItems/ServiceItemCard.jsx
suzubara: Yeah, I believe `icon` was deprecated because it didn't actually apply any styles in ReactUSWDS. if you want to introduce a `usa-button--icon` class for any custom styling then I think it should happen within MM.

src/components/Office/ShipmentContainer.module.scss
```diff
@@ -1,9 +1,11 @@
...
-    @include u-padding-x(2);
-    @include u-padding-bottom(2);
-    @include u-bg('white');
+  @include u-padding-top(0);
+  @include u-padding-x(2);
+  @include u-padding-bottom(3);
+  @include u-bg('white');
+  @include u-margin-bottom(
+    2
+  ); //this ensures in a vertically compressed window you can still scroll down to the bottom all the way and have some space below the card
```
suzubara, line 10: this is very nitpicky but I wonder if this comment is necessary (it feels like it could set a precedent for commenting all style declarations). I find it can also be helpful to use contextual selectors to add margins, which has the added effect of justifying why there is a margin (i.e., `.shipmentContainer:last-child` or `.shipmentContainer:last-of-type`)

src/components/Office/ReviewServiceItems/ServiceItemCard.jsx
Shkeating: yeah there is something funny going on with my local prettier. thanks for catching that. :/ `yarn lint --fix` sorted it out.

src/components/Office/ShipmentContainer.module.scss
Shkeating: i ended up removing the comment, it's misleading because it's also what puts the 16px space between shipment containers. That's why I didn't put a `last-child` on it in case you look at it and wonder!

src/components/Office/ReviewServiceItems/ServiceItemCard.jsx
```diff
@@ -156,10 +157,13 @@ const ServiceItemCard = ({
                       type="button"
                       unstyled
                       data-testid="clearStatusButton"
-                      className={styles.clearStatus}
+                      className={classnames(styles.clearStatus, styles.icon)}
```
suzubara, line 160: is this meant to be the global `icon` class? I don't see any `.icon` defined in the CSS module for this file.

src/components/Office/ShipmentContainer.module.scss
```diff
@@ -1,9 +1,9 @@
...
 .shipmentContainer {
-    @include u-padding-top(0);
-    @include u-padding-x(2);
-    @include u-padding-bottom(2);
-    @include u-bg('white');
+  @include u-padding-top(0);
+  @include u-padding-x(2);
+  @include u-padding-bottom(3);
+  @include u-bg('white');
+  @include u-margin-bottom(2);
```
suzubara, line 8: _usually_ I find it's useful to style margins contextually (for example, in the parent element styles rather than the child). I think this addition is leading to [this diff](https://github.com/transcom/mymove/pull/4613/files?short_path=18a2f60#diff-18a2f60517a0687f4d04d5f185d85e92) which I wouldn't have expected (the gap between service item cards is not evident in the designs)

PR #4614, opened 2020-08-13
MB-3543 integrate get mtoShipment endpoint with MyMove app

amitch23:
## Description

Integrates get MTOShipment endpoint with app (landing page, HHG details form page and Review Summary page).

Landing page now loads MTOShipments for a move_task_order_Id returned on the orders from the showLoggedInUser endpoint. This enables the "continue move setup" to route to the hhg-start page if there is a move with move type "HHG", and the Review page is the HHG exists and is complete. Previously the user would be routed to the hhg-start page even if there was an existing HHG.

HHG Details form page now loads the first mtoShipment for a move and displays the fields in the form, e.g. if a user hits "back" from the Review page after just creating an HHG. 
* "Use current residence" should toggle appropriately with the existing HHG pickupAddress and the currentResidenceAddress from the profile, if they are different. 
* Form does not hit the update endpoint because the updateMtoShipment service doesn't do what we need now, i.e. update the fields on the shipment. Left a TODO to hookup when ready and bypasses update. See: https://ustcdp3.slack.com/archives/CTNTFJSBA/p1597710038361500

Review/Summary page now gets the mtoShipments for a move that is returned from the `loadMove` payload, just like the PPMs are being returned. This is better than using the listShipments endpoint because it's one less API call! Summary now includes the HHG fields if they exist.

Refresh to your heart's content! 

Still needs:

- ~~merge from master to incorporate move consolidation work for Review page~~
- ~~update MTOShipment endpoint on HHG details form page for full functionality~~ - Added TODO to hookup when update service for SM ready
- ~~Fix routing logic in getworkflowroutes to route to appropriate pages with existence of HHG vs. PPM~~

## Reviewer Notes

## Setup

Run through the flow. Test the following use cases from the Landing page by clicking on "Continue Move Setup"

 * with Shipment type set on Move but no MTOShipment, goes to hhg-start 
 * With Shipment or PPM complete, goes to review page
  * with PPM type, goes to ppm-start 

Refresh on each page and ensure mtoShipment loads in the entities.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3543) for this change


src/scenes/Landing/index.jsx
```diff
@@ -54,8 +61,12 @@ export class Landing extends Component {
       }
     }
     if (prevProps.move && prevProps.move.id !== this.props.move.id) {
+      this.props.loadMTOShipments(this.props.moveTaskOrderID);
       this.props.loadPPMs(this.props.move.id);
     }
+    // if (prevProps.moveTaskOrderID !== this.props.moveTaskOrderID) {
```
amitch23, line 67: remove this

src/scenes/Landing/index.jsx
```diff
@@ -188,10 +199,15 @@ const mapStateToProps = (state) => {
   const user = selectCurrentUser(state);
   const serviceMember = get(state, 'serviceMember.currentServiceMember');
   const move = selectActiveOrLatestMove(state);
-  // TODO: use move_id from active or latest move instead of this once db reconciliation work done
-  const moveTaskOrderID = get(selectLoggedInUser(state), 'service_member.orders[0].move_task_order_id', '');
+  const orders = selectActiveOrLatestOrdersFromEntities(state);
+  const MTOID = get(orders, 'move_task_order_id', '');
+
+  // TODO: Use the move ID, not the MTOID, once moves have been consolidated to 1 table
```
amitch23, line 205: remove

pkg/models/move.go
```diff
@@ -196,6 +196,9 @@ func (m *Move) Cancel(reason string) error {
 func FetchMove(db *pop.Connection, session *auth.Session, id uuid.UUID) (*Move, error) {
 	var move Move
 	err := db.Q().Eager("PersonallyProcuredMoves.Advance",
+		"MTOShipments.MTOAgents.Email",
```
kimallen, line 199: I'm curious why only the Email is fetched, and not the other agent info?

src/scenes/MyMove/getWorkflowRoutes.jsx
```diff
@@ -220,20 +220,28 @@ const pages = {
   '/moves/:moveId/hhg-start': {
     isInFlow: (state) => inHhgFlow && state.selectedMoveType === SHIPMENT_OPTIONS.HHG,
     isComplete: ({ sm, orders, move, ppm, mtoShipment }) => {
-      return mtoShipment && every([mtoShipment.requestedPickupDate, mtoShipment.requestedDeliveryDate]);
+      return (
+        mtoShipment &&
+        every([
+          mtoShipment.requestedPickupDate,
+          mtoShipment.requestedDeliveryDate,
```
kimallen, line 227: If they do not have a requestedDeliveryDate, will this show up as not complete?  They should be able to consider this page completed even if it doesn't have a value.  I haven't tested the UI to see how this works out.

src/scenes/Landing/index.jsx
```diff
@@ -188,10 +199,15 @@ const mapStateToProps = (state) => {
   const user = selectCurrentUser(state);
   const serviceMember = get(state, 'serviceMember.currentServiceMember');
   const move = selectActiveOrLatestMove(state);
-  // TODO: use move_id from active or latest move instead of this once db reconciliation work done
-  const moveTaskOrderID = get(selectLoggedInUser(state), 'service_member.orders[0].move_task_order_id', '');
+  const orders = selectActiveOrLatestOrdersFromEntities(state);
+  const MTOID = get(orders, 'move_task_order_id', '');
```
kimallen, line 203: It's a little confusing that we are still getting a `move_task_order_id` here, but it's actually the `move_id`, which we are using in other places.  Would it make sense to change the return payload for Orders to name this `move_id` instead of `move_task_order_id`?  Or would that screw up a bunch of other stuff?

src/scenes/Review/Summary.jsx
```diff
@@ -140,15 +138,14 @@ Summary.propTypes = {
 };
 
 function mapStateToProps(state, ownProps) {
-  const moveID = state.moves.currentMove.id;
+  const moveID = ownProps.match.params.moveId;
```
kimallen, line 141: yay! (for moving away from `moves.currentMove`)

src/shared/Entities/modules/mtoShipments.js
```diff
@@ -38,6 +38,31 @@ export function createMTOShipment(mtoShipment, label = createMTOShipmentOperatio
   return swaggerRequest(getClient, createMTOShipmentOperation, { body: mtoShipment }, { label, schemaKey });
 }
 
+const updateMTOShipmentOperation = 'mtoShipment.updateMTOShipment';
```
kimallen, line 41: I really like that you put `Operation` in the variable name- I always forget what it refers to and have to go search for it since most other versions use `Label`.

pkg/models/move.go
amitch23: My understanding is that for Pop to fetch all the fields of a nested objects, we only need to specify one field at the  layer which we want to eagerly load. So this could have been '"MTOShipments.MTOAgents.CustomerRemarks"' or another field. I *think*

src/scenes/Landing/index.jsx
amitch23: You're right, we don't need this anymore and can use the move id from the move here ` const move = selectActiveOrLatestMove(state);`. I think it makes sense to revert the workaround for fetching the mto_id since it's now unnecessary. I'll take out that extra fetch mto that Ron added with the HHGDetailsForm PR and just use the move ID instead.

amitch23: (i.e. the changes that were added here: https://github.com/transcom/mymove/pull/4463/files)

amitch23:
> When I go through creating a shipment and get to the review page, click "Milmove" to go to the home page, and then choose `Continue Move Setup`, it is taking me to the `hhg-start`. I expected it to take me to the Review page. Can you double check that?

This was due to a typo on capitalizing fields on the mtoShipment that shouldn't have been in the "isComplete" of the hhg-start page. Oopsies!!

pkg/models/move.go
reggieriser: I don't think you want to be eagerly fetching individual fields that aren't themselves pointers to something else.  I believe eagerly fetching `MTOShipments.MTOAgents` is going to be no different than `MTOShipments.MTOAgents.Email` (but double check the `MTOAgents` struct after this `Find` call and make sure).  My concern with adding the individual field (even if it works) is that may give the impression that one must fetch each individual desired field.

amitch23: @reggieriser thanks for clarifying! :) I changed to now fetch the reference (i.e. Agents), which loads all the fields at that level.

src/components/Customer/HHGDetailsForm.jsx
```diff
@@ -50,112 +53,177 @@ const AgentSchema = Yup.object().shape({
...
-              ...initialValues,
-              ...currentValues,
-              pickupLocation: {
-                mailingAddress1: '',
-                mailingAddress2: '',
-                city: '',
-                state: '',
-                zip: '',
+          // eslint-disable-next-line no-lonely-if
+          if (match.params.moveId === initialValues.moveTaskOrderID) {
```
sojeri, line 150: adding comment for ease of linking to this line.

PR #4616, opened 2020-08-14
[MB-3596] Create and style Edit Shipment page (mobile view)

Ronolibert:
## Description

Add the EditShipment route and style mobile view

## Reviewer Notes
- I made a Fieldset component instead of using the react-uswds one cause for now it doesn't give the control I needed for the styling like adding a hint and being able to add styling to the container and the legend itself. I can look into updating the library itself in the near future
- this doesn't include any functionality, just adds the route/page and the mobile view styling
- route will be different once we add edit button to be able to navigate using the shipment's ID instead
- I know there was some discussion about how to use existing styles/variables in the customer side of things but I didn't want to just import every single shared stylesheet for the entire customer app incase those talks are still ongoing. I just imported the individual stylesheets I needed for the variables and copied some styling like `usa-button--unstyled` from a different stylesheet til a decision is made

## Setup
`make server_run`
`make client_run`
1. Log in to any account
2. Visit route `/moves/review/edit-shipment`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3596) for this change

## Screenshots

download-1


Ronolibert:
@mariaclaratgolfo Thanks!

I think I took care of most of the changes. I am using the storybook components but for some reason storybook has bold labels for all the inputs. Do we want them to be bold?

The remarks is the same style/size as all the other corresponding headers

I also changed it so we have the `Optional` next to the header instead on every input.

I already made a card for the desktop view for later when we finalize the layout!

src/components/Customer/EditShipment.scss
```diff
@@ -0,0 +1,54 @@
+@import 'shared/styles/_variables';
+@import 'shared/styles/colors';
+
```
kimallen, line 3: We should not create any new global styles unless it's intentional.  This should be an `EditShipment.module.scss` file.  See https://github.com/transcom/mymove/wiki/frontend#using-sass-and-css-modules.  This will ensure that the styles in here are localized.

Ronolibert: Some of it is intentional! Where should I put some shared styles for the customer app?

Ronolibert: Actually, maybe not for now since it's only being used in one place. Converted them to use CSS modules! Good catch

kimallen: If they will be used across the entire app, looks like `src/shared/styles`.  If it's only for the customer app, I'd create a `customer.module.scss` If it's going to go in multiple children of a parent component, create it for that parent component as a `module.scss`.  I'm not totally versed on the intentions of the latest on where styling goes, so probably @Shkeating or @suzubara could give a more definitive answer.

Ronolibert: Converted the styles to be used through css modules

src/shared/Hint/index.module.scss
```diff
@@ -0,0 +1,10 @@
+@import 'shared/styles/_variables';
+@import 'shared/styles/colors';
+
+@media (max-width: $mobile) {
+  .usa-hint {
```
kimallen, line 5: This looks like a USWDS class, but it's not since it's a css module, which appends a unique value. I'm thinking we should just call this `.hint`

src/components/Customer/EditShipment.jsx
```diff
@@ -0,0 +1,349 @@
...
+                    <li>Weapons or alcohol</li>
+                  </ul>
+                </div>
+
+                <Textarea
+                  label="Anything else you would like us to know?"
+                  data-testid="remarks"
+                  name="remarks"
+                  className={`${styles.remarks}`}
+                  placeholder="This is 500 characters of customer remarks placeholder"
```
kimallen, line 311: Do we know yet what we actually want this placeholder to say?

From the abstract, looks like design wants it to say '500 characters' just below the textarea box.

Ronolibert: Don't think we do yet, the design had some lorem ipsum in it. The 500 characters thing I think we're doing when we're working on having the form functional since its counting the number of letters and updating as they type

Ronolibert:
@mariaclaratgolfo I added the bold labels to the fields. I'm not so sure about the date inputs since we're using the same ones the TXO uses

@kimallen I just added bold to the inputs like it has in the storybooks. I didn't import the exact files the storybook has because it conflicts with some existing styling in the customer app so I just added the bold labels to the fields

kimallen:
Would be awesome if you made an update to the react/uswds library so that the Fieldset can use the adaptations you created in your custom component.

Ronolibert:
> Would be awesome if you made an update to the react/uswds library so that the Fieldset can use the adaptations you created in your custom component.

Yup! I think once we get the designs solidified I'll put up a PR to the original uswds lib with the extra support for customization

PR #4621, opened 2020-08-14
MB-3540 Remove references to MTOs in tests

monfresh:
Now that the Moves/MTOs consolidation is done, we can start some of
the cleanup work. In this PR, we are replacing references to
`MoveTaskOrder` with `Move`, but only in the testdatagen functions and
in tests that call those functions. Note that there are still
references to `MoveTaskOrder` within models. Those will be cleaned up
separately.

## Setup

```sh
make clean build
```

```sh
make server_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3540) for this change


migrations/app/migrations_manifest.txt
```diff
@@ -539,4 +539,4 @@
 20200807222709_update_mto_service_items_columns.up.sql
 20200811152728_move_task_orders_to_moves.up.sql
 20200812180333_add_contract_code_param_to_dpk.up.sql
-20200812202513_convert_schedule_params_to_integer.up.sql
\ No newline at end of file
+20200812202513_convert_schedule_params_to_integer.up.sql
```
monfresh, line 542: The precommit hook detected there was a missing new line at the end of the file, which is why this change is here.

pkg/testdatagen/make_move.go
```diff
@@ -18,14 +18,14 @@ func MakeMove(db *pop.Connection, assertions Assertions) models.Move {
...
 	var referenceID string
 	if assertedReferenceID == nil || *assertedReferenceID == "" {
 		referenceID, _ = models.GenerateReferenceID(db)
 	}
 
 	var contractorID uuid.UUID
-	mtoContractorID := assertions.MoveTaskOrder.ContractorID
+	mtoContractorID := assertions.Move.ContractorID
 	moveContractorID := assertions.Move.ContractorID
 	if mtoContractorID == nil || moveContractorID == nil {
```
reggieriser, line 30: After your change, `mtoContractorID` and `moveContractorID` point to the same thing, so we should be able to get rid of one of them now.  Another case like this in the `MakeMoveWithoutMoveType` function too.

migrations/app/migrations_manifest.txt
reggieriser: FYI, I checked with the engineer where the missing new line originated to see how it made it passed their pre-commit checks.  They think it may have happened when they used the GitHub UI to do a merge (which I assume bypassed our pre-commit checks).

pkg/testdatagen/make_move.go
monfresh: Good catch! I'll fix it.

PR #4623, opened 2020-08-14
MB-3685 | misc small cypress improvements :)

sojeri:
## Description

Since upgrading to 4.12.1 broke some of the tests, I'm redoing this PR as follows:
- add `yarn.lock` to cypress folder
- add cypress folder caching to circleCI cache flow
  - this has minimal speed gains, but any gain is a win at this point
- add cypress dep to root dev dependencies
  - this allows for root level opening of cypress test runner via yarn, which IMO is a huge win
  - it's much faster to use the test runner against already running docker pool rather than do a clean test/rebuild
    - __for local dev, unblocking fast test iteration > having a clean test environment for every cypress test run__
- add some dangerfile checks to ensure we're keeping cypress deps in sync
- update instructions in README around running e2e tests
  - call out more explicitly why `run_e2e_tests` may behave unexpectedly

## Reviewer Notes

RE this PR's original intent: __I will come back to the cypress upgrade in a trailing PR Thu or Fri!__

## Setup

Play around with the cypress test runner if you want. :)

Barry--- I added you as a token infra person. Any objections to my circleCI change? If you're not the right person, please help me add the right infra reviewer.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3685) for this change
* ~~[this cypress issue](https://github.com/cypress-io/cypress/issues/7930) contains more examples of the test slowdowns in versions around 4.8.0 (there are many similar issues).~~


esacteksab:
Thanks for the clarification on the add. Nothing sticks out to me, but I added the InfraSec team, as admittedly all of them have more experience with CircleCI than I do so maybe one of them has some thoughts. I think the only thing that stood out was that CircleCI has a Node [Orb](https://circleci.com/orbs/registry/orb/circleci/node). I don't know if we want to go all-in on Orbs or not, but I know on the Infra side of things, where they exist, we try and use them. YMMV.

cblkwell:
> Thanks for the clarification on the add. Nothing sticks out to me, but I added the InfraSec team, as admittedly all of them have more experience with CircleCI than I do so maybe one of them has some thoughts. I think the only thing that stood out was that CircleCI has a Node [Orb](https://circleci.com/orbs/registry/orb/circleci/node). I don't know if we want to go all-in on Orbs or not, but I know on the Infra side of things, where they exist, we try and use them. YMMV.

This seems okay to me -- I can't weigh in on the node specifics, but the CircleCI workflow seems fine.

package.json
```diff
@@ -89,13 +89,15 @@
...
     "test:coverage": "rescripts test --coverage --env=jsdom",
+    "test:e2e": "cypress open --config baseUrl=http://milmovelocal:3000",
     "prettier": "prettier --loglevel warn 'src/**/*.{js,jsx}'",
     "lint": "eslint --ext .js,.jsx src",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
     "spellcheck": "mdspell --ignore-numbers --ignore-acronyms --en-us --no-suggestions"
   },
   "version": "0.1.0",
+  "NOTE-RE-CYPRESS": "if you upgrade cypress here, upgrade in the /cypress package.json as well",
```
suzubara, line 100: I don't love using the `package.json` to document this... IMO depending on the intention, it should be in the docs/wiki, and enforced via something like the dangerfile.

sojeri: that's a good point. I'll look into how danger can enforce this.

cypress/support/constants.js
```diff
@@ -2,7 +2,7 @@ export const longPageLoadTimeout = 10000;
 export const fileUploadTimeout = 10000;
 
 // Base URLs
-const cypressClientPort = 4000; //change this to 3000 to run cypress against dev instance
+let cypressClientPort = Cypress.env('testClientPort') || 4000; //change this to 3000 to run cypress against dev instance
```
sojeri, line 5: @suzubara this + updating the incantation in package.json is how I fixed the port issue. I confirmed also that running it without the env option still defaults to 4000 as expected, but feel free to play with it more yourself to confirm! :)

PR #4628, opened 2020-08-17
MB-3556 Keep testdatagen functions organized

monfresh:
## Description

During our work to consolidate Moves and MTOs, we noticed this
`MakeAvailableMoveTaskOrder` function that fit better inside the
`testdatagen` package. This PR moves that function into the
`make_move.go` file and renames it to `MakeAvailableMove` since we
no longer have a Move Task Orders table.

### DIFF
--- a/pkg/testdatagen/make_move.go
+++ b/pkg/testdatagen/make_move.go
@@ -1,6 +1,8 @@
 package testdatagen
 
 import (
+	"time"
+
 	"github.com/go-openapi/swag"
 	"github.com/gobuffalo/pop"
 	"github.com/gofrs/uuid"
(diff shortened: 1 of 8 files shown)


pkg/testdatagen/make_move.go
```diff
@@ -104,6 +106,18 @@ func MakeMoveWithoutMoveType(db *pop.Connection, assertions Assertions) models.M
 	return move
 }
 
+// MakeAvailableMove makes a Move that is available to the prime at
+// the time of its creation
+func MakeAvailableMove(db *pop.Connection) models.Move {
+	now := time.Now()
+	mto := MakeMove(db, Assertions{
```
reggieriser, line 113: Nit: this should probably change from `mto` to `move` now.  Don't see any other `mto` references in this testdatagen maker.

reggieriser:
I just noticed while working on another PR that we have quite a few existing `MakeMove` calls that just set the `AvailableToPrimeAt` only, so they can probably be remapped to this function and simplify the code.

pkg/testdatagen/make_move.go
monfresh: Done.

monfresh:
@reggieriser I replaced the other instances you mentioned. I think I got everything.

reggieriser:
> @reggieriser I replaced the other instances you mentioned. I think I got everything.

Looks great!  Thanks for making the changes.

PR #4629, opened 2020-08-17
MB-3667 Create order with all fields at same time

monfresh:
## Description

During onboarding, we store the service member's rank and origin duty
station on the `service_members` table. However, on the TOO page,
those fields are assumed to come from the `Orders` table. Since those
fields are expected to be on the `Orders` table, they should be
populated at the same time as the order is first created. This was not
the case. The way those fields were added to the Order was via a
temporary hack.

During the moves and MTOs consolidation, we were able to remove one of
the temporary conversion functions. In this PR, we are able to
remove the remaining conversion-related code entirely by adding those
missing order fields to the order creation in the orders handler.

## Reviewer Notes

1. Go through onboarding to create a move and order
2. Sign in to the office app as a TOO
3. Click on the move you just created
4. Verify that the TOO can see the service member's grade, origin duty station, and weight allowances

## Setup

```sh
```


monfresh:
@kimallen This is a temporary solution to get rid of the converter. As for what's next, we decided we will keep this data in Orders only. As you noted, this will require coordination with Team Pamplemoose because it will require redesigning the onboarding flow. I will start a branch with that work and hand it off to your team once you are ready to pick it up.

monfresh:
The update issue is existing behavior, right? In this PR, we are aiming not to change existing behavior or fix any existing issues.

config/env/prod.app.env
```diff
@@ -12,7 +12,6 @@ DOD_CA_PACKAGE=/config/tls/Certificates_PKCS7_v5.6_DoD.der.p7b
 DOMAIN=move.mil
 EMAIL_BACKEND=ses
 FEATURE_FLAG_ACCESS_CODE=true
-FEATURE_FLAG_CONVERT_PROFILE_ORDERS_TO_GHC=false
```
reggieriser, line 15: I also see one of these in a `prd.app.env` (`prd` not `prod`) file.  Maybe that's a new GovCloud addition?

pkg/models/service_member.go
```diff
@@ -226,7 +226,12 @@ func (s ServiceMember) CreateOrder(db *pop.Connection,
 	ordersNumber *string,
 	tac *string,
 	sac *string,
-	departmentIndicator *string) (Order, *validate.Errors, error) {
+	departmentIndicator *string,
+	originDutyStation *DutyStation,
+	originDutyStationID *uuid.UUID,
+	grade *string,
+	entitlement *Entitlement,
+	entitlementID *uuid.UUID) (Order, *validate.Errors, error) {
```
reggieriser, line 234: Wouldn't the `originDutyStationID` and `entitlementID` be redundant to pass in here if we're already passing in an `originDutyStation` and `entitlement` (both of which should be carrying their respective `ID` fields)?

reggieriser (commented):
I noted a couple of small things, but generally this looks like it's doing the same thing the converter was doing in environments where this feature flag was turned on.  It looks like the flag/converter is off in prod, though -- will there be any issues with these fields being populated now on a production PPM moves if it occurs?  I doubt it, but just want to make sure we've thought that through.

config/env/prod.app.env
monfresh: Yeah, looks like it was added one day after I opened this PR.

pkg/models/service_member.go
monfresh: Good point. I fixed it.

config/env/prod.app.env
monfresh: This is fixed now.

monfresh:
@kim @lynzt Can you please give this one last look? Reggie made a good observation that the feature flag was not turned on in production. Was that for a reason? Will it cause issues if orders in production now automatically get these additional fields (entitlement, grade, origin duty station)?

kimallen (approved):

PR #4640, opened 2020-08-19
[MB-3842] Fix EIA fuel price param

garrettqmartin8:
## Description

Based on [this conversation](https://ustcdp3.slack.com/archives/CP6F568DC/p1597341388030100?thread_ts=1597338132.019800&cid=CP6F568DC), I've changed the EIA fuel price param to be an `INTEGER`, not a `DECIMAL`.


pearl-truss:
Looks good! I confirmed in psql-dev that EIAFuelPrice is now set as integer in the service_item_param_keys table

PR #4644, opened 2020-08-19
Add ADR to move service_member fields to orders

monfresh:


docs/adr/0056-move-service-members-fields-to-orders.md
```diff
@@ -0,0 +1,37 @@
...
+
+* `+` Does not require redesigning the onboarding flow
+* `-` Results in data integrity issues
+* `-` Increases code complexity when writing to two tables and worrying about how and when to keep them in sync
+
+### Remove the fields from `service_members` and only store them in `orders`
+
+* `+` Eliminates data integrity issues
+* `+` Simplifies code because we are only writing to one table
+* `-` Requires a redesign of the onboarding flow so that rank and origin duty station are captured during the order portion.
```
gidjin, line 37: Does it make sense to possibly leave the flow the same but create an orders record that only contains origin duty station and rank at that time until the full orders detail is captured later?

kimallen, line 37: Another - is that everywhere the rank and current station(service member) are currently utilized in both the frontend and backend need to be modified

docs/adr/0056-move-service-members-fields-to-orders.md
```diff
@@ -0,0 +1,37 @@
...
+
+The problem with storing that data in two tables is that we have to worry about keeping them in sync. For example, if a service member made a mistake when entering their origin duty station, they can update it (which will update it in the `service_members` table), but if the order was already created, it needs to be updated as well. We don't currently handle this scenario in the code.
+
+To resolve data integrity issues, we propose removing those fields from the `service_members` table and storing them on the `orders` table only.
+
+## Considered Alternatives
+
+* Keep the fields in both `service_members` and `orders`
+* Remove the fields from `service_members` and only store them in `orders`
+
```
kimallen, line 20: Was it considered to keep the fields only on the `service_members`, and have the TOO pull those values from the SM instead of the `orders` table?  That would require the least re-working, though not have the historical reference on the `orders`, though there was mention that the historical view would be available in the data warehouse.

docs/adr/0056-move-service-members-fields-to-orders.md
kimallen: There are currently required fields on the orders that would be needed to create an orders record (like issue date and report-by date, and orders type)

docs/adr/0056-move-service-members-fields-to-orders.md
monfresh: Wouldn't this cause data integrity issues if the service member updates their info?

kimallen: Not if we only save the data on the service member.  This option would basically be keeping the original setup for those fields in the db.  Not save those values on the orders.

monfresh: Would the following scenario ever happen? 

A customer populates their origin duty station and grade, creates a move, then the TOO uses this information to approve the move. Then, before the Prime sees it, the customer updates their info, and because the info is only stored in service members, the Prime is now seeing incorrect information.

monfresh: Or the service member changes the info while the TOO is still looking at it. Wouldn't it confuse the TOO to see changing information?

gidjin: Can the customer make changes after submitting the move for approval by the TOO? This might be a good question for someone on the @transcom/truss-design team to answer.

monfresh: I asked in #prac-design and the desire is to prevent the customer from making any changes that could affect pricing/MTO once they submit the move. So it sounds like the only outstanding question is about preserving the historical reference on the orders. I will ask about the data warehouse. Is there anything else we aren't thinking about in terms of storing the data on `service_members` only?

Ryan-Koch: I think the question of what happens after a move is closed out is a bit unclear (at least to me) as far as how it appears in our system. One side effect of a choice where we store this stuff on `service_members` is that if we choose to continue to show closed out items in the office app, that those items would show different information if that same person were to open a new move later. Another scenario that comes to mind is that if we're storing this information after it closes out, even if we aren't going to display it anymore in the interface, there's a non-zero chance the record could be requested for some sort of auditing situation. Even if we have audited changes and what not, they could still ask for the record. If that's a situation we need to be ready for, then the information changing could be problematic from that perspective.

monfresh: @kim @gidjin I think Ryan brings up a good point. I think it is safer to store the data in Orders due to the scenarios Ryan describes, and potentially other unforeseen ones. Thoughts?

gidjin: Having it on orders does seem better given some of the scenarios above. I don't know if that affects where rank or other data is stored. @kimallen did mention the Data Warehouse and I don't know enough about that to know what info goes there, but it sounds like it might allow for historical data. But I'm not familiar with the onboarding flow either so I'll defer to Kim and Ryan

monfresh:
@lynzt @tinyels Could I please get your thoughts on this? I think we've gotten all the feedback, and now we need someone to make a final decision please.

docs/adr/0056-move-service-members-fields-to-orders.md
tinyels: I would like to hear how design thinks the user conceptualized these values. Do they belong to the service member or to the orders?

lynzt: I'm in favor of storing the data in the Orders table.
I also can't think of a reason we'd need the service member's information without tying it to an order, but 🤷‍♀️

Ryan-Koch (approved):

docs/adr/0056-move-service-members-fields-to-orders.md
samadden: In term of how design thinks about this, we expect the customer to think of their Rank and Current duty station in their "Profile" section -- not the Orders section. They enter that information early in their flow, since they know these details without having to refer to the document. Later, they look to their Orders doc to indicate where/when they will move to next. So the "Profile" (service member) set of info is who/what/where they are now, whereas the "Orders" set of info is where/when they are being ordered to go next.

That said, we set it up this way to make it an easy, step-by-step flow for the service member at this stage -- the Orders documents do specify 1) rank/grade, 2) current duty station, and 3) new duty station. This is how the info gets reflected in the TOO interface: the Orders table shows the duty stations (current and new), and the Allowances table shows the Branch/rank; all of these fields are checked for accuracy against the Orders documents. (As Moncef pointed out above, these fields are editable by the TOO, not the customer.)

My opinion, lightly held, is that it makes sense to hold this info in the Orders table, since it reflects the info captured on the Orders documents.

tinyels: Is the orders object created when the service member is? If not, then we would need to do some work to allow it to exist  at that point. It also raises interesting questions when a user returns for moves in the future--I imagine we will need them to update this information then right?

tinyels (approved):
Approved with the caveat that @lynzt needs to confirm that her team can handle the impact of the change.

esacteksab:
Last comment is quite a while ago. Where's this at?

monfresh:
If I recall, during the last meeting we had with Team Pamplemoose, we agreed to create an [epic in JIRA](https://dp3.atlassian.net/browse/MB-4663) to track current issues related to updating service member data, based on service members only having one order ever. Once we start expecting service members in production having more than one order, then we will need to revisit this and write some more stories. I can close this for now and reopen later.

PR #4646, opened 2020-08-19
Replaced deprecated UUID library references

reggieriser:
## Description

I noticed a few references to `github.com/gobuffalo/uuid` that crept back into the code.  That library is [deprecated](https://github.com/gobuffalo/uuid) and we should use `github.com/gofrs/uuid` instead.

## Setup

`make server_test`


reggieriser:
> I wonder if we should add a danger js check to flag this automatically going forward since it somehow gets added back in by accident

That would be a great idea as this has happened a few times over the last year or so.  I haven't set up checks with danger yet, but I [made a story](https://dp3.atlassian.net/browse/MB-3719) for it.

PR #4653, opened 2020-08-20
MB-3583: Add pipeline for prd environment

cblkwell:
## Description

This adds a pipeline so that we can deploy to the prd environment; for now, I've made it so it has a placeholder branch with notes on how this will need to be updated when we're ready to make it a "real" pipeline.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1836,21 +1836,21 @@ workflows:
             - build_app # replace with approve_prod_deploy
           filters:
             branches:
-              only: DEPLOY_PRD_PLACEHOLDER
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1308,7 +1369,7 @@ jobs:
           health_check_hosts: gex.stg.move.mil,dps.stg.move.mil,orders.stg.move.mil
           ecr_env: stg
 
-  # `deploy_prod_migrations` deploys migrations to the staging environment
+  # `deploy_prod_migrations` deploys migrations to the prod environment
```
esacteksab, line 1372: <3

PR #4656, opened 2020-08-20
Kim mb 3230 refactor load transportation office

kimallen:
## Description

Creates a swaggerRequest for loadDutyStationTransportationOffice.
Removes old fn and entire ducks file and api file associated with it.

## Reviewer Notes

The call was being used in two places: the landing page (`Landing/index.jsx`) and `PaymentRequest.jsx`.  I removed the call altogether from `PaymentRequest.jsx` since we will always have access to the data retrieved from `getLoggedInUser`, which loads when the user enters the app, and upon any refresh.  At the point in the flow that the customer reaches the payment request page, they cannot edit any fields that would impact the transportation office info, so there is no need to load the office info separately.

## Setup

-Log in as a service member that is past the profile stage.  You should see the transportation office info on the right of the home page.

- Log in as a user that is requesting payment ( you can use`ppm@continue.requestingpayment`).  
- Go to the weight ticket page
- Choose a weight ticket type
- Click the checkbox for Full Weight "I'm missing this weight ticket"
- See the transportation office name and phone number listed.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3230) for this change


src/shared/Entities/modules/transportationOffices.js
```diff
@@ -0,0 +1,35 @@
+import { swaggerRequest } from 'shared/Swagger/request';
+import { getClient } from 'shared/Swagger/api';
+import { get, isNil } from 'lodash';
+
+export const loadDutyStationTransportationOfficeOperation = 'TransportationOffice.loadDutyStationTransportationOffice';
```
amitch23, line 5: I don't think this variable name is long enough, any way we can add a couple words? (;))

src/shared/Entities/modules/transportationOffices.js
```diff
@@ -0,0 +1,35 @@
...
+const dutyStationTransportationOfficeSchemaKey = 'transportationOffice';
+
+export function loadDutyStationTransportationOffice(
+  dutyStationId,
+  label = loadDutyStationTransportationOfficeOperation,
+  schemaKey = dutyStationTransportationOfficeSchemaKey,
+) {
+  console.log('inside load');
```
amitch23, line 13: reminder to remove the console.log

PR #4657, opened 2020-08-20
MB-3725 Add adr for using asdf to manage node, yarn, and golang development installs

gidjin:
## Description

It would be good to have a better way to manage golang versions locally than relying on `brew pin go` See the new adr `docs/adr/0056-use-asdf-to-manage-node-yarn-and-golang-versions-in-development.md`

## Reviewer Notes

This ADR also proposes to supersede 0046 on using nodenv for node management and combine it into the same tool [asdf](https://asdf-vm.com/) which can also manage yarn versions too.

This PR just adds the ADR, switching over is another story, if this adr is accepted.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3725) for this change
* [this slack thread](https://trussworks.slack.com/archives/CDJ9JAD26/p1597854026007300)
* [this slack thread](https://trussworks.slack.com/archives/C016LRGTZUJ/p1597769146030700)


gidjin:
For the discussion, I have this working locally with all three. But felt it was too much to add directions and make all the changes for the switch as part this PR. So if you'd like to see it in use let me know and I can pair with you.

gidjin:
@esacteksab and @suzubara I've made changes based on your feedback. However, I want to also add I'm not married to `asdf`. I saw value in using it for `node`, `yarn`, and `golang` in having less tools that do the same thing. But since it's only used occasionally I am open to changing this ADR to only use `asdf` for go, or to use another one for just go. Mostly I don't want to keep using brew to setup go.

esacteksab:
> @esacteksab and @suzubara I've made changes based on your feedback. However, I want to also add I'm not married to `asdf`. I saw value in using it for `node`, `yarn`, and `golang` in having less tools that do the same thing. But since it's only used occasionally I am open to changing this ADR to only use `asdf` for go, or to use another one for just go. Mostly I don't want to keep using brew to setup go.

I agree with you in about not wanting to keep using `brew` and I like the idea of one tool to rule them all. But I don't use `node` or `yarn` to know how it might impact or influence the workflow of others. So you totally have my support here, but not sure I'm the person that needs convincing.

gidjin:
@esacteksab that makes sense

@suzubara if you're not comfortable using asdf to manage node and yarn, at least not right now, I'm cool with changing the adr to use it just for golang and kick the one to rule them all feature down the road.

suzubara:
> @suzubara if you're not comfortable using asdf to manage node and yarn, at least not right now, I'm cool with changing the adr to use it just for golang and kick the one to rule them all feature down the road.

I'm really on the fence for this one. On one hand it looks fine on paper and would be useful to have one tool that ✨  just works ✨ .  On the other hand, it has implications to a user's system beyond MilMove, and when it comes to node/yarn I'm of the mindset that if it's not broke don't fix it. I think I would feel better if we try it out with just go for now, and if it goes well then potentially expand it to node/yarn (and maybe it becomes the Truss recommended way to install these)?

esacteksab:
> > @suzubara if you're not comfortable using asdf to manage node and yarn, at least not right now, I'm cool with changing the adr to use it just for golang and kick the one to rule them all feature down the road.
> 
> I'm really on the fence for this one. On one hand it looks fine on paper and would be useful to have one tool that ✨ just works ✨ . On the other hand, it has implications to a user's system beyond MilMove, and when it comes to node/yarn I'm of the mindset that if it's not broke don't fix it. I think I would feel better if we try it out with just go for now, and if it goes well then potentially expand it to node/yarn (and maybe it becomes the Truss recommended way to install these)?

I like this. I wonder if there is a way to raise awareness of `asdf` to reserve and let them take it for a spin and see how it works, maybe we can make it a part of our standard toolset/install and it can land in our Engineering Playbook

gidjin:
I'll update this ADR to reflect that, and maybe I'll drop a note to the reserve to get this on their radar or backlog.

gidjin:
@esacteksab and @suzubara I've updated the ADR as discussed. I also made the changes to use asdf for managing golang locally, not sure if they should be part of the ADR or not but figured since I had them I'd add them. If you could give a review again.

PR #4669, opened 2020-08-24
[MB-3656] ADR: Deprecate use of lodash over time

suzubara:
## Description

This is an ADR describing a strategy for gradually deprecating use of lodash in our frontend code. It also adds the bundle analyzer recommended by Create React App for analyzing our static JS.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3656) for this change


suzubara:
I'm looking for some initial feedback on the strategy I am proposing, and if accepted, I can amend this PR to add the optimization plugins and ESLint rules in a way that results in minimal initial code changes.

docs/adr/0056-lodash.md
```diff
@@ -0,0 +1,99 @@
...
+- Switch from native methods to only lodash methods
+- Don't change lodash usage, but add Babel & Webpack plugins
+- Do nothing
+
+## Decision Outcome
+
+- Chosen Alternative: Deprecate lodash methods that have close native equivalents.
+- Additionally, set up the [lodash-webpack-plugin](https://www.npmjs.com/package/lodash-webpack-plugin) and [babel-plugin-lodash](https://www.npmjs.com/package/babel-plugin-lodash) to optimize how lodash adds to our bundle size.
+- `+` Enforce one consistent style throughout the codebase
+- `+` Can replace existing lodash methods on a per-method basis using the [ESLint plugin](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore) (and prevent more instances from being added)
```
gidjin, line 70: Plus one to automated enforcement! And bundle size reductions

package.json
```diff
@@ -83,6 +83,7 @@
   "name": "client",
   "private": true,
   "scripts": {
+    "analyze": "source-map-explorer 'build/static/js/*.js'",
```
gidjin, line 86: This is neat!

.spelling
```diff
@@ -627,7 +627,32 @@ SwaggerClient
 polyfill
 TxO
 triaged
+lodash
+natively
+transpiled
+polyfilled
 # Put all custom terms BEFORE this comment, lest 'pre-commit' and 'make spellcheck' yield different errors.
  - ./docs/backend/route-planner.md
  - docs/adr/0050-doc-viewer-fork.md
```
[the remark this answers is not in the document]
suzubara: hm, I used the mdspell CLI and chose to ignore those words specifically in that file (instead of adding to the dictionary), so I assumed that was the syntax for that?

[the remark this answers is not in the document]
gidjin: I think it is but I thought the list of ignored files needed to be at the end of the `.spelling` file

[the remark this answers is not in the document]
suzubara: I think terms placed under each ignored file will be ignored specifically in those files (instead of ignoring the whole file)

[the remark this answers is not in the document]
gidjin: Yeah I looked some more and remembered that `pre-commit` and `make spellcheck` can sometimes conflict if the code passes if terms are after the list of files to ignore, but if locally it's fine and build is good then maybe it's not an issue.

suzubara:
I modified this PR to include the following:
- Set up the ESLint plugin with an initial ruleset and fix resulting issues
- Added babel lodash plugin to cherry-pick lodash modules

I started setting up the webpack lodash plugin as well, but it requires you explicitly enable certain feature sets which I found obscure, so I think it would be best to wait to turn this on until we've cut down on usage.

lynzt:
I'm all for getting rid of dependencies, but I'm am wondering if we want to address this now or focus on other (higher priority) things...??

suzubara:
> I'm all for getting rid of dependencies, but I'm am wondering if we want to address this now or focus on other (higher priority) things...??

this came up in a previous F/E check-in, because our current inconsistent usage is leading to some confusion around which syntax is preferred. additionally, it is easier for us to identify potential areas of optimization earlier on rather than waiting (at which point we may have bought deeper into things like lodash and they will be that much harder to extract). that is why I am proposing this strategy, where we declare our intention to scale back on it now, but not enforce anything that requires significant code changes right away. I anticipate turning on more and more of the ESLint rules over time so that the whole team can acclimate to the native syntax gradually.

suzubara:
> I modified this PR to include the following:
> 
> * Set up the ESLint plugin with an initial ruleset and fix resulting issues
> * Added babel lodash plugin to cherry-pick lodash modules
> 
> I started setting up the webpack lodash plugin as well, but it requires you explicitly enable certain feature sets which I found obscure, so I think it would be best to wait to turn this on until we've cut down on usage.

One more change: I commented out the babel lodash plugin because something about using a custom babel config with rescripts is not working, and messing up our SVG imports. I will create follow up stories to figure this out but for now I would just like to get the ADR & lint rules in. @gidjin do you mind doing another pass since this now includes code changes?

.rescriptsrc
```diff
@@ -1,8 +1,7 @@
...
-    '.eslintrc',
-    require.resolve('./terser-rescript.js'),
-    require.resolve('./case-sensitive-paths-rescripts.js'),
-  ],
-]
+  // ['use-babel-config', '.babelrc'], // TODO add babel lodash plugin
+  ['use-eslint-config', '.eslintrc'],
+  require.resolve('./terser-rescript.js'),
+  require.resolve('./case-sensitive-paths-rescripts.js'),
+  // require.resolve('./lodash-webpack-rescript.js') // TODO add this in once we've reduced lodash usage
```
gidjin, line 6: This format is what we had incorrect before? Good find, glad we're actually calling our things now. How will we know we've reduced our lodash usage enough to enable the `lodash-webpack-rescripts` plugin?

terser-rescript.js
```diff
@@ -27,14 +27,6 @@ module.exports = (webpackConfig) => {
...
           parallel: false,
           sourceMap: false,
-          chunkFilter: (chunk) => {
-            // Exclude uglification for the `vendor` chunk
-            if (chunk.name === 'vendor') {
-              return false;
-            }
-
-            return true;
-          },
```
gidjin, line 37: Do you know why we needed to exclude uglification for vendor chunk? I'm not sure I grok that whole statement.

docs/adr/0057-lodash.md
```diff
@@ -0,0 +1,99 @@
...
+- [pick](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_pick)
+- [reject](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_reject)
+- [some](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_some)
+- [sortBy](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_sortby-and-_orderby)
+- [startsWith](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_startsWith)
+- [toUpper](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_toupper)
+- [uniq](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_uniq)
+- [without](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_without)
+
+**Methods to retain (links to lodash docs):**
```
gidjin, line 42: Is this and the above a comprehensive list? I wonder if we should explicitly call out if it is or not.

docs/adr/0057-lodash.md
```diff
@@ -0,0 +1,99 @@
...
+- Deprecate lodash methods that have close native equivalents
+- Deprecate lodash entirely
+- Switch from native methods to only lodash methods
+- Don't change lodash usage, but add Babel & Webpack plugins
+- Do nothing
+
+## Decision Outcome
+
+- Chosen Alternative: Deprecate lodash methods that have close native equivalents.
+- Additionally, set up the [lodash-webpack-plugin](https://www.npmjs.com/package/lodash-webpack-plugin) and [babel-plugin-lodash](https://www.npmjs.com/package/babel-plugin-lodash) to optimize how lodash adds to our bundle size.
```
gidjin, line 68: Are there existing stories or docs of the tech debt that note these still need to happen?

gidjin (approved):
LGTM! I added a couple comments on areas I had questions but they are minor

terser-rescript.js
suzubara: no I don't, but this was something I had added awhile back when looking into webpack build failures in CI that was taken from another github discussion somewhere, and fixing the rescripts syntax surfaced the fact that this option was deprecated (and had maybe never been implemented??) 🤷‍♀️

.rescriptsrc
suzubara: the `lodash-webpack-plugin` works by completely [removing feature sets](https://github.com/lodash/lodash-webpack-plugin#feature-sets) from lodash (and things break if you remove something that you need), so I think once we have scoped down what we are using to a more understood subset, we can introduce this

docs/adr/0057-lodash.md
suzubara: it's comprehensive to everything we're currently using from lodash (not the entirety of lodash). I imagine this list may evolve over time.. I wasn't sure if it made sense to include in the ADR itself but I wanted to have an easy place to point people to see what the alternative syntax was

docs/adr/0057-lodash.md
suzubara: since it is currently unclear how to label this work in JIRA, I have added to a running tech debt doc I'm working on

docs/adr/0057-lodash.md
gidjin: I'm not opposed to updating this ADR's lists if we find more things to deprecate or retain.

jacquelineIO (approved):
looks good.

PR #4671, opened 2020-08-24
MB-3552: Removing prd safeties from upload-secure-migrations

cblkwell:
## Description

This switches on the functionality for uploading migrations into the bucket for the prd environment. I've added the appropriate permissions for engineers in the GovCloud `prd` account and tested to make sure they can do an `ls` of the bucket and `cp` the files appropriately. I am not sure if we want to merge this immediately, but this should close the loop and make it possible to move forward with `prd`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -839,7 +839,8 @@ tasks_post_file_to_gex: tasks_build_linux_docker ## Run post-file-to-gex from in
 #
 
 .PHONY: run_prod_migrations
-run_prod_migrations: run_com_prod_migrations run_gov_prod_migrations
+run_prod_migrations: run_com_prod_migrations
+	# run_gov_prod_migrations
 
 .PHONY: run_com_prod_migrations
 run_com_prod_migrations: bin/milmove db_deployed_migrations_reset ## Run Commercial Prod migrations against Deployed Migrations DB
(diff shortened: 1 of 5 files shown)


cblkwell:
> Please also change these lines that talk about this only working in exp/stg for govcloud: https://github.com/transcom/mymove/blob/master/Makefile#L855
> https://github.com/transcom/mymove/blob/master/scripts/generate-secure-migration#L77
> https://github.com/transcom/mymove/blob/master/scripts/pricing-import#L126
> https://github.com/transcom/mymove/blob/master/scripts/README.md
> 
> I think these should be in the same pr as this.

Hrm. Do engineers need to be able to *get* the migrations from the prod bucket as well? I didn't realize that. :/

kilbergr:
@cblkwell currently they do retrieve them from prod in commercial. There's a ticket to address pulling them from govcloud here: https://dp3.atlassian.net/browse/MB-3639
I don't think you have to do that work here.

cblkwell:
> https://dp3.atlassian.net/browse/MB-3639

Yeah, I updated the permissions. I've added the things I think you wanted me to fix -- I also made some other changes to the `Makefile` that I'm not sure you wanted, but seemed like it was a good idea, like making the `run_$ENV_migrations` targets run against both commercial and AWS -- if you think that's a bad idea I can revert those particular changes.

Makefile
```diff
@@ -852,20 +852,19 @@ run_com_prod_migrations: bin/milmove db_deployed_migrations_reset ## Run Commerc
...
+ 	DB_HOST=localhost \
+ 	DB_PORT=$(DB_PORT_DEPLOYED_MIGRATIONS) \
+ 	DB_NAME=$(DB_NAME_DEPLOYED_MIGRATIONS) \
+ 	DB_DEBUG=0 \
+ 	bin/milmove migrate
 
 .PHONY: run_staging_migrations
-run_staging_migrations: run_com_staging_migrations ## Currently: Run Commercial Staging migrations against Deployed Migrations DB in commercial
-	# run_gov_staging_migrations
+run_staging_migrations: run_com_staging_migrations i
```
kilbergr, line 866:
```suggestion
run_staging_migrations: run_com_staging_migrations
```

PR #4680, opened 2020-08-25
MB-3692 Check for destination address before referencing

abbyoung:
## Description

* Fixes bug where attempting to add weights to a shipment without a destination address would return a 500.
* If an estimated weight is updated on an `mtoShipment` and a `scheduledPickupDate` exists, `mto_shipment_updater` calculates a `requiredDeliveryDate`. This calculation requires a `destinationAddress`, which isn't required when creating a shipment. 
* This pr checks that a `destinationAddress` exists before trying to calculate the `requiredDeliveryDate`.
* It also adds a test that would've originally caught this bug.

## Setup

To run tests: `go test ./pkg/handlers/primeapi --testify.m "TestUpdateMTOShipmentHandler" -v`

You can also test manually:

* `make db_dev_e2e_populate`
* Pick an entry in the `mto_shipments` table and delete any estimated weights, plus the destination_address_id and secondary_delivery_address_id.
* Use the `/mto-shipments/:mtoShipmentId` endpoint to update the shipment with a payload similar to the following:

```json
{
    "primeEstimatedWeight": 600,
   "primeActualWeight": 500
}
```

You should receive a 200 and the updated payload. 

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3692) for this change

### DIFF
--- a/pkg/handlers/primeapi/mto_shipment_test.go
+++ b/pkg/handlers/primeapi/mto_shipment_test.go
@@ -285,6 +285,7 @@ func ClearNonUpdateFields(mtoShipment *models.MTOShipment) *primemessages.MTOShi
 func (suite *HandlerSuite) TestUpdateMTOShipmentHandler() {
 	primeEstimatedWeight := unit.Pound(500)
 	primeEstimatedWeightDate := testdatagen.DateInsidePeakRateCycle
+	primeActualWeight := unit.Pound(600)
 	move := testdatagen.MakeAvailableMove(suite.DB())
 	mtoShipment := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
 		Move: move,
@@ -292,6 +293,7 @@ func (suite *HandlerSuite) TestUpdateMTOShipmentHandler() {
 			Status: models.MTOShipmentStatusSubmitted,
 		},
 	})
+
 	mtoShipment.PrimeEstimatedWeight = &primeEstimatedWeight
 	mtoShipment.PrimeEstimatedWeightRecordedDate = &primeEstimatedWeightDate
 
@@ -300,6 +302,14 @@ func (suite *HandlerSuite) TestUpdateMTOShipmentHandler() {
 		Move: mtoNotAvailable,
 		MTOShipment: models.MTOShipment{
 			Status: models.MTOShipmentStatusSubmitted,
+		},
+	})
+
+	minimalMtoShipment := testdatagen.MakeMTOShipmentMinimal(suite.DB(), testdatagen.Assertions{
+		Move: move,
+		MTOShipment: models.MTOShipment{
+			Status:              models.MTOShipmentStatusDraft,
+			ScheduledPickupDate: &time.Time{},
 		},
 	})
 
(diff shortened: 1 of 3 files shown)


pkg/handlers/primeapi/mto_shipment_test.go
```diff
@@ -365,6 +375,26 @@ func (suite *HandlerSuite) TestUpdateMTOShipmentHandler() {
...
+			MtoShipmentID: *handlers.FmtUUID(minimalMtoShipment.ID),
+			Body:          ClearNonUpdateFields(&minimalMtoShipment),
+			IfMatch:       eTag,
+		}
+
+		response := handler.Handle(params)
+		suite.IsType(&mtoshipmentops.UpdateMTOShipmentOK{}, response)
+
+		okResponse := response.(*mtoshipmentops.UpdateMTOShipmentOK)
+		suite.Equal(minimalMtoShipment.ID.String(), okResponse.Payload.ID.String())
```
shimonacarvalho, line 395: Would be good to check that the response has the updated `PrimeEstimatedWeight` and `PrimeActualWeight`

pkg/handlers/primeapi/mto_shipment_test.go
```diff
@@ -365,6 +375,26 @@ func (suite *HandlerSuite) TestUpdateMTOShipmentHandler() {
 		suite.Equal(mtoShipment.ID.String(), okResponse.Payload.ID.String())
 	})
 
+	suite.T().Run("Successful PUT - Update weights on minimal shipment estimated weights", func(t *testing.T) {
+		minimalMtoShipment.PrimeEstimatedWeight = &primeEstimatedWeight
```
shimonacarvalho, line 379: Just wondering if you know for sure this minimalMTOShipment has a nil DestinationAddress and PickupAddress.

pkg/services/mto_shipment/mto_shipment_updater.go
```diff
@@ -120,9 +122,11 @@ func setNewShipmentFields(planner route.Planner, db *pop.Connection, oldShipment
 	}
 
 	// Updated based on existing fields that may have been updated:
-	if oldShipment.ScheduledPickupDate != nil && oldShipment.PrimeEstimatedWeight != nil {
+	if oldShipment.ScheduledPickupDate != nil && oldShipment.PrimeEstimatedWeight != nil && oldShipment.DestinationAddress != nil {
```
shimonacarvalho, line 125: My guess is the pickup address being nil also bombs out here, because we'd be dereferencing it before using. 
Can you check that? I will try to check as well.

pkg/testdatagen/make_mto_shipment.go
```diff
@@ -90,22 +90,19 @@ func MakeMTOShipmentMinimal(db *pop.Connection, assertions Assertions) models.MT
...
-		MoveTaskOrder:        moveTaskOrder,
-		MoveTaskOrderID:      moveTaskOrder.ID,
-		RequestedPickupDate:  &requestedPickupDate,
-		PickupAddress:        &pickupAddress,
-		PickupAddressID:      &pickupAddress.ID,
-		DestinationAddress:   &destinationAddress,
-		DestinationAddressID: &destinationAddress.ID,
-		ShipmentType:         shipmentType,
-		Status:               "SUBMITTED",
+		MoveTaskOrder:       moveTaskOrder,
```
shimonacarvalho, line 99: Do you think it might be a good idea to drop the pickupAddress as well so we can test that too in one test?

pkg/services/mto_shipment/mto_shipment_updater.go
abbyoung: i thought pickup address was required but let me check

pkg/testdatagen/make_mto_shipment.go
abbyoung: As far as I can tell, the pickup address is a required part of the minimal mto shipment. The UI requires it before allowing a user to submit, and the prime api requires it when creating. Do we still need to drop it?

pkg/services/mto_shipment/mto_shipment_updater.go
abbyoung: done!

pkg/testdatagen/make_mto_shipment.go
shimonacarvalho: No, that sounds like a good reason to keep.

PR #4069, opened 2020-05-08
MB-2438 Office app authorization

garrettqmartin8:
## Description

This PR adds functionality to restrict components and redirect users in the office app based on their user role. For instance, if I log in as a TOO, I'm taken to the TOO queue and will not be able to access the PPM office queue.

## Reviewer Notes

None of our existing office users have an explicit user role, so that presents us with the problem of not knowing where to redirect people in the system. This PR adds a migration to address that by assuming all existing office users in the system are "PPM office" users, and creates a corresponding `users_roles` association record for each user. We can manually add the TOO or TIO role to users in the admin app. [Slack context](https://ustcdp3.slack.com/archives/CP4979J0G/p1588788945076700).

## Setup

- `make db_dev_e2e_populate`
- `make server_run`
- `make client_run`
- Try logging into the office app as each user type. You should be taken to the correct screen for each role and shouldn't be able to access the screens of another role. For example, if you log in as a new TOO user and attempt to visit `officelocal:3000/queues/new`, you'll be redirected back to the TOO app.
- For users with multiple roles (most likely Truss folks), you'll see a "Change user role" link at the top left of your screen. From there, you can select the other roles that you have access to.

## Code Review Verification Steps

  * [ ] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://github.com/transcom/mymove/tree/master/docs/database.md#zero-downtime-migrations))
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2483) for this change

PR #3182, opened 2019-12-11
Authorization middleware

ralren:
## Description

This is the work done to check if the roles of an endpoint a user is accessing matches the roles for that user. If there's a match, the user is allowed to access that endpoint. Otherwise, an error will occur.

## Setup

You can not log on to the app with the feature flag on. 
Please verify that the app works normally when feature flag is off.
Also verify that role authorization hits hard when feature flag is on.
You can further test manually by appending an accepted role to the session.

```sh
make server_run
make client_run
make server_test
```

## Code Review Verification Steps
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-599?atlOrigin=eyJpIjoiMDU1ZmRkMjk4M2IwNGUwODllYzMyOTgxMDJmOTRjNjciLCJwIjoiaiJ9) for this change
* [this article](tbd) explains more about the approach used.

PR #3177, opened 2019-12-10
Redirect TOO on signup (local dev only)

mkrump:
## Description

The first time a user visits the office app and successfully creates a login.gov account, if they are not already in the `office_users` table the will be assumed to be a too and redirected to a page instructing them to wait for the verification of their account. This work is hidden behind the feature flag `FEATURE_FLAG_ROLE_BASED_AUTH`. 

## Reviewer Notes

The existing auth code was getting pretty hard to modify, so for the new auth work I've tried to encapsulate the authorization of unknown users (the function `authorizeUnknownUser`), in a new function `authorizeUnknownUserNew`. To verify that the the functionality of the old `authorizeUnknownUser` is unchanged after the feature flag is disabled i added a testing flag to the auth tests that you can run with `go test ./pkg/auth/authentication -args -new_auth=true` and all of the old `authorizeUnknownUser` will be run with the new function.

## Setup

`make db_dev_reset && make db_dev_migrate`
`make server_run`
`make client_run`

1)  Create a new office user on local dev using login.gov
2) The login should work as it normally does
3) In `envrc.local` set the feature flag to true `export FEATURE_FLAG_ROLE_BASED_AUTH=true`
4) Create a new office user on local dev using login.gov
5) You should be redirected to the `/verification-in-progress` page

## Code Review Verification Steps
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References
* [Jira story](https://dp3.atlassian.net/browse/MB-750) for this change

PR #4688, opened 2020-08-26
MB-3712 Remove unused role-based auth code

monfresh:
## Description

The code removed in this PR was written 8 months ago by Team Roci (
Matt Krump and Ren specifically), but the only way this code would
ever get executed is if the feature flag `FEATURE_FLAG_ROLE_BASED_AUTH`
was set to `true`. This flag defaulted to `false`, and is currently
not set in any environment, and has never been as far as I can tell.

The idea behind this code was to protect API endpoints so that only
users with the appropriate roles could access them. The implementation
was based on [ADR #0040](https://github.com/transcom/mymove/blob/master/docs/adr/0040-role-base-authorization.md). A
new "x-swagger-roles" tag was added to a couple of the Swagger YAML
files on each endpoint, indicating the list of roles that can access
the endpoint. The `RoleAuthMiddleware` was added to compare this list
with the roles the user has (by checking the session).

Here are the 3 original PRs that implemented this middleware-based
authorization:

https://github.com/transcom/mymove/pull/3177
https://github.com/transcom/mymove/pull/3182
https://github.com/transcom/mymove/pull/3217

Why this was never turned on, nobody seems to know. Since then, we
implemented something similar, but on the front end only, and only
for the office app. We check a user's role before giving access to
an office endpoint. Here is the original PR:
https://github.com/transcom/mymove/pull/4069

This code seems useful to me, so I'm surprised it was never used.
We should find out whether we do need these protections on all the
endpoints that were originally meant to be protected, and if we still
want to follow ADR 0040. If so, I would still merge this and open a
new story to rewrite the authentication so that it uses this
role-based middleware.

## Reviewer Notes

Are any of the Swagger endpoints to which we added the `x-swagger-roles` tag protected in some other way today? If not, do we want to protect them?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3712) for this change


suzubara (approved):
I am not very familiar with the context behind the original code, but can confirm the F/E code being deleted doesn't seem to be used at all

monfresh:
I looked into this some more, and I am confident we should remove this code. All of the `x-swagger-roles` that were added to the internal API wouldn't have an effect unless the feature flag was set in the handler context (even with the feature flag turned on), which was never added in the original code. After I set the feature flag in `serve.go` like this:
```golang
handlerContext.SetFeatureFlag(
    handlers.FeatureFlag{Name: cli.FeatureFlagRoleBasedAuth, Active: v.GetBool(cli.FeatureFlagRoleBasedAuth)},
)
```
I saw the authorization code being called. However, many endpoints seem to have been unnecessarily protected, or perhaps protected based on assumptions that didn't exist at the time. For example, the endpoint to GET `/users/is_logged_in` would only allow the roles `ppm_office_users`, `contracting_officer`, and `customer`. However, when a service member first signs in (which creates their entry in the `users` table), they don't automatically get a role. So, if this role-based auth feature is turned on, no service members can sign in.

Similarly, if you remove the `x-swagger-roles` from `users/logged_in` and `users/is_logged_in` so that you can actually sign in, then other protected endpoints will prevent you from creating a move or updating your service member info for example.

PR #4694, opened 2020-08-27
[MB-3835] Cypress optimizations

suzubara:
## Description

This PR makes some changes to Cypress custom commands and code organization in order to reduce test flakiness and memory consumption, and increase clarity for Cypress work going forward. The most effective change is the addition of a new command for signing in via the API, `apiSignInAsUser`, to replace `signInAsUserPostRequest`.

The reason for this is the `signInAsUserPostRequest` command, which made repeated calls to clear cookies and includes specific logic for testing CSRF cookies, was used excessively for logging in across many tests. The new command is more straightforward and performs the minimum required in order to log in with a given user ID. I did some informal benchmarking, and running the full test suite using the new command does seem to speed things up noticeably.

I also refactored the CSRF tests to be more explicit about the logic that was previously in `signInAsUserPostRequest`. Clearing cookies in between tests is not necessary because [Cypress already does this automatically](https://docs.cypress.io/guides/references/best-practices.html#Is-resetting-the-state-necessary).

Other Cypress changes include:
- General test organization, readability improvements (avoid over-abstracting into functions & commands for code that is not repeated or includes assertions), removed unused code
- Explicitly set base URL in `before` blocks for all test suites
- Un-comment-out TOO flow tests

I specifically wanted to avoid making changes to what the tests themselves are testing in this PR, because I am not yet very familiar with the PPM flows. Noted below, I think looking at those will be a good next step for improving Cypress perf.

A few additional code changes are called out in the PR comments.

## Reviewer Notes

During this effort, I came across a few other tasks that I think will help continue to improve our Cypress performance and reduce flakiness/memory issues. These include:
- Move specific UUIDs into constants file, document how they map to the seed data.
- Look into the tests themselves, remove unnecessary & redundant assertions. Avoid testing the same thing multiple times.
- Specifically investigate ppmCloseout, which caused out of memory errors in my local Chrome instance several times
- Find assertions in E2E tests that can be made redundant with Jest tests

Also, I want to note that this PR will probably _not_ eliminate flaky Cypress failures. But hopefully it will reduce the frequency of them, and as we move through the above work, they will be practically eliminated.

## Setup

You can run the Cypress tests locally (`yarn test:e2e-clean`) or just verify they passed in CI.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3835) for this change

## Screenshots


cypress/support/index.js
```diff
@@ -25,19 +25,3 @@ Cypress.on('window:load', (win) => {
   node.innerHTML = 'html { scroll-behavior: inherit !important; }';
   document.body.appendChild(node);
 });
-
```
suzubara, line 28: I removed this because it frequently caused errors when running tests locally, and the defaults (which screenshot the runner when a test fails) seem reasonable enough. If this was configured for a specific reason let me know and maybe we can revert.

cypress/support/index.js
```diff
@@ -25,19 +25,3 @@ Cypress.on('window:load', (win) => {
...
-  capture: 'fullPage',
-});
-
-afterEach(function () {
-  if (this.currentTest.state === 'failed') {
-    // Take another screenshot so we get the full page
-    cy.screenshot();
-  }
-});
-
```
suzubara, line 41: I removed this because it frequently caused errors when running tests locally, and the defaults (which screenshot the runner when a test fails) seem reasonable enough. If this was configured for a specific reason let me know and maybe we can revert.

cypress/support/utils.js
```diff
@@ -1,8 +0,0 @@
-import moment from 'moment';
-
-// convert nth day of current month to date string YYYY-MM-DD
-export const nthDayOfCurrentMonth = (n) => {
```
suzubara, line 4: this was not used anywhere

src/pages/Office/MoveDetails/MoveDetails.jsx
```diff
@@ -103,7 +103,7 @@ export class MoveDetails extends Component {
 
     sections.forEach((section) => {
       const sectionEl = document.querySelector(`#${section}`);
-      if (sectionEl.offsetTop <= distanceFromTop && sectionEl.offsetTop + sectionEl.offsetHeight > distanceFromTop) {
+      if (sectionEl?.offsetTop <= distanceFromTop && sectionEl?.offsetTop + sectionEl?.offsetHeight > distanceFromTop) {
```
suzubara, line 106: added defensive checks here, this seemed to cause test failures sometimes (maybe a render timing issue?)

src/scenes/Landing/index.jsx
```diff
@@ -33,10 +33,14 @@ import { selectActiveOrLatestMove } from 'shared/Entities/modules/moves';
 export class Landing extends Component {
   componentDidMount() {
     // Load user into entities
-    const { showLoggedInUser } = this.props;
-    showLoggedInUser();
+    const { isLoggedIn, showLoggedInUser } = this.props;
+    if (isLoggedIn) {
+      showLoggedInUser();
```
suzubara, line 38: added this check because if `showLoggedInUser` is attempted when not logged in (ie., on the customer / URL) it results in an uncaught exception (because SwaggerRequest throws fetch calls that are not `ok`) and makes Cypress thinks there's a code error

cypress/integration/mymove/ppm.js
```diff
@@ -1,58 +1,56 @@
-/* global cy */
+describe('the PPM flow', function () {
+  before(() => {
+    cy.prepareCustomerApp();
+  });
 
-import { milmoveAppName } from '../../support/constants';
+  it('can submit a PPM move', () => {
+    // profile@comple.te
```
sojeri, line 7: not for this PR, but I'm thinking at some point we should break these IDs out into a separate file so they can be more semantically used like profileCompleteUser and so on. just wanted to call out here in case you end up doing anotoher chunk of cypress work before me. :)

cypress/support/commands.js
```diff
@@ -126,103 +73,66 @@ Cypress.Commands.add('waitForReactTableLoad', () => {
...
 
-// Persist session cookies throughout a testing block (use in beforeEach)
-Cypress.Commands.add('persistSessionCookies', () => {
-  Cypress.Cookies.preserveOnce('masked_gorilla_csrf', 'office_session_token', '_gorilla_csrf');
+Cypress.Commands.add('signInAsNewTIOUser', () => {
+  cy.signInAsNewUser(TIOOfficeUserType);
 });
 
-Cypress.Commands.add(
-  'signInAsUserPostRequest',
```
sojeri, line 148: ✂️ ✂️ ✂️

cypress/support/index.js
sojeri: IDK who would own this but you may need to raise beyond this PR. maybe @travelar knows if this is a compliance thing we need to keep around?

suzubara: the default behavior is for Cypress to take a screenshot when the test fails. this code was effectively just changing the screenshot type from `runner` (the full test runner UI, which usually includes information about what failed) to `fullPage` (which is just the application UI and may not be as useful) (https://docs.cypress.io/api/cypress-api/screenshot-api.html). so I don't think there is anything compliance-related, especially since these are all running against test data and the screenshots are stored in CI, but I will ping on slack in case there's something I don't know.

cypress/integration/mymove/landingPages.js
```diff
@@ -1,30 +1,49 @@
-/* global cy */
-
-import { milmoveAppName } from '../../support/constants';
+// TODO - sometimes tests result in a 500 response to this URL;
+// http://milmovelocal:3000/internal/estimates/ppm?original_move_date=2020-08-27&origin_zip=72017&origin_duty_station_zip=50309&orders_id=fd5495d4-5105-46b8-89f8-0e6b155d900b&weight_estimate=8000
+// This causes the test to fail because of an uncaught exception (the fetch error is not handled)
```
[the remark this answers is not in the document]
suzubara: ohhh good call, I wasn't sure what this error was (admittedly didn't dig into it too much)

suzubara:
> have you tried the 5.0.0 upgrade against this branch? I'm curious if these refactors are enough to pass those tests. :)

I did (albeit on a new branch) and it seems to pass okay... I didn't want to tie all of these changes with that, so I will open a separate PR and we can see how that goes!

cypress/support/commands.js
donaldthai: Thank you for cleaning this up! 😀

cypress/integration/office/txo/tooFlows.js
```diff
@@ -1,84 +1,89 @@
...
+      });
+      // Select additional service items
+      cy.get('label[for="shipmentManagementFee"]').click();
+      cy.get('label[for="counselingFee"]').click();
+      // Open modal
+      const button = cy.contains('Approve selected shipments');
+      button.should('be.enabled');
+      button.click();
+      cy.get('#approvalConfirmationModal [data-testid="modal"]').then(($modal) => {
+        cy.get($modal).should('be.visible');
```
duncan-truss, line 65: One performance optimization I see here (I don't know if this is slow or not) on a page with a lot of content like the move details with the modal, would be to use the `cy.get('#approvalConfirmationModal')` and then use the within command to limit the scope since we don't want to search through all the non modal content.

https://docs.cypress.io/api/commands/get.html#Get-in-within

suzubara: I tried to avoid making changes to the tests themselves in this PR, and want to treat substantive changes in the tests as a separate task so it's easier to measure what changes are most effective

PR #4695, opened 2020-08-28
MB-3517 hookup Edit HHG page to get and patch

amitch23:
## Description

Issues:
From the EditShipment form:
* ~~there's a race condition with the update endpoint and the loadmove endpoint on the review page. The update payload returns and updates the entities. Then, loadMove on the Summary (Review page) hits the database and loads the old data - why is it loading the old data?? Refreshing the page fetches the updated HHG.~~ We handled this by making the update shipment call a promise. However, this is STILL the behavior on the HHGDetailsForm page. Once we move to supporting multiple HHGs in the next slice, we won't be prepopulating this form if the user hits "back" from the review page, so it might be a bug we can live with until then?

On the EditShipment Page:
* ~~unable to check the "use current address box"~~ was importing Checkbox from the wrong place
* ~~Textarea (customerRemarks) cannot be added to.~~ - Needed to pass in a handleSubmit from formik as a prop into TextArea
* ~~Submitting form hasn't been fully tested yet to ensure updating happens and updated HHG is displayed on the review page.~~ it works! NOTE: updating either address does NOT work, due to a known bug in the update service. Should this be highlighted visually to indicate that it is a known bug?

Cleanup issues:
* lots of repeat code in both forms - need to extract out the common functions and import those from both forms (e.g. reshaping the mtoshipment with the agents via setting initial state, etc.)
* Add e2e test for editing shipment from Review page

## Reviewer Notes

* What should we do about the known bug for address not updating?
* Are we OK with the buggy behavior of the HHGdetails form, given that it will soon change with multiple HHGs? (see above)

## Setup

Run the server/client, and go through an HHG flow. Edit a shipment and test all expected functionality from the review page. Click "back" from the review page after the HHG to see the buggy behavior described above.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/secure/RapidBoard.jspa?rapidView=28&projectKey=MB&modal=detail&selectedIssue=MB-3517) for this change

## Screenshots

editshipmentpage


cypress/integration/mymove/hhg.js
```diff
@@ -204,6 +204,29 @@ function customerSetsUpAnHHGMove() {
...
+
+  cy.location().should((loc) => {
+    expect(loc.pathname).to.match(/^\/moves\/[^/]+\/review\/edit-shipment/);
+  });
+
+  // Ensure remarks is displayed in form
+  cy.get(`[data-testid="firstName"]`).last().type('Johnson');
+
+  cy.get(`[data-testid="remarks"]`).contains('some customer remark');
+  // Edit remarks
```
kimallen, line 218: I'm wondering if we should also add a test for an agent update (since the backend is a bit more complicated since it updates a different table).  Or maybe we should just trust the handler will test that?

src/scenes/MyMove/index.jsx
```diff
@@ -119,7 +119,7 @@ export class AppWrapper extends Component {
                   {props.context.flags.hhgFlow && <ValidatedPrivateRoute exact path="/home-2" component={Home} />}
                   <ValidatedPrivateRoute exact path="/moves/:moveId/edit" component={Edit} />
                   <ValidatedPrivateRoute exact path="/moves/review/edit-profile" component={EditProfile} />
-                  <ValidatedPrivateRoute exact path="/moves/review/edit-shipment" component={EditShipment} />
+                  <ValidatedPrivateRoute exact path="/moves/:moveId/review/edit-shipment" component={HHGMoveSetup} />
```
kimallen, line 122: Can we remove `review` from the path?  I believe it should also be accessed from the new Home page, so it's not necessarily relevant to the review page.  (probably true for the other edits too, but you don't need to worry about that)

src/pages/MyMove/HHGMoveSetup.jsx
```diff
@@ -16,11 +18,15 @@ class HHGMoveSetup extends Component {
   }
```
kimallen, line 19: I know this is existing code, but wondering if it makes more sense to call this `HHGShipmentSetup`?

src/components/Customer/EditShipment.jsx
```diff
@@ -46,24 +54,79 @@ const AgentSchema = Yup.object().shape({
...
+    const { agents } = mtoShipment;
+    const formattedMTOShipment = { ...mtoShipment };
+    if (agents) {
+      const receivingAgent = agents.find((agent) => agent.agentType === 'RECEIVING_AGENT');
+      const releasingAgent = agents.find((agent) => agent.agentType === 'RELEASING_AGENT');
+
+      // Remove dashes from agent phones for expected form phone format
+      if (receivingAgent) {
+        const formattedAgent = cleanAgentPhone(receivingAgent);
+        if (!isEmpty(formattedAgent)) {
```
kimallen, line 115: There's an ADR to phase out of using lodash.  Can we do this (and other uses of lodash) in a native JS way?  (ex. `!!Object.keys(formattedAgent).length`).  (I'm passing on the feedback I got from Ron on my PR)

src/pages/MyMove/HHGMoveSetup.jsx
```diff
@@ -48,6 +54,9 @@ HHGMoveSetup.propTypes = {
     path: string.isRequired,
     url: string.isRequired,
   }).isRequired,
+  history: shape({
```
kimallen, line 57: I'm not sure where `history` is coming from in the props. If `push` comes from this same `history`, can we just pass in `history` to get `push` and `goBack`?

src/pages/MyMove/HHGMoveSetup.jsx
amitch23: changed to suggestion

PR #4699, opened 2020-08-28
MB-3830 Fix destination address lookup attempt

garrettqmartin8:
## Description

The demo dry run had [some issues](https://ustcdp3.slack.com/archives/CP6F568DC/p1598622660101200) related to pricing service items that didn't need a destination address, but failed when one wasn't set.

## Reviewer Notes

As far as I can tell, all that needed to be done was to reverse the conditional check. Am I missing anything?

## Setup

- `make db_dev_e2e_populate`
- `make server_run`
- `make client_run`
- Follow the instructions for acceptance testing payment requests in the [wiki](https://github.com/transcom/mymove/wiki/Acceptance-Testing-Payment-Requests). Make sure you submit a request for a `DPK` service item.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3830) for this change


pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go
```diff
@@ -76,10 +76,10 @@ func ServiceParamLookupInitialize(
 			return nil, services.NewNotFoundError(uuid.Nil, "looking for PickupAddressID")
 		}
 		pickupAddress = *mtoShipment.PickupAddress
-		if mtoShipment.DestinationAddressID == nil {
-			return nil, services.NewNotFoundError(uuid.Nil, "looking for DestinationAddressID")
+		if mtoShipment.DestinationAddressID != nil {
```
gidjin, line 79: Before the refactor, service items that required destination address would throw an error so this changes that. Can you throw the error from inside the specific lookup that does require destination to be present? Also, I'm wondering in real world will we ever not have the dest address when a payment request is made? Second I'm wondering if we need to do this with pickup address too

garrettqmartin8: > Can you throw the error from inside the specific lookup that does require destination to be present?

Yeah I'll try that. @pearl-truss and I ran into this being an issue.

> Also, I'm wondering in real world will we ever not have the dest address when a payment request is made?

Hmm, not sure.

> Second I'm wondering if we need to do this with pickup address too

Yeah, I guess if there may not be a destination, it might also be missing a pickup?

garrettqmartin8: @gidjin I updated the PR to restore the error upstream not run the lookups that require destination address if it's a `DPK` service item. If I can get your 👀 on verifying that you can create a `DPK` and a `DUPK` for good measure, that'd be awesome :)

PR #4711, opened 2020-08-31
Kim=mb 3713 home screen behavior

kimallen:
## Description

This adds button routing for Step 2 `Add orders`, Step 3 `Plan your shipment`, and Step 4 Confirm move request` and adds the checks for status of completion of each of the steps, along with showing the correct text based on the status.  This does not change the step number or add routing for the shipments, since they will be fetched in https://dp3.atlassian.net/browse/MB-3767.

This also replaces the dummy `move-landing` page with the Home-2 page, so that after submitting profile info, the user sees the page in the flow, and clicks `Add orders` to get to orders.

## Reviewer Notes

I added routing for the Edit buttons, though the edit was not working properly, so I commented out the edit for now along with the corresponding tests.

We will need more Cypress tests for the routing of all other buttons except for the `Add orders` button when the home page is integrated into the correct places.  Right now it's only accessible after profile orders, or by putting the `/home-2` in the url.

## Setup

Create a new customer user.  When arriving at the home page, click `Add orders`. 
Add the orders
Go to the `/home-2` page and check that the steps have changed and that `Plan your shipment` is now enabled, and click.
Choose a shipment and add a shipment.  
Once submitted, go to `/home-2`.  See the `Confirm move request` enabled.  Click and it should take you back to the review page.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3713) for this change


src/scenes/MyMove/getWorkflowRoutes.jsx
```diff
@@ -136,15 +136,11 @@ const pages = {
...
+  '/home-2': {
     isInFlow: (props) => myFirstRodeo(props) && inGhcFlow(props),
     isComplete: always,
-    render: (key, pages) => () => {
-      return (
-        <WizardPage handleSubmit={no_op} pageList={pages} pageKey={key}>
-          <MoveLanding />
-        </WizardPage>
-      );
+    render: (key, pages) => ({ history }) => {
```
Ronolibert, line 142: I don't think we want to add it to the workflow routes since we are already declaring this route in [this file](https://github.com/transcom/mymove/blob/102d4d8cea05d6558aedb54642aa18e04da43fa6/src/scenes/MyMove/index.jsx#L119). We could use `withRouter` instead to gain access to the `history` methods

src/pages/MyMove/Home/index.jsx
```diff
@@ -32,7 +33,27 @@ class Home extends Component {
     console.log('this is the shipment', shipment);
   };
 
+  handleNewPathClick = (path) => {
+    const { push } = this.props;
+    push(path);
```
Ronolibert, line 38: Could we use `history.push` instead of just `push`. Having `push` method on its own is a bit ambiguous on what it does

src/pages/MyMove/Home/index.jsx
```diff
@@ -32,7 +33,27 @@ class Home extends Component {
...
+
   render() {
+    const { move } = this.props;
+    const ordersPath = '/orders/';
+    const shipmentSelectionPath = `/moves/${move.id}/select-type`;
+    const confirmationPath = `/moves/${move.id}/review`;
+    const profileEditPath = '/moves/review/edit-profile';
+    const ordersEditPath = `/moves/${move.id}/review/edit-orders`;
+    const ordersCompleted = this.checkOrdersCompleted();
+    const hasShipment = !isNil(move.personally_procured_moves) || !isNil(move.shipments);
```
Ronolibert, line 54: Since we have an ADR going up that is moving away from lodash, can we opt out of using it here

src/scenes/MyMove/getWorkflowRoutes.jsx
Ronolibert: Just noticed while testing, do we want to go to the home screen in between every step? I was under the impression we would only be going to the home screen if the user logs out or presses the cancel button

kimallen: We talked, I'm going to leave it as is.

src/pages/MyMove/Home/index.jsx
kimallen: Removed lodash

PR #4722, opened 2020-09-02
[MB-3848] HHG Start page only creates shipments

Ronolibert:
## Description

HHG Start page only creates shipments

## Setup
`make server_run`
`make client_run`
Go through setup flow
Fill out HHG page and click next
Click back button
Fill out HHG page again
Finish submitting move
Visit /home-2 to see 2 shipments

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3848) for this change


cypress/integration/mymove/hhg.js
```diff
@@ -223,9 +228,16 @@ function customerSetsUpAnHHGMove() {
...
+
+function customerAddsAnotherShipment() {
+  cy.get('button.prev').should('be.enabled').click();
+  customerSetsUpAnHHGMove();
+}
 
+function customerCheckShipmentsOnHomepage(numOfShipments) {
   cy.visit('/home-2');
-  cy.get('[data-testid="shipment-list-item-container"]').contains('HHG');
+  cy.get('[data-testid="shipment-list-item-container"]', { timeout: 10000 }).should('have.length', numOfShipments);
```
Ronolibert, line 240: This was a flaky test in the CI server. Don't really like this type of solution but I guess its the recommended way to deal with flakiness in this scenario of waiting for a page to finish fully loading and extending the default time it retries looking for a DOM element. Default retry time is 4s before it times out so I increased it to 10. More info [here](https://docs.cypress.io/guides/core-concepts/retry-ability.html#Commands-vs-assertions).

sojeri: yeah, I've seen this sort of issue a lot in these longer running tests. I suspect the flakiness is related to the underlying memory leak issue in the cypress runner.

donaldthai: I think @suzubara mentioned something about using `wait` to wait for the shipments to load before checking the component. Would that work?

```js
beforeEach(() => {
  // Reference route alias
  cy.route('GET', '/internal/moves/**/mto_shipments').as('getMTOShipments');
});

cy.wait('@getMTOShipments');
```

Not a big fan of adding manual timeouts either.

Ronolibert: @donaldthai Interesting! I'll try that

donaldthai: There's at least one example in the same `hhg.js` file: https://github.com/transcom/mymove/blob/master/cypress/integration/mymove/hhg.js#L11

src/components/Customer/HHGDetailsForm.jsx
```diff
@@ -74,58 +74,10 @@ class HHGDetailsForm extends Component {
...
-
-  componentDidUpdate(prevProps) {
-    const { mtoShipment } = this.props;
-    // If refreshing page, need to handle mtoShipment populating from a promise
-    if (mtoShipment.id && prevProps.mtoShipment.id !== mtoShipment.id) {
-      this.setInitialState(mtoShipment);
-    }
   }
 
-  setInitialState = (mtoShipment) => {
```
sojeri, line 92: ✂️ ✂️ ✂️ 😋

cypress/integration/mymove/hhg.js
```diff
@@ -223,9 +229,17 @@ function customerSetsUpAnHHGMove() {
...
+
+function customerAddsAnotherShipment() {
+  cy.get('button.prev').should('be.enabled').click();
+  customerSetsUpAnHHGMove();
+}
 
+function customerCheckShipmentsOnHomepage(numOfShipments) {
   cy.visit('/home-2');
-  cy.get('[data-testid="shipment-list-item-container"]').contains('HHG');
+  cy.wait('@getMTOShipments');
```
sojeri, line 241: ah, nice. this seems more explicit than the timeout :)

cypress/integration/mymove/hhg.js
Ronolibert: @donaldthai Ok added it! Seems to work fine. I like this solution better since the response timeout default is 30s

kimallen (commented):
I think it would be helpful to mark down this PR somewhere- if we end up splitting the HHG form, we'll want to re-implement updating when someone goes back from HHG form 2 to HHG form 1.

Ronolibert:
@kimallen This same logic is in EditShipment.jsx! They were pretty much the same file with 2 minor changes with the button behavior but the data loading behavior was the same

cypress/integration/mymove/hhg.js
lynzt: Adding wait isn't great, but it will only wait the full time if it doesn't get a response... so there's that

PR #4729, opened 2020-09-02
[MB-3795] Re-enable e2e test that indirectly verifies BVS

garrettqmartin8:
## Description

Since BVS scores live in secure migrations and aren't loaded into our unit test environment, we're re-enabling an e2e test that was [commented out](https://github.com/transcom/mymove/pull/4588/files) until we could get data for the most recent performance period.

## Setup

`make e2e_test_docker`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3795) for this change


cypress/integration/mymove/ppmCloseout.js
```diff
@@ -134,8 +134,7 @@ function serviceMemberSubmitsPaymentRequestWithMissingDocuments() {
   );
 
   cy.get('.title').contains('Next step: Contact the PPPO office');
-  //TODO add this back in when we have BVS scores
-  // cy.get('.missing-label').contains('Unknown');
+  cy.get('.missing-label').contains('Unknown'); // If this test fails, we probably need to update our BVS data for the most recent performance period.
```
gidjin, line 137: Wasn't there a change to restrict it to a specific date too? I'm not sure if that happened or if that was just one of the ideas in the discussion but was never implemented. If this test isn't restricted to a specific date then I'm good with the change.

garrettqmartin8:
Shelving this work for now. Context: https://ustcdp3.slack.com/archives/CP6F568DC/p1599575344140500

tinyels (commented):
Justin and I talked and think we can just close this pr for now. 2 things:
1. I am surprised that only this one line fails when we can't price a PPM
2. He is going to find or create a story about updating PPMs to use the new rate data (instead of bvs data) when that story happens, we will want to restore this part of the test.

lynzt:
closing based on above comment

PR #4742, opened 2020-09-04
add cypress runner dep to root pkg as well

sojeri:
## Description

Unblock local e2e tests by adding new cypress runner dep to root package.json's devDeps.

## Reviewer Notes

If you `rm -rf node_modules/` (or `mv node_modules/ NOPE_modules/`), can you `yarn; yarn test:e2e` successfully?

## Setup

Lol oops. See reviewer notes. ;p

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,6 +115,7 @@
     "@testing-library/react-hooks": "^3.4.1",
     "case-sensitive-paths-webpack-plugin": "^2.3.0",
     "cypress": "~4.8.0",
+    "cypress-wait-until": "^1.7.1",
     "danger": "^10.3.0",
     "danger-plugin-jira-issue": "^1.4.1",
     "enzyme": "^3.11.0",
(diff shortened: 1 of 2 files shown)


suzubara (approved):
thanks for adding this!! Can confirm the instructions worked for me.

I think we might want to add a follow up chore to organize our JS dependencies using yarn workspaces (https://classic.yarnpkg.com/en/docs/workspaces/). Then we can have a root package.json with shared dependencies, and app- & Cypress-specific ones defined in their respective directories. I think the main thing will be to figure out if this plays nicely with CI installing deps & running Cypress.

sojeri:
@suzubara I've created the yarn workspaces story but not sure if the work will get prioritized because 1) it's probably not common that we'll add more dependencies to the e2e test runner, and 2) it does work as is as long as we update both places :)

PR #4774, opened 2020-09-11
Kim mb 3851 update tests for ppm home

kimallen:
## Description

Updates integration tests for service member app that expect to go to the home page (`/`) to now go to (`/ppm`).  Also adds some changes in the code where the path expects the ppm landing page at `/`.

## Reviewer Notes

I created an alternate sign in command to use in the SM tests `apiSignInAsPpmUser`.   When the new home page is hooked up to the flow, the tests will have to add the step of going to the new home page, and then clicking on PPM, and then starting at `/ppm`.  At that point, we can change usages of  `apiSignInAsPpmUser` back to `apiSignInAsUser`.

If I missed anywhere, the tests will still pass, since it will still be looking at the original page, so you may need to look for anything missed.  However, if I did miss anything, we will find out when we swap out the home page (and the test will break).

## Setup

run integration tests `make e2e_test` (although every time I ran it locally, there were failures, but in Circle they passed 🤷‍♀️ 

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3850) for this change


cypress/integration/mymove/orders.js
```diff
@@ -4,7 +4,7 @@ describe('orders entry', function () {
   });
 
   it('will accept orders information', function () {
-    cy.apiSignInAsUser('feac0e92-66ec-4cab-ad29-538129bf918e');
+    cy.apiSignInAsPpmUser('feac0e92-66ec-4cab-ad29-538129bf918e');
```
sojeri, line 7: if you know which user this is, can you add a note similar to the ones Suz added?

eg, `// profile@comple.te` on line 7 in `cypress/integration/mymove/ppm.js`

src/scenes/Legalese/index.jsx
```diff
@@ -54,10 +54,17 @@ export class SignedCertification extends Component {
 
   handleSubmit = () => {
     const pendingValues = this.props.values;
-    const { latestSignedCertification } = this.props;
+    const { latestSignedCertification, currentPpm } = this.props;
+    const landingPath = () => {
+      // TODO: change when supporting a combo ppm ?
+      if (Object.keys(currentPpm).length) {
+        return '/ppm';
```
sojeri, line 61: this is updating the user flow to return back to /ppm after submitting the form rather than root, right? 🎉

sojeri (commented):
I like what you have so far. Title still says WIP so I'm assuming more is coming? :)

sojeri:
> If I missed anywhere, the tests will still pass, since it will still be looking at the original page, so you may need to look for anything missed. However, if I did miss anything, we will find out when we swap out the home page (and the test will break).

I think this is fine for exactly the reason you stated: we'lll find out when we swap and whoever will fix the test to match the new correct behavior. :)

src/scenes/Legalese/index.jsx
kimallen: Yes!

cypress/integration/mymove/orders.js
kimallen: Done

kimallen:
> I like what you have so far. Title still says WIP so I'm assuming more is coming? :)

This is it!  I had forgotten to take WIP off after moving it out of draft state (I created a PR to see if tests passed in Circle)

lynzt (approved):
LGTM and like you mentioned, we'll find any missing tests when we convert this over and can update them then... 
`apiSignInAsPpmUser` 👍

🚢

PR #4775, opened 2020-09-11
Adding frontend guidelines

suzubara:
## Description

This PR adds some checks to encourage F/E best practices: write more unit tests, and proactively move/create files in the [new directory structure](https://github.com/transcom/mymove/blob/master/docs/adr/0048-frontend-file-org.md).

Changes are:
- CircleCI `client_test` job will now collect Jest coverage, and fail if coverage falls below thresholds specified in `package.json` (currently set to values just below our current numbers, with the intent that this number will creep up over time)
- DangerJS will fail PRs if new code is added in `src/shared, src/scenes`
- DangerJS will add a _warning_ if code in `src/shared, src/scenes` is modified (there are legit reasons for this, but the point is to encourage the PR author to move code outside of those places)
- Add design and @suzubara as reviewers for PRs that touch stories files & CSS files

## Setup

You can test the Dangerfile changes against another open PR if you check out this branch, and run:
```
➜ yarn danger pr <url-to-PR>
```

You can also test the code coverage threshold by changing the numbers in the `package.json` to be higher, and run
```
➜ yarn test:coverage
```
OR
```
➜ make client_test_coverage
```
locally to verify it fails due to not enough test coverage.

### DIFF
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -24,6 +24,20 @@ const fileChecks = () => {
   // load all modified and new files
   const allFiles = danger.git.modified_files.concat(danger.git.created_files);
 
+  const legacyFiles = danger.git.fileMatch('src/shared/**/*', 'src/scenes/**/*');
+
+  if (legacyFiles.created) {
+    fail(`New files have been created under one of the legacy directories
+(src/shared or src/scenes). Please refrain from adding files in these
+directories`);
+  }
+
+  if (legacyFiles.modified) {
+    warn(`Files located in legacy directories (src/shared or src/scenes) have
+been edited. Are you sure you don’t want to also relocate them for improved
+linting coverage?`);
+  }
+
   // Request changes to app code to also include changes to tests.
   const hasAppChanges = allFiles.some((path) => !!path.match(/src\/.*\.jsx?/));
   const hasTestChanges = allFiles.some((path) => !!path.match(/src\/.*\.test\.jsx?/));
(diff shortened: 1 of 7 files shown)


dangerfile.ts
```diff
@@ -24,6 +24,20 @@ const fileChecks = () => {
   // load all modified and new files
   const allFiles = danger.git.modified_files.concat(danger.git.created_files);
 
+  const legacyFiles = danger.git.fileMatch('src/shared/**/*', 'src/scenes/**/*');
+
+  if (legacyFiles.created) {
+    fail(`New files have been created under one of the legacy directories
+(src/shared or src/scenes). Please refrain from adding files in these
+directories`);
```
sojeri, line 32: Can you update the failure & warning below it to reference the appropriate ADR and/or wiki page? That way if this guidance is news to someone they can easilly read the context and find out more about the correct directory structure.

dangerfile.ts
```diff
@@ -66,36 +80,45 @@ const cypressUpdateChecks = async () => {
...
-    const rootPackageDiff = await danger.git.diffForFile(rootPackageFile)
+    const rootPackageDiff = await danger.git.diffForFile(rootPackageFile);
     if (rootPackageDiff && rootPackageDiff.diff.includes(cypressPackageName)) {
-      hasRootCypressDepChanged = true
+      hasRootCypressDepChanged = true;
 
-      let diff = rootPackageDiff.diff.split(cypressPackageName)[1]; // we don't care about diff before cypress
-      rootVersion = diff.match(versionRegex)[0]; // the first version # will be cypress's
+      const diff = rootPackageDiff.diff.split(cypressPackageName)[1]; // we don't care about diff before cypress
+      [rootVersion] = diff.match(versionRegex); // the first version # will be cypress's
```
sojeri, line 93: TIL!

dangerfile.ts
gidjin: Note using `fail` will block the pr as I understand. Should we reword to indicate that you cannot add new files there? I read `Please refrain from adding...` as you can if you have to, but the fail makes me think it's more of a you can't as it should go in the new locations.

dangerfile.ts
gidjin: I do like the destructuring syntax in JS

gidjin (approved):
Looks good to me. I agree with Jeri adding a link to the ADR would be great, but I don't need to re-approve after that. I did leave a note about wording of the message that you can take or leave as you like :)

Shkeating (approved):

PR #4785, opened 2020-09-14
MB-4226 Rename internal folders to payloads and move files up one level

sandy-wright:
## Description

This PR renames the `pkg/handlers/primeapi/internal` folder to `payloads`. This will allow the payload to model functions to be accessed from outside of the `primeapi` package, which will be needed for creating notifications for the prime (in the `pkg/services/event` package).

## Setup

The main way to test this code is to ensure that the tests are all still passing:

```sh
make server_test
```

You can also hit some of the Prime API endpoints to double-check that things are working as expected.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4226) for this change.


LeDeep (approved):
lgtm!

PR #4788, opened 2020-09-15
[MB-4223] Creates notification handler for mtoshipment events

shimonacarvalho:
## Description

Please see tickets [MB-3438](https://dp3.atlassian.net/browse/MB-3438) / [MB-4223](https://dp3.atlassian.net/browse/MB-4223) for full info 

Creates a notification handler for mtoshipment events. 
Note that I combined the paymentRequest and MTOshipment handlers into a default handler and only broke out the assemblePayload function. Originally i had thought we would break out the whole handler, but since that was only needed for future work, I decided it was best to have simpler, less duplicated code today.

Test created but cannot fully pass until successful creation of notification. Currently it passes but expects an error.

### DIFF
--- a/pkg/services/event/event.go
+++ b/pkg/services/event/event.go
@@ -51,6 +51,12 @@ const MTOShipmentCreateEventKey KeyType = "MTOShipment.Create"
 // MTOShipmentUpdateEventKey is a key containing MTOShipment.Update
 const MTOShipmentUpdateEventKey KeyType = "MTOShipment.Update"
 
+// MTOServiceItemCreateEventKey is a key containing MTOServiceItem.Create
+const MTOServiceItemCreateEventKey KeyType = "MTOServiceItem.Create"
+
+// MTOServiceItemUpdateEventKey is a key containing MTOServiceItem.Update
+const MTOServiceItemUpdateEventKey KeyType = "MTOServiceItem.Update"
+
 var eventModels map[KeyType]eventModel = map[KeyType]eventModel{
 	PaymentRequestCreateEventKey: {PaymentRequestCreateEventKey, models.PaymentRequest{}},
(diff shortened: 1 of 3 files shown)


abbyoung (approved):
lgtm!

pkg/services/event/event.go
```diff
@@ -45,9 +45,23 @@ const PaymentRequestCreateEventKey KeyType = "PaymentRequest.Create"
...
+const MTOServiceItemCreateEventKey KeyType = "MTOServiceItem.Create"
+
+// MTOServiceItemUpdateEventKey is a key containing MTOServiceItem.Update
+const MTOServiceItemUpdateEventKey KeyType = "MTOServiceItem.Update"
+
 var eventModels map[KeyType]eventModel = map[KeyType]eventModel{
 	PaymentRequestCreateEventKey: {PaymentRequestCreateEventKey, models.PaymentRequest{}},
 	PaymentRequestUpdateEventKey: {PaymentRequestUpdateEventKey, models.PaymentRequest{}},
+	MTOShipmentCreateEventKey:    {MTOShipmentCreateEventKey, models.MTOShipment{}},
+	MTOShipmentUpdateEventKey:    {MTOShipmentUpdateEventKey, models.MTOShipment{}},
```
sandy-wright, line 64: I think the MTO Service Item event keys need to be added here as well

sandy-wright (changes requested):
Looks good - thanks for adding the service item and the shipment event code! I think I found one tiny chunk that needs to be changed, but everything else is golden

pkg/services/event/event.go
shimonacarvalho: Yup abigail noticed that too.

shimonacarvalho: Done.

sandy-wright (approved):
thanks!! 🎉 🚀

PR #4789, opened 2020-09-15
Overview moved to wiki

shimonacarvalho:
## Description

I moved the overview to wiki, so it's easier to keep updated. Removed from readme, but kept a link to redirect reader.


gidjin (approved):
LGTM! I like the new over view page https://github.com/transcom/mymove/wiki/Overview-of-Milmove took a little bit to see but seems like all the info deleted is in that page. I wonder if we should link to that from the code as well, but not enough to block this pr.

PR #4795, opened 2020-09-15
Add L3 segment

garrettqmartin8:
## Description

Adds ability to parse/generate the `L3` EDI segment.

## Reviewer Notes

Do we need to do any more validation around the parameters?

## Setup

`make server_test`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4261) for this change

### DIFF
--- a/pkg/edi/segment/l3.go
+++ b/pkg/edi/segment/l3.go
@@ -8,7 +8,7 @@ import (
 // L3 represents the L3 EDI segment
 type L3 struct {
 	Weight          float64 `validate:"required_with=WeightQualifier"`
-	WeightQualifier string  `validate:"required_with=Weight"`
+	WeightQualifier string  `validate:"required_with=Weight omitempty,eq=B"`
 	PriceCents      int
 }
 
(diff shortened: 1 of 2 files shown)


pearl-truss (approved):
Looks good!

pkg/edi/segment/l3.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"fmt"
+	"strconv"
+)
+
+// L3 represents the L3 EDI segment
+type L3 struct {
+	Weight          float64 `validate:"required_with=WeightQualifier"`
+	WeightQualifier string  `validate:"required_with=Weight"`
+	PriceCents      int
```
reggieriser, line 12: Should we have a `validate` struct tag to make sure we have a non-negative `PriceCents` value?

reggieriser: Might also want some qualifiers on legal `Weight` and `WeightQualifier` values too.

garrettqmartin8: @reggieriser I'll address those 👍

garrettqmartin8: @reggieriser do we actually want to validate that the price isn't negative? @pearl-truss made a good point that fuel surcharge can possibly be negative.

garrettqmartin8: @reggieriser updated!

PR #4803, opened 2020-09-16
Adjust prod API flag settings and change support API default

reggieriser:
## Description

In doing some research for adding a new support API endpoint, I noticed a couple of potential issues:

- the new GovCloud `prd` environment appears to have some APIs turned on -- prime, support, and GHC -- that do not seem to have previously been on in the commercial `prod` environment.
- the `SERVE_SUPPORT_API` flag defaults to `true` unless an environment explicitly turns it off (which is opposite of what we do for all other API flags), potentially leading us to serving that API by accident.  This appears to have been the case even before the GovCloud migration.

## Reviewer Notes

- Am I missing some other place where the flags are set that may be affecting my assumptions?
- Did we intentionally decide to serve the APIs as part of the migration?
- I changed a few `docker-compose` files that would have previously had the support API on via the former `true` default.  I'm less familiar with how these are used, so if the support API should be off in them, let me know.
- I'm not sure what the best way to test this would be -- any suggestions?


config/env/prod.app-client-tls.env
```diff
@@ -26,5 +26,6 @@ MUTUAL_TLS_ENABLED=true
 REDIS_ENABLED=false
 SERVE_DPS=true
 SERVE_ORDERS=true
+SERVE_API_SUPPORT=false
```
shimonacarvalho, line 29: Shouldn't we add SERVE_API_PRIME=false here too? Although I guess I don't know if we have a commercial `prod` environemnt anymore?

reggieriser: I don't think we technically have to since the `SERVE_API_PRIME` defaults to false (it wasn't there before).  But the same could be said for `SERVE_API_SUPPORT` now that I've flipped the default to false for it as well.  But I think it's reasonable to add `SERVE_API_PRIME` here too since it's explicitly set in the `prd` version.

esacteksab (approved):

travelar (approved):
This is interesting and seems like a logical change.  I doubt that infrasec intentionally enabled these, although maybe they were enabled as part of the migration testing.

PR #4806, opened 2020-09-16
[MB-4232] Hide cancel buttons on orders page

Ronolibert:
## Description

Hide cancel buttons on the orders pages

## Setup
`make server_run`
`make client_run`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4232) for this change

## Screenshots

image
image


amitch23 (commented):
I'm a little out of the loop being on call and all, but why hide the cancel button on only the orders page and not all the pages? Also, if there are unit tests for these components, it would be good to add a check that the button is not there. (or e2e tests)

sojeri:
so 1) we do have multiple stories for removing cancel buttons, and 2) the story for this change specifically calls out that the button should be removed from the wizard page component. I'm not sure why Ron has added a new prop to remove the button here instead of removing as described in the Jira story.

Ronolibert:
@sojeri To remove the cancel buttons from the 2 pages of the orders flow, one uses `WizardPage/Form` and the other `WizardPage`

kimallen:
> I'm a little out of the loop being on call and all, but why hide the cancel button on only the orders page and not all the pages? Also, if there are unit tests for these components, it would be good to add a check that the button is not there. (or e2e tests)

I'm wondering about that too- I think during planning (which I know Ron was not at), we said that it was okay to just remove Cancel altogether from the Wizard because we are removing it from all wizard pages.  Either I'm wrong or that didn't get captured in the story.

Ronolibert:
@amitch23 I asked john if its fine to remove it from everything, but in the meantime I added client tests to ensure that the cancel button is missing from these 2 pages

Shkeating (approved):
i think everyone else covered the questions I'd have about just taking it out everywhere with the exception of the edit flow!

Ronolibert:
I had to migrate the Orders and UploadOrders components to the pages directory. Dangerjs detected that I changed a file that we don't lint and blocked the PR. I moved the components but I disabled the linting for the page because there were a lot of lint errors that changed the scope. I made a new card [here](https://dp3.atlassian.net/browse/MB-4416) to handle that work

PR #4809, opened 2020-09-16
Removes healthcheck since DPS went bye bye

esacteksab:
DPS got CNAMED to another entity. But we health check it, which now fails. This removes the health check.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1111,7 +1111,7 @@ jobs:
       - aws_vars_legacy
       - deploy_app_client_tls_steps:
           compare_host: '' # leave blank since we want experimental to be able to roll back
-          health_check_hosts: gex.experimental.move.mil,dps.experimental.move.mil,orders.experimental.move.mil
+          health_check_hosts: gex.experimental.move.mil,orders.experimental.move.mil
           ecr_env: legacy
 
   # `deploy_exp_migrations` deploys migrations to the exp environment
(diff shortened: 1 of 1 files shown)


Ryan-Koch (approved):

kilbergr (commented):
This looks good re: removing the health check from circleci. 
I noticed this healthcheck is also referenced in scripts here: https://github.com/transcom/mymove/blob/master/scripts/deploy-app-client-tls#L21 (3 refs) and here: https://github.com/transcom/mymove/blob/master/scripts/health-tls-check#L21 (3 refs). Shall we remove?
Finally, is this change also necessary for other references to dps.[x.]move.mil such as in config/env/XXX.app-client-tls.env? I'm not familiar enough with the whole set up to know whether that will cause issues.
If I'm heading into the weeds, please let me know :)

esacteksab:
> This looks good re: removing the health check from circleci.
> I noticed this healthcheck is also referenced in scripts here: https://github.com/transcom/mymove/blob/master/scripts/deploy-app-client-tls#L21 (3 refs) and here: https://github.com/transcom/mymove/blob/master/scripts/health-tls-check#L21 (3 refs). Shall we remove?
> Finally, is this change also necessary for other references to dps.[x.]move.mil such as in config/env/XXX.app-client-tls.env? I'm not familiar enough with the whole set up to know whether that will cause issues.
> If I'm heading into the weeds, please let me know :)

You're spot on as always. There's a ticket somewhere to remove all references to DPS as it is no longer used. Sadly, I can't find the ticket, but reach out to Paul.

PR #4815, opened 2020-09-17
MB-4235 Add UpdateMTOAgentHandler

sandy-wright:
## Description

This change adds a stub handler for the `updateMTOAgent` endpoint that returns a 501 "Not Implemented" response code. It also adds a small test that verifies the 501 response. 

*Note:* The MTO Agent updater object will be added to this handler in a later ticket.

## Setup

Follow [these instructions](https://github.com/transcom/prime_api_deliverable/wiki/Using-Postman) to test the endpoint using Postman (or a similar REST API client).

First, run the server:

```sh
make server_run
```

Grab a bogus UUID or two (such as the nil UUID `00000000-0000-0000-0000-000000000000`) to use in the path. Then you may either make up a fake MTO Agent body to test or send an empty body to verify that you get a 501 "Not Implemented" response when you hit this endpoint. 

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4235) for this change.
* [Prime API Documentation](https://github.com/transcom/prime_api_deliverable/wiki/Using-Postman)

### DIFF
--- a/pkg/handlers/primeapi/mto_agent.go
+++ b/pkg/handlers/primeapi/mto_agent.go
@@ -16,7 +16,6 @@ type UpdateMTOAgentHandler struct {
 
 // Handle updates an address on a shipment
 func (h UpdateMTOAgentHandler) Handle(params mtoshipmentops.UpdateMTOAgentParams) middleware.Responder {
-	msg := handlers.NotImplementedErrMessage
 	return mtoshipmentops.NewUpdateMTOAgentNotImplemented().WithPayload(
-		payloads.InternalServerError(&msg, h.GetTraceID()))
+		payloads.NotImplementedError(nil, h.GetTraceID()))
 }
--- a/pkg/handlers/primeapi/mto_agent.go
+++ b/pkg/handlers/primeapi/mto_agent.go
@@ -1,0 +1,22 @@
+package primeapi
+
+import (
+	"github.com/go-openapi/runtime/middleware"
+
+	mtoshipmentops "github.com/transcom/mymove/pkg/gen/primeapi/primeoperations/mto_shipment"
+	"github.com/transcom/mymove/pkg/handlers"
+	"github.com/transcom/mymove/pkg/handlers/primeapi/payloads"
+)
+
+// UpdateMTOAgentHandler is the handler to update an address
+type UpdateMTOAgentHandler struct {
+	handlers.HandlerContext
+	//MTOAgentUpdater services.MTOAgentUpdater TODO
+}
+
+// Handle updates an address on a shipment
+func (h UpdateMTOAgentHandler) Handle(params mtoshipmentops.UpdateMTOAgentParams) middleware.Responder {
+	msg := handlers.NotImplementedErrMessage
+	return mtoshipmentops.NewUpdateMTOAgentNotImplemented().WithPayload(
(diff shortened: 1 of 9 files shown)


pkg/handlers/primeapi/mto_agent.go
```diff
@@ -0,0 +1,22 @@
...
+// UpdateMTOAgentHandler is the handler to update an address
+type UpdateMTOAgentHandler struct {
+	handlers.HandlerContext
+	//MTOAgentUpdater services.MTOAgentUpdater TODO
+}
+
+// Handle updates an address on a shipment
+func (h UpdateMTOAgentHandler) Handle(params mtoshipmentops.UpdateMTOAgentParams) middleware.Responder {
+	msg := handlers.NotImplementedErrMessage
+	return mtoshipmentops.NewUpdateMTOAgentNotImplemented().WithPayload(
```
shimonacarvalho, line 20: Hmm I'm thinking since our payload titles usually match the type of error, we should probably create a new type of payload for Not implemented. Probably just like internalServerError but with better default strings. 

But I'm liking that our intuition was right on being able to make this function without too much dependency!

swagger/prime.yaml
```diff
@@ -1633,6 +1635,10 @@ responses:
     description: A server error occurred.
     schema:
       $ref: '#/definitions/Error'
+  NotImplemented:
+    description: The requested feature is still in development.
```
shimonacarvalho, line 1639: Nice!

pkg/handlers/primeapi/mto_agent_test.go
```diff
@@ -0,0 +1,45 @@
...
+		req := httptest.NewRequest("PUT", fmt.Sprintf("/mto-shipments/%s/agents/%s", agent.MTOShipmentID.String(), agent.ID.String()), nil)
+		params := mtoshipmentops.UpdateMTOAgentParams{
+			HTTPRequest:   req,
+			AgentID:       *handlers.FmtUUID(agent.ID),
+			MtoShipmentID: *handlers.FmtUUID(agent.MTOShipmentID),
+			Body:          payload,
+			IfMatch:       etag.GenerateEtag(agent.UpdatedAt),
+		}
+		// Run swagger validations
+		suite.NoError(params.Body.Validate(strfmt.Default))
```
shimonacarvalho, line 39: Yay!

shimonacarvalho (changes requested):
Looks good - hope the test creation didn't take too long since some of it will go away.

Could you add a payload for NotImplementedError? I think we'll be using it more often. 
(Unless our process changes next week I guess).

pkg/handlers/primeapi/mto_agent.go
sandy-wright: Ooh, this is a good idea; it will be useful to have a `NotImplemented` payload around. And yeah, it really wasn't so bad at all!

sandy-wright:
> Looks good - hope the test creation didn't take too long since some of it will go away.
> 
> Could you add a payload for NotImplementedError? I think we'll be using it more often.
> (Unless our process changes next week I guess).

Done! And the tests weren't bad at all - I figured most of it would be burned down once we link it up with the test data gen code 😄

shimonacarvalho (approved):
LGTM!

PR #4817, opened 2020-09-18
Yaml linting fixes for prime and support yaml, no effective code change

shimonacarvalho:
## Description

Tiny PR. Just cleaning up some issues using a linter for yaml, no effective code changes for these, just comments and yaml.

* According to openApi we should have all tags predeclared at the global level
* Examples for strings should be, well, strings
* There are not supposed to be siblings of the $ref tag according to openapi 
  - In fact after removing them, even those that I assumed would do something, there were no code changes. 
  - The descriptions were not being displayed on redoc either.
* Some of the x-nullable tags were not only $ref siblings but also misspelled.


sandy-wright (approved):
LGTM! Ran it through load testing as well, everything still works the same!

PR #4829, opened 2020-09-18
PrimeAPI Alphabetically sort the definitions

shimonacarvalho:
## Description

We had started with sorted alphabetical definitions and then slowly devolved into unsorted. 
This gets us back to sorted. 
Note this is only the `definitions` in the yaml. Not everything.
I attempted to make ZERO changes in any definitions during this sort but if you have a tool that can check that, would appreciate it.

### DIFF
--- a/swagger/prime.yaml
+++ b/swagger/prime.yaml
@@ -663,27 +663,6 @@ definitions:
       - title
       - detail
       - instance
-  CreatePaymentRequest:
-    type: object
-    properties:
-      isFinal:
-        default: false
-        type: boolean
-      moveTaskOrderID:
-        example: c56a4180-65aa-42ec-a945-5fd21dec0538
-        format: uuid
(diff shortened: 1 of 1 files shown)


swagger/prime.yaml
```diff
@@ -1197,30 +1038,7 @@ definitions:
...
-          pickupPostalCode:
-            type: string
-            format: zip
-            example: "90210"
-            pattern: '^(\d{5}([\-]\d{4})?)$'
-        required:
-          - reServiceCode
-          - reason
-          - pickupPostalCode
+
```
sandy-wright, line 1041: Oh, we could probably get rid of this empty line!

sandy-wright (approved):
LGTM, load tested and all! That diff sure does look goofy, but goland makes sense of the history somehow

PR #4832, opened 2020-09-18
Kim MB-3733 home routing from milmove

kimallen:
## Description

This PR continues the transition to using the new home page and combines two stories, as they ended up being dependent on each other.  When a customer logs in or logs out at any stage, they should return to the (new) home page OR where they left off in the profile flow if it was not completed.

The user should also be able to click on the "Milmove" icon in the upper left at any point and go to the Home page (except they should stay where they are if in the profile flow as current behavior).

Additional adjustments:  
- Previously if the user had completed the orders info but not uploaded, the home page "Add orders" button took them to the first page.  It now takes them to the unfinished upload page.
- The "Add a shipment" button on home page is now disabled after the move is submitted.

## Reviewer Notes

~Currently in draft PR- still need to update tests~ done

## Setup

Log in as a new user.  Work through the flow for setting up an HHG.  On every page, try the Milmove logo.  Return to where you left off, and trying logging in and logging out.  You should be taken to the home page in every stage except if you are within the profile setup.

Do the same for creating a PPM.  Note that there is currently not a link to the PPM flow on the Home page, so to return, go to `/ppm` through the url.

## References

* [Milmove Logo story](https://dp3.atlassian.net/browse/MB-3733) and [Login/logout story](https://dp3.atlassian.net/browse/MB-3732) for this change


src/pages/MyMove/Home/index.jsx
```diff
@@ -166,99 +275,145 @@ class Home extends Component {
...
+                  headerText="Shipment selection"
+                  secondaryBtn={this.hasShipment}
+                  secondaryClassName="margin-top-2"
+                  step="3"
+                >
+                  {this.hasShipment ? (
+                    <ShipmentList shipments={shipments} onShipmentClick={this.handleShipmentClick} />
+                  ) : (
+                    <Description>
+                      Tell us where you&apos;re going and when you want to get there. We&apos;ll help you set up
```
sojeri, line 379: looks like this was already the case here, but I'm curious why we're using `&apos;` here but plain `'` in many other files. do you know the context off the top of your head? :) pls ignore if you'll have to go digging

src/pages/MyMove/Home/index.jsx
```diff
@@ -166,99 +275,145 @@ class Home extends Component {
...
+                  containerClassName="margin-bottom-8"
+                  headerText="Confirm move request"
+                  completedHeaderText="Move request confirmed"
+                  onActionBtnClick={() => this.handleNewPathClick(confirmationPath)}
+                  step="4"
+                >
+                  <p className={styles.description}>
+                    {this.hasSubmittedMove
+                      ? 'Move submitted.'
+                      : 'Review your move details and sign the legal paperwork, then send the info on to your move counselor.'}
```
sojeri, line 397: also pre-existing so fine to ignore if you need to research:
- why do we need this string twice? (here & line 401/402)

sojeri (approved):
I don't have any comments on the overall change. it looks good to me so far aside from the behaviors we discussed offline which you've already fixed. TY for fixing on Friday! :)

I did leave a couple questions about code that was in place before this chance. if you know the context, I'd love to hear it. if not, please ignore them.

approving w/ the assumption that you'll fix tests before merge. at me or DM in slack if you need a rereview!

src/pages/MyMove/Home/index.jsx
```diff
@@ -497,6 +497,7 @@ const mapStateToProps = (state) => {
     orders: selectActiveOrLatestOrdersFromEntities(state),
     uploadedOrderDocuments: selectUploadedOrders(state),
     serviceMember,
+    backupContacts: serviceMember.backup_contacts || state.serviceMember.currentBackupContacts || [],
```
sojeri, line 500: nice! :)

src/pages/MyMove/Home/index.jsx
kimallen: I got a linter complaint- it previously just had `'`.  I don't know why it decided to start complaining now.

src/pages/MyMove/Home/index.jsx
kimallen: Hmm, strange.  We don't need that twice.   I can fix that.

src/pages/MyMove/Home/index.jsx
sojeri: ah, maybe we have new rules or dependabot timing hit us.

amanda5iu (approved):

PR #4839, opened 2020-09-21
Update make bin/* commands to depend on source

gidjin:
## Description

We have short cuts in our Makefile to create binary tools. The configuration for these tools is such that if the tool is built it will not be rebuilt. However, Make won't update the binary even after its code is updated. This is because we do not reference the source files in the make configuration. I've updated the Makefile to refer to the command line code directories and now Make will detect code changes. This might not detect changes to other packages that the commands depend on but it's a start in correcting our Makefile configuration.

## Reviewer Notes

None

## Setup

Pick any of the `make bin/*` commands, run the make, update the cli files in some way, run make again and it should rebuild.

For example:
```sh
make bin/generate-deploy-notes
make: `bin/generate-deploy-notes' is up to date.
// update cmd/generate-deploy-notes/main.go file
make bin/generate-deploy-notes
go build -ldflags "" -o bin/generate-deploy-notes ./cmd/generate-deploy-notes
```


reggieriser (commented):
Thanks for noticing and handling this!  I did see that when we have source in subdirectories as is the case for `prime-api-client` and `webhook-client` that editing a file in the subdirectory does *not* cause it to rebuild, but editing a file in the main directory does.  I suppose we could list all the subdirectories as dependencies too, but there may be a more general way.

I also noticed we're inconsistent with the directory naming for these `cmd` directories -- most of the time we use dashes, but a few times we use underscores.  That's probably outside the scope of this PR, but just noting it.

reggieriser (approved):
I'm going to go ahead and approve since this helps in many scenarios.  If refining to handle the cases I was talking about is easy, let's add it.  But don't want to hold it up for that. 🚢

garrettqmartin8 (approved):

gidjin:
> Thanks for noticing and handling this! I did see that when we have source in subdirectories as is the case for `prime-api-client` and `webhook-client` that editing a file in the subdirectory does _not_ cause it to rebuild, but editing a file in the main directory does. I suppose we could list all the subdirectories as dependencies too, but there may be a more general way.

Yeah I think we would have to list each sub directory that the cmd depends on. I think it'd be good to do, but was beyond the time I had for this PR.

> I also noticed we're inconsistent with the directory naming for these `cmd` directories -- most of the time we use dashes, but a few times we use underscores. That's probably outside the scope of this PR, but just noting it.

I noticed this too. I'd change it now but there would be a lot of announcing and docs to update hence I'm delaying it until another time.

PR #4840, opened 2020-09-21
Primeapi: Linter code fixes to prime.yaml

shimonacarvalho:
## Description

Cleaned up issues pointed out by the yaml linter

* `UpdatePaymentRequestStatus` is not a definition used in the primeapi - deleted it
* `createdAt` and `updatedAt` are readOnly values
* According to OpenAPI 2.0 there should be a defined set of global tags, added these and noticed some cleanup that could be done
* All other tag names are singular not plural so `paymentRequests` --> `paymentRequest`
* createUpload is specifically about uploading docs to payment requests so changed tag from `upload` --> `paymentRequest`

## Reviewer notes
There should be no change in behavior (other than not being able to send in createdAt, updatedAt)
However the tag name changes changed some package names, so there's cleanup of the imports.
Most changes are related to create payment request and uploads so testing those would be great.


sandy-wright (approved):
LGTM! Tested both of the payment request endpoints and ran load testing against it 👍

PR #4841, opened 2020-09-22
Display PPM in shipment list on home page

amitch23:
## Description

This adds the PPM card to the shipment list in step 3 on the Home page.  All shipments are ordered by `createdAt`.  If there is only one of a shipment type, it does not have a number next to it.  If there is more than one, they are numbered.

## Reviewer Notes

TODO:
* Verify handling of "shipmentType" on ppm (rather, the lack of that field...) - should this be a migration on the PPM table? Or should it be added at the PPM reducer level? --- **KA response:** I ended up making a copy of the ppm and applying the shipment type.  I don't think it's worth adding in db since all PPMs are always PPM shipment types.  Reducer level might be an ok option, but really it's just need to label the card, and we could also just check for no shipment type.

~* Jest component test verifying that the PPM shows in the step 3 shipment list~ DONE
* ~Double check that step 3 shipment list is correctly displaying on mobile and desktop views~ fixed by Shauna
* ~Design QA with Mariaclara to confirm- [request for some changes](https://ustcdp3.slack.com/archives/CTNTFJSBA/p1601064214059500)~  reviewed by Shauna

## Setup

Create multiple HHGs and a PPM.  Make sure they show up in created order and follow the numbering plan.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4316) for this change

## Screenshots

browser view
image

Phone view
image


src/pages/MyMove/Home/index.jsx
```diff
@@ -298,6 +305,22 @@ class Home extends Component {
     );
   };
 
+  sortAllShipments = (mtoShipments, currentPpm) => {
+    const allShipments = JSON.parse(JSON.stringify(mtoShipments));
```
kimallen, line 309: Does this need to be a deep copy? I have a feeling no since we aren't using any of the nested data for this page, but wasn't sure

src/pages/MyMove/Home/index.jsx
```diff
@@ -298,6 +305,22 @@ class Home extends Component {
...
+    if (Object.keys(currentPpm).length) {
+      const ppm = JSON.parse(JSON.stringify(currentPpm));
```
kimallen, line 311: Same question here about deep copy

Shkeating (approved):
I made some css changes that will not affect the layout if our ID #s are short, but should look alright if they are long.

Shkeating:
here's what it's looking like now!
image

kimallen:
> here's what it's looking like now!
> image

Great- thanks for working on it!

src/pages/MyMove/Home/index.jsx
Ronolibert: I think we do need a copy since `sort()` will end up mutating the original array but I don't think the copy has to be deep since we're not altering any of the values

src/pages/MyMove/Home/index.jsx
Ronolibert: I think it makes sense to have a deep copy here since we're changing/overwriting values

src/components/Customer/Home/ShipmentList/index.jsx
```diff
@@ -36,20 +37,49 @@ ShipmentListItem.propTypes = {
...
-    ))}
-  </div>
-);
+ShipmentListItem.defaultProps = {
+  showNumber: true,
+};
+
+const ShipmentList = ({ shipments, onShipmentClick }) => {
+  const shipmentNumbersByType = {};
+  const shipmentCountByType = {};
```
Ronolibert, line 49: I think we could simplify this logic because there will only ever be 1 PPM at most. We could declare a variable here for the shipment number starting at 0 then in the render method we could increment that number when we come across a `shipmentType` of `HHG` and pass that number into the shipment list item

kimallen: The reason I did it this way is that we could also have more than one of other types of shipments, and at some point in the future (post March 1), I believe we will support >1 PPM.

Ronolibert (approved):
Looks good! Works as expected

src/components/Customer/Home/ShipmentList/index.jsx
```diff
@@ -36,20 +39,49 @@ ShipmentListItem.propTypes = {
...
+  shipments.map((shipment) => {
```
suzubara, line 52: `Array.map` is used with a return value (i.e., to map array1 to array2). Here if you just want to loop through each item, you should use `Array.forEach`.

suzubara (approved):

src/pages/MyMove/Home/index.jsx
kimallen: I did add a value that didn't exist before (`ppm.shipmentType = SHIPMENT_OPTIONS.PPM;`).  Does that count as altering?

kimallen: I just tried using a shallow copy with `{...mtoShipments}` but for some reason it was erroring (allShipments was empty), so just left it for now.

PR #4846, opened 2020-09-22
MB-3938 Adds event trigger for CreateMTOServiceItem in GHC API

abbyoung:
## Description

* Creates an event when an mtoServiceItem is created in the ghc api
* Adds test to make sure the event and notification are getting created correctly

## Reviewer Notes

The endpoint itself is currently broken, so you can't manually test the service item creation.

## Setup

Tests:

```
$ make db_test_e2e_populate
$ go test ./pkg/handlers/ghcapi --testify.m "TestCreateMTOServiceItemHandler" -v

```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3938) for this change


pkg/handlers/ghcapi/mto_service_items.go
```diff
@@ -104,6 +105,18 @@ func (h CreateMTOServiceItemHandler) Handle(params mtoserviceitemop.CreateMTOSer
 		return mtoserviceitemop.NewCreateMTOServiceItemInternalServerError()
 	}
 
+	_, err = event.TriggerEvent(event.Event{
+		EventKey:        event.MTOServiceItemCreateEventKey,
+		MtoID:           (*createdServiceItems)[0].MoveTaskOrderID,
```
shimonacarvalho, line 110: I noticed there's a [0] index here. Does this endpoint only ever create one service item? If it creates multiple we may have to trigger the event multiple times. So that we send a notification to prime for each service item.

abbyoung: Confirmed with Donald: Only one mtoServiceItem is returned. Changes Roci made to the service object returning an array is why this handler now has an array of one to return.

shimonacarvalho (approved):
LGTM!

PR #4847, opened 2020-09-22
Supportapi: Linter code fixes to support.yaml

shimonacarvalho:
## Description

Cleaned up issues pointed out by the yaml linter

* `MoveOrders` is not a definition used in the supportapi (it's an array of move orders) - deleted it
* `UpdatePaymentRequest` is not used - deleted it
* Note `MTOAgents` is not used but was not deleted - we *should* be adding it to getMTO and listMTO so I left it in.
* `createdAt` and `updatedAt` are readOnly values
* According to OpenAPI 2.0 there should be a defined set of global tags, added these and noticed some cleanup that could be done
* All other tag names are singular not plural so `paymentRequests` --> `paymentRequest`

## Reviewer notes
There should be no change in behavior (other than not being able to send in createdAt, updatedAt)

However the tag name changes changed some package names, so there's cleanup of the imports.

Most changes are related to payment request and so testing updatePaymentRequestStatus would be great.


sandy-wright (approved):
🚀 ✨

PR #4852, opened 2020-09-23
[MB-4333] Create prime-api-client command and support API endpoint for EDI generation

reggieriser:
## Description

This PR adds a support API endpoint for viewing the EDI for a given payment request, along with a `prime-api-client` command (`support-get-payment-request-edi`) to invoke it.

## Reviewer Notes

- I refactored the EDI generator a little so that it exposes an interface similar to other service objects.  This allowed me to generate a mock for testing purposes.
- The EDI is returned as a JSON string in the returned payload.  Since there doesn't appear to be a way to have line breaks in a JSON string, you will see `\n` where there normally is a line break (although I show a way to below to have `jq` show it in "raw" mode).
- To avoid confusingly formatted endpoint URLs, I used the payment request *ID* as the identifier for the EDI to match what we do with other support API endpoints.  Our command line utility uses payment request *number*, however.  We may want to get them in sync at some point.

## Setup

1. `make db_dev_e2e_populate`
1. `make server_run`
1. Save the following payload in `create_payment_request.json`:
    ```
    {
      "body": {
        "isFinal": false,
        "serviceItems": [
          {
            "id": "722a6f4e-b438-4655-88c7-51609056550d"
          },
          {
          },
          {
          },
          {
          }
        ]
      }
    }
    ```
1. Create the payment request using the file above:
    ```
    go run ./cmd/prime-api-client --insecure create-payment-request --filename create_payment_request.json | jq '.id,.paymentRequestNumber'
    ```
    This will show two lines -- the first has the payment request ID and the second has the payment request number.
1. Save the following payload to `get_payment_request_edi.json`, replacing `<paymentRequestID>` with the ID that was output from the previous step:
    ```
    {
      "paymentRequestID": "<paymentRequestID>"
    }
    ```
1. Get the EDI for that payment request:
    ```
    go run ./cmd/prime-api-client --insecure support-get-payment-request-edi --filename get_payment_request_edi.json | jq
    ```
    Given that JSON can't contain embedded newlines, we can use `jq` to help us show it in its raw form:
    ```
    go run ./cmd/prime-api-client --insecure support-get-payment-request-edi --filename get_payment_request_edi.json | jq -r '.edi'
    ```
1. Locally, you can also test this against the EDI returned by our `generate-payment-request-edi` command below.  Replace the `<paymentRequestNumber>` with the payment request number output earlier (note that it's the *number* in this case not the *ID*).
    ```
    go run ./cmd/generate-payment-request-edi --payment-request-number <paymentRequestNumber>
    ```
    The EDI should match (except possibly some timestamps in the first few lines).

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4333) for this change
* [Acceptance Testing for Invoicing EDI](https://docs.google.com/document/d/1zC04O4WB4qgV2TqqWnlwRpfMy6WQVY3rxlY2bUiG65o/edit) design document


swagger/support.yaml
```diff
@@ -467,6 +467,37 @@ paths:
...
+    parameters:
+      - description: UUID of the payment request for which EDI should be generated.
+        format: uuid
+        in: path
+        name: paymentRequestID
+        required: true
+        type: string
+    get:
+      summary: getPaymentRequestEDI
+      description: Returns the EDI corresponding to the payment request identified by the given payment request number.
```
shimonacarvalho, line 480: `  480:20  warning  endpoint-description-length  The endpoint description should be at least 128 characters.`

We'd like a longer description of what the endpoint does. Potential additions
- What does EDI stand for
- Useful to point out any requirements that are more involved than what is listed as required in the Redoc (like payment request must be in approved state to retrieve edi -- I made that up)

swagger/support.yaml
```diff
@@ -1206,6 +1237,17 @@ definitions:
       - PAID
     title: Payment Request Status
     type: string
+  PaymentRequestEDI:
```
shimonacarvalho, line 1240: ` 1231:24    error  definitions-sort             Definitions MUST be sorted alphabetically`

We're keeping the definitions alphabetical. I think this one prob needs to move up.

swagger/support.yaml
```diff
@@ -1206,6 +1237,17 @@ definitions:
...
+    properties:
+      id:
+        example: c56a4180-65aa-42ec-a945-5fd21dec0538
+        format: uuid
+        type: string
+        readOnly: true
+      EDI:
```
shimonacarvalho, line 1247: ` 1247:11    error  body-properties-case         Body properties MUST be camelCase`

So since ID is `id`, EDI should probably be `edi`

shimonacarvalho (changes requested):
Ran the linter and found some issues. I've inserted them at the line they were found. 

Naturally they're a bit nitpicky as that's what linters like to be, but overall we hope they will keep our yamls up to a standard and consistent coding style.

pkg/handlers/supportapi/payment_request.go
```diff
@@ -165,3 +166,45 @@ func (h ListMTOPaymentRequestsHandler) Handle(params paymentrequestop.ListMTOPay
...
+}
+
+// Handle getting the EDI for a given payment request
+func (h GetPaymentRequestEDIHandler) Handle(params paymentrequestop.GetPaymentRequestEDIParams) middleware.Responder {
+	logger := h.LoggerFromRequest(params.HTTPRequest)
+
+	paymentRequestID, err := uuid.FromString(params.PaymentRequestID.String())
+	if err != nil {
+		logger.Error(fmt.Sprintf("Error parsing payment request id: %s", params.PaymentRequestID.String()), zap.Error(err))
+		return paymentrequestop.NewGetPaymentRequestEDIInternalServerError().WithPayload(payloads.InternalServerError(handlers.FmtString(err.Error()), h.GetTraceID()))
```
shimonacarvalho, line 184: Can someone trigger this just by sending a bad uuid? If so then we should return a 422 Unprocessable Entity. If we really think we should never hit this code (uuid validation is performed elsewhere) then Internal Server Error is valid. 
This is our strategy: https://github.com/transcom/mymove/wiki/API-Errors#http-status-codes

pkg/handlers/supportapi/payment_request.go
```diff
@@ -165,3 +166,45 @@ func (h ListMTOPaymentRequestsHandler) Handle(params paymentrequestop.ListMTOPay
...
+		return paymentrequestop.NewGetPaymentRequestEDINotFound().WithPayload(payloads.ClientError(handlers.NotFoundMessage, msg, h.GetTraceID()))
+	}
+
+	var payload supportmessages.PaymentRequestEDI
+	payload.ID = *handlers.FmtUUID(paymentRequestID)
+
+	edi858c, err := h.GHCPaymentRequestInvoiceGenerator.Generate(paymentRequest, false)
+	if err != nil {
+		logger.Error(fmt.Sprintf("Error generating EDI segments for payment request ID: %s: %s", paymentRequestID, err))
+		return paymentrequestop.NewGetPaymentRequestEDIInternalServerError().WithPayload(payloads.InternalServerError(handlers.FmtString(err.Error()), h.GetTraceID()))
```
shimonacarvalho, line 200: I guess the same is true here about wrapping all kinds of errors as Internal Server Errors. What we usually do is something like this (the case statement at the bottom) 
https://github.com/transcom/mymove/wiki/API-Errors#constructing-errors-and-responses

But that said, if this is an endpoint that will never move to production, laxer rules may be okay.

pkg/handlers/supportapi/payment_request_test.go
```diff
@@ -230,3 +235,128 @@ func (suite *HandlerSuite) TestListMTOPaymentRequestHandler() {
...
+		req := httptest.NewRequest("GET", fmt.Sprintf(urlFormat, invalidID), nil)
+
+		params := paymentrequestop.GetPaymentRequestEDIParams{
+			HTTPRequest:      req,
+			PaymentRequestID: strfmt.UUID(invalidID),
+		}
+
+		response := handler.Handle(params)
+
+		suite.IsType(paymentrequestop.NewGetPaymentRequestEDIInternalServerError(), response)
```
shimonacarvalho, line 319: So a bad script could hit staging causing 500s and resulting in an AWS on-call alert? Would be good to avoid that if possible. If you think swagger validation should catch this, you could call this on the params object before you call handler: 
```
		// Run swagger validations
```
And check the resulting error. If swagger returns a non-500, that's what would happen in staging and that would be okay - no on-call.

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -9,26 +9,33 @@ import (
...
+	}
 }
 
 const dateFormat = "20060102"
 const timeFormat = "1504"
 
 // Generate method takes a payment request and returns an Invoice858C
-func (g GHCPaymentRequestInvoiceGenerator) Generate(paymentRequest models.PaymentRequest, sendProductionInvoice bool) (ediinvoice.Invoice858C, error) {
+func (g ghcPaymentRequestInvoiceGenerator) Generate(paymentRequest models.PaymentRequest, sendProductionInvoice bool) (ediinvoice.Invoice858C, error) {
 	// TODO: seems ReferenceID is a *string but cannot be saved as nil, do we need to validate it's not nil here
```
shimonacarvalho, line 34: I know this wasn't your change, but is this comment still valid? I can't see any reference to ReferenceID here.

swagger/support.yaml
reggieriser: Fixed.

swagger/support.yaml
reggieriser: Fixed (hopefully) -- depends on how the sorting algorithm treats upper vs. lower case.

swagger/support.yaml
reggieriser: Fixed.  Note that this makes the generated Go code use `Edi` which looks a little odd.  There must be special handling for `id` to turn it into `ID`.

pkg/handlers/supportapi/payment_request.go
reggieriser: I had tested this out and our generated swagger code was catching bad UUIDs and returning a 422 before it got to the handler.  Another endpoint in the same file was handling it this same way, but I just looked at some other support API files and we sometimes use `FromStringOrNil` instead which just returns the "nil" UUID if this fails (which it shouldn't unless something strange is happening).  I think I'll just use that instead.

pkg/handlers/supportapi/payment_request_test.go
reggieriser: There's no `Body` with this endpoint, but I've changed this so that even if swagger validation didn't catch the problem, this now should return a non-500 error.  That seems to be how other support endpoints are handling the same situation.

pkg/handlers/supportapi/payment_request.go
reggieriser: I think we'll need to open a separate story to reconsider errors in `Generate` because it looks like there are certainly some things that should raise a 500 but it's possible some others might be more "normal" (cc: @gidjin ).  I don't think the calling code in the handler can discern that now since most error returns use a generic error rather than a specific error type (like your linked example).  Since this is a support endpoint only, we're hopefully OK to proceed for now with the ISE.

pkg/services/invoice/ghc_payment_request_invoice_generator.go
reggieriser: I'll defer to @gidjin since he wrote the comment, but I do see it being used on line 119: https://github.com/transcom/mymove/blob/7aa7d7549701f771739fbdb753a360ed24c1cf72/pkg/services/invoice/ghc_payment_request_invoice_generator.go#L119

jacquelineIO: @shimonacarvalho we're going to write a story to address this TODO

jacquelineIO (approved):
lgtm. Appreciate your review on how to handle the errors @shimonacarvalho

pkg/handlers/supportapi/payment_request.go
reggieriser: FYI, separate story for this has been created: https://dp3.atlassian.net/browse/MB-4422

shimonacarvalho (approved):
Looks good, thanks reggie.

PR #4654, opened 2020-08-20
MB-3687 Return an array of service items from CreateMTOServiceItem

sandy-wright:
## Description

This PR changes the Prime API to return an array of service items from the `createMTOServiceItem` endpoint. It also updates the service object `CreateMTOServiceItem` to return an array as well. The GHC API handler for their version of the endpoint was also updated to return the first item from this array.

## Reviewer Notes

Beyond the server tests, I wasn't able to test the GHC endpoint. So if anyone knows how to do that, please let me know! 

## Setup

[Prime API Documentation](https://github.com/transcom/prime_api_deliverable/wiki/Getting-Started)
[`CreateMTOServiceItem` Endpoint Documentation](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/transcom/mymove/sw-mb-3687-return-service-item-list/swagger/prime.yaml#operation/createMTOServiceItem)

Run the server:

```sh
make db_dev_e2e_populate && make server_run
```

Then test the `createMTOServiceItem` endpoint for the Prime API. For a successful 200 response, you should see a JSON array being returned (although it should still only contain one service item). 

## Code Review Verification Steps

* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3687) for this change.

PR #4859, opened 2020-09-25
MB-3616 TOO should only be able to add Mgmt & Counseling service items once

donaldthai:
## Description

> If the TOO has approved a shipment + Mgmt/Counseling service items, they should not be able to add them again.

This PR refactors the UI and backend service so the user can reject basic service items. This takes place during the time the TOO is approving shipments and basic service items.

## Reviewer Notes

- Halved the padding for the service item codes section. See comments for details.

## Setup

```sh
make clean db_dev_e2e_populate
make server_run
make office_client_run
```

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3616) for this change

## Screenshots

Oct-01-2020 12-28-05


pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -54,15 +54,21 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
...
+	}
 
 	// We can have two service items that come in from a MTO approval that do not have an MTOShipmentID
 	// they are MTO level service items. This should capture that and create them accordingly, they are thankfully
 	// also rather basic.
 	if serviceItem.MTOShipmentID == nil {
-		if serviceItem.ReService.Code == models.ReServiceCodeMS || serviceItem.ReService.Code == models.ReServiceCodeCS {
-			serviceItem.Status = "APPROVED"
-		}
 		verrs, err = o.builder.CreateOne(serviceItem)
```
duncan-truss, line 72: Is there a reason CreateMTOServiceItem returns an array of service items if we are only ever creating just one?

pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -54,15 +54,21 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
 	}
 	// set re service for service item
 	serviceItem.ReServiceID = reService.ID
-	serviceItem.Status = models.MTOServiceItemStatusSubmitted
+	// for basic service items
+	if serviceItem.ReService.Code == models.ReServiceCodeMS || serviceItem.ReService.Code == models.ReServiceCodeCS {
+		// set status as APPROVED if status is not set
+		if serviceItem.Status == "" {
```
duncan-truss, line 60: Do we still need this conditional if the api requires a status?  Is this getting called somewhere else for basic service items?

pkg/services/mto_service_item/mto_service_item_creator.go
donaldthai: @sandy-wright made some changes to the service object to return an array which will support future work to create additional service items I believe. Here's the PR that did the changes: https://github.com/transcom/mymove/pull/4654/files

pkg/services/mto_service_item/mto_service_item_creator.go
donaldthai: That's a good point. The creator is [being used in other places like the Prime api](https://github.com/transcom/mymove/blob/master/pkg/handlers/primeapi/mto_service_item.go#L73) but maybe it would be better to explicitly pass in the status and just make sure the status is being passed back. I think we already have model validation before we create the service item.

pkg/services/mto_service_item/mto_service_item_creator.go
duncan-truss: Sounds good thanks for digging that back up

src/constants/mtoServiceItemStatusTypes.js
```diff
@@ -0,0 +1,4 @@
+export default {
+  APPROVED: 'APPROVED',
+  REJECTED: 'REJECTED',
+};
```
duncan-truss, line 4: I think these are already defined under SERVICE_ITEM_STATUS in the `shared/constants` folder

donaldthai: ah, I just didn't know. I can just use the one in the shared/contants.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -53,15 +54,39 @@ func (o moveTaskOrderUpdater) MakeAvailableToPrime(moveTaskOrderID uuid.UUID, eT
...
+
+		if index < len(mto.MTOServiceItems) && mto.MTOServiceItems[index].ReService.Code == code {
+			// x is present at data[i]
+			return true
+		}
+
+		// x is not present in data, but index is the index where it would be inserted.
+		return false
+	}
+
```
donaldthai, line 74: Didn't want to just do a double for loop to search for existing service items. Sorts the service items by code ascending order and the searching algorithm uses a binary search (assuming the collection is sorted).

pkg/handlers/ghcapi/move_task_order.go
```diff
@@ -54,17 +54,16 @@ type UpdateMoveTaskOrderStatusHandlerFunc struct {
...
-			serviceItemCodes = append(serviceItemCodes, models.ReServiceCodeMS)
 
+	// include mto level service items to create if requested
+	serviceItemCodes := make(map[models.ReServiceCode]bool)
+	if params.ServiceItemCodes != nil {
+		if params.ServiceItemCodes.ServiceCodeCS {
+			serviceItemCodes[models.ReServiceCodeCS] = true
+		}
+		if params.ServiceItemCodes.ServiceCodeMS {
+			serviceItemCodes[models.ReServiceCodeMS] = true
```
donaldthai, line 65: I felt like this took away the ambiguity of what service item codes are passed in. LMK if this doesn't work.

src/components/Office/RequestedShipments/RequestedShipments.module.scss
```diff
@@ -18,7 +18,7 @@
   }
 
   .serviceItems {
-    @include u-padding-top(4);
+    @include u-padding-top(2);
```
donaldthai, line 21: @transcom/truss-design I halved the padding since it looked a bit off to me. Let me know if that doesn't work.

donaldthai: I think you should be able to see the difference in happo now.

duncan-truss:
@donaldthai @brooksr4 If I'm looking at this right we are not displaying or saving to the database any basic service item that was rejected anymore?

donaldthai:
> @donaldthai @brooksr4 If I'm looking at this right we are not displaying or saving to the database any basic service item that was rejected anymore?

Nope, we're creating and displaying the ones that are selected for the first pass. On a second pass, users can only approve a shipment.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -53,15 +54,39 @@ func (o moveTaskOrderUpdater) MakeAvailableToPrime(moveTaskOrderID uuid.UUID, eT
 		}
 	}
 
+	// sort by service code first, asc
+	sort.Slice(mto.MTOServiceItems, func(i int, j int) bool {
```
duncan-truss, line 58: This is MTOServiceItems array is going to include all of the accessorial service items (mainly) that are attached to a previous shipment not just any basic service items.  I feel like sorting all of those just to find if any basic service items exists is a lot of work, where maybe we can make some assumptions to simplify instead?  Or is there something changing in the future I don't know about?

Assumptions I think we know about MakeAvailableToPrime:
- Only basic services items (not accessorial) will ever be created (identifiable by their code or because they are special and have no shipment id associated)
- Basic service items will only ever be created the first time a Move Task Order becomes available to the prime.  After that they cannot be added when subsequent shipments are approved.
- No service items exist the first time a Move Task Order is being made available to the prime.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -53,15 +54,39 @@ func (o moveTaskOrderUpdater) MakeAvailableToPrime(moveTaskOrderID uuid.UUID, eT
...
+		return mto.MTOServiceItems[i].ReService.Code < mto.MTOServiceItems[j].ReService.Code
+	})
+	serviceItemExists := func(code models.ReServiceCode) bool {
+		index := sort.Search(len(mto.MTOServiceItems), func(i int) bool {
```
duncan-truss, line 62: I forgot how terrible searching was in Go 🤣

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -29,7 +30,7 @@ func NewMoveTaskOrderUpdater(db *pop.Connection, builder UpdateMoveTaskOrderQuer
 
 //MakeAvailableToPrime updates the status of a MoveTaskOrder for a given UUID to make it available to prime
 func (o moveTaskOrderUpdater) MakeAvailableToPrime(moveTaskOrderID uuid.UUID, eTag string,
-	mtoApprovalServiceItems *[]models.ReServiceCode) (*models.Move, error) {
+	mtoApprovalServiceItems *map[models.ReServiceCode]bool) (*models.Move, error) {
```
duncan-truss, line 33: I only see us iterating through this map, do you think it make sense to keep it a map if we're not doing lookups or ever getting a false bool value?

pkg/services/move_task_order/move_task_order_updater.go
donaldthai: > I feel like sorting all of those just to find if any basic service items exists is a lot of work, where maybe we can make some assumptions to simplify instead? Or is there something changing in the future I don't know about?

Hmm that's a good point. I definitely also think that we can just use `available_to_prime_at` column to determine creating the service items or not. Thanks, I'll refactor.

pkg/services/move_task_order/move_task_order_updater.go
donaldthai: I think if the user "intentionally" passes in a false value then we could encounter that. Otherwise, maybe instead of iterating, I could just pass in the bool values in the params?

donaldthai: Cool, just refactored to pass in bool values in the param.

pkg/services/move_task_order/move_task_order_updater.go
donaldthai: Alright, we're only creating service items only if the `available_to_prime_at` is nil. Much simpler for sure.

duncan-truss: Thanks Donald, the changes I think reduced the complexity a lot. I saw that we are still calling update on the move task order, does that do anything if `available_to_prime_at` already has a value?

src/components/Office/RequestedShipments/RequestedShipments.jsx
```diff
@@ -45,24 +45,30 @@ const RequestedShipments = ({
...
-        approveMTO(moveTaskOrder.id, moveTaskOrder.eTag, mtoApprovalServiceItemCodes),
-      ])
+      ];
+      const mtoApprovalServiceItemCodes = {};
+      if (values.shipmentManagementFee) {
+        mtoApprovalServiceItemCodes.serviceCodeMS = true;
+      }
+      if (values.counselingFee) {
+        mtoApprovalServiceItemCodes.serviceCodeCS = true;
+      }
```
duncan-truss, line 61: I think this can be one line now if we're being nitpicky:
```
const mtoApprovalServiceItemCodes = {
    serviceCodeMS: values.shipmentManagementFee,
    serviceCodeCS: values.counselingFee,
}
```

pkg/services/move_task_order/move_task_order_updater.go
duncan-truss: 👍 I see the conditional check in the form onSubmit now that doesn't even call approveMTO if it's already been set.

duncan-truss (approved):
👍

src/components/Office/RequestedShipments/RequestedShipments.jsx
```diff
@@ -45,24 +45,30 @@ const RequestedShipments = ({
...
+      // if mto is not yet approved, add request to approve it
+      if (!moveTaskOrder.availableToPrimeAt) {
+        requests.push(approveMTO(moveTaskOrder.id, moveTaskOrder.eTag, mtoApprovalServiceItemCodes));
+      }
+
+      Promise.all(requests)
         .then((results) => {
           if (
-            results[1].response.status === 200 &&
+            results[1]?.response?.status === 200 ||
```
suzubara, line 71: why change this to `||`? Is there a scenario where one request can fail and it still is considered successful?

src/components/Office/RequestedShipments/RequestedShipments.jsx
```diff
@@ -128,21 +138,25 @@ const RequestedShipments = ({
...
-                  onChange={formik.handleChange}
-                />
-                <Checkbox
-                  id="counselingFee"
-                  label="Counseling fee"
-                  name="counselingFee"
-                  onChange={formik.handleChange}
-                />
-              </Fieldset>
+              {mtoServiceItems?.length <= 0 && (
```
suzubara, line 141: To clarify this logic, we should show these checkboxes if there are _no_ `mtoServiceItems`? In that case this might be better as:
```
!mtoServiceItems?.length && (...
```
that will eval to `true` if `mtoServiceItems` is undefined or if the length is 0. as-is, it will eval to `false` if mtoServiceItems is undefined, which I don't think is what we want..

brooksr4 (changes requested):
Everything looks good, except the border-line for the service items should go all the way to the right.

donaldthai:
> Everything looks good, except the border-line for the service items should go all the way to the right.

@brooksr4 ok, just made the change to do full width for that service item area. Should be viewable once Happo is done generating differences.

for reference:
image

src/components/Office/RequestedShipments/RequestedShipments.jsx
donaldthai: oh that was a quick fix to refresh the page since the request to update the MTO status may not be included. Let me clean up the logic to make it clearer though.

src/components/Office/RequestedShipments/RequestedShipments.jsx
donaldthai: Squirrelly logic here...I wanted to show the button based on two cases:
1. show button if don't have any mto service items
2. show button if we haven't sent to prime

Probably easier to just check if we already sent to the prime or not. LOL. I'll refactor.

src/components/Office/RequestedShipments/RequestedShipments.jsx
donaldthai: Nice catch, I'll use that.

brooksr4 (approved):
looks good now

src/components/Office/RequestedShipments/RequestedShipments.jsx
donaldthai: Okay, I moved the `results[1]?.response?.status === 200` logic inside the `if` and now we only need to reload the page if we're setting the status for the MTO. :D

src/components/Office/RequestedShipments/RequestedShipments.jsx
donaldthai: ok, `availableToPrimeAt` is now being used to determine to show the checkboxes. Much better I'd say.

PR #4860, opened 2020-09-25
MB-4234 Add an MTOAgentUpdater service object

sandy-wright:
## Description

This PR adds the `MTOAgentUpdater` service object to the `services` package and creates a set of validators for this object (in `mto_agent_validators.go`). The code for the service object itself, including the interface and the generated mock files, is in line with the way we write most other service objects. The code for the validators is new, however.

### Validators Outline

The base of the validators is a struct called `updateMTOAgentData` - the general naming scheme being **"[service object name]Data"**. This struct should define all the objects you need to be able to validate this action. For an update, that would likely be, at minimum, the old object, the update object, and `verrs` (validation errors). In this case, we also need the MTO Checker service object to check if the MTO Agent is Prime-available:

```go
type updateMTOAgentData struct {
    updatedAgent        models.MTOAgent
    oldAgent            models.MTOAgent
    availabilityChecker services.MoveTaskOrderChecker
    verrs               *validate.Errors
}
```

Functions that check this data are then defined for this struct. These should be distinct and modular. For example, `checkShipmentID` only validates the `MTOShipmentID` field, and `checkPrimeAvailability` only checks if the MTO Agent is Prime-available. 

The next level of validation is defining the structs that implement the interface `updateMTOAgentValidator`:

```go
type updateMTOAgentValidator interface {
    validate(agentData *updateMTOAgentData) error
}
```

The `validate` method for these structs should call all the `checkX` functions on the data object that are required for that level of validation. For example, the base level of validation for an MTO Agent just checks the `MTOShipmentID` field, but the Prime-level of validation needs to check the `MTOShipmentID` field, the agent's Prime-availability, and that the minimum required contact info was provided.

Then some constant string values are created to indicate which level of validation is requested for this action:

```go
// UpdateMTOAgentBaseValidator is the key for generic validation on the MTO Agent
const UpdateMTOAgentBaseValidator string = "UpdateMTOAgentBaseValidator"

// UpdateMTOAgentPrimeValidator is the key for validating the MTO Agent for the Prime contractor
const UpdateMTOAgentPrimeValidator string = "UpdateMTOAgentPrimeValidator"

// UpdateMTOAgentValidators is the map connecting the constant keys to the correct validator
var UpdateMTOAgentValidators = map[string]updateMTOAgentValidator{
    UpdateMTOAgentBaseValidator:  new(baseUpdateMTOAgentValidator),
    UpdateMTOAgentPrimeValidator: new(primeUpdateMTOAgentValidator),
}
```

This string is passed into the service object as the `validatorKey`, which is how the service object can now be used with multiple layers of validation.

## Reviewer Notes

A few things I'd appreciate feedback on:

- Do the validators seem usable and scalable? Please let me know if anything is confusing!
- How do you feel about the organization of the code in this package? Does it make sense to have these validators in their own file?
- Do you think this setup could be mimicked effectively for other service objects (like shipments, for example)?
- Anything else you want to say! This was a tricky problem to solve and I know this is a meaty solution.

I'd also appreciate extra scrutiny on my use of pointers. I tried my darnedest, but I'm still learning.

## Setup

Right now we only have the tests to go off of for this code. Feel free to write an adhoc script to call this service object against the dev database, but the tests should be fairly comprehensive. I tried my best to do TDD when I could!

You can run:

```sh
make server_test
```

for the full suite of tests. Or you can run:

```sh
go test ./pkg/services/mto_agent -v
```

for more detail.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4234) for this change.


pkg/services/mto_agent/mto_agent_validators.go
```diff
@@ -0,0 +1,167 @@
+package mtoagent
```
chrisrcoles, line 1: I really liked the update you made in moving this to a different file - this feels like a really nice way to separate concerns!

pkg/services/mto_agent/mto_agent_validators.go
```diff
@@ -0,0 +1,167 @@
...
+		return services.NewInvalidInputError(v.updatedAgent.ID, nil, v.verrs, "Invalid input found while validating the agent.")
+	}
+
+	return nil
+}
+
+// setNewMTOAgent compares updatedAgent and oldAgent and updates a new MTOAgent instance with all data
+// (changed and unchanged) filled in
+func (v *updateMTOAgentData) setNewMTOAgent(newAgent *models.MTOAgent) error {
+	err := deepcopy.Copy(&newAgent, &v.oldAgent)
```
chrisrcoles, line 145: why are we deep copying? as opposed to setting new attributes on the model instance with the base methods? (not a blocker but maybe something to thing about)

pkg/services/mto_agent/mto_agent_validators_test.go
```diff
@@ -0,0 +1,188 @@
...
+			updatedAgent:        successAgent, // as-is, should succeed
+			oldAgent:            oldAgent,
+			availabilityChecker: checker,
+			verrs:               validate.NewErrors(),
+		}
+		err := agentData.getVerrs()
+
+		suite.NoError(err)
+		suite.NoVerrs(agentData.verrs)
+	})
```
chrisrcoles, line 187: these tests are awesome! thank you so much for adding these!

chrisrcoles (approved):
Hey - thanks for spending some time with me yesterday (9/24) to walk through this PR. This looks great and provides a great foundation for adding more complex validations. I do acknowledge that it is alot of code and adds complexity but should make adding validations easier. Maybe think about creating some documentation/training on how to add these validations so the wider project knows. I added one nitpick comment re: deepcopy that is *not* a blocker. Thanks again for getting this over the finish line. 

LGTM!

pkg/services/mto_agent/mto_agent_validators.go
sandy-wright: This a really good question - I saw this convention in one of the shipment service objects (`UpdateMTOShipment`, I think), but I'm actually not convinced it needs to happen here. I'm definitely willing to change it, which might simplify the code some as well.

If anyone else has other opinions about this, I'd be happy to hear them!

abbyoung (commented):
This looks great! Excited to have the new validator pattern -- will be really useful.

PR #4862, opened 2020-09-25
Migration for new cac

LeDeep:
## Description

Migration for new cac.

## Setup

```sh
download-secure-migration 20200925155922_deep_cac.up.sql
make run_prod_migrations
```


reggieriser (approved):
I verified that local migrations worked and the `client_certs` table was updated appropriately.  Looked on GovCloud AWS S3 and saw appropriate migrations in exp/stg and a no-op in prod.  I'm not sure what's going on with the pre-commit checks, but hopefully a re-run (or a merge and re-run) will take care of them.

PR #4864, opened 2020-09-25
[MB-4419] Added precommit hook for spectral linter

shimonacarvalho:
## Description

**Please don't approve if the test in Reviewer Notes isn't super smooth! Let me know** 

Spectral ruleset, script and linter added to the repo and precommit. 

By default all openapi 2.0 rules are tested, I’ve only disabled one, which requires we include a host tag at the top level.
Here are the additional rules I’ve set up. 

Generally speaking, the MUSTs are errors - they cause the linter to fail; the shoulds are warnings that allow the linter to pass:

- Endpoint MUST include description
- The endpoint description should be at least 128 characters.
- Path operationId MUST be camelCase
- Paths MUST be kebab-case and {params} in path MUST be camelCase
- Endpoints MUST include produces property
- Allowable types for produces are [application/json]
- Endpoints that [POST,PUT,PATCH] methods must include consumes property
- Allowable types for consumes are [application/json, multipart/form-data]
- Body properties MUST be camelCase
- Property createdAt MUST be readOnly
- Property updatedAt MUST be readOnly
- Property eTag MUST be readOnly
- Property of type date should not give example, due to ReDoc quirk
- Definitions MUST be sorted alphabetically
- Definition names MUST not end in Payload

## Reviewer Notes

I'd really like to check that the install of spectral through the regular make and precommit infrastructure works properly. 
So please try removing spectral from your machine (`yarn spectral` should return an error) and then try:

```
  make clean
  make build # should install spectral
   # either make a fake commit by changing a file and the hooks should run
   # OR this command runs just spectral but through the precommit functionality
  pre-commit run spectral --all-files
```


abbyoung (approved):
Tested both paths listed in Review Notes and worked with zero issues. Looks great!

PR #4867, opened 2020-09-26
MB-4384 Display empty string for 0 SIT allowance

monfresh:
## Description

By default, a service member's SIT (storage in transit) allowance is
90 days. However, the `StorageInTransit` field does not have a default
value in the DB, and we are not explicitly setting it to 90 days
when the customer creates their move. This was causing the frontend
to display `undefined days` as the SIT Allowance in the Move Details
page in the TOO interface. This is because the `StorageInTransit`
field was not being returned at all as part of the payload when it is
zero. Note that there is a separate JIRA story to set the default
SIT allowance.

In this PR, we are fixing the frontend so that it displays an empty
string if SIT is zero. To do that, we set the `x-nullable` attribute
for `storageInTransit` to `true`, which allows the frontend to check
if it is zero or non-zero. If it is zero, we display an empty string.

If it's ever possible for the SIT allowance to only be one day, we
might consider using some kind of pluralization helper so that it
would display the singular `day` as opposed to the hardcoded plural
`days` that we have now.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4384) for this change

## Screenshots

Before:

Screen Shot 2020-09-23 at 10 05 35 AM

After:

Screen Shot 2020-09-25 at 9 37 05 PM

### DIFF
--- a/src/components/Office/AllowancesTable.test.jsx
+++ b/src/components/Office/AllowancesTable.test.jsx
@@ -26,3 +26,11 @@ describe('Allowances Table', () => {
     expect(wrapper.find({ 'data-testid': 'dependents' }).text()).toMatch('Authorized');
   });
 });
+
+describe('Allowances Table when SIT is 0', () => {
+  it('displays an empty string for the SIT allowance', () => {
+    info.storageInTransit = 0;
+    const wrapper = shallow(<AllowancesTable info={info} />);
+    expect(wrapper.find({ 'data-testid': 'storageInTransit' }).text()).toEqual('');
+  });
+});
--- a/pkg/gen/ghcapi/embedded_spec.go
+++ b/pkg/gen/ghcapi/embedded_spec.go
@@ -2071,6 +2071,7 @@ func init() {
         },
         "storageInTransit": {
           "type": "integer",
+          "x-nullable": true,
           "example": 90
         },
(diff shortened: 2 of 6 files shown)


src/components/Office/AllowancesTable.test.jsx
```diff
@@ -26,3 +26,11 @@ describe('Allowances Table', () => {
     expect(wrapper.find({ 'data-testid': 'dependents' }).text()).toMatch('Authorized');
   });
 });
+
+describe('Allowances Table when SIT is 0', () => {
+  it('displays an empty string for the SIT allowance', () => {
+    info.storageInTransit = 0;
```
duncan-truss, line 32: If we ever get a value of zero and not null then we are still going to see a blank field instead of `0 days` because it will be evaluated as falsey.  Not sure that matters just something to note.

duncan-truss (approved):

src/components/Office/AllowancesTable.test.jsx
monfresh: Yeah, that's the expected result, hence this test. Alexi wanted a blank field. When the `StorageInTransit` field is not set, it becomes `0` in the payload.

PR #4875, opened 2020-09-29
[MB-3946] Notification handler for MoveTaskOrder

shimonacarvalho:
## Description

Added the `assembleMTOPayload` function. Also had to add a payload definition and a custom model_to_payload because the MoveTaskOrder payload in primeapi contains unnecessary items, like MoveOrders, MTOShipments, PaymentRequests. 

Some of them, like MTOShipments and PaymentRequests, if unpopulated, just show up as empty arrays which would be sort of ok. But MoveOrders shows up as an object with empty fields and that looks broken.

I tried a few things to avoid having to define a new payload but none were successful 
- I looked for a way to tell the json unmarshall to ignore the MoveOrders object but couldn't find anything that wouldn't require retooling the MoveTaskOrder object. But it's gen'ed by swagger so we can't control it that much.
- I tried splitting the MoveTaskOrder yaml definition into a minimal and extended version and recombine them using `allof` . That failed, it actually generated pkg/gen code that was uncompile-able. I think the polymorphism of MTOServiceItems does not play well with the allOf functionality.
- Would be interested to hear other ideas.

## Setup

Run the unit test added. 
## References

* [ [MB-3946] Push Enable: Add notification handler for MoveTaskOrder - Jira](https://dp3.atlassian.net/browse/MB-3946) for this change

[MB-3946]: https://dp3.atlassian.net/browse/MB-3946

### DIFF
--- a/pkg/services/event/event.go
+++ b/pkg/services/event/event.go
@@ -39,11 +39,11 @@ type Event struct {
 	logger          handlers.Logger         // The logger
 }
 
-// PaymentRequestCreateEventKey is a key containing PaymentRequest.Create
-const PaymentRequestCreateEventKey KeyType = "PaymentRequest.Create"
+// MoveTaskOrderCreateEventKey is a key containing MoveTaskOrder.Create
+const MoveTaskOrderCreateEventKey KeyType = "MoveTaskOrder.Create"
 
-// PaymentRequestUpdateEventKey is a key containing PaymentRequest.Update
-const PaymentRequestUpdateEventKey KeyType = "PaymentRequest.Update"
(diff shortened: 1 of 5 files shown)


sandy-wright (approved):
Looks good! I think it might actually be a good thing that you had to rewrite the MoveTaskOrder payload because it sounds like we might be rethinking that endpoint soon, so the payload in the Prime API may be changing.

shimonacarvalho:
> Looks good! I think it might actually be a good thing that you had to rewrite the MoveTaskOrder payload because it sounds like we might be rethinking that endpoint soon, so the payload in the Prime API may be changing.

Good point! Hadn't thought of that.

PR #4879, opened 2020-09-29
[MB-4427] MS and CS service items should not error out when generating EDI

garrettqmartin8:
## Description

Creating an EDI from a payment request for only MS and CS fails due to missing dates.  

## Setup

1. `make server_run`
2. Create a payment request. Create a file called `create_payment_request.json` in the project root and paste in the JSON payload provided below:

```json
{
  "body": {
    "isFinal": false,
    "serviceItems": [
      {
        "id": "722a6f4e-b438-4655-88c7-51609056550d"
      },
      {
      }
    ]
  }
}
```

2. Use the Prime API Client to create the payment request and return the payment request number by running the command below:

```sh
prime-api-client --insecure create-payment-request --filename  create_payment_request.json | jq '.paymentRequestNumber'
```
3. Use the CLI utility to generate the EDI858 and view it in your terminal. Remember to pass the payment request number of the payment request you just created.

```sh
go run ./cmd/generate-payment-request-edi/ --payment-request-number [$paymentRequestNumber]
```

You should see a valid EDI in standard out with MS and CS service items.

## References

[Jira story](https://dp3.atlassian.net/browse/MB-4427)


pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -155,31 +155,31 @@ func (g ghcPaymentRequestInvoiceGenerator) Generate(paymentRequest models.Paymen
...
-	originDestinationSegments, err := g.createOriginAndDestinationSegments(paymentRequest.ID, moveTaskOrder.Orders)
-	if err != nil {
-		return ediinvoice.Invoice858C{}, err
+		// Add origin and destination details to header
+		var originDestinationSegments []edisegment.Segment
+		originDestinationSegments, err = g.createOriginAndDestinationSegments(paymentRequest.ID, moveTaskOrder.Orders)
+		if err != nil {
+			return ediinvoice.Invoice858C{}, err
+		}
+		edi858.Header = append(edi858.Header, originDestinationSegments...)
```
gidjin, line 181: Why is this here, I would expect that even MS and CS only payment requests still have service member orders object so we can still include origin and destination. Basically I expect that an MTO will have an Orders object even if there are not shipments.

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -499,3 +517,15 @@ func (g ghcPaymentRequestInvoiceGenerator) generatePaymentServiceItemSegments(pa
 
 	return segments, nil
 }
+
+func msAndCsOnly(paymentServiceItems models.PaymentServiceItems) bool {
```
gidjin, line 521: This method is named confusingly to me. The name `msAndCsOnly` implies that it should return true if we only have MS and CS, but it does the opposite of that. I would change it to return true if we only MS or CS and rename it to `msOrCsOnly` since a PaymentRequest could be just one as far as I understand.

gidjin (changes requested):
Changes work, but there are a couple of issues that I left in the comments.

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -499,3 +517,15 @@ func (g ghcPaymentRequestInvoiceGenerator) generatePaymentServiceItemSegments(pa
...
+	var codesWithoutMsOrCs []string
+	for _, psi := range paymentServiceItems {
+		code := string(psi.MTOServiceItem.ReService.Code)
+		if code != "MS" && code != "CS" {
+			codesWithoutMsOrCs = append(codesWithoutMsOrCs, code)
```
reggieriser, line 526: A couple of things here:
- Let's do the comparison against the constants like `ReServiceCodeMS` and `ReServiceCodeCS` rather than converting to a string and testing against string literals.
- Nit: Since you only care if there's a code that's not either MS or CS, couldn't you just return false immediately in the loop as soon as you see one rather than building up another slice and testing its length.

pkg/services/invoice/ghc_payment_request_invoice_generator.go
garrettqmartin8: Yep, this is wrong. I'll move that out of the conditional. Good catch!

pkg/services/invoice/ghc_payment_request_invoice_generator.go
garrettqmartin8: Yep, both of those changes make sense. I'll address them now.

gidjin:
You'll also want to pull master to resolve the `pre_test` failure

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -257,6 +248,33 @@ func (g ghcPaymentRequestInvoiceGenerator) createServiceMemberDetailSegments(pay
...
+		Where("pr.id = ?", paymentRequestID).
+		Where("sipk.key = ?", models.ServiceItemParamNameRequestedPickupDate).
+		First(&requestedPickupDateParam)
+
+	if err != nil {
+		return nil, fmt.Errorf("Couldn't find requested pickup date: %s", err)
+	}
+
+	requestedPickupDateSegment := edisegment.G62{
+		DateQualifier: 86,
```
gidjin, line 269: When pairing with Reggie we noticed this was `86` which is for Actual Pickup date vs `10` for Requested Pickup date. But I'm fine if this is fixed as part of Reggie's PR which will add the additional dates.

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -155,27 +155,27 @@ func (g ghcPaymentRequestInvoiceGenerator) Generate(paymentRequest models.Paymen
...
+		All(&paymentServiceItems)
 	if err != nil {
-		return ediinvoice.Invoice858C{}, fmt.Errorf("Couldn't find requested pickup date: %s", err)
+		return ediinvoice.Invoice858C{}, fmt.Errorf("Could not find payment service items: %w", err)
 	}
 
-	requestedPickupDateSegment := edisegment.G62{
-		DateQualifier: 86,
-		Date:          requestedPickupDateParam.Value,
+	if !msOrCsOnly(paymentServiceItems) {
```
gidjin, line 167: Small nitpick, I don't like `if !boolean` because the `!` is easy to miss, but not sure there is a clear better way to adjust this so I'm fine if it stays.

gidjin (approved):
LGTM, thanks for making those changes.

reggieriser (approved):
LGTM -- thanks for making the changes!

pkg/services/invoice/ghc_payment_request_invoice_generator.go
garrettqmartin8: Yeah I toyed around with names for this and I think it's important that we make it clear that the situation in which this conditional is met has to do with their only being MS or CS. I unfortunately couldn't come up with something that I liked better.

PR #4880, opened 2020-09-29
MMDM-1630 Remove rd's cac access

abbyoung:
## Description
Removes CAC access. This doesn't have to be a secure migration because we're only identifying the sha256 digest in dev/experimental/staging/prod. And if it doesn't exist in those environments this is a noop.

## Reviewer Notes
NOTE: This needs to be deployed manually to experimental to apply in that environment.

## Setup
`make db_dev_run db_dev_migrate`


gidjin (approved):
LGTM

LeDeep (approved):

PR #4883, opened 2020-09-29
Kim mb 4125 selectable card component

kimallen:
## Description

Creates a SelectableCard component for a radio and adds it to storybook.  Three states are displayed in storybook: unselected, selected, and disabled.

## Reviewer Notes

I did not create a checkbox option or separate component.  it sounded like it may not be necessary ([see thread](https://ustcdp3.slack.com/archives/CTNTFJSBA/p1601334712127300)) right now.

## Setup

`make storybook`

Look under Customer Components --> SelectableCard

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4125) for this change.

## Screenshots
phone view:
image

browser view (updated with text alignment):
image

selected (updated):
image

disabled:
image


Shkeating (changes requested):
This is great! I am noticing is the drop shadow is missing when the card is selected. I've spent some time this morning to see if there was a css-only way to acheive this, but I think we need to add a class/id to the `.cardContainer` div in order to have a way to target the div when the radio is selected. when we have a way to target it, we should be able to add the shadow with `@include u-shadow(2);`

The only other thing I am noticing this that the icon in the top left corner is missing, but it seems that icon has not been added to the repo for use (if it were, you would find it here: https://storybook.dp3.us/?path=/story/global-icons--all) I will export it out of the Sketch file and get it added for you ASAP!

suzubara:
> The only other thing I am noticing this that the icon in the top left corner is missing, but it seems that icon has not been added to the repo for use (if it were, you would find it here: https://storybook.dp3.us/?path=/story/global-icons--all) I will export it out of the Sketch file and get it added for you ASAP!

Since we're not implementing the actual handler for this icon in this story, it seems like it would make more sense to add the icon itself when we do add the tooltip/modal?

src/components/Customer/SelectableCard.jsx
```diff
@@ -0,0 +1,41 @@
...
+    </div>
+  );
+};
+
+SelectableCard.propTypes = {
+  id: string.isRequired,
+  label: string.isRequired,
+  name: string.isRequired,
+  value: string.isRequired,
+  cardText: string,
```
suzubara, line 29: We will probably want `cardText` to be `PropTypes.node` so that it can include rich text. Then the usage would be something like:
```
<SelectableCard cardText={<p>My card description...</p>} ... />
```

src/components/Customer/SelectableCard.module.scss
```diff
@@ -0,0 +1,20 @@
+@import '../../shared/styles/basics';
+@import '../../shared/styles/mixins';
+@import '../../shared/styles/colors';
+
+.cardContainer {
+  @include u-padding-top(2);
+  @include u-padding-right(2);
+  @include u-padding-bottom(3);
+  @include u-padding-left(2);
+  border-radius: 0.26667rem;
```
suzubara, line 10: I think using px or units for border radius will be easier to keep track of.. it looks like this translates to `border-radius: 2px`

src/components/Customer/SelectableCard.module.scss
```diff
@@ -0,0 +1,20 @@
...
+  margin: 1.6rem;
```
suzubara, line 11: We will probably not want to set global margins on this component, because it will depend a lot on the context within which it's rendered. I would recommend removing this. However it _can_ be useful to set margins in between repeated components, ie.:

```
.cardContainer + .cardContainer {
  @include u-margin-top(1);
}
```

That will set a 1 unit top margin to SelectableCards that follow other SelectableCards

src/components/Customer/SelectableCard.module.scss
```diff
@@ -0,0 +1,20 @@
...
+  background-color: $bg-white;
+  border:  2px solid $base-lighter;
+}
+
+.cardText {
+  font-weight: normal;
+  color: $base;
+  font-size: 13px;
```
suzubara, line 19: Looking at the designs, I think this also needs `line-height: 1.25;` to achieve a 16px line height

src/components/Customer/SelectableCard.test.jsx
```diff
@@ -0,0 +1,30 @@
...
+  onChange: jest.fn(),
+};
+function mountSelectableCard(props = defaultProps) {
+  return mount(
+    <Provider store={store}>
+      <SelectableCard {...props} />
+    </Provider>,
+  );
+}
+describe('SelectableCard component', () => {
```
suzubara, line 25: This component should not need to be mounted with the Redux provider & store. It can just be a basic:
```
mount(<SelectableCard {...props} />)
```

suzubara (changes requested):
This is in really good shape! I did leave some comments but overall great job :D

src/components/Customer/SelectableCard.module.scss
kimallen: TIL!

kimallen:
> This is great! I am noticing is the drop shadow is missing when the card is selected. I've spent some time this morning to see if there was a css-only way to achieve this, but I think we need to add a class/id to the `.cardContainer` div in order to have a way to target the div when the radio is selected. when we have a way to target it, we should be able to add the shadow with `@include u-shadow(2);`
> 
> The only other thing I am noticing this that the icon in the top left corner is missing, but it seems that icon has not been added to the repo for use (if it were, you would find it here: https://storybook.dp3.us/?path=/story/global-icons--all) I will export it out of the Sketch file and get it added for you ASAP!

oof- you're right- I totally forgot about the color change and didn't see the drop shadow when selected.  Will work on that.

Do you mean to informational icon in the top right?  The story specifically says not to worry about that- we will cover it when we build the modal in another story.

src/components/Customer/SelectableCard.module.scss
```diff
@@ -0,0 +1,29 @@
...
+  @include u-padding-bottom(3);
+  @include u-padding-left(2);
+  border-radius: 2px;
+  background-color: $bg-white;
+  border:  2px solid $base-lighter;
+}
+
+.selected {
+  @include u-shadow(2);
+  background-color: #ecf1f7;
```
kimallen, line 17: @Shkeating - I didn't find a close enough color in the colors.scss.  I should probably add this color into that file-- do you have any recommendations for where that should go or for a name?

kimallen: `$link-light` is the closest existing color I could find, but it was a a little darker

amanda5iu (changes requested):
Thanks for including screenshots Kim. This looks great! 

My only note is that on desktop, the grey text should be left-aligned with the radio button label. After that, looks like we're good to go!

amanda5iu (approved):

Shkeating (approved):
i'm going to approve so this can move forward. Once I find out what color variable actually should be here for the selected box's background color, I'll open up a branch and fix it! The drop shadow looks great!

src/components/Customer/SelectableCard.module.scss
kimallen: @Shkeating -thx for asking in #prac-design.  I ended up added two colors from the palette that were missing. (info-light and info).

suzubara (approved):
Updates look good! 💯

PR #4891, opened 2020-09-30
MB-4385 Allow MTO tab to render all the time

monfresh:
**Description**
Previously, if you clicked on the Move task order tab when looking
at a Move as a TOO, and when no shipments had been approved yet,
it would throw an exception. This was because the code was assuming
there would always be service items present.

To fix this, we can use the Null Propagation operator so that if
something is null, it doesn't throw an `undefined` exception. In
addition, we added a conditional around the ShipmentContainer so that
only approved shipments appear.

### Scenario where no approved shipments exist
1. Run the setup commands in the section below
2. Visit http://officelocal:3000/devlocal-auth/login
3. Click the `Login` button next to `too_role@office.mil (PPM office)`
4. Click on a move that doesn't have any approved shipments yet
5. Click on the Move task order tab

Expected Result: You don't see an exception. Instead, the MTO tab renders, but with nothing displayed.

### Scenario where at least one approved shipment exists
1. Run the setup commands in the section below
2. Visit http://officelocal:3000/devlocal-auth/login
3. Click the `Login` button next to `too_role@office.mil (PPM office)`
4. Click on, or create a move that has 2 shipments: one that is approved, and one that is not approved.
5. Click on the Move task order tab

Expected Result: Only the approved shipment, along with its service items, appears in the MTO tab.

## Setup

```sh
make db_dev_e2e_populate
make server_run
make office_client_run (in a separate terminal tab)
```

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4385) for this change

## Screenshots

Screen Shot 2020-09-30 at 3 17 08 PM

### DIFF
--- a/src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
+++ b/src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
@@ -102,7 +102,7 @@ export const MoveTaskOrder = ({ match }) => {
   if (isLoading) return <LoadingPlaceholder />;
   if (isError) return <SomethingWentWrong />;
 
-  const serviceItems = mtoServiceItemsArr.map((item) => {
+  const serviceItems = mtoServiceItemsArr?.map((item) => {
     const newItem = { ...item };
     newItem.code = item.reServiceCode;
     newItem.serviceItem = item.reServiceName;
@@ -125,6 +125,17 @@ export const MoveTaskOrder = ({ match }) => {
     setIsModalVisible(true);
   };
 
+  const approved = (shipment) => shipment.status === 'APPROVED';
+  const mtoShipmentsArr = Object.values(mtoShipments);
+
+  if (!mtoShipmentsArr.some(approved)) {
+    return (
+      <div>
+        <p>This Move does not have any approved shipments yet.</p>
+      </div>
+    );
+  }
+
   return (
     <div className={styles.tabContent}>
       <GridContainer className={styles.gridContainer} data-testid="too-shipment-container">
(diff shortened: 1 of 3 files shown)


src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -144,74 +144,83 @@ export const MoveTaskOrder = ({ match }) => {
...
           );
-          const approvedServiceItems = serviceItemsForShipment.filter(
+          const approvedServiceItems = serviceItemsForShipment?.filter(
             (item) => item.status === SERVICE_ITEM_STATUS.APPROVED,
           );
-          const rejectedServiceItems = serviceItemsForShipment.filter(
+          const rejectedServiceItems = serviceItemsForShipment?.filter(
             (item) => item.status === SERVICE_ITEM_STATUS.REJECTED,
           );
+          if (mtoShipment.status === 'APPROVED') {
```
duncan-truss, line 157: I think what you have should work fine.  My suggestion would be we can move this condition check to the top of the block as more of a guard clause.  So if the status is not equal to APPROVED then return the basic ShipmentContainer otherwise filter all of the status items which you know will exist for the approved shipment and we won't need the optional operator?

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -144,74 +144,83 @@ export const MoveTaskOrder = ({ match }) => {
...
+                    handleUpdateMTOServiceItemStatus={handleUpdateMTOServiceItemStatus}
+                    handleShowRejectionDialog={handleShowRejectionDialog}
+                    statusForTableType={SERVICE_ITEM_STATUS.REJECTED}
+                  />
+                )}
+              </ShipmentContainer>
+            );
+          }
           return (
             <ShipmentContainer
```
duncan-truss, line 219: I'm not sure what the point is of returning the empty container at all unless we're going to display some information.  We could return false to skip including it in the resulting array.  In the react query I wonder if we should expand it to take a status for mto shipments then just avoid those statuses all together.

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
duncan-truss: Nice work! You could display an info message box maybe like the one in storybook but I'm guessing it will be temporary https://storybook.dp3.us/?path=/story/components-alerts--info ?

Ryan-Koch (approved):

PR #4902, opened 2020-10-01
Migration for updated CAC

reggieriser:
## Description

This PR adds a migration to update environments with my updated CAC information.

## Setup

- `db_dev_migrate` to make sure local migration works fine.  Ensure the `CN` looks as expected.
- `download-secure-migration 20201001183657_rriser_cac.up.sql`, then review downloaded files to make sure staging/experimental has the update (with correct `CN`) and prod has a no-op.


gidjin (approved):
Changes look good, verified the uploaded secure migrations look good too.

LeDeep (approved):
migrations ran as expected, downloaded secure migration and it looks good.

PR #4909, opened 2020-10-02
Update Review page design using storybook components

Ronolibert:
## Description

Update customer review page to use new storybook components

## Reviewer Notes
- After a design review with Mariaclara, I changed the shipment ID to be 8 characters and uppercase to be a little more design friendly until we have a dedicated value for that slot
- The weight estimate for PPM does not currently work/does not get saved in the flow so for now it is a static value until we fix that bug (card already made). I added a yellow section to denote it as todo in future cards
- The edit path for the PPM weight estimate does not go anywhere until we fix the weight estimate bug
- Had to use !important to override a specific rule adding margin to h2

## Setup
`make server_run`
`make client_run`
Go through normal setup flow until you reach the review page

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4300) for this change

## Screenshots

image


src/shared/utils.js
```diff
@@ -161,3 +161,16 @@ export function dateSort(field, direction) {
 export function formatCustomerDate(date) {
   return moment(date).format('DD MMM YYYY');
 }
+
+// Format orders type (ex: PERMANENT_CHANGE_OF_STATION => Permanent change of station)
```
suzubara, line 165: It would be great to stop adding new code under `shared`, can we start a file under `src/utils` for any new formatters?

src/components/Customer/Review/ShipmentCard/HHGShipmentCard/index.jsx
```diff
@@ -72,20 +75,30 @@ const HHGShipmentCard = ({
           </div>
           <div className={styles.row}>
             <dt>Destination</dt>
-            <dd>{destinationZIP}</dd>
+            <dd>
```
suzubara, line 78: Maybe we can create a shared function for formatting names & addresses, since this probably occurs in a lot of places throughout the app?

suzubara (commented):
left some general comments since this is still a WIP!

src/scenes/Review/Review.css
```diff
@@ -110,3 +110,7 @@
 .hhg-shipment-summary .not-implemented {
   display: none;
 }
+
+.review-move-container {
```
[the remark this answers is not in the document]
Ronolibert: Moved this to the module

src/shared/utils.js
Ronolibert: Made the formatters file in utils

src/components/Customer/Review/OrdersTable/index.jsx
```diff
@@ -1,15 +1,24 @@
...
+  onEditClick,
+  orderType,
+  reportByDate,
+  uploads,
+}) => (
   <div className={reviewStyles['review-container']}>
     <div className={reviewStyles['review-header']}>
       <h2>Orders</h2>
-      <Button unstyled className={reviewStyles['edit-btn']}>
+      <Button unstyled className={reviewStyles['edit-btn']} onClick={() => onEditClick(editPath)}>
```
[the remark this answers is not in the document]
Ronolibert: i'm not sure what you mean by this, do you mean handling the logic of what path it would navigate to in the handler itself? I just did it this way because the alternative was passing in the `moveId` to the component itself just to be used as an argument for the handler

src/components/Customer/Review/ShipmentCard/HHGShipmentCard/index.jsx
Ronolibert: Added a function for the addresses!

src/components/Customer/Review/OrdersTable/index.jsx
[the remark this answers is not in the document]
suzubara: So looking at the Summary component, where `OrdersTable` is rendered, it would look like this:
```
<OrdersTable onEditClick={() => this.handleEditClick(editOrdersPath) } ...
```

this is assuming that the edit handler for `OrdersTable` will be constant, but that is my impression based on the code.

mariaclaratgolfo:
This is looking great!

- In Happo, it appears we now have two ## signs appearing ahead of the shipment code for the shipment cards. This is appearing on the Review page and the Home page shipment cells. Let's be sure that's not the case

- Let's make sure the padding below each subheader [Profile, Orders, Move setup] is the same. It should be 3 units below each. Right now it appears the space below Move setup is smaller than the others.

- The buttons at the bottom of the page should be side by side [Back to the left, Next to the right]

mariaclaratgolfo (changes requested):
Left a comment with a couple of things to double check

src/components/Customer/Review/OrdersTable/index.jsx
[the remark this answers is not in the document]
Ronolibert: Ahh yeah, I guess this is a case of a little bit of premature optimization. I'll change this so that the handler passed in is responsible for where it is going

kimallen:
> This is looking great!
> 
> * In Happo, it appears we now have two ## signs appearing ahead of the shipment code for the shipment cards. This is appearing on the Review page and the Home page shipment cells. Let's be sure that's not the case
> * Let's make sure the padding below each subheader [Profile, Orders, Move setup] is the same. It should be 3 units below each. Right now it appears the space below Move setup is smaller than the others.
> * The buttons at the bottom of the page should be side by side [Back to the left, Next to the right]

We should clarify about the side-by-side v. stacked.  Last week there was various conversation about it (there's different ways they show up in designs), and it was determined to do them stacked in phone view and side by side in browser view. @amanda5iu @mariaclaratgolfo

src/components/Customer/Review/OrdersTable/index.jsx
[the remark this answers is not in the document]
Ronolibert: Simplified this logic

mariaclaratgolfo (approved):

src/components/Customer/Review/ShipmentCard/ShipmentCard.module.scss
```diff
@@ -3,6 +3,10 @@
 @import 'shared/styles/_variables';
 @import 'shared/styles/colors';
 
+.todo {
```
suzubara, line 6: if this is a style that should be able to applied anywhere to indicate a WIP, it might actually be easier to add to a global CSS file instead of a CSS module 🤷‍♀️

src/components/Customer/Review/Summary/index.test.jsx
```diff
@@ -2,14 +2,16 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 
-import { Summary } from './Summary';
+import { Summary } from '.';
```
suzubara, line 5: can this be `'./index'` just so it's a little clearer?

src/pages/MyMove/Review/index.jsx
```diff
@@ -0,0 +1,64 @@
+import React, { Component } from 'react';
+import { arrayOf, bool, string } from 'prop-types';
+import { connect } from 'react-redux';
+
+import styles from './Review.module.scss';
+
+import { hasShortHaulError } from 'shared/incentive';
+import { no_op as noOp } from 'shared/utils';
+import scrollToTop from 'shared/scrollToTop';
+/* eslint-disable import/no-named-as-default */
```
suzubara, line 10: I'd prefer if we didn't ignore this ESLint rule. to get around it, you have to name the default export something that is not already a named export. Usually this means the default export is the component with Redux and/or other HOCs, so a common convention is to do something like `import ConnectedWizardPage from 'shared/WizardPage';`. then it's clearer what is being imported.

src/components/Customer/Review/Summary/Summary.module.scss
```diff
@@ -0,0 +1,9 @@
+@import 'shared/styles/_basics';
+@import 'shared/styles/_mixins';
+@import 'shared/styles/_variables';
+
+.moveSetup {
+  // using important to override rule that has more specificity on h2
+  margin: 0 !important;
```
suzubara, line 7: I don't see what effective styles these are implementing.. there's already a margin bottom and margin top on the h2 sibling elements, and removing these styles doesn't seem to change anything. It would be best to avoid using `!important` as much as possible. Can we remove this?

cypress/integration/mymove/hhg.js
```diff
@@ -21,7 +21,7 @@ describe('HHG Setup flow', function () {
     customerChoosesAnHHGMove();
     customerSetsUpAnHHGMove();
     customerReviewsMoveDetailsAndEditsHHG();
-    customerSubmitsMove();
+    // customerSubmitsMove();
```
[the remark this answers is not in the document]
Ronolibert: Oops! Uncommenting this, was just debugging tests. Good catch!

suzubara (approved):

PR #4912, opened 2020-10-02
MB-4126 | add another shipment logic

sojeri:
## Description

this change updates the add a shipment logic. see more in setup and in attached jira story, but tl;dr--- since user can only add 1 PPM and can only add HHGs prior to submitting move, there are several states which indicated disabled options in the shipment selection flow & updated language to educate the user re: why these new options are disabled.

## Reviewer Notes

~1. I need to finish updating component tests to cover some of these new scenarios. if you have thoughts on whether I should do this by mocking the full state vs just using the new props like `isHhgSelectable`, I'd love to hear that.~

~2. I have heard mixed things from design re: whether add a shipment button should be disabled or removed in certain scenarios.~

## Setup

there are several scenarios to test. I will share some screenshots at the end of this PR description of a couple of these.

- given: you have some shipments already but have not submitted your move
  - then: does the shipment eyebrow on shipment selection screen show the right number?

- given: you have an HHG shipment and have signed/submitted your move
  - then: is the add a shipment button visible or enabled?
  - also: is HHG language updated on shipment selection page & the option disabled?
  - and: does the shipment eyebrow on shipment selection screen show the right number?

- given: you have added a PPM shipment already but have not submitted your move
  - and: is the PPM language updated on shipment selection page & the option disabled?
  - and: does the shipment eyebrow on shipment selection screen show the right number?

- given: you have an HHG+ (1 or more) and PPM shipments on an already signed/submitted move
  - then: is the add a shipment button hidden?
  - and: if you can get to the shipment selection page, is the language for both PPM & HHG updated?
  - and: if you can get to the shipment selection page, does the shipment eyebrow show the right number?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4126) for this change

### given: you have already added a PPM and an HHG

Screen Shot 2020-10-02 at 2 54 36 PM

### and: the shipment eyebrow should read 3 (2 for the PPM & HHG + 1 for the one you're about to create)

Screen Shot 2020-10-02 at 2 54 21 PM

### given: you have already submitted a move not containing a PPM

Screen Shot 2020-10-02 at 3 05 03 PM

### and: the shipment eyebrow should read 2 (1 for the HHG in submitted move + 1 for the PPM you're about to create)

Screen Shot 2020-10-02 at 3 05 13 PM


src/pages/MyMove/Home/index.jsx
```diff
@@ -117,14 +117,19 @@ class Home extends Component {
...
   }
 
+  get doesPpmAlreadyExist() {
+    const { move } = this.props;
+    return !!move.personally_procured_moves?.length;
+  }
+
   get shipmentActionBtnLabel() {
-    if (this.hasSubmittedMove) {
+    if (this.hasSubmittedMove && this.doesPpmAlreadyExist) {
```
sojeri, line 126: this will cause this step 3 button to disappear if move is submitted and PPM already exists. see reviewer notes #2 for more text.

src/pages/MyMove/Home/index.jsx
```diff
@@ -117,14 +117,19 @@ class Home extends Component {
...
       return '';
     }
     if (this.hasShipment) {
-      return 'Plan your shipments';
+      return 'Add another shipment';
```
sojeri, line 130: note to self: this should be `this.hasShipment || this.doesPpmAlreadyExist` b/c any shipment should trigger Add vs Plan

src/pages/MyMove/SelectMoveType.jsx
```diff
@@ -43,35 +53,55 @@ export class SelectMoveType extends Component {
...
-              </ul>
+              {isPpmSelectable && (
+                <ul>
+                  <li>This is a PPM - “personally procured move”</li>
+                  <li>You arrange to move some or all of your belongings</li>
+                  <li>The government pays you an incentive based on weight</li>
+                  <li>DIY or hire your own movers</li>
+                </ul>
+              )}
+              {!isPpmSelectable && (
```
sojeri, line 75: @suzubara is there a more intuitive way to do this? I went for the easy route here, but in reading this file as whole it's weird to see thing && content, !thing && content

src/pages/MyMove/SelectMoveType.jsx
```diff
@@ -28,8 +37,9 @@ export class SelectMoveType extends Component {
     return updateMove(match.params.moveId, moveType);
   };
 
+  // TODO: question PR reviewers on checked behavior
```
sojeri, line 40: re: checked behavior: if you have added an HHG shipment and you go back to selection page to add a new one, the current behavior (regardless of my changes) is that the HHG item will be checked. is this expected/desired? it felt weird to me, but I understand it can get complex otherwise if we need to maintain this checked behavior if/when customer hits back button from the details form page.

src/pages/MyMove/SelectMoveType.test.jsx
```diff
@@ -13,6 +13,9 @@ describe('SelectMoveType', () => {
     updateMove: () => {},
     push: () => {},
     selectedMoveType: SHIPMENT_OPTIONS.PPM,
+    isPpmSelectable: true,
```
sojeri, line 16: this is where my first reviewer's note starts to matter. would you rather see this, or is there a better way here I should hydrate more of the actual state so this gets calculated in mapStateToProps based on actual PPM / HHG shipments state.

src/pages/MyMove/SelectMoveType.jsx
suzubara: you can do:
```
{isPpmSelectable ? ppmSelectableContent : ppmUnselectableContent}
```

(I also recommend moving content into variables like that just so it's a little easier to read with some separation of content / logic)

src/pages/MyMove/SelectMoveType.jsx
suzubara: that is weird to me. Navigating back to the selection page should mean `SelectMoveType` gets re-mounted, and there shouldn't be any initial "checked" state (unless one is passed in explicitly via props)

suzubara: oh just looked at it real quick, SelectMoveType is initializing `moveType` state based on `props.selectedMoveType` which is a [common React anti-pattern](https://stackoverflow.com/questions/28785106/reactjs-why-is-passing-the-component-initial-state-a-prop-an-anti-pattern). I realize this is not related to your PR, but wanted to point it out as something that should probably be removed. What is the use case for a move type to already be selected when. navigating to this page?

src/pages/MyMove/SelectMoveType.jsx
suzubara: (also I imagine this will end up looking a little different once @kimallen 's PR is in)

src/pages/MyMove/SelectMoveType.jsx
sojeri: that's exactly what I'm saying @suzubara 
> is this expected/desired? it felt weird to me, but I understand it can get complex otherwise if we need to maintain this checked behavior if/when customer hits back button from the details form page.

kimallen: the `selectedMoveType` is saved on the move in the db, so if you are remounting, that value is still on the move.  Whatever the last one selected is will be the value.  An alternative is that when a shipment or PPM is fully created (but move is not yet submitted), then we set the value of `move.selectedMoveType` to ''.  The reason we store this in the db is that we need to know what their selection was if they were mid-creating one so that we can start them where they left off if they log out and log back in.

suzubara (commented):
Left some comments to hopefully answer your questions! I also tested the scenarios you mentioned and they look good. One comment is that when I _do_ access the Select Move Type page when I'm not supposed to (i.e., already submitted a move), it seems like the back & next buttons should be disabled. Clicking on them produces an API call with a 403 response. But since the user shouldn't end up on this page in the first place, I don't know how much of an issue that is.

src/pages/MyMove/Home/index.jsx
```diff
@@ -117,14 +117,19 @@ class Home extends Component {
     return !!Object.keys(move).length && move.status !== 'DRAFT';
   }
 
+  get doesPpmAlreadyExist() {
```
kimallen, line 120: [nit] I wonder if we can simplify this name to `hasPpm`

src/pages/MyMove/SelectMoveType.jsx
suzubara: I would argue that this value should not be saved in `SelectMoveType` state at all, and that `move.selectedMoveType` can be saved in the DB (and Redux) when the user clicks "Next". But how are we resolving this field on the move now that a move can have multiple shipments of varying types?

src/pages/MyMove/Home/index.jsx
sojeri: for this PR. my logic change above means that this string will be displaying incorrectly on the button if not resolved. fixing it now. :)

kimallen: `this.hasShipment` already checks if there is a Ppm or not, so it does not need `this.PpmAlreadyExist` added.

src/pages/MyMove/SelectMoveType.jsx
kimallen: A more accurate name for the field would be `lastSelectedMoveType`.  I think what you said is correct.  Currently, if a user started entering their ppm (since there are multiple pages), and log out in the middle, then when they log back in, it will take them back to the last unfinished page (`selectedMoveType` is used in `getWorkflowRoutes` in helping determine which pages in the flow are completed.

src/pages/MyMove/Home/index.jsx
```diff
@@ -107,7 +107,7 @@ class Home extends Component {
     return !!Object.keys(orders).length && !uploadedOrderDocuments.length;
   }
 
-  get hasShipment() {
+  get hasMtoShipments() {
     const { mtoShipments, currentPpm } = this.props;
     return (this.hasOrders && !!mtoShipments.length) || Object.keys(currentPpm).length;
```
kimallen, line 112: Now that you changed the title to include MTO, do you want to remove the check for Ppms and adjust anywhere else that would be impacted?

src/pages/MyMove/SelectMoveType.jsx
```diff
@@ -28,9 +37,41 @@ export class SelectMoveType extends Component {
...
   render() {
-    const { pageKey, pageList, match, push } = this.props;
+    const { pageKey, pageList, match, push, isPpmSelectable, isHhgSelectable, shipmentNumber } = this.props;
     const { moveType } = this.state;
+    const ppmSelectableContent = (
```
kimallen, line 44: I actually covered all of the content changes on my PR.  It's confusing because the content spreadsheet was linked on both of our stories, but my understanding was that this story was to cover the disabled/enabled states of the selectType page, and my story included the copy changes. https://dp3.atlassian.net/browse/MB-4302

kimallen (changes requested):
I believe I saw all the correctly enabled/disabled radios on the Shipment Selection page.  I also saw on the home page that the Add shipments button was not there when there was a ppm+hhg combo submitted move, it was enabled when they did not yet have a Ppm (whether submitted or not).  That seems correct based on what I can glean from the story and the links.

Requested change: According to the spreadsheet, the home page step 3 should show 'If you need to add shipments, let your movers know.' when the submitted move has a disappeared button. 

Requested change: changes are needed around get `hasMtoShipments` Either the get action or the name need to change-- see comment.

src/pages/MyMove/Home/index.jsx
kimallen: For example, as currently written, it correctly shows the `ShipmentList` if there is any kind of shipment or ppm.  Your name change implies that it does not include Ppm, so either the name needs to be inclusive of Ppms or the check for ppms removed and other uses of `hasMtoShipments` as conditional check need to be adjusted to check for a Ppm.

sojeri: ahhh, nope. I did an automatic rename, and I think it was a mistake! I will revert this part of the change / updated it to be hasAnyShipment for clarity.

src/pages/MyMove/SelectMoveType.jsx
sojeri: So I think there were some crossed wires here and some scope creep on your story, which was supposed to cover only incorporating the new shipment card onto the page and nothing else.

If you look at the description and AC in my story, you can see that the state and content updates were part of MB-4126 (story for this PR):

> Incorporate logic into the shipment selection page that enables, and disables, shipments based on what the user has already requested within the flow.
> - Educational content for customers to understand what they can do about it

I think part of how this happened is our story templates are encouraging adding too many references. If no content changes are indicated, the content spreadsheet shouldn't be in the story. :/

sojeri: Though FWIW I am only changing the disabled copy here, which may be in line with both our stories if you're only updating the enabled copy. In that case, I think we're fine and whoever merges second can deal with the merge conflict.

sojeri: Thinking about this more, and I made a reminder to bring up at either next retro or the one after. I think something went wrong when we split these stories out and we could have done a better job on both for lining out what wasn't covered and then pointing to the other story which should cover it. We might have still had a merge conflict, but at least we'd have been less confused about the overlap. :)

mariaclaratgolfo (changes requested):
I believe we've incorporated the new selectable cards for the Shipment selection page. Is there a reason we're not leveraging them here? 

A few additional notes:

- For the "given: you have already submitted a move not containing a PPM" scenario, once a user submits their move, the edit icon on the Home HHG cell + the edit link in the Review page should be removed. They can no longer make edits to HHG shipments post submission.

- Can you share screenshots of what the Home page and Review page looks like after they've submitted and already requested a PPM? The Add a shipment buttons should be removed from both pages in this scenario. Happy to pair to walk through these if that's easier.

- Echoing Kim's note above, Post-submission — on the Home and Review pages, when they can no longer add another shipment, we have language letting them know "If you need to add shipments, let your movers know."

src/pages/MyMove/SelectMoveType.jsx
kimallen: Yes, I agree.  I think neither story is very clear or delineated between expected tasks.  And maybe one was meant to be finished before the other (they were in different sprints).  We'll figure it all out in the merge conflicts.

src/components/Customer/Home/ShipmentList/ShipmentList.test.jsx
```diff
@@ -47,6 +49,7 @@ describe('ShipmentList component', () => {
     const props = {
       shipments,
       onShipmentClick,
+      moveSubmitted: false,
```
suzubara, line 52: non-blocking, minor suggestion: if you update the `mountShipmentList` fn to return the following, you don't have to worry about passing in default props for each test case:
```
mount(<ShipmentList {...defaultProps} {...props} />)
```

src/components/Customer/Home/ShipmentList/index.jsx
```diff
@@ -6,8 +6,9 @@ import styles from './ShipmentList.module.scss';
 
 import { ReactComponent as EditIcon } from 'shared/icon/edit.svg';
 
-const ShipmentListItem = ({ shipment, onShipmentClick, shipmentNumber, showNumber }) => {
+const ShipmentListItem = ({ shipment, onShipmentClick, shipmentNumber, canEdit, showNumber }) => {
   function handleEnterOrSpace(event) {
+    if (!canEdit) return;
```
suzubara, line 11: this is not related to this PR, but I want to voice a strong opinion that this element should be a `<button>` element, then we don't have to add a keyDown handler and can use the `disabled` attribute to prevent the action.

src/components/Customer/Home/ShipmentList/index.jsx
```diff
@@ -19,7 +20,7 @@ const ShipmentListItem = ({ shipment, onShipmentClick, shipmentNumber, showNumbe
     <div
       className={`${styles['shipment-list-item-container']} ${shipmentClassName}`}
       data-testid="shipment-list-item-container"
-      onClick={() => onShipmentClick(shipment, shipmentNumber, shipment.shipmentType)}
+      onClick={() => onShipmentClick(shipment, shipmentNumber, shipment.shipmentType, canEdit)}
```
suzubara, line 23: minor, non-blocking: I would recommend wrapping `onShipmentClick` handler to implement the `canEdit` early return in one place, rather than having to pass it into the handler itself and also handle it in the keyDown. ie:

```
const handleShipmentClick = () => {
  if (!canEdit) return
  onShipmentClick(shipment, shipmentNumber, shipment.shipmentType)
}

<div onClick={handleShipmentClick}...
```

(but again, this wouldn't be needed if we changed it to a `<button>`!)

src/components/Customer/Home/ShipmentList/index.jsx
```diff
@@ -68,12 +70,14 @@ const ShipmentList = ({ shipments, onShipmentClick }) => {
...
         const shipmentNumber = shipmentNumbersByType[shipmentType];
+        const canEdit = moveSubmitted ? shipmentType === 'PPM' : true;
         return (
           <ShipmentListItem
             key={shipment.id}
             shipmentNumber={shipmentNumber}
             showNumber={shipmentCountByType[shipmentType] > 1}
-            onShipmentClick={() => onShipmentClick(shipment.id, shipmentNumber, shipmentType)}
+            canEdit={canEdit}
+            onShipmentClick={() => onShipmentClick(shipment.id, shipmentNumber, shipmentType, canEdit)}
```
suzubara, line 80: Blocking: this is also not strictly related to this PR, but the way this this handler is passed around is actually a bug IMO. The parent component `<ShipmentList>` is passing a fn that accepts no parameters, and calls `onShipmentClick` into `ShipmentListItem`. That means that the parameters passed into `onShipmentClick` from `ShipmentListItem` are redundant, and are not actually being used at all. The parameters should be passed into the handler in only one place, so one of these implementations should be changed to `onShipmentClick={onShipmentClick}`

src/pages/MyMove/Home/index.jsx
```diff
@@ -335,18 +337,27 @@ class Home extends Component {
...
+                  secondaryBtn={this.hasAnyShipments}
                   secondaryClassName="margin-top-2"
                   step="3"
                 >
-                  {this.hasShipment ? (
-                    <ShipmentList shipments={allSortedShipments} onShipmentClick={this.handleShipmentClick} />
+                  {this.hasAnyShipments ? (
+                    <div>
+                      {this.hasSubmittedMove && !this.doesPpmAlreadyExist && (
+                        <p className={styles.descriptionExtra}>If you need to add shipments, let your movers know.</p>
```
kimallen, line 353: I would have expected this text below the list of shipments where the other messages appear (in place of the text of <Description>, but if the placement is okayed by @mariaclaratgolfo, then fine by me.  The abstract I found didn't contain the text at all, so is unclear.

kimallen (approved):
Changes approved!  left a note about placement of some text, but I don't really know what is desired.

src/pages/MyMove/Home/index.jsx
sojeri: @kimallen yeah, Mariaclara had to link me directly to the mockup. here it is with the text above: https://share.goabstract.com/54bf497e-72bf-44b9-a169-3946e56ff710?collectionLayerId=c963c8d3-0dbb-4f6a-85b8-cb492c62ace1&mode=design

src/pages/MyMove/SelectMoveType.mapStateToProps.test.jsx
```diff
@@ -0,0 +1,90 @@
+// NOTE: mock setup has to happen before import of file with dep that needs mocking
```
suzubara, line 1: I'm curious why you ended up adding this test file instead of testing `ConnectedSelectMoveType`? this is not a pattern I've ever seen before. It also seems like the tests could be part of `SelectMoveType.test.jsx` since usually we've been using one test file to test all exports from a file.

src/pages/MyMove/SelectMoveType.test.jsx
```diff
@@ -33,4 +44,30 @@ describe('SelectMoveType', () => {
     // HHG button should be checked on page load
     expect(wrapper.find(Radio).at(1).find('.usa-radio__input').html()).toContain('checked');
   });
+
+  it('should disable PPM form option if PPM is already submitted', () => {
+    defaultProps.isPpmSelectable = false;
```
suzubara, line 49: Instead of mutating `defaultProps` I think it would follow React paradigm more to do:
```
const wrapper = mount(<SelectMoveType {...defaultProps} isPpmSelectable={false} />);
```

src/components/Customer/Home/ShipmentList/index.jsx
sojeri: yessssss so much. I didn't want to refactor now, but as I started looking at these click and event handlers I was feeling the same thing.

src/components/Customer/Home/ShipmentList/index.jsx
sojeri: yeah, I see what you mean. I was trying to do the least work here because I agree it should be a button. I've made a slight adjustment to move all the canEdit early returns to the ShipmentList file. let me know what you think!

src/pages/MyMove/SelectMoveType.mapStateToProps.test.jsx
sojeri: I separated them to keep this import mocking from altering or complicating anything for the other tests.

as for not testing `ConnectedSelectMoveType`, I had no idea that was a thing. still learning redux! if you think it's worth blocking this PR and Kim's, I can look into refactoring this to test `ConnectedSelectMoveType` instead.

I will say, though, I like being able to test this logic directly rather than checking for side effects in the resulting rendered component.

src/components/Customer/Home/ShipmentList/index.jsx
sojeri: I don't understand what you mean. you may need to walk me through this.

what I'm seeing is:
1. in `pages/MyMove/Home`, we have `handleShipmentClick` which has 4 params
2. still in `pages/MyMove/Home`, we pass into ShipmentList a reference to that function
3. in `components/Customer/Home/ShipmentList`, we map the shipments to `ShipmentListItem`s, in the process defining arrow functions which will call the reference `handleShipmentClick` and passing in the relevant 4 args
4. still in `components/Customer/Home/ShipmentList` we have the second entry point into the `handleShipmentClick` ref, this time via the event handler function.

what am I misunderstanding or missing here? :)

sojeri (commented):
@suzubara responded to several of your comments & have pushed changes for those I could handle right away :)

mariaclaratgolfo (approved):
Notes that we talked through adding during our pairing session:

- No padding appearing below "add another shipment message" 
- Mention during Standup that a Design review will be needed for your work on Shipmenet selection page once new cards are in
- Review page — post submission, button disappears and new language

Sounds like you got them all in — in which case, approved

suzubara (approved):

PR #4920, opened 2020-10-06
Update cac migration for shimona

shimonacarvalho:
## Description

Migrating my cac due to renewal.


LeDeep (approved):
ran the migration and checked the secure migration files. Looks good!

PR #4935, opened 2020-10-07
[MB-4422] fix edi error handling

pearl-truss:
## Description

Currently the EDI generator code returns generic errors. This PR updates the code to return more specific errors for the purposes of making it easier for a client to know what type of error status code to throw. 

## Reviewer Notes

There are now 5 types of errors in the EDI generator code

1. `NotFoundError`: for errors after querying the database that are due to "sql: no rows in result set"
2. `QueryError`: for any error that happens after querying a database that doesn't fall under a `NotFoundError`
3. `BadDataError`: thrown when the code needs a certain field to not be `nil`
4. `InvalidInputError`: thrown when a function returns an error
5. `fmt.Errorf`: generic error thrown in use cases not covered above. Currently only used to handle errors from `strconv` methods

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4422) for this change
* [API Error Documentation](https://github.com/transcom/mymove/wiki/API-Errors#constructing-errors-and-responses) explains more about the approach used.


jacquelineIO (commented):
@pearl-truss  this looks good. The one thing is missing is that the  caller of the function `Generate` only returns `InternalServerError`... 

```
	edi858c, err := h.GHCPaymentRequestInvoiceGenerator.Generate(paymentRequest, false)
	if err != nil {
		logger.Error(fmt.Sprintf("Error generating EDI segments for payment request ID: %s: %s", paymentRequestID, err))
		return paymentrequestop.NewGetPaymentRequestEDIInternalServerError().WithPayload(payloads.InternalServerError(handlers.FmtString(err.Error()), h.GetTraceID()))
	}

```

pearl-truss:
> @pearl-truss this looks good. The one thing is missing is that the caller of the function `Generate` only returns `InternalServerError`...
> 
> ```
> 	edi858c, err := h.GHCPaymentRequestInvoiceGenerator.Generate(paymentRequest, false)
> 	if err != nil {
> 		logger.Error(fmt.Sprintf("Error generating EDI segments for payment request ID: %s: %s", paymentRequestID, err))
> 		return paymentrequestop.NewGetPaymentRequestEDIInternalServerError().WithPayload(payloads.InternalServerError(handlers.FmtString(err.Error()), h.GetTraceID()))
> 	}
> ```

From my understanding of @reggieriser 's comment the internal server errors being returned are okay since that endpoint isn't used in production https://github.com/transcom/mymove/pull/4852#discussion_r494588943

jacquelineIO:
> From my understanding of @reggieriser 's comment the internal server errors being returned are okay since that endpoint isn't used in production [#4852 (comment)](https://github.com/transcom/mymove/pull/4852#discussion_r494588943)

I understand. I don't have an opinion about the support end point.

jacquelineIO:
@pearl-truss  can you create 2 stories? 1 to handle these errors in the prime API and 1 to handle them in the support API. Then we can discuss if we want to do the support API during planning.

jacquelineIO (approved):
Thank you. Looks good. 🚀

pearl-truss:
> @pearl-truss can you create 2 stories? 1 to handle these errors in the prime API and 1 to handle them in the support API. Then we can discuss if we want to do the support API during planning.

I've created:
https://dp3.atlassian.net/browse/MB-4556 
https://dp3.atlassian.net/browse/MB-4557

And thanks for reviewing the story! I will merge whenever master is passing again

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -39,7 +39,10 @@ func (g ghcPaymentRequestInvoiceGenerator) Generate(paymentRequest models.Paymen
 			Where("id = ?", paymentRequest.MoveTaskOrderID).
 			First(&moveTaskOrder)
 		if err != nil {
-			return ediinvoice.Invoice858C{}, fmt.Errorf("cannot load MTO %s for PaymentRequest %s: %w", paymentRequest.MoveTaskOrderID, paymentRequest.ID, err)
+			if err.Error() == "sql: no rows in result set" {
```
reggieriser, line 42: Can we use this constant instead of the test against the explicit string literal? https://github.com/transcom/mymove/blob/ebe96dd7e69c3cd1a10a926278218d6da48c804b/pkg/models/errors.go#L41

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -155,7 +168,7 @@ func (g ghcPaymentRequestInvoiceGenerator) Generate(paymentRequest models.Paymen
 	// Add service member details to header
 	serviceMemberSegments, err := g.createServiceMemberDetailSegments(paymentRequest.ID, moveTaskOrder.Orders.ServiceMember)
 	if err != nil {
-		return ediinvoice.Invoice858C{}, fmt.Errorf("Could not create service: %s", err)
+		return ediinvoice.Invoice858C{}, services.NewInvalidInputError(paymentRequest.ID, err, nil, "unable to service member segments")
```
reggieriser, line 171: Nit: Think the error message is missing a word -- "create" maybe?

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -165,14 +178,17 @@ func (g ghcPaymentRequestInvoiceGenerator) Generate(paymentRequest models.Paymen
...
+		}
+		return ediinvoice.Invoice858C{}, services.NewQueryError("PaymentServiceItems", err, fmt.Sprintf("Could not find payment service items: %s", err))
 	}
 
 	if !msOrCsOnly(paymentServiceItems) {
 		var g62Segments []edisegment.Segment
 		g62Segments, err = g.createG62Segments(paymentRequest.ID)
 		if err != nil {
-			return ediinvoice.Invoice858C{}, err
+			return ediinvoice.Invoice858C{}, services.NewInvalidInputError(paymentRequest.ID, err, nil, "unable to create G62 segments")
```
reggieriser, line 191: In this case, aren't we wrapping an `InvalidInputError` inside of another `InvalidInputError`?  I would think that if the private method we're calling is returning the right error types, we could just return `err` here.  We're making the assumption that every error in the private method is actually an `InvalidInputError`.  This may be the case for other private methods too.

reggieriser, line 191: In cases like this where the private method is already returning the appropriate error type, would it be better to return the `err` unwrapped as opposed to trying to bring them all under one type?  In this case, it looks like a `NotFoundError` or a `NewQueryError` will all get wrapped in an `InvalidInputError` which doesn't seem quite right.

reggieriser: Just to add to this ... I would think the caller of the generator might want to behave differently for a `NewQueryError` (which I'd expect to be a 500 response since that shouldn't normally happen) and an `InvalidInputError` (which might be a 4xx code because the JSON we were given had invalid input, for instance).  I think this extra wrapping could hide that distinction.

pearl-truss: That makes sense and has been updated

reggieriser (approved):
LGTM, pending CI tests eventually passing.  Thanks for making the suggested changes!

PR #4946, opened 2020-10-08
[MB-4529] MyMove API data load

suzubara:
## Description

The goal of this PR is to establish a new pattern for loading API data into the Customer app that is de-coupled from rendering UI, and can be called easily as needed throughout the flow, using Redux actions.

### New pattern is:
- Created [`rootCustomerSaga` for bootstrapping sagas that should only happen in the customer app](https://github.com/transcom/mymove/pull/4946/files#diff-c5bb6481b319d418e4307f67a9b1e8ecR10)
- Wrote [`initializeOnboarding` saga to handle hydrating onboarding user data when a customer logs in](https://github.com/transcom/mymove/pull/4946/files#diff-d41bccd15113a160657933168162f6c9R50). This should only happen once-per-page load, so [it is triggered at the highest level of the customer app (`MyMove/index componentDidMount`)](https://github.com/transcom/mymove/pull/4946/files#diff-6eb50225743dcbdf87fad2a63022f3a4R66), and it handles the following:
  - Get logged in user
  - If user doesn't have a service member, create one
  - If user has moves, load MTO shipments also
  - Populate user data & MTO shipments into Entities
- All API calls should be defined in `src/services` and use the [new `makeSwaggerRequest` function](https://github.com/transcom/mymove/pull/4946/files#diff-eadb579699721add54d5fa7d9853ef3cR78)
  - Existing API calls should be relocated from entities files as needed
  - `makeSwaggerRequest` accepts a `normalize` option (defaults to true) so that API calls can be used to populate both entities _and_ ducks state for backwards compatibility, and to safely proceed with Redux refactor work.

### Proof-of-concepts:
- Replaced `loadMTOShipments` in `HHGShipmentSetup componentDidMount` with new `fetchCustomerData` action, to demonstrate how to refresh _all_ customer data into entities
- Added `initializeOnboarding` action to `MyMove/index componentDidMount` to kick off new onboarding API flow 
- Removed `showLoggedInUser` and `loadMTOShipments` in `MyMove/Home/index componentDidMount, componentDidUpdate` since those calls will happen as part of `initializeOnboarding`
- Removed `createServiceMember` in `MyMove/Home/index componentDidUpdate` since that will happen as part of `initializeOnboarding`
- Added `updateServiceMember` saga to populate service member entities throughout onboarding steps in order to pass tests (this will get cleaned up when service member ducks are refactored)

### Next steps:
- The customer app is calling `getLoggedInUser` twice (once in `initializeOnboarding` and once in `loadUser`) -- clean this up so only one call is made, in a way that is compatible with the office app.
- Identify any other API calls that should be triggered when a customer user logs in (in addition to MTO shipments)
- Identify other places where re-fetch API calls may not be needed anymore and can be deleted, or replaced with `fetchCustomerData` (or something smaller)
- **Continue to replace API calls with sagas so they can update data in ducks & entities simultaneously for backwards compatibility**
  - Once this is completed, we can continue to remove ducks code with more reliability 🎉 

## Reviewer Notes

Test cases:
- Log in as existing MilMove customer & verify your data loads and you are directed to the right place
- Log in as new MilMove customer & verify your service member is created and you are directed to the right place

## Setup

```
make server_run
make client_run
```

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4529) for this change

## Screenshots

Here is a fun diagram I made that illustrates the saga flows: 
image

### DIFF
--- a/src/sagas/onboarding.js
+++ b/src/sagas/onboarding.js
@@ -1,0 +1,32 @@
+import { takeLatest, put, call } from 'redux-saga/effects';
+
+import { INIT_ONBOARDING, initOnboardingFailed, initOnboardingComplete } from 'store/onboarding/actions';
+import { getLoggedInUser, getMTOShipmentsForMove } from 'services/internalApi';
+import { addEntities } from 'shared/Entities/actions';
+
+export function* initializeOnboarding() {
+  try {
+    // First load the user & store in entities
+    const user = yield call(getLoggedInUser);
+    yield put(addEntities(user));
+
+    // TODO - create service member if doesn't exist
+
+    // Load MTO shipments if there is a move
+    const { moves } = user;
+    if (moves && Object.keys(moves).length > 0) {
+      const [moveId] = Object.keys(moves);
+      // User has a move, load MTO shipments & store in entities
+      const mtoShipments = yield call(getMTOShipmentsForMove, moveId);
+      yield put(addEntities(mtoShipments));
+    }
+
+    yield put(initOnboardingComplete());
+  } catch (error) {
+    yield put(initOnboardingFailed(error));
+  }
+}
+
+export function* watchInitializeOnboarding() {
+  yield takeLatest(INIT_ONBOARDING, initializeOnboarding);
+}
--- a/src/sagas/onboarding.js
+++ b/src/sagas/onboarding.js
@@ -6,14 +6,20 @@ import {
   initOnboardingFailed,
   initOnboardingComplete,
 } from 'store/onboarding/actions';
-import { getLoggedInUser, getMTOShipmentsForMove } from 'services/internalApi';
+import {
+  getLoggedInUser,
+  getMTOShipmentsForMove,
+  createServiceMember as createServiceMemberApi,
+} from 'services/internalApi';
 import { addEntities } from 'shared/Entities/actions';
+import { CREATE_SERVICE_MEMBER } from 'scenes/ServiceMembers/ducks';
 
 export function* fetchCustomerData() {
   // First load the user & store in entities
   const user = yield call(getLoggedInUser);
   yield put(addEntities(user));
 
+  // TODO - fork/spawn additional API calls
   // Load MTO shipments if there is a move
   const { moves } = user;
   if (moves && Object.keys(moves).length > 0) {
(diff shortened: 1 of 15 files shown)


src/sagas/onboarding.js
```diff
@@ -0,0 +1,76 @@
...
+    const [moveId] = Object.keys(moves);
+    // User has a move, load MTO shipments & store in entities
+    const mtoShipments = yield call(getMTOShipmentsForMove, moveId);
+    yield put(addEntities(mtoShipments));
+  }
+
+  return user;
+}
+
+export function* watchFetchCustomerData() {
```
Ronolibert, line 36: Are there any other types of methods used in watchers other than taking the latest action fired?

suzubara: there can be! for example, you might use `all` in order to `take` multiple actions in parallel, and you could also use `take` to only handle the action _once_ or `takeEvery` to handle it every time it's called. `takeLatest` is used often in watchers because it's the same as `takeEvery`, but if an action is dispatched multiple times without enough time to complete the handler, only the most recent one will be handled. ultimately, I think the convention is that watcher sagas are use to `take` actions and trigger a handler for them - [this is how the Redux Saga docs define it](https://redux-saga.js.org/docs/Glossary.html)

src/scenes/ServiceMembers/ducks.js
```diff
@@ -121,7 +121,7 @@ export function serviceMemberReducer(state = initialState, action) {
   switch (action.type) {
     case GET_LOGGED_IN_USER.success:
       return Object.assign({}, state, {
-        currentServiceMember: reshape(action.payload.service_member),
+        currentServiceMember: reshape(action.payload.service_member) || state.currentServiceMember,
```
gidjin, line 124: I'm not clear how ducks work, but why do we need to do an `|| state.currentServiceMember`? Is this to handle a case when this has been triggered again after the user is already logged in?

gidjin (approved):
I was able to log in locally as an existing customer and a new customer and things looked good. Overall the pattern seems better to me.

src/sagas/onboarding.js
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export function* updateServiceMember(action) {
+  const { payload } = action;
+  const normalizedData = normalizeResponse(payload, 'serviceMember');
+  yield put(addEntities(normalizedData));
+}
+
+export function* watchUpdateServiceMember() {
+  yield takeLatest('UPDATE_SERVICE_MEMBER_SUCCESS', updateServiceMember);
```
Ronolibert, line 47: [nit] We could use `UPDATE_SERVICE_MEMBER.success` here from `src/scenes/ServiceMembers/ducks.js`

Ronolibert (approved):
Looks good!

src/scenes/ServiceMembers/ducks.js
suzubara: I ended up adding this because of an edge case that was causing spurious integration test failures :( because `getLoggedInUser` is currently called twice in the customer app (which is a known issue that I want to resolve subsequently...), sometimes the second call would resolve _after_ service member creation would complete (when a new user logs in), and end up overwriting the newly created service member. So this is a less than ideal bandaid that will hopefully be resolved when either we fix the double log in call, or refactor service member ducks, whichever happens first!

src/sagas/onboarding.js
```diff
@@ -0,0 +1,76 @@
...
+export function* fetchCustomerData() {
+  // First load the user & store in entities
+  const user = yield call(getLoggedInUser);
+  yield put(addEntities(user));
+
+  // TODO - fork/spawn additional API calls
+  // Load MTO shipments if there is a move
+  const { moves } = user;
+  if (moves && Object.keys(moves).length > 0) {
+    const [moveId] = Object.keys(moves);
```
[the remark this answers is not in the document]
suzubara: follow up convo at: https://ustcdp3.slack.com/archives/CTNTFJSBA/p1602784247489900

PR #4948, opened 2020-10-08
[MB-4457] TOO Move Queues API Endpoint

duncan-truss:
## Description

This PR creates a new endpoint in the GHC api for the TOO moves queue.  The handler authorizes the office user id from the session request before querying the moves from the move order service fetcher.  Once the list of moves is returned it is transformed into the queue payload and sent as the response.

## Reviewer Notes

I left out any handler parameters because we aren't yet doing any filtering, sorting, or paginating of the results.

We're strictly only returning moves that have a matching GBLOC.  What should be done for the default 'XXX' GBLOC that doesn't match any transportation offices?

I am only fetching moves with selectedMoveType HHG, is this the right behavior?

I did not attempt to work out the status field for the queue yet because that seemed a bit too complex at this stage, I'm assuming that work will be captured elsewhere.

I decided since the DoDID is part of the Customer that we can then pull out the name and DoDID from that object rather than only return those 2 fields making it less prescriptive.

I added the page, perPage, and totalCount fields to the response but am not populating them yet.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4457) for this change
* [GHC Queues API Design Doc](https://docs.google.com/document/d/1b6p39UL36dSJtwWWqHZ3bvIDlRqegazBYEkvkD2Zqtk/edit#) 

## Screenshots

Screen Shot 2020-10-08 at 6 11 50 PM

Screen Shot 2020-10-09 at 10 54 49 AM


duncan-truss:
Let me try to add instructions for manually testing the endpoint since there is no way to call it yet.

Ryan-Koch (approved):
Nice work! I agree with the choice on holding off on worrying about the status for now, we can handle that in an upcoming story. Also very much appreciate test coverage of the potential http responses.

noelledusahel (approved):

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	response := handler.Handle(params)
+	suite.IsNotErrResponse(response)
+
+	suite.Assertions.IsType(&queues.GetMovesQueueOK{}, response)
+	payload := response.(*queues.GetMovesQueueOK).Payload
+
+	suite.Len(payload.Results, 0)
+}
```
monfresh, line 207: These tests looks great. My one comment would be to avoid using the DB. Perhaps that can be a follow-up ticket we can pick up to show how to use the `Stub` assertion I showed today.

duncan-truss: Good idea, I'd be happy to explore that pattern more going forward.  I remember doing that with factorybot only returning the stub when it wasn't necessary.  Maybe we can confine it to the create methods so it's in fewer places.

PR #4950, opened 2020-10-09
MB-3109, MB-4542, MB-4543 | refactor create hhg form & create new nts(r) forms

sojeri:
## Description

original checklist is still here but hidden. click `... > edit` to reveal. :)

## Reviewer Notes

1. Kim has already signed off on the shipment selection page changes. These are not entirely related to this change but were needed to unblock manual testing of this PR and also just to cleanup some places where we were calculating the same thing for a second or third time.

2. Another change I made was to add new NTS/NTSr types for the server and FE. The server side added them to `selectedMoveType`, a change which is required to unblock moving from the shipment selection page to the create form. The FE side is just the value used from the server. These could have been different, but I went for the easy route. In some of the other stories like adding NTS/r support in the home page's ShipmentList--- the work will now be converting this raw type to a more human readable one. __I need to sync with other teams on this one, because it has changed a lot of TIO/TOO components as well which were built on the original selectedMove.NTS == 'NTS' value.__

3. The new high level components are all in storybook now, so you can test by manually setting things up OR by playing with storybook. Do whatever you prefer! I kept most of the component tests for these very simple-- just verifying that child components were as expected.

4. The create/edit merge we discussed (during last week's sprint planning IIRC)--- I intend to tackle that in a trailing PR along with Suz's requested changes to remove redux connectors from the components in src/components.

## Setup

...

## References

* [parent user story: SM can enter in their NTS and NTSr shipment details](https://dp3.atlassian.net/browse/MB-3019)
  * [sub task: refactor hhg form](https://dp3.atlassian.net/browse/MB-4542)
  * [sub task: use new components to create new nts(r) forms](https://dp3.atlassian.net/browse/MB-4543)

## Screenshots

...


mariaclaratgolfo:
Are any screenshots or visual changes meant to be attached here for review? Or does this PR approval not require frontend review?

src/components/Customer/MtoShipments/DeliveryDetails.jsx
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
```
suzubara, line 1: can this be called `DeliveryDetailsFields` so the name is easy to distinguish from a component that might just display read-only delivery details? I think that would be a good convention to start for components that are forms/inputs/fieldsets

src/components/Customer/MtoShipments/HHGDetailsForm.jsx
```diff
@@ -0,0 +1,252 @@
...
+};
+
+const mapDispatchToProps = {
+  createMTOShipment: createMTOShipmentAction,
+  updateMTOShipment: updateMTOShipmentAction,
+  showLoggedInUser: showLoggedInUserAction,
+};
+
+export { HHGDetailsForm as HHGDetailsFormComponent };
+export default connect(mapStateToProps, mapDispatchToProps)(HHGDetailsForm);
```
suzubara, line 252: I know this was already like this, but since you're refactoring anyways... I think we should make a rule that no files in `src/components` should be connected to Redux (or reliant on any other global app context). These props should be passed in by the parent component instead, in order to more consistently separate logic from UI. It should also simplify testing & rendering components in Storybook.

src/components/Customer/MtoShipments/HHGDetailsForm.test.jsx
```diff
@@ -4,7 +4,7 @@ import { mount } from 'enzyme';
 import { Provider } from 'react-redux';
 import { ConnectedRouter } from 'connected-react-router';
 
-import { history, store } from '../../shared/store';
+import { history, store } from '../../../shared/store';
```
suzubara, line 7: it's worth noting you _can_ change this to `src/shared/store` since we have absolute imports configured..

src/components/Customer/MtoShipments/NTSDetailsForm.jsx
```diff
@@ -0,0 +1,221 @@
...
+};
+
+const mapDispatchToProps = {
+  createMTOShipment: createMTOShipmentAction,
+  updateMTOShipment: updateMTOShipmentAction,
+  showLoggedInUser: showLoggedInUserAction,
+};
+
+export { NTSDetailsForm as NTSDetailsFormComponent };
+export default connect(mapStateToProps, mapDispatchToProps)(NTSDetailsForm);
```
suzubara, line 221: same comment re: above about not using Redux connect inside `src/components`

src/components/Customer/MtoShipments/NTSrDetailsForm.jsx
```diff
@@ -0,0 +1,177 @@
...
+};
+
+const mapDispatchToProps = {
+  createMTOShipment: createMTOShipmentAction,
+  updateMTOShipment: updateMTOShipmentAction,
+  showLoggedInUser: showLoggedInUserAction,
+};
+
+export { NTSrDetailsForm as NTSrDetailsFormComponent };
+export default connect(mapStateToProps, mapDispatchToProps)(NTSrDetailsForm);
```
suzubara, line 177: same as above

src/components/Customer/MtoShipments/formTypes.js
```diff
@@ -0,0 +1,41 @@
+import * as Yup from 'yup';
```
suzubara, line 1: I would like to differentiate between the types we've been defining in `src/types` (which are just abstractions built on top of `PropTypes`) and form validation schema. while they have similarities, they're also not 100% the same and are used for different purposes. Perhaps we can call this file something like `validationSchemas`?

src/components/Customer/MtoShipments/propShapes.js
```diff
@@ -0,0 +1,23 @@
+import { string, shape } from 'prop-types';
```
suzubara, line 1: I think all types should be defined globally, so this should be in `src/types` (and organized accordingly.. ie, added to `src/types/address` and so on)

src/components/Customer/MtoShipments/utils.js
```diff
@@ -0,0 +1,63 @@
+import { isEmpty } from 'lodash';
```
suzubara, line 1: same as above comment for utils.. can this be moved to `src/utils`?

src/components/Customer/MtoShipments/utils.js
```diff
@@ -0,0 +1,63 @@
...
+
+import { MTOAgentType } from 'shared/constants';
+import { formatSwaggerDate } from 'shared/formatters';
+
+function formatAgent(agent) {
+  const agentCopy = { ...agent };
+  Object.keys(agentCopy).forEach((key) => {
+    /* eslint-disable security/detect-object-injection */
```
suzubara, line 9: You can pass this lint rule by asserting the type of `key`, i.e.: `agentCopy[`${key}`]

suzubara (commented):
I know this is a WIP but I left some initial comments! I also wonder about the usefulness of nesting inside of `Customer/MtoShipments` .. that strikes me as a data model-oriented way to organize the code, and IMO F/E code is more easily organized around a user flow or concern. In this case I might suggest something like `Customer/Onboarding` or `Customer/Forms`

sojeri:
> Are any screenshots or visual changes meant to be attached here for review? Or does this PR approval not require frontend review?

@mariaclaratgolfo this is a draft PR which is not ready for review.

src/components/Customer/MtoShipments/propShapes.js
sojeri: why do we want global prop shapes rather than having them defined alongside their components?

kimallen (commented):
Changes to `SelectMoveType` and its tests look good.  Thanks for noticing the refactor needs.

src/components/Customer/MtoShipments/HHGDetailsForm.jsx
sojeri: I agree this is a much better pattern. given how much I'm already changing, though, I'd like to do this in the trailing PR to refactor the EditShipment component out of existence. will that work for you? I'd be picking up that work as soon as this PR merges / is ready to merge, so it's not punting it for long & will probably still happen this week.

src/components/Customer/MtoShipments/utils.js
sojeri: not sure what you mean. I'll follow up tomorrow. :)

src/components/Customer/MtoShipments/HHGDetailsForm.jsx
suzubara: Yup sounds good!

src/components/Customer/MtoShipments/propShapes.js
suzubara: good question... I guess it depends on how we are viewing them -- as a defined type that represents data models that are shared throughout the app, or one that is bespoke to a specific component. I don't think it has to be either/or, for example I just remembered that I went against my own rule when I created `types` for the `DocumentViewer` component. I guess these shapes (addresses & agents) seem like they may be used in other places as well, so it seemed preferable to defining duplicate addressShapes in multiple places.

suzubara (commented):
I'm not completely sure what the end goal is for the create/edit refactor, but I do want to say while they should use the same form components, I do think it's important to still differentiate between wizard pages (create) vs. non-wizard pages (edit) -- especially assuming the routes for creating a shipment & editing a shipment will be different.

src/components/Customer/MtoShipments/HHGDetailsForm.jsx
```diff
@@ -0,0 +1,248 @@
...
+});
+
+class HHGDetailsForm extends Component {
+  constructor(props) {
+    super(props);
+    const hasDeliveryAddress = get(props.mtoShipment, 'destinationAddress', false);
+    this.state = {
+      hasDeliveryAddress,
+      useCurrentResidence: false,
+      initialValues: {},
```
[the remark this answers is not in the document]
sojeri: this was an existing pattern. I'm with you on it feeling weird to have this in state and to have these onChange handlers, but without these--- what's the right way to prefill data in the form depending on a checkbox or radio button?

sandy-wright (approved):
Looks good from the Prime perspective! Our endpoint is still working smoothly 👍

suzubara (approved):
outstanding changes to be made in future refactor efforts

PR #4958, opened 2020-10-09
[MB-4517 and MB-3904] Select NTS move

Ronolibert:
## Description

Add NTS and NTSR options to shipment selection screen and descriptions

## Reviewer Notes
- Both options are NTS for now as we cannot save an NTSR shipment right now. There is another task to add NTSR as a shipment type

## Setup
`make client_run`
`make server_run`
Go through customer flow and see 2 additional shipment types

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4517) and [other jira story](https://dp3.atlassian.net/browse/MB-3904) for this change

## Screenshots
image


mariaclaratgolfo (approved):
🎉🎉🎉

src/scenes/MyMove/getWorkflowRoutes.jsx
```diff
@@ -220,6 +221,11 @@ const pages = {
       <HHGShipmentSetup pageList={pages} pageKey={key} match={match} history={history} />
     ),
   },
+  '/moves/:moveId/nts-start': {
+    isInFlow: (state) => inHhgFlow && state.selectedMoveType === SHIPMENT_OPTIONS.NTS,
```
sojeri, line 225: it looks like [this const](https://github.com/transcom/mymove/blob/master/src/shared/constants.js#L60) is holding on to the wrong value (`NTS` instead of `HHG_INTO_NTS_DOMESTIC`). can you update the shipment options constant to include the correct  NTS and NTSr types?

here's the go model ref with the correct values:
- https://github.com/transcom/mymove/blob/5707bda7a9f368476b006c82f043684e017edcea/pkg/models/mto_shipments.go#L28
- https://github.com/transcom/mymove/blob/5707bda7a9f368476b006c82f043684e017edcea/pkg/models/mto_shipments.go#L29
- https://github.com/transcom/mymove/blob/5707bda7a9f368476b006c82f043684e017edcea/pkg/models/mto_shipments.go#L30
- https://github.com/transcom/mymove/blob/5707bda7a9f368476b006c82f043684e017edcea/pkg/models/mto_shipments.go#L31

cypress/integration/mymove/nts.js
```diff
@@ -0,0 +1,150 @@
...
+    cy.route('POST', '/internal/service_members').as('createServiceMember');
+    cy.route('PATCH', '**/internal/mto-shipments/**').as('patchShipment');
+    cy.route('GET', '/internal/moves/**/mto_shipments').as('getMTOShipments');
+    cy.route('GET', '/internal/users/logged_in').as('getLoggedInUser');
+  });
+
+  it('Sets up an NTS shipment', function () {
+    cy.signInAsNewMilMoveUser();
+    customerFillsInProfileInformation();
+    customerFillsOutOrdersInformation();
```
sojeri, line 20: is there a seed account we can use which already has profile/orders?

or else is there a way to import similar methods from another integration test so these don't need to be redefined?

src/pages/MyMove/SelectMoveType.jsx
```diff
@@ -138,8 +141,30 @@ export class SelectMoveType extends Component {
                   ? 'How do you want this group of things moved?'
                   : 'How do you want to move your belongings?'}
               </h1>
+              <h2>Choose 1 shipment at a time.</h2>
+              <p>You can add more later</p>
               {hasPpm ? selectPpmHasPpm : selectPpmHasNoPpm}
               {hasSubmittedMove ? selectHhgSubmittedMove : selectHhgDefault}
+              <h3>Long-term storage</h3>
+              <p>These shipments do count against your weight allowance for this move.</p>
+              <SelectableCard
```
sojeri, line 150: do these cards not have changed text when they're no longer valid options? eg, HHG card changes to `Talk with your movers directly if you want to add or change shipments.` when customer can no longer add more HHG shipments.

src/scenes/MyMove/getWorkflowRoutes.jsx
sojeri: or if you don't want to make this change I can make it in my create NTS/r branch :)

cypress/integration/mymove/nts.js
Ronolibert: I had a seed account, but it uncovered some session problems that were going to dig me into a rabbit hole specifically with the feature flag interactions. I can extract these methods to a separate file for reuse

src/pages/MyMove/SelectMoveType.jsx
Ronolibert: From the copy doc, they do have separate text if the user has already submitted an nts(r) but I figured we would add that when we can add that information. I'll add a boolean that we can fill in later when we have real data

cypress/integration/mymove/hhg.js
```diff
@@ -1,4 +1,4 @@
-import { fileUploadTimeout } from '../../support/constants';
+import { customerFillsInProfileInformation, customerFillsOutOrdersInformation } from './utilities/customer';
```
sojeri, line 1: thanks :)

src/pages/MyMove/SelectMoveType.module.scss
```diff
@@ -21,7 +21,14 @@ h1.selectTypeHeader.header {
...
 
 .gridContainer {
-  :global(.grid-row){
+  :global(.grid-row) {
     @include u-padding-top(3);
   }
+
+  h2 + p,
+  h3 + p {
+    @include u-margin-top('05');
```
sojeri, line 30: lolsob at `'05'` and `3`

(no action necessary)

sojeri (commented):
looks good to me aside from nts/r type which it sounds like I'll be handling in my create PR.

double checking now that we have a story for coming back to this page to fix the disabled states. will approve after confirming we have that work tracked somewhere. :)

sojeri (approved):

PR #4959, opened 2020-10-09
MB-4061 Filter payment requests by GBLOC

monfresh:
## Description

When a TIO signs in, they should only see the payment
requests associated with a transportation office that has the same
GBLOC as the transporation office the office user belongs to.

This PR also adds authorization to the list payment requests endpoint
so that it only allows users who have the TIO role.

In order to speed up tests and avoid using the DB when not needed, I
added a new top-level `Stub` assertion type to `testdatagen` so that
you can bypass the DB when setting it to `true`. The testdatagen
functions are very convenient, but sometimes we just want plain models
without saving them to the DB. By setting the `Stub` assertion, we can
take advantage of that convenience without having to write the models
from scratch in tests.

## Reviewer Notes

Steps to test:
1. Run the setup commands in the section below
2. Visit http://officelocal:3000/devlocal-auth/login
2. Click the `Login` button next to `tio_role@office.mil (PPM office)`
3. Verify that you can see payment requests
5. Sign in to http://milmovelocal:3000 with your `@truss.works` email that you use for testing
6. Sign in to the admin interface and create a new office user for that same email, and assign it to a transportation office whose GBLOC is not "LKNQ", then add the TOO and TIO roles.
7. Sign in to http://officelocal:3000
8. Click the "change user roles" link in the top left
9. Click the TIO button

Expected Result: you should not see any payment requests

## Setup

```sh
make db_dev_e2e_populate
make server_run
make office_client_run (in a separate terminal tab)
```

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4061) for this change

## Screenshots


pkg/handlers/ghcapi/payment_request.go
```diff
@@ -31,13 +33,25 @@ type ListPaymentRequestsHandler struct {
 
 // Handle lists payment requests
 func (h ListPaymentRequestsHandler) Handle(params paymentrequestop.ListPaymentRequestsParams) middleware.Responder {
-	// TODO: add authorizations
-	logger := h.LoggerFromRequest(params.HTTPRequest)
+	request := params.HTTPRequest
+	session, logger := h.SessionAndLoggerFromRequest(request)
+	var officeUserID uuid.UUID
```
duncan-truss, line 38: If we aren't setting this inside of a conditional flow I think we can declare and assign it all in the same line below

pkg/handlers/ghcapi/payment_request.go
```diff
@@ -31,13 +33,25 @@ type ListPaymentRequestsHandler struct {
...
+	}
+
+	officeUserID = session.OfficeUserID
 
-	paymentRequests, err := h.FetchPaymentRequestList()
+	paymentRequests, err := h.FetchPaymentRequestList(officeUserID)
 	if err != nil {
-		logger.Error("Error listing payment requests err", zap.Error(err))
-		return paymentrequestop.NewListPaymentRequestsInternalServerError()
+		logger.Error("listing payment requests for", zap.String("office user ID", officeUserID.String()), zap.Error(err))
```
duncan-truss, line 48: I thought if you're adding fields to the logger the zap.String key of officeUserID should be a camel or snake case key?  Or did you really want to add it to the message?

pkg/testdatagen/make_office_user.go
```diff
@@ -47,7 +49,9 @@ func MakeOfficeUser(db *pop.Connection, assertions Assertions) models.OfficeUser
 
 	mergeModels(&officeUser, assertions.OfficeUser)
 
-	mustCreate(db, &officeUser)
+	if assertions.Stub != true {
```
duncan-truss, line 52: shouldn't need the compare to boolean

migrations/app/schema/20201012143646_add_index_to_gbloc.up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX transportation_offices_gbloc_idx ON transportation_offices USING btree(gbloc);
```
duncan-truss, line 1: 👍

pkg/services/payment_request/payment_request_list_fetcher_test.go
```diff
@@ -3,25 +3,64 @@ package paymentrequest
...
+	paymentRequestListFetcher := NewPaymentRequestListFetcher(suite.DB())
+	officeUser := testdatagen.MakeDefaultOfficeUser(suite.DB())
 
-	suite.T().Run("Successful fetch of payment requests", func(t *testing.T) {
-		expectedPaymentRequests := []models.PaymentRequest{}
-		numToMake := 3
-		for i := 0; i < numToMake; i++ {
-			testdatagen.MakeDefaultPaymentRequest(suite.DB())
-		}
+	suite.T().Run("No results when GBLOC does not match", func(t *testing.T) {
```
duncan-truss, line 36: Your handler says a 404 NotFound might be returned but here there is no error with empty results.  Which did we want it to be?

duncan-truss (approved):
LGTM, no blocking comments.

Ryan-Koch (approved):

pkg/testdatagen/make_office_user.go
monfresh: That's what I thought, but when I did `!assertions.Stub` the first time around, it didn't work for some reason. I'll try it again.

pkg/handlers/ghcapi/payment_request.go
monfresh: Good point! I was probably following another similar example, like `zap.String("New station name", newStation.Name)`. Looking through our usage of this, we're not being consistent at all. I've seen pretty much every combination: `user-agent`, `userID`, `user_id`, `tlsVersion`, `ID`. I wonder if we should pick one and stick with it?

pkg/services/payment_request/payment_request_list_fetcher_test.go
monfresh: Thanks for the reminder. I was gonna fix that. It looks like if there aren't matches with the gbloc, no error is returned, so I don't think we need the 404 case. What do you think about the frontend implementing logic that displays something like "no payment requests found for your GBLOC" when the array is empty?

duncan-truss: Yea that makes sense, I think that's an open question for design what to do if there are no results, or there is an error, etc.

PR #4983, opened 2020-10-13
Allow finding matching duty stations and TOs

monfresh:
In the October slice, TTV is working on filtering the TOO and TIO
queues based on the GBLOC. When we demo this work, we have to make
sure that the Origin Duty Station chosen by the service member
corresponds to a transportation office with the same GBLOC as the
one the TOO and TIO are assigned to.

Similarly, to demo the scenario where the GBLOCs don't match, we need
to choose a transportation office for the TOO and TIO that is in a
different GBLOC than the one associated with the duty station.

To make it easier to find the names of matching transportation
offices or duty stations, I've created these two scripts.

The `ds-for-gbloc` script expects a transportation office name, and it
will show you all the duty stations with a matching GBLOC.

The `to-for-gbloc` script expects a duty station name, and it
will show you all the transportation offices with a matching GBLOC.

A usage example during demo prep would be:

1. A TOO/TIO already exists, and is assigned to transportation office
X
2. The person playing the role of the service member will run
`ds-for-gbloc <name of transportation office from step 1>`, and will
pick any of the results as their origin duty station

Alternatively:

1. The person playing the role of the service member will agree to
pick duty station X
2. The person playing the role of the TOO/TIO will run
`to-for-gbloc <name of duty station from step 1>` and update their
transportation office (via the admin interface) to any office in the
script results.

## Reviewer Notes

Is there a better name for the scripts?

## Usage example

```sh
scripts/to-for-gbloc "Dover AFB"
```

## Screenshots

Screen Shot 2020-10-13 at 5 07 01 PM

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -94,6 +94,7 @@ application testing
 | Script Name                     | Description                                                                                       |
 | ------------------------------- | ------------------------------------------------------------------------------------------------- |
 | `check-docker-size`             | Script to check the available disk space Docker has used                                          |
+| `ds_for_gbloc`                  | Helper script to find duty stations with the same GBLOC as the given transportation office        |
 | `ensure-application`            | Ensure APPLICATION is set to `app` or `orders` and matches input value                            |
 | `export-obfuscated-tspp-sample` | Export a subset of rows from the `transportation_service_provider_performances` table             |
 | `find-invoices`                 | This script will use available API endpoints to find invoices in whatever environment you specify |
@@ -103,6 +104,7 @@ application testing
 | `make-test`                     | A script to test common developer make targets.                                                   |
 | `prime-api`                     | A script to connect to endpoints on the Prime API.                                                |
 | `prime-api-demo`                | A script to demo the Prime API.                                                                   |
+| `to_for_gbloc`                  | Helper script to find transportation offices with the same GBLOC as the given duty station        |
 | `update-docker-compose`         | Update branch name before running docker-compose                                                  |
 
 ### Building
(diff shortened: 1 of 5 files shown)


scripts/README.md
```diff
@@ -94,6 +94,7 @@ application testing
 | Script Name                     | Description                                                                                       |
 | ------------------------------- | ------------------------------------------------------------------------------------------------- |
 | `check-docker-size`             | Script to check the available disk space Docker has used                                          |
+| `ds_for_gbloc`                  | Helper script to find duty stations with the same GBLOC as the given transportation office        |
```
duncan-truss, line 97: I think the rest of these scripts use kebab case

scripts/ds_for_gbloc
```diff
@@ -0,0 +1,25 @@
...
+  exit 1
+fi
+
+name=$1
+
+printf "These are the names of duty stations with the same GBLOC as %s: \\n\\n" "$name"
+
+# The -t option prints just the names as opposed to the default formatting that
+# includes the column name underlined
+psql -t postgres://postgres:mysecretpassword@localhost:5432/dev_db << EOF
```
duncan-truss, line 16: Seems like these db values are readable from the environment instead of hardcoding?

scripts/README.md
monfresh: Done!

scripts/ds_for_gbloc
monfresh: Done!

duncan-truss (approved):

PR #4988, opened 2020-10-14
truss-sr-eng own ADRs

tinyels:
## Description

Giving ownership of ADRs to the github group Truss-Sr-Eng which includes team leads and all level 4 app engs.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -7,7 +7,7 @@
 # Add language specific code owners if it becomes relevant
 
 # ADRs are architectural decisions records that should all be approved by all teams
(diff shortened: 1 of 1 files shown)


lynzt (approved):
🚢 

question: should we have an infra peeps in this group so they know what we're up to or is this app-eng specific?

tinyels:
> 🚢
> 
> question: should we have an infra peeps in this group so they know what we're up to or is this app-eng specific?

@esacteksab is in this group for awareness, but I think for the most part ADRs are the responsibility of the app team. Obviously, if there is an ADR that affects Infra (or any specific team or group), we should tag them too.

PR #4991, opened 2020-10-14
MB-4158 too queue status field

Ryan-Koch:
## Description
Adds in ability to translate status to needed values in the queue payload. Also enables the TOO shipment approval to change the move status to approved. Lastly, filters out non SUBMITTED and APPROVED moves from what's being provided by the service.

## Setup

Make sure you run a new e2e populate (like below). I made a status adjustment to one of the moves so we'd wind up with two in the queue after the filtering is applied. Afterward, login to the office app as a TOO user and then navigate over to `/moves/queue` in order to see the queue with the statuses. At first you should see both moves as 'New move' for their status. You'll have to approve the shipments on one to see it change to 'Approvals requested' and once you approve all the service items it will switch to 'Move approved.'

```sh
make db_dev_e2e_populate
```

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4158) for this change


pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -467,16 +467,14 @@ func QueueMoves(moveOrders []models.Order) *ghcmessages.QueueMoves {
...
 		if order.DepartmentIndicator != nil {
 			deptIndicator = *order.DepartmentIndicator
 		}
 		queueMoveOrders[i] = &ghcmessages.QueueMove{
-			ID:       *handlers.FmtUUID(order.ID),
-			Customer: Customer(&customer),
-			// TODO Add status calculation logic here or at service/query level
-			Status:                 ghcmessages.QueueMoveStatus("NEW"),
+			Customer:               Customer(&customer),
+			Status:                 ghcmessages.QueueMoveStatus(queueMoveStatus(hhgMove)),
```
duncan-truss, line 476: I would think the swagger type enum would need updated with these statuses.  You may need to return one as the default instead of `move.Status` in the default case since that wouldn't match.

Ryan-Koch: Oh good point, I did forget to do that. Oddly, swagger didn't yell at me, but it really should reflect what could actually pass through. Will update that and push it up.

pkg/services/mto_shipment/mto_shipment_updater.go
```diff
@@ -524,6 +524,25 @@ func (o *mtoShipmentStatusUpdater) UpdateMTOShipmentStatus(shipmentID uuid.UUID,
 	}
 
 	if shipment.Status == models.MTOShipmentStatusApproved {
+		// If we're approving a shipment, we also need the move the shipment is in to change statuses
+		moveToUpdate := shipment.MoveTaskOrder
+		moveToUpdate.Status = models.MoveStatusAPPROVED
```
duncan-truss, line 529: So if a move has multiple shipments it will be approved even if not all of it's shipments are approved?

Ryan-Koch: Yeah that's how this would operate. I believe that's how it would flow, as this is also the time we're making the information available to the prime to start acting on the move. So to me if a shipment is approved then the move is approved to start having things happen. Totally down to look at doing something different though if I'm off base.

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -39,11 +39,14 @@ func (f moveOrderFetcher) ListMoveOrders(officeUserID uuid.UUID) ([]models.Order
 		"OriginDutyStation",
 		"Entitlement",
 		"Moves.MTOShipments",
+		"Moves.MTOServiceItems",
```
duncan-truss, line 42: This might be an overly complex optimization but it seems possible we could add an Eager("Moves.MTOServiceItems").Where("mto_service_items.status = 'SUBMITTED'").  I notice that status column isn't indexed though.

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -39,11 +39,14 @@ func (f moveOrderFetcher) ListMoveOrders(officeUserID uuid.UUID) ([]models.Order
...
 	).InnerJoin("moves", "orders.id = moves.orders_id").
 		InnerJoin("mto_shipments", "moves.id = mto_shipments.move_id").
 		InnerJoin("duty_stations", "orders.origin_duty_station_id = duty_stations.id").
 		InnerJoin("transportation_offices", "duty_stations.transportation_office_id = transportation_offices.id").
 		Where("transportation_offices.gbloc = ?", gbloc).
+		// TODO: Let's include the status in filters that are passed into this service once we build that feature for the TXO queue (instead of it being hardcoded like it is below right now).
+		Where("moves.status NOT IN ('DRAFT', 'CANCELLED')").
```
duncan-truss, line 49: Not sure it matters is there a reason we're doing the negation instead of `IN ('SUBMITTED', 'APPROVED')`?

pkg/services/move_order/move_order_fetcher.go
Ryan-Koch: That's a fair point, though I think I'm going to leave it for now. I suspect we're going to wind up messing about with this query when we get to the filtering story anyhow.

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
Ryan-Koch: Added the things to swagger!

pkg/services/mto_shipment/mto_shipment_updater.go
duncan-truss, line 529: I think we're encouraged to use the state machine transition for statuses https://github.com/transcom/mymove/blob/43afd3e1f4ffac11f455b52fefe58dfeabb0c9bc/pkg/models/move.go#L155

swagger/ghc.yaml
```diff
@@ -2407,6 +2407,9 @@ definitions:
       - NEW
       - APPROVED
       - APPROVALS_REQUESTED
+      - New move
```
duncan-truss, line 2410: You can remove the duplicated CAPITALIZED enum keys above

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -103,6 +106,42 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandler() {
 	suite.Equal(order.NewDutyStation.ID.String(), result.DestinationDutyStation.ID.String())
 	suite.Equal(hhgMove.Locator, result.Locator)
 	suite.Equal(int64(1), result.ShipmentsCount)
+	suite.Equal(ghcmessages.QueueMoveStatus("New move"), result.Status)
+
+	// let's test for the Move approved status
```
duncan-truss, line 111: I think this makes this test a bit too long can we separate this out into a test for the statuses?

pkg/services/mto_shipment/mto_shipment_updater.go
Ryan-Koch: Sure thing, transitioned that over.

pkg/handlers/ghcapi/queues_test.go
Ryan-Koch: Done

pkg/testdatagen/scenario/e2ebasic.go
```diff
@@ -1029,8 +1030,11 @@ func (e e2eBasicScenario) Run(db *pop.Connection, userUploader *uploader.UserUpl
 	})
 	mto := testdatagen.MakeMove(db, testdatagen.Assertions{
 		Move: models.Move{
```
duncan-truss, line 1032: This move shows up in the queue but has a `SelectedMoveType` of PPM is that what you wanted?

Ryan-Koch: It's what our e2e test is looking for, so I just adjusted it so that it would continue working after the status filter was put into place (removes draft, which is the default for moves that are made in testdatagen). I can change the type though so it makes sense.

duncan-truss (approved):
I tested the different states with the COMBOS move and it works as expected!

donaldthai (approved):
Totally unrelated to your changes, but I noticed:
1. that there's an [extra `if statement` in the fetcher that could be removed](https://github.com/transcom/mymove/blob/215955106e382fb6f6d6bd99c5046aa2aecad2c9/pkg/services/move_order/move_order_fetcher.go#L26). 
2. Took me a while to remember and realize new/submitted moves need to match the `GBLOC` of the office user. Is this something that it'd be better to have a description on the queue that says so?

Otherwise, LGTM

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -84,6 +87,124 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandler() {
...
+		},
+		Order: excludedOrder,
+	})
+
+	testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+		Move: excludedMove,
+		MTOShipment: models.MTOShipment{
+			Status: models.MTOShipmentStatusSubmitted,
+		},
+	})
```
monfresh, line 154: I think lines 131-154 can be simplified to this:
```golang
testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
 		TransportationOffice: models.TransportationOffice{
 			Name:  "Fort Punxsutawney",
 			Gbloc: "AGFM",
 		},
 		MTOShipment: models.MTOShipment{
 			Status: models.MTOShipmentStatusSubmitted,
 		},
        Move: models.Move{
 			SelectedMoveType: &hhgMoveType,
 		},
})
```

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -84,6 +87,124 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandler() {
...
 	suite.Equal(hhgMove.Locator, result.Locator)
 	suite.Equal(int64(1), result.ShipmentsCount)
+
+}
+
+func (suite *HandlerSuite) TestGetMoveQueuesHandlerStatuses() {
+	officeUser := testdatagen.MakeDefaultOfficeUser(suite.DB())
+	officeUser.User.Roles = append(officeUser.User.Roles, roles.Role{
+		RoleType: roles.RoleTypeTOO,
+	})
```
monfresh, line 97: We have a helper method `MakeTOOOfficeUser` if you want to make this less verbose. Also, I would suggest passing in the `Stub: true` option because we don't need this user to be in the database.
```golang
officeUser := testdatagen.MakeTOOOfficeUser(suite.DB(), testdatagen.Assertions{
    Stub: true,
})
```

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -84,6 +87,124 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandler() {
...
+	response = handler.Handle(params)
+	suite.IsNotErrResponse(response)
+
+	suite.Assertions.IsType(&queues.GetMovesQueueOK{}, response)
+	payload = response.(*queues.GetMovesQueueOK).Payload
+
+	result = payload.QueueMoves[0]
+
+	suite.Equal(ghcmessages.QueueMoveStatus("Approvals requested"), result.Status)
+
```
monfresh, line 207: Given that we are testing logic based on various move statuses, it seems like we should be able to test this using mocks and stubs only, without ever needing the DB, right?

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -39,11 +36,14 @@ func (f moveOrderFetcher) ListMoveOrders(officeUserID uuid.UUID) ([]models.Order
 		"OriginDutyStation",
 		"Entitlement",
 		"Moves.MTOShipments",
+		"Moves.MTOServiceItems",
 	).InnerJoin("moves", "orders.id = moves.orders_id").
 		InnerJoin("mto_shipments", "moves.id = mto_shipments.move_id").
 		InnerJoin("duty_stations", "orders.origin_duty_station_id = duty_stations.id").
 		InnerJoin("transportation_offices", "duty_stations.transportation_office_id = transportation_offices.id").
 		Where("transportation_offices.gbloc = ?", gbloc).
+		// TODO: Let's include the status in filters that are passed into this service once we build that feature for the TXO queue (instead of it being hardcoded like it is below right now).
```
monfresh, line 45: Can we please write a JIRA story for this so we don't forget? TODOs in code are generally forgotten. If it's important, it should be in JIRA so it can be tracked.

Ryan-Koch: We have a user story for this filtering work: https://dp3.atlassian.net/browse/MB-4060, I'll add a note there so that this TODO is clear there also.

PR #5005, opened 2020-10-15
[MB-4620] Multiple NTS validation

Ronolibert:
## Description

Add error if move already has a submitted NTS

## Setup
`make server_test`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4620) for this change


pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -69,6 +69,13 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
 		return nil, services.NewNotFoundError(moveID, "for move")
 	}
 
+	for _, mtoShipment := range move.MTOShipments {
```
kimallen, line 72: I don't have strong feelings about it, but wonder if it would be more clear if this was called `existingShipment` instead of `mtoShipment`.

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -69,6 +69,13 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
...
+		if shipment.ShipmentType == models.MTOShipmentTypeHHGIntoNTSDom &&
+			(mtoShipment.ShipmentType == models.MTOShipmentTypeHHGIntoNTSDom && mtoShipment.Status == models.MTOShipmentStatusSubmitted) {
```
kimallen, line 74: I'm wondering why the status has to be Submitted.  If it were in DRAFT state (which I think we won't have unless we split the form), would they be able to create another shipment?  If it is ACCEPTED or REJECTED, can they still create an additional NTS?

pkg/testdatagen/make_mto_shipment.go
```diff
@@ -12,13 +12,23 @@ import (
...
 	}
 
+	if mtoShipment.ShipmentType != "" {
+		shipmentType = mtoShipment.ShipmentType
+	}
+
+	if mtoShipment.Status != "" {
+		shipmentStatus = mtoShipment.Status
+	}
+
```
kimallen, line 31: Nice adjustment on these

kimallen (commented):
Looks good with my eyes (didn't run it). Just leaving comments since I'll be out tomorrow so someone else can dig deeper and approve.

pkg/services/mto_shipment/mto_shipment_creator.go
Ronolibert: Will change this!

pkg/services/mto_shipment/mto_shipment_creator.go
Ronolibert: I think currently, a shipment is in draft as soon as they complete the SelectMoveType step. So this was just guarding that case where they'd select a shipment but never really follow through with it. Draft shipments aren't included in any of the collections that we use for displaying though

lynzt (approved):
LGTM
With the caveat of I can't test this on the front end so relying on unit tests...

lynzt (approved):
LGTM
with the caveat that we're relying on the unit tests and can't test in the app till the create NTS functionality is in the system

pkg/services/mto_shipment/mto_shipment_creator.go
kimallen: Ok, I think that check is fine to have in there.  FWIW, I don't think we actually create a shipment after the shipment type is selected- we update the `selectedMoveType` on the move, but don't create the shipment until the "Next" button is clicked on the form page.

PR #5016, opened 2020-10-16
Create MakeStubbedUser for faster tests

monfresh:
This PR includes a couple of ways to speed up our tests. I tried to
break them down into small chunks in seperate commits to make it
easier to review.

The first change was to create a `MakeStubbedUser` function in
testdatagen that uses the new `Stub` assertion (which was added in
a previous PR) so that it doesn't touch the database. I then replaced
all instances of `MakeDefaultUser` with `MakeStubbedUser` where we
don't actually need a user in the DB.

The second change was to add the `-failfast` and `-vet=off` flags to
our `go test` command so that we don't have to wait for the entire
test suite to finish just to tell us that a test failed. Note that the
failfast flag is only enabled locally. In Circle CI, we want to know
about all test failures at once.

The third commit was to expand the ability to stub models to all the
ones in testdatagen that call `mustCreate`.

From my unscientific benchmarks, the time savings aren't yet significant,
but every little bit helps. If we get in the habit of using stubbed models
when we don't need the DB, it will prevent our tests from slowing down
unnecessarily.

### DIFF
--- a/pkg/testdatagen/make_personally_procured_move.go
+++ b/pkg/testdatagen/make_personally_procured_move.go
@@ -2,6 +2,7 @@ package testdatagen
 
 import (
 	"github.com/gobuffalo/pop"
+	//"github.com/gofrs/uuid"
 
 	"github.com/transcom/mymove/pkg/models"
 )
@@ -34,7 +35,7 @@ func MakePPM(db *pop.Connection, assertions Assertions) models.PersonallyProcure
 	// Overwrite values with those from assertions
 	mergeModels(&ppm, assertions.PersonallyProcuredMove)
 
-	mustCreate(db, &ppm)
+	mustCreate(db, &ppm, assertions.Stub)
 
 	// Add the ppm we just created to the move.ppm array
 	ppm.Move.PersonallyProcuredMoves = append(ppm.Move.PersonallyProcuredMoves, ppm)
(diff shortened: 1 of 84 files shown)


pkg/testdatagen/make_user.go
```diff
@@ -19,13 +19,17 @@ func MakeUser(db *pop.Connection, assertions Assertions) models.User {
...
 }
 
 // MakeDefaultUser makes a user with default values
 func MakeDefaultUser(db *pop.Connection) models.User {
 	return MakeUser(db, Assertions{})
 }
+
+// MakeStubbedUser returns a user without hitting the DB
+func MakeStubbedUser(db *pop.Connection) models.User {
+	return MakeUser(db, Assertions{Stub: true})
```
duncan-truss, line 34: I would think we would still want the ability to pass other assertions unless you only want this to replace MakeDefaultUser.

monfresh: Yes, we only want to replace `MakeDefaultUser`. We want `MakeStubbedUser` to be used when you only need a mock user with the default attributes.

pkg/testdatagen/make_personally_procured_move.go
```diff
@@ -45,7 +46,8 @@ func MakePPM(db *pop.Connection, assertions Assertions) models.PersonallyProcure
 // MakeDefaultPPM makes a PPM with default values
 func MakeDefaultPPM(db *pop.Connection) models.PersonallyProcuredMove {
 	advance := models.BuildDraftReimbursement(1000, models.MethodOfReceiptMILPAY)
-	mustCreate(db, &advance)
+	mustSave(db, &advance)
```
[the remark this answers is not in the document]
monfresh: `mustCreate` now requires a third argument based on Assertions, which would mean requiring `MakeDefaultPPM` to take in that argument as well, and I thought it was easier to change to `mustSave` instead of changing all the places where `MakeDefaultPPM` is called in the tests.

lynzt (approved):
LGTM. 
i'm all for mocking things and not having to hit the db when we can. i'm curious if you looked run times before this change and after on circle?

🚢

monfresh:
I did not look at Circle CI in depth yet. It's on my to-do list to see how to get meaningful data. At a quick glance, I compared the `server_test` portion in this PR to the most recent build that passed in master, and my PR was faster by 17 seconds. I think we need more data though before reaching conclusions.

PR #5020, opened 2020-10-16
[MB-3963] Change comparison of Postgres error strings to error codes

reggieriser:
## Description

As part of the Pop v5 upgrade, the underlying Postgres driver is changing from `lib/pq` to `jackc/pgx`.  Because the driver is changing, the error strings returned by the driver are also changing.  We have some cases where we compare against explicit strings to validate a DB error, so those need to be altered.  But rather than just change the string to another hard-coded string literal, this PR introduces error codes from `jackc/pgerrcode` and compares against them instead, which should be less brittle.  We also use the `pgconn.PgError` error struct to pull out specific table and/or constraint names as well and test against them if appropriate.

For more info about the error checking strategy, [see this Wiki page](https://github.com/jackc/pgx/wiki/Error-Handling).

## Reviewer Notes

- Note that this PR is requesting a merge into `rr-MB-2467-pop5` (the long-running branch for the Pop upgrade), not into master. Eventually we will merge the long-running branch into master after all tests are passing.
- There are a few error string comparisons in `pkg/migrate` and `pkg/services/dbtools` that I didn't change yet (you'll see a TODO, however).  These will be revisited as part of the next story for this feature where we rework the current `lib/pq` dependency in this code.
- I noticed that we have 20 or so cases of comparisons against a `sql: no rows in result set` string that we already have a `models.RecordNotFoundErrorString` constant to represent, so I remapped those.  Note that this particular DB error doesn't originate from the driver, but instead from [`ErrNoRows`](https://golang.org/pkg/database/sql/#pkg-variables) in Go's `database/sql` package.  In fact, we could probably simplify all comparisons to `RecordNotFoundErrorString` to instead do direct comparison against `ErrNoRows` instead, but I didn't want to add all of those changes to this PR/feature.

## Setup

- `make server_test`
- Verify that all tests complete without any DB error string issues.
- Note that you will still see errors in the packages below for now.  I believe these are all issues caused by the direct `lib/pq` `COPY FROM` functionality that we will be refactoring to use the `jackc/pgx` variant in the next story.
    - github.com/transcom/mymove/pkg/migrate
    - github.com/transcom/mymove/pkg/parser/pricing
    - github.com/transcom/mymove/pkg/services/dbtools

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3963) for this change


pearl-truss:
@reggieriser it's hanging for me after the `github.com/transcom/mymove/pkg/paperwork` test. I will try again on Monday

pearl-truss (approved):
Disregard my last comment! I'm seeing the expected test failures now and no lingering error string comments

reggieriser:
> @reggieriser it's hanging for me after the `github.com/transcom/mymove/pkg/paperwork` test. I will try again on Monday

It's the `github.com/transcom/mymove/pkg/parser/pricing` package that comes after `paperwork` that's the problem.  I thought it was hung too, but it will eventually finish -- it's getting stuck in a goroutine that will eventually timeout after 10 minutes (I think) and let the tests proceed.  I believe this is related to the `COPY FROM` stuff that needs to be changed in the next story since the `pricing` package uses that as well.

reggieriser:
I decided to add some helper methods and abstract out the DB error checks to better isolate the specific error type dependency (so we can switch drivers easier during this migration).  Also makes things a little more DRY.  @pearl-truss , could you please re-review?

pearl-truss:
> I decided to add some helper methods and abstract out the DB error checks to better isolate the specific error type dependency (so we can switch drivers easier during this migration). Also makes things a little more DRY. @pearl-truss , could you please re-review?

@reggieriser I just re-ran the server test and everything still looks good

pearl-truss (approved):

PR #5035, opened 2020-10-20
Specify node version in pre-commit config

monfresh:
**Why**: By default, pre-commit expects a language to be available
globally. We use nodenv to manage node versions, and we shouldn't
expect folks to set their global version to always match the one we
use on the project.

To tell pre-commit to use a specific version of a language, you can
use the `language_version` key. This works for `node`, but not for
`golang` unfortunately. When I tried it, I got this message:

```
An unexpected error has occurred: AssertionError: For now, pre-commit
requires system-installed golang
```

I verified that when I didn't have `node` installed globally, the
`pre-commit install-hooks` command hung when trying to install the
node-based hooks. After I added the `language_version`, everything
installed as expected without hanging.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,7 @@
+default_language_version:
+  # this should match the version in .node-version at the root of this project
+  node: 12.16.3
+
 repos:
   - repo: local
     hooks:
@@ -39,7 +43,6 @@ repos:
     hooks:
       - id: markdownlint
         entry: markdownlint --ignore .github/*.md
-        language_version: 12.16.3
 
   - repo: git://github.com/detailyang/pre-commit-shell
     rev: 1.0.5
@@ -52,7 +55,6 @@ repos:
         name: prettier
         entry: node_modules/.bin/prettier --write
         language: node
-        language_version: 12.16.3
         files: \.(js|jsx)$
 
   - repo: local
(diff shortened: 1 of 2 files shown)


lynzt (approved):
this looks like a good thing to add based on the Moncef's description and the docs: https://pre-commit.com/#overriding-language-version

🚢

monfresh:
Thanks for the link to the docs! This reminds me we can set the default version once at the top instead of in multiple places.

.pre-commit-config.yaml
```diff
@@ -51,6 +52,7 @@ repos:
         name: prettier
         entry: node_modules/.bin/prettier --write
         language: node
+        language_version: 12.16.3
```
gidjin, line 55: Instead of setting this for each node based tool can you just use `default_language_version` to set it for all of them at once?

https://pre-commit.com/#overriding-language-version

gidjin (changes requested):
I'm unclear how this works for node, though I can confirm that it does. I would like to see if using the default setup would work so marking this as changes req for that reason. Please also add documentation to a wiki page explaining why we need this for node based tools and note that it needs to be kept in sync with `.node-version` as we upgrade. I'd say link to the new wiki page from the page on how to upgrade node but I don't think there is a doc for that though we probably should have one like we do for go since it's more complex now than it used to be.

Too bad it won't work for go, but marginal gains are still gains

.pre-commit-config.yaml
monfresh: Done.

gidjin (approved):
Changes look good, though I think a wiki page for further context is worth while.

monfresh:
Yeah, for sure. I'm working on the wiki now.

PR #5052, opened 2020-10-21
Kim mb 4407 only one nts

kimallen:
## Description

This limits the front-end user to being able to create more than one NTS or NTSr by disabling the selection radio when either has already been created.

If both have been created, alternative text is shown instead of the selection cards.

If the move has been submitted, whether they have created either of the long-term storage types, they will see the alternative text.

## Reviewer Notes

This includes 3 stories.

## Setup

`make db_dev_e2e_populate`

Use `sm_no_move_type@example.com`

Choose an NTS and create it.  
`Add another shipment`
See that the NTS option is disabled.
Choose an NTSr and create it.
`Add another shipment`
See that both cards are gone and are replaced with alternate text.

Log in as a different user: `hhg@only.unsubmitted`
Choose an NTSr and create one.
From review page, `Add a shipment` and see that it is now disabled.
Use browser back.  From review page, choose `Next` and submit the full move.
From Home screen, `Add a shipment`.
See that the cards are not there and the alternative text is there.

## References

* [Only one NTS](https://dp3.atlassian.net/browse/MB-4407) for this change
* [Only one NTSr](https://dp3.atlassian.net/browse/MB-4513) for this change
* [Neither available](https://dp3.atlassian.net/browse/MB-4575) for this change

## Screenshots

has both NTS and NTSr
image

has NTS
image


src/pages/MyMove/SelectMoveType.jsx
```diff
@@ -154,29 +157,34 @@ export class SelectMoveType extends Component {
...
-                label="Get things out of long-term storage"
-                value={SHIPMENT_OPTIONS.NTSR}
-                id={SHIPMENT_OPTIONS.NTSR}
-                cardText={isNtsSelectable ? ntsrCardText : ntsrDisabledText}
-                checked={moveType === SHIPMENT_OPTIONS.NTSR && isNtsrSelectable}
-                disabled={!isNtsrSelectable}
-                onHelpClick={this.toggleStorageModal}
-              />
+              <h3 data-testid="long-term-storage-heading">Long-term storage</h3>
+              {(!isNtsSelectable && !isNtsrSelectable) || move.status !== MOVE_STATUSES.DRAFT ? (
```
sojeri, line 161: shouldn't need to separately check for draft. this should be part of is-selectable calculation

sojeri (commented):
looking good aside from one small comment. will pull locally to manually test FE before approving.

lynzt (changes requested):
I think we need to remove this txt too when we hide the NTS(r) buttons...

Screen_Shot_2020-10-22_at_2_02_32_PM

https://app.zeplin.io/project/5f6e33d2169b8832e39d06b7/screen/5f6e34eeb1c71c8bb1a02549

src/pages/MyMove/SelectMoveType.jsx
```diff
@@ -71,11 +72,13 @@ export class SelectMoveType extends Component {
     const ntsrCardText =
       'Movers pick up things you put into NTS during an earlier move and ship them to your new destination. This is an NTS-R (non-temporary storage release) shipment.';
     const ntsDisabledText =
-      "You've already requested a long-term storage shipment for this move. Talk to your movers to change or add to your request.";
+      'You‘ve already requested a long-term storage shipment for this move. Talk to your movers to change or add to your request.';
```
suzubara, line 75: nit: these apostrophes look backwards, I think they should be: `’`

kimallen: Hmm, @suzubara I can't actually see the difference.  I"m trying to use the shortcuts you suggested.  Should it be option+shift+] OR option+]?

kimallen: I think I picked the right one now. . .

lynzt (approved):
let's address the wording for the different shipments in another pr and merge this as is

🚢

src/pages/MyMove/SelectMoveType.jsx
sojeri: can we just `&rsquo;` ?

src/pages/MyMove/SelectMoveType.jsx
```diff
@@ -51,18 +51,19 @@ export class SelectMoveType extends Component {
...
-      push,
-      isPpmSelectable,
-      isHhgSelectable,
-      isNtsSelectable,
-      isNtsrSelectable,
-      shipmentNumber,
-    } = this.props;
+    const { pageKey, pageList, match, push, move, mtoShipments } = this.props;
     const { moveType, showStorageInfoModal } = this.state;
+    const hasNTS = mtoShipments.some((shipment) => shipment.shipmentType === SHIPMENT_OPTIONS.NTS);
```
sojeri, line 56: instead of moving this all to render and dropping the tests, can we move it to a util function that gets called during render and which we can separately test?

mariaclaratgolfo (approved):

src/pages/MyMove/SelectMoveType.jsx
kimallen: I did add and adjust tests in the client test to make sure we are getting the right values.  I do see where testing that logic directly will be better if something goes wrong.  I do think adding those utils and tests will bump this story into the next sprint for me.  I'm going to make a story combining your request along with adding the additional layer `SelectMoveTypeForm`.

kimallen:
> I think we need to remove this txt too when we hide the NTS(r) buttons...
> 
> Screen_Shot_2020-10-22_at_2_02_32_PM
> 
> https://app.zeplin.io/project/5f6e33d2169b8832e39d06b7/screen/5f6e34eeb1c71c8bb1a02549

Adding this task as a separate story, along with pulling out the logic values into a util fn with tests:
https://dp3.atlassian.net/browse/MB-4935

PR #5060, opened 2020-10-22
Edit Shipment F/E work

suzubara:
## Description

F/E cleanup around Edit Shipment form:
- hoisted API calls & Redux connect up to the `CreateOrEditShipment` page
-  used Formik props (`values, setValues`) to handle form input logic around the delivery address inputs & current residence
- pass `shipmentNumber` into `MtoShipmentFormFields`
- don't store `initialValues` in state

### DIFF
--- a/src/components/Customer/MtoShipmentForm/MtoShipmentForm.jsx
+++ b/src/components/Customer/MtoShipmentForm/MtoShipmentForm.jsx
@@ -15,29 +15,12 @@ import {
 } from 'shared/Entities/modules/mtoShipments';
 import { selectActiveOrLatestOrdersFromEntities } from 'shared/Entities/modules/orders';
 import { selectServiceMemberFromLoggedInUser } from 'shared/Entities/modules/serviceMembers';
-import { showLoggedInUser as showLoggedInUserAction } from 'shared/Entities/modules/user';
 import { WizardPage } from 'shared/WizardPage';
 import { AddressShape, SimpleAddressShape } from 'types/address';
 import { HhgShipmentShape, MatchShape, HistoryShape, PageKeyShape, PageListShape } from 'types/customerShapes';
 import { formatMtoShipmentForAPI, formatMtoShipmentForDisplay } from 'utils/formatMtoShipment';
 
 class MtoShipmentForm extends Component {
-  componentDidMount() {
-    const { showLoggedInUser } = this.props;
-    showLoggedInUser();
-    // TODO - move this to the parent component instead
-
-    // TODO: confirm this block should exist
-    // If refreshing edit page, need to handle mtoShipment populating from a promise
-    /*
-    if (!isCreatePage && mtoShipment.id) {
-      this.setState({
-        initialValues,
-        hasDeliveryAddress: initialValues.hasDeliveryAddress,
-      });
-    } */
-  }
-
   submitMTOShipment = ({ shipmentType, pickup, delivery, customerRemarks }) => {
     const {
       createMTOShipment,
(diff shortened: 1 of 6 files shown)


src/components/Customer/MtoShipmentForm/MtoShipmentForm.jsx
```diff
@@ -212,29 +88,82 @@ class MtoShipmentForm extends Component {
...
+                // use current residence
+                setValues({
+                  ...values,
+                  pickup: {
+                    ...values.pickup,
+                    address: currentResidence,
+                  },
+                });
+              } else if (match.params.moveId === mtoShipment?.moveTaskOrderId) {
+                // TODO - what is the purpose of this check?
```
suzubara, line 104: @sojeri forgot to ask you this... I'm not convinced we need this but left it as is because I wasn't sure

src/components/Customer/MtoShipmentForm/MtoShipmentForm.jsx
```diff
@@ -212,29 +88,82 @@ class MtoShipmentForm extends Component {
...
+                      postal_code: '',
+                    },
+                  },
+                });
+              }
+            };
+
+            if (isCreatePage) {
+              // return MTO Shipment form in the wizard
+              return (
```
sojeri, line 133: nested returns?! 🤯

src/utils/formatMtoShipment.js
```diff
@@ -41,11 +41,13 @@ function formatAddressForAPI(address) {
 
   if (formattedAddress.state) {
     formattedAddress.state = formattedAddress.state?.toUpperCase();
+    delete formattedAddress.id;
```
suzubara, line 44: @sojeri I added this because I was getting 422 validation errors from the create/update shipment calls saying `address.id` could not be part of the payload. I'm not sure where that was introduced but this seemed like a safe way to resolve it.

src/pages/MyMove/CreateOrEditMtoShipment.jsx
```diff
@@ -18,32 +24,43 @@ class CreateOrEditMtoShipment extends Component {
...
-function mapStateToProps(state, ownProps) {
-  const props = {
-    mtoShipment: selectMTOShipmentForMTO(state, ownProps.match?.params.moveId),
-  };
-  return props;
-}
-
-function mapDispatchToProps(dispatch) {
-  return bindActionCreators({ fetchCustomerData: fetchCustomerDataAction }, dispatch);
+    return <LoadingPlaceholder />;
```
sojeri, line 62: oh nice! I'll have to remember too revert my change to the get display options fn (I removed throwing an error, but with this change it seems still useful to have)

sojeri (approved):
looks good to me. want to merge? :)

src/utils/formatMtoShipment.js
sojeri: we may need to dig into this to see why the BE is doing that. if we're updating an MTO shipment, it should be valid to update an already existing address rather than creating a new one. this is something we can cleanup after merging this main PR though.

src/components/Customer/MtoShipmentForm/MtoShipmentForm.jsx
sojeri: yeah, it was preexisting code. when I read it, I was thinking this was protection against the server accidentally sending an mtoShipment that belonged to another move/user.

suzubara: cool, probably something we can remove down the road then

PR #5072, opened 2020-10-23
MB-4842 Allow filtering by branch

monfresh:
## Description

As part of the new TOO queue page, we are adding the ability to filter by various parameters. This PR only adds the backend functionality to support these queries. Adding the actual dropdowns and text fields to the queues page will be performed separately.

## Reviewer Notes

To test:

1. Run the setup steps below if you haven't already seeded your local DB
2. Go to http://officelocal:3000/devlocal-auth/login
3. Click on the Login button next to `too_role@office.mil (PPM office)`
4. Visit `http://officelocal:3000/ghc/v1/queues/moves` in Firefox or with a browser extension that makes it easier to read JSON data
5. Note the amount of results you get. With our current setup, this should be 2.
6. Append `?branch=ARMY` to the URL

Expected: No results appear.

7.  Append `?branch=AIR_FORCE` to the URL

Expected: only results where the `departmentIndicator` is "AIR_FORCE" should appear

**Important Note**
Based on the current moves queue displaying the "Branch" value from the order's department indicator, that is how I implemented the filtered query. It has since come to my attention that we most likely implemented this column incorrectly. The Branch is supposed to represent the service member's `Affiliation`. This will require front end changes, which we will most likely do in a separate PR.

## Setup

```sh
make db_dev_e2e_populate
make server_run
make office_client (in a separate terminal tab)
```

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4842)

### DIFF
--- a/pkg/handlers/ghcapi/queues.go
+++ b/pkg/handlers/ghcapi/queues.go
@@ -86,7 +86,7 @@ func (h GetPaymentRequestsQueueHandler) Handle(params queues.GetPaymentRequestsQ
 func branchFilter(params queues.GetMovesQueueParams) FilterOption {
 	return func(query *pop.Query) {
 		if params.Branch != nil {
-			query = query.Where("orders.department_indicator = ?", *params.Branch)
+			query = query.InnerJoin("service_members", "service_members.id = orders.service_member_id").Where("service_members.affiliation = ?", *params.Branch)
 		}
 	}
 }
--- a/pkg/handlers/ghcapi/queues.go
+++ b/pkg/handlers/ghcapi/queues.go
@@ -1,9 +1,6 @@
 package ghcapi
 
 import (
-	"encoding/json"
-	"fmt"
-
 	"github.com/go-openapi/runtime/middleware"
 	"github.com/gobuffalo/pop"
 	"go.uber.org/zap"
@@ -23,11 +20,6 @@ type GetMovesQueueHandler struct {
 	services.MoveOrderFetcher
 }
 
-// Filter defines all the possible filter parameters
-type Filter struct {
-	Branch string `json:"branch"`
-}
-
 // FilterOption defines the type for the functional arguments passed to ListMoveOrders
 type FilterOption func(*pop.Query)
 
@@ -39,9 +31,7 @@ func (h GetMovesQueueHandler) Handle(params queues.GetMovesQueueParams) middlewa
 		return queues.NewGetMovesQueueForbidden()
 	}
 
-	queryFilters := h.generateQueryFilters(params.Filter, logger)
-
-	branchQuery := branchFilter(queryFilters)
+	branchQuery := branchFilter(params)
 
 	orders, err := h.MoveOrderFetcher.ListMoveOrders(session.OfficeUserID, branchQuery)
 	if err != nil {
@@ -93,28 +83,10 @@ func (h GetPaymentRequestsQueueHandler) Handle(params queues.GetPaymentRequestsQ
 	return queues.NewGetPaymentRequestsQueueOK().WithPayload(result)
 }
 
-// generateQueryFilters is helper to convert filter params from a json string
-// of the form `{"status": "Submitted" "branch": "Army"}` to an array of services.QueryFilter
-func (h GetMovesQueueHandler) generateQueryFilters(filters *string, logger handlers.Logger) Filter {
-	f := Filter{}
-	if filters == nil {
-		return f
-	}
-	b := []byte(*filters)
-	err := json.Unmarshal(b, &f)
-	if err != nil {
-		fs := fmt.Sprintf("%v", filters)
-		logger.Warn("unable to decode param", zap.Error(err),
-			zap.String("filters", fs))
-	}
-
-	return f
-}
-
-func branchFilter(filter Filter) FilterOption {
+func branchFilter(params queues.GetMovesQueueParams) FilterOption {
 	return func(query *pop.Query) {
-		if filter.Branch != "" {
-			query = query.Where("orders.department_indicator = ?", filter.Branch)
+		if params.Branch != nil {
+			query = query.Where("orders.department_indicator = ?", *params.Branch)
 		}
 	}
 }
--- a/pkg/handlers/ghcapi/queues.go
+++ b/pkg/handlers/ghcapi/queues.go
@@ -1,7 +1,11 @@
 package ghcapi
 
 import (
+	"encoding/json"
+	"fmt"
+
 	"github.com/go-openapi/runtime/middleware"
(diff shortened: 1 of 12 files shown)


pkg/handlers/ghcapi/queues.go
```diff
@@ -75,4 +81,12 @@ func (h GetPaymentRequestsQueueHandler) Handle(params queues.GetPaymentRequestsQ
 	}
 
 	return queues.NewGetPaymentRequestsQueueOK().WithPayload(result)
-}
\ No newline at end of file
+}
+
+func branchFilter(params queues.GetMovesQueueParams) FilterOption {
```
duncan-truss, line 86: I think this should be moved into the services package instead of living in the handler.  The handler shouldn't need to know how the db/ORM works we are leaking that here and it will complicate the tests.

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -90,6 +90,59 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandler() {
 
 }
 
+func (suite *HandlerSuite) TestGetMoveQueuesFilter() {
+	officeUser := testdatagen.MakeDefaultOfficeUser(suite.DB())
```
duncan-truss, line 94: I think we have MakeTOOOfficeUser that adds the role

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -90,6 +90,59 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandler() {
...
+	}
+
+	response := handler.Handle(params)
+	suite.IsNotErrResponse(response)
+
+	suite.Assertions.IsType(&queues.GetMovesQueueOK{}, response)
+	payload := response.(*queues.GetMovesQueueOK).Payload
+
+	suite.Equal(1, len(payload.QueueMoves))
+}
```
duncan-truss, line 144: Can we test that it's the right branch in the payload?

pkg/testdatagen/make_order.go
```diff
@@ -46,7 +46,11 @@ func MakeOrder(db *pop.Connection, assertions Assertions) models.Order {
 
 	ordersNumber := "ORDER3"
 	TAC := "F8E1"
-	departmentIndicator := "AIR_FORCE"
+	defaultDepartmentIndicator := "AIR_FORCE"
```
duncan-truss, line 49: Thanks for filling out the testdata!

swagger/ghc.yaml
```diff
@@ -1100,6 +1100,10 @@ paths:
       operationId: getMovesQueue
       tags:
         - queues
+      parameters:
+        - in: query
+          name: branch
+          type: string
```
duncan-truss, line 1106: Should this be an array type to support multiple values?  I think we have an enum too but it may be the one that combines navy and marines.

pkg/handlers/ghcapi/queues_test.go
monfresh: Yeah, but it expects assertions. We could either modify it so that it doesn't require assertions, or create a `MakeDefaultTOOOfficeUser`.

swagger/ghc.yaml
monfresh: The JIRA issue says this should be a single selector with values: `All, Army, Navy, Marine Corps, Air Force, Coast Guard`. Note that these are supposed to map to the Service Member's affiliation, not the Order's Department Indicator. I opened a new JIRA issue to fix that on the front end. I also just updated this PR so that the backend looks up the affiliation.

pkg/handlers/ghcapi/queues_test.go
monfresh: Yes we can!

pkg/handlers/ghcapi/queues.go
monfresh: Sure, we can do that. Can you explain how it will complicate tests? Either way, the handler will be passing in several of these types of filters to `ListMoveOrders`. Does the test I wrote seem more complicated than it should be?

Ryan-Koch (approved):
It looks like this does the thing. I saw the branch thing is in another issue, so this work should do the trick for the scope of this story.

pkg/handlers/ghcapi/queues.go
duncan-truss: I meant more in general that if you wanted to start mocking things out or only testing the service it would make the inputs a little messier to set up.

swagger/ghc.yaml
duncan-truss: Oh so the only multiselect in the queue is the status?  I kind of assumed they were all the same status, branch, and gbloc.

Ryan-Koch: That's correct @duncan-truss , status is the only one that's using the multi-select functionality 👍.

pkg/handlers/ghcapi/queues.go
monfresh: Ah yeah, that makes sense.

donaldthai (approved):
LGTM besides Duncan's comment. 👍

pkg/handlers/ghcapi/queues_test.go
monfresh: Done!

pkg/handlers/ghcapi/queues.go
monfresh: This is more complicated than it seemed. We have 2 handlers that call the same `ListMoveOrders` function, and they use different types for the params. I believe the intent is to remove `pkg/handlers/ghcapi/move_orders.go` once we are satisfied with the new queues, so we can address that then?

Also, for some reason, I was not able to get the query working when I put it in the service. Perhaps I'm missing some Golang concepts. I did something like this:
```golang
query := f.db.Q().Eager(.... etc)
branchQuery := branchFilter(params)
query = filteredOrders(query, branchQuery)

func filteredOrders(baseQuery *pop.Query, options ...func(query *pop.Query)) *pop.Query {
	for _, option := range options {
		if option != nil {
			option(baseQuery)
		}
	}
	return baseQuery
}

func branchFilter(params queues.GetMovesQueueParams) FilterOption {
		if params.Branch != nil {
			query = query.InnerJoin("service_members", "service_members.id = orders.service_member_id").Where("service_members.affiliation = ?", *params.Branch)
		}
	}
}
```
I also tried it without reassigning `query` and calling `filteredOrders` directly without having it return the query.

monfresh: Ha! It was my test that was not set up properly. It's working now.

PR #5076, opened 2020-10-23
mb-4886 | migrate selected move type nts raw value

sojeri:
## Description

in PR #4950 I updated the raw value of move.selectedMoveType for NTS shipments, but I did not add a migration for this change to update any existing moves with NTS shipments. this PR fixes that.

add migration to reflect new NTS raw value for move.selectedMoveType
- including column comment update
- also fixed a broken link

## Reviewer Notes

Am I doing this right? :)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4886) for this change


migrations/app/schema/20201023202845_update_selectedMoveType_NTS.up.sql
```diff
@@ -0,0 +1,8 @@
+UPDATE
+    moves
+SET
+    selected_move_type = 'HHG_INTO_NTS_DOMESTIC'
+WHERE
+    selected_move_type = 'NTS';
+
+COMMENT ON COLUMN moves.selected_move_type IS 'The type of Move the customer is choosing. Allowed values are HHG, PPM, UB, POV, HHG_INTO_NTS_DOMESTIC, HHG_OUTOF_NTS_DOMESTIC HHG_PPM (but only HHG and PPM appear to be used currently).';
```
kimallen, line 8: I think the last part of the comment can be removed "(but only HHG and PPM appear to be used currently)".  Or you can add in HHG_INTO_NTS_DOMESTIC and HHG_OUTOF_NTS_DOMESTIC to that statement.

migrations/app/schema/20201023202845_update_selectedMoveType_NTS.up.sql
```diff
@@ -0,0 +1,8 @@
+UPDATE
+    moves
+SET
+    selected_move_type = 'HHG_INTO_NTS_DOMESTIC'
```
kimallen, line 4: Do we also need to set 'HHG_OUTOF_NTS_DOMESTIC', or was that already done previously?

lynzt (changes requested):
LGTM, but let's first make at least 1 NTS on experimental... and then run this and make sure things don't break...

migrations/app/schema/20201023202845_update_selectedMoveType_NTS.up.sql
lynzt: also, good question ☝️

sojeri: we don't need to set it because NTSR is new. NTS was an existing type which got updated, so we need this migration to convert old values to new.

sojeri:
@lynzt & @kimallen, here's the results of my adding a couple moves on exp:

- none of them show up in TIO/TOO regardless of whether there's an NTS shipment included or not
- adding a PPM shipment does show in the PPM office app
- saw no other unexpected issues, though the experience of submitting a PPM after submitting the main move is not great

lynzt (approved):
with the test of the migration in experimental, LGTM 🚢

PR #5077, opened 2020-10-23
MB-4845 TOO queue can filter status

donaldthai:
## Description

As a TOO, I want to be able to filter certain columns in my queue, so that I can narrow my focus on relevant moves.
> Given that I am a TOO logged into MM
>  When I look at my work queue and want to view a subset of data
>  Then I can use filters on the following columns:
>       Status

Notes

There are a few different types of filters for the various columns:

MoveStatus - Multi drop down selector, with options:

- New move

- Approvals requested

- Move approved

Data should be updated with a query when the user click’s away from the element they are searching with.

## Reviewer Notes

- The status is determined after the results come back from pop. So the filter is applied to the result set after we get back the data. We may want to refactor the logic to determine the status in the pop query instead later.
- @monfresh, @noelledusahel and I decided to individually include filter params since we don't think we'll be using the `query_builder` so therefore, single params to capture filters should be well documented.

## Setup

```sh
make server_run
make office_client_run
```

Test:
1. Make sure you log in to the office app as a TOO role first.
2. To filter, the status param must be included in the query. Here's a query that includes all the statuses:  `http://officelocal:3000/ghc/v1/queues/moves?status=New%20move%2CApprovals%20requested%2CMove%20approved`

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4845) for this change


pkg/handlers/ghcapi/queues_test.go
```diff
@@ -241,6 +243,157 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandlerStatuses() {
...
+			Name:  "Fort Punxsutawney",
+			Gbloc: "AGFM",
+		},
+		MTOShipment: models.MTOShipment{
+			Status: models.MTOShipmentStatusSubmitted,
+		},
+		Move: models.Move{
+			SelectedMoveType: &hhgMoveType,
+		},
+	})
```
monfresh, line 349: Since the GBLOC filtering is already tested elsewhere, what do you think about only creating the minimum amount of data to test just the status filtering?

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -241,6 +243,157 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandlerStatuses() {
...
+
+	request := httptest.NewRequest("GET", "/move-task-orders/{moveTaskOrderID}", nil)
```
monfresh, line 351: Shouldn't the endpoint be `/queues/moves`? It doesn't look like it affects the test, but for documentation purposes, I would vote for specifying the correct endpoint.

monfresh:
Now that the TOO can filter by statuses, we should remove this line from `move_order_fetcher.go`:
```golang
// TODO: Let's include the status in filters that are passed into this service once we build that feature for the TXO queue (instead of it being hardcoded like it is below right now).
Where("moves.status NOT IN ('DRAFT', 'CANCELLED')").
```

pkg/handlers/ghcapi/queues_test.go
donaldthai: Works for me!

donaldthai:
> Now that the TOO can filter by statuses, we should remove this line from `move_order_fetcher.go`:
> 
> ```go
> // TODO: Let's include the status in filters that are passed into this service once we build that feature for the TXO queue (instead of it being hardcoded like it is below right now).
> Where("moves.status NOT IN ('DRAFT', 'CANCELLED')").
> ```

Sure, I can move that logic into the filter function.

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -241,6 +243,157 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandlerStatuses() {
...
+		Order: order3,
+	})
+
+	testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+		Move: move3,
+		MTOShipment: models.MTOShipment{
+			Status: models.MTOShipmentStatusApproved,
+		},
+	})
+
```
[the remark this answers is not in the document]
donaldthai: Sweet, I just put in that change.

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -241,6 +243,114 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandlerStatuses() {
...
+	testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+		Move: models.Move{
+			SelectedMoveType: &hhgMoveType,
+			Status:           models.MoveStatusDRAFT,
+		},
+		MTOShipment: models.MTOShipment{
+			Status: models.MTOShipmentStatusSubmitted,
+		},
+	})
+	// Move approved
```
monfresh, line 297: I think you meant Move canceled or maybe no comment since the one above refers to both draft and canceled?

monfresh (approved):
LGTM once the comment thing is fixed.

pkg/handlers/ghcapi/queues_test.go
donaldthai: oops, yeap

donaldthai: removing that

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -42,8 +42,6 @@ func (f moveOrderFetcher) ListMoveOrders(officeUserID uuid.UUID, options ...func
 		InnerJoin("duty_stations", "orders.origin_duty_station_id = duty_stations.id").
 		InnerJoin("transportation_offices", "duty_stations.transportation_office_id = transportation_offices.id").
 		Where("transportation_offices.gbloc = ?", gbloc).
-		// TODO: Let's include the status in filters that are passed into this service once we build that feature for the TXO queue (instead of it being hardcoded like it is below right now).
-		Where("moves.status NOT IN ('DRAFT', 'CANCELLED')").
```
duncan-truss, line 46: I think our status logic has to be SQL for this to be reasonably performant in the future.

Ryan-Koch (approved):

pkg/services/move_order/move_order_fetcher.go
monfresh: Yeah, that's where we were originally heading with the functional arguments. @donaldthai Did it end up not working? We should open a follow-up JIRA story to get this working with SQL.

donaldthai: Yea, didn't wanna take up too much time converting the logic from the [queueMoveStatus](https://github.com/transcom/mymove/blob/master/pkg/handlers/ghcapi/internal/payloads/model_to_payload.go#L500) function that determines the status. I can create that chore to convert this to the pop query.

donaldthai: @monfresh we could just include it in the chore you created: https://dp3.atlassian.net/browse/MB-4941

PR #5078, opened 2020-10-25
MB-4461 Customer Info Filters for Queues Table -  Backend

noelledusahel:
## Description

## Reviewer Notes

## Setup

```sh
echo "Code goes here"
```

 have been satisfied.

## References

## Screenshots


pkg/handlers/ghcapi/queues.go
```diff
@@ -75,4 +96,46 @@ func (h GetPaymentRequestsQueueHandler) Handle(params queues.GetPaymentRequestsQ
...
+			nameSearch := fmt.Sprintf("%s%%", *params.LastName)
+			query = query.Where("service_members.last_name ILIKE ?", nameSearch)
+		}
+	}
+}
+
+func firstNameFilter(params queues.GetMovesQueueParams) FilterOption {
+	return func(query *pop.Query) {
+		if params.FirstName != nil {
+			nameSearch := fmt.Sprintf("%s%%", *params.FirstName)
```
noelledusahel, line 113: 'Starts with' pattern

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -30,22 +30,32 @@ func (f moveOrderFetcher) ListMoveOrders(officeUserID uuid.UUID) ([]models.Order
...
+	query := f.db.Q().Eager(
 		"ServiceMember",
 		"NewDutyStation.Address",
+		"NewDutyStation.Name",
 		"OriginDutyStation",
 		"Entitlement",
 		"Moves.MTOShipments",
 		"Moves.MTOServiceItems",
 	).InnerJoin("moves", "orders.id = moves.orders_id").
+		InnerJoin("service_members", "orders.service_member_id = service_members.id").
```
noelledusahel, line 42: i would want to explore conditionally joining tables based on filters present, but none of that is implemented here.

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -188,6 +188,203 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandlerStatuses() {
...
+	request = suite.AuthenticateOfficeRequest(request, officeUser)
+
+	context := handlers.NewHandlerContext(suite.DB(), suite.TestLogger())
+	handler := GetMovesQueueHandler{
+		context,
+		officeuser.NewOfficeUserFetcher(query.NewQueryBuilder(context.DB())),
+		moveorder.NewMoveOrderFetcher(suite.DB()),
+	}
+
+	suite.Run("loads unfiltered results", func() {
```
Ryan-Koch, line 303: Nice, thanks for creating cases for the different filter types.

pkg/handlers/ghcapi/queues.go
Ryan-Koch: We may wind up wanting to fiddle with it once we learn more about how folks are going to use it, but I think for now this gets us going with something that does the thing.

Ryan-Koch (approved):

pkg/services/move_order/move_order_fetcher.go
monfresh: Have you tried specifying the inner join within the individual filters? For example:
```golang
func lastNameFilter(params queues.GetMovesQueueParams) FilterOption {
 		if params.LastName != nil {
 			nameSearch := fmt.Sprintf("%s%%", *params.LastName)
 			query = query.InnerJoin("service_members", "orders.service_member_id = service_members.id").Where("service_members.last_name ILIKE ?", nameSearch)
 		}
 	}
 }

File: pkg/services/move_order/move_order_fetcher.go [hunk 4]
```diff
@@ -30,22 +30,32 @@ func (f moveOrderFetcher) ListMoveOrders(officeUserID uuid.UUID) ([]models.Order
 
 	gbloc := transportationOffice.Gbloc
 
-	err = f.db.Q().Eager(
+	query := f.db.Q().Eager(
 		"ServiceMember",
 		"NewDutyStation.Address",
+		"NewDutyStation.Name",
```
monfresh, line 36: I don't think this does anything. Eager loading is for associations. Name is a column on the NewDutyStation table, it's not a separate table.

noelledusahel: oops, meant to take that out

pkg/services/move_order/move_order_fetcher.go
noelledusahel: great idea 👍

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -188,6 +188,202 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandlerStatuses() {
...
+		OfficeUser: models.OfficeUser{
+			TransportationOffice:   office,
+			TransportationOfficeID: office.ID,
+		},
+	})
+
+	officeUser.User.Roles = append(officeUser.User.Roles, roles.Role{
+		RoleType: roles.RoleTypeTOO,
+	})
+
```
monfresh, line 224: Could we simplify this to:
```golang
dutyStation1 := testdatagen.MakeDutyStation(suite.DB(), testdatagen.Assertions{
     DutyStation: models.DutyStation{
         Name: "This Other Station",
     },
 })

dutyStation2 := testdatagen.MakeDefaultDutyStation(suite.DB())

officeUser := testdatagen.MakeTOOOfficeUser(suite.DB(), testdatagen.Assertions{})
```
The reasoning is that for this test, the GBLOC of the transportation office is not relevant. All data created in tests should be relevant. If it's not, it can cause confusion as to why it's specified in the test. The default data generators automatically use the same GBLOC for the office user and orders, so it's not necessary to specify a GBLOC in tests that are testing other filters.

noelledusahel: yep makes sense.

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -188,6 +188,202 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandlerStatuses() {
...
+		},
+		Order: order2,
+	})
+
+	testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+		Move: move2,
+		MTOShipment: models.MTOShipment{
+			Status: models.MTOShipmentStatusSubmitted,
+		},
+	})
```
monfresh, line 272: What do you think about trimming off a few lines between 244-291 by doing this:
```golang
move := models.Move{
    SelectedMoveType: &hhgMoveType,
    Status:           models.MoveStatusSUBMITTED,
}
shipment := models.MTOShipment{
  Status: models.MTOShipmentStatusSubmitted,
}

testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
  Move: move,
  Order: models.Order{
    NewDutyStation:      dutyStation1,
    NewDutyStationID:    dutyStation1.ID,
  },
  MTOShipment: shipment,
  ServiceMember: serviceMember1,
})

testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
  Move: move,
  Order: models.Order{
    NewDutyStation:      dutyStation2,
    NewDutyStationID:    dutyStation2.ID,
  },
  MTOShipment: shipment,
  ServiceMember: serviceMember2,
})
```

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -188,6 +188,183 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandlerStatuses() {
...
+			HTTPRequest: request,
+			FirstName:   serviceMember1.FirstName,
+		}
+
+		response := handler.Handle(params)
+		suite.IsNotErrResponse(response)
+
+		payload := response.(*queues.GetMovesQueueOK).Payload
+
+		suite.Len(payload.QueueMoves, 1)
```
monfresh, line 308: Can we also verify that the name in the payload is the expected one? Same for all the tests below if we can verify that the correct entry got returned.

pkg/handlers/ghcapi/queues.go
```diff
@@ -19,15 +22,33 @@ type GetMovesQueueHandler struct {
...
 		logger.Error("user is not authenticated with TOO office role")
 		return queues.NewGetMovesQueueForbidden()
 	}
 
-	orders, err := h.MoveOrderFetcher.ListMoveOrders(session.OfficeUserID)
+	moveIDQuery := moveIDFilter(params)
+	dodIDQuery := dodIDFilter(params)
+	firstNameQuery := firstNameFilter(params)
+	lastNameQuery := lastNameFilter(params)
+	dutyStationQuery := destinationDutyStationFilter(params)
```
duncan-truss, line 41: I said this on Moncef's PR too that this pop query logic feels better suited for the service layer than the handler and would simplify only passing params or something similar to the ListMoveOrders fetcher.  Handlers shouldn't be doing any database related work, it's better to separate those concerns.

swagger/ghc.yaml
```diff
@@ -1100,6 +1100,22 @@ paths:
...
+      parameters:
+        - in: query
+          name: moveID
+          type: string
+        - in: query
+          name: lastName
+          type: string
+        - in: query
+          name: firstName
+          type: string
```
duncan-truss, line 1112: Did we decide to separate the name into first and last name columns in the queue?  If it's still one column we would need the query to do `WHERE first_name ILIKE %firstName% OR last_name ILIKE %lastName%` because right now it is an AND if they are in separate WHERE clauses

pkg/handlers/ghcapi/queues.go
monfresh: I agree. To minimize back and forth and merge conflicts, I propose that we merge this PR and Donald's PR after the minor test cleanup comments are addressed, and then I can resolve all conflicts on my PR and move all of the queries to the service.

swagger/ghc.yaml
noelledusahel: So the first pass of the filter will only search on the last name with a 'starts with' pattern rather than a wildcard. after the convo in sprint planning about not filtering on the first name, I may just remove the first name filter backend.

pkg/handlers/ghcapi/queues.go
```diff
@@ -75,4 +94,37 @@ func (h GetPaymentRequestsQueueHandler) Handle(params queues.GetPaymentRequestsQ
...
+			nameSearch := fmt.Sprintf("%s%%", *params.LastName)
+			query = query.InnerJoin("service_members", "orders.service_member_id = service_members.id").Where("service_members.last_name ILIKE ?", nameSearch)
+		}
+	}
+}
+
+func dodIDFilter(params queues.GetMovesQueueParams) FilterOption {
+	return func(query *pop.Query) {
+		if params.DodID != nil {
+			query = query.InnerJoin("service_members", "orders.service_member_id = service_members.id").Where("service_members.edipi ILIKE ?", params.DodID)
```
duncan-truss, line 111: We should add a migration creating an index for the edipi column on the service_members table.  I'm surprised it's not a unique column to begin with.

pkg/handlers/ghcapi/queues.go
```diff
@@ -75,4 +94,37 @@ func (h GetPaymentRequestsQueueHandler) Handle(params queues.GetPaymentRequestsQ
...
 	return queues.NewGetPaymentRequestsQueueOK().WithPayload(result)
-}
\ No newline at end of file
+}
+
+func lastNameFilter(params queues.GetMovesQueueParams) FilterOption {
+	return func(query *pop.Query) {
+		if params.LastName != nil {
+			nameSearch := fmt.Sprintf("%s%%", *params.LastName)
+			query = query.InnerJoin("service_members", "orders.service_member_id = service_members.id").Where("service_members.last_name ILIKE ?", nameSearch)
```
duncan-truss, line 103: We can index the last_name column in the service_members table since there is no wildcard at the beginning of the match.

pkg/handlers/ghcapi/queues.go
Ryan-Koch: We should, but let's capture that in another ticket. I don't want the scope of this one to creep and push us way over into the new sprint.

pkg/handlers/ghcapi/queues.go
duncan-truss, line 103: @noelledusahel Since lastNameFilter and dodIDFilter join to the same table is that a problem when they are both used?

pkg/handlers/ghcapi/queues_test.go
duncan-truss: 💯Great work testing

pkg/handlers/ghcapi/queues.go
noelledusahel: i was concerned about this...dont know the answer but i'll add it back into the query in the moveorderfetcher,

pkg/handlers/ghcapi/queues_test.go
monfresh: Agreed. This is a great set. One thing I see missing is a test where all filters are used at the same time. That would be good to test that the combination of queries works fine.

pkg/handlers/ghcapi/queues.go
```diff
@@ -86,7 +101,40 @@ func (h GetPaymentRequestsQueueHandler) Handle(params queues.GetPaymentRequestsQ
...
+		if params.DodID != nil {
+			query = query.Where("service_members.edipi ILIKE ?", params.DodID)
+		}
+	}
+}
+
+func moveIDFilter(params queues.GetMovesQueueParams) FilterOption {
+	return func(query *pop.Query) {
+		if params.MoveID != nil {
+			query = query.Where("moves.locator ILIKE ?", *params.MoveID)
```
monfresh, line 129: I asked the team and the consensus is that this should be an exact match. Same with the DOD ID filter.

pkg/handlers/ghcapi/queues.go
```diff
@@ -91,7 +106,40 @@ func (h GetPaymentRequestsQueueHandler) Handle(params queues.GetPaymentRequestsQ
...
+			nameSearch := fmt.Sprintf("%s%%", *params.LastName)
+			query = query.Where("service_members.last_name ILIKE ?", nameSearch)
+		}
+	}
+}
+
+func dodIDFilter(params queues.GetMovesQueueParams) FilterOption {
+	return func(query *pop.Query) {
+		if params.DodID != nil {
+			query = query.Where("service_members.edipi = ?", params.DodID)
```
monfresh, line 126: I'm curious why this works without `*params.DodID`. It's a string pointer like all the other params.

monfresh (approved):
LGTM

pkg/handlers/ghcapi/queues.go
donaldthai: It's because the `where` clauses and arguments gets converted into a string in the end. There's this line in the pop source code:
```go
sql = fmt.Sprintf("%s WHERE %s", sql, wc.Join(" AND "))
```

PR #5092, opened 2020-10-27
[MB-4931] on Select move type page, conditionally show footer text only when SM does not already have a move

SirenaBorracha:
## Description

Currently the Select move type page shows helper text regarding the move counselor regardless of whether or not the service member already has a move. This PR makes that helper text conditionally show only when the service member does not already have a move or storage request.

(I also confirmed that the helper text no longer shows regardless if a new SM profile selects any of the move/storage options and then clicks through to add another shipment.)

## Setup

```sh
make server_run
make client_run
```
Local login (I used needs@orde.rs) and follow through adding orders until you progress to the select move type page. See that the helper text is displayed regarding move choices and questions for move counselors. Progress through, adding a move or a storage request, then click to add another shipment. On the select move type page, see that the service member is no longer shown the helper text.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4931) for this change

## Screenshots

**Helper text showing for a brand new service member profile without any shipments:**
Screen Shot 2020-10-26 at 5 12 38 PM

**Helper text no longer showing on the same profile once a move has been added:** 
Screen Shot 2020-10-26 at 5 14 58 PM

### DIFF
--- a/src/pages/MyMove/SelectMoveType.jsx
+++ b/src/pages/MyMove/SelectMoveType.jsx
@@ -131,7 +131,7 @@ export class SelectMoveType extends Component {
     const footerText = (
       <div>
         {!hasMove && (
-          <div className={`${styles.footer} grid-col-12`}>
+          <div data-testid="helper-footer" className={`${styles.footer} grid-col-12`}>
             It’s OK if you’re not sure about your choices. Your move counselor will go over all your options and can
             help make changes if necessary.
           </div>
--- a/src/pages/MyMove/SelectMoveType.jsx
+++ b/src/pages/MyMove/SelectMoveType.jsx
@@ -64,6 +64,7 @@ export class SelectMoveType extends Component {
     const ppmCount = hasPpm ? 1 : 0;
     const mtosCount = mtoShipments?.length || 0;
     const shipmentNumber = 1 + ppmCount + mtosCount;
+    const hasMove = shipmentNumber > 1;
     const ppmCardText =
       'You pack and move your things, or make other arrangements, The government pays you for the weight you move.  This is a a Personally Procured Move (PPM), sometimes called a DITY.';
     const hhgCardText =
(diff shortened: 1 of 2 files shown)


sojeri:
loving the no helper text for you :3

src/pages/MyMove/SelectMoveType.jsx
```diff
@@ -62,6 +62,7 @@ export class SelectMoveType extends Component {
     const ppmCount = hasPpm ? 1 : 0;
     const mtosCount = mtoShipments?.length || 0;
     const shipmentNumber = 1 + ppmCount + mtosCount;
+    const hasMove = shipmentNumber > 1;
```
gidjin, line 65: This confuses me some, I see that shipmentNumber defined on the line above as `1 + ppmCount + mtosCount`, but below on lines 162-164 it seems that `shipmentNumber > 1` means more than move, which means that `shipmentNumber == 1` would mean a single move vs no moves. Is the condition correct in both places? This could also be my lack of understanding of how this page is supposed to work too.

Lines 162-164
```javascript
          {shipmentNumber > 1
                  ? 'How do you want this group of things moved?'
                  : 'How do you want to move your belongings?'}	
```

kimallen, line 65: I would change this to `hasShipment`.  Even though the file is called "SelectMoveType", it's not actually accurate.  We are selecting a shipment type, not a move type.  At this point, the move has already been created.

kimallen: @gidjin - the shipment number on this page (applied in the top left "eyebrow") indicates which shipment they are about to select and create, which is why it's 1 when they don't have any shipments.  So the logic is correct. That said, your confusion indicates that it may be confusing for others reading the code. 

 Ari, to avoid that confusion, I'd suggest changing the definition of `hasMove` (or `hasShipment` as suggested below) to `(ppmCount + mtosCount) > 0`.

gidjin: Thanks for the explanation! I agree with your suggestion will make it more readable to those outside that section of the code.

kimallen (approved):
Looks good- works as expected. Nice work!

One process tip- I think rather than resolving the comments yourself, it's helpful for me as reviewer to resolve them after I verify that they are resolved.  This was easy because there were only small change requests and I could remember what they are.  For a larger PR, I may not remember what the requested changes were, so better to be able to see them easily.

PR #5106, opened 2020-10-27
[MB-4966] - Add SHIPMENT to signed cert type, submit NTS/NTS-R moves

amitch23:
## Description

Fixes the bug where the NTS/NTS-R shipments error when the user tries to submit the move (i.e. hitting next on the signed certification page). 

Adds a signed cert type of "SHIPMENT" to the API, which covers all current types (PPM, HHG, NTS, NTS-R), and sets a const in the f/e to pass the SHIPMENT type to the endpoint. 

## Reviewer Notes

- Assumes that we _will_ need to keep the existing types for the moves already submitted in prod et al, since the signed cert text varies per type (i.e. PPM has had it's own text, as has HHG, etc.). Because of this, we will need to be able to fetch submitted moves' signed certs, and be able to distinguish between whether they were a PPM or an HHG, or both for historical reasons. 

So, this does not replace the types, but adds a 3rd type, with comments about a date in the model for signed cert for when this new type was added. 

## Setup

Login locally as the NTS or HHG user, submit your move. Make sure the server doesn't error and you see a success banner on the Home page after submission.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=MB&modal=detail&selectedIssue=MB-4966&quickFilter=60) for this change


pkg/models/signed_certification.go
```diff
@@ -18,20 +18,24 @@ import (
...
-	// SignedCertificationTypeHHG captures enum value "HHG"
+	// SignedCertificationTypeHHG captures enum value "HHG" - deprecated 10/2020
 	SignedCertificationTypeHHG SignedCertificationType = "HHG"
+
+	// SignedCertificationTypeSHIPMENT captures enum value "SHIPMENT" for all shipment types from 10/2020
+	SignedCertificationTypeSHIPMENT SignedCertificationType = "SHIPMENT"
 )
 
 var signedCertifications = []string{
 	string(SignedCertificationTypePPM),
```
kimallen, line 35: I'm wondering if we can remove PPM and HHG here since this is used only in the validation, and we only CREATE and don't UPDATE the certifications, so removing wouldn't impact any fetching.

kimallen (commented):
Looks good, worked as expected.

I noticed that there is a `MakeSignedCertification` fn in testdatagen that uses `SignedCertificationTypePPM`.  Let's change that to SHIPMENT

Same with `models/signed_certification_test.go` that uses PPM and HHG types

swagger/internal.yaml
```diff
@@ -1625,6 +1625,7 @@ definitions:
       - PPM
       - PPM_PAYMENT
       - HHG
+      - SHIPMENT
```
kimallen, line 1628: I wonder if it would be reasonable to create two different `SignedCertificationType`s, one with the deprecated vaules, and one without.  We could use the one with only HHG and SHIPMENT in the `CreateSignedCertificationPayload`, and then use the one with deprecated values in the payload used in the return payloads `SignedCertificationPayload`.  I don't know what best practice is when there are deprecated values that still could be fetched, just an idea.

sojeri: I like this! I'd + if GH had an explicit + emoji.

kimallen: What is GH?

amitch23: github I think. Jeri is just saying that Github's emoji game is weak ;)

amitch23: @kimallen both the create handler (CreateSignedCertificationHandler) and the fetch handler (IndexSignedCertificationsHandler) both return the same payload (payloadForSignedCertificationModel) in the signed_certifications.go handler. This is currently a problem if we want to return a specific cert type depending on which handler is getting the request. Is your idea to create 2 different payloads that return 2 different types?

amitch23: resolved offline - Both handlers will return the same payload, this is just to limit the createpayload on the request coming in in the internal.yaml spec

kimallen (approved):
Thanks for the changes- LGTM!

PR #5107, opened 2020-10-28
MB-4546 add event trigger for updateMoveOrder

NamibiaLT:
## Description

Add an event trigger for the UpdateMoveOrderHandler and test that a notification is sent to Prime.

## Reviewer Notes

Please take a look at whether the query errors are properly handled and logged. 

## Setup

Nothing to setup.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4546) for this change


pkg/handlers/ghcapi/move_orders.go
```diff
@@ -144,6 +146,28 @@ func (h UpdateMoveOrderHandler) Handle(params moveorderop.UpdateMoveOrderParams)
 			return moveorderop.NewUpdateMoveOrderInternalServerError()
 		}
 	}
+	// Get move.orderID:
+	fmt.Println("moveOrder.ID", updatedOrder.ID)
```
shimonacarvalho, line 150: This was mine, pretty sure, but do remove it. We don't print anything in the normal functioning of the server.

pkg/handlers/ghcapi/move_orders.go
```diff
@@ -144,6 +146,28 @@ func (h UpdateMoveOrderHandler) Handle(params moveorderop.UpdateMoveOrderParams)
...
+	// Find the record where orderID matches moveOrder.ID
+	var move models.Move
+	query := h.DB().Where("orders_id = ?", updatedOrder.ID)
+	err = query.First(&move)
+
```
shimonacarvalho, line 155: Might be a good idea to log an error here, it's unusual to have an orders object where there is no move associated.

pkg/handlers/ghcapi/move_order_test.go
```diff
@@ -195,6 +195,61 @@ func (suite *HandlerSuite) TestUpdateMoveOrderHandlerIntegration() {
...
+	handler := UpdateMoveOrderHandler{
+		context,
+		moveorder.NewMoveOrderUpdater(suite.DB(), queryBuilder),
+	}
+
+	traceID, err := uuid.NewV4()
+	handler.SetTraceID(traceID)        // traceID is inserted into handler
+	response := handler.Handle(params) // This step also saves traceID into DB
+	suite.IsNotErrResponse(response)
+	moveOrderOK := response.(*moveorderop.UpdateMoveOrderOK) // Type assertion: Take the response, change the type to *moveorderop.UpdateMoveOrderOK, and set the variable moveOrderOK to the new type
```
shimonacarvalho, line 242: Maybe delete my explanation comment :)

pkg/handlers/ghcapi/move_order_test.go
```diff
@@ -195,6 +195,61 @@ func (suite *HandlerSuite) TestUpdateMoveOrderHandlerIntegration() {
...
+	moveOrdersPayload := moveOrderOK.Payload
+
+	suite.FatalNoError(err, "Error creating a new trace ID.")
+
+	suite.Assertions.IsType(&moveorderop.UpdateMoveOrderOK{}, response)
+	suite.Equal(moveOrdersPayload.ID, strfmt.UUID(moveOrder.ID.String()))
+	suite.HasWebhookNotification(moveOrder.ID, traceID)
+	// suite.HasWebhookNotification(updatedOrder.ID, traceID)
```
shimonacarvalho, line 250: Please remove commented line

pkg/handlers/ghcapi/move_order_test.go
```diff
@@ -195,6 +195,61 @@ func (suite *HandlerSuite) TestUpdateMoveOrderHandlerIntegration() {
 	suite.Equal(body.Sac, moveOrdersPayload.Sac)
 }
 
+// Test that a move order notification got stored Successfully
+func (suite *HandlerSuite) TestUpdateMoveOrderEventTrigger() {
```
shimonacarvalho, line 199: Thanks for adding a test!

shimonacarvalho (changes requested):
Awesome job, some changes requested. Let me know if you want to talk about any of them. 

Couple of notes on review conventions in our team:
Can be useful to mention if there's a test that someone can run. 
Add the label `roci` to the labels section, that way it shows up when we look for pending prs.

NamibiaLT:
Thanks @shimonacarvalho I added the changes. Can you let me know if I wrote the error logger correctly? Really appreciate the feedback!

pkg/handlers/ghcapi/move_orders.go
shimonacarvalho: Hey namibia, I'm still seeing the fmt.Println?

pkg/handlers/ghcapi/move_orders.go
shimonacarvalho: If you don't find the move, move will be an empty struct. Then you use `move.id` in the call below to `TriggerEvent`. Not sure what happens then - does it send an zero id? Or a nil? Or cause a nil dereference?

Ideally i think if you couldn't find the move, you want to send a `nil` in that value, and have the `TriggerEvent` handle it however it will.

shimonacarvalho (changes requested):
Hey Namibia, the log message itself looks good but there's an issue with what happens after you detect the error. I've commented at the line.

pkg/handlers/ghcapi/move_orders.go
sandy-wright: I was wondering - is it an error if there's no move? Maybe the order was created without the move for some reason? In that case, I might do an if statement like:

```
if err == nil {
    // do the event trigger code here
} else {
    log.Warning("ghcapi.UpdateMoveOrderHandler could not find move - no event created")
}
```

I'm not sure if that's right, but maybe that could be a good solution!

pkg/handlers/ghcapi/move_orders.go
NamibiaLT: Ok, it has been removed.

pkg/handlers/ghcapi/move_orders.go
shimonacarvalho: Hey Sandy - yeah it looks like in this case they could be updating the moveOrders prior to creating the first move and that's okay. However we are going to trigger the _event_ but add a ticket to make sure the notification package handles the situation correctly.

shimonacarvalho (approved):
Awesome work, looks good to me!

sandy-wright (approved):
LGTM!

PR #5114, opened 2020-10-28
[MB-4523] add service payment_request_reviewed_processor

jacquelineIO:
## Description

Implement service object `ProcessReviewedPaymentRequest`. 

ProcessApprovedPaymentRequests service object
* Fetch list of approved PRs
* Call generate EDI
* Send to Syncada/GEX
* Update PR status

## Reviewer Notes

Eventually there will be an API or async task that can call this method. Right now you can only run unit test to verify it. 

Also trying a raw query for updates that I haven't seen used in our code before. The idea is coming from https://stackoverflow.com/a/18799497 and is used here: https://github.com/transcom/mymove/pull/5114/files#diff-80e10c88689cf867733c8a3b904f05c67b66f62387fa6e88b738d312b21d0972R90

This is an example of the output produced by the unit test.
```sql
UPDATE payment_requests AS pr SET
    status = c.status
FROM (VALUES
	('dae242fb-987f-485c-9eed-5d8eee4736c9'::uuid,'SENT_TO_GEX'::payment_request_status),
        ('34e542e8-75fb-4aae-9fe1-fc35c0545542'::uuid,'SENT_TO_GEX'::payment_request_status),
        ('e4b148ee-3483-444a-ba71-1e037a37627d'::uuid,'SENT_TO_GEX'::payment_request_status),
        ('ba7d18e3-87e6-4dfb-bb6d-c6ee27e38361'::uuid,'SENT_TO_GEX'::payment_request_status)
    ) AS c(id, status)
WHERE c.id = pr.id;
```

## Setup

```sh
echo "Code goes here"
```

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4523) for this change

## Screenshots

n/a


pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -0,0 +1,134 @@
...
+// NewPaymentRequestReviewedProcessor returns a new payment request reviewed processor
+func NewPaymentRequestReviewedProcessor(db *pop.Connection, logger Logger, generator services.GHCPaymentRequestInvoiceGenerator) services.PaymentRequestReviewedProcessor {
+	return &paymentRequestReviewedProcessor{db, logger, generator}
+}
+
+func (p *paymentRequestReviewedProcessor) ProcessReviewedPaymentRequest() error {
+
+	// Fetch all payment request that have been reviewed
+	reviewedPaymentRequestFetcher := NewPaymentRequestReviewedFetcher(p.db)
+	reviewedPaymentRequests, err := reviewedPaymentRequestFetcher.FetchReviewedPaymentRequest()
```
reggieriser, line 30: Since that's a service object, should we make it part of the `paymentRequestReviewedProcessor` struct so it can be injected like `ediGenerator`?  That way, you could mock it in tests and test the path where it fails.

jacquelineIO: @reggieriser  great point! made update. thanks

pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/transcom/mymove/pkg/services"
+)
+
+type paymentRequestReviewedProcessor struct {
+	db                            *pop.Connection
+	logger                        Logger
+	reviewedPaymentRequestFetcher services.PaymentRequestReviewedFetcher
+	ediGenerator                  services.GHCPaymentRequestInvoiceGenerator
+}
+
```
jacquelineIO, line 21: @harshamurthy  your PR is good. I'm going to have to re-work it a tiny bit to make `NewSyncadaSFTPSession` as  part of my struct here. I need to be able to mock it out. So the function `func SendToSyncada(edi string) error {`) might get moved to this file. There is nothing for you do, I'm just thinking about what I will need to do once it is in my PR.

pkg/payment_request/send_to_syncada.go
```diff
@@ -2,27 +2,31 @@ package paymentrequest
...
 	"time"
 
-	"github.com/transcom/mymove/pkg/services/invoice"
+	"github.com/transcom/mymove/pkg/services"
 )
 
 // SendToSyncada send EDI file to Syncada for processing
-func SendToSyncada(edi string) error {
-	//TODO: Send to Syncada via GEX needs to be implemented
+func SendToSyncada(edi string, gexSender services.GexSender, sftpSender services.SyncadaSFTPSender, sendEDIFile bool, logger Logger) error {
```
reggieriser, line 12: This feels like the type of scenario where both `GexSender` and the `SyncadaSFTPSender` could implement a common `EDISender` interface with a `SendEDI` method, then you wouldn't need this type of switching logic to determine what to do in each case.  I believe the SFTP solution is temporary, though, so that may be an abstraction we won't ultimately need.

jacquelineIO: @reggieriser that did cross my mind. Since we aren't using GEX right now, a good user story explaining this as a option would be good. I also don't want to jump through hoops for that either. So only if it makes sense.

pkg/payment_request/payment_request_helper.go
```diff
@@ -4,5 +4,6 @@ import "github.com/gobuffalo/pop/v5"
 
 // RequestPaymentHelper is a helper to connect to the DB
 type RequestPaymentHelper struct {
-	DB *pop.Connection
+	DB     *pop.Connection
+	Logger Logger
```
reggieriser, line 8: Where do we use `Logger`?  I don't see it being set when in any of the places where this struct is created.

pkg/services/invoice/syncada_sftp_sender.go
```diff
@@ -30,6 +31,11 @@ func NewSyncadaSFTPSession(port string, userID string, remote string, password s
 	}
 }
 
+// InitNewSyncadaSFTPSession initialize a NewSyncadaSFTPSession and return services.SyncadaSFTPSender
+func InitNewSyncadaSFTPSession() services.SyncadaSFTPSender {
+	return NewSyncadaSFTPSession(os.Getenv("SYNCADA_SFTP_PORT"), os.Getenv("SYNCADA_SFTP_USER_ID"), os.Getenv("SYNCADA_SFTP_IP_ADDRESS"), os.Getenv("SYNCADA_SFTP_PASSWORD"), os.Getenv("SYNCADA_SFTP_INBOUND_DIRECTORY"))
```
reggieriser, line 36: I think we typically do this type of setup (via viper) when the server starts up and then put it into the context.  At least, that's the way the gex sender is done.  If this is just temporary, it's probably OK.  But if longer lived, we should probably have a story to refactor it.

pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -0,0 +1,173 @@
...
+		gexSender:                     gexSender,
+		sftpSender:                    sftpSender,
+		runSendToSyncada:              runSendToSyncada}
+}
+
+// InitNewPaymentRequestReviewedProcessor initialize NewPaymentRequestReviewedProcessor for production use
+func InitNewPaymentRequestReviewedProcessor(db *pop.Connection, logger Logger, sendToSyncada bool) services.PaymentRequestReviewedProcessor {
+	reviewedPaymentRequestFetcher := NewPaymentRequestReviewedFetcher(db)
+	generator := invoice.NewGHCPaymentRequestInvoiceGenerator(db)
+	SFTPSession := invoice.InitNewSyncadaSFTPSession()
```
reggieriser, line 50: Nit: I'd probably use `sftpSession` here since this is a local variable just to avoid confusion with it looking like something exported.

pkg/payment_request/payment_request_helper.go
jacquelineIO: @reggieriser that part of the code was removed when I merged in Harsha's PR. Good catch

pkg/services/invoice/syncada_sftp_sender.go
jacquelineIO: @reggieriser hmm OK this is also coming in from @harshamurthy  PR. I wonder if @harshamurthy  knows why he was calling it this way. I was curious but thought it might have been because this isn't a CLI, but I should have asked instead of assuming.

harshamurthy: @reggieriser are you talking about the environment variables? Because there are other examples in the code where we use the os.Getenv pattern which is why I did it this way here.

harshamurthy: It also seems the gex sender is only called in the context of a CLI utility? I may not fully understand where viper can be used, but I thought it was used to parse command line flags.

pkg/services/payment_request/payment_request_reviewed_processor.go
jacquelineIO: fixed in update

pkg/payment_request/payment_request_helper.go
jacquelineIO: fixed in update

pkg/services/invoice/syncada_sftp_sender.go
reggieriser: I was talking about how we [configure the GexSender here](https://github.com/transcom/mymove/blob/3423a6bc6314d2e07c0f98bba125fa9b0428b876/cmd/milmove/serve.go#L585-L610).  But if this is going to be run as part of a separate task, maybe that doesn't apply here or applies in a different way.  It looks like we still [use viper as part of a task](https://github.com/transcom/mymove/blob/e01d317e6bcef025ddb16df8898e85d5789678c9/cmd/milmove-tasks/post_file_to_gex.go#L133-L139).  I'm fine with what's here for now, but just something to investigate with later stories.

pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -0,0 +1,173 @@
...
+		sftpSession)
+}
+
+func (p *paymentRequestReviewedProcessor) ProcessReviewedPaymentRequest() error {
+
+	// Fetch all payment request that have been reviewed
+	reviewedPaymentRequests, err := p.reviewedPaymentRequestFetcher.FetchReviewedPaymentRequest()
+	if err != nil {
+		return fmt.Errorf("function ProcessReviewedPaymentRequest failed call to FetchReviewedPaymentRequest: %w", err)
+	}
```
reggieriser, line 70: It might be worth testing the length of `reviewedPaymentRequests` here to see if it's zero, then just returning right away rather than having to go through all the other stuff below.

pkg/services/payment_request/payment_request_reviewed_processor_test.go
```diff
@@ -0,0 +1,305 @@
...
+		err := paymentRequestReviewedProcessor.ProcessReviewedPaymentRequest()
+		suite.NoError(err)
+
+	})
+
+	suite.T().Run("process reviewed payment request, successfully test init function", func(t *testing.T) {
+		// Run init with no issues
+		_ = InitNewPaymentRequestReviewedProcessor(suite.DB(), suite.logger, false)
+	})
+}
```
reggieriser, line 305: A nice-to-have may be a test that fails on some but succeeds on others in the same `ProcessReviewedPaymentRequest` call.  I think you can tell the mock to fail just on the first payment request, for instance.

reggieriser (approved):
I added some comments that I don't think affect whether this works or not.  Going to go ahead and approve -- the tests ran fine -- but if something I noted seems worthwhile, we can add another story to consider it.

PR #5115, opened 2020-10-28
MB-3095 fix shipment date check

NamibiaLT:
## Description
Fix check on requestShipmentDate to verify that it returns zero rather than nil. Added an unhappy path test for this fix.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3095) for this change


pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -136,7 +136,7 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
 		}
 
 		// check that required items to create shipment are present
-		if shipment.RequestedPickupDate == nil {
+		if shipment.RequestedPickupDate.IsZero() {
```
sojeri, line 139: this behaving correctly may break the customer side of the demo, so please hold off on merging until #5116 gets merged. if you want, I'm happy to add a test case to this branch for the NTS-R edge case. I'm not familiar enough with the other mto shipment types to know whether any of them might also not require pickup details, but we can cross that bridge when we get there. :)

NamibiaLT: @sandy-wright , my incination is to branch off of mine to add the test case. What do you think?

sandy-wright: @NamibiaLT I think that's a good idea. Thanks for spotting this so quickly Jeri!

sandy-wright: Although I am also curious about the failing integration tests - it might be related to the NTS-R shipments being unable to get past this step and not routine flakiness this time. If that's the case, we may need to flip it around and have this PR branch off of @sojeri's

pkg/services/mto_shipment/mto_shipment_creator_test.go
```diff
@@ -70,6 +71,36 @@ func (suite *MTOShipmentServiceSuite) TestCreateMTOShipmentRequest() {
 		suite.NotEmpty(invalidErr.ValidationErrors)
 	})
 
+	// Unhappy path
+	suite.T().Run("When required requested pickup dates are zero (required for NTS & HHG shipment types)", func(t *testing.T) {
+		// this check verifies that if requestedPuckupDate is Zero it returns an error
```
sandy-wright, line 76: Oh, there's a small typo here we missed! Should be `requestedPickupDate`

sandy-wright (commented):
This looks good to me! I'm gonna hold off on an approval until https://github.com/transcom/mymove/pull/5116 gets merged and this branch gets updated, but it should be good to go after that 👍

pkg/services/mto_shipment/mto_shipment_creator.go
sojeri: I suspect they'll all pass once y'all merge the lastest from master which now includes #5116, but please at here or on slack if that's not the case and you want to pair on resolving any remaining nts/r issues.

sandy-wright (approved):
LGTM! ✨

PR #5116, opened 2020-10-28
fix nts-r logic hole

sojeri:
## Description

Fix NTS-R logic hole by only throwing an error if requested pickup date is nil on non NTS-R type shipments.

## Reviewer Notes

@NamibiaLT -- lmk if you want me to branch off your branch to add another test case for this NTS-R type edge case. since we weren't checking the zero value before, I think this error check wasn't causing any problems--- but once your change merges this will become an issue for the customer side of creating NTS-R shipments.

### DIFF
--- a/pkg/services/mto_shipment/mto_shipment_creator.go
+++ b/pkg/services/mto_shipment/mto_shipment_creator.go
@@ -123,8 +123,8 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
 				return fmt.Errorf("failed to create pickup address %#v %e", verrs, err)
 			}
 			shipment.PickupAddressID = &shipment.PickupAddress.ID
-		} else if shipment.ShipmentType == models.MTOShipmentTypeHHG {
-			return services.NewInvalidInputError(uuid.Nil, nil, nil, "PickupAddress is required to create an HHG type MTO shipment")
+		} else if shipment.ShipmentType != models.MTOShipmentTypeHHGOutOfNTSDom {
+			return services.NewInvalidInputError(uuid.Nil, nil, nil, "PickupAddress is required to create an HHG or NTS type MTO shipment")
 		}
 
 		if shipment.DestinationAddress != nil {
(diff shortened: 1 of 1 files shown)


pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -123,8 +123,8 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
 				return fmt.Errorf("failed to create pickup address %#v %e", verrs, err)
 			}
 			shipment.PickupAddressID = &shipment.PickupAddress.ID
-		} else if shipment.ShipmentType == models.MTOShipmentTypeHHG {
-			return services.NewInvalidInputError(uuid.Nil, nil, nil, "PickupAddress is required to create an HHG type MTO shipment")
+		} else if shipment.ShipmentType != models.MTOShipmentTypeHHGOutOfNTSDom {
```
sojeri, line 126: this is checking if not NTS-R

lynzt:
@sojeri can you add some setup instructions to test this change?

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -136,8 +136,8 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
 		}
```
sandy-wright, line 136: Do we need to add a check here to see if the shipment type is NTSr and then return an invalid input error if so? I was under the impression that delivery details were required for NTSr shipments, but I also know very little about them (so could very easily be wrong).

sojeri:
@lynzt you'd have to pull all three branches locally to repro a difference. this branch, Namibia's branch, and my branch off Namibia's branch.

sandy-wright (approved):
This looks good, thanks for making these changes! I had a question about whether or not the `destinationAddress` needs to be required for NTSr shipments, so that may or may not be an issue. It might also be nice to either make or update some of the test cases for the address checks, but I'm not going to make that a blocker. 

Either way, I ran a bunch of tests and everything is still working smoothly on the Prime side, so these updates are solid!

pkg/services/mto_shipment/mto_shipment_creator.go
sandy-wright: Wow, I sure picked the worst line to comment under. This is the block of code in question:

```
		if shipment.DestinationAddress != nil {
			verrs, err = txBuilder.CreateOne(shipment.DestinationAddress)
			if verrs != nil || err != nil {
				return fmt.Errorf("failed to create destination address %#v %e", verrs, err)
			}
			shipment.DestinationAddressID = &shipment.DestinationAddress.ID
		}
```

sojeri: @sandy-wright destination address is not required for NTSr shipments. it's the same experience as HHG where the customer can say they don't know their address and the move counselor will follow up later to get those details.

lynzt (approved):
LGTM 🚢

pkg/services/mto_shipment/mto_shipment_creator.go
sandy-wright: great, thanks for the clarification!

sandy-wright (approved):
Looks good! 🚀

PR #5118, opened 2020-10-28
MB-4934 TOO can filter status column - UI

donaldthai:
## Description

This PR adds a status column filter for the TOO Move Queue. This uses a multi select checkbox.
Includes [Date Submitted Filter](https://github.com/transcom/mymove/pull/5132)
## Reviewer Notes

- There's a weird issue where when you have more than two options selected and then focus on the input again the selected options text jumps to the left. Not sure what's going on but I couldn't find a solution right away. Will make a story bug for this.

## Setup

```sh
make server_run
make office_client_run
```

1. Navigate to officelocal:3000, sign in as TOO
2. Go to `officelocal:3000/moves/queue`
3. Filter status should be there and working

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4934) for this change
* [this article](https://react-select.com/home) explains more about the approach used.

## Screenshots

Oct-30-2020 10-09-42


brooksr4 (changes requested):
The form inputs function well, although they look kind of funny because they should be on a separate row. See image below:
image

brooksr4 (dismissed):
One more thing, for the status filter if we can middle align the checkbox since there's a carriage return, that would make it look a little cleaner.

suzubara (approved):
I left a few minor comments but nothing is blocking. Playing with the filters locally, it feels like there's probably some UX finessing that could be done (like adding an empty state, and UI improvements) but those can probably be taken care of later. This is a really great start to a complex feature!

donaldthai:
@brooksr4 Should be all good now! 
image

src/constants/queues.js
```diff
@@ -0,0 +1,20 @@
...
+  Paid: 'Paid',
+};
+
+export const BRANCH_OPTIONS = {
+  ARMY: 'Army',
+  NAVY: 'Navy',
+  MARINES: 'Marine Corps',
+  AIR_FORCE: 'Air Force',
+  COAST_GUARD: 'Coast Guard',
+};
```
monfresh, line 20: What do y'all think about making this an explicit hash with `value` and `label` keys, so that it's clear which is which? Right now, with some of them having the same text on both sides, it's not immediately clear which one you would need to update.

donaldthai: So instead of the object key value you would want to define the `value` and `label`. So pre-made like this?
```javascript
var branchFilterOptions = [
{ label: 'Air Force', value: 'AIR_FORCE' },
...
];
```

monfresh: Exactly.

monfresh: It doesn't have to be in this PR. I can write a JIRA story for later.

donaldthai: Since we aren't using the options for anything else, I don't mind that format.

amanda5iu (approved):
Approved pending removal of the `x` from the filter fields, [as discussed in Slack](https://ustcdp3.slack.com/archives/CP4979J0G/p1604347471088900).

donaldthai:
> Approved pending removal of the `x` from the filter fields, [as discussed in Slack](https://ustcdp3.slack.com/archives/CP4979J0G/p1604347471088900).

Removed the `x`:
image

PR #5119, opened 2020-10-28
add ntsr requestedPickupDate test

sojeri:
## Description

- adds ntsr requestedPickupDate test 
- fixes a few variable names to reference their shipment types
- updates testdatagen's create mto shipment method to only set valid data fields on a per shipment type basis

note that this new test will continue failing until #5116 is merged.

## Reviewer Notes

- Sandy & Namibia -- added you because we've been discussing this across slack and the other two PRs.
- Ron -- added you b/c you added the other NTSr test. Thought you might have thoughts on the testdatagen change. :)

### DIFF
--- a/pkg/services/mto_shipment/mto_shipment_creator_test.go
+++ b/pkg/services/mto_shipment/mto_shipment_creator_test.go
@@ -1,7 +1,6 @@
 package mtoshipment
 
 import (
-	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/services/mto_shipment/mto_shipment_creator_test.go
```diff
@@ -179,24 +194,24 @@ func (suite *MTOShipmentServiceSuite) TestCreateMTOShipmentRequest() {
 	})
 
 	suite.T().Run("If the move already has a submitted NTSr shipment, it should return a validation error", func(t *testing.T) {
-		ntsShipment := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+		ntsrShipment := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
```
sojeri, line 197: lmk if you want me to update this test and the new one above to create an MTO shipment with the data fields as they would actually be for NTSr shipments. this test is using the default data which includes a bunch of fields that are only in HHG shipments. the same issue is true for the NTS test which I think is right above this one in code. (you'll have to click to expand unchanged code.)

sojeri: went ahead and did this b/c it felt right

sandy-wright (approved):
These updates look really good, and I totally support the changes to the `testdatagen` function to make it more flexible based on shipment type. This is super awesome, thanks!! ✨ 

(On a side note, the fact that we use "destination" and then "delivery" in the address field names will always make me weep inside)

sojeri:
@NamibiaLT looks like Sandy has signed off. are you cool if I merge this to your branch now?

sojeri:
> (On a side note, the fact that we use "destination" and then "delivery" in the address field names will always make me weep inside)

yes, same for me every time. :sob:

PR #5124, opened 2020-10-30
Mb 4969 submit transition error

kimallen:
## Description

Uncovered by the issue before the MB-4966 update, we discovered that the creation of the signed certification could fail, but the submission could still go through.  This PR moves the creation of the signed certification into a transaction on the backend, rather than calling it as a separate call on the frontend.

## Reviewer Notes

Additional updates include the following:
- Removes ppmSubmit date from SubmitForApproval and utilizes server side current time instead
- Makes a copy of `SaveMoveDependencies` from the model and applies it as a function of the `SubmitMoveForApproval handler`.  (the move model version of `SaveMoveDependencies` I left because it is used in many other parts of the application and is outside the scope of this story to refactor)
- saves the certificate information within the db transaction inside of `h.saveMoveDependencies`
- removes the `latestSignedCertificate` code in `Legalese/index.jsx` because users will never return to this page after certificate creation

The CreateSignedCertification handler and the corresponding frontend code remains because it is still utilized for a PPM Payment request.  `move.SaveMoveDependencies` is used on the PPM office side to cancel a move.

## Setup

To manually test, create a validation error on the certificate, and see that nothing gets saved.  You can change the value of `certificate.certificationType` in `Legalese/index.jsx` to an invalid string (like 'PPM').  

When tests are ready. . .
`go test ./pkg/handlers/internalapi`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4969) for this change


src/scenes/Legalese/index.jsx
```diff
@@ -33,31 +33,22 @@ export class SignedCertification extends Component {
     this.props.loadPPMs(this.props.moveId);
   }
 
-  submitCertificate = () => {
-    const signatureTime = moment().format();
+  handleSubmit = () => {
+    const pendingValues = this.props.values;
```
sojeri, line 37: why do we need two refs to values? I'm seeing pendingValues here and values on next line.

pkg/handlers/internalapi/moves_test.go
```diff
@@ -179,42 +179,87 @@ func (suite *HandlerSuite) TestShowMoveWrongUser() {
...
+		context.SetNotificationSender(notifications.NewStubNotificationSender("milmovelocal", suite.TestLogger()))
+		handler := SubmitMoveHandler{context}
+		response := handler.Handle(params)
+
+		// Then: expect a 200 status code
+		suite.Assertions.IsType(&moveop.SubmitMoveForApprovalOK{}, response)
+		okResponse := response.(*moveop.SubmitMoveForApprovalOK)
+
+		// And: Returned query to have an approved status
+		suite.Assertions.Equal(internalmessages.MoveStatusSUBMITTED, okResponse.Payload.Status)
```
sojeri, line 260: should we also verify before submission that status is draft?

pkg/handlers/internalapi/moves_test.go
```diff
@@ -179,42 +179,87 @@ func (suite *HandlerSuite) TestShowMoveWrongUser() {
...
+		context := handlers.NewHandlerContext(suite.DB(), suite.TestLogger())
+		context.SetNotificationSender(notifications.NewStubNotificationSender("milmovelocal", suite.TestLogger()))
+		handler := SubmitMoveHandler{context}
+		response := handler.Handle(params)
+
+		// Then: expect a 200 status code
+		suite.Assertions.IsType(&moveop.SubmitMoveForApprovalOK{}, response)
+		okResponse := response.(*moveop.SubmitMoveForApprovalOK)
+
+		// And: Returned query to have an approved status
```
sojeri, line 218: nit: this isn't approved; it's submitted

pkg/handlers/internalapi/moves_test.go
```diff
@@ -179,42 +179,87 @@ func (suite *HandlerSuite) TestShowMoveWrongUser() {
...
+			Signature:                swag.String("Jane Doe"),
+		}
+		newSubmitMoveForApprovalPayload := internalmessages.SubmitMoveForApprovalPayload{Certificate: &certificate}
+
+		params := moveop.SubmitMoveForApprovalParams{
+			HTTPRequest:                  req,
+			MoveID:                       strfmt.UUID(move.ID.String()),
+			SubmitMoveForApprovalPayload: &newSubmitMoveForApprovalPayload,
+		}
+		// And: a move is submitted
```
sojeri, line 208: When: a move is submitted?

pkg/handlers/internalapi/moves.go
```diff
@@ -160,14 +169,17 @@ func (h SubmitMoveHandler) Handle(params moveop.SubmitMoveForApprovalParams) mid
 	}
 	logger = logger.With(zap.String("moveLocator", move.Locator))
 
-	submitDate := time.Time(*params.SubmitMoveForApprovalPayload.PpmSubmitDate)
-	err = move.Submit(submitDate)
+	err = move.Submit()
```
sojeri, line 172: why calculate a new datetime in move.Submit when we're already calculating one here in saveMoveDependencies?

src/scenes/Legalese/index.jsx
```diff
@@ -33,31 +33,22 @@ export class SignedCertification extends Component {
...
     const { currentPpm, moveId, values } = this.props;
+    const landingPath = '/';
+    const submitDate = moment().format();
```
sojeri, line 40: are we actually using this date? I see lots of datetime generation happening server side.

sojeri (commented):
left a couple comments about dates. I'm not sure I understand the changes you've made related to those or why we need to separately generate so many different ones for a single transaction.

I think my remaining comments are little nits in the test code which you can disregard if you think they don't add anything.

src/scenes/Legalese/index.jsx
kimallen: Nope.  It's been a balance between making lots of refactors and just making sure I take care of the requests on the ticket.  We were also passing in a 'now' date into the submitMove handler directly, but I took that out.

amitch23 (approved):
Great job, works as expected! The validation error is clear and helpful as well. :) I have a couple questions that may be outside the scope of the PR, or are just due to lack of understanding:
1. I second Jeri's questions about the dates being set/passed in both the F/E and backend. I see that the date is being pre-populated in the legalese form and is passed to the API, so it seems like we're opening up to a small risk of those 2 things being out of sync. 
2. Since I see the PPM ID passed in as a param, I would expect to see the mtoShipmentID as well, but I don't. I don't totally understand why we're storing the PPMId on the signed certification but not the mtoShipment - is this because of being able to submit a signed cert of type PPM_PAYMENT, or handling the saving of the advance?

pkg/handlers/internalapi/moves_test.go
kimallen: We do check that it's the correct status in the `Submit` function.  Do you mean should we check on the frontend before they submit?

pkg/handlers/internalapi/moves_test.go
kimallen: (this was the existing test- just looks like I wrote it because I put it into a subtest)

pkg/handlers/internalapi/moves.go
kimallen: Refactored it.

src/scenes/Legalese/index.jsx
kimallen: After discussion, I'm leaving this submit date since it's existing behavior to date the document from when the user landed on the page Discussion: https://ustcdp3.slack.com/archives/CTNTFJSBA/p1604092636412600?thread_ts=1604085524.412200&cid=CTNTFJSBA

src/scenes/Legalese/index.jsx
kimallen: > Great job, works as expected! The validation error is clear and helpful as well. :) I have a couple questions that may be outside the scope of the PR, or are just due to lack of understanding:
> 
> 1. I second Jeri's questions about the dates being set/passed in both the F/E and backend. I see that the date is being pre-populated in the legalese form and is passed to the API, so it seems like we're opening up to a small risk of those 2 things being out of sync.

Good point about the date being pre-populated. Keeping the legalese signature to the frontend date based on Slack convo.

> 2. Since I see the PPM ID passed in as a param, I would expect to see the mtoShipmentID as well, but I don't. I don't totally understand why we're storing the PPMId on the signed certification but not the mtoShipment - is this because of being able to submit a signed cert of type PPM_PAYMENT, or handling the saving of the advance?

I actually don't think we need to pass in the ppmId at all.  However, I think refactoring that is outside the scope of this PR.  When we delve into PPM, I think we could revisit.

pkg/handlers/internalapi/moves_test.go
sojeri: I just mean earlier in this test.

src/scenes/Legalese/index.jsx
sojeri: I think you left your comments in the wrong place, Kim?

cc @amitch23 some replies to your comments here

pkg/handlers/internalapi/moves.go
```diff
@@ -160,14 +169,18 @@ func (h SubmitMoveHandler) Handle(params moveop.SubmitMoveForApprovalParams) mid
 	}
 	logger = logger.With(zap.String("moveLocator", move.Locator))
 
-	submitDate := time.Time(*params.SubmitMoveForApprovalPayload.PpmSubmitDate)
+	submitDate := time.Now()
```
sojeri, line 172: shouldn't this one be using the time from the payload as well?

src/scenes/Legalese/index.jsx
kimallen: Yeah, I don't know how they ended up here.  I quote-replied to her comment 🤷

sojeri (approved):
changes look good! I think you missed one place where the date used to be from the payload but is now being recalculated. other than that, I think this is ready to go.

approving with the assumption that you'll fix that date. :)

src/scenes/Legalese/index.jsx
sojeri: GH madness, haha. I wish they made it easier to thread off comments in the main feed which aren't attached to lines of code!

pkg/handlers/internalapi/moves.go
kimallen: I don't think so?  I did some refactoring so that we are using server-generated date for both the ppm and the move submit dates.  Previously, we were getting the ppm submit date from the client and the move submit date from the server.  So I was keeping the submitDate coming from the server, while the certificate signing date comes from the client.

sojeri: so the line of code you're changing was pulling the data from the FE client's params......

pkg/models/move.go
```diff
@@ -127,17 +127,18 @@ func (m *Move) ValidateUpdate(tx *pop.Connection) (*validate.Errors, error) {
...
 	m.Status = MoveStatusSUBMITTED
-	m.SubmittedAt = swag.Time(time.Now())
+	submitDate := swag.Time(submittedDate)
+	m.SubmittedAt = submitDate
 
 	// Update PPM status too
 	for i := range m.PersonallyProcuredMoves {
 		ppm := &m.PersonallyProcuredMoves[i]
-		err := ppm.Submit(submitDate)
+		err := ppm.Submit(*submitDate)
```
sojeri, line 141: so this date used to come from the client and now it's from the server. are we sure that's OK?

kimallen: I had brought this up earlier about whether it was better to get dates from the client v. server.  Suz said server.  I didn't see any reason ppm should use a client date and the move should use a server date, so I refactored it.  But I guess there could be something I'm not thinking of.  I was trying to improve the code, but can revert if there's reason to.

sojeri: nah, I think I  missed that part of the thread. merge away :)

if we have opinions later for legal reasons, presumably all the different dates will come up then. :)

PR #5149, opened 2020-11-03
MMDM-1742 Add migration to remove Lee's CAC access

amitch23:
## Description

Removes CAC access. This doesn't have to be a secure migration because we're only identifying the sha256 digest in dev/experimental/staging/prod. And if it doesn't exist in those environments this is a noop.

## Reviewer Notes

NOTE: This needs to be deployed manually to experimental to apply in that environment.

## Setup

```make db_dev_run db_dev_migrate```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1323,7 +1323,7 @@ workflows:
           # if testing on experimental, you can disable these tests by using the commented block below.
           filters:
             branches:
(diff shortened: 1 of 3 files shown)


reggieriser (approved):
LGTM!  Ran migrations locally and saw that Lee's record in `client_certs` had been deleted as expected.

PR #5150, opened 2020-11-04
MB-5119 TOO Pagination. - Feature Branch

noelledusahel:
## Description

- Build a storybook component for pagination on the TXO queue
- [MB-5120 Hook up the backend](https://github.com/transcom/mymove/pull/5189)
## Reviewer Notes

how does the pagination bar behave on mobile? 
Screen Shot 2020-11-11 at 11 13 31 AM

*the gif below has an option for 1 result per page only to show pagination working on a small set of results*
pagination
## Setup

For the Storybook Pagination Component:
```sh
make storybook
```

For in app pagination, sign in as TOO office user
:
```sh
make db_dev_e2e_populate
make server_run
make office_client_run
``` 

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/secure/RapidBoard.jspa?rapidView=27&modal=detail&selectedIssue=MB-5119) for this change

## Screenshots


src/components/Table/Table.jsx
```diff
@@ -2,52 +2,92 @@
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 
+import { ReactComponent as ChevronLeft } from '../../shared/icon/chevron-left.svg';
+import { ReactComponent as ChevronRight } from '../../shared/icon/chevron-right.svg';
```
donaldthai, line 6: Heard from @suzubara that we're moving away from icons we upload in place of fontawesome. Found the chevron icons for fontawesome: https://fontawesome.com/icons?d=gallery&q=chevron&m=free

Here's an example of how we import them: https://github.com/transcom/mymove/blob/master/src/components/Table/Filters/MultiSelectCheckBoxFilter.jsx#L6

src/components/Table/Table.jsx
```diff
@@ -2,52 +2,92 @@
...
+              <p>rows per page</p>
+            </div>
+          </div>
+          <div className={styles.tableControlPagination}>
+            <c disabled className={styles.usaButtonUnstyled}>
+              <span className="icon">
+                <ChevronLeft />
+              </span>
+              <span>Prev</span>
+            </c>
```
donaldthai, line 76: What's `<c></c>` suppose to be? The console is throwing an error for this: 
image

src/components/Table/Table.jsx
```diff
@@ -2,52 +2,92 @@
...
+            <select className={styles.usaSelect} name="table-pagination">
+              <option value="1">1</option>
+              <option value="2">2</option>
+              <option value="3">3</option>
+            </select>
+            <c className={styles.usaButtonUnstyled}>
```
donaldthai, line 82: Same here, what's `<c></c>`?

src/components/Table/Table.jsx
```diff
@@ -2,52 +2,92 @@
...
+              <span className="icon">
+                <ChevronLeft />
+              </span>
+              <span>Prev</span>
+            </c>
+            <select className={styles.usaSelect} name="table-pagination">
+              <option value="1">1</option>
+              <option value="2">2</option>
+              <option value="3">3</option>
+            </select>
```
donaldthai, line 81: Mind replacing the `<select>` with our TrussWorksUSWDS component? I think it'll be easier to maintain and don't have to worry about styling too much if we use that component. Here's an example: https://github.com/transcom/mymove/blob/master/src/components/Table/Filters/SelectFilter.jsx#L4

donaldthai: I also notice the icon is different than the USWDS version: 
regular `<select>` - single chevron down
image

USWDS - double chevron up down
image

src/components/Table/Table.module.scss
```diff
@@ -51,4 +51,54 @@
...
+.paginationSectionWrapper {
+ display: flex;
+ flex-wrap: wrap;
+ justify-content: space-between;
+ flex-grow: 1;
+
+   .paginationTableWrapper {
+    margin: 20px;
+    min-width: 173px;
+      font-size: 12px;
```
donaldthai, line 68: I think we should be able to remove this `font-size` since the font-size is set globally at the body.

donaldthai: I think if we import the TrussWorksUSWDS component, we won't need to worry about font-size either when the page resizes.

src/components/Table/Table.module.scss
```diff
@@ -51,4 +51,54 @@
...
+  .tableControlPagination {
+    display: flex;
+    align-items: center;
+    @include u-margin-y(1);
+    .usaSelect {
+      width: 89px;
+      height: 38px;
+      @include u-margin(0);
+      @include u-margin-x('105');
+    }
```
donaldthai, line 94: Should be able to remove `display:flex` from `.tableControlPagination` and `tableControlRowsPerPage` since `display:flex` is set in `.paginationSectionWrapper`

donaldthai (changes requested):
Looking good so far. I do think there's some changes we could do though. Comments on code. Let me know if you wanna pair or have questions.

src/components/Table/Table.jsx
noelledusahel: tbh i got it from here https://storybook.dp3.us/?path=/story/components-tables--table-elements, but i'm seeing in our code `Button` from uswds is being used

src/components/Table/Table.jsx
```diff
@@ -2,52 +2,92 @@
...
-                      {cell.render('Cell')}
-                    </td>
-                  );
-                })}
-              </tr>
-            );
-          })}
-        </tbody>
-      </table>
+    <div className={styles.inlineBlock}>
```
suzubara, line 13: What is the reason for wrapping a Table inside an inline-block div?

src/components/Table/Table.jsx
```diff
@@ -2,52 +2,92 @@
...
+                        {cell.render('Cell')}
+                      </td>
+                    );
+                  })}
+                </tr>
+              );
+            })}
+          </tbody>
+        </table>
+        <div className={styles.paginationSectionWrapper}>
```
suzubara, line 59: Should the pagination markup be rendered for every table? What about tables without pagination?

src/components/Table/Table.jsx
suzubara: I'm getting the same error. Maybe this is suppose to be a `Button`?

src/components/Table/Table.module.scss
```diff
@@ -51,4 +51,54 @@
...
+  }
+
+
+  .tableControlRowsPerPage {
+    display: flex;
+    align-items: center;
+    .usaSelect {
+      width: 89px;
+      height: 38px;
```
suzubara, line 79: We shouldn't be setting explicit height/width on elements unless absolutely necessary.

src/components/Table/Table.jsx
```diff
@@ -2,52 +2,92 @@
...
+          <div className={styles.paginationTableWrapper}>
+            <div className={styles.tableControlRowsPerPage}>
```
suzubara, line 61: I don't think both of these divs are necessary

src/components/Table/Table.stories.jsx
```diff
@@ -120,6 +120,37 @@ const CreatedTableWithFilters = () => {
...
+  );
+  // eslint-disable-next-line react/jsx-props-no-spreading
+  return <Table {...propsWithPagination} />;
+};
+
 export const TXOTable = () => <CreatedTable />;
 
 export const TXOTableFilters = () => <CreatedTableWithFilters />;
+
+export const TXOTablePagination = () => <CreateTableWithPagination />;
```
suzubara, line 156: I am seeing pagination controls for all of the table stories, when I would expect to only see it for this one. Should there be Table stories where no pagination is displayed?

also, it might be good to include enough test data in the pagination table story for the pagination to be used.

suzubara (commented):
I left some comments.. It looks like the pagination markup was copied from the table elements stories (https://storybook.dp3.us/?path=/story/components-tables--table-elements) but unfortunately this is a misleading example. The code in that story is just an example of how it should look, but doesn't implement it as a functional React component, so I would expect some changes are needed.

src/components/Table/Table.jsx
noelledusahel: tbh i got it from here https://storybook.dp3.us/?path=/story/components-tables--table-elements, but i'm seeing in our code Button from uswds is being used

src/components/Table/Table.module.scss
```diff
@@ -1,10 +1,12 @@
 @import 'shared/styles/_basics';
 @import 'shared/styles/_mixins';
+// @import 'shared/styles/_overrides';
```
donaldthai, line 3: nit: Let's remove this if we don't need it anymore

src/components/Table/Table.module.scss
```diff
@@ -51,4 +53,60 @@
...
-}
\ No newline at end of file
+
+.inlineBlock {
+  display: inline-block
+}
+
+.displayFlex {
+  justify-content: space-between;
+}
```
donaldthai, line 63: nit: These can be removed since it's not being used.

donaldthai (approved):
Some non-blocking comments but LGTM!

src/components/Table/Table.jsx
```diff
@@ -48,22 +69,91 @@ const Table = ({ handleClick, getTableProps, getTableBodyProps, headerGroups, ro
...
+  nextPage: undefined,
+  setPageSize: undefined,
+  gotoPage: undefined,
+  canNextPage: undefined,
+  pageCount: undefined,
+  state: undefined,
+  pageIndex: 0,
+  pageSize: 20,
+  pageOptions: [],
+  perPage: [10, 20, 50],
```
noelledusahel, line 156: @brooksr4  this is where you'd add an option for 1 per page: [1, 20, 50]

src/components/Table/Table.jsx
```diff
@@ -48,22 +69,91 @@ const Table = ({ handleClick, getTableProps, getTableBodyProps, headerGroups, ro
...
+              {perPage.map((page, index) => (
+                <option value={page} key={`page-size-${index}`}>
+                  {page}
+                </option>
+              ))}
+            </Dropdown>
+            <div>rows per page</div>
+          </div>
+          <div className={styles.tableControlPagination}>
+            <Button className={styles.usaButtonUnstyled} onClick={previousPage} disabled={!canPreviousPage}>
```
suzubara, line 92: all `<Button>`s should have the `type="button"` prop

src/components/Table/Table.module.scss
```diff
@@ -51,4 +52,56 @@
       }
     }
   }
-}
\ No newline at end of file
+
+.inlineBlock {
```
suzubara, line 56: Doesn't look like this is used anymore, can we delete it?

src/components/Table/Table.module.scss
```diff
@@ -51,4 +52,56 @@
...
+   justify-content: space-between;
+   margin: 20px;
+
+    .tableControlPagination {
+      display: flex;
+      align-items: center;
+      flex-grow: 1;
+      justify-content: flex-end;
+
+      .usaButtonUnstyled {
```
suzubara, line 72: Instead of re-implementing `usaButtonUnstyled` in CSS modules, it should be applied using the existing global CSS. Since this is the ReactUSWDS Button component, adding the `unstyled` prop will achieve this ([docs](https://trussworks.github.io/react-uswds/?path=/story/button--unstyled)):
```<Button unstyled ... />```

If you need to override or add more styling then those can be added with another CSS class and styles in this module file.

File: src/components/Table/Table.module.scss [hunk 18]
```diff
@@ -51,4 +52,56 @@
...
+        @include u-margin-x(0);
+        @include u-margin-bottom(0);
+
+        &:disabled {
+          @include u-text('base-light');
+        }
+      }
+
+
+      .usaSelect {
```
suzubara, line 87: similar to my comment above. It's fine to override CSS with new classes, but giving them the same name as existing USWDS classes is confusing, so can this be called something else? Alternatively, it can be selected using the global class like:
```
:global(.usa-select) { width: unset; }
```

suzubara (commented):
💯  this is looking really solid! I left a few minor comments

src/scenes/Office/office.scss
```diff
@@ -28,6 +28,14 @@
   margin: 0 0.3em 0 0.5em;
 }
 
+.icon-right {
```
noelledusahel, line 31: @suzubara i included these because i'd expect pagination icons' margins to mirror each other, but i realize not all icons have this style. is this fine? or better, a specific class (like pagination-icon-right' )

brooksr4 (approved):
Thanks for showing me where to add it. Looks good to me!

PR #5158, opened 2020-11-05
MB-4468 Add pagination to TOO queue handler

Ryan-Koch:
## Description

Adds pagination functionality to the TOO queue handler. Right now this just uses the pop provided Paginator interface. We may want to readdress this later as we learned that we don't necessarily need the ability to go to a specific page: https://ustcdp3.slack.com/archives/CP4979J0G/p1604594350011000?thread_ts=1604586429.001000&cid=CP4979J0G . For now though this should suffice as a first pass until we can evaluate moving to something that would perform faster with that new knowledge.

## Reviewer Notes

## Setup

Run the below and then head on over here once the office app is loaded locally: http://officelocal:3000/ghc/v1/queues/moves?page=1&perPage=1 , you should get 1 result. If you change the page to equal 2 it should show the second result.

```sh
make db_dev_e2e_populate
```

 have been satisfied.


Ryan-Koch:
Still working on this, will switch to ready to review when that's the case

pkg/handlers/ghcapi/queues.go
```diff
@@ -60,12 +66,15 @@ func (h GetMovesQueueHandler) Handle(params queues.GetMovesQueueParams) middlewa
 	queueMoves := payloads.QueueMoves(orders)
 	// ToDo - May want to move this logic into the pop query later.
 	// filter queueMoves by status
+	queueMovesBeforeStatusFilterCount := len(*queueMoves)
 	queueMoves = movesFilteredByStatus(params.Status, queueMoves)
+	queueMovesAfterStatusFilterCount := len(*queueMoves)
+	count = count - (queueMovesBeforeStatusFilterCount - queueMovesAfterStatusFilterCount)
```
monfresh, line 72: This could be because I'm tired, but I'm not following this count logic. Could you walk me through it please? In my mind, the total count represents the total amount of records for any given query, including all the filters. Whatever gets returned after applying all the filters is the total count. The total count is going to change based on which filters have been selected. I think what we had before is what we want.

pkg/services/move_order/move_order_fetcher_test.go
```diff
@@ -185,9 +187,34 @@ func (suite *MoveOrderServiceSuite) TestListMoveOrdersWithEmptyFields() {
...
 
 	suite.FatalNoError(err)
 	suite.Nil(moveOrders)
 
 }
+
+func (suite *MoveOrderServiceSuite) TestListMoveOrdersWithPagination() {
+	officeUser := testdatagen.MakeDefaultOfficeUser(suite.DB())
+
+	for i := 0; i < 22; i++ {
```
monfresh, line 200: One of the testing best practices I agree with is to use the DB as little as possible, and create the minimum amount of entries. If we have to create 22 entries to test this, that's a smell to me. What do you think about making the per page variable instead of hardcoding to 20? That way, we only need to create 2 entries for example, and in the test we can specify 1 item per page, then test that the Total Count is 2 but that only one entry is returned on the first page. This might also be better as a handler test so we can pass in the page and per page parameters and test the TotalCount in the payload.

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -14,7 +14,7 @@ type moveOrderFetcher struct {
 	db *pop.Connection
 }
 
-func (f moveOrderFetcher) ListMoveOrders(officeUserID uuid.UUID, options ...func(query *pop.Query)) ([]models.Order, error) {
+func (f moveOrderFetcher) ListMoveOrders(officeUserID uuid.UUID, page *int, options ...func(query *pop.Query)) ([]models.Order, int, error) {
```
monfresh, line 17: The intent from our work in the previous sprints was to refactor this so that it only accepts the `params` from the handler. That way, the various queries can be defined in the service, as opposed to the handler. We don't want to end up adding more and more arguments to this function. Ideally, we would do this now, but if we are tight on time, then I think we should make sure this refactor is part of next week's sprint.

Ryan-Koch: That's fair, I don't know that I want to loop that into this work with the time constraint. But happy to talk about what needs to happen for next sprint.

pkg/services/move_order/move_order_fetcher_test.go
Ryan-Koch: That's a good point, making that variable would save this pain point. I'll make some adjustments here.

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -49,14 +49,31 @@ func (f moveOrderFetcher) ListMoveOrders(officeUserID uuid.UUID, options ...func
...
+			return []models.Order{}, 0, services.NotFoundError{}
 		default:
-			return []models.Order{}, err
+			return []models.Order{}, 0, err
+		}
+	}
+
+	// If we have a page argument let's do the pagination thing.
+	if page != nil {
+		err = query.GroupBy("orders.id").Paginate(*page, 20).All(&moveOrders)
```
monfresh, line 66: What do you think about defaulting to page 1 instead of the else statement?

monfresh: The way I understand it, we always want to limit results to the first 20. If the TOO chooses the next page, then we would use that page instead of the first page.

pkg/handlers/ghcapi/queues.go
Ryan-Koch: Oh so I was trying to maintain the count that includes all records with the filters but not pagination. By the time we are filtering the statuses there we already have a set of records impacted by pagination. In the service though I did the count separately and passed it as a return. So here I was looking at the difference so I could subtract the number of records filtered out from the paginated total from the actual total count (so that it changes with the status filter too).

pkg/services/move_order/move_order_fetcher.go
Ryan-Koch: That could work. I was initially trying to make it so that someone could still use the service to pull the whole list if they wanted to. I suppose right now we don't really have anything that *needs* that though. Yeah I'll go ahead and do that.

pkg/testdatagen/scenario/e2ebasic.go
```diff
@@ -2128,4 +2128,17 @@ func (e e2eBasicScenario) Run(db *pop.Connection, userUploader *uploader.UserUpl
 			ID: uuid.FromStringOrNil("1130e612-94eb-49a7-973d-72f33685e551"), // MS - Move Management
 		},
 	})
+
+	for i := 0; i < 22; i++ {
```
monfresh, line 2132: This entire file is run before running integration tests. The more data we add to it, the longer it takes to set up the integration tests. One of the improvements I mentioned when Erin asked for things to address was separating data needed for tests from data needed for testing things locally. I think we should stop adding things to this file that are not used by integration tests. By making the per page parameter adjustable, we can easily test this locally with the existing data that gets generated.

Ryan-Koch: When I do the perPage as a variable thing I'll remove this along with that change.

pkg/handlers/ghcapi/queues.go
monfresh: Thanks, that matches what I figured out after I looked some more. Another benefit of having the new move statuses defined in the DB (that I proposed in Slack the other day) is that we wouldn't need this logic. The filter would be already applied in ListMoveOrders, and we would only need to return the count before the pagination.

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -49,14 +49,31 @@ func (f moveOrderFetcher) ListMoveOrders(officeUserID uuid.UUID, options ...func
 			option(query)
 		}
 	}
+	var moveOrdersCountObj models.Orders
+	count, err := query.GroupBy("orders.id").Count(&moveOrdersCountObj)
```
monfresh, line 53: I'm not a SQL expert. Is the `GroupBy` necessary to get the count, or even at all here? We are repeating the GroupBy down below for pagination, so I wonder if we can skip it.

pkg/services/move_order/move_order_fetcher.go
Ryan-Koch: Done

pkg/services/move_order/move_order_fetcher_test.go
Ryan-Koch: Went ahead and put the perPage handler param in, that's now being passed into the service with its own default value.

pkg/testdatagen/scenario/e2ebasic.go
Ryan-Koch: Removed.

pkg/services/move_order/move_order_fetcher.go
Ryan-Koch: Good catch, yeah I don't think we need that for the count. Removed it on this last push.

pkg/services/move_order/move_order_fetcher_test.go
```diff
@@ -185,9 +187,34 @@ func (suite *MoveOrderServiceSuite) TestListMoveOrdersWithEmptyFields() {
...
+	}
+
+	moveOrderFetcher := NewMoveOrderFetcher(suite.DB())
+	moveOrders, count, err := moveOrderFetcher.ListMoveOrders(officeUser.ID, swag.Int(1), swag.Int(1))
+
+	suite.NoError(err)
+	suite.Equal(1, len(moveOrders))
+	suite.Equal(2, count)
+
+}
```
monfresh, line 220: Could we also have a handler test in `queues_test` that makes sure the Total Count is correct when the status filter is used? Or maybe move this test to be a handler test instead?

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -49,14 +51,35 @@ func (f moveOrderFetcher) ListMoveOrders(officeUserID uuid.UUID, options ...func
...
+	}
+
+	// If we have a page argument let's do the pagination thing.
+	if page == nil {
+		page = swag.Int(1)
+	}
+
+	if perPage == nil {
+		perPage = swag.Int(20)
+	}
```
monfresh, line 73: It looks like we can get rid of this as well, since Pop defaults to 1 and 20 for page and perPage respectively. I tested this locally by printing out `query.Paginator.Page` and `query.Paginator.PerPage`. See: https://github.com/gobuffalo/pop/blob/944158f9015b0c41b2fd4f26164a9c9c74f4f090/paginator.go#L55-L60

pkg/handlers/ghcapi/queues.go
monfresh: I tested this and it's not the right formula. The problem is that `queueMoves` is only a subset of the results, based on the pagination params, so the end count could be different based on the params. For example, if we add Page: 1 and PerPage: 1 to the test labeled `loads results with all STATUSes selected`, we expect the total count to be 3 out of the 5 entries in the DB. However, we get 4 because the total number of moves is 5 and `queueMovesBeforeStatusFilterCount` is 1 because we are only showing 1 per page, and `queueMovesAfterStatusFilterCount` is 0 because the first move that was returned was a DRAFT move, which we are ignoring. So 5 - (1 - 0) = 4.

This tells us that we really need to move this status filtering into the query, not after the fact. Even if we go back to using the where clause: `Where("moves.status NOT IN ('DRAFT', 'CANCELED')")`, that doesn't solve the problem when we are only looking for one type of status. 

For example, in the test `loads results with one STATUS selected`, if we use Page: 1 and PerPage: 1, the total count from Pop will be 3, but we expect the total count in the payload to be 1 because we only asked for 1 out of the 3 moves. In this case, both the before and after queueMoves will be 1, so 3 - (1 - 1) = 3, but we want it to be 1.

So, as long as the per page is less than the total results, this calculation will be off.

Ryan-Koch: Yeah you're right. I started into a refactor early this morning. Trying to write out some subqueries to do the inference based on service items. Thinking about just biting it and making the changes needed to make the database have that 1:1 type relationship you have talked about on slack. Seems like it's turning into a similar amount of complexity to do this as it would be to change the couple of places that impact serviceItems to change the move status as a side effect.

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -49,14 +51,35 @@ func (f moveOrderFetcher) ListMoveOrders(officeUserID uuid.UUID, options ...func
...
+	count, err := query.Count(&moveOrdersCountObj)
+
+	if err != nil {
+		switch err {
+		case sql.ErrNoRows:
+			return []models.Order{}, 0, services.NotFoundError{}
+		default:
+			return []models.Order{}, 0, err
+		}
+	}
```
monfresh, line 64: Weird. I left a comment last night about this section but I'm not seeing it anymore. I found out that Pop has a Paginator struct you can call on query to get the count, so we don't need this block from lines 54-64. We can do this instead:
```golang
query = query.GroupBy("orders.id").Paginate(*page, *perPage)
err = query.All(&moveOrders)
if err != nil {
  blah blah
}
```
Reference: https://github.com/gobuffalo/pop/blob/944158f9015b0c41b2fd4f26164a9c9c74f4f090/paginator.go#L28-L41

pkg/handlers/ghcapi/queues.go
Ryan-Koch: Did this (we had paired through some of it)

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -485,7 +485,7 @@ func QueueMoves(moveOrders []models.Order) *ghcmessages.QueueMoves {
 
 		queueMoveOrders[i] = &ghcmessages.QueueMove{
 			Customer:               Customer(&customer),
-			Status:                 ghcmessages.QueueMoveStatus(queueMoveStatus(hhgMove)),
+			Status:                 ghcmessages.QueueMoveStatus(hhgMove.Status),
```
monfresh, line 488: Yay for deleted code!

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -343,8 +339,10 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandlerFilters() {
 		params := queues.GetMovesQueueParams{
 			HTTPRequest: request,
 			Status: []string{
-				modelToPayload.QueueMoveStatusNEWMOVE,
+				string(models.MoveStatusSUBMITTED),
 			},
+			Page:    swag.Int64(1),
+			PerPage: swag.Int64(1),
```
monfresh, line 345: This is looking great so far. Could we also please add a test where the total count is greater than the PerPage? I think the one below is an example where we expect the Total to be 3, but if we set the PerPage to 1, we should expect the payload length to be 1 on the first page.

Ryan-Koch: Done! ✅

pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -148,6 +149,18 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
 		return nil, verrs, services.NewQueryError("unknown", err, "")
 	}
 
+	if move.Status != models.MoveStatusAPPROVALSREQUESTED {
+		err := move.SetApprovalsRequested()
+		if err != nil {
+			return nil, nil, err
+		}
+		verrs, err := o.builder.UpdateOne(&move, nil)
+		fmt.Println(move.Status)
```
monfresh, line 158: Did you mean to leave this print statement?

pkg/services/mto_service_item/mto_service_item_creator_test.go
```diff
@@ -23,12 +23,17 @@ type testMTOServiceItemQueryBuilder struct {
...
 	fakeTransaction func(func(tx *pop.Connection) error) error
+	fakeUpdateOne   func(models interface{}, eTag *string) (*validate.Errors, error)
 }
 
 func (t *testMTOServiceItemQueryBuilder) CreateOne(model interface{}) (*validate.Errors, error) {
 	return t.fakeCreateOne(model)
 }
 
+func (t *testMTOServiceItemQueryBuilder) UpdateOne(model interface{}, eTag *string) (*validate.Errors, error) {
+	return nil, nil
```
monfresh, line 34: shouldn't this be returning `t.fakeUpdateOne(model, eTag)`?

Ryan-Koch: yes that's right, good catch. Fixed it in the last push.

pkg/services/mto_service_item/mto_service_item_creator.go
Ryan-Koch: Good eye, that one slipped through. Removed now.

pkg/services/mto_service_item/mto_service_item_updater.go
```diff
@@ -74,5 +74,57 @@ func (p *mtoServiceItemUpdater) UpdateMTOServiceItemStatus(mtoServiceItemID uuid
...
+			return nil, services.NewInvalidInputError(move.ID, err, verrs, "")
+		}
+
+		if err != nil {
+			if errors.Cause(err).Error() == models.RecordNotFoundErrorString {
+				return nil, services.NewNotFoundError(move.ID, "")
+			}
+		}
+	}
+
```
monfresh, line 128: Do we have a test for thses status changes?

Ryan-Koch: The updater service object for `mto_service_items` doesn't seem to have its own set of tests, which I am not quite apt to dive into fixing on this PR. That said we should have something, so I added in a check on the happy path integration test on the handler side that will check to verify that the move status transitioned over to `APPROVED`.

monfresh:
Can we also add the DB indexes please? I believe we need one for every column that we call in a `Where` clause. What I'm not sure about is if we need to make them all composite indexes that also include the `status` (since it's used for sorting), given that we also search by status. In other words, if we have a separate index for status, do we also need to add it to the other indexes?

Ryan-Koch:
> Can we also add the DB indexes please? I believe we need one for every column that we call in a `Where` clause. What I'm not sure about is if we need to make them all composite indexes that also include the `status` (since it's used for sorting), given that we also search by status. In other words, if we have a separate index for status, do we also need to add it to the other indexes?

Would you mind if we did that in a separate PR? I was going to write up a quick chore to knock out the indexes for both sides of the queues if that works.

monfresh:
That works. I believe I already wrote a chore, but I can't remember if it's for both TOO and TIO. Lemme find it,

monfresh:
Here it is: https://dp3.atlassian.net/browse/MB-5022

Ryan-Koch:
> Here it is: https://dp3.atlassian.net/browse/MB-5022

Awesome! Thanks for writing that up. I moved it to the top of the backlog so we can include it in the planning chat today.

monfresh (approved):
LGTM

PR #5163, opened 2020-11-05
[MB-4544] support api calls process reviewed payment request

pearl-truss:
## Description

This PR is to create a support api endpoint to trigger the processing of  reviewed payment request. 

- If the sendToSyncada flag is true, it calls the ProcessReviewedPaymentRequest service. 
- If the sendToSyncada flag is false and a paymentRequestId is provided, it will find and update the status of that particular payment request. By default the status is switched to Sent_To_Gex. (There is a status flag, if provided it will change the status to match the flag)
- If the sendToSyncada flag is false and no paymentRequestId is provided, it will find all payment request with a reviewed status and switch the status of each to match the status flag. If no status flag provided, they're switched to a Sent_To_Gex status

## Reviewer Notes
I've edited out all references to an eTag. However, it's convention to use eTags when sending patch and put requests.

## Setup

Be sure to start by running `make db_dev_e2e_populate` and then `make server_run` 

### Case 1: Send To Syncada false (will update ALL payment request in a reviewed status)

JSON
```json
{
  "body": {
    "sendToSyncada": false
  }
}
```

Command
```sh
prime-api-client --insecure support-reviewed-payment-requests --filename <filename.json>
```

Expected output
```sh
[{"eTag":"MjAyMC0xMS0wOVQxOToxNjoxMC41MTAzNDFa","id":"ba6b175c-e017-43d5-88a1-da67d14718b0","isFinal":false,"moveTaskOrderID":"dafcd344-0073-49be-90f3-918549800972","paymentRequestNumber":"9390-9705-1","status":"SENT_TO_GEX"}]
```

### Case 2: Send To Syncada false, test a particular payment request 

JSON
```json
{
  "body": {
    "paymentRequestID" : "cfd110d4-1f62-401c-a92c-39987a0b4228",
    "sendToSyncada": false
  }
}
```

Command
```sh
prime-api-client --insecure support-reviewed-payment-requests --filename <filename.json>
```

Expected output
```sh
[{"eTag":"MjAyMC0xMS0wOVQxOToxNjoxMC41MTAzNDFa","id":"ba6b175c-e017-43d5-88a1-da67d14718b0","isFinal":false,"moveTaskOrderID":"dafcd344-0073-49be-90f3-918549800972","paymentRequestNumber":"9390-9705-1","status":"SENT_TO_GEX"}]
```

### Case 3: Send To Syncada true (will update ALL payment request in a reviewed status)

JSON
```json
{
  "body": {
    "sendToSyncada": true
  }
}
```

Command
```sh
prime-api-client --insecure support-reviewed-payment-requests --filename <filename.json>
```

Expected output
```sh
[]
```

 have been satisfied.

## References


pkg/testdatagen/scenario/e2ebasic.go
```diff
@@ -1413,6 +1429,35 @@ func (e e2eBasicScenario) Run(db *pop.Connection, userUploader *uploader.UserUpl
 		MTOServiceItem: serviceItemMS,
 	})
 
+	paymentRequest9 := testdatagen.MakePaymentRequest(db, testdatagen.Assertions{
+		PaymentRequest: models.PaymentRequest{
+			ID:            uuid.FromStringOrNil("345c9ebb-972f-4711-acb2-5911f52acpp9"),
```
reggieriser, line 1434: This is a bad UUID, so it's returning `nil` and creating a new UUID each time this is re-run.

pkg/testdatagen/scenario/e2ebasic.go
```diff
@@ -1413,6 +1429,35 @@ func (e e2eBasicScenario) Run(db *pop.Connection, userUploader *uploader.UserUpl
...
+			ID:     uuid.FromStringOrNil("4fba4454-b5aa-4c29-8448-66aa07ac9657"),
+			Status: models.MTOServiceItemStatusApproved,
+		},
+		Move: move8,
+		ReService: models.ReService{
+			ID: uuid.FromStringOrNil("1130e612-94eb-49a7-973y-72f33685e453"), // MS - Move Management
+		},
+	})
+
+	testdatagen.MakePaymentServiceItem(db, testdatagen.Assertions{
```
reggieriser, line 1453: It looks like we're going to need to add some service item params here if we want the third case in your instructions to work.  Right now, I'm getting this instead:

```json
{
  "Payload": {
    "detail": "function ProcessReviewedPaymentRequest failed call to generator.Generate: id: 00000000-0000-0000-0000-000000000000 not found for ContractCode",
    "instance": "4987f1e9-7de1-4377-87e2-89fc59b2b560",
    "title": "Internal Server Error"
  }
}
```

reggieriser (commented):
Nice job on a complex endpoint!  I left some comments on some things to review.  I was able to run the first two cases, but not the third (think it's just an E2E data setup issue).

pkg/handlers/supportapi/payment_request.go
```diff
@@ -225,3 +225,116 @@ func (h GetPaymentRequestEDIHandler) Handle(params paymentrequestop.GetPaymentRe
...
+				RequestedAt:          pr.RequestedAt,
+				ReviewedAt:           &reviewedDate,
+				SentToGexAt:          &sentGexDate,
+				ReceivedByGexAt:      &recGexDate,
+				PaidAt:               &paidAtDate,
+				PaymentRequestNumber: pr.PaymentRequestNumber,
+				SequenceNumber:       pr.SequenceNumber,
+			}
+			var nilEtag string
+			updatedPaymentRequest, err := h.PaymentRequestStatusUpdater.UpdatePaymentRequestStatus(&paymentRequestForUpdate, nilEtag)
```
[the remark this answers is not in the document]
pearl-truss: UpdatePaymentRequestStatus requires an eTag, but we decided we need more information on if/how this enpoint should use etags so it doesn't have a valid eTag to pass in. So instead of tabling the whole story, it was decided to make this endpoint not require an eTag for now. https://ustcdp3.slack.com/archives/CP6F568DC/p1604697280339500?thread_ts=1604690916.327700&cid=CP6F568DC

swagger/support.yaml
```diff
@@ -1255,6 +1300,26 @@ definitions:
...
+        format: uuid
+        type: string
+        readOnly: true
+      status:
+        $ref: '#/definitions/PaymentRequestStatus'
+      sendToSyncada:
+        example: true
+        type: boolean
+        x-nullable: true
+      # eTag:
```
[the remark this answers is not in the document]
pearl-truss: I can remove it but I'd intentionally left it in since we're going to be coming back to this endpoint and adding/requiring an eTag

pkg/handlers/supportapi/payment_request.go
```diff
@@ -225,3 +225,116 @@ func (h GetPaymentRequestEDIHandler) Handle(params paymentrequestop.GetPaymentRe
...
+		case "PAID":
+			newPaymentRequestStatus = models.PaymentRequestStatusPaid
+			paidAtDate = time.Now()
+		default:
+			newPaymentRequestStatus = models.PaymentRequestStatusSentToGex
+			sentGexDate = time.Now()
+		}
+		// Update each payment request to have the given status
+		for _, pr := range paymentRequests {
+			paymentRequestForUpdate := models.PaymentRequest{
```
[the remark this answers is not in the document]
pearl-truss: I can look into it. For context, I followed the same convention as the `UpdatePaymentRequestStatusHandler` found at the top of the file https://github.com/transcom/mymove/pull/5163/files/30bd36ee0bf369a7291c197fd7ef9e28f4c3d91c#diff-f0420efb25f24d852a5557d1a735e23438f664baaccf1211d5e631f586301617L91

pkg/handlers/supportapi/payment_request.go
```diff
@@ -225,3 +225,116 @@ func (h GetPaymentRequestEDIHandler) Handle(params paymentrequestop.GetPaymentRe
...
+				ID:                   pr.ID,
+				MoveTaskOrder:        pr.MoveTaskOrder,
+				MoveTaskOrderID:      pr.MoveTaskOrderID,
+				IsFinal:              pr.IsFinal,
+				Status:               newPaymentRequestStatus,
+				RejectionReason:      pr.RejectionReason,
+				RequestedAt:          pr.RequestedAt,
+				ReviewedAt:           &reviewedDate,
```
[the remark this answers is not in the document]
pearl-truss: same as above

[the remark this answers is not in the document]
reggieriser: Hmm, I think this is still a potential issue (and maybe in the other code too).  Consider this scenario.  A payment request has been reviewed by the TIO and has a `ReviewedAt` time set.  Then this code runs and sets the `sentGexData` local variable only in the case above.  That means th elocal `reviewedDate` is still the zero time.  We'd overwrite the legit reviewed at date with the zero date I believe.

pkg/handlers/supportapi/payment_request_test.go
```diff
@@ -381,3 +381,298 @@ func (suite *HandlerSuite) TestGetPaymentRequestEDIHandler() {
 		suite.Contains(*errResponse.Payload.Detail, errStr)
 	})
 }
+
+const testDateFormat = "060102"
+
+func (suite *HandlerSuite) createPaymentRequest(num int) models.PaymentRequests {
+	var prs models.PaymentRequests
+	for i := 0; i < 4; i++ {
```
jacquelineIO, line 389: Did you mean to use the `num int` passed in to control the loop count here?

pkg/testdatagen/scenario/e2ebasic.go
```diff
@@ -1413,6 +1413,108 @@ func (e e2eBasicScenario) Run(db *pop.Connection, userUploader *uploader.UserUpl
...
+	// 	},
+	// })
+
+	// testdatagen.MakePaymentServiceItem(db, testdatagen.Assertions{
+	// 	PaymentServiceItem: models.PaymentServiceItem{
+	// 		PriceCents: &dlhCost9,
+	// 	},
+	// 	PaymentRequest: paymentRequest9,
+	// 	MTOServiceItem: serviceItemDLH9,
+	// })
```
jacquelineIO, line 1472: don't forget to remove commented out code

pkg/handlers/supportapi/api.go
```diff
@@ -24,7 +24,7 @@ import (
 // NewSupportAPIHandler returns a handler for the Prime API
 func NewSupportAPIHandler(context handlers.HandlerContext) http.Handler {
 	queryBuilder := query.NewQueryBuilder(context.DB())
-
+	var logger paymentrequest.Logger
```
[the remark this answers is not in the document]
pearl-truss: @reggieriser this is still outstanding. I've run into type issues when trying to get a value for this logger.

pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -47,7 +47,8 @@ func NewPaymentRequestReviewedProcessor(db *pop.Connection,
 func InitNewPaymentRequestReviewedProcessor(db *pop.Connection, logger Logger, sendToSyncada bool) services.PaymentRequestReviewedProcessor {
 	reviewedPaymentRequestFetcher := NewPaymentRequestReviewedFetcher(db)
 	generator := invoice.NewGHCPaymentRequestInvoiceGenerator(db)
-	sftpSession := invoice.InitNewSyncadaSFTPSession()
+	var sftpSession services.SyncadaSFTPSender
+	sftpSession = nil
```
pearl-truss, line 51: I want to call out this change. I removed the call to initialize a Syncada session because from my understanding we don't have the environment variables setup in `invoice.InitNewSyncadaSFTPSession` to actually establish a connection to syncada. Which was causing a "can't assign requested address" issue. 

@transcom/truss-a-team

gidjin: We should have variables for connecting to Syncada via sftp stored chamber. They likely won't exist in a test context though.

pkg/handlers/supportapi/payment_request.go
```diff
@@ -225,3 +226,135 @@ func (h GetPaymentRequestEDIHandler) Handle(params paymentrequestop.GetPaymentRe
...
+				} else {
+					return paymentrequestop.NewProcessReviewedPaymentRequestsBadRequest().WithPayload(payloads.ClientError(handlers.BadRequestErrMessage, "bad request, an invalid status type was used", h.GetTraceID()))
+				}
+			}
+
+			// Update payment request values to relect new status
+			pr.Status = newPaymentRequestStatus
+			pr.ReviewedAt = &reviewedDate
+			pr.SentToGexAt = &sentGexDate
+			pr.ReceivedByGexAt = &recGexDate
```
reggieriser, line 337: I did a test where I set the `requested_at` and `reviewed_at` columns in the payment request to actual dates, then ran case 1 from the setup instructions.  I noticed in the DB that those two columns stayed with their original values (good), the `sent_to_gex_at` was set to the current timestamp (good), but `received_by_gex_at` and `paid_at` changed to be the zero value timestamp rather than `null` (bad).  I've got some ideas about how we may be able to simplify the code above and get past this issue if you'd like to pair on it.

pkg/handlers/supportapi/payment_request.go
```diff
@@ -225,3 +226,135 @@ func (h GetPaymentRequestEDIHandler) Handle(params paymentrequestop.GetPaymentRe
...
+				recGexDate = time.Now()
+			case "PAID":
+				newPaymentRequestStatus = models.PaymentRequestStatusPaid
+				paidAtDate = time.Now()
+			default:
+				if len(paymentRequestStatus) == 0 {
+					newPaymentRequestStatus = models.PaymentRequestStatusSentToGex
+					sentGexDate = time.Now()
+				} else {
+					return paymentrequestop.NewProcessReviewedPaymentRequestsBadRequest().WithPayload(payloads.ClientError(handlers.BadRequestErrMessage, "bad request, an invalid status type was used", h.GetTraceID()))
```
reggieriser, line 329: Nit: This is subjective, but it might be a little more readable to use two `case` statements here: a `case ""` for the first part of the `if` and a `default` for the else part.  The `if` then goes away.

pkg/handlers/supportapi/payment_request.go
```diff
@@ -225,3 +226,135 @@ func (h GetPaymentRequestEDIHandler) Handle(params paymentrequestop.GetPaymentRe
...
+	paymentRequestStatus := params.Body.Status
+	var newPaymentRequestStatus models.PaymentRequestStatus
+	var paymentRequests models.PaymentRequests
+	var updatedPaymentRequests models.PaymentRequests
+
+	if sendToSyncada == nil {
+		return paymentrequestop.NewProcessReviewedPaymentRequestsBadRequest().WithPayload(payloads.ClientError(handlers.BadRequestErrMessage, "bad request, sendToSyncada flag required", h.GetTraceID()))
+	}
+	if *sendToSyncada {
+		err := reviewedPaymentRequestProcessor.ProcessReviewedPaymentRequest()
```
reggieriser, line 262: I noticed when running case 3 that the status changed to `SENT_TO_GEX`, but the `sent_to_gex_at` timestamp was still `null` in the DB.  Should that service be setting that?  If so, we may need a story to fix the service.

reggieriser (changes requested):
We're almost there -- blocking until we sort out the issues I noted with the DB updates.

pkg/handlers/supportapi/payment_request.go
pearl-truss: I think I found a fix for this (I verified the null timestamps after running case 1)

pearl-truss: Also, thanks for catching this!

pkg/handlers/supportapi/payment_request.go
pearl-truss: yeah that would be my understanding. I created https://dp3.atlassian.net/browse/MB-5474 to address

pkg/services/payment_request/payment_request_reviewed_processor.go
pearl-truss: @gidjin I'm just catching your comment. Yes we have: 

```
func InitNewSyncadaSFTPSession() services.SyncadaSFTPSender {
	return NewSyncadaSFTPSession(os.Getenv("SYNCADA_SFTP_PORT"), os.Getenv("SYNCADA_SFTP_USER_ID"), os.Getenv("SYNCADA_SFTP_IP_ADDRESS"), os.Getenv("SYNCADA_SFTP_PASSWORD"), os.Getenv("SYNCADA_SFTP_INBOUND_DIRECTORY"))
}
```

Is the expectation that these variables should work for the support api?

reggieriser (approved):
Thanks for making the changes!  I tried case 1 again and this time see that `null` values stay that way in the DB.  👍

pkg/services/payment_request/payment_request_reviewed_processor.go
gidjin: 🤔  yes, I would expect that those variables need to be set when the server is started otherwise the support api will not be able to contact Syncada. However, looking at it some now I'm realizing that there will likely need to be some additional work to get those variables through. First, we need to make sure our local docker setup is ready to pass those in. Second, we'll need to ensure they are setup in chamber for stg and exp and ensure the deploy passes them in properly. All of that is worth a new story. We could make this check for the presence of those vars and if present create the new session and if not return nil or an error so it can be skipped gracefully. I'm fine if that is just left to be part of the new story

pearl-truss: @gidjin thanks for the clarification. I'm not really sure how I would test for the presence of the variables. If it's okay to move that into a new story I can create a ticket for it

gidjin (approved):
Changes look good, :shipit: and create a new story for connecting syncada environment vars

pkg/services/payment_request/payment_request_reviewed_processor.go
gidjin: Yeah I'm good to put that into the new story as well. Let me know if you need feedback on the story once it's created

pearl-truss: Great, I created https://dp3.atlassian.net/browse/MB-5483

PR #5187, opened 2020-11-10
[MB-5339] Updated webhook-client to retry notifications 3 times

shimonacarvalho:
## Description

This adds 3 retries to the webhook client. On one run of the engine we will try 3 times to send, so as to hopefully mark as SENT and continue. Rather than sleep until the next run.

Please see ticket and document for full details.

### Unit Test
A number of tests were added/updated to check this, each test has a comment describing the scenario
But you can run all with `go test ./cmd/webhook-client/webhook/ -v`

### UI Test
Also you can test in the UI as follows
Run the server, office client in two different tabs
```
make server_run
make office_client_run
```

Then in the office, sign in as TIO and pick a payment request to approve. Approve it, this should put a notification in the db.

Then KILL the server (which acts as the recipient for the notifications in our devlocal)
Then run the webhook-client
`webhook-client db-webhook-notify --period 60 --insecure`

You should see the webhook client try to send 3 times, fail. It should wait a minute (period 60) and then try again. It will continue to try and fail in this state as we've not coded up the increasing severity actions.

Finally you should be able to check in the db and see the FAILING status for the notification in the webhook_notifications table, and the `first_attempted_at` date populated.

### Reference
* [Jira story](https://dp3.atlassian.net/browse/mb-5339) for this change


abbyoung (approved):
LGTM! Works as expected. Thanks for adding the subscription test data, too.

PR #5190, opened 2020-11-10
MB-5022 Add DB indexes for TXO queues

monfresh:
## Description

In the TXO queues, we are searching on various fields, some of which
already have indexes, and others that didn't. This PR adds the missing
indexes.

I verified that we already have indexes for moves.status,
moves.locator, duty_station.name, and transportation_office.gbloc.

## Reviewer Notes

Ideally, we would add a unique constraint for the EDIPI column, but unfortunately, we are violating that constraint in the `e2ebasic` script, and so the migration will fail until we fix that, which I think we should do in a separate PR.

## Setup

```sh
make db_dev_migrate
```

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5022) for this change

### DIFF
--- a/migrations/app/schema/20201110162431_add_indexes_for_txo_queues.up.sql
+++ b/migrations/app/schema/20201110162431_add_indexes_for_txo_queues.up.sql
@@ -1,0 +1,5 @@
+CREATE INDEX on service_members (affiliation);
+CREATE INDEX on service_members (last_name text_pattern_ops);
+CREATE INDEX on service_members (edipi);
+CREATE INDEX on payment_requests (created_at);
+CREATE INDEX on payment_requests (status);
--- a/migrations/app/migrations_manifest.txt
+++ b/migrations/app/migrations_manifest.txt
@@ -573,3 +573,4 @@
 20201103225532_remove_mr337_cac.up.sql
 20201103231542_add_failing_to_wh_subscriptions_status.up.sql
 20201105195546_add_index_for_webhook_notifications.up.sql
+20201110162431_add_indexes_for_txo_queues.up.sql


migrations/app/schema/20201110162431_add_indexes_for_txo_queues.up.sql
```diff
@@ -0,0 +1,5 @@
+CREATE INDEX on service_members (affiliation);
+CREATE INDEX on service_members (last_name);
```
duncan-truss, line 2: After some digging I think we need to change this to be:
```
CREATE INDEX on service_members (last_name text_pattern_ops);
```
According to https://www.postgresql.org/docs/12/indexes-opclass.html because our database locale is en_US.utf8 which you can find by running `SHOW LC_COLLATE;` the query won't use the btree index unless we specify the opclass.

monfresh: Thanks! Fixed.

duncan-truss (approved):
At some point we may want to add the direction to the index if we're adding limits for pagination https://www.postgresql.org/docs/12/indexes-ordering.html.  I don't think we know enough about how sorting is going to be used yet.  My guess would be most likely on the created_at dates.

monfresh:
I think we want to sort on the move status for TOO and payment request status for TIO.

duncan-truss:
> I think we want to sort on the move status for TOO and payment request status for TIO.

I was thinking more what the user would choose to sort on.  When we get to those sort feature stories we'll have to indicate what these defaults selected are instead of hardcoding them I imagine.

Ryan-Koch (approved):

lynzt (commented):
I think this looks good for now and we look at optimizing performance once we have more records in the system.

This looks good to me, but I haven't tested any of the indexes...

PR #5202, opened 2020-11-11
[MB-5245] Print legal agreement

Ronolibert:
## Description

Print legal agreement from the homepage

## Setup
`make server_run`
`make client_run`
1. Completely set up a move
2. Print legal agreement on homepage
Expected result: Print preview should have legalese, signature and the date you signed it

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5245) for this change

## Screenshots

image

### DIFF
--- a/src/pages/MyMove/Home/index.jsx
+++ b/src/pages/MyMove/Home/index.jsx
@@ -36,7 +36,7 @@ import {
 } from 'shared/Entities/modules/signed_certifications';
 import { selectActiveOrLatestMove } from 'shared/Entities/modules/moves';
 import { selectMTOShipmentsByMoveId, selectMTOShipmentForMTO } from 'shared/Entities/modules/mtoShipments';
-import { SHIPMENT_OPTIONS } from 'shared/constants';
+import { SHIPMENT_OPTIONS, MOVE_STATUSES } from 'shared/constants';
 import { selectActivePPMForMove } from 'shared/Entities/modules/ppms';
 import {
   selectCurrentUser,
@@ -64,6 +64,13 @@ Description.defaultProps = {
 };
 
 class Home extends Component {
+  componentDidMount() {
+    const { move, getSignedCertification } = this.props;
+    if (Object.entries(move).length && move.status === MOVE_STATUSES.SUBMITTED) {
+      getSignedCertification(move.id);
+    }
+  }
+
   componentDidUpdate(prevProps) {
     const { serviceMember, loggedInUserSuccess, isProfileComplete, move, getSignedCertification } = this.props;
     if (!prevProps.loggedInUserSuccess && loggedInUserSuccess) {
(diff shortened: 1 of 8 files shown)


mariaclaratgolfo (approved):

src/pages/MyMove/Home/index.jsx
```diff
@@ -270,122 +288,133 @@ class Home extends Component {
...
+                        step="4"
+                      >
+                        {this.hasSubmittedMove ? (
+                          <Description
+                            className={styles.moveSubmittedDescription}
+                            dataTestId="move-submitted-description"
+                          >
+                            Move submitted {formatCustomerDate(move.submitted_at)}.<br />
+                            <Button unstyled onClick={this.handlePrintLegalese} className={styles.printBtn}>
+                              Prink the legal agreement
```
kimallen, line 393: "Prink" --> "Print"

kimallen (commented):
I know the ticket says to treat this print button the same as we do on the signature page, which it does; however, I noticed that on Firefox, it does not show a preview, it just shows this below.  @mariaclaratgolfo I'm wondering if that's acceptable behavior?  It does show a preview in Edge, still trying to get IE running to check there.
image

It could also be that FIrefox allows to you adjust settings to be able to see the preview or not, I don't know.

UPDATE: I tested IE and similar to Firefox I don't see a preview 
image

kimallen (changes requested):
If I submit a move and then go to print from the home page directly, the signature and date are not populated but are expected to be:
image

However, if I log out and log back in, the values are populated.

Change requested for this and a small typo, but the code otherwise looks good and works.

src/pages/MyMove/Home/index.jsx
Ronolibert: Oops! Fixed this

Ronolibert:
> If I submit a move and then go to print from the home page directly, the signature and date are not populated but are expected to be:
> image
> 
> However, if I log out and log back in, the values are populated.
> 
> Change requested for this and a small typo, but the code otherwise looks good and works.

Good catch! The lifecycle method wasn't catching this. Added logic to fetch the certification if the move being provided to the homepage is already submitted

Ronolibert:
> I know the ticket says to treat this print button the same as we do on the signature page, which it does; however, I noticed that on Firefox, it does not show a preview, it just shows this below. @mariaclaratgolfo I'm wondering if that's acceptable behavior? It does show a preview in Edge, still trying to get IE running to check there.
> image
> 
> It could also be that FIrefox allows to you adjust settings to be able to see the preview or not, I don't know.
> 
> UPDATE: I tested IE and similar to Firefox I don't see a preview
> image

I think this is just default behavior for certain browsers but the print should still print the correct information. There are some workarounds we can explore for browsers like firefox like opening the file in a separate tab so it exposes firefox's print toolbar that does give the print preview but it's a a little more involved

kimallen (approved):
It works!  I'm approving, though also wonder if we need a client test for the `PrintableLegalese.jsx` component (which might have caught the earlier bug)

Ronolibert:
> It works! I'm approving, though also wonder if we need a client test for the `PrintableLegalese.jsx` component (which might have caught the earlier bug)

It's tricky because the PrintableLegalese doesn't always have data. It only has it in 2 situations, having submitted a move and directly after submitting a move, and to test both you need some real API data flow. I was trying to do an integration test but they don't have an API to check the print preview since its technically invisible up until you try to print. So I couldn't find any valuable tests to make

PR #5203, opened 2020-11-11
[MB-3553] New office user unauthorized on first callback from login.gov 

duncan-truss:
## Description

This PR attempts to fix the longstanding bug where when creating an office user from the admin interface would result in an unauthorized error coming back from login.gov on first sign in.  The admin made it seem like you could assign roles but this was silently failing because the underlying user record was never created until the login.gov account was created so the office user would exist with no roles.

I've changed the flow to drop the not null requirement on the `login_gov_uuid` column allowing us to create a users table record when we first create the office_users record from the admin.  This allows the roles to be saved and assigned properly.  Now on first login to login.gov we now perform an update to fill in the login_gov_uuid value instead of needing to create the whole user record.

I added the roles checkbox group to the office user creation screen instead of having to save and then go back and update the record for easier use.  Or if we wanted to use the create api in the future for adding bulk records creation it won't require multiple queries.

Finally there is a migration that looks for office_users and admin_users that have never logged in. We'll create users table records for them so if they ever do log in there will be a user record that can be updated with the login.gov id.  This migration also removes all test office and admin users that were created as a result of secure migrations.  We would need to fix these users, at least giving the office users roles because they have none, and they seem to be a lot more inflexible than the ones we can create in devlocal and in the e2e seeds.

## Reviewer Notes

I probably could have been more thorough in handling admin error cases, we haven't been good about displaying informative error messages there.

There is probably more cleanup to be done here but I was trying to keep this PR narrow and then leave some refactoring later.

One thing that occurred to me while making these changes is we don't handle if a login.gov user changes their email (or has multiple emails) at some point and we'll need to update that value in our records.

Are we allowing someone with the same email to be an admin, office, service member users?

## Setup

```sh
make db_dev_e2e_populate
make server_run
make admin_client_run
```

1. Login as an admin user (devlocal is fine) to adminlocal:3000
2. Click on the Office Users tab
3. Click the `+ Create` button to make a new office user
4. Fill in values for all of the office user fields (with an email you+authtest@truss.works) including the roles and save the record
5. Logout of the admin and go to officelocal:3000 and choose sign in with login.gov
6. Create a new login.gov account with the email used previously (you+authtest@truss.works)
7. On completion you should be brought back to the office app and redirected depending on your role to PPM/TOO/TIO queue instead of seeing an unauthorized error
8. In the database the user record should have been updated with the login_gov_uuid value as well.

These changes affect admin user creation as well.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3553) for this change

## Screenshots

Screen Shot 2020-11-17 at 12 00 06


pkg/auth/authentication/auth_test.go
```diff
@@ -866,11 +868,14 @@ func (suite *AuthSuite) TestAuthorizeUnknownUserAdminNotFound() {
 }
 
 func (suite *AuthSuite) TestAuthorizeUnknownUserAdminLogsIn() {
+	user := testdatagen.MakeDefaultUser(suite.DB())
 	// user is in admin_users but has not logged into the app before
-	adminUser := testdatagen.MakeAdminUserWithNoUser(suite.DB(), testdatagen.Assertions{
+	adminUser := testdatagen.MakeAdminUser(suite.DB(), testdatagen.Assertions{
```
monfresh, line 873: Don't we still want to test the scenario where the the admin or office user hasn't logged in yet, and doesn't have a corresponding entry in the users table?

duncan-truss: I guess that's my point in the Reviewer Notes about whether we should create a migration so we aren't maintaining that old flow.  Are you suggesting that I add code to make that scenario still work?

monfresh: Ah, yes. If we're sure that going forward, we should never be in that state, then I think we need the migration for the existing users that haven't logged in yet.

pkg/handlers/internalapi/users_test.go
```diff
@@ -38,14 +38,20 @@ func (suite *HandlerSuite) TestServiceMemberLoggedInUserRequiringAccessCodeHandl
...
 	suite.NoError(suite.DB().Save(&smRole))
 	sm := testdatagen.MakeExtendedServiceMember(suite.DB(), testdatagen.Assertions{
 		ServiceMember: models.ServiceMember{
 			FirstName:          &firstName,
 			RequiresAccessCode: true,
-		},
-		User: models.User{
-			Roles: []roles.Role{smRole},
+			UserID:             user.ID,
+			User:               user,
```
monfresh, line 54: I wonder why this test had to change.

duncan-truss: I think it had to do with how the MakeExtendedServiceMember was calling MakeUser (or wasn't) and that to get the roles to add properly I had to change the setup.

pkg/services/office_user/office_user_creator_test.go
```diff
@@ -61,7 +61,7 @@ func (suite *OfficeUserServiceSuite) TestCreateOfficeUser() {
 			fakeFetchOne: fakeFetchOne,
 		}
 
-		creator := NewOfficeUserCreator(builder)
+		creator := NewOfficeUserCreator(suite.DB(), builder)
```
monfresh, line 64: Do we want to test that the office user has a corresponding user now? And maybe also a test to make sure the transaction rollback works?

pkg/testdatagen/make_service_member.go
```diff
@@ -32,7 +32,7 @@ func MakeServiceMember(db *pop.Connection, assertions Assertions) models.Service
 		if assertions.User.LoginGovEmail == "" {
 			assertions.User.LoginGovEmail = email
 		}
-		user = MakeUser(db, assertions)
+		user = MakeDefaultUser(db)
```
monfresh, line 35: If we're no longer using the assertions, do we still need the 3 lines above this?

pkg/services/office_user/office_user_creator_test.go
duncan-truss: Good idea, I meant to focus on new tests now that I fixed all the existing ones.

pkg/testdatagen/make_user.go
```diff
@@ -11,7 +11,6 @@ import (
 func MakeUser(db *pop.Connection, assertions Assertions) models.User {
 
 	user := models.User{
-		LoginGovUUID:  uuid.Must(uuid.NewV4()),
```
monfresh, line 14: It looks like removing this is requiring us to define this field in `e2ebasic`. Would keeping this here allow us to not have to change `e2ebasic`?

duncan-truss: Yes always setting that would reduce all the e2e changes.  My reasoning was that we'd catch more issues if we didn't have that default always set, which it isn't for office users and admins upon creation.  Maybe for service member users it makes sense to always have that field set on creation.

monfresh:
I noticed that when we create a new admin user via devlocal, it automatically gets a `login_gov_uuid`, and I verified that when I signed in with that admin user via login.gov, my `login_gov_uuid` was updated to the one from login.gov. Is there a reason why we wouldn't do the same for office users? That would be better than removing the NOT NULL constraint.

Also, not related to this PR, but I noticed we're not enforcing uniqueness of the user email. I wonder why that is. Seems like we should.

duncan-truss:
> I noticed that when we create a new admin user via devlocal, it automatically gets a `login_gov_uuid`, and I verified that when I signed in with that admin user via login.gov, my `login_gov_uuid` was updated to the one from login.gov. Is there a reason why we wouldn't do the same for office users? That would be better than removing the NOT NULL constraint.
> 
> Also, not related to this PR, but I noticed we're not enforcing uniqueness of the user email. I wonder why that is. Seems like we should.

Right, a devlocal user is supposed to be one that never logs in from login.gov though.  A new admin user doesn't get a login_gov_uuid though if you create it through the admin ui just like an office user.  I don't think we ever update a login_gov_uuid except on initial account creation when `authorizeUnknownUser` gets called when the openID doesn't have a database match but does have a matching email.  After that it will match the openID with the login_gov_uuid and call `authorizeKnownUser` instead.

monfresh:
Right. What I'm saying asking is why can't we create a random `login_gov_uuid` as part of creating the user for the admin user or office user, and then when they sign in for the first time, it will automatically update it to the correct value from login.gov.

duncan-truss:
> Right. What I'm asking is why can't we create a random `login_gov_uuid` as part of creating the user for the admin user or office user, and then when they sign in for the first time, it will automatically update it to the correct value from login.gov.

@monfresh Gotcha, I think that would fundamentally alter the core logic which uses the login_gov_uuid search to know if the user exists in the system.  If that value is populated with a dummy value we'll have no way of knowing if they've ever logged into the app.  The FetchUserIdentity function in the user model is doing most of the work here.  We would have to then switch to matching the email and then checking to see if the login_gov_uuid differed on every sign in instead of just the first one.  I mentioned this in the reviewer notes too that in the future you should be able to change your login.gov email and that uuid will be stable so we can discover that change.  We don't want to start relying heavily on that email.

duncan-truss:
I realized I have to fix this for Service Member account creation now 🤦

pkg/handlers/internalapi/users_test.go
Ryan-Koch: Ah yeah, `MakeExtendedServiceMember` does call `MakeServiceMember` which then calls `MakeUser` like so:

```
	// ID is required because it must be populated for Eager saving to work.
	if isZeroUUID(assertions.ServiceMember.UserID) {
		if assertions.User.LoginGovEmail == "" {
			assertions.User.LoginGovEmail = email
		}
		user = MakeUser(db, assertions)
	}
	if assertions.User.LoginGovEmail != "" {
		email = assertions.User.LoginGovEmail
	}
```
I think by having the non-nil userID there in your call you manage to avoid that `MakeUser` call. Which is likely what you want seeing that you're running it above there.

monfresh:
@duncan-truss Here is the way I understand the flow:

1. New office user is created in the admin interface, along with a corresponding user in the users table
1. office user signs in via login.gov
2. We check the users table for an entry that matches the UUID provided by login.gov
3. If we don't find that UUID, then that means this is the first time they are signing in via login.gov, and so we call `authorizeUnknownUser`
4. We look up the office user by the email in the session (which comes from login.gov), and get the corresponding user
5. We update the `login_gov_uuid` field in the users table to match the one provided by login.gov
6. Going forward, they will always have the same login.gov UUID, even if they change their email, so from now on, they are known users

What about this current flow prevents us from setting a dummy UUID in step 1, since it will get updated with the correct UUID in step 6?

duncan-truss:
> @duncan-truss Here is the way I understand the flow:
> 
>     1. New office user is created in the admin interface, along with a corresponding user in the users table
> 
>     2. office user signs in via login.gov
> 
>     3. We check the users table for an entry that matches the UUID provided by login.gov
> 
>     4. If we don't find that UUID, then that means this is the first time they are signing in via login.gov, and so we call `authorizeUnknownUser`
> 
>     5. We look up the office user by the email in the session (which comes from login.gov), and get the corresponding user
> 
>     6. We update the `login_gov_uuid` field in the users table to match the one provided by login.gov
> 
>     7. Going forward, they will always have the same login.gov UUID, even if they change their email, so from now on, they are known users
> 
> 
> What about this current flow prevents us from setting a dummy UUID in step 1, since it will get updated with the correct UUID in step 6?

@monfresh What you outline matches my understanding.  I was wrong to say that it would require changing to use emails.  I stand by my point that having that value be null allows us to know who has never signed in and filling it in obscures it's purpose.  If you just want to make the test factories simpler I think we can do that.

pkg/models/user.go
```diff
@@ -96,6 +95,26 @@ func CreateUser(db *pop.Connection, loginGovID string, email string) (*User, err
 	return &newUser, nil
 }
 
+// UpdateUser is called upon the first successful login.gov verification of a new user
+func UpdateUser(db *pop.Connection, user *User, loginGovID string) error {
```
monfresh, line 99: What do you think about renaming this to `UpdateUserLoginGovUUID`?

monfresh:
In what context do we need to know if a user has every signed in? Is it for informational or auditing purposes? As far as I can tell, it doesn't affect the authentication flow. If a field is meant to be required, it should have a DB constraint. That's my main issue. I don't think we should lose that safety if there is a way around that. If we need to be able to find out if a user has ever signed in, there are other ways to do that, such as adding a `last_signed_in_at` field. This is actually a STIG requirement that says users need to be able to see for themselves when they last signed in, but because the login.gov profile provides that information, we think that might be enough to satisfy that requirement.

duncan-truss:
> In what context do we need to know if a user has ever signed in? Is it for informational or auditing purposes? As far as I can tell, it doesn't affect the authentication flow. If a field is meant to be required, it should have a DB constraint. That's my main issue. I don't think we should lose that safety if there is a way around that. If we need to be able to find out if a user has ever signed in, there are other ways to do that, such as adding a `last_signed_in_at` field. This is actually a STIG requirement that says users need to be able to see for themselves when they last signed in, but because the login.gov profile provides that information, we think that might be enough to satisfy that requirement.

@monfresh For something that gets audited like user records, I don't think having garbage data in that uuid field is the right route and isn't inherently more safe than a null.  We had it not null before and this is the state we got to of silent errors. I think this reflects the business logic more explicitly whereas having that field with a nonexistent value will lead to confusion.  I am all for adding first class tracking of last signed in and other features.

monfresh:
How about instead of leaving it null, we populate it with a zero UUID? Would that be a good compromise? That would preserve the data integrity knowing that we would never end up with an empty UUID in the database, and at the same time, it makes it easy to differentiate between real UUIDs and placeholders ones for users that were newly created via the admin interface.

duncan-truss:
> How about instead of a random UUID, we populate it with a zero UUID? Would that be a good compromise? That would preserve the data integrity knowing that we would never end up with an empty UUID in the database, and at the same time, it makes it easy to differentiate between real UUIDs and placeholder ones for users that were newly created via the admin interface.

That sounds like we're reinventing NULL but now can't add a uniqueness constraint if we want.

monfresh:
Oh yeah! I didn't think about that! I think it helped to talk it through in Slack, so we should be fine with your approach.

migrations/app/schema/20201112222339_create_user_records_for_office_and_admin_users_without.up.sql
```diff
@@ -0,0 +1,50 @@
...
+WITH admin_associate_user AS (
+    SELECT users.*
+    FROM admin_users
+    JOIN users ON admin_users.email = users.login_gov_email
+    WHERE admin_users.user_id IS NULL
+)
+UPDATE admin_users
+SET user_id = admin_associate_user.id
+FROM admin_associate_user
+WHERE admin_users.email = admin_associate_user.login_gov_email;
```
monfresh, line 50: I tested this locally for office users and it seems to work fine, but I noticed that the `active` column is set to `false`. Should it be set to `true`?

monfresh: On the users table that is.

duncan-truss: Yes that should be true thank you 🙌

duncan-truss: Although the active column exists on both the `admin_users` and `office_users` tables as well.

pkg/services/admin_user/admin_user_creator_test.go
```diff
@@ -48,6 +63,44 @@ func (suite *AdminUserServiceSuite) TestCreateAdminUser() {
 		suite.Equal(userInfo.Email, adminUser.User.LoginGovEmail)
 	})
 
+	// Reuses existing user if it's already been created for an office or service member
+	suite.T().Run("Finds existing user by email and associates with office user", func(t *testing.T) {
```
monfresh, line 67: Minor nit: should that say "associates with admin user" instead of office user?

pkg/testdatagen/make_service_member.go
monfresh: @duncan-truss It looks like this is still an open question. Could you double check please?

monfresh (approved):
LGTM % comments

Ryan-Koch (approved):

PR #5207, opened 2020-11-12
[MB-3605] Fixes UpdateMTOPostCounselingInfoHandler payload response

chrisrcoles:
## Description

This PR introduces a fix to ensure that the API response for `UpdateMTOPostCounselingInfoHandler` includes `moveOrder.customer` and `moveOrder.destinationDutyStation`. Prior to this work, there were not values being returned for these attributes. 

## Reviewer Notes

In-line comments

## Setup

```
# Ensure tests are passing successfully
mylaptop ~ %: cd /mymove
mylaptop ~ %: make server_run
mylaptop ~ %: go test ./pkg/handlers/ghcapi --testify.m "TestMoveTaskOrderUpdater_UpdatePostCounselingInfo" -v 

# Ensure payload includes `moveOrder.customer` and `moveOrder.destinationDutyStation`
# Note: Substitute `moveTaskOrderID`and `etag`
curl --location --request PATCH 'https://primelocal:9443/prime/v1/move-task-orders/{moveTaskOrderID}/post-counseling-info' \
--header 'If-Match: {eTag}' \
--header 'Content-Type: application/json' \
--data-raw '{
    "ppmEstimatedWeight": 1500,
    "ppmType": "FULL"
}'
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3605) 

## Screenshots
Screen Shot 2020-11-11 at 9 18 27 PM

### DIFF
--- a/pkg/services/move_task_order/move_task_order_updater.go
+++ b/pkg/services/move_task_order/move_task_order_updater.go
@@ -116,6 +116,14 @@ func (o *moveTaskOrderUpdater) UpdatePostCounselingInfo(moveTaskOrderID uuid.UUI
 		return nil, services.NewNotFoundError(moveTaskOrderID, "while looking for moveTaskOrder.")
 	}
 
+	err = o.db.Q().Eager(
+		"Orders.NewDutyStation.Address",
+		"Orders.ServiceMember").Find(&moveTaskOrder, moveTaskOrderID)
+
+	if err != nil {
+		return nil, services.NewNotFoundError(moveTaskOrderID, "while looking for moveTaskOrder.")
+	}
+
 	estimatedWeight := unit.Pound(body.PpmEstimatedWeight)
 	moveTaskOrder.PPMType = &body.PpmType
 	moveTaskOrder.PPMEstimatedWeight = &estimatedWeight
--- a/pkg/services/move_task_order/move_task_order_updater.go
+++ b/pkg/services/move_task_order/move_task_order_updater.go
@@ -107,18 +107,12 @@ type UpdateMoveTaskOrderQueryBuilder interface {
 func (o *moveTaskOrderUpdater) UpdatePostCounselingInfo(moveTaskOrderID uuid.UUID, body movetaskorderops.UpdateMTOPostCounselingInformationBody, eTag string) (*models.Move, error) {
 	var moveTaskOrder models.Move
 
-	queryFilters := []services.QueryFilter{
-		query.NewQueryFilter("id", "=", moveTaskOrderID),
-	}
-	err := o.builder.FetchOne(&moveTaskOrder, queryFilters)
-
-	if err != nil {
-		return nil, services.NewNotFoundError(moveTaskOrderID, "while looking for moveTaskOrder.")
-	}
-
-	err = o.db.Q().Eager(
+	err := o.db.Q().Eager(
 		"Orders.NewDutyStation.Address",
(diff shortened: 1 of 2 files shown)


pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -116,6 +116,14 @@ func (o *moveTaskOrderUpdater) UpdatePostCounselingInfo(moveTaskOrderID uuid.UUI
 		return nil, services.NewNotFoundError(moveTaskOrderID, "while looking for moveTaskOrder.")
 	}
 
+	err = o.db.Q().Eager(
```
chrisrcoles, line 119: Previously, we were [only using `builder.FetchOne` to return only the `moveTaskOrder`](https://github.com/transcom/mymove/blob/380b1e558432730bcccdc8e8de3a68e1b9336dc8/pkg/services/move_task_order/move_task_order_updater.go#L113). This implementation excludes returning any associated data. In this case, `Orders.NewDutyStation.Address` and `Orders.ServiceMember` were not being queried from the database and thus not populating the payload correctly. Given that 1) since the new Pop 5 update now we are able to selectively query only the related data we want and 2) neither [`FetchOne`](https://github.com/transcom/mymove/blob/380b1e558432730bcccdc8e8de3a68e1b9336dc8/pkg/services/move_task_order/move_task_order_updater.go#L113) nor [`UpdateOne`](https://github.com/transcom/mymove/blob/master/pkg/services/query/query_builder.go#L360) return associated data, the [Eager() implementation](https://godoc.org/github.com/gobuffalo/pop#Connection.Eager) made the most sense. Because this happens before the update, we can be assured that we are able to query all the necessary data before updating the record.

pkg/services/move_task_order/move_task_order_updater_test.go
```diff
@@ -0,0 +1,60 @@
...
+		suite.NotZero(actualMTO.Orders)
+		suite.NotNil(expectedMTO.ReferenceID)
+		suite.NotNil(expectedMTO.Locator)
+		suite.Nil(expectedMTO.AvailableToPrimeAt)
+		suite.NotEqual(expectedMTO.Status, models.MoveStatusCANCELED)
+
+		suite.NotNil(expectedMTO.Orders.ServiceMember.FirstName)
+		suite.NotNil(expectedMTO.Orders.ServiceMember.LastName)
+		suite.NotNil(expectedMTO.Orders.NewDutyStation.Address.City)
+		suite.NotNil(expectedMTO.Orders.NewDutyStation.Address.State)
```
chrisrcoles, line 50: these checks ensure that the previously nil fields have been fixed

pkg/services/move_task_order/move_task_order_updater.go
shimonacarvalho: Hey Chris, tested and it works yay. I agree I think selective Eager is a good choice here. 

It looks to me that you are overwriting moveTaskOrder with this query, is there any reason why you are keeping the other call to the query builder FetchOne on line 113? Seems like we could rip out the entire Fetcher as we are not using it any more? Unless I'm missing some other point in the function where we are?

shimonacarvalho (changes requested):

pkg/services/move_task_order/move_task_order_updater.go
chrisrcoles: Hey Shimona - good catch. I've removed it.

abbyoung (approved):
lgtm!

PR #5208, opened 2020-11-12
[MB-5340] Update subscriptions based on notifications status

shimonacarvalho:
## Description

This ticket adds on to the previous mb-5339, by updating the subscription when a notification is failing. It also adds a test which covers mb-5341. 
The expected behaviour is that if the notification is failing the sub is set to `FAILING`. 
When the notification is sent, the sub is set to `ACTIVE`
Please see ticket and document for full details.

### Unit Test
A number of tests were updated to check this, each test has a comment describing the scenario
The `TestEngineRunFailingSub` specifically tests this case.
But you can run all with `go test ./cmd/webhook-client/webhook/ -v`

### UI Test

**Setup**
Also you can test in the UI as follows
Run the server, office client in two different tabs
```
make server_run
make db_dev_e2e_populate
```
and
```
make office_client_run
```
Check the webhook_subscriptions table in the db to find the sub.

**Add notification**
Then in the office site, sign in as TIO and pick a payment request to approve. 
Approve it, this should put a notification in the db.
Check the webhook_notifications table in the db to find the notification.
If it does not, it's possible that payment request is associated with an mto not available to prime. In that case, pick a different payment request.

**Make the webhook client fail**
Then KILL the server (which acts as the recipient for the notifications in our devlocal)
And run the webhook-client
`webhook-client db-webhook-notify --period 60 --insecure`

**Check results**
You should see the webhook client try to send 3 times, fail. It should wait a minute and then try again. It will continue to try and fail in this state as we've not coded up the increasing severity actions.

You should be able to check in the db and see the FAILING status for the subscription in the webhook_subscriptions table.

**Make the webhook client succeed**
Run the server again, so the client can send to it `make server_run`
Watch to see the webhook client succeed.

**Check results**
You should be able to check in the db and see the ACTIVE status for the subscription in the webhook_subscriptions table.

### Reference
* [Jira story](https://dp3.atlassian.net/browse/mb-5340) for this change
* [Jira story](https://dp3.atlassian.net/browse/mb-5341) for the test

### DIFF
--- a/cmd/webhook-client/webhook/webhook.go
+++ b/cmd/webhook-client/webhook/webhook.go
@@ -44,11 +44,16 @@ func (eng *Engine) processNotifications(notifications []models.WebhookNotificati
 		// search for subscription
 		foundSub := false
 		for _, sub := range subscriptions {
+			sub := sub
 			if sub.EventKey == notif.EventKey {
 				foundSub = true
 				// If found, send  to subscription
 				// #nosec G601 TODO needs review
 				err := eng.sendOneNotification(&notif, &sub)
+				errDB := eng.updateSubscriptionStatus(&notif, &sub)
+				if errDB != nil {
+					eng.Logger.Error("Webhook Subscription update failed", zap.Error(err))
+				}
 				if err != nil {
 					eng.Logger.Error("Webhook Notification send failed", zap.Error(err))
 					return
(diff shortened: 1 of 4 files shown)


cmd/webhook-client/webhook/webhook.go
```diff
@@ -69,6 +74,41 @@ func (eng *Engine) processNotifications(notifications []models.WebhookNotificati
 	}
 }
 
+// updateSubscriptionStatus updates the subscription based on the status of the last notification.
+// Returns nil if nothing to update or update succeeds, returns error if error found
+func (eng *Engine) updateSubscriptionStatus(notif *models.WebhookNotification, sub *models.WebhookSubscription) error {
```
abbyoung, line 79: Testing locally, I'm seeing the expected behavior for the notification but my subscription is never marked 'FAILING'. Investigating now 🔍

abbyoung: JK I forgot to rebuild the client 🙄 works great! 😄

cmd/webhook-client/webhook/webhook_test.go
```diff
@@ -382,6 +383,91 @@ func (suite *WebhookClientTestingSuite) Test_EngineRunInactiveSub() {
...
+	suite.Equal(models.WebhookNotificationFailing, updatedNotifs[1].Status)
+	suite.False(updatedNotifs[1].FirstAttemptedAt.IsZero())
+	// Subscription should be set to FAILING
+	suite.DB().Find(&subscriptions[1], subscriptions[1].ID)
+	suite.Equal(models.WebhookSubscriptionStatusFailing, subscriptions[1].Status)
+
+	// Third notification should be PENDING
+	suite.Equal(models.WebhookNotificationPending, updatedNotifs[2].Status)
+	suite.True(updatedNotifs[2].FirstAttemptedAt.IsZero())
+
```
abbyoung, line 470: Thank you for all the comments in the test!

abbyoung (approved):
lgtm! so exciting to see this come together

PR #5220, opened 2020-11-13
Refactor devseed for easier batch creation

monfresh:
## Description

By creating functions for each type of entry, it makes the file more
readable, and makes it a lot easier to create many instances of a
particular entry. For example, to test pagination in the TXO queues,
you can now easily loop over one of the functions.

I removed some entries that didn't seem to be used for local testing.

I also noticed that in some cases, we are using the same email for a
bunch of users, which doesn't reflect reality. Similarly, we are using
the default ARMY affiliation for all service members. Ideally, we would 
make email unique and choose a random affiliation, which will make it
easier to test sorting and filtering in the TXO queues. I will do that 
in a separate PR.

## Reviewer Notes

Run `make db_dev_e2e_populate`. It should succeed without failures.
Sign in as a TOO. You should see 15 moves.
Verify that the data you expect to see locally is there.
If there are entries we no longer need for local testing, please flag them.

 have been satisfied.

### DIFF
--- a/pkg/testdatagen/scenario/devseed.go
+++ b/pkg/testdatagen/scenario/devseed.go
@@ -27,8 +27,11 @@ type devSeedScenario NamedScenario
 // DevSeedScenario Is the thing
 var DevSeedScenario = devSeedScenario{"dev_seed"}
 
-// Run does that data load thing
-func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUploader, primeUploader *uploader.PrimeUploader, logger Logger, storer *storage.Filesystem) {
+var estimatedWeight = unit.Pound(1400)
+var actualWeight = unit.Pound(2000)
+var hhgMoveType = models.SelectedMoveTypeHHG
+
+func createPPMOfficeUser(db *pop.Connection) {
 	/*
 	 * Basic user with office access
 	 */
@@ -38,7 +41,7 @@ func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUplo
 		log.Panic(fmt.Errorf("Failed to find RoleTypePPMOfficeUsers in the DB: %w", err))
 	}
 
-	email := "officeuser1@example.com"
+	email := "ppm_role@office.mil"
 	userID := uuid.Must(uuid.FromString("9bfa91d2-7a0c-4de0-ae02-b8cf8b4b858b"))
 	testdatagen.MakeOfficeUser(db, testdatagen.Assertions{
 		User: models.User{
@@ -53,35 +56,14 @@ func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUplo
 			Active: true,
 		},
 	})
+}
 
-	/*
-	 * Service member with no uploaded orders
-	 */
-	email = "needs@orde.rs"
-	uuidStr := "feac0e92-66ec-4cab-ad29-538129bf918e"
-	testdatagen.MakeUser(db, testdatagen.Assertions{
-		User: models.User{
-			ID:            uuid.Must(uuid.FromString(uuidStr)),
-			LoginGovEmail: email,
-			Active:        true,
-		},
-	})
-
-	testdatagen.MakeExtendedServiceMember(db, testdatagen.Assertions{
-		ServiceMember: models.ServiceMember{
-			ID:            uuid.FromStringOrNil("c52a9f13-ccc7-4c1b-b5ef-e1132a4f4db9"),
-			UserID:        uuid.FromStringOrNil(uuidStr),
-			FirstName:     models.StringPointer("NEEDS"),
-			LastName:      models.StringPointer("ORDERS"),
-			PersonalEmail: models.StringPointer(email),
-		},
-	})
-
+func createPPMWithAdvance(db *pop.Connection, userUploader *uploader.UserUploader) {
 	/*
 	 * Service member with uploaded orders and a new ppm
 	 */
-	email = "ppm@incomple.te"
-	uuidStr = "e10d5964-c070-49cb-9bd1-eaf9f7348eb6"
+	email := "ppm@incomple.te"
+	uuidStr := "e10d5964-c070-49cb-9bd1-eaf9f7348eb6"
 	testdatagen.MakeUser(db, testdatagen.Assertions{
 		User: models.User{
 			ID:            uuid.Must(uuid.FromString(uuidStr)),
@@ -113,12 +95,14 @@ func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUplo
 	})
 	ppm0.Move.Submit(time.Now())
 	models.SaveMoveDependencies(db, &ppm0.Move)
+}
 
+func createPPMWithNoAdvance(db *pop.Connection, userUploader *uploader.UserUploader) {
 	/*
 	 * Service member with uploaded orders, a new ppm and no advance
 	 */
-	email = "ppm@advance.no"
-	uuidStr = "f0ddc118-3f7e-476b-b8be-0f964a5feee2"
+	email := "ppm@advance.no"
+	uuidStr := "f0ddc118-3f7e-476b-b8be-0f964a5feee2"
 	testdatagen.MakeUser(db, testdatagen.Assertions{
 		User: models.User{
 			ID:            uuid.Must(uuid.FromString(uuidStr)),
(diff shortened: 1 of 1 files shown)


pkg/testdatagen/scenario/devseed.go
```diff
@@ -633,608 +279,248 @@ func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUplo
...
 	})
 
-	email = "profile2@complete.draft"
-	uuidStr = "3b9360a3-3304-4c60-90f4-83d687884077"
+	move.PersonallyProcuredMoves = models.PersonallyProcuredMoves{ppm}
+	move.Submit(time.Now())
+	models.SaveMoveDependencies(db, &move)
+}
+
+// func createUnsubmittedHHGMove(db *pop.Connection) {
```
kimallen, line 315: Please don't remove unsubmited moves.  I use them all the time in testing onboarding.  Not sure if I'm understanding what the changes here are, but please keep hhg@only.unsubmitted and nts@ntsr.unsubmitted

pkg/testdatagen/scenario/devseed.go
```diff
@@ -2202,11 +1531,43 @@ func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUplo
...
+	// PPM Office Queue
+	createPPMOfficeUser(db)
+	createPPMWithAdvance(db, userUploader)
+	createPPMWithNoAdvance(db, userUploader)
+	createPPMWithPaymentRequest(db, userUploader)
+	createCanceledPPM(db, userUploader)
+	createPPMReadyToRequestPayment(db, userUploader)
+
+	// Unsubmitted moves - I don't think we need these.
+	// createUnsubmittedHHGMove(db)
```
kimallen, line 1547: We need unsubmitted moves for onboarding

pkg/testdatagen/scenario/devseed.go
```diff
@@ -2202,11 +1531,43 @@ func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUplo
...
+	// Should this one below be submitted so we can test NTS and NTS-R shipments in the TOO queue?
+	// createUnsubmittedMoveWithNTSAndNTSR(db)
```
kimallen, line 1549: This one was created for onboarding testing- if you need ones in submitted state, please create a separated one

pkg/testdatagen/scenario/devseed.go
```diff
@@ -508,75 +216,14 @@ func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUplo
...
-			Locator: "BLABLA",
-			Status:  models.MoveStatusSUBMITTED,
-		},
-		UserUploader: userUploader,
-	})
-
-	/*
-	 * A service member with orders and a move, but no move type selected
-	 */
-	email = "sm_no_move_type@example.com"
```
kimallen, line 549: Can we please keep this as well?  We often need a record that has a customer that has a completed profile and orders, but has not created any moves

monfresh: Yes! This is exactly the feedback I was looking for. Thanks for confirming that you need these.

pkg/testdatagen/scenario/devseed.go
monfresh: Will do. Thanks for confirming.

monfresh:
@kim I added them back in. Please take another look. Thanks!

pkg/testdatagen/scenario/devseed.go
```diff
@@ -53,35 +56,14 @@ func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUplo
 			Active: true,
 		},
 	})
+}
 
-	/*
-	 * Service member with no uploaded orders
-	 */
-	email = "needs@orde.rs"
```
kimallen, line 60: I'd like to keep needs@ord.rs as well.  This allows us to start with a service member that has already gone through the profile step, but not yet the orders step.

kimallen:
Thanks for keeping the ones I requested.  Regarding your note about randomly populating the service member email, I'd like to NOT do that randomly.  When we log in from Local Sign In, the email is how we determine which user we want to log in as, which tells us at what stage they are in the flow.

monfresh:
Yes, exactly. Sorry for not being clearer. The problem I noticed is that we are not enforcing uniqueness of the user email, which I'm pretty sure we should. When I said random, I meant the ability to generate unique emails for each user to reflect reality.

All of the moves from the TXO queues also create a Service Member, but with the same name and email, which is what you're seeing.

pkg/testdatagen/scenario/devseed.go
monfresh: Done!

kimallen: thanks, though I don't see the needs@ord.rs user in the http://milmovelocal:3000/devlocal-auth/login page.

monfresh: I see it on my computer. Maybe you need to pull this branch down again or maybe run `make clean build`?

pkg/testdatagen/scenario/devseed.go
```diff
@@ -146,152 +130,14 @@ func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUplo
 	})
 	ppmNoAdvance.Move.Submit(time.Now())
 	models.SaveMoveDependencies(db, &ppmNoAdvance.Move)
```
duncan-truss, line 132: This may get fixed in the static analysis work but we have a bad habit of not checking error return values in these scenarios.  I've seen it silently obscure some mistakes.

pkg/testdatagen/scenario/devseed.go
```diff
@@ -1871,9 +1221,23 @@ func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUplo
...
+}
 
+func createTXO(db *pop.Connection) {
 	/* A user with both too and tio roles */
-	email = "too_tio_role@office.mil"
+	email := "too_tio_role@office.mil"
+	tooRole := roles.Role{}
+	err := db.Where("role_type = $1", roles.RoleTypeTOO).First(&tooRole)
+	if err != nil {
+		log.Panic(fmt.Errorf("Failed to find RoleTypeTOO in the DB: %w", err))
```
duncan-truss, line 1232: In some places we are using the builtin "log" and others we are using the zap Logger, do you think we should make it consistent?

pkg/testdatagen/scenario/devseed.go
```diff
@@ -2202,16 +1587,44 @@ func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUplo
...
-			CallbackURL: "https://primelocal:9443/support/v1/webhook-notify",
-		},
-	})
-	// Create multiple payment requests to fill queue
-	testdatagen.MakeMultiPaymentRequestWithItems(db, testdatagen.Assertions{
-		Move:          move8,
-		PrimeUploader: primeUploader,
-	}, 35)
+// Run does that data load thing
+func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUploader, primeUploader *uploader.PrimeUploader, logger Logger, storer *storage.Filesystem) {
```
duncan-truss, line 1593: It doesn't look like devseed or e2e basic uses the storer argument should we remove it from the arguments list?

pkg/testdatagen/scenario/devseed.go
```diff
@@ -2202,16 +1587,44 @@ func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUplo
...
+
+	// TXO Queues
+	createTOO(db)
+	createTIO(db)
+	createTXO(db)
+
+	// This allows testing the pagination feature in the TXO queues.
+	// Feel free to comment out the loop if you don't need this many moves.
+	for i := 1; i < 12; i++ {
+		createHHGMoveWithPaymentRequest(db, userUploader, primeUploader, logger)
```
duncan-truss, line 1616: I would say it's out of scope for this PR but maybe the scenario struct could hold the db, uploader, and logger instances then these methods could be instances of the scenario:
```
func(e devSeedScenario) createHHGMoveWithPaymentRequest() {
    testdatagen(e.db, testdatagen.Asserstions{}
    ...
        Uploader: e.userUploader,
    ...
}
```

pkg/testdatagen/scenario/devseed.go
kimallen: Dope!  I see it now.  I won't even tell you why 🤦

kimallen (approved):
LGTM!  Thanks for the adjustments!

pkg/testdatagen/scenario/devseed.go
monfresh: Good point. I will make a note of it for a follow-up PR.

pkg/testdatagen/scenario/devseed.go
monfresh: That sounds like a good idea. I made a note of it.

pkg/testdatagen/scenario/devseed.go
monfresh: Yeah, I noticed that as well. I'll remove it in a future PR and see if anything breaks.

pkg/testdatagen/scenario/devseed.go
monfresh: Agreed!

pkg/testdatagen/scenario/devseed.go
monfresh: From what I can tell, if there are any errors or validation errors, they will be caught in `testdatagen/shared.go` when attempting to create or save the models, before `models.SaveMoveDependencies(db, &ppmNoAdvance.Move)` gets called. I don't see anything that would cause that line to fail.

More importantly, it's not clear to me why we're manually changing states in all of these PPM functions, when we can set the final state that we want directly in the `Move` assertion. As far as I can tell, it's the same end result.

PR #5225, opened 2020-11-13
Kim mb 5485 insecure random number source

kimallen:
## Description

gosec linter revealed security vulnerabilities regarding G404, Insecure random number source (rand).  This PR replaces math/rand usage with crypto/rand, a cryptographically secured generator.  It also instantiates seeding in a couple places with crypto/rand, while utilizing the math/rand tools.

## Reviewer Notes

disabling of #nosec for the 404 finding has been removed in all cases but two, which contain annotation as to the mitigation/justification for disabling.

Many of the usages of math/rand were in places that were inconsequential, such as test data generation.  I still mitigated them to avoid engineers looking for randomization examples and using them in places where it would be a security risk.

Rather than change the signature for all functions where crypto/rand's Int() returns an error, I used a logger Panic.  Is that acceptable?

## Setup

pre-commit run -v --all-files golangci-lint

## References

* [Static anaylysis story](https://dp3.atlassian.net/browse/MB-5485) for this change
* [Remediation subtask](https://dp3.atlassian.net/browse/MB-5487) for this change

* [this article](https://yourbasic.org/golang/crypto-rand-int/) and [this article](https://blog.gopheracademy.com/advent-2017/a-tale-of-two-rands/)explain more about the approach used for the seed.
* [Previously used method now used in combo](https://www.calhoun.io/creating-random-strings-in-go/)
*[Fortify documentation for this finding](https://vulncat.fortify.com/en/detail?id=desc.semantic.abap.insecure_randomness#Golang)


cmd/generate-shipment-summary/main.go
```diff
@@ -104,7 +104,7 @@ func main() {
 
 	err = checkConfig(v, logger)
 	if err != nil {
-		logger.Fatal("invalid configuration", zap.Error(err))
+		logger.Panic("invalid configuration", zap.Error(err))
```
sojeri, line 107: why change from fatal to panic? isn't fatal sufficient to halt the flow?

pkg/models/move.go
```diff
@@ -577,12 +578,17 @@ func GenerateReferenceID(db *pop.Connection) (string, error) {
 
 // GenerateReferenceID creates a random ID for an MTO. Format (xxxx-xxxx) with X being a number 0-9 (ex. 0009-1234. 4321-4444)
 func generateReferenceIDHelper(db *pop.Connection) (string, error) {
-	min := 0
-	max := 9999
-	// #nosec G404 TODO needs review
-	firstNum := rand.Intn(max - min + 1)
-	// #nosec G404 TODO needs review
-	secondNum := rand.Intn(max - min + 1)
+	min := int64(0)
```
sojeri, line 581: it seems like we have several instances of needing two random numbers based on a low and a high input. can getting those numbers be abstracted into a util method similar to how you created the seed method `NewCryptoSeededSource`?

cmd/generate-shipment-summary/main.go
kimallen: oops, that was a misplaced change- will change it back.

pkg/models/move.go
kimallen: good point- I created some functions, need to work on tests for them.

kimallen: Tests added for the two new fns.

pkg/auth/authentication/login_gov.go
```diff
@@ -89,10 +91,17 @@ func (p LoginGovProvider) RegisterProvider(milHostname string, milClientID strin
...
 func generateNonce() string {
 	nonceBytes := make([]byte, 64)
+	//RA Summary: gosec - G404 - Insecure random number source (rand)
+	//RA gosec detected use of the insecure package math/rand rather than the more secure cryptographically secure pseudo-random number generator crypto/rand.
+	//RA This particular usage is mitigated by sourcing the seed from crypto/rand in order to create the new random number using math/rand.
+	//RA Developer Status: False Positive
+	//RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+	//RA Validator: jneuner@mitre.org
+	//RA Modified Severity: TBD
 	// #nosec G404 TODO needs review
```
gidjin: If this is here due to a false positive, as I understand from the above `//RA` comments, then please remove the `TODO needs review` possibly even replace it with `false positive due to secure seed` or similar

pkg/testingsuite/pop_suite.go
```diff
@@ -22,8 +24,16 @@ import (
...
-var seededRand = rand.New(rand.NewSource(time.Now().UnixNano()))
+//RA Summary: gosec - G404 - Insecure random number source (rand)
+//RA gosec detected use of the insecure package math/rand rather than the more secure cryptographically secure pseudo-random number generator crypto/rand.
+//RA This particular usage is mitigated by sourcing the seed from crypto/rand in order to create the new random number using math/rand.
+//RA Second, as part of the testing suite, the need for a secure random number here is not necessary.
+//RA Developer Status: False Positive
+//RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+//RA Validator: jneuner@mitre.org
+//RA Modified Severity: TBD
+// #nosec G404 TODO needs review
```
gidjin: Same here as earlier about removing `TODO needs review` for the false positive

pkg/iampostgres/iampostgres.go
```diff
@@ -90,10 +91,15 @@ func EnableIAM(host string, port string, region string, user string, passTemplat
...
 
 		// Add some entropy to this value so all instances don't fire at the same time
-		minDur := 100
-		maxDur := 5000
-		// #nosec G404 TODO needs review
-		wait := time.Millisecond * time.Duration(rand.Intn(maxDur-minDur)+minDur)
+		minDur := int64(100)
+		maxDur := int64(5000)
+		randDur := big.NewInt(maxDur - minDur)
+		randInt, err := rand.Int(rand.Reader, randDur)
```
gidjin, line 97: Small nit, but this is similar to the code I put in `pkg/db/sequence/random_sequencer.go`. Can it be refactored into a helper in the new `pkg/random/random.go` as you did with `GetRandomIntAddend`?

gidjin (commented):
Left a couple comments but it looks good so far. As for Panic vs Fatal, I found this post and sounds like Panic is what we would want. Though if another person more familiar with go could confirm or inform otherwise that would be appreciated.

jneuner-mitre (commented):
False Positive is only used when the linter has malfunctioned.  Looks like it is working fine here, so I changed the dev status to mitigated.

lgtm otherwise

pkg/iampostgres/iampostgres.go
kimallen: Done!

pkg/random/random_test.go
```diff
@@ -0,0 +1,35 @@
+package random
+
+import (
+	"testing"
+)
+
+func TestGetRandomInt(t *testing.T) {
+	intMax := 4
+	for i := 0; i < intMax*20; i++ {
```
kimallen, line 9: I made up a number of how many times to run this to ensure that we would have gotten a number out of range if the function allowed it.  Not sure how otherwise to test or choose a number.

gidjin: maybe just testing that `result` is within 0 and intMax exclusive of max?

pkg/iampostgres/iampostgres.go
gidjin: Thanks! is it not possible to use that function here too?

kimallen: oh yes!  I thought there was another one! thx

pkg/testdatagen/make_service_member.go
```diff
@@ -13,9 +15,12 @@ import (
 // randomEdipi creates a random Edipi for a service member
 func randomEdipi() string {
 	low := 1000000000
-	high := 9999999999
-	// #nosec G404 TODO needs review
-	return strconv.Itoa(low + rand.Intn(high-low))
+	high := 649999999999
```
sojeri, line 18: should this high value really change? I suspect it's a type from when you had int64 >_>

pkg/iampostgres/iampostgres.go
sojeri: +1 I'm surprised this file isn't using the new random fns

sojeri (commented):
I saw one change that I think might be a typo from when you had int64s, and there's one other place where I think we still need to use your new random int functions. aside from that, everything looks great! :)

sojeri:
also @gidjin can you link the post you mentioned in your earlier comment re: panic/fatal?

kimallen:
> Left a couple comments but it looks good so far. As for Panic vs Fatal, I found this post and sounds like Panic is what we would want. Though if another person more familiar with go could confirm or inform otherwise that would be appreciated.

For the changes I made, I used Panic.  The one I accidentally changed was existing code, and the other errors in that file used Fatal, so I changed it back and figure it's outside the scope of this PR to adjust that file.

pkg/testdatagen/make_service_member.go
kimallen: ooh, good catch.  yes, random cursor placement

kimallen:
> also @gidjin can you link the post you mentioned in your earlier comment re: panic/fatal?

Here's the one I found when I was doing research. https://stackoverflow.com/questions/58010284/should-i-use-log-panic-or-log-fatal-with-os-open

gidjin:
> > also @gidjin can you link the post you mentioned in your earlier comment re: panic/fatal?
> 
> Here's the one I found when I was doing research. https://stackoverflow.com/questions/58010284/should-i-use-log-panic-or-log-fatal-with-os-open (negative points for the question, but the answers seemed reasonable)

I can't believe I forgot to share the link, https://quasilyte.dev/blog/post/log-fatal-vs-log-panic/

gidjin (approved):
Thanks for refactoring the rands into a helper! LGTM :shipit:

PR #5226, opened 2020-11-13
[MB-5252] Display edit button on Step component

suzubara:
## Description

Display the Edit button on the Home screen Profile step (at all times) and Orders step (pre-submission only). 

## Reviewer Notes

Note: there is currently [a bug related to editing Orders in the API](https://dp3.atlassian.net/browse/MB-5239). This is not addressed in this PR. So when testing, verify you are able to access the Edit Orders form, but be aware that changing Orders information in the form will not work.

## Setup

```sh
make server_run
make client_run
```

Verify:
- User who has not submitted their move should be able to edit both their Profiles and Orders after initially adding them
- User who has submitted their move should still be able to edit Profile, but not Orders

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5252) for this change

## Screenshots

image


kimallen:
I found some buggy behavior where, if I have a user with profile and orders, and I edit the profile from the home link, when I return, the Add Orders button is visible and enabled and there's no edit button.

suzubara:
> I found some buggy behavior where, if I have a user with profile and orders, and I edit the profile from the home link, when I return, the Add Orders button is visible and enabled and there's no edit button.
> 
> Otherwise, code looks good and the other behavior was as expected.

Ok I looked into this and determined the reason is because the `uploaded_orders` returned by the PATCH service member call is returning an empty data set, even though the service member has uploaded orders documents. I created a bug ticket for this issue: https://dp3.atlassian.net/browse/MB-5529

We can resolve the UX by re-fetching orders data on the Home screen on mount, but this is a bandaid and I'm worried it would make the underlying bug less visible when it seems like an important thing to fix.

kimallen:
> > I found some buggy behavior where, if I have a user with profile and orders, and I edit the profile from the home link, when I return, the Add Orders button is visible and enabled and there's no edit button.
> > Otherwise, code looks good and the other behavior was as expected.
> 
> Ok I looked into this and determined the reason is because the `uploaded_orders` returned by the PATCH service member call is returning an empty data set, even though the service member has uploaded orders documents. I created a bug ticket for this issue: https://dp3.atlassian.net/browse/MB-5529
> 
> We can resolve the UX by re-fetching orders data on the Home screen on mount, but this is a bandaid and I'm worried it would make the underlying bug less visible when it seems like an important thing to fix. @kimallen what do you think?

Moving it to a bug sounds good to me, bandaid is unnecessary.

kimallen (approved):
LGTM!  Thanks for writing up that bug.

PR #5238, opened 2020-11-16
Use yarn frozen lockfile feature and avoid caching node_modules

gidjin:
## Description

The best practice recommendation for consistent builds when using `yarn` for managing dependencies is to use the `--frozen-lockfile` flag and rely on `yarn` to use its cache. Currently our build caches `node_modules` and the yarn cache which duplicates data caching. To speed up our builds and reduce the number of things we are caching this PR updates the jobs to use `--frozen-lockfile` and not cache `node_modules`.

This PR also removes the network concurrency flag as we now have releases of `react-uswds` so downloads can happen concurrently again.

## Reviewer Notes

Is this worth applying?

## Setup

Check out the builds

* [X] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)

## References

* [Slack Thread](https://trussworks.slack.com/archives/C2360K10T/p1605563953084300) for this change
* [this article](https://github.com/actions/cache/issues/67) explains more about the approach used.
* [this stack overflow](https://stackoverflow.com/questions/58482655/what-is-the-closest-to-npm-ci-in-yarn/64783982#64783982) explains more about the approach used.
* [yarn frozen lockfile docs](https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-frozen-lockfile) explains more about the approach used.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -502,26 +502,15 @@ jobs:
       - restore_cache:
           keys:
             - v3-cache-yarn-v3-{{ checksum "yarn.lock" }}
-      - restore_cache:
-          keys:
-            - v3-mymove-node-modules-{{ checksum "yarn.lock" }}
       - run:
           name: Install YARN dependencies
-          # setting network concurrency to 1 because using the default failed with errors trying to extract package tar files saying they were corrupt.
-          # this was workaround that seemed to fix the issue See for details https://github.com/yarnpkg/yarn/issues/7212
-          # This is caused by a timing issue when using react-uswds branch. It can be removed once we switch to a released version
-          command: yarn install --network-concurrency 1
+          command: yarn install --frozen-lockfile
       - run: scripts/check-generated-code yarn.lock
       # `v3-cache-yarn-v3-{{ checksum "yarn.lock" }}` is used to cache yarn sources
       - save_cache:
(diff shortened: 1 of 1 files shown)


gidjin:
FYI This is currently on experimental

sojeri (changes requested):
this removes node_modules caching altogether rather than using the yarn cache instead of node_modules. please implement the full recommended solution?

https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops#nodejsnpm

sojeri (approved):
ohohhhh jk I can't read. I see it now-- where yarn cache is being used.

:nothingtodohere:

.circleci/config.yml
```diff
@@ -1323,7 +1297,7 @@ workflows:
           # if testing on experimental, you can disable these tests by using the commented block below.
           filters:
             branches:
-              ignore: placeholder_branch_name
```
suzubara, line 1326: Should this change get merged to master?

suzubara (approved):
left a question, but the changes look good to me!

.circleci/config.yml
gidjin: No that's there because this is on experimental for testing, I'll revert it shortly

PR #5239, opened 2020-11-16
MB-5431 - Add annotation for detect private key pre-commit hook

amitch23:
## Description

Based on discovery by @kimallen , all the keys except one case, are fake keys used for local development. Therefore, including an annotation indicating as much is the desired outcome.

The exception is in pkg/cli/auth.go which does not include an actual key- just the formatting for interpolating one.

This PR adds the annotation comment in the pre-commit config for the hook detect-private-key.

## Reviewer Notes

It was not straightforward to include inline annotations for this pre-commit hook in the .ssh and .go files detected (see jira ticket for full list of files). As such, it seemed the most straightforward to use the exclude block in the pre-commit config, as it was intended to be used.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/secure/RapidBoard.jspa?rapidView=38&modal=detail&selectedIssue=MB-5431) for this change

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,6 +19,12 @@ repos:
       - id: check-yaml
         exclude: config/database.yml # database.yml is not a valid yaml file, it contains go templating
       - id: detect-private-key
+        #RA Summary: detect-private-key - Private key found.
+        #RA: detect-private-key detected a private key, but this key is a fake key intended to be used for a local environment only.
+        #RA Developer Status: Mitigated
+        #RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+        #RA Validator: jneuner@mitre.org
+        #RA Modified Severity:
         exclude: >
           (?x)^(
             config/tls/devlocal-.+\.key$|


sojeri (commented):
don't we need to separately annotate every single instance? what's to stop me tomorrow (or in 3 months) from checking in a real private key?

.pre-commit-config.yaml
```diff
@@ -19,6 +19,12 @@ repos:
       - id: check-yaml
         exclude: config/database.yml # database.yml is not a valid yaml file, it contains go templating
       - id: detect-private-key
+        #RA Summary: detect-private-key - Private key found.
+        #RA: detect-private-key detected a private key, but this key is a fake key intended to be used for a local environment only.
```
kimallen, line 23: I might add that `cli/auth.go` instance is triggered because of string interpolation.
This is the line where it is triggered: `privateKeyFormat := "-----BEGIN PRIVATE KEY-----\n%s\n-----END PRIVATE KEY-----"`

amitch23:
> don't we need to separately annotate every single instance? what's to stop me tomorrow (or in 3 months) from checking in a real private key?

@sojeri I originally tried to bypass the keys inline in the key (ssh) files and the auth.go file, but "#nosec" is a bypass code specific to the gosec linter, so that didn't work. I failed to find information about excluding or bypassing rules for the detect-private-key pre-commit hook inline, and everything led me to the pre-commit config file "exclude" param to handle these cases. I'm happy to pair on finding another way to do it inline, or spend some more time researching today (being conscious of the goals we have for this week).

To answer your question, the thing that would stop you would be that if you did add a key and didn't modify the pre-commit config, the pre-commit hook would fail. If you then modified the pre-commit config, the codeowners rule would automatically add a necessary review from the "cATO static analysis" group, once the dangerJS PR changes go in (https://github.com/transcom/mymove/pull/5240/files), which should hopefully be enough to catch the change.

.pre-commit-config.yaml
```diff
@@ -19,6 +19,12 @@ repos:
...
+        #RA Developer Status: Mitigated
+        #RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+        #RA Validator: jneuner@mitre.org
+        #RA Modified Severity:
         exclude: >
           (?x)^(
             config/tls/devlocal-.+\.key$|
```
jneuner-mitre, line 30: Each of these lines needs its own risk assessment.  If multiple findings are identified by one line (e.g. the first line), make sure the assessment is correct for all of the findings.

The assessment should usually be three sentences.
RA: <Why did the linter flag this line of code?>
RA: <Why is this line of code valuable?>
RA: <What mitigates the risk of negative impact?>

jneuner-mitre (changes requested):
We need to be more explicit.

.pre-commit-config.yaml
jneuner-mitre: I added a bit more detail to the template here:
https://docs.google.com/spreadsheets/d/1eH5ZYkKv7_c-p_D_y5G5ULdWgI_Zrwg_xL4WbLNZoqw/edit#gid=1701318393

amitch23: thanks @jneuner-mitre for the info. I don't think it's possible to add comments to a yaml file in the mapping block, as I tried to add risk assessments for each file/line, and am getting `InvalidConfigError`. I will look into this more tomorrow to see how we can be more granular/specific for each line while still adhering to acceptable config syntax.

sojeri: would it work to have a much more granular / specific regex and to annotate the group but call out each file as explicitly spelled out in the regex?

jneuner-mitre: @sojeri I think security types will always prefer explicit. ;)
@amitch23 If the current template doesn't work, feel free to propose an alternative way to annotate.

amitch23:
I accidentally merged this PR without intending to, but am not seeing the changes in master, oddly. It seems I cannot reopen this PR, so will make a new one

PR #5241, opened 2020-11-17
MB-5262 update SIT codes and add new ones

NamibiaLT:
## Description

Update prime.yaml with new SIT codes and remove unused fields.

## Reviewer Notes

Make sure sitEntryDate field was added correctly

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5262) for this change
* [Service code wiki](https://github.com/transcom/prime_api_deliverable/wiki/Service-Codes) explains more about the approach used.


swagger/prime.yaml
```diff
@@ -1051,8 +1051,7 @@ definitions:
             description: Service code allowed for this model type.
             enum:
               - DDFSIT # Domestic Destination First Day SIT
-          type:
-            $ref: '#/definitions/CustomerContactType'
```
sandy-wright, line 1055: This whole definition (`CustomerContactType` on line 757) can be removed - it isn't used anywhere else.

sandy-wright (approved):
LGTM - I had one comment about deleting the whole `CustomerContactType` definition, but otherwise looks great! You will need to regenerate the auto-gen go swagger files in order to merge this into master though. There are a couple of commands that can do this for you, but my go-to is `make server_run`. There will be ~8-17 files (I don't remember exactly 😅) that are changed by this and will need to be committed.

Approved so it can be merged once the gen files are pushed up!

swagger/prime.yaml
NamibiaLT: Just removed it.

PR #5253, opened 2020-11-17
Remediate all instances of gosec G304

kimallen:
## Description

This remediates all occurrences of gosec G304: Potential file inclusion via variable.  Although some of the instances could be justified because either they did not include user input or were only for tests, I went ahead and used the fix using `filepath.Clean()` suggested in the [securego.io documentation](https://securego.io/docs/rules/g304.html).

## Reviewer Notes

Is it extraneous to use filepath.Clean where it's not necessary?  Or ok to just use it for consistency and it doesn't hurt?
Although Discovery subtast suggested using annotation for ones that were acceptable and remediation for others, I chose to just do remediation for all.

[filepath.Clean() documentation](https://golang.org/pkg/path/filepath/#Clean)

This touches the serve.go- any reason to test in experimental first?

## Setup

`pre-commit run -v --all-files golangci-lint | grep G304`

There should be no occurrences found of G304.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4961) for this change

### DIFF
--- a/cmd/health-checker/main.go
+++ b/cmd/health-checker/main.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	neturl "net/url"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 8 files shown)


lynzt (approved):
LGTM 🚢 

I think using `filepath.Clean` is best practice and should be used even where not necessarily needed. The consistency is nice and I can't see it having a performance impact that isn't balanced by better security.

sojeri (approved):
this is a great solution :)

PR #5254, opened 2020-11-17
Add annotation for all instances of G101

kimallen:
## Description

All occurrences of G101: Password Management: Hardcoded Password I categorized as false positives, as the linter was flagging the use of the words "token" and "secret", and they were all references to names rather than the keys/secrets themselves.

## Reviewer Notes

## Setup

`pre-commit run -v --all-files golangci-lint | grep G101` and you will still see 5 occurrences.  However, all have annotation.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4970) for this change

### DIFF
--- a/pkg/auth/cookie.go
+++ b/pkg/auth/cookie.go
@@ -41,9 +41,9 @@ func (e *errInvalidHostname) Error() string {
 //RA: This line is used to identify the name of the token. GorillaCSRFToken is the name of the base CSRF token.
 //RA: This variable does not store an application token.
 //RA Developer Status: Mitigated
-//RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+//RA Validator Status: Mitigated
 //RA Validator: jneuner@mitre.org
-//RA Modified Severity:
+//RA Modified Severity: CAT III
 
 // GorillaCSRFToken is the name of the base CSRF token
 const GorillaCSRFToken = "_gorilla_csrf" // #nosec G101
--- a/pkg/auth/cookie.go
+++ b/pkg/auth/cookie.go
@@ -39,8 +39,8 @@ func (e *errInvalidHostname) Error() string {
 //RA Summary: gosec - G101 - Password Management: Hardcoded Password
 //RA: This line was flagged because it detected use of the word "token"
 //RA: This line is used to identify the name of the token. GorillaCSRFToken is the name of the base CSRF token.
-//RA: This variable does not identify the actual token, just the name, so is not a risk.
-//RA Developer Status: False Positive
+//RA: This variable does not store an application token.
+//RA Developer Status: Mitigated
 //RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
 //RA Validator: jneuner@mitre.org
 //RA Modified Severity:
--- a/pkg/auth/cookie.go
+++ b/pkg/auth/cookie.go
@@ -36,6 +36,15 @@ func (e *errInvalidHostname) Error() string {
 	return fmt.Sprintf("invalid hostname %s, must be one of %s, %s, or %s", e.Hostname, e.MilApp, e.OfficeApp, e.AdminApp)
 }
 
+//RA Summary: gosec - G101 - Password Management: Hardcoded Password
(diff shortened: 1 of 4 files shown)


pkg/auth/cookie.go
```diff
@@ -36,6 +36,15 @@ func (e *errInvalidHostname) Error() string {
...
+//RA Summary: gosec - G101 - Password Management: Hardcoded Password
+//RA: This line was flagged because it detected use of the word "token"
+//RA: This line is used to identify the name of the token.
+//RA: This variable does not identify the actual token, just the name, so is not a risk.
+//RA Developer Status: False Positive
+//RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+//RA Validator: jneuner@mitre.org
+//RA Modified Severity:
+
 // GorillaCSRFToken is the name of the base CSRF token
```
sojeri, line 48: I would put this into the annotation instead of or in addition to line 41. it explains in clearer detail what's going on here.

pkg/cli/auth.go
```diff
@@ -11,9 +11,17 @@ import (
...
+	//RA Summary: gosec - G101 - Password Management: Hardcoded Password
+	//RA: This line was flagged because of use of the word "secret"
+	//RA: This line is used to identify the name of the flag.
+	//RA: This value of this variable does not identify the actual secret, so is not a risk.
+	//RA Developer Status: False Positive
+	//RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+	//RA Validator: jneuner@mitre.org
+	//RA Modified Severity:
+
+	// ClientAuthSecretKeyFlag is the Client Auth Secret Key Flag // #nosec G101
```
sojeri, line 23: so for all of these-- IMO this part should be in the annotation instead of or as part of line 16 here

amitch23 (approved):
apart from Jeri's comments, lgtm! (I updated the description for checking to be g101 instead of a different one ;))

pkg/auth/cookie.go
kimallen: I can add it in addition.  If I remove this comment, I get a complaint because it expects a comment above it.

kimallen: I also had to add the line break between the annotation and the explanatory comment because it thought the first line of the annotation was the comment.

sojeri: that is rude, lol. thanks golint 😅

sojeri (approved):
🚀

pkg/cli/dps.go
```diff
@@ -15,15 +15,35 @@ const (
 	HTTPSDDCPortFlag string = "http-sddc-port"
 	// HTTPDPSServerNameFlag is the HTTP DPS Server Name Flag
 	HTTPDPSServerNameFlag string = "http-dps-server-name"
-	// DPSAuthSecretKeyFlag is the DPS Auth Secret Key Flag #nosec G101
+
+	//RA Summary: gosec - G101 - Password Management: Hardcoded Password
+	//RA: This line was flagged because of use of the word "secret"
+	//RA: This line is used to identify the name of the flag. DPSAuthSecretKeyFlag is the DPS Auth Secret Key Flag.
+	//RA: The value of this variable does not identify the actual secret, so is not a risk.
```
jneuner-mitre, line 22: We should not assert whether something is or is not a risk.  Just state why the risk is reduced.

jneuner-mitre:
```suggestion
	//RA: This variable does not store an application secret.
```

pkg/cli/dps.go
```diff
@@ -15,15 +15,35 @@ const (
...
+	//RA Developer Status: False Positive
```
jneuner-mitre, line 23: The linter appears to be working correctly, so a False Positive status is incorrect.  
The finding has been assessed, and risk appears to be Mitigated because a application secret was not found.

jneuner-mitre:
```suggestion
	//RA Developer Status: Mitigated
```

jneuner-mitre (changes requested):
Applying these recommended changes to all the G101 findings in this commit and I'll concur.

jneuner-mitre (approved):
lgtm

PR #2308, opened 2019-06-20
Use migrations manifest

pjdufour-truss:
## Description

Rather than running whatever files are in the migrations folders, `milmove migrate` will now error on any files that were not listed in the migrations manifest.

Added a pre-commit hook, that updates the manifest.  To avoid issues with merge commits, the new pre-commit hook doesn't run in CI.  However, if a commit is missing from the manifest the server tests will fail.

We'll have the added benefit of an easy to view changelog of migrations.

## Reviewer Notes

**Local Migrations**

```shell
make db_dev_destroy && make db_dev_start && DB_TIMEOUT=30 make db_dev_create && make db_dev_migrate
```

## Setup

None

## Code Review Verification Steps

* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/tree/master/docs/backend.md#logging)
* [ ] The requirements listed in
 [Querying the Database Safely](https://github.com/transcom/mymove/tree/master/docs/backend.md#querying-the-database-safely)
 have been satisfied.
* Any new migrations/schema changes:
  * [ ] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://github.com/transcom/mymove/tree/master/docs/database.md#zero-downtime-migrations))
  * [ ] Have been communicated to #dp3-engineering
  * [ ] Secure migrations have been tested using `scripts/run-prod-migrations`
* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [ ] Have the Pivotal acceptance criteria been met for this change?

## References

None

## Screenshots

None

PR #5257, opened 2020-11-17
Mb 5518 static analysis g306

kimallen:
## Description

Adds annotation to justify usage of permission level 644 for the writefile (the linter expects 600 or lower).  There is only 1 occurrence of G306.

## Reviewer Notes

 Permissions of 600 mean that the owner has full read and write access to the file, while no other user can access the file. Permissions of 644 mean that the owner of the file has read and write access, while the group members and other users on the system only have read access.In this situation, we want to protect from any changes to migration files once they have been deployed, and changes to the db should happen only in a new migration file.

## Setup

`pre-commit run -v --all-files golangci-lint | grep G306`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5518) for this change
* [Article](https://www.linode.com/docs/guides/modify-file-permissions-with-chmod/) about permission levels.

Migration to add new fake TSP office to test the migration file with perms 600, deployed to experimental and visible in the admin console:
Screen Shot 2020-11-19 at 3 26 55 PM


sojeri:
Thanks for including the file permissions article! This one is a new concept for me and it's great to have the ref handy. Will review when I'm done reading, though without additional context the annotation seems good.

cmd/milmove/gen.go
```diff
@@ -77,7 +77,18 @@ func writeEmptyFile(migrationPath, filename string) error {
...
+	//RA: Permissions of 600 mean that the owner has full read and write access to the file, while no other user can access the file.
+	//RA: Permissions of 644 mean that the owner of the file has read and write access, while the group members and other users on the system only have read access.
+	//RA: In this situation, we want to protect from any changes to migration files once they have been deployed, and changes to the db should happen only in a new migration file.
+	//RA: Therefore, 644 is the appropriate permission level in this case.
+	//RA Developer Status: Mitigated
+	//RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+	//RA Validator: jneuner@mitre.org
+	//RA Modified Severity:
+	// #nosec G306
 	err := ioutil.WriteFile(path, []byte{}, 0644)
```
sojeri, line 92: huh. I traced this line back to its origin in `@pjdufour-truss`'s PR https://github.com/transcom/mymove/pull/2308/files, but noboody there or on the inbetween PRs had the same question I did: why are we setting 0644 here? is there some reason we should not write this file with 0600 or even 0200 and then chmod it to read only / 0444?

sojeri: ok, nope, I am super wrong. I asked about this in a couple truss eng channels, and the short answer here is since the files being generated by this are then committed to the repo which is open source --- as opposed to being created and used only on the server -- then we don't really care about this file permission. it's not a security risk to have these files set to `rw-r-r`, because anyone can pull the repo, check out a branch, and make modifications to the file anyway.

jneuner-mitre: What process needs to read this file?
Why not set the process group, and use 0640?
Does the process still work if we used 0600?

lynzt (approved):
LGTM 🚢 

To Jarod's point, the commit worked locally but unsure how the file permission would work w/ circle. Because we can't "see" the comments, might be good to try w/ a table on experimental that we have access to via the admin interface. We could add a test SM and see if that works?

PR #5262, opened 2020-11-18
MB-5308 Add sorting to TOO queue backend

Ryan-Koch:
## Description

Adds the ability to include a sort and order parameter to requests for TOO queue data.

## Setup
Once you populate with data using the command below and log into the TOO side of the office app you can visit here:
http://officelocal:3000/ghc/v1/queues/moves?page=1&perPage=20&order=asc&sort=locator

Which should give you a list of moves with the COMBOS locator ID item first. If you flip the order to `desc` it should show it second.

```sh
make db_dev_e2e_populate
```

 have been satisfied.

Jira: https://dp3.atlassian.net/browse/MB-5308


pkg/services/move_order/move_order_fetcher.go
```diff
@@ -215,3 +215,14 @@ func gblocFilter(gbloc string) FilterOption {
 		query = query.Where("transportation_offices.gbloc = ?", gbloc)
 	}
 }
+
+func sortOrder(sort *string, order *string) FilterOption {
```
noelledusahel, line 219: this makes me wanna rename FilterOption, in favor of QueryOption, since its not being applied to a non-filter

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -215,3 +215,14 @@ func gblocFilter(gbloc string) FilterOption {
...
+	return func(query *pop.Query) {
+		// If we have a sort and order defined let's use it. Otherwise we'll use our default status desc sort order.
+		if sort != nil && order != nil {
+			query = query.Order(fmt.Sprintf("%s %s", *sort, *order))
```
noelledusahel, line 223: @duncan-truss  had a good point, we're going to need to create a special sort query for customer name that sorts both on last and first name

pkg/handlers/ghcapi/queues.go
```diff
@@ -44,6 +44,8 @@ func (h GetMovesQueueHandler) Handle(params queues.GetMovesQueueParams) middlewa
 		Status:                 params.Status,
 		Page:                   params.Page,
 		PerPage:                params.PerPage,
+		Sort:                   params.Sort,
+		Order:                  params.Order,
 	}
 
 	// Let's set default values for page and perPage if we don't get arguments for them. We'll use 1 for page and 20
```
noelledusahel, line 51: i had to add some if-nil conditions [here](https://github.com/transcom/mymove/pull/5264/files#diff-3396d23c61ae3630ac718a7090ac0733d634c3132d4e00b71ae2050dafa47edcR110) because without param for sort and order, the queue breaks. it seems like the same thing is happening with your TOO queue

noelledusahel: oh wait, on second look its not breaking 👍

pkg/handlers/ghcapi/queues_test.go
```diff
@@ -336,6 +336,29 @@ func (suite *HandlerSuite) TestGetMoveQueuesHandlerFilters() {
...
+
+		payload := response.(*queues.GetMovesQueueOK).Payload
+		suite.EqualValues(3, payload.TotalCount)
+		suite.Len(payload.QueueMoves, 3)
+		// test that the moves are sorted by status descending
+		suite.Equal(ghcmessages.QueueMoveStatus("APPROVALS REQUESTED"), payload.QueueMoves[0].Status)
+		suite.Equal(ghcmessages.QueueMoveStatus("APPROVED"), payload.QueueMoves[1].Status)
+		suite.Equal(ghcmessages.QueueMoveStatus("SUBMITTED"), payload.QueueMoves[2].Status)
+	})
+
```
monfresh, line 361: Since the sorting is being handled by the service, and since we already have a test in move_order_fetcher_test.go, what do you think about removing this test? I don't think we need 2 tests that hit the DB.

pkg/services/move_order/move_order_fetcher_test.go
```diff
@@ -265,3 +266,33 @@ func (suite *MoveOrderServiceSuite) TestListMovesWithPagination() {
...
+
+	moveOrderFetcher := NewMoveOrderFetcher(suite.DB())
+	params := services.ListMoveOrderParams{Page: swag.Int64(1), PerPage: swag.Int64(2), Sort: swag.String("locator"), Order: swag.String("asc")}
+	moves, _, err := moveOrderFetcher.ListMoveOrders(officeUser.ID, &params)
+
+	suite.NoError(err)
+	suite.Equal(2, len(moves))
+	suite.Equal(expectedMoveLocators[0], moves[0].Locator)
+	suite.Equal(expectedMoveLocators[1], moves[1].Locator)
+}
```
monfresh, line 298: Is this PR meant to only support status ordering, or all the other possible sort options? If the latter, could we please add subtests for all the other options?

pkg/services/move_order/move_order_fetcher.go
noelledusahel: https://github.com/transcom/mymove/pull/5264/files#diff-5d8ca3b207795fe25ec589bd2306d395c4641c1970d66f9bc00294b861dc6c0cR110

pkg/services/move_order/move_order_fetcher_test.go
Ryan-Koch: Yeah I can add the other subtests. Will push that up soon and update this thread.

pkg/handlers/ghcapi/queues_test.go
Ryan-Koch: That's fair, I'll remove this and we can lean on the service tests.

pkg/services/move_order/move_order_fetcher.go
Ryan-Koch: Went ahead and used what you're doing on the TIO side so it's consistent.

pkg/services/move_order/move_order_fetcher.go
Ryan-Koch: That's a good point. Renamed it to QueryOption.

pkg/services/move_order/move_order_fetcher_test.go
Ryan-Koch: Added in tests for each field going each direction

pkg/handlers/ghcapi/queues_test.go
Ryan-Koch: Done!

pkg/services/move_order/move_order_fetcher_test.go
```diff
@@ -265,3 +266,147 @@ func (suite *MoveOrderServiceSuite) TestListMovesWithPagination() {
...
+		// Only moves with shipments are returned, so we need to add a shipment
+		// to the move we just created
+		testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+			Move: expectedMove,
+			MTOShipment: models.MTOShipment{
+				Status: models.MTOShipmentStatusSubmitted,
+			},
+		})
+
+		if i == 2 {
```
duncan-truss, line 298: This block won't run if your condition is i < 2

Ryan-Koch: Ack that's a typo. Thanks for pointing that out. Will fix.

pkg/services/move_order/move_order_fetcher_test.go
```diff
@@ -265,3 +266,147 @@ func (suite *MoveOrderServiceSuite) TestListMovesWithPagination() {
...
+	suite.NoError(err)
+	suite.Equal(2, len(moves))
+	suite.Equal(expectedServiceMemberNames[0], *moves[0].Orders.ServiceMember.LastName+", "+*moves[0].Orders.ServiceMember.FirstName)
+	suite.Equal(expectedServiceMemberNames[1], *moves[1].Orders.ServiceMember.LastName+", "+*moves[1].Orders.ServiceMember.FirstName)
+
+	params = services.ListMoveOrderParams{Page: swag.Int64(1), PerPage: swag.Int64(2), Sort: swag.String("service_members.last_name"), Order: swag.String("desc")}
+	moves, _, err = moveOrderFetcher.ListMoveOrders(officeUser.ID, &params)
+	suite.NoError(err)
+	suite.Equal(2, len(moves))
+	suite.Equal(expectedServiceMemberNames[0], *moves[0].Orders.ServiceMember.LastName+", "+*moves[0].Orders.ServiceMember.FirstName)
```
duncan-truss, line 349: This should now be the reverse of the previous test right?
```
suite.Equal(expectedServiceMemberNames[0], *moves[1].Orders.ServiceMember.LastName+", "+*moves[1].Orders.ServiceMember.FirstName)
suite.Equal(expectedServiceMemberNames[1], *moves[0].Orders.ServiceMember.LastName+", "+*moves[0].Orders.ServiceMember.FirstName)
```

Ryan-Koch: Correct, I made a bunch of changes here that should have fixed this as well.

pkg/services/move_order/move_order_fetcher_test.go
Ryan-Koch: fixed

pkg/services/move_order/move_order_fetcher_test.go
```diff
@@ -256,3 +256,114 @@ func (suite *MoveOrderServiceSuite) TestListMovesWithPagination() {
...
+
+	expectedMove2 := expectedShipment2.MoveTaskOrder
+
+	// Sort by service member name
+	params := services.ListMoveOrderParams{Sort: swag.String("lastName"), Order: swag.String("asc")}
+	moves, _, err := moveOrderFetcher.ListMoveOrders(officeUser.ID, &params)
+
+	suite.NoError(err)
+	suite.Equal(2, len(moves))
+	suite.Equal("Spacemen, Leo", *moves[0].Orders.ServiceMember.LastName+", "+*moves[0].Orders.ServiceMember.FirstName)
```
noelledusahel, line 299: ohh was the problem before that every time tests ran the order of ‘expectedNames’ array changed because it re-sorted?

Ryan-Koch: There was a combo of that and accidentally creating 3 moves when running through the loop, so after pairing a bit with Moncef I wound up simplifying it a bit since it was really just a comparison of 2 moves I was going for.

pkg/services/move_order/move_order_fetcher_test.go
```diff
@@ -256,3 +256,114 @@ func (suite *MoveOrderServiceSuite) TestListMovesWithPagination() {
...
+			Status: models.MTOShipmentStatusSubmitted,
+		},
+	})
+
+	expectedShipment2 := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+		MTOShipment: models.MTOShipment{
+			Status: models.MTOShipmentStatusSubmitted,
+		},
+		Move: models.Move{
+			Status:  models.MoveStatusSUBMITTED,
```
noelledusahel, line 283: i think one of these moves should have a different status so that you are able to test sorting

noelledusahel: it LGTM after this bit. on [MB-5306](https://github.com/transcom/mymove/pull/5262/files#diff-6e59306980fce1889ad0c1b6a0015634dc8f5aa4507777e03b750aaf66df569bR299) i encountered status sorting in a way i didn't expect though.

Ryan-Koch: good catch, pushed the change.

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -211,8 +211,33 @@ func moveStatusFilter(statuses []string) FilterOption {
...
+	}
+
+	return func(query *pop.Query) {
+		// If we have a sort and order defined let's use it. Otherwise we'll use our default status desc sort order.
+		if sort != nil && order != nil {
+			sortTerm := parameters[*sort]
+			if sortTerm == "service_members.last_name" {
+				query = query.Order(fmt.Sprintf("service_members.last_name %s, service_members.first_name %s", *order, *order))
+			} else {
+				query = query.Order(fmt.Sprintf("%s %s", sortTerm, *order))
```
donaldthai, line 237: Would `order` be better if we turn it into a bool type in the swagger definition? Passing a `*string` type is a bit concerning at first glance until I saw the swagger definition restricting the query field to enums.

I think changing the `order` query field to bool would make it less ambiguous. Eg. True, order by asc. False, order by desc.

What do you think?

noelledusahel (approved):

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -211,8 +211,33 @@ func moveStatusFilter(statuses []string) FilterOption {
...
 }
+
+func sortOrder(sort *string, order *string) QueryOption {
+	parameters := map[string]string{
+		"lastName":               "service_members.last_name",
+		"dodID":                  "service_members.edipi",
+		"branch":                 "service_members.affiliation",
+		"moveID":                 "moves.locator",
+		"status":                 "moves.status",
+		"destinationDutyStation": "duty_stations.name",
```
duncan-truss, line 227: Did you not add shipmentsCount here because we wait to calculate that in the payload?

pkg/services/move_order/move_order_fetcher.go
donaldthai: Something to consider. Doesn't have to be this PR.

pkg/services/move_order/move_order_fetcher.go
Ryan-Koch: Agreed, we should get that moved over. I think I'm going to take an action item to write something up for that.

duncan-truss:
Most of the columns sorted well for me when I integrated what I had from the frontend.  The destinationDutyStation column was coming back out of order for both sorts.

PR #5263, opened 2020-11-18
Add SIT fields to MTOServiceItem model

sandy-wright:
## Description

This PR adds the new SIT fields that we added to the `mto_service_items` table last sprint to the `MTOServiceItem` model.

## Setup

You can run server tests to check that everything is still functional as expected.


abbyoung (approved):
lgtm!

PR #5266, opened 2020-11-18
[MB-4063] TIO frontend sorting of payment requests queue

duncan-truss:
## Description

This PR adds the ability to sort the TIO payment request queue table columns.  The default sort on load should show the table sorted by age in descending order.  The Origin GBLOC column should be disabled and not sortable since for all TOO users this will have the same value except for the special Marine Corps transportation office.

## Reviewer Notes

I added some flexbox styling because when filtering the queue to one row I noticed the height would cut off the date picker when expanded.  I made sure the queue and various containers were growing in the vertical direction to fix this bug.

## Setup

```sh
make db_dev_e2e_populate
make server_run
make office_client_run
```

1. Login as the TIO office user
2. View the queue sorted presorted by the age column in the descending direction
3. Clicking on the age column should toggle to the unsorted state or clicking on another column should sort them in the ascending direction.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4063) for this change

## Screenshots

Screen Recording 2020-11-23 at 11 19 08 AM mov


noelledusahel (approved):
Looks good! I wish we didn't have the clunky " third click to remove sorting" functionality though

amanda5iu (approved):

donaldthai (approved):
Code looks good to me. Sorting works great!

One thing I did noticed was that on the third click of a column, the sort gets unset. According to the story, we may not want this behavior? Don't know how important it is though.

> Then that column toggles between sorting
>    ascending and descending

duncan-truss:
@noelledusahel @donaldthai Yea that is the default react-table behavior with the toggle unsetting the column.  I explored it but did not feel we had the time to implement our own toggleSort handler properly and test it.  It's definitely possible but there's a good bit going on in that TableQueue component already.

src/components/Table/Table.jsx
```diff
@@ -62,7 +74,12 @@ const Table = ({
                 {row.cells.map((cell, index) => {
                   return (
                     // eslint-disable-next-line react/no-array-index-key
-                    <td key={`cell${index}`} data-testid={`${cell.column.id}-${cell.row.id}`} {...cell.getCellProps()}>
+                    <td
+                      key={`cell${index}`}
```
sojeri, line 78: if you don't mind, I'd love to see this one fixed as well. this linter error is firing because indexes do not make good keys in react. the point of a key is to have a static ID which doesn't change when a list is reordered or when some items are removed / added. when you use an index, you're giving a dynamic value instead which potentially changes every time the list is updated in some way. you should instead pass in an object which has an immutable ID, and use that ID for the key.

this is a cat III as well.

src/components/Table/Table.jsx
```diff
@@ -35,11 +36,22 @@ const Table = ({
...
-                {headerGroup.headers.map((column, headerIndex) => (
-                  <th key={`header${headerIndex}`} data-testid={column.id} {...column.getHeaderProps()}>
-                    {column.render('Header')}
-                  </th>
-                ))}
+                {headerGroup.headers.map((column, headerIndex) => {
+                  const sortProperties = column.canSort ? column.getSortByToggleProps() : {};
+                  return (
+                    <th
+                      key={`header${headerIndex}`}
```
sojeri, line 43: this is another cat III array index as key issue. instead of a dynamic index value, this should be an immutable object id.

sojeri (changes requested):
👋 hello! I am here, summoned by dangerJS because you have added a new eslint disable line. I left 3 related comments. the one I am requesting changes for is the new eslint line you've added re: unused variables. the other two I would also super appreciate if you fix now, but those fixes aren't blocking this PR since they are already known issues in the static analysis backlog.

duncan-truss:
> 👋 hello! I am here, summoned by dangerJS because you have added a new eslint disable line. I left 3 related comments. the one I am requesting changes for is the new eslint line you've added re: unused variables. the other two I would also super appreciate if you fix now, but those fixes aren't blocking this PR since they are already known issues in the static analysis backlog.

Cthulhu has been awoken 🙈

src/components/Table/TableQueue.jsx
```diff
@@ -9,19 +9,42 @@ import Table from 'components/Table/Table';
...
+  disableMultiSort,
+  defaultCanSort,
+  disableSortBy,
+  defaultSortedColumns,
+  handleClick,
+  useQueries,
+  showFilters,
+  showPagination,
+}) => {
+  // eslint-disable-next-line no-unused-vars
```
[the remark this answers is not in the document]
duncan-truss: I just fixed this in another branch!

src/components/Table/Table.jsx
```diff
@@ -36,11 +36,11 @@ const Table = ({
           {headerGroups.map((headerGroup, hgIndex) => (
             <Fragment key={`headerGroup${hgIndex}`}>
               <tr {...headerGroup.getHeaderGroupProps()}>
-                {headerGroup.headers.map((column, headerIndex) => {
+                {headerGroup.headers.map((column) => {
```
sojeri, line 39: it won't let me comment there, but I think you can remove the linter disable line at the top of this file now. thanks for making this change! :)

sojeri (approved):
looks like you can remove one more eslint disable line, but that's not required for merge -- just nice to have.

thanks for making a couple tweaks for me!

🚀

src/components/Table/Table.jsx
duncan-truss: I think I tried that and it flagged the headerGroup on line 37 (which should only ever have one row) and some others related to the page numbers which are basically indexes 🙁

PR #5267, opened 2020-11-18
[MB-5354][MB-5536] Fix bug in fetch-mto-updates causing identical destination addresses

reggieriser:
## Description

This PR fixes a bug in the prime API's `fetch-mto-updates` (and potentially other code that uses the Prime API's model-to-payload converter).  In particular, every destination duty station address returned by `fetch-mto-updates` was being set to the same street/city/state/zip regardless of what address was actually stored in the database (essentially, the last one through the loop "won" as all the addresses were pointing to the same chunk of memory).  I also found that it affected the `isFinal` field of a payment request if more than one payment request was attached to a given move.

The root cause was that we were taking the address of a loop variable set via a range, but not appreciating that Go instantiates the loop variable once and then just copies each item to the same loop variable each iteration.  Taking the address of that loop variable results in the same pointer every iteration which can lead to some hard-to-detect issues.

I have also added [documentation](https://github.com/transcom/mymove/wiki/Using-loop-iterator-variables-in-go) about this to our Wiki.

Some general references about the issue:
- https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable
- https://stackoverflow.com/questions/48826460/using-pointers-in-a-for-loop

## Reviewer Notes

- Note that this issue is flagged by the new version of the gosec linter as `G601`, so that should help identify these scenarios.
- Make sure the fix (and the comment next to it) seem clear enough for others that run into this code.
- We may not have ever noticed this with our dev seed data because all of our available-to-prime moves were using the same address.  I added another move with a unique address in our dev seed data to make the issue more obvious.

## Setup

- `make server_build db_dev_e2e_populate`
- I made a test case that demonstrates the issue.  If you roll back the changes to `model_to_payload.go`, you can see the test fail and demonstrate the problem.  Reapply the changes and all tests should pass.
- This problem was hidden in our dev seed data because the three available to prime moves there all were set up with the same address.  I added another one with a unique address in this PR.  If you roll back the changes to `model_to_payload.go` again, then run a `prime-api-client --insecure fetch-mto-updates`, you should see the identical address problem.  Reapply and the issue should go away.  Comparing the two resulting JSON responses should make that clear.
- Until this is merged, you can also see the issue running `fetch-mto-updates` against staging.

## References

* [Parent story](https://dp3.atlassian.net/browse/MB-5343) for this change
* [Subtask story](https://dp3.atlassian.net/browse/MB-5354) for the unit test
* [Subtask story](https://dp3.atlassian.net/browse/MB-5536) for the fix
* [Wiki documentation](https://github.com/transcom/mymove/wiki/Using-loop-iterator-variables-in-go)

### DIFF
--- a/pkg/handlers/primeapi/move_task_order_test.go
+++ b/pkg/handlers/primeapi/move_task_order_test.go
@@ -7,6 +7,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/go-openapi/swag"
+
 	mtoserviceitem "github.com/transcom/mymove/pkg/services/mto_service_item"
 
 	"github.com/transcom/mymove/pkg/gen/primemessages"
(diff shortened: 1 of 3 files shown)


pkg/handlers/primeapi/move_task_order_test.go
```diff
@@ -262,6 +264,137 @@ func (suite *HandlerSuite) TestListMoveTaskOrdersHandlerReturnsUpdated() {
...
+
+func (suite *HandlerSuite) equalPaymentRequest(expected models.PaymentRequest, actual *primemessages.PaymentRequest) {
+	suite.Equal(expected.ID.String(), actual.ID.String())
+	suite.Equal(expected.MoveTaskOrderID.String(), actual.MoveTaskOrderID.String())
+	suite.Equal(expected.IsFinal, *actual.IsFinal)
+	suite.Equal(expected.Status.String(), string(actual.Status))
+	suite.Equal(expected.RejectionReason, actual.RejectionReason)
+	suite.Equal(expected.PaymentRequestNumber, actual.PaymentRequestNumber)
+}
+
```
monfresh, line 316: Excellent use of helper functions!

pkg/testdatagen/scenario/devseed.go
```diff
@@ -1655,4 +1692,7 @@ func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUplo
 	// This move below is a PPM move in DRAFT status. It should probably
 	// be changed to an HHG move in SUBMITTED status to reflect reality.
 	createMoveWithBasicServiceItems(db, userUploader)
+	// Sets up a move with a non-default destination duty station address
+	// (to more easily spot issues with addresses being overwritten).
+	createMoveWithUniqueDestinationAddress(db)
```
monfresh, line 1697: Not a blocker for this PR, but I've noted other places where we use the same data, such as the same email for service members, where it would be better to have unique data using some kind of faker package.

reggieriser: Agreed.  I think we miss some bugs like this because of the similar nature of the default objects we're creating.

monfresh (approved):
LGTM. I verified that with this PR, I see Augusta instead of Columbia.

pkg/handlers/primeapi/payloads/model_to_payload.go
```diff
@@ -53,8 +53,8 @@ func MoveTaskOrders(moveTaskOrders *models.Moves) []*primemessages.MoveTaskOrder
 	payload := make(primemessages.MoveTaskOrders, len(*moveTaskOrders))
 
 	for i, m := range *moveTaskOrders {
-		// #nosec G601 TODO needs review
-		payload[i] = MoveTaskOrder(&m)
+		copyOfM := m // Make copy to avoid implicit memory aliasing of items from a range statement.
+		payload[i] = MoveTaskOrder(&copyOfM)
```
jacquelineIO, line 57: The other way to do is `MoveTaskOrder(m)`. This would do copy by value which is essentially what you are with the combination of lines 56 and 57.

reggieriser: True, although that changes the signatures all the way down this chain which would have a ripple effect on any callers since these are public methods (I haven't dug in to see how painful it would be, so it may not end up being that bad).  But I will add something to the Wiki page noting this as another option.

PR #5287, opened 2020-11-19
MB-5052: update call to chamber 

mdawn:
## [Update Chamber Refs](https://dp3.atlassian.net/browse/MB-5052)

How should I deal with references to other dependencies inside the `.envrc` file that may need `transcom-ppp` to access them, like [maybe this bucket](https://github.com/transcom/mymove/blob/c99ff6be7d873d3d9c0896e7df2d0a2e6b224a92/.envrc#L176)?

I thought about removing the `export AWS_PROFILE=...` on line 50 entirely, but it seems pulling on that thread will cause wailing and gnashing of teeth. Not sure where to go from here. Please advise.

For this ^ and even MOAR thought-provoking questions, please see the ticket. 🙂


travelar (approved):
This seems benign, but I think merging this will break other things in flight. Given the inter-dependencies of the devlocal ecosystem, we may need to  work this series of tickets in separate branch to avoid derailing app engineers, but I don't know that for sure. There will probably be some trials and tribulations here, so tread lightly and be prepared to roll-back if necessary.

PR #5296, opened 2020-11-20
MB-5342 TXO see usmc moves

donaldthai:
## Description

> For this one we’re just worried about making sure the branch dropdown on the front end is able to adapt to the backend work that was done for this in MB-4692: TOO & TIO can be assigned to handle USMC moves and see USMC moves in their queue (backend)
> 
> For this we will likely need to look at what we’re returning from the API side for the office user info and expand it to include what we need to handle these cases appropriately.
> 
> For non-USMC office users we should remove the USMC filter option from the branch drop down
> 
> As a USMC TOO/TIO, I want to be able to handle all USMC moves, so that…
> 
> Given that I am a TOO/TIO assigned to the USMC GBLOC
>  When I look at my move queue
>  Then I see all moves for USMC service members
> Given that I am *not* a TOO/TIO assigned to the USMC GBLOC
>  When I look at my move queue
>  Then I *don't* see any moves for USMC service members
>  
> 
> USTC is in the process of establishing the USMC PPSO Queue in MilMove.
> 
> GBLOC:  USMC
> Location:  Camp LeJeune, NC
> PPSO POC:  (need name, phone, email, etc.)
> Name for new record should be: Camp LeJeune (USMC)

## Reviewer Notes

- Did a bit of work to refactor old payload code to make it more reusable 
- Added an new user that has both TOO/TIO roles but with the gbloc USMC to test
- Opted to load the user with react-query instead of trying to hook up with redux to avoid over complicating the components

## Setup

```sh
make db_dev_e2e_populate
make server_run
make office_client_run
```

**To test non-USMC gblocs**
1. Login as `too_tio_role@office.mil`, local
2. Click on branch dropdown to check that `Marine Corps` is not added
3. Switch roles to check the same

**To test USMC gblocs**
1. Login as `too_tio_role_usmc@office.mil`, local, new from devseed
2. Check to see branch filtering is removed
3. Switch roles to check the same

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5342) for this change

## Screenshots

User not of USMC gbloc, Marine Corps removed from options
image

User of USMC gbloc, branch filtering is removed
image


duncan-truss:
@donaldthai This looks good! Should we have the branch filter menu at all for a USMC office user?

donaldthai:
> @donaldthai This looks good! Should we have the branch filter menu at all for a USMC office user?

Not sure. @Ryan-Koch what do you think?

pkg/handlers/internalapi/users.go
```diff
@@ -40,14 +42,27 @@ func (h ShowLoggedInUserHandler) Handle(params userop.ShowLoggedInUserParams) mi
...
 	if !session.IsServiceMember() {
+		var officeUser *models.OfficeUser
+		if session.OfficeUserID != uuid.Nil {
+			fetchedOfficeUser, err := h.officeUserFetcherPop.FetchOfficeUserByID(session.OfficeUserID)
+			if err != nil {
+				logger.Error("Error retrieving office_user", zap.Error(err))
+				return userop.NewIsLoggedInUserInternalServerError()
+			}
+
+			officeUser = &fetchedOfficeUser
```
duncan-truss, line 53: You shouldn't need this if officeUser isn't defined as a pointer right?

donaldthai: Oh yea, that's a good catch. 

So something like this then:
```golang
var officeUser models.OfficeUser
var err error
if session.OfficeUserID != uuid.Nil {
	officeUser, err = h.officeUserFetcherPop.FetchOfficeUserByID(session.OfficeUserID)
	if err != nil {
		logger.Error("Error retrieving office_user", zap.Error(err))
		return userop.NewIsLoggedInUserInternalServerError()
	}
}
```

src/pages/Office/MoveQueue/MoveQueue.jsx
```diff
@@ -56,8 +46,9 @@ const columns = [
     {
       id: 'branch',
       isFilterable: true,
-      // eslint-disable-next-line react/jsx-props-no-spreading
```
sojeri, line 59: it's fine to disable `react/jsx-props-no-spreading`, but I don't understand why you're removing this and adding the disable whole file one at the top (vs not touching it like the one on line 44/34). since it's better for our annotated linter warnings to be inline with the offending code, I recommend keeping this next line disable statement. that way we can follow the same pattern everywhere. :)

src/pages/Office/PaymentRequestQueue/PaymentRequestQueue.jsx
```diff
@@ -1,7 +1,8 @@
+/* eslint-disable react/jsx-props-no-spreading */
```
sojeri, line 1: same for this file-- for consistency please use only the disable next line for now. :)

sojeri (approved):
the linter disable lines in this PR are acceptable. I flagged a couple places where I believe we'd prefer to use next-line disable statements rather than whole-file ones.

monfresh:
My brain has been foggy lately, so it's possible I'm missing something. What is the reasoning for changing the internal API and introducing a second DB query to fetch the office user when the GHC API already fetches the office user's GBLOC  via `FetchGblocForOfficeUser`? Why can't we return the GBLOC to the TXO handlers from `ListMoveOrders` and `FetchPaymentRequestList`?

monfresh (changes requested):
Please see my question on the main PR page.

monfresh (approved):

donaldthai:
> My brain has been foggy lately, so it's possible I'm missing something. What is the reasoning for changing the internal API and introducing a second DB query to fetch the office user when the GHC API already fetches the office user's GBLOC via `FetchGblocForOfficeUser`? Why can't we return the GBLOC to the TXO handlers from `ListMoveOrders` and `FetchPaymentRequestList`?

TL;DR;

@monfresh and I chatted in a zoom meeting and I explained my thoughts on the solution given:
- Wanted to keep the user data consistent to models we have
- Introduced a react-query call since we're in a state where old code is using `redux` and new pages are using `react-query`
- Future proofing with `react-query` calls since we have the ability to turn on caching later

amanda5iu (approved):

PR #5307, opened 2020-11-23
[MB-4060] TOO fronted moves queue sorting

duncan-truss:
## Description

This PR adds sorting to the TOO moves queue table component.  The default sort should be on the status ascending direction when initially loaded.  You can then toggle between the reverse (descending) sort and then click again to unapply the sort on the column.  This is the default behavior of the react-table library, which is easier to keep for the time being than try to prevent the user from unapplying sort.

All columns should be sortable except for the Origin GBLOC column, since for most TOOs these will all have the same value unless they are the Marine Corps TOO.

I also needed to add logic to sort the destination duty station by name once I noticed it was off in the backend PR.  There was ambiguity and we were actually sorting by the origin duty station name instead.

There is a new storybook story with sorting enabled called TOO/TIO Components -> Table -> TXO Table Sortable

## Reviewer Notes

The border styling is not perfectly aligned and does cause some shifting of the row below.  This is due to the table display styling of cell data.  We are also collapsing all borders of the table so that is part of the reason for the overlap as well. I tried a few different ways to wrap the cell or row but couldn't find a working solution.  Hopefully we can clean up this styling in a subsequent PR.

The jest react tests were giving me a too many nested update levels deep error.  I think I've resolved that by turning off the autoResetSortBy flag on the react-table so should be now able to write some client side tests.

## Setup

```sh
make db_dev_e2e_populate
make server_run
make office_client_run
```

1. Login as the TOO user to the office app
2. Initial sorting should be status ascending
3. Toggle the status column to descending and unsorted
4. Sort the other columns, Origin GBLOC should do nothing when clicked
5. Try applying the sort along with filters and pagination

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4060) for this change

## Screenshots
Screen Recording 2020-11-23 at 8 59 13 AM mov


brooksr4 (approved):
Looks good!

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -67,8 +67,10 @@ func (f moveOrderFetcher) ListMoveOrders(officeUserID uuid.UUID, params *service
 	).InnerJoin("orders", "orders.id = moves.orders_id").
 		InnerJoin("service_members", "orders.service_member_id = service_members.id").
 		InnerJoin("mto_shipments", "moves.id = mto_shipments.move_id").
-		InnerJoin("duty_stations", "orders.origin_duty_station_id = duty_stations.id").
-		InnerJoin("transportation_offices", "duty_stations.transportation_office_id = transportation_offices.id").
+		InnerJoin("duty_stations as origin_ds", "orders.origin_duty_station_id = origin_ds.id").
+		InnerJoin("transportation_offices as origin_to", "origin_ds.transportation_office_id = origin_to.id").
+		LeftJoin("duty_stations as dest_ds", "dest_ds.id = orders.new_duty_station_id").
+		LeftJoin("transportation_offices as dest_to", "dest_to.id = dest_ds.id").
```
monfresh, line 73: I don't see `dest_to` being used. Is this something we will need in the future?

duncan-truss: Oh good point, it was late on Friday when I added that I think we can remove that join!

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -193,7 +201,7 @@ func destinationDutyStationFilter(destinationDutyStation *string) QueryOption {
 	return func(query *pop.Query) {
 		if destinationDutyStation != nil {
 			nameSearch := fmt.Sprintf("%s%%", *destinationDutyStation)
-			query = query.InnerJoin("duty_stations as destination_duty_station", "orders.new_duty_station_id = destination_duty_station.id").Where("destination_duty_station.name ILIKE ?", nameSearch)
+			query = query.Where("dest_ds.name ILIKE ?", nameSearch)
```
monfresh, line 204: Nice.

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -93,7 +95,13 @@ func (f moveOrderFetcher) ListMoveOrders(officeUserID uuid.UUID, params *service
...
 	}
 
-	err = query.GroupBy("moves.id, service_members.id, orders.id, duty_stations.id").Paginate(int(*params.Page), int(*params.PerPage)).All(&moves)
+	var groupByColumms []string
+	groupByColumms = append(groupByColumms, "service_members.id", "orders.id", "origin_ds.id")
+	if params.Sort != nil && *params.Sort == "destinationDutyStation" {
+		groupByColumms = append(groupByColumms, "dest_ds.name")
+	}
+
+	err = query.GroupBy("moves.id", groupByColumms...).Paginate(int(*params.Page), int(*params.PerPage)).All(&moves)
```
monfresh, line 104: I like the conditional grouping!

src/components/Table/Table.module.scss
```diff
@@ -21,22 +21,35 @@
...
+      @include u-margin-top('05');
+      @include u-margin-bottom('05');
+    }
+
     th {
       color: $base-darker;
       @include u-text('bold');
+
+      &.sortAscending {
+        @include u-border-top('05', 'black');
```
monfresh, line 47: Is there a rule we follow for when to spell out color names like this versus using an existing color variable?

pkg/services/move_order/move_order_fetcher.go
duncan-truss: Pop overwrites the GroupBy if you call it multiple times I found.  And you can't call `query.GroupBy(groupByColumns...)` because the first arg is a string followed by the variadiac.  I guess we could change it to `query.GroupBy(groupByColumns[0], groupByColumns[1:]...)` but we are querying moves so I thought that wasn't a big deal.

src/components/Table/Table.module.scss
duncan-truss: I think for the uswds helpers you can't pass a SASS variable like $custom-color so in those cases I usually default to ditching the helper and write it out explicitly.  In this case since we are using a predefined color 'black' I thought it was easiest to use the spacing token '05'.  If it was a different color maybe I would have broken up the border-top-color into it's own definition.

src/components/Table/TableQueue.jsx
```diff
@@ -9,19 +9,42 @@ import Table from 'components/Table/Table';
...
+  disableMultiSort,
+  defaultCanSort,
+  disableSortBy,
+  defaultSortedColumns,
+  handleClick,
+  useQueries,
+  showFilters,
+  showPagination,
+}) => {
+  // eslint-disable-next-line no-unused-vars
```
monfresh, line 28: What is eslint complaining about here? From what I can tell, both `paramSort` and `setParamSort` are being used.

duncan-truss: That can be deleted now! That was a placeholder until I could integrate fully with the api call.

monfresh (approved):
LGTM

PR #5309, opened 2020-11-23
[MB-5578] Edit profile expected behavior

Ronolibert:
## Description

Previously, when you Edit Profile and save, it would revert the client to a stale state. Saving Edit Profile now have the expected behavior. `Orders` were being loaded but auxiliary pieces of data, that we need, like `moves` and `uploads` were not. This loads the correct data

## Setup
`make server_run`
`make client_run`
1. Set up a service member
2. Continue through flow and set up entire move
3. Edit Profile
4. Save
Expected behavior: Home page should be in the correct state

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5578) for this change


lynzt (approved):
works as expected 🚢

PR #5325, opened 2020-11-24
[MB 5474] set sent to gex timestamp in ProcessReviewedPaymentRequest service

pearl-truss:
## Description

Update the ProcessReviewedPaymentRequest service so that it updates/adds a sent_to_gex_at timestamp to payment request that have been successfully processed

## Setup

Be sure to start by running `make db_dev_e2e_populate` and then `make server_run` 

In a database viewer run this query:

`select * from payment_requests where id='cfd110d4-1f62-401c-a92c-39987a0b4228';`

Confirm that the status is REVIEWED and the sent_to_gex_at field is nil

Then run the process reviewed payment request support api to test sending to syncada (this will call the ProcessReviewedPaymentRequest service that is now updating the sent_to_gex_at field)

JSON
```json
{
  "body": {
    "sendToSyncada": true
  }
}
```

Command
```sh
prime-api-client --insecure support-reviewed-payment-requests --filename <filename.json>
```

Search the database for the updated payment request

`select * from payment_requests where id='cfd110d4-1f62-401c-a92c-39987a0b4228';`

Confirm that the `sent_to_gex_at` field is populated

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5474) for this change


pkg/services/payment_request/payment_request_reviewed_processor_test.go
```diff
@@ -198,11 +200,18 @@ func (suite *PaymentRequestServiceSuite) TestProcessReviewedPaymentRequest() {
 			SFTPSession)
 		err := paymentRequestReviewedProcessor.ProcessReviewedPaymentRequest()
 		suite.NoError(err)
+
+		// Ensure that sent_to_gex_at timestamp has been added
+		fetcher := NewPaymentRequestFetcher(suite.DB())
+		for _, pr := range prs {
+			paymentRequest, _ := fetcher.FetchPaymentRequest(pr.ID)
+			suite.NotNil(paymentRequest.SentToGexAt)
```
gidjin, line 208: Feels like it's worth it to also check that the time stamp in `SentToGexAt` is not a zero time, since it's a pointer it can be nil or zero value both which would be invalid. https://golang.org/pkg/time/#Time.IsZero

pkg/services/payment_request/payment_request_reviewed_processor_test.go
```diff
@@ -20,9 +20,9 @@ import (
 
 const testDateFormat = "060102"
 
-func (suite *PaymentRequestServiceSuite) createPaymentRequest(num int) {
-
-	for i := 0; i < 4; i++ {
+func (suite *PaymentRequestServiceSuite) createPaymentRequest(num int) models.PaymentRequests {
+	var paymentRequests models.PaymentRequests
+	for i := 0; i < num; i++ {
```
gidjin, line 25: Nice catch, looks like this was unused before?

gidjin (approved):
Code looks good and works as expected. I left one minor comment on the tests, but I don't need to re-review it.

pkg/services/payment_request/payment_request_reviewed_processor_test.go
pearl-truss: It was used in some of the other test in this suite

PR #5329, opened 2020-11-24
MB-5359 Create DOASIT and DOPSIT when creating DOFSIT service item

abbyoung:
## Description

When the Prime created a DOFSIT Service Item, a DOASIT and DOPSIT service item should also be created. This PR adds the necessary validation to the CreateMTOServiceItem service object, and adds test coverage.

## Setup

The server tests have been updated to add test cases for this functionality.
`go test ./pkg/services/mto_service_item --testify.m "TestCreateOriginSITServiceItem" -v`

You can also create a DOFSIT Service Item using Postman.
* You must first make the associated move APPROVED in the db.
* The DOASIT and DOPSIT will be returned with the json, but some fields may be null (addressed in future stories).

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5259) for this change

### DIFF
--- a/pkg/services/mto_service_item/mto_service_item_creator.go
+++ b/pkg/services/mto_service_item/mto_service_item_creator.go
@@ -110,13 +110,6 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
 		}
 	}
 
-	if serviceItem.ReService.Code == models.ReServiceCodeDOPSIT {
-		verrs = validate.NewErrors()
-		verrs.Add("reServiceCode", fmt.Sprintf("%s cannot be created", serviceItem.ReService.Code))
-		return nil, nil, services.NewInvalidInputError(serviceItem.ID, nil, verrs,
-			fmt.Sprintf("A service item with reServiceCode %s cannot be manually created.", serviceItem.ReService.Code))
-	}
-
 	for index := range serviceItem.CustomerContacts {
 		createCustContacts := &serviceItem.CustomerContacts[index]
 		err = validateTimeMilitaryField(createCustContacts.TimeMilitary)
@@ -125,7 +118,7 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
 		}
 	}
 
-	if serviceItem.ReService.Code == models.ReServiceCodeDDDSIT {
+	if serviceItem.ReService.Code == models.ReServiceCodeDDDSIT || serviceItem.ReService.Code == models.ReServiceCodeDOPSIT {
 		verrs = validate.NewErrors()
 		verrs.Add("reServiceCode", fmt.Sprintf("%s cannot be created", serviceItem.ReService.Code))
 		return nil, nil, services.NewInvalidInputError(serviceItem.ID, nil, verrs,
--- a/pkg/services/mto_service_item/mto_service_item_creator.go
+++ b/pkg/services/mto_service_item/mto_service_item_creator.go
@@ -110,16 +110,6 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
 		}
 	}
 
-	if serviceItem.ReService.Code == models.ReServiceCodeDOFSIT {
-		extraServiceItems, errSIT := o.validateFirstDaySITServiceItem(serviceItem)
-		if errSIT != nil {
-			return nil, nil, errSIT
-		}
-
-		requestedServiceItems = append(requestedServiceItems, *extraServiceItems...)
-
-	}
-
 	if serviceItem.ReService.Code == models.ReServiceCodeDOPSIT {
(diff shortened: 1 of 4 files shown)


pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -345,10 +362,15 @@ func (o *mtoServiceItemCreator) validateFirstDaySITServiceItem(serviceItem *mode
 
 	// create the extra service items for first day SIT
 	var reServiceCodes []models.ReServiceCode
-	if serviceItem.ReService.Code == models.ReServiceCodeDDFSIT {
+
+	switch serviceItem.ReService.Code {
```
abbyoung, line 366: @sandy-wright I changed this from an if/else to a switch, to cover any case where the `validateFirstDaySITServiceItem` fn may be called incorrectly and we don't want to return any items. Does this make sense? Also, is the error returned correct, or should it be something else?

sandy-wright: The switch looks good! I think NotFound might not make sense here... Maybe InvalidInput? Because the reServiceCode set on the given service item wasn't right somehow

pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -110,6 +110,23 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
...
+		extraServiceItems, errSIT := o.validateFirstDaySITServiceItem(serviceItem)
+		if errSIT != nil {
+			return nil, nil, errSIT
+		}
+
+		requestedServiceItems = append(requestedServiceItems, *extraServiceItems...)
+
+	}
+
+	if serviceItem.ReService.Code == models.ReServiceCodeDOPSIT {
```
sandy-wright, line 123: You could combine this if statement with the DDDSIT one on line 138:

`if serviceItem.ReService.Code == models.ReServiceCodeDOPSIT || serviceItem.ReService.Code == models.ReServiceCodeDDDSIT` ... 

which should work without any other changes since the error message doesn't have the ReServiceCode hardcoded into it 😄

sandy-wright (commented):
Looks good! I added a couple of comments/suggestions, and I think to get the server tests to pass you'll have to add a `MakeDOFSITReServiceCode` function like I did to use in other tests. 

The issue is that with the way it works in the service object now, the DOASIT and DOPSIT service codes need to exist in the database whenever the DOFSIT is created, but tests in other packages don't add those codes to the test DB beforehand. So I made that function for DDFSIT to bundle the creation of all three and make it a little simpler for those tests. Let me know if you'd like to pair on that!

pkg/services/mto_service_item/mto_service_item_creator.go
abbyoung: that was my other choice! 😄 thank you!

pkg/services/mto_service_item/mto_service_item_creator.go
abbyoung: b r i l l i a n t. thanks for catching that!

abbyoung: (altho i think you meant DDFSIT?)

pkg/testdatagen/make_re_service.go
```diff
@@ -74,3 +74,32 @@ func MakeDDFSITReService(db *pop.Connection) models.ReService {
...
 }
+
+// MakeDOFSITReService creates the three origin SIT service codes: DOFSIT, DOPSIT, DOASIT. Returns DOFSIT only.
+func MakeDOFSITReService(db *pop.Connection, assertions Assertions) models.ReService {
+	assertionsDOFSIT := Assertions{
+		ReService: models.ReService{
+			Code: models.ReServiceCodeDOFSIT,
+		},
+	}
+	// Any assertions passed in should be applied to the DOFSIT only
```
sandy-wright, line 85: oh this is better than what I did! I'll make a teeny ticket to go back and update that one later

sandy-wright: (the DDFSIT one)

pkg/services/mto_service_item/mto_service_item_creator.go
sandy-wright: oh well heck, I meant both actually! Would ya look at that. DDDSIT and DOPSIT are both the departure SITs, and the logic for not being able to update them is the same. The DDDSIT one is on line 128

sandy-wright (approved):
LGTM! One small comment about combing the DDSIT + DOPSIT line, but it looks good to go either way 😄

pkg/services/mto_service_item/mto_service_item_creator.go
abbyoung: OH! I SEE. On it! Thank you!

PR #5331, opened 2020-11-24
MB-4829 Include dependents in weight allowance

monfresh:
## Description

Previously, the GHC API payload, when returning an order, was setting
the `totalWeight` key to the weight allowance based only on the
service member's rank, regardless of whether or not they have
dependents.

This PR fixes the bug in the simplest way as a first pass, by adding
a conditional to check if the entitlement includes dependents, and if
so, it uses the `TotalWeightSelfPlusDependents` key in the lookup
table in `entitlements.go`. Otherwise, it uses `TotalWeightSelfPlus`
as before.

I added 2 handler tests for both scenarios.

## Reviewer Notes

1. Log in as a TOO locally
2. Click on the first move.
3. Assuming it shows the service member rank as E-1, the weight allowance should show 8000 lbs.

Note that the authorized weight is still wrong. It should also be 8000 lbs (they should both the same initially). That bug is being fixed in a [separate JIRA story](https://dp3.atlassian.net/browse/MB-4830).

## Setup

```sh
make db_dev_e2e_populate
make office_client_run (in a separate tab)
```

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4829)

## Screenshots

**Before:**
Screen Shot 2020-11-23 at 3 09 25 PM

**After:**
Screen Shot 2020-11-24 at 3 09 29 PM

### DIFF
--- a/pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
+++ b/pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
@@ -134,7 +134,11 @@ func Entitlement(entitlement *models.Entitlement) *ghcmessages.Entitlements {
 	if entitlement.WeightAllotment() != nil {
 		proGearWeight = int64(entitlement.WeightAllotment().ProGearWeight)
 		proGearWeightSpouse = int64(entitlement.WeightAllotment().ProGearWeightSpouse)
-		totalWeight = int64(entitlement.WeightAllotment().TotalWeightSelf)
+		if *entitlement.DependentsAuthorized {
+			totalWeight = int64(entitlement.WeightAllotment().TotalWeightSelfPlusDependents)
+		} else {
+			totalWeight = int64(entitlement.WeightAllotment().TotalWeightSelf)
+		}
 	}
 	var authorizedWeight *int64
 	if entitlement.AuthorizedWeight() != nil {
--- a/pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
+++ b/pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
@@ -131,13 +131,13 @@ func Entitlement(entitlement *models.Entitlement) *ghcmessages.Entitlements {
 		return nil
 	}
 	var proGearWeight, proGearWeightSpouse, totalWeight int64
-	if entitlement.WeightAllotment() != nil {
-		proGearWeight = int64(entitlement.WeightAllotment().ProGearWeight)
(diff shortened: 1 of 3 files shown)


pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -134,7 +134,11 @@ func Entitlement(entitlement *models.Entitlement) *ghcmessages.Entitlements {
 	if entitlement.WeightAllotment() != nil {
 		proGearWeight = int64(entitlement.WeightAllotment().ProGearWeight)
 		proGearWeightSpouse = int64(entitlement.WeightAllotment().ProGearWeightSpouse)
-		totalWeight = int64(entitlement.WeightAllotment().TotalWeightSelf)
+		if *entitlement.DependentsAuthorized {
+			totalWeight = int64(entitlement.WeightAllotment().TotalWeightSelfPlusDependents)
+		} else {
+			totalWeight = int64(entitlement.WeightAllotment().TotalWeightSelf)
+		}
```
duncan-truss, line 141: since we're calling entitlement.WeightAllotment() ~3 times what do you think about making the top conditional:
```
if weightAllotment := entitlement.WeightAllotment(); weightAllotment != nil {
    ...
}
```

monfresh: Sounds good. Hopefully, we'll refactor this later so that we would save the value in the DB, and just call something like `entitlement.WeightAllowance`.

monfresh: Done!

pkg/testdatagen/make_entitlement.go
```diff
@@ -33,3 +34,11 @@ func MakeEntitlement(db *pop.Connection, assertions Assertions) models.Entitleme
 
 	return entitlement
 }
+
+func setDependentsAuthorized(assertionDependentsAuthorized *bool) *bool {
+	dependentsAuthorized := swag.Bool(true)
```
duncan-truss, line 39: I'm just curious why the default is true if no assertion is specified?

duncan-truss (approved):

pkg/testdatagen/make_entitlement.go
monfresh: That's a great question, one that I wondered about as well. That's how it was before, so I left it like that. I didn't want to go down the path of seeing what would break if we made the default false.

PR #5332, opened 2020-11-24
Remediate staticcheck SA1017

kimallen:
## Description

The error for staticcheck SA1017 says "the channel used with signal.Notify should be buffered".  This adds a buffer of 1.

## Reviewer Notes

https://golang.org/pkg/os/signal/#Notify
The os/signal package uses non-blocking channel sends when delivering signals. If the receiving end of the channel isn't ready and the channel is either unbuffered or full, the signal will be dropped. To avoid missing signals, the channel should be buffered and of the appropriate size. For a channel used for notification of just one signal value, a buffer of size 1 is sufficient.

## Setup

I'm not sure how to test this.  Maybe @shimonacarvalho can offer me assistance?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5454) for this change
* [this article](https://golang.org/pkg/os/signal/#Notify) explains more about the approach used.


cmd/webhook-client/db_webhook_notify.go
```diff
@@ -73,7 +73,7 @@ func dbWebhookNotify(cmd *cobra.Command, args []string) error {
 
 	// Wait for interrupt signal to gracefully shutdown the server with
 	// a timeout of 5 seconds.
-	quit := make(chan os.Signal)
+	quit := make(chan os.Signal, 1)
```
sojeri, line 76: it matches your docs, but I'd love to be a fly on the wall if you pair on testing this with @shimonacarvalho! I haven't worked much with buffering data in go so I'd love to learn more. :)

shimonacarvalho (approved):
Looks good to me!

PR #5334, opened 2020-11-24
Ignore dangerjs if acceptable rule

kimallen:
## Description

This adds to the dangerjs trigger for linter bypass changes.  It does not trigger a warning if the code adds certain disable rules.

## Reviewer Notes

ok eslint rules:
    'no-underscore-dangle',
    'prefer-object-spread',
    'object-shorthand',
    'camelcase',
    'react/jsx-props-no-spreading',
    'react/destructuring-assignment',
    'react/forbid-prop-types',
    'react/prefer-stateless-function',
    'react/sort-comp',
    'import/no-extraneous-dependencies',
    'import/order',
    'import/prefer-default-export',
    'import/no-named-as-default'

~This PR does not cover every scenario.  For example, it's possible to disable multiple rules in one disable line, and the rules after the first would not get checked.~ This now covers multiple rules in one disable line

The prior dangerjs version was triggering even when disabling code was removed (subtracted diff).  This should check only code that has been added or modified.

## Setup

It's a bit challenging to test this, since the diff includes strings I'm telling it to trigger a warning.  You can add a commit that contains an eslint disable line (acceptable rule or unacceptable rule)  Then remove #nosec from the `bypassCodes` array in your local file, and then dangerjs won't get triggered by #nosec in the file.  Run the below line to verify expected result.

Run `yarn danger pr https://github.com/transcom/mymove/pull/5334`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5683) for this change


dangerfile.ts
```diff
@@ -67,20 +67,39 @@ View the [frontend file org ADR](https://github.com/transcom/mymove/blob/master/
...
+    'destructuring-assignment',
+    'forbid-prop-types',
+    'prefer-stateless-function',
+    'sort-comp',
+    'import/no-extraneous-dependencies',
+    'order',
+    'prefer-default-export',
+    'no-named-as-default',
+  ];
   let addedByPassCode = false;
```
sojeri, line 90: I know this is still a draft, but can you update all the different bypass variables to use consistent casing? you don't have to follow this exact naming, but something like:
- `addedByPassCode` => `isBypassFound`
- `diffsWithbypassCodes` => `bypassDiffs`

dangerfile.ts
```diff
@@ -67,20 +67,39 @@ View the [frontend file org ADR](https://github.com/transcom/mymove/blob/master/
     warn(`${message} - <i>${idea}</i>`);
   }
 };
-
+// eslint-disable jsx-no-target-blank test
 const bypassingLinterChecks = async () => {
   // load all modified and new files
   const allFiles = danger.git.modified_files.concat(danger.git.created_files);
-  const bypassCodes = ['#nosec', 'eslint-disable', 'eslint-disable-next-line'];
+  const bypassCodes = ['eslint-disable', 'eslint-disable-next-line', '#nosec'];
```
sojeri, line 74: doesn't have to be changed in this PR, but I'm thinking we should add a story to fail if eslint-disable is found outside of eslint-disable-next-line, because disabling the whole file can hide new issues.

kimallen: Sometimes `eslint-disable` is used to disable the same line.  So we'd need to check if it was at the top of the file, somehow

sojeri: I think any plain eslint-disable is for the whole file regardless of where it's placed--- even when scoped to a single type, it's still the whole file.

kimallen: Ok, I see. I just check.  disabling the same line uses `eslint-disable-line`, which we are missing from the list of bypassCodes.  I'll add that.

sojeri (changes requested):
I like where this is going, but there's a few issues in the current iteration which will either cause bugs or allow folks to slip through these new checks. I've left more detailed comments about these issues on relevant lines of code. :)

dangerfile.ts
```diff
@@ -71,20 +71,57 @@ View the [frontend file org ADR](https://github.com/transcom/mymove/blob/master/
...
+  ];
+  let hasBypass = false;
   const diffs = await Promise.all(allFiles.map((f) => danger.git.diffForFile(f)));
 
   for (let i = 0; i < diffs.length; i += 1) {
     const diff = diffs[Number(i)];
-    const diffsWithbypassCodes = bypassCodes.find((b) => diff.diff.includes(b));
-    if (diffsWithbypassCodes) {
-      addedByPassCode = true;
+    const bypassCodesInDiff = bypassCodes.map((b) => {
```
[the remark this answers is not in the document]
kimallen: From what I understand, `find` returns only the first encounter that meets the callback criteria, and we need to know all bypass codes that occur in the diff.  I also would expect that map would only return the bypass codes of the diff that exist (returns a new array of the returned values).  It's definitely possible I'm thinking about it wrong.  Might be helpful to walk through it together.

kimallen:
I'm honestly wondering if it makes more sense, rather than spending a bunch of time on this, to just list the acceptable rules in the dangerjs message while it's just a warning, and then people can determine if they are in violation or not.

dangerfile.ts
```diff
@@ -71,20 +71,57 @@ View the [frontend file org ADR](https://github.com/transcom/mymove/blob/master/
...
+      if (diff.diff.includes(b)) {
+        return b;
+      }
+    });
+    if (bypassCodesInDiff.length > 0) {
+      for (const code in bypassCodesInDiff) {
+        const codeValue = bypassCodesInDiff[code];
+        if (codeValue !== '#nosec') {
+          for (const rule in okBypassRules) {
+            if (diff.diff.includes(codeValue + ' ' + okBypassRules[`${rule}`])) {
```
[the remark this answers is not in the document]
kimallen: Now splitting on commas and checking against rules.

dangerfile.ts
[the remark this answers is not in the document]
kimallen: Walked through- I now understand the issue.

dangerfile.ts
```diff
@@ -69,28 +69,129 @@ View the [frontend file org ADR](https://github.com/transcom/mymove/blob/master/
...
+    prohibitedOverrideMsg = 'Must specify the rule you are disabling';
+    return prohibitedOverrideMsg;
+  }
+
+  // rules format: ['no-jsx', 'no-default']
+  let rules = disablingStringParts.slice(1);
+  for (let r in rules) {
+    const rule = rules[r];
+    if (!okBypassRules.includes(rule)) {
+      prohibitedOverrideMsg = `Contains a rule that is not in the permitted eslint list. Ok rules: (\n${okBypassRules.map(
```
sojeri, line 187:
```suggestion
      prohibitedOverrideMsg = `Contains a rule that is not in the permitted eslint list. You are free to disable only: (\n${okBypassRules.map(
```

dangerfile.ts
```diff
@@ -69,28 +69,129 @@ View the [frontend file org ADR](https://github.com/transcom/mymove/blob/master/
...
+  // split along commas and/or spaces and remove surrounding spaces
+  let disablingStringParts = disablingString
+    .trim()
+    .split(/[\s,]+/)
+    .map((item) => item.trim());
+  // disablingStringParts format: ['eslint-disable-next-line', 'no-jsx', 'no-default']
+  if (disablingStringParts[0] === 'eslint-disable') {
+    // fail because don't disable whole file please!
+    prohibitedOverrideMsg =
+      'Found "eslint-disable": this disables the whole file. Please specify rules if disable is permitted.';
```
sojeri, line 173:
```suggestion
      'Found `eslint-disable`. This disables the whole file, which is bad practice because it can allow security issues to slip in unnoticed. Please specify exact rules on a line by line basis, eg `eslint-disable-next-line no-underscore-dangle`.';
```

dangerfile.ts
```diff
@@ -69,28 +69,129 @@ View the [frontend file org ADR](https://github.com/transcom/mymove/blob/master/
...
+  const dangerMsgSegment = checkPRHasProhibitedLinterOverride(diffsByFile);
+  if (dangerMsgSegment) {
+    warn(
+      `It looks like you are attempting to bypass a linter rule, which is not within
+      security compliance rules.\n** ${dangerMsgSegment} **\n Please remove the bypass code and address the underlying issue. cc: @transcom/Truss-Pamplemoose`,
+    );
+  }
+};
+
+// fn for bypassingLinterChecks
```
sojeri, line 83: nitpick: can you remove or clarify these `fn for bypassingLinterChecks` comments? if you want to keep this context, I'd recommend creating a new file `bypassingLinterChecks.ts` with all relevant code & then import just `bypassingLinterChecks`.

dangerfile.ts
```diff
@@ -69,28 +69,129 @@ View the [frontend file org ADR](https://github.com/transcom/mymove/blob/master/
...
-  for (let i = 0; i < diffs.length; i += 1) {
-    const diff = diffs[Number(i)];
-    const diffsWithbypassCodes = bypassCodes.find((b) => diff.diff.includes(b));
-    if (diffsWithbypassCodes) {
-      addedByPassCode = true;
+  const diffsByFile = await Promise.all(allFiles.map((f) => danger.git.diffForFile(f)));
+  const dangerMsgSegment = checkPRHasProhibitedLinterOverride(diffsByFile);
+  if (dangerMsgSegment) {
+    warn(
+      `It looks like you are attempting to bypass a linter rule, which is not within
```
sojeri, line 77: # 📨 📬
I LOVE THESE MESSAGES. They are so helpful and so much more specific than before! I think these will be really helpful to eng who get the dangerjs warnings.

dangerfile.ts
```diff
@@ -69,28 +69,129 @@ View the [frontend file org ADR](https://github.com/transcom/mymove/blob/master/
...
+      security compliance rules.\n** ${dangerMsgSegment} **\n Please remove the bypass code and address the underlying issue. cc: @transcom/Truss-Pamplemoose`,
+    );
+  }
+};
+
+// fn for bypassingLinterChecks
+function checkPRHasProhibitedLinterOverride(dangerJSDiffCollection) {
```
sojeri, line 84: if & only if it's not a huge chunk of work (maybe <15min?), can you update the signature of this fn to return the file name as well? if that's something we can do, it will make reviewing relevant changes easier. I'm thinking of this because a couple days ago the old version of this code pulled me into a 90-file review. having the offending file name would have made that a lot easier -- both for reviewer & the dev who first reads the dangerjs warning.

if it's a large amount of work, I can create a chore for it and we can do it later.

sojeri (approved):
❇️   ,,  ❇️ 
\\______/

I am so happy about where this ended up. I know [you had said](https://github.com/transcom/mymove/pull/5334#issuecomment-736131924) it was a lot of work, but I think this will be really helpful to MilMove eng AND it will save security reviewers a lot of time. Thanks for completely rewriting this from scratch after posting your first iteration. 🖤

I left a few small comments, but they are not blocking changes. Two of them are copy suggestions, so if you agree with my proposed language you can click them to commit directly to the branch. :)

dangerfile.ts
sojeri: also if you do this & want a new review of those changes--- please rerequest my review or at mention me. either should trigger a github notification for me. :)

dangerfile.ts
```diff
@@ -71,20 +71,57 @@ View the [frontend file org ADR](https://github.com/transcom/mymove/blob/master/
...
     const diff = diffs[Number(i)];
-    const diffsWithbypassCodes = bypassCodes.find((b) => diff.diff.includes(b));
-    if (diffsWithbypassCodes) {
-      addedByPassCode = true;
+    const bypassCodesInDiff = bypassCodes.map((b) => {
+      if (diff.diff.includes(b)) {
+        return b;
+      }
+    });
+    if (bypassCodesInDiff.length > 0) {
```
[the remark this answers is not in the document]
kimallen: resolving because no longer relevant due to giant refactor. . .

dangerfile.ts
```diff
@@ -71,20 +71,57 @@ View the [frontend file org ADR](https://github.com/transcom/mymove/blob/master/
...
+      for (const code in bypassCodesInDiff) {
+        const codeValue = bypassCodesInDiff[code];
+        if (codeValue !== '#nosec') {
+          for (const rule in okBypassRules) {
+            if (diff.diff.includes(codeValue + ' ' + okBypassRules[`${rule}`])) {
+              hasBypass = false;
+              break;
```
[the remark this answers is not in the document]
kimallen: no longer relevant-- refactored

dangerfile.ts
kimallen: aah, yes, that would be a better way.  I was feeling like all my functions were making the dangerjs file really cluttered and harder to read, hence the clarifying comments.

kimallen: I like the idea of putting stuff into a new file, but also don't want to clutter the project root with it.  Anywhere else you see it being a fitting place for it?

sojeri: ah, yeah. you could add a new danger/ subdir like we have cypress?

dangerfile.ts
kimallen: Ok, I spent more time than I should have, and ran into a blocker, though I mostly know what I'd do to implement it.  We should create a story for it.

dangerfile.ts
```diff
@@ -71,20 +71,57 @@ View the [frontend file org ADR](https://github.com/transcom/mymove/blob/master/
...
+            }
+          }
+        } else {
+          hasBypass = true;
+          break;
+        }
+        if (hasBypass) {
```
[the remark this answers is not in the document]
kimallen: outdated with refactor

PR #5335, opened 2020-11-24
[MB-5712] Dont exclude potential hhg ppm combo move in the move order fetcher

duncan-truss:
## Description

The TOO moves queue is supposed to exclude PPM moves but the change we made also excluded a combo HHG and PPM move when you create the PPM last in the customer flow.

We were relying on the value of the `selected_move_type` in the moves table to indicate when a move was an HHG or a PPM but this was inaccurate because it could end up with the value of PPM on a combo move.

We are already joining to the mto_shipments table, which should exclude moves that are only PPM but we have to remove the PPM selected_move_type from the exclusion list because it could be a combo move with an HHG. 

## Reviewer Notes

## Setup

```sh
make db_dev_e2e_populate
make server_run
make client_run
```

1. Create a new service member at milmovelocal:3000
2. Complete the user profile with an origin duty station in the 'LKNQ' GBLOC such as Yuma AFB
2. Next create an HHG shipment
3. Add another shipment of type PPM
4. Sign and submit your move
5. Login as the TOO user and the move should now appear in the queue

I've also changed the combo HHG+PPM move in the devseed test data to use the PPM SelectedMoveType so you can verify it appears in the TOO queue still.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5712) for this change

## Screenshots

Screen Shot 2020-11-25 at 10 24 02 AM

### DIFF
--- a/pkg/services/move_order/move_order_fetcher_test.go
+++ b/pkg/services/move_order/move_order_fetcher_test.go
@@ -3,6 +3,8 @@ package moveorder
 import (
 	"testing"
 
+	"github.com/gofrs/uuid"
+
 	"github.com/transcom/mymove/pkg/services"
 
 	"github.com/go-openapi/swag"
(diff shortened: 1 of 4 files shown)


pkg/testdatagen/make_move.go
```diff
@@ -186,6 +186,10 @@ func MakeHHGPPMMoveWithShipment(db *pop.Connection, assertions Assertions) model
 		Stub: assertions.Stub,
 	})
 
+	if !assertions.Stub {
+		mustSave(db, &move)
+	}
+
```
monfresh, line 192: Why is this necessary?

pkg/services/move_order/move_order_fetcher_test.go
```diff
@@ -68,33 +68,40 @@ func (suite *MoveOrderServiceSuite) TestListMoves() {
...
-		suite.Equal(expectedMove.Orders.NewDutyStationID, move.Orders.NewDutyStation.ID)
-		suite.NotNil(move.Orders.Entitlement)
-		suite.Equal(*expectedMove.Orders.EntitlementID, move.Orders.Entitlement.ID)
-		suite.Equal(expectedMove.Orders.OriginDutyStation.ID, move.Orders.OriginDutyStation.ID)
-		suite.NotNil(move.Orders.OriginDutyStation)
-		suite.Equal(expectedMove.Orders.OriginDutyStation.AddressID, move.Orders.OriginDutyStation.AddressID)
-		suite.Equal(expectedMove.Orders.OriginDutyStation.Address.StreetAddress1, move.Orders.OriginDutyStation.Address.StreetAddress1)
+		suite.Equal(2, moveCount)
+
+		for i, move := range moves {
```
monfresh, line 88: Are we confident that the order will always be the same? Also, I think that with the current pagination params, moves length will only be 1. I think we should remove pagination from our tests unless we're specifically testing it. It has caused confusion during test failures.

monfresh: Maybe it would be better to have a separate test to make sure combo moves are included. That test would only test that the moveCount is correct since all we're trying to verify is that certain types of moves are returned. I'm not sure what value is provided by testing the model attributes unless we're specifically testing ordering, which I think we have a separate test for. If there are only 2 entries created in the DB, and 2 are returned, doesn't that imply that we got the correct entries?

pkg/testdatagen/make_move.go
duncan-truss: 🤦 Yesterday was a long day

pkg/services/move_order/move_order_fetcher_test.go
duncan-truss: No I wasn't confident about the order and was likely going to specify an explicit sort order.  Happy to break it out into it's own test.

pkg/services/move_order/move_order_fetcher_test.go
```diff
@@ -121,6 +124,21 @@ func (suite *MoveOrderServiceSuite) TestListMoves() {
...
+
+		suite.FatalNoError(err)
+		suite.Equal(2, moveCount)
+		suite.Len(moves, 2)
+
+		moveIDs := []uuid.UUID{moves[0].ID, moves[1].ID}
+
+		suite.Contains(moveIDs, expectedComboMove.ID)
+	})
+
```
monfresh, line 141: Yay!

monfresh (approved):
LGTM

PR #5336, opened 2020-11-24
MB-5591 Create the definition of the updateMTOServiceItem endpoint

shimonacarvalho:
## Description

Added the UpdateMtoServiceItem endpoint to the yaml. Rather than make an endpoints specifically to update departureDate, we are making a generic endpoint to update MTO service items. 

Currently it only returns NotImplemented error. Goal for next tix is that it only updates departure date on SITs but we do believe that there will be more service items that will need updates which is why we are doing it this way. 
Functionality will be limited to what Prime is ALLOWED to update. 

I found that we needed to make a separate parent/child definition in the yaml called updateMTOServiceItemxxx because the polymorphism makes it impossible to easily share the definitions between the creation and update endpoints.

I also updated the documentation to make the redoc clearer on what each endpoint allows/supports. 

## Reviewer Notes

Please check the redoc to review the result. 
Also try to hit with Postman, it should return the 501/Not Implemented.

Also created a test for the NotImplemented response, which can be run with 
`go test ./pkg/handlers/primeapi/ --testify.m TestUpdateMTOServiceItemHandler -v`


swagger/prime.yaml
```diff
@@ -367,12 +367,19 @@ paths:
...
+        This endpoint supports different body definitions. In the modelType field below, select the modelType corresponding
+         to the service item you wish to create and the documentation will update with the new definition.
+
+        Upon creation these items are associated with a Move Task Order and an MTO Shipment.
+        The request must include UUIDs for the MTO and MTO Shipment connected to this service item. Some service item types require
         additional service items to be autogenerated when added - all created service items, autogenerated included,
         will be returned in the response.
+
+        To update a service item, please use [updateMTOServiceItem](#operation/updateMTOServiceItem) endpoint.
+
```
sandy-wright, line 382: Thanks for updating this description too!

sandy-wright (approved):
LGTM, thanks! ✨

PR #5350, opened 2020-11-25
[MB-4237] Homepage move code

Ronolibert:
## Description

Adds the move code to then homepage and does some slight styling to accommodate for mobile and desktop

## Reviewer Notes
- I wanted to refactor a bit, but there is also a CustomerHeader component that I think we could make use of in the near future

## Setup
`make server_run`
`make client_run`
1. Log in
2. Set up a move
3. Submit a move
Expected result: Homepage should have move locator and the views should be different in desktop and mobile

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4237) for this change

## Screenshots

image

image


mariaclaratgolfo:
Desktop will need some changes to reflect the [proposed design](https://zpl.io/29r7YKz)
The layout on desktop is meant to mimic that in the TOO experience. It should utilize [this component](https://storybook.dp3.us/?path=/story/components-headers-customer-header--customer) 

- Current duty station, destination duty station and report by date should be displayed horizontally in a data format [vs. conversational format on mobile]

- Padding and font rules should match the component

mariaclaratgolfo (changes requested):
Added notes in comment

Ronolibert:
@mariaclaratgolfo Migrating to the new customer header will be done in https://dp3.atlassian.net/browse/MB-5720

src/pages/MyMove/Home/Home.module.scss
```diff
@@ -68,3 +69,52 @@
     display: none;
   }
 }
+
+@media (max-width: $tablet) {
+  .subheaderContainer {
+    margin-top: 12px;
```
sojeri, line 75: just double checking my understanding here: don't we have unit measures that we prefer to use in styles like this rather than using raw pixel amounts?

src/pages/MyMove/Home/Home.module.scss
```diff
@@ -68,3 +69,52 @@
...
+    .subheaderWeightAllowance {
+      margin-right: 15px;
+    }
+
+    .subheaderWeightAllowance,
+    .subheaderMoveCode {
+      display: inline-block;
+
+      > * {
+        display: inline-block;
```
sojeri, line 103: why do we need all these inline-block styles? would it make more sense to use display: flex here?

src/pages/MyMove/Home/index.jsx
```diff
@@ -200,7 +200,18 @@ class Home extends Component {
         <strong>{serviceMember.current_station.name}.</strong> Report by{' '}
         <strong>{moment(orders.report_by_date).format('DD MMM YYYY')}.</strong>
         <br />
-        Weight allowance: <strong>{serviceMember.weight_allotment.total_weight_self} lbs</strong>
+        <div className={styles.subheaderContainer}>
+          <div className={styles.subheaderWeightAllowance}>
+            <div>Weight allowance</div>
```
sojeri, line 205: it would be more semantic to have this be a p tag containing the strong element.

in addition, the top level p tag (line 198) seems like it'd be better off as a fragment with the first three lines wrapped still in a p

src/pages/MyMove/Home/index.jsx
```diff
@@ -200,7 +200,18 @@ class Home extends Component {
...
+            <strong>{serviceMember.weight_allotment.total_weight_self} lbs</strong>
+          </div>
+          {move.locator && (
+            <div className={styles.subheaderMoveCode}>
+              <div>Move code</div>
```
sojeri, line 210: same comment here re: semantic HTML and p containing strong

lynzt (approved):
I see the `move code` on the home screen after submitting a move
🚢

sojeri (changes requested):
left a few small comments about using more semantic HTML and styles.

src/pages/MyMove/Home/Home.module.scss
suzubara: I agree, display flex seems better to use here. also, is there a reason to have distinct classes for `subheaderWeightAllowance` and `subheaderMoveCode`? it seems like they could use a more generic subheader class

src/pages/MyMove/Home/index.jsx
suzubara: agreed, there are a lot of invalid DOM nesting errors in the console when I view this page (divs not allowed inside p, for example)

suzubara (changes requested):
+1 to @sojeri 's comments about the HTML structure. I know the plan is to refactor this and use the Customer header component eventually, but it would be good to resolve the DOM errors at least.

src/pages/MyMove/Home/Home.module.scss
Ronolibert: Yup, I'll use the em value instead. Forgot about this since design didn't have the purple boxes just the raw pixel value

mariaclaratgolfo (approved):

Ronolibert:
@sojeri @suzubara Thanks! Rushed a bit too much on this, simplified the HTML/CSS and used more semantic HTML

sojeri (commented):
the HTML looks a lot better now! left one more comment about the last trailing DOM issue :)

suzubara (approved):
Adding to this PR since @Ronolibert is OOO. I updated the markup & used `dl` and also re-organized the CSS a little to not use any max-width media queries. need to add my own approval since I had requested changes earlier.

sojeri (approved):
🚀

PR #5351, opened 2020-11-25
MB-4830 Bug initial authorized weight should be same as weight allowance

donaldthai:
## Description

> According to LTC Ryan, weight entitlement/allowance should not include pro gear. We still want to display the amounts for pro gear and spouse pro gear, like we already to in the TOO interface. However, only rank and dependents should be used to determine the initial weight allowance and authorized allowance. 
> 
> ## Acceptance Criteria
> 
> Write a test to verify the existing broken behavior
> 
> Make the test pass by modifying the `GetEntitlement` function to only take the rank and dependents and return the appropriate number
> 
> Modify all the places where we call `GetEntitlement`. 

## Reviewer Notes

## Setup

```sh
make db_dev_e2e_populate
make server_run
make office_client_run
```

## To test
1. Log in to office app as TOO role, locally
2. Click on any test move (non of the moves should have a set authorized weight)
3. Scroll down to `Allowances`
image
4. Weight Allowance and Authorized Weight should be the same

Notes:
- To test for dependents_authorized is false, you can quickly change the value in the database to see the changes in the `entitlements` table

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4830) for this change


pkg/handlers/internalapi/moves_test.go
```diff
@@ -354,7 +354,6 @@ func (suite *HandlerSuite) TestShowMoveDatesSummaryHandler() {
 	suite.Equal(moveDate, *okResponse.Payload.MoveDate)
 
 	pack := []strfmt.Date{
-		strfmt.Date(time.Date(2018, 10, 4, 0, 0, 0, 0, time.UTC)),
```
donaldthai, line 357: Was able to remove this because the weights are now based on `TotalWeightSelf` or `TotalWeightSelfWithDependents` which is a lot less than with pro-gear and pro-gear spouse. The formula used is an estimated `5000lbs` per day. So a rank E1 with dependents entitlement weight is 8000, that would mean an estimated 2 days for packing.

pkg/handlers/internalapi/entitlements_test.go
```diff
@@ -40,8 +40,8 @@ func (suite *HandlerSuite) TestValidateEntitlementHandlerReturns200() {
 	move := ppm.Move
 
 	// When: rank is E1, the orders have dependents and spouse gear, and
```
noelledusahel, line 42: nit: is this comment still accurate re: including pro gear?

donaldthai: Nice catch! I'll remove it.

monfresh:
@donaldthai I think there's still some missing logic. For an E-1 rank with dependents (which I think is what most moves locally have), both weight allowance and authorized weight should be 8000. I fixed the weight allowance in my branch, so if you could please merge master into your branch, we can verify that here as well.

Before, if the authorized weight was empty in the DB, like it is in all the devseed moves, then it would fall back to `TotalWeightSelf`, which is 5000, which is wrong when there are dependents. There should be the same conditional I added in my PR.

donaldthai:
> @donaldthai I think there's still some missing logic. For an E-1 rank with dependents (which I think is what most moves locally have), both weight allowance and authorized weight should be 8000. I fixed the weight allowance in my branch, so if you could please merge master into your branch, we can verify that here as well.
> 
> Before, if the authorized weight was empty in the DB, like it is in all the devseed moves, then it would fall back to `TotalWeightSelf`, which is 5000, which is wrong when there are dependents. There should be the same conditional I added in my PR.

@monfresh what a coincidence! I just pushed the latest code fixing the exact thing you're talking about!

pkg/models/ghc_entitlements.go
```diff
@@ -42,6 +42,9 @@ func (e *Entitlement) AuthorizedWeight() *int {
 	case e.DBAuthorizedWeight != nil:
 		return e.DBAuthorizedWeight
 	case e.WeightAllotment() != nil:
+		if e.DependentsAuthorized != nil && *e.DependentsAuthorized == true {
+			return &e.WeightAllotment().TotalWeightSelfPlusDependents
+		}
```
monfresh, line 47: Can we update the comment too, please? It still refers to `TotalWeightSelf` only

donaldthai: Done!

pkg/models/ghc_entitlements_test.go
```diff
@@ -1,25 +1,37 @@
...
+	suite.T().Run("with dependents authorized, TotalWeightSelfPlusDependents is AuthorizedWeight", func(t *testing.T) {
+		dependentsAuthorized := true
+		entitlement := models.Entitlement{DependentsAuthorized: &dependentsAuthorized}
+		entitlement.SetWeightAllotment("E_1")
+		err := suite.DB().Create(&entitlement)
+		suite.NoError(err)
 
-	suite.Equal(entitlement.WeightAllotment().TotalWeightSelf, *entitlement.AuthorizedWeight())
+		suite.Equal(entitlement.WeightAllotment().TotalWeightSelfPlusDependents, *entitlement.AuthorizedWeight())
+	})
```
monfresh, line 36: Can we use plain models here without saving them to the DB since we're only testing logic based on attributes?

donaldthai: Oh yea, that's definitely a good route to go with. I can refactor this.

donaldthai: Yup, looks like removing the db create call does the trick. I'll push up the code.

pkg/notifications/move_submitted.go
```diff
@@ -69,7 +69,7 @@ func (m MoveSubmitted) emails(ctx context.Context) ([]emailContent, error) {
 		return emails, err
 	}
 
-	totalEntitlement, err := models.GetEntitlement(*serviceMember.Rank, orders.HasDependents, orders.SpouseHasProGear)
+	totalEntitlement, err := models.GetEntitlement(*serviceMember.Rank, orders.HasDependents)
```
monfresh, line 72: In what context does this get used? I wonder if this is expected to include pro gear like in the summary worksheet?

donaldthai: I think this refers to when a move is submitted and an email is sent to the service member. @kimallen knows more though.

donaldthai: The email copy could probably use an update to include the pro-gear and pro-gear spouse. Not sure if this PR is the right place to do it.

monfresh (approved):
LGTM. It would be good to have @transcom/truss-pamplemoose  review as well.

pkg/notifications/move_submitted.go
lynzt: @monfresh Donald is right... this is getting the `entitlement` to display on the email we send to the SM after they submit their move...

lynzt (commented):
LGTM... 
I have an outstanding question if we need to do anything in regards to existing moves in our system that used the old entitlement numbers or if this is only affecting moves going forward.

pkg/models/shipment_summary_worksheet_test.go
```diff
@@ -110,9 +110,11 @@ func (suite *ModelSuite) TestFetchDataShipmentSummaryWorksheet() {
 	suite.Equal(unit.Pound(rankWtgAllotment.ProGearWeight), ssd.WeightAllotment.ProGear)
 	suite.Equal(unit.Pound(0), ssd.WeightAllotment.SpouseProGear)
 	suite.Require().NotNil(ssd.ServiceMember.Rank)
-	totalEntitlement, err := models.GetEntitlement(*ssd.ServiceMember.Rank, ssd.Order.HasDependents, ssd.Order.SpouseHasProGear)
+	weightAllotment := models.GetWeightAllotment(*ssd.ServiceMember.Rank)
+	// E_9 rank, no dependents, no spouse pro-gear
+	totalWeight := weightAllotment.TotalWeightSelf + weightAllotment.ProGearWeight
```
kimallen, line 115: Should this also include `weightAllotment.ProGearWeightSpouse`?

kimallen: oh nm, I see the comment specifying this case

kimallen (approved):
Looks good!  Thanks for doing due-diligence and cross-team communication to make sure PPM stuff didn't get impacted.

PR #5352, opened 2020-11-25
[MB-5251][MB-5572] Origin Contact Info on home screen

Ronolibert:
## Description

Originally, there was static data for the contact info on the home screen. This adds dynamic contact info and additional copy for when the move is submitted

## Setup
`make server_run`
`make client_run`
1. Log in
2. Set up a move and submit it
3. Scroll to bottom of home screen
Expected result: Contact component should have correct origin duty station and additional text

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5251) for this change
* Also handles [this Jira story](https://dp3.atlassian.net/browse/MB-5572)

## Screenshots

image

### DIFF
--- a/src/components/Customer/Home/Contact/index.jsx
+++ b/src/components/Customer/Home/Contact/index.jsx
@@ -4,8 +4,8 @@ import { string, bool } from 'prop-types';
 import styles from './Contact.module.scss';
 
 const Contact = ({ header, dutyStationName, moveSubmitted, officeType, telephone }) => (
-  <div className={styles['contact-container']}>
-    <h6 className={styles['contact-header']}>{header}</h6>
+  <div className={styles.contactContainer}>
+    <h6 className={styles.contactHeader}>{header}</h6>
     <p>
       <strong>{dutyStationName}</strong>
       <br />
@@ -15,7 +15,7 @@ const Contact = ({ header, dutyStationName, moveSubmitted, officeType, telephone
     </p>
     {moveSubmitted && (
       <p data-testid="move-submitted-instructions">
-        After you hear from your move counselor, they should be your first resource for questions.
+        Talk to your move counselor or directly with your movers if you have questions during your move.
       </p>
     )}
   </div>
--- a/src/components/Customer/Home/Contact/index.jsx
+++ b/src/components/Customer/Home/Contact/index.jsx
@@ -13,16 +13,7 @@ const Contact = ({ header, dutyStationName, moveSubmitted, officeType, telephone
       <br />
       <span>{telephone}</span>
     </p>
-    <p>
-      {moveSubmitted && (
-        <>
-          You&apos;ll hear from a move counselor and the movers themselves within the next few days.
-          <br />
-          <br />
-          Talk to either of them with questions about your move.
-        </>
-      )}
-    </p>
+    {moveSubmitted && <p>After you hear from your move counselor, they should be your first resource for questions.</p>}
   </div>
 );
 
--- a/src/components/Customer/Home/Contact/index.jsx
+++ b/src/components/Customer/Home/Contact/index.jsx
@@ -1,9 +1,9 @@
 import React from 'react';
-import { string } from 'prop-types';
+import { string, bool } from 'prop-types';
 
 import styles from './Contact.module.scss';
 
-const Contact = ({ header, dutyStationName, officeType, telephone }) => (
+const Contact = ({ header, dutyStationName, moveSubmitted, officeType, telephone }) => (
   <div className={styles['contact-container']}>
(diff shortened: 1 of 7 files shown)


src/components/Customer/Home/Contact/index.jsx
```diff
@@ -13,12 +13,14 @@ const Contact = ({ header, dutyStationName, officeType, telephone }) => (
       <br />
       <span>{telephone}</span>
     </p>
+    {moveSubmitted && <p>After you hear from your move counselor, they should be your first resource for questions.</p>}
```
gidjin, line 16: Is there a unit or other test for this change? If not I would like to see a unit test added

src/pages/MyMove/Home/index.jsx
```diff
@@ -415,9 +416,10 @@ class Home extends Component {
                     </SectionWrapper>
                     <Contact
                       header="Contacts"
-                      dutyStationName="Seymour Johnson AFB"
+                      dutyStationName={current_station.transportation_office.name}
                       officeType="Origin Transportation Office"
-                      telephone="(919) 722-5458"
+                      telephone={current_station.transportation_office.phone_lines[0]}
+                      moveSubmitted={move.status === MOVE_STATUSES.SUBMITTED}
```
gidjin, line 422: Are there unit or other tests for these changes? If not I would like to see a unit test added

src/components/Customer/Home/Contact/index.jsx
Ronolibert: Good call! Added additional unit tests

src/pages/MyMove/Home/index.jsx
```diff
@@ -197,7 +197,7 @@ class Home extends Component {
     return (
       <p>
         You&apos;re moving to <strong>{orders.new_duty_station.name}</strong> from{' '}
-        <strong>{serviceMember.current_station.name}.</strong> Report by{' '}
+        <strong>{serviceMember?.current_station?.name}.</strong> Report by{' '}
```
sojeri, line 200: why is serviceMember nullable here but not in other places in the file like line 203?

Ronolibert: You're right! The service member will never be a nullable. At most an empty object.  Removing the optional chaining on the serviceMember.

src/pages/MyMove/Home/index.test.jsx
```diff
@@ -74,6 +75,29 @@ describe('Home component', () => {
...
+          },
+        },
+        move: {
+          status: MOVE_STATUSES.SUBMITTED,
+        },
+      };
+
+      const wrapper = mountHome(props);
+      expect(wrapper.find('[data-testid="move-submitted-instructions"]').text()).toBe(
+        'After you hear from your move counselor, they should be your first resource for questions.',
```
suzubara, line 96: Since we're already testing for this text content in the Contact unit test, I think the test in the Home page component would be more relevant if it asserted something like `expect(wrapper.find(Contact).prop('moveSubmitted')).toEqual(true)`

suzubara (commented):
left a couple minor comments!

mariaclaratgolfo (approved):

sojeri (commented):
I noticed something else about the optional chaining. I guess that's my jam on this PR. 😅

src/pages/MyMove/Home/index.jsx
suzubara: FWIW I think it doesn't hurt to use optional chaining even when it shouldn't be necessary, especially at this point in the code when the client-side data is a little bit in flux.

src/pages/MyMove/Home/index.jsx
```diff
@@ -415,9 +417,10 @@ class Home extends Component {
                     </SectionWrapper>
                     <Contact
                       header="Contacts"
-                      dutyStationName="Seymour Johnson AFB"
+                      dutyStationName={current_station?.transportation_office?.name}
                       officeType="Origin Transportation Office"
-                      telephone="(919) 722-5458"
+                      telephone={current_station?.transportation_office?.phone_lines[0]}
```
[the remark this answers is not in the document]
suzubara: IMO in that case it would mean transportation office data is missing, and is likely an API or data error. Since AFAIK we don't have that use case designed, I wouldn't consider handling it to be part of this story. I did add a storybook story for this situation, in case we decide to update it in the future.

suzubara (approved):
I'm handling this PR since Ron is OOO. I addressed outstanding PR comments, and also noticed the static copy for post-submission state was not accurate so I updated that as well.

sojeri (approved):
🚀

gidjin (approved):
Thanks for adding tests! Code looks good to me

PR #5367, opened 2020-11-30
[MB-5716] Approve MTO prior to MTO Shipment status update and service item creation

duncan-truss:
## Description

After the introduction of an additional move status `APPROVALS_REQUESTED`, related to the TOO moves queue work, we began getting 500 errors when approving the MTO (making it available to the prime) with shipments and service items.  Instead of the approve modal dismissing it would stay open until you refreshed the page because of a non 200 success response.

We were making the MTO approval status update and the mto shipment status update requests in parallel using promises in our form submit handler.  Meanwhile the create mto service item service required that the MTO be in approved status to transition to the `APPROVALS_REQUESTED` status.  Because these calls could resolve in a non-deterministic order there would often be an error returned.

The interim solution is to make the mto shipment status updates wait until the initial approval of the mto has completed successfully.  Also in the event that we do hit these errors again, they should now return more gracefully as conflict errors instead of internal server errors.

Due to the closeness of the demo and amount of work being done in this part of the codebase I restrained myself from refactoring further.  Others have pointed out these multiple calls should likely be captured in a transaction.  I would also point out that the flows for basic service item creation and shipment accessorial service items seem different enough to split out their logic.  We're also continually requerying for the MTO object which we already have and are doing redundant work there. 

## Reviewer Notes

I tried mocking out the approveMTOShipment function in the jest component test but it wasn't detecting it getting called for some reason.

## Setup

```sh
make db_dev_e2e_populate
make server_run
make office_client_run
```

1. Login as the TOO user in the office app
2. Find a move in the queue that is in the 'New Move' status, ideally one with 2 shipments
3. Select at least one of the shipments and a service item for approval
4. Click the Approve and send button on the confirmation modal
(Optional) 4b. Select the other shipment for approval given the MTO status is already available to the prime
5. The multiple network requests to update the mto and shipment status should no longer return error rsponses
6. The modal should close and the page should refresh on success.
7. Navigate back to the queue and the move should now have the status 'Approvals requested'

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5716) for this change

### DIFF
--- a/pkg/handlers/ghcapi/move_task_order.go
+++ b/pkg/handlers/ghcapi/move_task_order.go
@@ -17,7 +17,7 @@ import (
 	"github.com/transcom/mymove/pkg/services/audit"
 )
 
-// GetMoveTaskOrderHandler fetches a of a Move Task Order
+// GetMoveTaskOrderHandler fetches a Move Task Order
 type GetMoveTaskOrderHandler struct {
 	handlers.HandlerContext
 	moveTaskOrderFetcher services.MoveTaskOrderFetcher
--- a/pkg/handlers/ghcapi/move_task_order.go
+++ b/pkg/handlers/ghcapi/move_task_order.go
@@ -17,13 +17,13 @@ import (
 	"github.com/transcom/mymove/pkg/services/audit"
 )
 
-// GetMoveTaskOrderHandler updates the status of a Move Task Order
+// GetMoveTaskOrderHandler fetches a of a Move Task Order
 type GetMoveTaskOrderHandler struct {
 	handlers.HandlerContext
 	moveTaskOrderFetcher services.MoveTaskOrderFetcher
 }
 
-// Handle updates the status of a MoveTaskOrder
+// Handle fetches a single MoveTaskOrder
 func (h GetMoveTaskOrderHandler) Handle(params movetaskorderops.GetMoveTaskOrderParams) middleware.Responder {
 	logger := h.LoggerFromRequest(params.HTTPRequest)
 
@@ -76,6 +76,8 @@ func (h UpdateMoveTaskOrderStatusHandlerFunc) Handle(params movetaskorderops.Upd
 			return movetaskorderops.NewUpdateMoveTaskOrderStatusBadRequest()
 		case services.PreconditionFailedError:
 			return movetaskorderops.NewUpdateMoveTaskOrderStatusPreconditionFailed().WithPayload(&ghcmessages.Error{Message: handlers.FmtString(err.Error())})
+		case services.ConflictError:
+			return movetaskorderops.NewUpdateMoveTaskOrderStatusConflict().WithPayload(&ghcmessages.Error{Message: handlers.FmtString(err.Error())})
 		default:
 			return movetaskorderops.NewUpdateMoveTaskOrderStatusInternalServerError()
 		}
--- a/pkg/services/move_task_order/move_task_order_updater.go
+++ b/pkg/services/move_task_order/move_task_order_updater.go
@@ -53,7 +53,7 @@ func (o moveTaskOrderUpdater) MakeAvailableToPrime(moveTaskOrderID uuid.UUID, eT
 
 		verrs, err = o.builder.UpdateOne(mto, &eTag)
 		if verrs != nil && verrs.HasAny() {
-			return &models.Move{}, services.InvalidInputError{}
+			return &models.Move{}, services.NewInvalidInputError(mto.ID, nil, verrs, "")
 		}
 		if err != nil {
 			switch err.(type) {
--- a/pkg/services/move_task_order/move_task_order_updater.go
+++ b/pkg/services/move_task_order/move_task_order_updater.go
@@ -1,6 +1,7 @@
 package movetaskorder
 
 import (
+	"errors"
 	"time"
 
 	"github.com/gobuffalo/pop/v5"
@@ -46,6 +47,19 @@ func (o moveTaskOrderUpdater) MakeAvailableToPrime(moveTaskOrderID uuid.UUID, eT
 		if mto.Status == models.MoveStatusSUBMITTED {
 			err = mto.Approve()
 			if err != nil {
+				return &models.Move{}, services.NewConflictError(mto.ID, err.Error())
+			}
+		}
+
+		verrs, err = o.builder.UpdateOne(mto, &eTag)
+		if verrs != nil && verrs.HasAny() {
+			return &models.Move{}, services.InvalidInputError{}
+		}
+		if err != nil {
+			switch err.(type) {
+			case query.StaleIdentifierError:
+				return nil, services.NewPreconditionFailedError(mto.ID, err)
+			default:
 				return &models.Move{}, err
 			}
 		}
@@ -63,6 +77,17 @@ func (o moveTaskOrderUpdater) MakeAvailableToPrime(moveTaskOrderID uuid.UUID, eT
 				ApprovedAt:      &now,
 			})
 		}
+
+		if err != nil {
+			if errors.Is(err, models.ErrInvalidTransition) {
+				return &models.Move{}, services.NewConflictError(mto.ID, err.Error())
+			}
+			return &models.Move{}, err
+		}
+		if verrs != nil {
+			return &models.Move{}, verrs
+		}
+
 		if includeServiceCodeCS {
 			// create if doesn't exist
 			_, verrs, err = o.serviceItemCreator.CreateMTOServiceItem(&models.MTOServiceItem{
@@ -75,22 +100,19 @@ func (o moveTaskOrderUpdater) MakeAvailableToPrime(moveTaskOrderID uuid.UUID, eT
 		}
 
 		if err != nil {
+			if errors.Is(err, models.ErrInvalidTransition) {
+				return &models.Move{}, services.NewConflictError(mto.ID, err.Error())
+			}
 			return &models.Move{}, err
 		}
 		if verrs != nil {
 			return &models.Move{}, verrs
 		}
-	}
 
-	verrs, err = o.builder.UpdateOne(mto, &eTag)
-	if verrs != nil && verrs.HasAny() {
-		return &models.Move{}, services.InvalidInputError{}
-	}
-	if err != nil {
-		switch err.(type) {
-		case query.StaleIdentifierError:
-			return nil, services.NewPreconditionFailedError(mto.ID, err)
-		default:
+		// CreateMTOServiceItem may have updated the mto status so refetch as to not return incorrect status
+		// TODO: Modify CreateMTOServiceItem to return the updated move or refactor to operate on the passed in reference
+		mto, err = o.FetchMoveTaskOrder(moveTaskOrderID)
+		if err != nil {
 			return &models.Move{}, err
 		}
 	}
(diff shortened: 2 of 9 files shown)


pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -46,6 +47,19 @@ func (o moveTaskOrderUpdater) MakeAvailableToPrime(moveTaskOrderID uuid.UUID, eT
 		if mto.Status == models.MoveStatusSUBMITTED {
 			err = mto.Approve()
 			if err != nil {
+				return &models.Move{}, services.NewConflictError(mto.ID, err.Error())
+			}
+		}
+
+		verrs, err = o.builder.UpdateOne(mto, &eTag)
```
duncan-truss, line 54: Initially we were changing the status only in memory and not saving the move to the database until after the creation of the basic service items which was confusing.  The create mto service item service refetches the move, which is inefficient but is being called from other places so I didn't want to break that behavior.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -63,6 +77,17 @@ func (o moveTaskOrderUpdater) MakeAvailableToPrime(moveTaskOrderID uuid.UUID, eT
...
+
+		if err != nil {
+			if errors.Is(err, models.ErrInvalidTransition) {
+				return &models.Move{}, services.NewConflictError(mto.ID, err.Error())
+			}
+			return &models.Move{}, err
+		}
+		if verrs != nil {
+			return &models.Move{}, verrs
+		}
```
duncan-truss, line 89: We were not checking each error response, so it was possible the first service item creation call could fail and we would still proceed to create the next one.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -75,22 +100,19 @@ func (o moveTaskOrderUpdater) MakeAvailableToPrime(moveTaskOrderID uuid.UUID, eT
...
-		return &models.Move{}, services.InvalidInputError{}
-	}
-	if err != nil {
-		switch err.(type) {
-		case query.StaleIdentifierError:
-			return nil, services.NewPreconditionFailedError(mto.ID, err)
-		default:
+		// CreateMTOServiceItem may have updated the mto status so refetch as to not return incorrect status
+		// TODO: Modify CreateMTOServiceItem to return the updated move or refactor to operate on the passed in reference
+		mto, err = o.FetchMoveTaskOrder(moveTaskOrderID)
```
duncan-truss, line 114: For basic service item creation the move status shouldn't be updated within CreateMTOServiceItem, only when creating service items with a shipment will the status change.  Still to be safe let's make sure we're not returning a stale MTO object.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -63,6 +77,17 @@ func (o moveTaskOrderUpdater) MakeAvailableToPrime(moveTaskOrderID uuid.UUID, eT
 				ApprovedAt:      &now,
 			})
 		}
+
+		if err != nil {
+			if errors.Is(err, models.ErrInvalidTransition) {
```
sandy-wright, line 82: ooh I haven't seen `Is()` before ✨

src/components/Office/RequestedShipments/RequestedShipments.jsx
```diff
@@ -45,36 +45,38 @@ const RequestedShipments = ({
...
+        approveMTOShipment(moveTaskOrder.id, shipment.id, 'APPROVED', shipment.eTag),
+      );
+
       const mtoApprovalServiceItemCodes = {
         serviceCodeMS: values.shipmentManagementFee,
         serviceCodeCS: values.counselingFee,
       };
 
       // if mto is not yet approved, add request to approve it
       if (!moveTaskOrder.availableToPrimeAt) {
```
duncan-truss, line 58: I am going to adjust this for when the TOO approves shipments one at a time.

pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -189,12 +189,11 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
 		return nil, verrs, services.NewQueryError("unknown", err, "")
 	}
 
-	// TODO: Fix error message to be conflict error instead of server error so user knows the Move status is wrong
 	// TODO: Determine if this should be in the same transaction as the service items (they get created even if this fails)
 	if move.Status != models.MoveStatusAPPROVALSREQUESTED {
 		err := move.SetApprovalsRequested()
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, services.NewConflictError(move.ID, err.Error())
```
sandy-wright, line 196: 🙌

pkg/services/mto_shipment/mto_shipment_updater_test.go
```diff
@@ -564,4 +565,21 @@ func (suite *MTOShipmentServiceSuite) TestUpdateMTOShipmentStatus() {
...
+			},
+		})
+		eTag = etag.GenerateEtag(shipment7.UpdatedAt)
+
+		updatedShipment, err := updater.UpdateMTOShipmentStatus(shipment7.ID, models.MTOShipmentStatusApproved, nil, eTag)
+		suite.Nil(updatedShipment)
+		suite.Error(err)
+		suite.IsType(services.ConflictError{}, err)
+		suite.True(strings.Contains(err.Error(), "Cannot move to Approvals Requested when the Move is not in an Approved state"))
+	})
```
sandy-wright, line 584: thanks for adding this test!

pkg/services/move_task_order/move_task_order_updater.go
sandy-wright: I definitely agree with this! We may want to coordinate in later sprints to update this behavior because we might see more issues later (for example, the `SetApprovalsRequested()` will return an error if the Move is _already_ in the APPROVALS_REQUESTED status, which maybe shouldn't cause an error?). In any case, this is the best way to do it for now 👍

sandy-wright (approved):
LGTM! Behavior in the app seems to be fixed, everything else looks good too! Thanks, Duncan 😄

pkg/services/move_task_order/move_task_order_updater.go
duncan-truss: It's handy when errors get wrapped!

duncan-truss:
> LGTM! Behavior in the app seems to be fixed, everything else looks good too! Thanks, Duncan 😄

Appreciate all your help @sandy-wright!

pkg/handlers/ghcapi/move_task_order.go
```diff
@@ -17,13 +17,13 @@ import (
 	"github.com/transcom/mymove/pkg/services/audit"
 )
 
-// GetMoveTaskOrderHandler updates the status of a Move Task Order
+// GetMoveTaskOrderHandler fetches a of a Move Task Order
```
monfresh, line 20: should this say "fetches a Move"?

pkg/services/mto_shipment/mto_shipment_updater_test.go
```diff
@@ -564,4 +565,21 @@ func (suite *MTOShipmentServiceSuite) TestUpdateMTOShipmentStatus() {
...
 	})
+
+	suite.T().Run("When move status is not already approved", func(t *testing.T) {
+		submittedMTO := testdatagen.MakeMove(suite.DB(), testdatagen.Assertions{Move: models.Move{Status: models.MoveStatusSUBMITTED}})
+		shipment7 := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+			Move: submittedMTO,
+			MTOShipment: models.MTOShipment{
+				Status: models.MTOShipmentStatusSubmitted,
+			},
+		})
```
monfresh, line 576: What do you think about using the `MakeHHGMoveWithShipment` helper instead?

pkg/services/mto_shipment/mto_shipment_updater_test.go
monfresh: It's a little confusing that we are adding a test here when mto_shipment_updater hasn't changed, and that we're checking for an error related to a move when updating a shipment. Is it because of the current nature of the code that this was the best place to add this test?

src/components/Office/RequestedShipments/RequestedShipments.jsx
```diff
@@ -45,36 +45,51 @@ const RequestedShipments = ({
...
-        .catch(() => {
-          // TODO: Decided if we wnat to display an error notice, log error event, or retry
-          setSubmitting(false);
-        });
+          })
+          .catch(() => {
+            // TODO: Decide if we want to display an error notice, log error event, or retry
+            setSubmitting(false);
+          });
+      }
```
monfresh, line 92: I would think this should have been caught by one of our Cypress tests, no? Are we missing a test, or should the existing one be updated?

duncan-truss: It was a race condition since we were calling `Promise.all` so it wouldn't happen 100% of the time.  Also since they are independent calls, it means that the MTO would be approved with the basic service items but the modal wouldn't close and the page wouldn't refresh.  It would fail to move the status from "New move" to "Approvals requested" I guess.  We could test for that assuming that would have caught it.

monfresh: Right, I'm wondering why we haven't seen intermittent Cypress failures. The `tooFlows.js` test seems to be going through this workflow. Is it a test data issue? One strange thing I'm seeing is that the move that the `tooFlows` test is using (it's the one with `Locator` "TEST12") has a shipment in Submitted status, but it already has an ApprovedDate: https://github.com/transcom/mymove/blob/master/pkg/testdatagen/scenario/e2ebasic.go#L1295

Would that affect the test?

duncan-truss: Yea you would think it would fail if the modal is still visible since that's in the test.  I saw that timestamp set as well but the shipments were moving from sumbitted to approved properly, it was just the parent move status that was failing to transition from approved to approvals requested because it would still be in the submitted status when the shipments were being approved.

pkg/services/mto_shipment/mto_shipment_updater_test.go
duncan-truss: I had to add the shipments back to the move after creation so it didn't require requerying the move.

pkg/services/mto_shipment/mto_shipment_updater_test.go
duncan-truss: This was mostly the part about making sure that the previous 500 error is now a 409 conflict.  We don't expect this to happen now but if the move has the wrong status at least we are not defaulting to 500 when we know what the issue is with the invalid state transition.

It was the order of the api calls in the form submission handler that was causing the issue. I tried testing the mock call of approveMTOShipment in the react component but I think it's too complex with promises after the first one resolves.

duncan-truss:
@monfresh Thanks for reviewing this, I want to see if I can merge it before the dry run if possible.  This is a stopgap solution for the demo, so I hope we can make it more robust in cypress and elsewhere soon.

monfresh:
Sounds good. Didn't mean to hold things up!

PR #5373, opened 2020-12-01
first pass at reverting PR #5309; see TODOs

sojeri:
## Description

Partial work on reverting PR #5309. I have some open questions in `pkg/services/invoice/syncada_sftp_sender.go`.

## References

- [Jira story](https://dp3.atlassian.net/browse/MB-5744)
- [slack a-team pairing thread](https://ustcdp3.slack.com/archives/CP6F568DC/p1606778239067800)

### DIFF
--- a/cmd/send-to-syncada-via-sftp/main.go
+++ b/cmd/send-to-syncada-via-sftp/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"fmt"
 	"log"
 	"os"
@@ -10,6 +11,7 @@ import (
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 	"go.uber.org/zap"
+	"golang.org/x/crypto/ssh"
 
 	"github.com/transcom/mymove/pkg/cli"
 	"github.com/transcom/mymove/pkg/logging"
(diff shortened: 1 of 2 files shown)


pkg/services/invoice/syncada_sftp_sender.go
```diff
@@ -58,6 +60,10 @@ func InitNewSyncadaSFTPSession() (services.SyncadaSFTPSender, error) {
 		return nil, fmt.Errorf("Invalid credentials sftp missing SYNCADA_SFTP_INBOUND_DIRECTORY")
 	}
 
+	// TODO: handle hostKey?
```
gidjin, line 63: I would recommend pulling the host key from an environment variable as the other parameters in this function. The values are stored securely in chamber.

pkg/services/invoice/syncada_sftp_sender.go
```diff
@@ -58,6 +60,10 @@ func InitNewSyncadaSFTPSession() (services.SyncadaSFTPSender, error) {
...
+	// TODO: why do we have this reimplementation of the viper CLI layer??
```
gidjin, line 64: This service is currently called in a cli, however it will be called from a job that will run automatically in our deployed environments eventually. It will need the credentials to connect to Syncada which are stored in chamber and populated into the environment vars at deploy time.

pkg/services/invoice/syncada_sftp_sender.go
```diff
@@ -58,6 +60,10 @@ func InitNewSyncadaSFTPSession() (services.SyncadaSFTPSender, error) {
...
+	// TODO: or else why do we need both init and new??
```
gidjin, line 65: I think it was convenience to allow creating an instance without looking to the env, but could be that we can consolidate them into one method.

cmd/send-to-syncada-via-sftp/main.go
```diff
@@ -48,6 +50,39 @@ func initFlags(flag *pflag.FlagSet) {
 	flag.SortFlags = false
 }
 
+func getHostKey(host string) ssh.PublicKey {
```
sojeri, line 53: from pairing: instead of this, we probably want to get the value and save it in chamber. retrieve from chamber instead of known hosts. the key is unlikely to change. we'd need to talk to the vendor if it did to confirm that were expected.

sojeri:
Closing this one as it is now superseded by John's PR #5384. 😄

PR #5382, opened 2020-12-01
MB-5723 Allow log level to be configurable

monfresh:
**Description**

It is a STIG requirement for the application log levels to
be configurable, and for them to default to `INFO`. It is also an
industry best practice. Rails, for example, provides this out of the
box.

**How**

* Add new logging level flags to `pkg/cli/logging.go`

* Modify our existing `logging/log.go` package to accept a string that
defines the desired log level, based on the `LOGGING_LEVEL` env var.
Previously, we were limiting the configuration to the default log
levels provided by zap (`DEBUG` for development, and `INFO` for
production), and the ability to set the log level to `DEBUG` via a
boolean based on the `DEBUG_LOGGING` env var defined in
`pkg/cli/verbose.go`.

* Update all calls to `logging.Config` to replace the boolean with
the string from the new logging level flag.

* Remove `pkg/cli/verbose.go` and associated tests.

* Remove `DEBUG_LOGGING=true` from prd, stg, and exp `.env` files, and
replace with `LOGGING_LEVEL=info`, which is the desired default level
in deployed environments per the STIG.

* Make the logging levels case-insensitive because it's the nice
thing to do

This allows us to change the log level to any of the 5 acceptable
values (`FATAL`, `ERROR`, `WARN`, `INFO`, `DEBUG`), without having to
deploy new code. We only need to change the env var and restart the
app.

**Reviewer Notes**

1. From the master branch, run `make server_run` locally
- [ ] Verify that you see `DEBUG` entries in the log, such as
"Server DOD Key Pair Loaded"
3. Stop the server, and switch to this branch
4. Run `make server_run`
- [ ] Verify that you only see log items marked as `INFO`.
6. Add `export LOGGING_LEVEL=DeBUG` to your `.envrc.local`
7. Stop the server
8. Run `direnv allow`
9. Run `make server_run`
- [ ] Verify that you see `DEBUG` entries
10. Update the log level in your `.envrc.local` to a bogus value
11. Repeat steps 7-9
- [ ] Verify that the server crashes and lets you know that the
value is not allowed, and what the allowed values are.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5723) for this change
* [zap Prod Config](https://github.com/uber-go/zap/blob/v1.16.0/config.go#L120)
* [zap Dev Config](https://github.com/uber-go/zap/blob/v1.16.0/config.go#L161)


pkg/cli/logging.go
```diff
@@ -27,25 +43,55 @@ var (
...
+		return &errInvalidLoggingLevel{Value: str, LoggingLevels: allLoggingLevels}
+	}
 	return nil
 }
+
+// LogLevelIsDebug is a helper for functions that require a boolean to determine
+// log versbosity
+func LogLevelIsDebug(v *viper.Viper) bool {
+	logLevel := strings.ToLower(v.GetString(LoggingLevelFlag))
+	return logLevel == "debug"
```
duncan-truss, line 96: I think we can compare to the constant here LoggingLevelDebug

duncan-truss:
@monfresh For the stig requirement I think they said it's okay to require a restart but not a redeploy.  In that case would we need the LOGGING_LEVEL to be set in chamber or overridden somewhere else?  Having to modify the *.env would mean a redeploy correct?

cmd/send-to-gex/main.go
```diff
@@ -111,7 +111,7 @@ func main() {
 
 	logger.Println(ediString)
 
-	certLogger, err := logging.Config("development", true)
+	certLogger, err := logging.Config("development", "debug")
```
duncan-truss, line 114: We should make this one configurable even if the default is LoggingLevelDebug?

cmd/webhook-client/command_test.go
```diff
@@ -23,7 +23,7 @@ type WebhookClientTestingSuite struct {
 }
 
 func TestWebhookClientTestingSuite(t *testing.T) {
-	logger, _ := logging.Config("development", true)
+	logger, _ := logging.Config("development", "debug")
```
duncan-truss, line 26: ditto

cmd/webhook-client/webhook/webhook_test.go
```diff
@@ -28,7 +28,7 @@ type WebhookClientTestingSuite struct {
 }
 
 func TestWebhookClientTestingSuite(t *testing.T) {
-	logger, _ := logging.Config("development", true)
+	logger, _ := logging.Config("development", "debug")
```
duncan-truss, line 31: Same question

cmd/send-to-gex/main.go
monfresh: I wanted to preserve the previous behavior in this first pass. I suppose it doesn't hurt to make them configurable, and I can update `.envrc` to set the default log level to debug locally. I believe all of these files in the `/cmd` directory are meant to be run locally, right? Are any of these used in production?

cmd/webhook-client/webhook/webhook_test.go
monfresh: This is a test. We would have to add new dependencies and change the signature to use viper. Do you think we need this?

monfresh:
> @monfresh For the stig requirement I think they said it's okay to require a restart but not a redeploy. In that case would we need the LOGGING_LEVEL to be set in chamber or overridden somewhere else? Having to modify the *.env would mean a redeploy correct?

That's a good question. I was under the impression that the env vars in the *.env files could be overridden on AWS. I will ask infra.

monfresh:
@duncan-truss This PR has everything needed on the app side to allow for the log level to be changed in production. The way it works is that any ENV var we define via our CLI package can automatically be changed via AWS SSM. The only thing missing is for Infra to create the AWS SSM parameter called `logging-level`, and once that's done, we can test this in experimental.

duncan-truss:
> @duncan-truss This PR has everything needed on the app side to allow for the log level to be changed in production. The way it works is that any ENV var we define via our CLI package can automatically be changed via AWS SSM. The only thing missing is for Infra to create the AWS SSM parameter called `logging-level`, and once that's done, we can test this in experimental.

Sounds good thank you! I was casually following along to the slack discussions.

duncan-truss (approved):

PR #5390, opened 2020-12-02
MB-5731 ghcapi payment requests by move handler

Ryan-Koch:
## Description

Adds a handler at `/moves/[locator]/payment-requests/` so that we can pull the payment requests for a specific move from the `GHCAPI`.

I wound up reusing and lightly extending the service object we're using for the Payment Requests queue. The adjustment allows one to pass in 0s for both the `perPage` and `Page` parameters, which then should cause the query to run without pagination but with the locator/move code value as the filter.

## Setup

If you populate your db using:

```sh
make db_dev_e2e_populate
```

Then you can go here after the office app is up (as a logged in TIO): http://officelocal:3000/ghc/v1/moves/CFCDFG/payment-requests/ 

You should get json for two payment requests.

 have been satisfied.


duncan-truss:
@Ryan-Koch In terms of API design it would make more sense to me if the api call was `move/{locator}/payment-requests`.  If you didn't want to nest them maybe `payment-requests/?moveLocator={locator}` with a query variable instead of a path one.

I'm not convinced that reusing the queue call is not more trouble than it's worth.  There's a bunch of things we don't need in there such as gbloc filtering, extra joins, and grouping.

Ryan-Koch:
> @Ryan-Koch In terms of API design it would make more sense to me if the api call was `move/{locator}/payment-requests`. If you didn't want to nest them maybe `payment-requests/?moveLocator={locator}` with a query variable instead of a path one.
> 
> I'm not convinced that reusing the queue call is not more trouble than it's worth. There's a bunch of things we don't need in there such as gbloc filtering, extra joins, and grouping.

That's fair, I'll switch the url structure around. I do think we need the gbloc filtering and looking ahead at abstract we will need some of the joins. The query is running without the grouping and pagination for this case also (it'll do this when you pass 0s for page and perpage).

The reason for the gbloc filtering is that would prevent someone from typing in a move code/locator for a move they don't have permission for and then accessing it that way. Having this (and the USMC logic) built in already is what made me think it might be worthwhile to reuse this service.

It occurs to me that we could also pass something in as an optional argument if we want to conditionally limit the joins. Would that do for reducing the concern around extra joins?

duncan-truss:
> > @Ryan-Koch In terms of API design it would make more sense to me if the api call was `move/{locator}/payment-requests`. If you didn't want to nest them maybe `payment-requests/?moveLocator={locator}` with a query variable instead of a path one.
> > I'm not convinced that reusing the queue call is not more trouble than it's worth. There's a bunch of things we don't need in there such as gbloc filtering, extra joins, and grouping.
> 
> That's fair, I'll switch the url structure around. I do think we need the gbloc filtering and looking ahead at abstract we will need some of the joins. The query is running without the grouping and pagination for this case also (it'll do this when you pass 0s for page and perpage).
> 
> The reason for the gbloc filtering is that would prevent someone from typing in a move code/locator for a move they don't have permission for and then accessing it that way. Having this (and the USMC logic) built in already is what made me think it might be worthwhile to reuse this service.
> 
> It occurs to me that we could also pass something in as an optional argument if we want to conditionally limit the joins. Would that do for reducing the concern around extra joins?

I think allowing perPage and page zero values to turn off pagination is dangerous then someone could request all data within their gbloc.  Adding an unknown perPage or page value should always go to our defaults.  If we must I say make it explicit and don't reuse any parameters coming from the client.

I just think the queries are going to evolve in different ways and it will be harder to follow all the exclusions for when we really don't want queue data returned.

That's a good point about authorization, I think it is slightly different here because we would be checking that the move exists in the gbloc prior to querying for the payment requests.  We would want to return a not found error or a not authorized error in that case instead of an empty array result I think.

Curious how others feel @monfresh @noelledusahel ?

Ryan-Koch:
> > > @Ryan-Koch In terms of API design it would make more sense to me if the api call was `move/{locator}/payment-requests`. If you didn't want to nest them maybe `payment-requests/?moveLocator={locator}` with a query variable instead of a path one.
> > > I'm not convinced that reusing the queue call is not more trouble than it's worth. There's a bunch of things we don't need in there such as gbloc filtering, extra joins, and grouping.
> > 
> > 
> > That's fair, I'll switch the url structure around. I do think we need the gbloc filtering and looking ahead at abstract we will need some of the joins. The query is running without the grouping and pagination for this case also (it'll do this when you pass 0s for page and perpage).
> > The reason for the gbloc filtering is that would prevent someone from typing in a move code/locator for a move they don't have permission for and then accessing it that way. Having this (and the USMC logic) built in already is what made me think it might be worthwhile to reuse this service.
> > It occurs to me that we could also pass something in as an optional argument if we want to conditionally limit the joins. Would that do for reducing the concern around extra joins?
> 
> I think allowing perPage and page zero values to turn off pagination is dangerous then someone could request all data within their gbloc. Adding an unknown perPage or page value should always go to our defaults. If we must I say make it explicit and don't reuse any parameters coming from the client.
> 
> I just think the queries are going to evolve in different ways and it will be harder to follow all the exclusions for when we really don't want queue data returned.
> 
> That's a good point about authorization, I think it is slightly different here because we would be checking that the move exists in the gbloc prior to querying for the payment requests. We would want to return a not found error or a not authorized error in that case instead of an empty array result I think.
> 
> Curious how others feel @monfresh @noelledusahel ?

Hm yeah that's a good point, it's likely not great to allow someone to manually pull the whole gbloc if they figure that out. I think what might be a reasonable compromise is punching up another public function as part of the service that has access to the private FilterOption functions. That way I can reuse that filtering logic without it being wrapped up with those other issues.

pkg/handlers/ghcapi/payment_request.go
```diff
@@ -23,6 +25,37 @@ import (
...
 
+// GetPaymentRequestForMoveHandler gets payment requests associated with a move
+type GetPaymentRequestForMoveHandler struct {
+	handlers.HandlerContext
+	services.PaymentRequestListFetcher
+}
+
+// Handle handles the HTTP handling for GetPaymentRequestForMoveHandler
+func (h GetPaymentRequestForMoveHandler) Handle(params paymentrequestop.GetPaymentRequestsForMoveParams) middleware.Responder {
+	session, logger := h.SessionAndLoggerFromRequest(params.HTTPRequest)
```
monfresh, line 36: Do we want to add authorization here for the right type of user and roles? I'm noticing that our other handlers in this file aren't protected. Shouldn't they be?

pkg/handlers/ghcapi/payment_request_test.go
```diff
@@ -80,6 +80,56 @@ func (suite *HandlerSuite) TestFetchPaymentRequestHandler() {
...
+		context := handlers.NewHandlerContext(suite.DB(), suite.TestLogger())
+		handler := GetPaymentRequestForMoveHandler{
+			HandlerContext:            context,
+			PaymentRequestListFetcher: paymentRequestListFetcher,
+		}
+		response := handler.Handle(params)
+		suite.Assertions.IsType(&paymentrequestop.GetPaymentRequestNotFound{}, response)
+	})
+}
+
```
monfresh, line 132: Thanks for stubbing and mocking!

pkg/handlers/ghcapi/payment_request.go
Ryan-Koch: Oh good point, I forgot to add the role check in. Pushed it on this last one. Going to see if it's a big pain to just fix it for the whole file.

monfresh:
I agree with @duncan-truss about keeping the 2 separate, and not being able to override the pagination, and returning a 403 if they are trying to access things outside their jurisdiction.

pkg/handlers/ghcapi/payment_request.go
Ryan-Koch: Added in checks for the other handlers in the file and adjusted their tests accordingly.

Ryan-Koch:
> I agree with @duncan-truss about keeping the 2 separate, and not being able to override the pagination, and returning a 403 if they are trying to access things outside their jurisdiction.

I wound up going that route with separate function in the same service 👍

noelledusahel:
i like the    path taken here, but i'm not as familiar with the security issues that come out of overriding pagination. i like the idea of introducing conditional logic to limit joins though

swagger/ghc.yaml
```diff
@@ -919,6 +919,45 @@ paths:
...
+          description: Validation error
+          schema:
+            $ref: '#/definitions/ValidationError'
+        '500':
+          description: A server error occurred
+          schema:
+            $ref: '#/responses/ServerError'
+      tags:
+        - paymentRequests
+        - gov
```
duncan-truss, line 957: I don't think we need the gov tag here

pkg/services/payment_request/payment_request_list_fetcher.go
```diff
@@ -116,6 +116,54 @@ func (f *paymentRequestListFetcher) FetchPaymentRequestList(officeUserID uuid.UU
...
+func (f *paymentRequestListFetcher) FetchPaymentRequestListByMove(officeUserID uuid.UUID, locator string) (*models.PaymentRequests, error) {
+	gblocFetcher := officeuser.NewOfficeUserGblocFetcher(f.db)
+	gbloc, gblocErr := gblocFetcher.FetchGblocForOfficeUser(officeUserID)
+	if gblocErr != nil {
+		return &models.PaymentRequests{}, gblocErr
+	}
+
+	paymentRequests := models.PaymentRequests{}
+
+	query := f.db.Q().
```
duncan-truss, line 128: Do we need to EagerPreload the PaymentServiceItems association? I don't see that coming back in the payload.

Ryan-Koch: Yes, and I forgot to include that 😅. Just pushed that change up.

swagger/ghc.yaml
Ryan-Koch: Removed

pkg/handlers/ghcapi/payment_request.go
```diff
@@ -23,6 +25,37 @@ import (
...
+
+	if !session.IsOfficeUser() || !session.Roles.HasRole(roles.RoleTypeTIO) {
+		logger.Error("user is not authenticated with TIO office role")
```
monfresh, line 39: To double check, this tab is meant to be accessible for folks who have both the TOO and TIO roles, right? In order to see the Move details page, they need the TOO role, then to see the payment requests tab, they need the TIO role. If someone only has the TOO role, they should be able to see the Move details, but not the payment requests, right?

pkg/handlers/ghcapi/payment_request.go
```diff
@@ -23,6 +25,37 @@ import (
...
+
+	paymentRequests, err := h.FetchPaymentRequestListByMove(session.OfficeUserID, locator)
+	if err != nil {
+		logger.Error(fmt.Sprintf("Error fetching Payment Request for locator: %s", locator), zap.Error(err))
+		return paymentrequestop.NewGetPaymentRequestNotFound()
+	}
+
+	returnPayload, err := payloads.PaymentRequests(paymentRequests, h.FileStorer())
+	if err != nil {
+		logger.Error(fmt.Sprintf("Error building payment requests payload for locator: %s", locator), zap.Error(err))
```
monfresh, line 53: do we want to return a non-200 error to the front end if we can't build the payload?

pkg/handlers/ghcapi/payment_request_test.go
```diff
@@ -36,13 +38,17 @@ func (suite *HandlerSuite) TestFetchPaymentRequestHandler() {
 		UpdatedAt: time.Now(),
 	}
 
+	officeUser := testdatagen.MakeDefaultOfficeUser(suite.DB())
```
monfresh, line 41: stub?

pkg/handlers/ghcapi/payment_request_test.go
```diff
@@ -59,11 +65,38 @@ func (suite *HandlerSuite) TestFetchPaymentRequestHandler() {
 		okResponse := response.(*paymentrequestop.GetPaymentRequestOK)
 		suite.Equal(paymentRequestID.String(), okResponse.Payload.ID.String())
 	})
+
+	suite.T().Run("failed fetch for payment request - forbidden", func(t *testing.T) {
+		officeUserTOO := testdatagen.MakeDefaultOfficeUser(suite.DB())
```
monfresh, line 70: stub?

pkg/handlers/ghcapi/payment_request_test.go
```diff
@@ -80,8 +113,85 @@ func (suite *HandlerSuite) TestFetchPaymentRequestHandler() {
 	})
 }
 
+func (suite *HandlerSuite) TestGetPaymentRequestsForMoveHandler() {
+	prUUID, _ := uuid.NewV4()
+	paymentRequests := models.PaymentRequests{models.PaymentRequest{ID: prUUID}}
+	officeUser := testdatagen.MakeDefaultOfficeUser(suite.DB())
```
monfresh, line 119: stub?

pkg/handlers/ghcapi/payment_request_test.go
```diff
@@ -80,8 +113,85 @@ func (suite *HandlerSuite) TestFetchPaymentRequestHandler() {
...
+			PaymentRequestListFetcher: paymentRequestListFetcher,
+		}
+		response := handler.Handle(params)
+		suite.Assertions.IsType(&paymentrequestop.GetPaymentRequestsForMoveForbidden{}, response)
+	})
+}
+
 func (suite *HandlerSuite) TestUpdatePaymentRequestStatusHandler() {
 	paymentRequestID, _ := uuid.FromString("00000000-0000-0000-0000-000000000001")
+	officeUser := testdatagen.MakeDefaultOfficeUser(suite.DB())
```
monfresh, line 191: stub?

pkg/handlers/ghcapi/payment_request_test.go
```diff
@@ -120,6 +229,39 @@ func (suite *HandlerSuite) TestUpdatePaymentRequestStatusHandler() {
 
 	})
 
+	suite.T().Run("failed status update of payment request - forbidden", func(t *testing.T) {
+		officeUserTOO := testdatagen.MakeDefaultOfficeUser(suite.DB())
```
monfresh, line 233: stub?

pkg/services/payment_request/payment_request_list_fetcher_test.go
```diff
@@ -248,6 +275,13 @@ func (suite *PaymentRequestServiceSuite) TestFetchPaymentRequestListWithPaginati
 	suite.NoError(err)
 	suite.Equal(1, len(*expectedPaymentRequests))
 	suite.Equal(2, count)
+
+	// Pagination disabled using zero values
+	expectedPaymentRequests, count, err = paymentRequestListFetcher.FetchPaymentRequestList(officeUser.ID, &services.FetchPaymentRequestListParams{Page: swag.Int64(0), PerPage: swag.Int64(0)})
+
+	suite.NoError(err)
+	suite.Equal(2, len(*expectedPaymentRequests))
+	suite.Equal(2, count)
```
monfresh, line 284: What are we trying to test here? That if someone sets page and per page to 0, it uses the defaults of page=1 and per_page = 20? If so, I don't think this test communicates that, because it makes it look like we're supposed to get all results, which I don't think is what we want. It would be better if we could verify that Pop is receiving `Paginate` with arguments `(1, 20)`.

Ryan-Koch: Oh this test should go away, thanks for pointing this out. This isn't scenario isn't part of the implementation anymore.

pkg/handlers/ghcapi/payment_request_test.go
Ryan-Koch: I had initially done that with some of these. Stubbing them seems to cause the handler to fail the permissions check though unfortunately. I'll poke at it a bit to see if I can work around it.

pkg/handlers/ghcapi/payment_request.go
Ryan-Koch: I believe this is correct. The TIO role would be what you need to ultimately get payment request info.

pkg/handlers/ghcapi/payment_request.go
Ryan-Koch: yes, that's a good point. Pushed up a change to do that.

pkg/handlers/ghcapi/payment_request_test.go
Ryan-Koch: Ah I figured it out, the check to see if we're using an office user is looking for the OfficeUserID uuid field, which is a zero value if you don't set it. Setting it ahead of time before passing it into the creation method with the stub argument does the trick. Just pushed up this change.

duncan-truss (approved):
Thanks for reworking this it looks good to me! Could we add the createdAt field to the PaymentRequest payload?  It's needed for the date display in the component.

Ryan-Koch:
> Thanks for reworking this it looks good to me! Could we add the createdAt field to the PaymentRequest payload? It's needed for the date display in the component.

Just pushed the change to add this to the payload. I'll merge once the tests pass.

monfresh (approved):
LGTM. Thanks for fixing the stubs.

PR #5392, opened 2020-12-02
MB-5590 Add UpdateMTOServiceItem service object

sandy-wright:
## Description

This PR adds an `UpdateMTOServiceItem` function to the services package. It also sets up this service object with the validator pattern used for `UpdateMTOAgent` (old PR): https://github.com/transcom/mymove/pull/4860

In addition to adding this service object, I also made some updates to the `UpdateMTOAgent` service object and validators that align with feedback I've gotten in the past. I did this so that the validation pattern could stay updated as I implemented it again with `UpdateMTOServiceItem`.

Note that there is no endpoint to test this service object currently. 

## Reviewer Notes

I wanted to keep validation bare so that I wasn't making too many assumptions, but I wasn't sure if I was missing anything for the `SITDeparture` validation in particular. Let me know if you think of any other rules for this update that I should implement! (Especially @shimonacarvalho)

## Setup

Right now we only have the tests to go off of for this code. Feel free to write an adhoc script to call this service object against the dev database, but the goal is for the tests to be as comprehensive as possible.

You can run:

```
make server_test
```

for the full suite of tests. Or you can run:

```
go test ./pkg/services/mto_service_item -v
```

for the `mto_service_item` package's tests. If you want to run each set of tests individually, these are the functions you'll need to call (using the `-testify.m` flag):

```
TestMTOServiceItemUpdater
TestValidateUpdateMTOServiceItem
TestUpdateMTOServiceItemData
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5590) for this change.


pkg/services/mto_agent/mto_agent_validators_test.go
```diff
@@ -149,6 +139,8 @@ func (suite *MTOAgentServiceSuite) TestUpdateMTOAgentData() {
 			availabilityChecker: checker,
 			verrs:               validate.NewErrors(),
 		}
+		_ = agentData.checkShipmentID() // this test should pass regardless of potential errors here
```
shimonacarvalho, line 142: If you don't need the return value, do you need `_ =` ? Perhaps you can just do 
`agentData.checkShipmentID()`

pkg/services/mto_agent/mto_agent_validators.go
```diff
@@ -140,12 +137,9 @@ func (v *updateMTOAgentData) getVerrs() error {
...
 // setNewMTOAgent compares updatedAgent and oldAgent and updates a new MTOAgent instance with all data
-// (changed and unchanged) filled in
-func (v *updateMTOAgentData) setNewMTOAgent(newAgent *models.MTOAgent) error {
-	err := deepcopy.Copy(&newAgent, &v.oldAgent)
-	if err != nil {
-		return fmt.Errorf("error copying agent data %w", err)
-	}
+// (changed and unchanged) filled in. Does not return an error, data must be checked for validation before this step.
+func (v *updateMTOAgentData) setNewMTOAgent() *models.MTOAgent {
+	newAgent := v.oldAgent
```
shimonacarvalho, line 142: Does this make a copy? or are they pointing at the same thing?

pkg/services/mto_service_item/mto_service_item_validators.go
```diff
@@ -0,0 +1,264 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = serviceItemData.checkPaymentRequests()
+	if err != nil {
+		return err
+	}
+
+	err = serviceItemData.checkSITDeparture()
```
sandy-wright, line 72: Add one sentence comment with each of these to explain what these checks are doing

pkg/services/mto_service_item/mto_service_item_validators.go
```diff
@@ -0,0 +1,264 @@
...
+	return nil
+}
+
+// checkPrimeAvailability checks that the service item is connected to a Prime-available move
+func (v *updateMTOServiceItemData) checkPrimeAvailability() error {
+	// NOTE: We cannot use the MoveTaskOrderChecker here because this service uses QueryBuilder and doesn't have access
+	// to the DB
+	var move models.Move
+	queryFilters := []services.QueryFilter{
+		query.NewQueryFilter("id", "=", v.updatedServiceItem.MoveTaskOrderID),
```
sandy-wright, line 114: Use the MoveTaskOrderID from the oldServiceItem

pkg/services/mto_service_item/mto_service_item_validators.go
```diff
@@ -0,0 +1,264 @@
...
+func (v *updateMTOServiceItemData) checkSITDeparture() error {
+	if v.updatedServiceItem.SITDepartureDate == nil || v.updatedServiceItem.SITDepartureDate == v.oldServiceItem.SITDepartureDate {
+		return nil // the SITDepartureDate isn't being updated, so we're fine here
+	}
+
+	if v.oldServiceItem.ReService.Code == models.ReServiceCodeDDDSIT || v.oldServiceItem.ReService.Code == models.ReServiceCodeDOPSIT {
+		return nil // the service item is a SIT departure service, so we're fine
+	}
+
+	// TODO do we need to check anything else?
```
sandy-wright, line 160: Remove this, this check is fine

pkg/services/mto_service_item/mto_service_item_validators.go
```diff
@@ -0,0 +1,264 @@
...
+		query.NewQueryFilter("mto_service_item_id", "=", v.updatedServiceItem.ID),
+	}
+
+	err := v.builder.FetchOne(&paymentServiceItem, queryFilters)
+	if err == nil && paymentServiceItem.ID != uuid.Nil {
+		return services.NewConflictError(v.updatedServiceItem.ID,
+			"- this service item has an existing payment request and can no longer be updated.")
+	} else if err != nil && !strings.Contains(err.Error(), "sql: no rows in result set") {
+		return err
+	}
```
sandy-wright, line 180: Add a comment that clarifies that the third error case is good because it means we didn't find anything

pkg/services/mto_agent/mto_agent_validators_test.go
sandy-wright: Turns out that the Go compiler complains about omitting the underscores because it views it as an "unhandled error," so I'm going to keep these after all to prevent adding warnings. But it was a good idea :/

chrisrcoles (approved):
LGTM! I love all the added validators and the tests added! Thanks so much for the attention to detail.

pkg/services/mto_agent/mto_agent_validators_test.go
shimonacarvalho: Huh interesting!

PR #5395, opened 2020-12-03
[MB-5588 // MB-5656 // MB-5657] Add an endpoint that call service to check MTOs against fake data

harshamurthy:
## Description

This PR adds an endpoint `/move-task-orders/hide` that runs the Move task order hide non fake service found in [this PR](https://github.com/transcom/mymove/pull/5397/files)

## Set Up

`make db_dev_e2e_populate && make server_run`
```ssh
curl \
    --insecure \
    --cert ./config/tls/devlocal-mtls.cer \
    --key ./config/tls/devlocal-mtls.key \
    --request "PATCH" \
    "https://primelocal:9443/support/v1/move-task-orders/hide"
```

Expected output is an array of MTOs that did not include fake data. It's a lot of data back but can be reviewed to see fields not using fake data https://docs.google.com/spreadsheets/d/1u1NO_ZWvKJc2ylOSF5-4mcm6Eg5X2zu7c_P-X4lDrE4/edit#gid=524168919

 have been satisfied.

## References

* [Jira story MB-5588](https://dp3.atlassian.net/browse/MB-5588) for this change
* [Jira story MB-5656](https://dp3.atlassian.net/browse/MB-5588) for this change
* [Jira story MB-5657](https://dp3.atlassian.net/browse/MB-5588) for this change

## Screenshots


pkg/handlers/supportapi/move_task_order.go
```diff
@@ -74,6 +74,28 @@ func (h MakeMoveTaskOrderAvailableHandlerFunc) Handle(params movetaskorderops.Ma
 	return movetaskorderops.NewMakeMoveTaskOrderAvailableOK().WithPayload(moveTaskOrderPayload)
 }
 
+// HideNonFakeMoveTaskOrdersHandlerFunc calls service to hide MTOs that are not using fake data
+type HideNonFakeMoveTaskOrdersHandlerFunc struct {
+	handlers.HandlerContext
+	services.MoveTaskOrderHider
+}
+
+// Handle updates the prime availability of a MoveTaskOrder
```
pearl-truss, line 83: Nit, can the comment be updated to reflect this function

swagger/support.yaml
```diff
@@ -188,6 +188,40 @@ paths:
           $ref: '#/responses/UnprocessableEntity'
         '500':
           $ref: '#/responses/ServerError'
+  '/move-task-orders/hide':
+    parameters: []
+    get:
```
reggieriser, line 193: We should probably make this a `patch` instead of a `get` since `get` is supposed to not change any state.  Looks like we do a `patch` for the `available-to-prime` endpoint which feels similar.

pkg/handlers/supportapi/move_task_order_test.go
```diff
@@ -62,6 +65,43 @@ func (suite *HandlerSuite) TestListMTOsHandler() {
 
 }
 
+func (suite *HandlerSuite) TestHideNonFakeMoveTaskOrdersHandler() {
+	request := httptest.NewRequest("GET", "/move-task-orders/hide", nil)
```
reggieriser, line 69: `PATCH` here too (see other comment about `GET` vs `PATCH`).

pkg/handlers/supportapi/move_task_order_test.go
```diff
@@ -62,6 +65,43 @@ func (suite *HandlerSuite) TestListMTOsHandler() {
...
+	var moves models.Moves
+	moves = append(moves, testdatagen.MakeMove(suite.DB(), testdatagen.Assertions{}))
+
+	suite.T().Run("successfully hide fake moves", func(t *testing.T) {
+		mockHider := &mocks.MoveTaskOrderHider{}
+		handler := HideNonFakeMoveTaskOrdersHandlerFunc{
+			context,
+			mockHider,
+		}
+		mockHider.On("Hide").Return(moves, nil)
```
reggieriser, line 83: I tend to test the golden paths by calling the actual function instead of a mock and just use mocks for hard-to-test paths (like error conditions).  Is that possible here without too much additional data setup?

reggieriser (commented):
@pearl-truss Left a few comments on things @harshamurthy and I talked about in person yesterday.

pkg/handlers/supportapi/move_task_order_test.go
```diff
@@ -62,6 +65,51 @@ func (suite *HandlerSuite) TestListMTOsHandler() {
...
 }
 
+func (suite *HandlerSuite) TestHideNonFakeMoveTaskOrdersHandler() {
+	request := httptest.NewRequest("PATCH", "/move-task-orders/hide", nil)
+	params := move_task_order.HideNonFakeMoveTaskOrdersParams{
+		HTTPRequest: request,
+	}
+	context := handlers.NewHandlerContext(suite.DB(), suite.TestLogger())
+
+	suite.T().Run("successfully hide fake moves", func(t *testing.T) {
```
jacquelineIO, line 75: Is this just really calling the API endpoint successfully? doesn't seem like moves are being removed?

pkg/handlers/supportapi/move_task_order_test.go
```diff
@@ -62,6 +65,51 @@ func (suite *HandlerSuite) TestListMTOsHandler() {
...
+
+		for idx, mto := range mtoRequestsPayload {
+			suite.Equal(strfmt.UUID(moves[idx].ID.String()), mto.ID)
+		}
+	})
+
+	suite.T().Run("unsuccessfully hide fake moves", func(t *testing.T) {
+		var moves models.Moves
+		moves = append(moves, testdatagen.MakeMove(suite.DB(), testdatagen.Assertions{}))
+		mockHider := &mocks.MoveTaskOrderHider{}
```
jacquelineIO, line 100: @pearl-truss @gidjin  what's the pros of calling a mock here?

swagger/support.yaml
```diff
@@ -188,6 +188,40 @@ paths:
           $ref: '#/responses/UnprocessableEntity'
         '500':
           $ref: '#/responses/ServerError'
+  '/move-task-orders/hide':
+    parameters: []
```
jacquelineIO, line 192: @reggieriser @pearl-truss  does this mean this is a patch call but we have no parameters, however we are changing data with this call?

pkg/handlers/supportapi/move_task_order_test.go
pearl-truss: the moves shouldn't be removed but they should have the show flag switched to false. I can add a test for that

pearl-truss: Oh I remember, there are test to check that moves are hidden inside of the Hide service (https://dp3.atlassian.net/browse/MB-5588). Do we need to still test that here?

pkg/handlers/supportapi/move_task_order_test.go
pearl-truss: I added a mock since this service is tested in another file

pkg/handlers/supportapi/move_task_order_test.go
jacquelineIO: No, I think this makes sense, I just wasn't following. So only moves returned are the ones that have been hidden. And that's what this test is showing makes sense. 

It does make sense to verify the flag, not sure if that is happening some where else so I'm not certain it needs to be here.

jacquelineIO (approved):
The return params seem like too much for this endpoint that it isn't very useful. I'm not clear why this approach was taken. But overall the code is good and is holding up other PRs.

swagger/support.yaml
reggieriser: The initial implementation used a GET, but this actually changes some state, so it seemed a PATCH was better.  I'm guessing the `parameters` section here could be removed entirely since we're not passing any parameters as part of the URL for instance.

PR #5397, opened 2020-12-03
[MB-5654] Create a service to hide fake data

pearl-truss:
## Description

This PR introduces a service to hide move task orders that contain data that violates our fake data values and conventions. 

## Reviewer Notes

Some questions that were brought up

- Error handling inside of `Hide`, what to return when there's an error?
- Which records and relations need to be Eager loaded?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5654) for this change


cheatsheet
```diff
@@ -0,0 +1,11 @@
+# make support-update-payment-request-status
```
jacquelineIO, line 1: I'm guessing this file was checked in on accident

jacquelineIO (commented):
Looking for file `move_task_order_updater.go`. I don't see it in the files for the PR. But when I look in Goland IDE and view the PR I see it. 

Specifically I was looking to see the comments left around lines 112-113
```
		// CreateMTOServiceItem may have updated the mto status so refetch as to not return incorrect status
		// TODO: Modify CreateMTOServiceItem to return the updated move or refactor to operate on the passed in reference
```
and follow-on chores or something to clean it up later.

jacquelineIO (approved):
Looks good. I don't think I can kick the tires yet by running the CLI since that isn't merged in here.

pkg/services/move_task_order/move_task_order_hide_non_fake.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	if s.SecondaryPickupAddress != nil {
+		ok, err := isValidFakeModelAddress(s.SecondaryPickupAddress)
+		if err != nil {
+			return false, err
+		}
+		if ok == false {
+			return false, nil
+		}
```
jacquelineIO, line 148: I know this has already been merged, but is this allowing and empty string?

jacquelineIO: We don't need to hide moves if it just an empty string.

PR #5408, opened 2020-12-04
[MB-5594] Add Updatemtoserviceitem handler 

shimonacarvalho:
## Description

We now have a live endpoint for updating MTOServiceItems, it connects to the updateMTOServiceItem service object. 
You can use this to update a DDDSIT or DOPSIT departure date. 

### Unit test version
Two tests were added
`go test ./pkg/handlers/primeapi/ --testify.m TestCreateMTOServiceItemDDFSITHandler -v`
`go test ./pkg/handlers/primeapi/ --testify.m TestUpdateMTOServiceItemDOPSIT -v`

### Real world version
Use postman or prime-api-client. Pick a move and shipment to which you will add the SIT. 
You will need to ensure that the move you selected has status set to APPROVED.

Hit the createMTOServiceItem endpoint with a DOFSIT, this should create all three items, DOFSIT, DOASIT and DOPSIT. However the response is not perfect yet as this is being worked on in another ticket. 

Once you have a DOPSIT in your DB, hit the updateMTOServiceItem end point using the ID of the DOPSIT item. 
Send a "sitDepartureDate" value and you should see it updated in the response. 

## Reference
* [MB-5594 - SIT Service Items: updateMTOServiceItem endpoint - Add Prime API handler function](https://dp3.atlassian.net/browse/MB-5594)

### DIFF
--- a/pkg/handlers/primeapi/mto_service_item_test.go
+++ b/pkg/handlers/primeapi/mto_service_item_test.go
@@ -565,3 +565,128 @@ func (suite *HandlerSuite) TestUpdateMTOServiceItemDDDSIT() {
 	})
 
 }
+
+func (suite *HandlerSuite) TestUpdateMTOServiceItemDOPSIT() {
+
+	// Under test: updateMTOServiceItemHandler.Handle function
+	//             MTOServiceItemUpdater.Update service object function
+	// SETUP
+	// Create the service item in the db for dofsit and DOPSIT
+	timeNow := time.Now()
+	dopsit := testdatagen.MakeMTOServiceItem(suite.DB(), testdatagen.Assertions{
+		Move: models.Move{
+			AvailableToPrimeAt: &timeNow,
+		},
+		MTOServiceItem: models.MTOServiceItem{
+			SITEntryDate: swag.Time(time.Now()),
+		},
+		ReService: models.ReService{
+			Code: "DOPSIT",
+		},
+	})
+
+	// Create the payload with the desired update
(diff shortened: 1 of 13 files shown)


abbyoung (commented):
It works!!! 🥳 

I have a Q about the endpoint's body -- why do we need to include the id in the body, as well as a param in the path? I can't figure out why we need it here but not in every endpoint.

pkg/handlers/primeapi/mto_service_item_test.go
```diff
@@ -439,36 +441,252 @@ func (suite *HandlerSuite) TestCreateMTOServiceItemDDFSITHandler() {
...
-	//             Receive a 501 - Not Implemented Error
+	// Under test: updateMTOServiceItemHandler.Handle function
+	//             MTOServiceItemUpdater.Update service object function
 	// SETUP
-	// Create the payload
-	id := uuid.Must(uuid.NewV4())
-	payload := &primemessages.UpdateMTOServiceItemSIT{
-		ReServiceCode:    "DDFSIT",
-		SitDepartureDate: *handlers.FmtDate(time.Now()),
+	// Create the service item in the db for dofsit and dddsit
```
abbyoung, line 449: i think might be copypasta and should say ddasit and dddsit?

abbyoung: actually i'm not sure, i've confused myself!!

abbyoung (approved):
looks good and works as expected! 

let's pair to resolve the payload/model merge conflicts with my PR once they're both ready.

PR #5422, opened 2020-12-08
Remove md-toc because it's not used

monfresh:
## Description

In a previous PR, we added both `markdown-toc` and `md-toc` to our package.json in order to fix a Circle CI issue that was blocking all builds. We added both dependencies because that's what worked for the Infra team, who ran into the same issue. Now that we're unblocked, we can remove `md-toc` because it doesn't look like we're using it.


monfresh:
@suzubara @gidjin It looks like `md-toc` is needed after all. Removing it causes the same issue as before.

PR #5431, opened 2020-12-08
[MB-4634] Edit allowances route

Ronolibert:
## Description

Add Edit Allowances button to Allowances table and the corresponding page

## Reviewer Notes
- did some slight refactoring to fix the styling for the allowances table
- will verify with design, but I hid the edit button on the ShipmentApprovalPreview page

## Setup
`make db_dev_e2e_populate`
`make office_client_run`
`make server_run`
Log into TOO user
Visit any move
Scroll down to Allowances Table and click `Edit Allowances`
Expected result: Should be taken to allowances page which has a sidebar containing Save/Cancel button

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4634) for this change

## Screenshots

image
image


brooksr4 (approved):
Looks good and meets acceptance criteria

src/components/Office/AllowancesTable/AllowancesTable.jsx
```diff
@@ -22,11 +25,18 @@ const AllowancesTable = ({ info }) => {
...
   return (
-    <div>
+    <div className={styles.MoveDetailTable}>
       <div className="stackedtable-header">
         <div>
           <h4>Allowances</h4>
         </div>
+        {showEditBtn && (
+          <div>
+            <Link className="usa-button usa-button--secondary" data-testid="edit-allowances" to="allowances">
```
duncan-truss, line 35: Do we want to use the react-uswds Link here since we don't care about whether the link is an active route? https://trussworks.github.io/react-uswds/?path=/docs/link--styled-as-button#styled-as-button

src/pages/Office/TXOMoveInfo/TXOMoveInfo.jsx
```diff
@@ -63,10 +64,15 @@ const TXOMoveInfo = () => {
           <Route path="/moves/:moveOrderId/details" exact>
             <MoveDetails />
           </Route>
+
           <Route path="/moves/:moveOrderId/orders" exact>
             <MoveOrders />
           </Route>
 
+          <Route path="/moves/:moveOrderId/allowances" exact>
```
duncan-truss, line 72: We should add this to the hideNav matchPath at the top of the file similar to what we do for the orders route.

https://github.com/transcom/mymove/pull/5431/files#diff-6f4d707640a1126e3e9aeb3c5a8a260c5bf4b8e2e90c5435bf8c07684b9a0ae5R21

src/components/Office/AllowancesTable/AllowancesTable.jsx
Ronolibert: Oooh, I didn't know that existed. What are the advantages of using the react-uswds link over the react-router one? I looked and it seems the react-uswds component is just an `a` tag

One thing we do lose if we go with the react-uswds component is all the built in props like `to`. Instead of providing a path of `allowances` which inherently uses the baseUrl of the router, we'd need to pass in the full path of `moves/:moveId/details`.Also, I think it'd be easier to stay in the react-router as the router instead of using react-router in one place and regular links in the other just to keep api usage consistent. The times we've used regular anchor tags is for linking to external docs

src/pages/Office/MoveAllowances/MoveAllowances.test.jsx
```diff
@@ -0,0 +1,121 @@
...
+          url: '/storage/user/1/uploads/2?contentType=application%2Fpdf',
+        },
+      },
+    };
+  },
+}));
+
+describe('MoveAllowances page', () => {
+  const wrapper = mount(
+    <MockProviders initialEntries={['moves/1000/orders']}>
```
duncan-truss, line 107: I think you meant this to be `moves/1000/allowances`

src/pages/Office/TXOMoveInfo/TXOMoveInfo.jsx
Ronolibert: Didn't see that! Added that now

src/components/Office/AllowancesTable/AllowancesTable.jsx
duncan-truss: Yea I saw we are almost exclusively using the `react-router-dom` version.  I see your point, I'm not entirely sure what props we really need if it matches our Router in Office/index.jsx that component will get the :moveOrderId parameter right?

cypress/integration/office/txo/tooFlows.js
```diff
@@ -158,4 +158,28 @@ describe('TOO user', () => {
...
+    cy.wait(['@getSortedMoveOrders']);
+    cy.contains(moveLocator).click();
+    cy.url().should('include', `/moves/${moveOrderId}/details`);
+
+    // Move Details page
+    cy.wait(['@getMoveTaskOrders', '@getMTOShipments', '@getMTOServiceItems']);
+
+    // Edit allowances page | Save
+    cy.get('[data-testid="edit-allowances"]').contains('Edit Allowances').click();
+    cy.url().should('include', `/moves/${moveOrderId}/allowances`);
```
duncan-truss, line 176: Thanks for adding this test! Do we need to wait for the document and orders queries to resolve here similar the Move Details page.

src/components/Office/MoveDetailTable.module.scss
```diff
@@ -0,0 +1,36 @@
+@import 'shared/styles/_basics';
+@import 'shared/styles/_mixins';
+@import 'shared/styles/colors';
+@import 'shared/styles/_custom';
+
+.MoveDetailTable {
```
duncan-truss, line 6: Nice thanks for consolidating these styles, it fixed the approval preview layout as well.

src/pages/Office/MoveAllowances/MoveAllowances.jsx
```diff
@@ -0,0 +1,97 @@
...
+import { HistoryShape, MatchShape } from 'types/router';
+import { useOrdersDocumentQueries } from 'hooks/queries';
+
+const MoveAllowances = ({ history, match }) => {
+  const { moveOrderId } = match.params;
+
+  const { upload, isLoading, isError } = useOrdersDocumentQueries(moveOrderId);
+
+  const handleClose = () => {
+    history.push(`/moves/${moveOrderId}/details`);
```
duncan-truss, line 22: I could see a case where we come here from the TIO payment-requests page so it might not make sense to return you to the details page.  Just a thought (not blocking) we might want to turn this handler into a configurable prop.

duncan-truss (approved):
Great work!

src/components/Office/MoveDetailTable.module.scss
```diff
@@ -0,0 +1,36 @@
...
+  @include u-padding-y(2);
+  :global .stackedtable-header {
+    max-width: unset;
+    h4 {
```
duncan-truss, line 10: I think we may be missing the bold styling here according to the mock https://share.goabstract.com/7b0ee857-f97a-465d-9243-5005f91e9ea4?collectionLayerId=b2b6c7bb-2be5-48d5-b102-9dbff3d4fb03&mode=design

src/components/Office/AllowancesTable/AllowancesTable.jsx
Ronolibert: Yup! The params are caught in the same way, its just that the Link component from react-router-dom has some convenience props like `to` for cleaner paths or `replace` to overwrite the history stack and things like that

cypress/integration/office/txo/tooFlows.js
Ronolibert: I think so, just for a reliability standpoint. There could be a situation where the API isn't done loading but the UI is in its initial render and it could introduce some flakiness

src/pages/Office/MoveAllowances/MoveAllowances.jsx
Ronolibert: Right! Was thinking about that initially but currently its only referenced in that one place but we can keep it mind once its a part of more flows

PR #5437, opened 2020-12-08
[MB-5737] add cli to hide mtos

pearl-truss:
## Description

This PR introduces a new command support-hide-mto to hide moves not using fake data

## Reviewer Notes

Any reviewer that has a CAC can test this in experimental. Just let me know when you're ready to do so and I can deploy to experimental. Or alternatively, anyone else can also do this. 

I added steps below on how to test both locally and on experimental 

## Setup

```sh
```

Locally 
```sh
prime-api-client --insecure support-hide-non-fake-mtos
```

Experimental (once deployed https://github.com/transcom/mymove/wiki/deploy-to-experimental)
```sh
prime-api-client  --cac --hostname api.exp.move.mil --port 443  support-hide-non-fake-mtos | jq 
```


gidjin:
> Experimental (once deployed https://github.com/transcom/mymove/wiki/deploy-to-experimental)
> 
> ```shell
> prime-api-client  --cac --hostname api.exp.move.mil --port 443  support-hide-mtos | jq 
> ```

Should this second command also be `support-hide-non-fake-mtos` vis `support-hide-mtos`?

gidjin (commented):
Code looks good and confirmed it works locally, it hide all moves as expected. I created 2 moves one with valid fake data one with invalid and it hid just the invalid one as expected. I can test this on exp as well. I'll approve once that's done unless someone else can before me

gidjin (changes requested):
Okay so I ran this in exp and had a failure I'll slack you the log trace. From what I can tell from the log trace `pkg/handlers/supportapi/internal/payloads/model_to_payload.go` line 39 is where the failure came from and seems to indicate that the ContractorID is nil for some of the records, which could be for some old data. Since this is likely true in staging as well I'm thinking we need to handle the error in the handler at least. What are your thoughts on this @pearl-truss ?

pearl-truss:
> Okay so I ran this in exp and had a failure I'll slack you the log trace. From what I can tell from the log trace `pkg/handlers/supportapi/internal/payloads/model_to_payload.go` line 39 is where the failure came from and seems to indicate that the ContractorID is nil for some of the records, which could be for some old data. Since this is likely true in staging as well I'm thinking we need to handle the error in the handler at least. What are your thoughts on this @pearl-truss ?

This is ready for review again. I updated the handler to nil check the ContractorID

pkg/handlers/supportapi/move_task_order.go
```diff
@@ -91,6 +91,12 @@ func (h HideNonFakeMoveTaskOrdersHandlerFunc) Handle(params movetaskorderops.Hid
 		return movetaskorderops.NewHideNonFakeMoveTaskOrdersInternalServerError().WithPayload(payloads.InternalServerError(handlers.FmtString(err.Error()), h.GetTraceID()))
 	}
 
+	for _, mto := range mtos {
+		if mto.ContractorID == nil {
+			return movetaskorderops.NewHideNonFakeMoveTaskOrdersConflict().WithPayload(payloads.ClientError(handlers.ConflictErrMessage, "ContractorID must be present on the MTO", h.GetTraceID()))
```
jacquelineIO, line 96: @pearl-truss @gidjin Do we really want to error out here? I don't think this is really our concern. I would think we should put out a warning message, but doesn't seem like this is a reason for this code to error out. I would think to skip it and continue if possible, but if this isn't possible then I guess that is a reason why.

gidjin: I don't think we want to error out, I think we want to continue, since there wouldn't be much we could do in this case. Currently it threw an error even though moves were hidden. Is it possible to catch the error and still render a json response of the moves affected?

pearl-truss: Okay, kind of combining the 2 meesages, it sounds like we want to

-  Check if the ContractorID is nil, and log a warning message 
- Return all MTOs that had fake data, minus the ones missing a contractorID

gidjin (approved):
The changes look good, tested locally by changing a move with unapproved fake data to have a null contractor id and there were no errors running and the move with the null contractor id was marked `show == f`

PR #5442, opened 2020-12-09
MB-5722 Make stack trace length configurable

monfresh:
## Description

One of the STIG requirements is to limit the number of lines in a
stack trace to only show the first 3 code paths, and to make the
number of lines configurable, in case we need to increase or decrease
it. The configuration should be via an ENV var that can be changed
without having to deploy new code. This requires creating a new AWS
SSM parameter for this, which the Infra team kindly already took care
of.

**How**

* In order to get access to the stack trace, we need to define a new
custom Zap encoder, and do the stack trace manipulation in the
encoder's `EncodeEntry` function. We need to create two custom
encoders because development uses the console encoder, and deployed
environments use the JSON encoder.

* Instead of adding yet another parameter to the logging `Config`
function, which is a code smell, especially when there are multiple
parameters of the same type, I refactored `Config` to accept
functional options instead. Advantages:
- The order of the options doesn't matter, so no more risk of messing
things up
- You don't have to pass in all the options if you don't need them
all. This means if we add a new option in the future that is only
needed in `serve.go`, we won't have to make changes everywhere
`logging.Config` is called
- Options with descriptive names makes it clear what you are passing

* In order to limit the stack trace to the first 3 code paths, and
also keep it useful for troubleshooting, I modified the stack trace
so that it only shows code paths within the "mymove" packages.

**Bonus**
I updated the dev logger to colorize the log levels, so it's easier
to spot the ERRORs in red.

## Reviewer Notes

1. Replace `pkg/services/move_order/move_order_fetcher.go` with the contents of [this gist](https://gist.github.com/monfresh/c04364b04c47387c95e0dd4f1ed90fb9)
1. Run `make server_run`
2. In a new terminal tab, run `make office_client_run`
3. Use the local sign in to sign in with the TOO user
- [ ] Verify that you see a stack trace, and that it only contains
code paths from "mymove/pkg", and that it only shows 6 lines by
default
- [ ] Verify that all log lines include the `git-commit` and `git-branch` fields
4. Add `export STACKTRACE_LENGTH=10` to your `.envrc.local`
5. Stop the server
6. Run `direnv allow`
7. Run `make server_run`
8. Bring focus back on the browser tab/window that has the office app
- [ ] Verify that the stack trace is now 10 lines long
9. Update STACKTRACE_LENGTH in your `.envrc.local` to a string instead of an integer
10. Repeat steps 5-7
- [ ] Verify that there is a FATAL error that says the length has to be greater than 6.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5722) for this change
* https://www.sohamkamani.com/golang/options-pattern/
* https://dave.cheney.net/2019/09/24/be-wary-of-functions-which-take-several-parameters-of-the-same-type
* https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

## Screenshots
**After**:
Screen Shot 2020-12-09 at 1 43 23 PM

**Before**:
Screen Shot 2020-12-09 at 1 46 42 PM


monfresh:
I just noticed a bug by looking at the screenshots. My PR lost the `git-branch` and `git-commit` fields. I will look into it.

pkg/cli/logging.go
```diff
@@ -95,3 +106,14 @@ func LogLevelIsDebug(v *viper.Viper) bool {
 	logLevel := strings.ToLower(v.GetString(LoggingLevelFlag))
 	return logLevel == LoggingLevelDebug
 }
+
+// ValidateStacktraceLength validates STACKTRACE_LENGTH is an integer between 1 and 50
+func ValidateStacktraceLength(v *viper.Viper, flagname string) error {
+	stacktraceLength := v.GetInt(flagname)
+
+	if stacktraceLength <= 0 || stacktraceLength > 50 {
```
duncan-truss, line 114: So there's no way to turn the stacktrace limit off, we're capped at 50?

monfresh: Yes, but I wanted a way to validate that the env var was set to an integer, and not a string, for example, while also allowing us to set lower and upper limits. I meant to update this to have a minimum of 6. Anything less than that I don't think would result in a helpful stack trace. I chose 50 arbitrarily. We can certainly pick a higher limit. From my testing with the GetMovesQueue handler exception, the stack trace is only 28 lines long. I don't think we'll need more than 50, but I'm happy to bump this up to 100.

pkg/logging/log.go
```diff
@@ -4,22 +4,81 @@ import (
...
+		c.LoggingLevel = level
+	}
+}
+
+// WithStacktraceLength provides an option to pass in the stack trace length
+func WithStacktraceLength(length int) ZapConfigOption {
+	return func(c *ZapConfig) {
+		c.StacktraceLength = length
+	}
+}
```
duncan-truss, line 42: I'm not sure what the real benefit of these option functions are if they are just assigning values to struct fields that are already public?  With a basic constructor we can see the 3 args in the signature and then make the struct fields private.

monfresh: I was trying to make it clearer which option is being passed in, without having to worry about the order of arguments, given that 2 of them are strings, and to be able to pass in only some options if they're not all needed. Can that be achieved without using option functions?

pkg/logging/log.go
```diff
@@ -4,22 +4,81 @@ import (
...
+
 // Config configures a Zap logger based on the environment string and debugLevel
-func Config(env string, loggingLevel string) (*zap.Logger, error) {
+func Config(opts ...ZapConfigOption) (*zap.Logger, error) {
+	const (
+		defaultEnvironment      = "development"
+		defaultLoggingLevel     = "info"
+		defaultStacktraceLength = 6
+	)
```
duncan-truss, line 50: Why are these necessary, they don't seem to differ from the defaults at https://github.com/transcom/mymove/pull/5442/files#diff-51bef48edb06413d6896391f5f1cbdbb1ac7d668a387498fb151306e05b84f04R80

monfresh: They're necessary if you don't pass in any particular option. For example, if you call `logging.Config` with only the environment and logging level, like this:

```golang
logger, err = logging.Config(
    logging.WithEnvironment(v.GetString(cli.LoggingEnvFlag)),
    logging.WithLoggingLevel(v.GetString(cli.LoggingLevelFlag)),
)
```
Then, if we don't set a default stack trace length, it will be 0, which will show the full stack trace. I could go either way on this. One option would be to imply that if you want to make sure there is a stack trace limit, then pass in the stack trace length env var, otherwise, leave it off, and you'll get the full stack trace.

Alternatively, we could say, if you leave it off, there's no harm because it will default to 6 to match the env var default value, so that we comply with STIG rules all the time. If you want to make sure it's configurable to a different value, then pass in the option.

I think I prefer the latter for safety.

monfresh: After I typed that and though about it some more, I think I changed my mind. The whole point of the functional options is that you can leave some out if you don't need them. For example, if some files are only ever used in development, then they don't need to pass in the stack trace length, and the assumption would be that it should default to the full stack trace. Otherwise, the only way to get a bigger stack trace, is to pass in the option so that the env var can be used. OK, I will remove those defaults.

pkg/logging/log.go
duncan-truss: I guess you could create a params struct if you really want keyed args.

pkg/logging/log.go
duncan-truss: Okay before you were forced with the constructor though to pass all of the args right?  One thing I think I learned on Saber was that infra prefers to fail fast explicitly but maybe that's not true across the board.

I agree it's a good idea to enforce the defaults for the STIG, it's tempting to pass the viper var where those defaults already exist but it's probably not good to introduce that dependency to the log package.

duncan-truss:
Adding the colors was a nice touch 🌈 I can confirm the server stacktrace was filtered to the number set in the flag and that it caught invalid values.  In development (and elsewhere) do we want the ability to not filter by the milmove package as a separate flag?  That feels baked into the lines limit.

pkg/logging/log.go
monfresh: Yeah, and even if we wanted to import the `cli` package, we couldn't because that would cause a circular dependency because `cli_test.go` imports the `logging` package. As for failing fast, that will still happen if the env var is not set to the right value, but if we forget to pass in the stack trace option, then it would default to showing the full stack trace. One way to solve that, while keeping the flexibility of passing in only the needed options, is to throw an error if the environment is *not* development, and the option has not been passed in. Or, we could just pass it in everywhere. What I wanted to avoid is that, if in the future we add some more logging options that don't apply everywhere, we would have to make changes to more files than necessary.

pkg/logging/log.go
monfresh: Wouldn't that require creating the struct every time I want to pass in options to `logging.Config`? I don't think that's great. Dave Cheney wrote about the disadvantages, and he prefers the functional options pattern:
https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

monfresh:
>  In development (and elsewhere) do we want the ability to not filter by the milmove package as a separate flag? That feels baked into the lines limit.

Yeah, I thought about that. I don't have strong feelings. Perhaps we could ask folks if they recall needing more than `mymove` stuff to troubleshoot a stack trace.

pkg/cli/logging.go
gidjin: I'm not really clear on why it has to be in the range 1-50. If the intention is just to validate that it is an int and not a string seems like there are other ways to check the type is a parsable int and then make sure it's just greater than 0. I don't have an example handy but I want to say I've run into some stack traces that are over 50 lines long, but I don't think that was in anything we run on the cli so idk if it's really an issue.

monfresh: That's a fair point. How about we just have a minimum set? Anything less than 6 I don't think would be helpful.

monfresh: Also, note that we are only keeping code paths from mymove, which shortens the stack trace.

pkg/logging/log.go
duncan-truss: His issue with the params struct was zero values but we could use pointers or other null types if we wanted to overcome that ambiguity.

Ryan-Koch (approved):

PR #5446, opened 2020-12-09
[MB-5532] Replace Qualifying BadDataError with ConflictError in the GHC EDI Generator

pearl-truss:
## Description

This PR was created to replace qualifying use of BadDataError with ConflictError in the GHC EDI Generator
https://github.com/transcom/mymove/wiki/API-Errors#http-status-codes

## References

According to the documentation:

> This error (ConflictError) is generated when we cannot process the request due to the current state of the server. For e.g. if the PrimeEstimatedWeight is not set on a shipment, we cannot create certain service items on that shipment. If we get a request to create such a service item, we have a conflict error.


reggieriser (commented):
The changes to this file LGTM.  But based on the [API Errors page](https://github.com/transcom/mymove/wiki/API-Errors#http-status-codes) you linked to, I'm wondering if we need to change callers to look for this type of error and return a 409 status code?  Is that what we're expecting to happen?

Specifically, I'm looking at [this handler](https://github.com/transcom/mymove/blob/b4c0007a8450a054457a072aae5f68f5b7529da9/pkg/handlers/supportapi/payment_request.go#L193-L225).

pkg/handlers/supportapi/payment_request.go
```diff
@@ -208,6 +208,11 @@ func (h GetPaymentRequestEDIHandler) Handle(params paymentrequestop.GetPaymentRe
 			return paymentrequestop.NewGetPaymentRequestEDIUnprocessableEntity().
 				WithPayload(payloads.ValidationError(handlers.ValidationErrMessage, h.GetTraceID(), e.ValidationErrors))
 
+			// ConflictError -> Conflict Error reponse
```
reggieriser, line 211: Super nit: Comment misaligned.

pkg/handlers/supportapi/payment_request.go
```diff
@@ -208,6 +208,11 @@ func (h GetPaymentRequestEDIHandler) Handle(params paymentrequestop.GetPaymentRe
...
+		case services.ConflictError:
+			return paymentrequestop.NewGetPaymentRequestEDIConflict().
```
reggieriser, line 213: Should/can we add a test for this error path?  Looks like we have tests around some of the other error paths.

reggieriser (approved):
Wondering about a test for this new error path, but not going to block on that.  Otherwise, LGTM!

PR #5447, opened 2020-12-09
[MB-5461] Upgrade react scripts & remove rescripts

suzubara:
## Description

This PR updates `react-scripts` to v4 (4.0.1) and makes necessary changes. Since eslint, jest, and prettier are all packaged within react-scripts, most of the changes are related to those.

I added in-line comments to the PR to explain most changes but here is a summary:
- Disable new default prettier eslint rules that caused a lot of new failures. These rules are just around code style/formatting, so shouldn't be a concern and we can always turn them back on when we decide to.
- Updated related packages (anything related to JS unit tests, eslint, and prettier) to latest versions
- Minor changes in code to satisfy new ESLint rules, such as ordering of imports and methods. One major change is no longer allowing anonymous default exports (so a default export has to be defined as a variable before being exported).

Additionally, I have removed rescripts packages and restored the CRA scripts in package.json. Rescripts was added in the early days of MM to provide a workaround for customizing Create React App config. There are a lot of similar tools around now, but CRA itself has also evolved, and the current recommended way of customizing your config is to maintain a fork of react-scripts. Truss actually has a [fork](https://github.com/trussworks/create-react-app) that I started back on a previous project, but haven’t kept up with at all -- but I think that is the best way forward as far as customizing CRA projects at Truss.

The custom config that rescripts was providing was:
- **Custom ESLint config** -- CRA now supports this out of the box, and I verified that our custom rules are still working as expected
- **Terser Webpack plugin** -- CRA also includes this with almost identical configuration to what we had. I think we can safely leave this out, and if it becomes an issue again address it with our custom fork of react-scripts. (We had originally configured this because we were trying to resolve webpack build memory failures in CI)
- **Case Sensitive Paths Webpack plugin** - This is not included in this PR. My inclination is to remove it for the time being (especially since at this time, I don't think we are doing as much moving around of files) and re-add it when we are using the custom react-scripts fork. I would like input on this decision since it is the one thing that I am intentionally _not_ migrating, but my opinion is that moving off of rescripts is of higher urgency.

## Reviewer Notes

Most of the changes here will affect the build, test, and lint pipeline, so if CI is passing that's a pretty good sign. It's worth checking out on your local machine to make sure your typical dev flow still works.

## Setup

- Pull down this branch and `yarn install` to update packages
- If you were running any client processes (`make client_run, yarn test, make storybook`) you will want to restart these and make sure they work as expected
- If any weirdness occurs, you might need to try `rm -rf node_modules` and installing all JS modules from scratch.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5461) for this change


.eslintrc
```diff
@@ -4,6 +4,7 @@
     "react-app",
     "airbnb",
     "prettier",
+    "prettier/prettier",
```
suzubara, line 7: I added this because the CRA update includes an update to `prettier` which turns on the `arrow-body-style` rule and this was causing a lot of new eslint failures. We can fix them in the future, but for now I followed the [instructions from `eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier/blob/main/README.md#arrow-body-style-and-prefer-arrow-callback) to disable the new rule for now.

.gitignore
```diff
@@ -41,6 +41,7 @@ cypress/cypress
 .vscode
 *.swp
 *.sublime-*
+.eslintcache
```
suzubara, line 44: this is a new file generated by running CRA that we can ignore

package.json
```diff
@@ -90,12 +90,12 @@
...
-    "start": "rescripts start",
-    "test": "rescripts test --env=jsdom",
-    "test:debug": "rescripts --inspect-brk test --runInBand --env=jsdom",
-    "test:coverage": "rescripts test --coverage --env=jsdom --watchAll=false --coverageDirectory=coverage",
+    "build": "react-scripts build",
+    "eject": "react-scripts eject",
+    "start": "react-scripts start",
+    "test": "react-scripts test --env=jsdom",
+    "test:debug": "react-scripts --inspect-brk test --runInBand --env=jsdom",
+    "test:coverage": "react-scripts test --coverage --env=jsdom --watchAll=false --coverageDirectory=coverage",
```
suzubara, line 98: down with `rescripts`, long live `react-scripts`

package.json
```diff
@@ -161,6 +158,7 @@
     "not op_mini all"
   ],
   "jest": {
+    "resetMocks": false,
```
suzubara, line 161: this flag newly defaults to `true` in CRA 4 and is breaking `jest.fn()`, it might be reverted. Source: https://github.com/facebook/create-react-app/issues/9935
https://github.com/facebook/create-react-app/issues/10171

src/shared/styles/_overrides.scss
```diff
@@ -248,8 +248,8 @@
     background: white;
     @include u-font-size('body', 'sm');
   }
+
   &:hover {
-    background-image: url('/static/media/arrow-both.0bfe2193.svg'), linear-gradient(transparent, transparent);
```
suzubara, line 252: @Shkeating I had to remove this because it was breaking the Storybook build. I don't think it was necessary in the first place because it seems to be a duplicate style declaration from USWDS, and there is no resulting visual diff.

src/styles/colors.module.scss
```diff
@@ -0,0 +1,49 @@
+@import 'shared/styles/colors.scss';
```
suzubara, line 1: I had to move the `:export` block into a `module.scss` file because of this bug:
https://github.com/facebook/create-react-app/issues/10155
https://github.com/facebook/create-react-app/issues/10047

suzubara:
My proposal for follow-up work:
- create a story to switch MM over to use the [Truss fork of react-scripts](https://github.com/trussworks/create-react-app)(https://github.com/trussworks/create-react-app)
- create issues in that repo for updating it, setting up dependabot so it stays up to date, and adding customization support for the webpack plugins we want (case sensitive paths, lodash, SASS resources, etc.)

gidjin (approved):
Changes don't seem to be too drastic. I ran `make clean` to delete everything and then I started the server and client and created a move and approved it. That all worked. I also ran the js unit tests and everything passed. Storybook appeared to function too. So looks good to me. 

As for the case sensitive plugin, I like it but not enough to say we should delay this upgrade for the minor guard rail of case sensitive file names. So I say :shipit:  but feel free to wait for more people, especially those actively in the front end.

Shkeating (approved):

PR #5432, opened 2020-12-08
[MB-5797] Add migration to create zip3_distances table

reggieriser:
## Description

This PR adds a migration to create a `zip3_distances` table that will be storing zip3-to-zip3 distances (data coming in a subsequent subtask/migration).  This data will eventually be used for pricing moves.

See the [design doc](https://docs.google.com/document/d/1RWh1hGYZltWKYpiejauoBambRL_REcZJVfBnLN3P0_M/edit#) for this work.

## Reviewer Notes

- We only store unique zip3-to-zip3 pairs in one direction (since presumably the opposite direction should have the same distance and our input file only gives us one direction).  The `zip3_distances_ordering` constraint and `zip3_distances_unique_zip3s` unique index enforce that.
- I put a default of `NOW()` on the created/updated timestamps.  We don't do that on many other tables since Pop sends it anyway, but Pop isn't involved in the data load.  Since the data load will be a large file, using the defaults helps to reduce the file size.
- We haven't historically documented constraints/indexes, but these two seemed like good candidates to have further explanation.

## Setup

- `make db_dev_reset db_dev_migrate` to make sure the migration runs without errors and you see the schema changed appropriately.
- `make schemaspy`, then open the `index.html` file in the generated web site (link at end of `make` output).  You should see the comments appear for that table.  Alternatively, use GoLand or some other database tool that shows Postgres comments.

## Code Review Verification Steps

* Any new migrations/schema changes:
  * [ ] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://github.com/transcom/mymove/wiki/migrate-the-database#zero-downtime-migrations))
  * [ ] Have been communicated to #g-database
  * [ ] Secure migrations have been tested following the instructions in our [docs](https://github.com/transcom/mymove/wiki/migrate-the-database#secure-migrations)
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5797) for this change
* [Design Doc](https://docs.google.com/document/d/1RWh1hGYZltWKYpiejauoBambRL_REcZJVfBnLN3P0_M/edit#)

PR #5461, opened 2020-12-10
[MB-5798] Migration to load zip3_distances data

reggieriser:
## Description

This PR adds a secure (and local obfuscated) migration for loading zip3-to-zip3 distance data into the `zip3_distances` table which will be used in pricing.  PR #5432 created the table and related objects.

See the [design doc](https://docs.google.com/document/d/1RWh1hGYZltWKYpiejauoBambRL_REcZJVfBnLN3P0_M/edit#) for this work to understand more about the approach taken.

## Reviewer Notes

- This is a large set of records; is there any way to speed up this load?  It currently takes about 5 seconds locally but YMMV.  I expect that to be a good bit longer when I test this on experimental after initial PR review.  **UPDATE:** I got it down to about 2.5 seconds locally by dropping/restoring indexes.

## Setup

- `db_dev_migrate` to make sure local migration works fine.
- `make run_prd_migrations` (or replace `prd` with `stg` or `exp` for those environments) to make sure secure migration works fine.
- Compare the results in `zip3_distances` and verify that the real migration reflects the data from the CSV and the obfuscated migration is randomized (see design doc for a link to the data file and details about randomization).

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5798) for this change
* [Design Doc](https://docs.google.com/document/d/1RWh1hGYZltWKYpiejauoBambRL_REcZJVfBnLN3P0_M/edit#)


reggieriser:
I did some testing and was able to speed up the data load by 50% or so (at least locally) by dropping the indexes and restoring them after the load was done.  Hopefully that helps in the deployed environments too.

lynzt (approved):
This looks good to me... The design doc you linked answered a lot of my questions.

Question about maintenance: Is this a table we'll have to keep "up to date"? Just something to think about when this might change?

reggieriser:
> Question about maintenance: Is this a table we'll have to keep "up to date"? Just something to think about when this might change?

I don't anticipate that this data set will change very often, but we could have some updates if new zip3s appear or new roads result in a different mileage.  That said, I believe the ultimate goal is to use a web service instead of this static data, so hopefully it's not a problem we'll have to worry about eventually.

> As for the time it takes, since we don't have active users using the system, I'm kinda like meh... but if this is something that we need to do in the future, we might want to look at the insert speed/locking.

I spent some time this morning researching/testing ways to speed it up and the big one is to use `COPY` like we're already doing.  But dropping/recreating the indexes was another one and that helped quite a bit (about 50% locally).  I'm interested to see how long this takes in experimental when I test it (waiting to do that last after getting a couple of reviews).

I think our biggest speed-up overall for migrations is going to be to 1) figure out a way to squash them since we build up a lot of old things that we turn around and tear down, and 2) get rid of some past data loads of performance data if we're allowed to.  If we squash migrations, I'd like to try the drop/restore indexes trick I did here to see how much that helps overall.

Thanks for reviewing!

gidjin (approved):
Assuming 411,778 records is the correct number after the discussed data clean up, this data looks good, as does the secure migration. 🚢

Ryan-Koch (approved):
This approach seems reasonable to me. Any questions I had pop up in my brain wound up being answered by that design doc you linked on slack.

PR #5477, opened 2020-12-11
MB-5922 Remove unused model type for sit departure

abbyoung:
## Description

* Removes `MTOSeriveItemSITDepartur model type from prime.yaml

Cleaning up unused model type `MTOServiceItemSITDeparture` in `prime.yaml`.

The sitDepartureDate implementation changed and resulted in an `UpdateMTOServiceItemSIT` model type and endpoint, so we no longer need this previous model type.

Once we remove this unused model type, it will resolve a high volume of failures for the `createMTOServiceItem` endpoint in load testing.

## Setup

You can test the updated yaml with load testing by taking the url to the branch's `prime.yaml` and substituting it in the `parsers.py` file, like so:

```python
class PrimeAPIParser(APIParser):
    """ Parser class for the Prime API. """

    api_file = "https://raw.githubusercontent.com/transcom/mymove/mb-5922-remove-unused-model-type-sit/swagger/prime.yaml"
```

Then run `make load_test_prime`. The should be no 400 errors similar to:

```
{
    "title": "Bad Request",
    "instance": "2ebef0f0-57d3-404c-af0e-3a86121ffcec",
    "detail": "parsing body body from \"\" failed, because invalid modelType value: \"MTOServiceItemSITDeparture\""
}
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5922) for this change

### DIFF
--- a/pkg/gen/primeapi/embedded_spec.go
+++ b/pkg/gen/primeapi/embedded_spec.go
@@ -1402,8 +1402,7 @@ func init() {
         "MTOServiceItemOriginSIT",
         "MTOServiceItemDestSIT",
         "MTOServiceItemShuttle",
-        "MTOServiceItemDomesticCrating",
-        "MTOServiceItemSITDeparture"
+        "MTOServiceItemDomesticCrating"
       ]
     },
     "MTOServiceItemOriginSIT": {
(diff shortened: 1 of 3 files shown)


shimonacarvalho (approved):
LGTM. I think that was added in prep for updating the departure but then we went another way and addd UpdateMTOServiceItemModelType

PR #5487, opened 2020-12-14
Output yarn logs on install of Cypress

suzubara:
## Description

Changes the `yarn install` command when setting up Cypress to run integration tests in Docker so it doesn't suppress logs. Making this change so we can find out more about why `yarn install` is occasionally failing in CI on all branches including master.


pearl-truss (approved):
LGTM!

PR #5488, opened 2020-12-14
Roc mb-5846 edit authorized weight

Ronolibert:
## Description

Changes update move order handler to be able to edit the authorized weight

## Reviewer Notes
- Had to convert the update move order service to be a transaction because it seems like `ValidateAndUpdate` only updates associations if you are assigning new associations not changing columns of an association

## Setup
`make server_run`
`make server_test`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5846) for this change

### DIFF
--- a/pkg/handlers/ghcapi/move_orders.go
+++ b/pkg/handlers/ghcapi/move_orders.go
@@ -4,6 +4,8 @@ import (
 	"database/sql"
 	"time"
 
+	"github.com/go-openapi/swag"
+
 	"github.com/transcom/mymove/pkg/gen/ghcmessages"
 	"github.com/transcom/mymove/pkg/gen/internalmessages"
 	"github.com/transcom/mymove/pkg/models"
@@ -159,8 +161,14 @@ func MoveOrder(payload ghcmessages.UpdateMoveOrderPayload) (models.Order, error)
 	}
 
 	departmentIndicator := string(payload.DepartmentIndicator)
+	var entitlement models.Entitlement
+
+	if payload.AuthorizedWeight != nil {
+		entitlement.DBAuthorizedWeight = swag.Int(int(*payload.AuthorizedWeight))
+	}
 
 	return models.Order{
+		Entitlement:         &entitlement,
 		IssueDate:           time.Time(*payload.IssueDate),
 		ReportByDate:        time.Time(*payload.ReportByDate),
 		OrdersType:          internalmessages.OrdersType(payload.OrdersType),
--- a/pkg/handlers/ghcapi/move_orders.go
+++ b/pkg/handlers/ghcapi/move_orders.go
@@ -161,8 +161,8 @@ func MoveOrder(payload ghcmessages.UpdateMoveOrderPayload) (models.Order, error)
 	}
 
 	departmentIndicator := string(payload.DepartmentIndicator)
-	var entitlement models.Entitlement
 
+	var entitlement models.Entitlement
 	if payload.AuthorizedWeight != nil {
 		entitlement.DBAuthorizedWeight = swag.Int(int(*payload.AuthorizedWeight))
 	}
--- a/pkg/services/move_order/move_order_updater.go
+++ b/pkg/services/move_order/move_order_updater.go
@@ -38,7 +38,7 @@ func (s *moveOrderUpdater) UpdateMoveOrder(moveOrderID uuid.UUID, eTag string, m
 
 	existingETag := etag.GenerateEtag(existingOrder.UpdatedAt)
 	if existingETag != eTag {
-		return nil, query.StaleIdentifierError{StaleIdentifier: eTag}
+		return nil, services.NewPreconditionFailedError(moveOrder.ID, query.StaleIdentifierError{StaleIdentifier: eTag})
 	}
 
 	transactionError := s.db.Transaction(func(tx *pop.Connection) error {
@@ -88,6 +88,8 @@ func (s *moveOrderUpdater) UpdateMoveOrder(moveOrderID uuid.UUID, eTag string, m
 			switch updateErr.(type) {
 			case query.StaleIdentifierError:
 				return services.NewPreconditionFailedError(moveOrder.ID, err)
+			default:
+				return updateErr
 			}
 		}
 
@@ -95,7 +97,7 @@ func (s *moveOrderUpdater) UpdateMoveOrder(moveOrderID uuid.UUID, eTag string, m
 	})
 
 	if transactionError != nil {
-		return nil, services.InvalidInputError{}
+		return nil, transactionError
 	}
 
 	return existingOrder, err
--- a/pkg/services/move_order/move_order_updater.go
+++ b/pkg/services/move_order/move_order_updater.go
@@ -6,6 +6,7 @@ import (
 	"github.com/gofrs/uuid"
 
 	"github.com/transcom/mymove/pkg/etag"
+
 	"github.com/transcom/mymove/pkg/models"
 	"github.com/transcom/mymove/pkg/services"
 	"github.com/transcom/mymove/pkg/services/query"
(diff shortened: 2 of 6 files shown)


pkg/services/move_order/move_order_updater.go
```diff
@@ -34,45 +35,71 @@ func (s *moveOrderUpdater) UpdateMoveOrder(moveOrderID uuid.UUID, eTag string, m
...
+			if fetchErr != nil {
+				return services.NewInvalidInputError(moveOrder.ID, fetchErr, nil, "unable to find origin duty station")
+			}
+			existingOrder.OriginDutyStationID = moveOrder.OriginDutyStationID
+			existingOrder.OriginDutyStation = &originDutyStation
+		}
 
-	verrs, err := s.builder.UpdateOne(existingOrder, &eTag)
+		if moveOrder.NewDutyStationID != existingOrder.NewDutyStationID {
+			newDutyStation, fetchErr := models.FetchDutyStation(s.db, *&moveOrder.NewDutyStationID)
```
duncan-truss, line 64: You can fix remove the unnecessary `*&` that I definitely wasn't the author of 🙈

pkg/services/move_order/move_order_updater.go
```diff
@@ -34,45 +35,71 @@ func (s *moveOrderUpdater) UpdateMoveOrder(moveOrderID uuid.UUID, eTag string, m
...
 		}
-		existingOrder.OriginDutyStationID = moveOrder.OriginDutyStationID
-		existingOrder.OriginDutyStation = &originDutyStation
-	}
 
-	if moveOrder.NewDutyStationID != existingOrder.NewDutyStationID {
-		newDutyStation, fetchErr := models.FetchDutyStation(s.db, *&moveOrder.NewDutyStationID)
-		if fetchErr != nil {
-			return nil, services.NewInvalidInputError(moveOrder.ID, fetchErr, nil, "unable to find destination duty station")
+		if moveOrder.Entitlement.DBAuthorizedWeight != nil &&
```
duncan-truss, line 44: Do we need to check that all values are necessarily different (we'd have to also add dependents authorized eventually too). It should be okay and overwrite and we'll still touch the `updated_at` value because they clicked save?

pkg/services/move_order/move_order_updater.go
```diff
@@ -34,45 +35,71 @@ func (s *moveOrderUpdater) UpdateMoveOrder(moveOrderID uuid.UUID, eTag string, m
 		return nil, services.NewNotFoundError(moveOrder.ID, "while looking for moveOrder")
 	}
 
-	if moveOrder.OriginDutyStationID != existingOrder.OriginDutyStationID {
-		originDutyStation, fetchErr := models.FetchDutyStation(s.db, *moveOrder.OriginDutyStationID)
-		if fetchErr != nil {
-			return nil, services.NewInvalidInputError(moveOrder.ID, fetchErr, nil, "unable to find origin duty station")
+	transactionError := s.db.Transaction(func(tx *pop.Connection) error {
+		existingETag := etag.GenerateEtag(existingOrder.UpdatedAt)
```
duncan-truss, line 39: We could check this even prior to the transaction right?  I always have a fear of bloated transactions.

pkg/services/move_order/move_order_updater.go
```diff
@@ -34,45 +35,71 @@ func (s *moveOrderUpdater) UpdateMoveOrder(moveOrderID uuid.UUID, eTag string, m
...
+				return services.NewPreconditionFailedError(moveOrder.ID, err)
+			}
 		}
+
+		return nil
+	})
+
+	if transactionError != nil {
+		if t, ok := transactionError.(query.StaleIdentifierError); ok {
+			return nil, services.NewPreconditionFailedError(existingOrder.ID, t)
```
duncan-truss, line 99: Aren't we already returning a PreconditionFailedError type from the transaction here if there is a stale identifier error https://github.com/transcom/mymove/pull/5488/files#diff-181ed3d7d35ad776c2dff959e256260422c8c43324723a4965dce75745994d40R89

pkg/services/move_order/move_order_updater.go
```diff
@@ -34,45 +35,71 @@ func (s *moveOrderUpdater) UpdateMoveOrder(moveOrderID uuid.UUID, eTag string, m
...
+		}
+		return nil, services.NewQueryError("moveOrder", transactionError, "")
```
duncan-truss, line 101: Simliar question why are we wrapping InvalidInput errors as QueryErrors?

swagger/ghc.yaml
```diff
@@ -1695,6 +1695,11 @@ definitions:
         x-nullable: true
       departmentIndicator:
         $ref: '#/definitions/DeptIndicator'
+      authorizedWeight:
```
duncan-truss, line 1698: What do you think about adding a `minimum: 1` to match the client side form validation?  And maybe specify the units as lbs in a description?

pkg/services/move_order/move_order_updater.go
Ronolibert: Oh you're right. I guess we don't need to check that its different necessarily since if it is nil then it'd just default to the original calculations

pkg/services/move_order/move_order_updater.go
Ronolibert: We are but isn't that error just assigned to the transactionError but the service itself hasn't returned anything yet

Ronolibert: Actually, I'll remove this since I'm moving the optimistic logic outside of the transaction

pkg/services/move_order/move_order_updater.go
Ronolibert: You're right! Changed this to invalid input error, was a hold over from a copy paste

pkg/services/move_order/move_order_updater.go
duncan-truss: Right we still need to know if the transaction value returned an error but it should already be of the type PreconditionFailedError if you checked for that inside the transaction.  There is the unlikely case that even if the etags match another update could sneak in between the time we check and attempt to save.

pkg/handlers/ghcapi/move_orders.go
```diff
@@ -159,8 +161,14 @@ func MoveOrder(payload ghcmessages.UpdateMoveOrderPayload) (models.Order, error)
 	}
 
 	departmentIndicator := string(payload.DepartmentIndicator)
+	var entitlement models.Entitlement
```
duncan-truss, line 164: I think Entitlement has it's own model to payload function and it looks like one for MoveOrder already exists in the `internal/payloads/model_to_payload.go` file

pkg/services/move_order/move_order_updater.go
```diff
@@ -34,45 +36,67 @@ func (s *moveOrderUpdater) UpdateMoveOrder(moveOrderID uuid.UUID, eTag string, m
...
+			case query.StaleIdentifierError:
+				return services.NewPreconditionFailedError(moveOrder.ID, err)
+			}
 		}
+
+		return nil
+	})
+
+	if transactionError != nil {
+		return nil, services.InvalidInputError{}
```
duncan-truss, line 98: I think we can just return the transactionError directly because they could be of a variety of types (Precondition, InvalidInput, and other unexpected sql errors.

pkg/services/move_order/move_order_updater.go
```diff
@@ -34,45 +36,67 @@ func (s *moveOrderUpdater) UpdateMoveOrder(moveOrderID uuid.UUID, eTag string, m
...
+		verrs, updateErr := s.builder.UpdateOne(existingOrder, &eTag)
+
+		if verrs != nil && verrs.HasAny() {
+			return services.NewInvalidInputError(moveOrder.ID, err, verrs, "")
+		}
+
+		if updateErr != nil {
+			switch updateErr.(type) {
+			case query.StaleIdentifierError:
+				return services.NewPreconditionFailedError(moveOrder.ID, err)
```
duncan-truss, line 90: we need a default case here to still return the updateErr directly if it doesn't match the other cases.

pkg/handlers/ghcapi/move_orders.go
Ronolibert: I'm not sure I understand, isn't this a `payload_to_model` situation

Ronolibert:
@duncan-truss I fixed all the error handling stuff and had one last question on the entitlement payload comment

pkg/handlers/ghcapi/move_orders.go
duncan-truss: You're right disregard what I said I was thrown off that it was in the handler file and thought we were duplicating what existed in model_to_payload.  Thanks for setting me straight.

duncan-truss (approved):
LGTM!

PR #5490, opened 2020-12-14
MB-4393 | Removing AWS_PROFILE from .envrc

esacteksab:
## Description

Removes the last remaining reference to AWS commercial. No impact expected as previous efforts worked around this.

## Reviewer Notes

_All_ the things! 

* [MB-4393](https://dp3.atlassian.net/browse/MB-4393) for this change

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -179,9 +179,9 @@ export TZ="UTC"
 if [ "$STORAGE_BACKEND" == "s3" ]; then
   export AWS_S3_BUCKET_NAME="transcom-gov-dev-app-devlocal-us-gov-west-1"
   export AWS_PROFILE=transcom-gov-dev
+  export AWS_S3_REGION="us-gov-west-1"
 fi
 
-export AWS_S3_REGION="us-gov-west-1"
 export AWS_DEFAULT_REGION="us-gov-west-1"
 export AWS_S3_KEY_NAMESPACE=$USER
 export AWS_SES_DOMAIN="devlocal.dp3.us"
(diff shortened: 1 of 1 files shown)


travelar (approved):
🚨

kilbergr (commented):
Ok I'm being a little horrible in this PR bc it's total insider knowledge, but with this happening, I think you should move line 184/182 (export AWS_DEFAULT_REGION="us-gov-west-1") OUT of the conditional since we will henceforth be using exclusively govcloud, right? And all the scripts do expect the default region, I'm finding. What do you think?

kilbergr (changes requested):
In fact, I'm going to block on my comment as I think it is necessary. Happy to walk thru if it doesn't make sense to you/you have objections.

esacteksab:
> In fact, I'm going to block on my comment as I think it is necessary. Happy to walk thru if it doesn't make sense to you/you have objections.

Let me know if I captured what you were hoping to achieve.

.envrc
```diff
@@ -180,11 +178,11 @@ export TZ="UTC"
 # Setting region and profile conditionally while we migrate from com to govcloud.
 if [ "$STORAGE_BACKEND" == "s3" ]; then
   export AWS_S3_BUCKET_NAME="transcom-gov-dev-app-devlocal-us-gov-west-1"
-  export AWS_S3_REGION="us-gov-west-1"
```
kilbergr, line 183: Sorry--I think we should leave the S3 region in the conditional, but move the default region out.

esacteksab: I'm curious why.

kilbergr: Oh sure--because I believe S3 region is only being used when we set the STORAGE_BACKEND to S3, whereas the AWS_DEFAULT_REGION is being used by a bunch of other scripts unrelated to the STORAGE_BACKEND. Make sense?

esacteksab: It does. Thank you.

kilbergr (approved):
🥇

PR #5491, opened 2020-12-15
MB-5952: add a command to run the webhook client in docker

carterjones:
## Description

This adds two make commands to Makefile:

1. `make webhook_client_start`: this brings up the dev database, runs migrations, and prepares it to a state that the webhook-client container can use
1. `make webhook_client_start_standalone`: this runs a pre-existing webhook-client container, which can be generated using the pre-existing `make webhook_client_docker` command

## Reviewer Notes

I tried to follow the pattern regarding the mTLS certificate & key for local environments. I'm not exactly sure how to go about handling certificate & key in the non-local Docker image, since baking certificate keys into images isn't a best practice. So if there is a better pattern that should be used for the mTLS keys in this image, I am happy to follow it. I'm not concerned about secrecy for these specific certificates given that they are already public; I'm more concerned with the pattern of baking certificate keys into Docker images; it just feels like there is perhaps a better way. I brainstormed an approach where we would pass in the repo by using Docker volumes, but then that has the potential to get complex for users who are trying to use the container from outside of the repository. If anyone has opinions on how/if I should do things differently with all this, please let me know.

If there are any other commonly used environment variables (command line options converted into environment variables) that should be configured in the docker container, please let me know. There are a whole bunch of options in `bin/webhook-client`, but I'm not sure which ones should be included here when spinning up the Docker container locally.

## Setup

This demonstrates the default configuration:

```console
$ make webhook_client_start_standalone
Starting the webhook client...
docker run \
                -e LOGGING_LEVEL=debug \
                -e DB_HOST="database" \
                -e DB_NAME \
                -e DB_PORT \
                -e DB_USER \
                -e DB_PASSWORD \
                -e PERIOD \
                --link="milmove-db-dev:database" \
                webhook-client:latest
2020-12-15T02:14:14.129Z        INFO    webhook-client/main.go:50       Checking config and initializing
2020-12-15T02:14:14.129Z        INFO    cli/dbconn.go:258       Connecting to the database      {"url": "postgres://postgres:*****@database:5432/dev_db?sslmode=disable", "db-ssl-root-cert": ""}
2020-12-15T02:14:14.129Z        INFO    cli/dbconn.go:363       Starting database ping....
2020-12-15T02:14:14.132Z        INFO    cli/dbconn.go:370       ...DB ping successful!
2020-12-15T02:14:14.138Z        DEBUG   webhook/webhook.go:258  Notification Check:     {"Num notifications found": 0}
2020-12-15T02:14:19.136Z        DEBUG   webhook/webhook.go:258  Notification Check:     {"Num notifications found": 0}
^C2020/12/15 02:14:21 Shutdown Server ...
2020-12-15T02:14:21.000Z        INFO    webhook-client/db_webhook_notify.go:82  Db connection closed
2020/12/15 02:14:21 Listener exiting
```

This is a demonstration of environment variables in the laptop's shell being passed down into the Docker container. In this case, it is using the `PERIOD` environment variable and using that to determine how frequently to check the database for updates:

```text
$ PERIOD=1 make webhook_client_start_standalone
Starting the webhook client...
docker run \
                -e LOGGING_LEVEL=debug \
                -e DB_HOST="database" \
                -e DB_NAME \
                -e DB_PORT \
                -e DB_USER \
                -e DB_PASSWORD \
                -e PERIOD \
                --link="milmove-db-dev:database" \
                webhook-client:latest
2020-12-15T02:14:26.550Z        INFO    webhook-client/main.go:50       Checking config and initializing
2020-12-15T02:14:26.550Z        INFO    cli/dbconn.go:258       Connecting to the database      {"url": "postgres://postgres:*****@database:5432/dev_db?sslmode=disable", "db-ssl-root-cert": ""}
2020-12-15T02:14:26.551Z        INFO    cli/dbconn.go:363       Starting database ping....
2020-12-15T02:14:26.556Z        INFO    cli/dbconn.go:370       ...DB ping successful!
2020-12-15T02:14:26.564Z        DEBUG   webhook/webhook.go:258  Notification Check:     {"Num notifications found": 0}
2020-12-15T02:14:27.559Z        DEBUG   webhook/webhook.go:258  Notification Check:     {"Num notifications found": 0}
2020-12-15T02:14:28.559Z        DEBUG   webhook/webhook.go:258  Notification Check:     {"Num notifications found": 0}
2020-12-15T02:14:29.559Z        DEBUG   webhook/webhook.go:258  Notification Check:     {"Num notifications found": 0}
2020-12-15T02:14:30.559Z        DEBUG   webhook/webhook.go:258  Notification Check:     {"Num notifications found": 0}
^C2020/12/15 02:14:30 Shutdown Server ...
2020-12-15T02:14:30.977Z        INFO    webhook-client/db_webhook_notify.go:82  Db connection closed
2020/12/15 02:14:30 Listener exiting
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5952) for this change

### DIFF
--- a/Dockerfile.webhook_client
+++ b/Dockerfile.webhook_client
@@ -3,6 +3,4 @@ FROM gcr.io/distroless/static:latest
 COPY bin/rds-ca-us-gov-west-1-2017-root.pem /bin/rds-ca-us-gov-west-1-2017-root.pem
 COPY bin/webhook-client /bin/webhook-client
 
-ENTRYPOINT ["/bin/webhook-client"]
-
-CMD ["db-webhook-notify"]
+CMD ["/bin/webhook-client", "db-webhook-notify"]
--- a/Dockerfile.webhook_client_local
+++ b/Dockerfile.webhook_client_local
@@ -19,9 +19,14 @@ RUN make bin/webhook-client
 
 FROM gcr.io/distroless/static:latest
 
+# mTLS authentication self-signed certificate for use in development and test environments.
+COPY --from=builder --chown=root:root /home/circleci/project/config/tls/devlocal-mtls.cer /config/tls/devlocal-mtls.cer
+COPY --from=builder --chown=root:root /home/circleci/project/config/tls/devlocal-mtls.key /config/tls/devlocal-mtls.key
+
+# Public root certificate for RDS in us-gov-west-1.
 COPY --from=builder --chown=root:root /home/circleci/project/bin/rds-ca-us-gov-west-1-2017-root.pem /bin/rds-ca-us-gov-west-1-2017-root.pem
+
+# The main webhook-client binary.
 COPY --from=builder --chown=root:root /home/circleci/project/bin/webhook-client /bin/webhook-client
 
-ENTRYPOINT ["/bin/webhook-client"]
-
-CMD ["db-webhook-notify"]
+CMD ["/bin/webhook-client", "db-webhook-notify"]
--- a/Makefile
+++ b/Makefile
@@ -893,6 +893,25 @@ run_exp_migrations: bin/milmove db_deployed_migrations_reset ## Run GovCloud exp
 webhook_client_docker:
 	docker build -f Dockerfile.webhook_client_local -t $(WEBHOOK_CLIENT_DOCKER_CONTAINER):latest .
 
+.PHONY: webhook_client_start
+webhook_client_start: db_dev_e2e_populate webhook_client_start_standalone
+
+.PHONY: webhook_client_start_standalone
+webhook_client_start_standalone:
+	@echo "Starting the webhook client..."
+	# More environment variables can be added here that correlate with the command line options for
+	# the webhook-client binary.
+	docker run \
+		-e LOGGING_LEVEL=debug \
+		-e DB_HOST="database" \
+		-e DB_NAME \
+		-e DB_PORT \
+		-e DB_USER \
+		-e DB_PASSWORD \
+		-e PERIOD \
+		--link="$(DB_DOCKER_CONTAINER_DEV):database" \
+		$(WEBHOOK_CLIENT_DOCKER_CONTAINER):latest
+
 .PHONY: webhook_client_test
 webhook_client_test:
 	echo "This is a placeholder for webhook-client tests"


travelar (approved):
I don't know the best path forward regarding certs. It's possible that we could pull them from chamber, but is that any better or worse than this approach?

Dockerfile.webhook_client
```diff
@@ -3,6 +3,4 @@ FROM gcr.io/distroless/static:latest
 COPY bin/rds-ca-us-gov-west-1-2017-root.pem /bin/rds-ca-us-gov-west-1-2017-root.pem
 COPY bin/webhook-client /bin/webhook-client
 
-ENTRYPOINT ["/bin/webhook-client"]
-
-CMD ["db-webhook-notify"]
+CMD ["/bin/webhook-client", "db-webhook-notify"]
```
mdawn, line 6: Why do you no longer need this ENTRYPOINT?

Dockerfile.webhook_client_local
```diff
@@ -19,9 +19,14 @@ RUN make bin/webhook-client
 
 FROM gcr.io/distroless/static:latest
 
+# mTLS authentication self-signed certificate for use in development and test environments.
+COPY --from=builder --chown=root:root /home/circleci/project/config/tls/devlocal-mtls.cer /config/tls/devlocal-mtls.cer
```
mdawn, line 23: Did you have to manually add this to circleci? Or was it already there?

Makefile
```diff
@@ -893,6 +893,25 @@ run_exp_migrations: bin/milmove db_deployed_migrations_reset ## Run GovCloud exp
 webhook_client_docker:
 	docker build -f Dockerfile.webhook_client_local -t $(WEBHOOK_CLIENT_DOCKER_CONTAINER):latest .
 
+.PHONY: webhook_client_start
+webhook_client_start: db_dev_e2e_populate webhook_client_start_standalone
+
+.PHONY: webhook_client_start_standalone
+webhook_client_start_standalone:
```
mdawn, line 900: When would this need to be run standalone? Is this just best practice to include the standalone command? Is this mostly just for your own testing/debugging?

Dockerfile.webhook_client
carterjones: Practically speaking, I could have used `ENTRYPOINT` or `CMD` here. However, when experimenting/iterating locally with running different commands, I found that I preferred this syntax (enabled by the setting the command in `CMD`):

```sh
docker run --rm -it webhook-client:latest /bin/whatever
```

over this syntax, which is required when setting the command in `ENTRYPOINT`:

```sh
docker run --rm -it --entrypoint /bin/whatever webhook-client:latest
```

I preferred modifying the end of the command when iterating rather than having to bounce around on the command line. With the `ENTRYPOINT` approach, I needed to put the `--entrypoint /bin/whatever` before the `webhook-client:latest` portion, since the `docker run` usage is:


```text
Usage:  docker run [OPTIONS] IMAGE [COMMAND] [ARG...]
```

So this is purely a mechanism of saving time when moving the cursor around on the command line.

From a Docker best practices perspective, I see that the Internet says many things, none of which I am strongly opposed to. If people say that we should be using one or the other (or perhaps both in tandem) in some particular way, I would have no objections.

Dockerfile.webhook_client_local
carterjones: It was already there given these two statements:

```
COPY --chown=circleci:circleci . /home/circleci/project
WORKDIR /home/circleci/project
```

The certificates are stored at `config/tls/` in the mymove repo so when the root directory directory (`.`) was copied, it copied the certificates to `/home/circleci/project/config/tls/`.

Makefile
carterjones: You wouldn't need to run in standalone, but it helps speed up iteration. `make webhook_client_start` completely resets the database and re-runs all the migrations, which can take some time. In my experience, it's around a couple minutes. So if the database is already up and running, we can just run `make webhook_client_start_standalone` and it'll use the database as it currently exists. Additionally, if devs are making changes to the database, they'll likely not want to reset and rebuild the database, but instead would want to use/test whatever custom changes they've made to the database. Using the standalone command will let them do that without blowing away whatever changes they made to the database.

As for the word `standalone`, there's nothing fancy about it to my knowledge. I saw it referenced a few other places in Makefile, so I followed that pattern.

carterjones:
> I don't know the best path forward regarding certs. It's possible that we could pull them from chamber, but is that any better or worse than this approach?

@travelar I like that we at least have options. I'm not entirely sure if it's better or worse than this approach. And the trade-offs will be different for local vs ECS containers. I'll need to think about it a bit more, but having that option is certainly good.

travelar:
> > I don't know the best path forward regarding certs. It's possible that we could pull them from chamber, but is that any better or worse than this approach?
> 
> @travelar I like that we at least have options. I'm not entirely sure if it's better or worse than this approach. And the trade-offs will be different for local vs ECS containers. I'll need to think about it a bit more, but having that option is certainly good.

@carterjones I feel an ADR brewing here.

carterjones:
Agreed, but before we go down the ADR route, I think it would be worth us getting an understanding of how secrets are _currently_ managed in containers. I'm guessing a pattern exists, but I'm just not aware of the pattern(s).

PR #5510, opened 2020-12-18
[MB-5550] Use map instead of slice for EDI segments

mboren:
## Description

Previously, we used slices to hold segments in [ghc_payment_request_generator.go](https://github.com/transcom/mymove/blob/master/pkg/services/invoice/ghc_payment_request_invoice_generator.go).
  This created unit tests that had to reference slices at specific indices, which made writing and updating tests annoying.
 It also leaves us with errors from the validator that reference `edi.Header[9] Missing Name` which is hard to debug.

We originally planned to use a `map[string]Segment` for Invoice858C.Header, but this required us to separately keep track of the order to write segments out in, and to keep the string keys in sync across a few places. We tried an ordered map library, but it required us to write a custom validation function to make it work with our validation rules, and lost some type safety. Eventually, we settled on just using a separate struct for the header. This introduces new forms of awkwardness, but it does make the tests and error messages better.

I also created a struct for the repeated segments in ServiceItems.
 
 
## Reviewer Notes
- I had to make a lot of changes to tests.
- In particular, moving all the header items into a struct caused validation rules to be run, so the minimal invoice that was being used for `TestEdiString` was failing validation. I adapted some sample data from the sample EDI file attached to [MB-5934](https://dp3.atlassian.net/browse/MB-5934)
    - I noticed a discrepancy between this sample file and what our code produces: for N9 and N4 segments, we print out 2 asterisks at the end of the line (for 2 fields that we dont use), while the sample data omits those field separators for the empty fields.
        - our output: `N9*1W*Leo, Spacemen**`
        - example file: `N9*1W*Leo, Spacemen`
    - **I would love to know if my sample data is OK and whether or not the extra `**` at the end of lines matters.**
- `ediinvoice.NonEmptySegments` and `ediinvoice.Size` are both a bit awkward and inefficient. Putting all the header fields in a struct makes it harder to iterate over all the segments. More reflection could be used here (but at what cost? 😨 ). **Any suggestions for how to do this better?**

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5550)

### DIFF
--- a/pkg/edi/invoice/generator_test.go
+++ b/pkg/edi/invoice/generator_test.go
@@ -81,6 +81,9 @@ func MakeValidEdi() Invoice858C {
 		DateQualifier: 10,
 		Date:          "20200909",
 	}
+	ediHeader := make(map[string]edisegment.Segment)
+	ediHeader["G62_RequestedPickupDate"] = &date
+
 	n4 := edisegment.N4{
 		CityName:            "San Francisco",
 		StateOrProvinceCode: "CA",
@@ -126,9 +129,7 @@ func MakeValidEdi() Invoice858C {
 			TransactionSetIdentifierCode: "858",
 			TransactionSetControlNumber:  "ABCDE",
 		},
-		Header: []edisegment.Segment{
-			&date,
-		},
+		Header: ediHeader,
 		ServiceItems: []edisegment.Segment{
 			&l3total,
 			&n4,
--- a/pkg/edi/invoice/generator_test.go
+++ b/pkg/edi/invoice/generator_test.go
@@ -5,8 +5,6 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/emirpasic/gods/maps/linkedhashmap"
-
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/suite"
@@ -60,9 +58,26 @@ func (suite *InvoiceSuite) TestEDIString() {
 		suite.Equal(`ISA*00*0084182369*00*0000000000*ZZ*MILMOVE        *12*8004171844     *201002*1504*U*00401*000009999*0*T*|
 GS*SI*MILMOVE*8004171844*20190903*1617*1*X*004010
 ST*858*ABCDE
+BX*00*J*PP*3351-1123b*BLKW**4
+N9*CN*3351-1123-1b**
+N9*CT*TRUSS_TEST**
+N9*1W*Leo, Spacemen**
+N9*ML*E_1**
+N9*3L*ARMY**
 G62*10*20200909**
+N1*BY*BuyerOrganizationName*92*LKNQ
+N1*SE*SellerOrganizationName*2*BLKW
+N1*ST*DestinationName*10*CNNQ
+N4*Augusta*GA*30813*US**
+N1*SF*Uoe1WjuUjU*10*LKNQ
+N4*Des Moines*IA*50309*US**
+HL*1**I
+N9*PO*3351-1123c**
+L5*1*CS*TBD*D**
+L0*1**********
+FA1*DF
+FA2*TA*1234
 L3*300.000*B***100
-N4*San Francisco*CA*94123*USA**
 SE*12345*ABCDE
 GE*1*1234567
 IEA*1*000009999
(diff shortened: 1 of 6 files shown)


pkg/edi/invoice/generator_test.go
```diff
@@ -58,9 +58,26 @@ func (suite *InvoiceSuite) TestEDIString() {
 		suite.Equal(`ISA*00*0084182369*00*0000000000*ZZ*MILMOVE        *12*8004171844     *201002*1504*U*00401*000009999*0*T*|
 GS*SI*MILMOVE*8004171844*20190903*1617*1*X*004010
 ST*858*ABCDE
+BX*00*J*PP*3351-1123b*BLKW**4
+N9*CN*3351-1123-1b**
+N9*CT*TRUSS_TEST**
+N9*1W*Leo, Spacemen**
```
gidjin, line 64: I can confirm that the extra `*`'s at the end are optional when the data in those fields is omitted. However, the way our code generates the EDI ti would be a pain to take them out.

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -238,42 +231,38 @@ func (g ghcPaymentRequestInvoiceGenerator) Generate(paymentRequest models.Paymen
...
 		ReferenceIdentificationQualifier: "3L",
 		ReferenceIdentification:          string(*branch),
 	}
-	serviceMemberDetails = append(serviceMemberDetails, &serviceMemberBranch)
 
-	return serviceMemberDetails, nil
+	return nil
 }
 
-func (g ghcPaymentRequestInvoiceGenerator) createG62Segments(paymentRequestID uuid.UUID) ([]edisegment.Segment, error) {
```
gidjin, line 276: Since this method is modifying the header that is passed in, is there any reason to still return the slice of segments vs just an error?

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -323,80 +312,74 @@ func (g ghcPaymentRequestInvoiceGenerator) createG62Segments(paymentRequestID uu
...
 	var destinationDutyStation models.DutyStation
 	if orders.NewDutyStationID != uuid.Nil {
 		destinationDutyStation, err = models.FetchDutyStation(g.db, orders.NewDutyStationID)
 		if err != nil {
-			return []edisegment.Segment{}, services.NewInvalidInputError(orders.NewDutyStationID, err, nil, "unable to find new duty station")
+			return services.NewInvalidInputError(orders.NewDutyStationID, err, nil, "unable to find new duty station")
 		}
 	} else {
-		return []edisegment.Segment{}, services.NewConflictError(orders.ID, "Invalid Order, must have NewDutyStation")
+		return services.NewBadDataError("Invalid Order, must have NewDutyStation")
```
gidjin, line 368: Should remain ConflictError as above

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -238,42 +231,38 @@ func (g ghcPaymentRequestInvoiceGenerator) Generate(paymentRequest models.Paymen
...
 		ReferenceIdentificationQualifier: "1W",
 		ReferenceIdentification:          serviceMember.ReverseNameLineFormat(),
 	}
-	serviceMemberDetails = append(serviceMemberDetails, &serviceMemberName)
 
 	// rank
 	rank := serviceMember.Rank
 	if rank == nil {
-		return []edisegment.Segment{}, services.NewConflictError(serviceMember.ID, fmt.Sprintf("no rank found for ServiceMember ID: %s Payment Request ID: %s", serviceMember.ID, paymentRequestID))
+		return services.NewBadDataError(fmt.Sprintf("no rank found for ServiceMember ID: %s Payment Request ID: %s", serviceMember.ID, paymentRequestID))
```
gidjin, line 245: Should be ConflictError as above

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -238,42 +231,38 @@ func (g ghcPaymentRequestInvoiceGenerator) Generate(paymentRequest models.Paymen
...
 		ReferenceIdentificationQualifier: "ML",
 		ReferenceIdentification:          string(*rank),
 	}
-	serviceMemberDetails = append(serviceMemberDetails, &serviceMemberRank)
 
 	// branch
 	branch := serviceMember.Affiliation
 	if branch == nil {
-		return []edisegment.Segment{}, services.NewConflictError(serviceMember.ID, fmt.Sprintf("no branch found for ServiceMember ID: %s Payment Request ID: %s", serviceMember.ID, paymentRequestID))
+		return services.NewBadDataError(fmt.Sprintf("no branch found for ServiceMember ID: %s Payment Request ID: %s", serviceMember.ID, paymentRequestID))
```
gidjin, line 255: Should be ConflictError as above

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -323,80 +312,74 @@ func (g ghcPaymentRequestInvoiceGenerator) createG62Segments(paymentRequestID uu
...
 
 	if orders.OriginDutyStationID != nil && *orders.OriginDutyStationID != uuid.Nil {
 		originDutyStation, err = models.FetchDutyStation(g.db, *orders.OriginDutyStationID)
 		if err != nil {
-			return []edisegment.Segment{}, services.NewInvalidInputError(*orders.OriginDutyStationID, err, nil, "unable to find origin duty station")
+			return services.NewInvalidInputError(*orders.OriginDutyStationID, err, nil, "unable to find origin duty station")
 		}
 	} else {
-		return []edisegment.Segment{}, services.NewConflictError(orders.ID, "Invalid Order, must have OriginDutyStation")
+		return services.NewBadDataError("Invalid Order, must have OriginDutyStation")
```
gidjin, line 332: This should remain a ConflictError not a BadDataError. BadData implies the prime sent in data that was bad, but this is a case where data in the system won't allow us to continue. See this doc for details https://github.com/transcom/mymove/wiki/API-Errors#error-types-and-error-responses

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -449,24 +431,23 @@ func (g ghcPaymentRequestInvoiceGenerator) createOriginAndDestinationSegments(pa
...
```
gidjin, line 437: Should remain ConflictError as above

gidjin (changes requested):
Changes look good. I don't think it's worth adding more reflection at this point either, so you can leave the inefficiency of Size and NonEmptySegments. Thanks for improving the generator test too. I'm marking this as Request Changes because of the change in some ConflictError's to BadDataErrors.

pkg/services/invoice/ghc_payment_request_invoice_generator.go
mboren: Thanks for catching these, I was using macros for a lot of these repetitive changes and this slipped by me.

pkg/services/invoice/ghc_payment_request_invoice_generator.go
mboren: No reason at all, I'll fix that

gidjin (approved):
Thanks for making updates! LGTM

PR #5525, opened 2020-12-21
Switch cypress dockerfile to use milmove-cypress image

gidjin:
## Description

Currently every build of cypress tests in circle ci causes an install of cypress in building the `Dockerfile.cypress` image. However, sometimes that install fails causing the build to fail. Also we should only reinstall cypress in the docker image if we have requested a new version. This PR switches the Dockerfile to use the new `milmove-cypress` image from [the circleci-docker repo](https://github.com/transcom/circleci-docker/pull/63).

## Reviewer Notes

The `cypress/package.json` and `cypress/yarn.lock` files are included in the milmove-cypress image. In the future changes to the version of cypress needs to be done in the circleci-docker repo.

## Setup

This uses the new image from PR https://github.com/transcom/circleci-docker/pull/63

```sh
make e2e_test_docker
```

## References

* [this slack thread](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1608150863124600) explains more about the approach used.


duncan-truss:
@gidjin This looks great, although do we lose dependabot scanning for npm package updates now for the cypress dependencies?

dangerfile.ts
```diff
@@ -183,51 +183,25 @@ const cypressUpdateChecks = async () => {
 
   // check if relevant package.jsons have changed
   const rootPackageFile = 'package.json';
-  const cypressPackageFile = 'cypress/package.json';
   const rootPackageChanged = allFiles.includes(rootPackageFile);
-  const cypressPackageChanged = allFiles.includes('cypress/package.json');
   const cypressPackageName = '"cypress":';
```
suzubara, line 187: We may want this to be an array of all the dependencies listed in the [Cypress image dependencies](https://github.com/transcom/circleci-docker/pull/63/files#diff-1939631fae0ef26991d0dfb2d03c243b0aa82485947115e4c1c3db113ecbebddR1)

suzubara (commented):
awesome!! left one suggestion but I am really excited for this improvement

suzubara:
> @gidjin This looks great, although do we lose dependabot scanning for npm package updates now for the cypress dependencies?

@duncan-truss the Cypress deps are duplicated in the root package.json of this repo in order to run it locally, and it looks like dependabot is also set up in https://github.com/transcom/circleci-docker so I think we're good here!

duncan-truss:
> > @gidjin This looks great, although do we lose dependabot scanning for npm package updates now for the cypress dependencies?
> 
> @duncan-truss the Cypress deps are duplicated in the root package.json of this repo in order to run it locally, and it looks like dependabot is also set up in https://github.com/transcom/circleci-docker so I think we're good here!

That sounds good to me, I don't think dependabot is set to look at the milmove-cypress deps there https://github.com/transcom/circleci-docker/blob/master/.dependabot/config.yml but maybe here is enough?

gidjin:
> > > @gidjin This looks great, although do we lose dependabot scanning for npm package updates now for the cypress dependencies?
> > 
> > 
> > @duncan-truss the Cypress deps are duplicated in the root package.json of this repo in order to run it locally, and it looks like dependabot is also set up in https://github.com/transcom/circleci-docker so I think we're good here!
> 
> That sounds good to me, I don't think dependabot is set to look at the milmove-cypress deps there https://github.com/transcom/circleci-docker/blob/master/.dependabot/config.yml but maybe here is enough?

Thanks for catching that! I'll update the other pr!

dangerfile.ts
gidjin: That's probably a good idea, today's conflict was `cypress-audit` so I'll make the appropriate change

suzubara (approved):

duncan-truss (approved):

PR #5528, opened 2020-12-21
[MB-6010] create model for Rand McNally data

pearl-truss:
## Description

In order to integrate the Rand McNally database table (zip3_distances) with the GHC Planner, we need a model for the RM data. I used this [Rand McNally doc](https://docs.google.com/document/d/1RWh1hGYZltWKYpiejauoBambRL_REcZJVfBnLN3P0_M/edit#heading=h.xk6y5kr3aui9) to create the struct for the data model

## Set Up
Confirm that the server test in `./pkg/models` are passing. 

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6010) for this change
* [Rand McNally Data Doc](https://docs.google.com/document/d/1RWh1hGYZltWKYpiejauoBambRL_REcZJVfBnLN3P0_M/edit#) used as a basis for determining the schema

### DIFF
--- a/pkg/models/zip3_distance_test.go
+++ b/pkg/models/zip3_distance_test.go
@@ -1,0 +1,64 @@
+package models_test
+
+import (
+	"testing"
+
+	"github.com/transcom/mymove/pkg/models"
+)
+
+func (suite *ModelSuite) TestZip3DistanceValidations() {
+	suite.T().Run("test valid Zip3Distance", func(t *testing.T) {
+		validZip3Distance := models.Zip3Distance{
+			FromZip3:      "010",
+			ToZip3:        "011",
+			DistanceMiles: 24,
+		}
+		expErrors := map[string][]string{}
+		suite.verifyValidationErrors(&validZip3Distance, expErrors)
+	})
+
+	suite.T().Run("test invalid Zip3Distance", func(t *testing.T) {
+		emptyZip3Distance := models.Zip3Distance{}
+		expErrors := map[string][]string{
+			"from_zip3":      {"FromZip3 not in range(3, 3)"},
+			"to_zip3":        {"ToZip3 not in range(3, 3)"},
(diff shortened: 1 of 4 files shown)


gidjin (approved):
Code looks good 🚢

pkg/models/zip3_distance_test.go
```diff
@@ -0,0 +1,64 @@
...
+	suite.T().Run("test when distance_miles is not provided", func(t *testing.T) {
+		invalidDistance := models.Zip3Distance{
+			FromZip3: "010",
+			ToZip3:   "011",
+		}
+		expErrors := map[string][]string{
+			"distance_miles": {"DistanceMiles can not be blank."},
+		}
+		suite.verifyValidationErrors(&invalidDistance, expErrors)
+	})
```
gidjin, line 63: Is it worth / possible to add a test where FromZip3 was greater than ToZip3 violating the db table constraint?

pearl-truss: I've looked into this and I think the problem is that there isn't a way in the model to validate FromZip3 < ToZip3 since these fields are strings. 

@reggieriser do you have any thoughts on this?

reggieriser: Glancing through the built-in validators and our custom ones, I don't think there is an existing validator that would apply.  There is something similar for times -- `TimeIsBeforeTime` -- that you could use as a model for creating a `StringIsBeforeString` validator if you wanted.

That said, I'm not sure it's necessarily worth it since we don't (at the moment, at least) anticipate creating/updating these records through the model.  Plus, the constraint would catch the error as a safety if we ever did.

pearl-truss: Thanks! It sounds like more effort than it's worth to address and there's a constraint in place that can catch the error, so I'm going to merge the code as it is

PR #5531, opened 2020-12-21
Fixes for Happo Github workflow

suzubara:
Fixes the node version to our .node-version (12.16.3) when running the Happo Github workflow. Also splits the master Happo run into a separate workflow to bypass the paths-filter logic (so Happo runs on all master pushes).

### DIFF
--- a/.github/workflows/happo-tests-master.yml
+++ b/.github/workflows/happo-tests-master.yml
@@ -1,4 +1,4 @@
-name: Happo CI
+name: Happo CI (master)
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


.github/workflows/happo-tests-master.yml
```diff
@@ -0,0 +1,26 @@
+name: Happo CI (master)
+
+on:
+  push:
```
duncan-truss, line 4: You changed the other one to `pull_requests` does this one stay the `push` event?

duncan-truss (approved):

.github/workflows/happo-tests-master.yml
suzubara: Yeah, I split the single workflow into two (one for triggering Happo on PRs to master, and one for triggering Happo on all pushes to master)

duncan-truss: Ahh ok that push is the merge

PR #5508, opened 2020-12-17
[MB-5925] DTOD SOAP Connection

jacquelineIO:
## Description

First part of the DTOD service object that will get distance mileage from the DTOD SOAP service. 
This is able to connect successfully to the service and send a request. 

**Not implemented**
The response is not parsed for this PR, but will be part of the [next story](https://dp3.atlassian.net/browse/MB-5926) to implement parsing the response.

## Reviewer Notes

Is there anything you would like reviewers to give additional scrutiny?

### Setup local env

To run these changes locally, you'll have to make sure the `DTOD_API_*` envs are available in your environment. Following [`Setup:Direnv`](https://github.com/transcom/mymove#setup-direnv).

**Check if you already have the envs set**
```sh
grep DTOD_API .envrc.local
```
If they are set, make sure they match the ones from chamber. Verify the ones in dev chamber by running
```sh
 DISABLE_AWS_VAULT_WRAPPER=1 AWS_REGION=us-gov-west-1 aws-vault exec transcom-gov-dev -- chamber env app-devlocal --retries=1 | grep DTOD
```

**Add DTOD envs to your local envrc**
```
 DISABLE_AWS_VAULT_WRAPPER=1 AWS_REGION=us-gov-west-1 aws-vault exec transcom-gov-dev -- chamber env app-devlocal --retries=1 | grep DTOD >> .envrc.local
```
### Run local dev UT

Since this is a service that connects to an outside service. There are no already running UT. But there is a UT that can be un-commented out to test the code locally. 

After running `make server_test` un-comment out the code in the the test function `TestDTODZip5Distance` then run the test:
```sh
go test ./pkg/services/dtod_planner -v
```

**Additional Setup**
To view the request and response to the SOAP service un-comment out [lines 182-188 in the file dtod_zip5_distance.go](https://github.com/transcom/mymove/blob/3c446334bdc3841bdd0a5abeadaafb74161f1232/pkg/services/dtod_planner/dtod_zip5_distance.go#L183). And comment out the line using  `gosoap.SoapClient`.

**Format response XML**
If you run with `gosoap.SoapClientWithConfig` you see both the request and response on the screen. To format the response. Copy the response and do the following:
```
echo '<PASTE XML HERE>'  | xmllint --format -
```

## Code Review Verification Steps

* [ ] If the change is risky, it has been tested in experimental before merging.
* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#logging)
[...]

PR #5501, opened 2020-12-16
[MB-4584, MB-5940, MB-5941, MB-5942] Adds skeleton for GHC planner

reggieriser:
## Description

This PR sets up the skeleton for the new GHC planner.  We'll later be adding in the implementations that talk to DTOD (for zip5-to-zip5) and RM (for zip3-to-zip3).

See the [design doc](https://docs.google.com/document/d/1ApDDnPpwuV3s-XwAcK1XFspEg34UGox4ehKlCaTmnQM/edit) for more details.

## Reviewer Notes

Should I go any further in this story beyond what's here?  It's truly a skeleton right now, but glad to add more to it if needed.

## Setup

`make server_test` -- as this is just a skeleton, just make sure the tests pass (the tests just make sure the unimplemented methods panic for now).

## Code Review Verification Steps

* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#logging)
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4584) for this change.
  * [Subtask 1](https://dp3.atlassian.net/browse/MB-5940)
  * [Subtask 2](https://dp3.atlassian.net/browse/MB-5941)
  * [Subtask 3](https://dp3.atlassian.net/browse/MB-5942)
* [Design doc](https://docs.google.com/document/d/1ApDDnPpwuV3s-XwAcK1XFspEg34UGox4ehKlCaTmnQM/edit) explains more about the approach used.

PR #5543, opened 2020-12-23
[MB-5926,MB-5948, MB-5951, MB-6016] Parse/test DTOD response and integrate into GHC planner

reggieriser:
## Description

This PR expands on the work started in #5508 to connect to the DTOD SOAP API.  At a high level, this PR adds the following:
- parses the response from DTOD
- adds unit tests that use a mocked DTOD API
- integrates DTOD into the GHC planner skeleton from #5501 

## Reviewer Notes

I ran into several circular dependency problems that I had to work around.

One prevented me from keeping this in the services package (since something else in services already refers back to the `route` package).  To deal with that one, I moved the DTOD work into the `route` package itself (we had a similar problem with the RM integration).  This should be OK since the planner is really a service for the entire app.  I think it even makes some of the initialization more straightforward too.

The other prevented me from generating a mock for `SoapCaller` in the `mocks` directory because a different mock referred back to the `route` package already.  To work around this, I created a `ghcmocks` directory/package instead and placed `SoapCaller` in it.

If there's another straightforward way to handle either of these, please let me know.  I didn't want to get into a huge refactoring effort in this PR since we may have to refactor unrelated code to fix this without moving DTOD-related code.

Also, I noticed there were some old mocks that were in the `pkg/services/mocks` directory as I cleaned up one of mine in this PR after the refactor.  I am deleting those as well in this PR.

## Setup

The GHC planner is not hooked up anywhere yet.  That is coming in a future story.

If you haven't already set up your `envrc.local` before with DTOD-related environment variables, then see the `Setup local env` section of #5508.

You can `make server_test` and see that the mocked SOAP API works.

To see a real call go through, you could temporarily edit `serve.go` and add in the following lines after the GHC planner is instantiated:

```
distance, err := ghcRoutePlanner.Zip5TransitDistance("05030", "05091")
logger.Debug("DTOD distance", zap.Int("distance", distance))
```

When you rebuild (`rm bin\milmove && make server_build`) and then run your server (`make server_run`), you should see a couple of debug-level statements related to distance in the log output.  Change the zips as desired to get different results.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5926) for this change


pkg/route/ghc_planner.go
```diff
@@ -52,9 +57,12 @@ func (p *ghcPlanner) Zip3TransitDistance(source string, destination string) (int
 }
 
 // NewGHCPlanner constructs and returns a Planner for GHC routing.
-func NewGHCPlanner(db *pop.Connection, logger Logger) Planner {
+func NewGHCPlanner(logger Logger, db *pop.Connection, soapClient SoapCaller, dtodUsername string, dtodPassword string) Planner {
```
pearl-truss, line 60: The GHC Planner could be called just for Rand McNally,  in that case will it know the dtodUsername and password? Those seem to be coming from the CLI?

pkg/route/dtod_zip5_distance.go
```diff
@@ -0,0 +1,130 @@
+package route
+
+/*******************************************
```
pearl-truss, line 3: Thanks, this is great documentation to have in this file

pkg/route/ghc_planner.go
gidjin: The dtodUsername and password are stored in chamber and are passed via environment variables. I believe you can override that system with CLI flags to the milmove serve command.

pearl-truss (commented):
After going through the set up steps and re-running my server I get back a distance of 0 for the distance between these zips "05030" and "05091"

`2020-12-23T21:01:31.306Z        DEBUG   milmove/serve.go:585    DTOD distance   {"git_branch": "rr-MB-5926-dtod-response", "git_commit": "d23100e3de7cbf043a74fcfb2d54060059d40f86", "distance": 0}`

pkg/route/ghc_planner.go
gidjin: That said this is typical of our other such secrets for the HERE api for example, as I recall.

reggieriser: Right, the real GHC planner is instantiated at startup of the web server and configured for both the zip3 (RM) or zip5 (DTOD) calculations.  Similarly, DTOD calls don't need the DB connection that RM does (for now), but we need to include it as part of the planner setup.

pearl-truss: "real GHC planner is instantiated at startup" okay that makes sense!

reggieriser:
> After going through the set up steps and re-running my server I get back a distance of 0 for the distance between these zips "05030" and "05091"

Oh, I wonder if you don't have the `env.local` variables related to DTOD yet.  I just added a note about this to the setup instructions.  See if that fixes the problem.

gidjin (approved):
LGTM, Tried this out and seems to return correct distances for the values I tried.

pkg/route/dtod_zip5_distance.go
```diff
@@ -0,0 +1,130 @@
...
+	params := gosoap.Params{
+		"DtodRequest": map[string]interface{}{
+			"AuthToken": map[string]interface{}{
+				"Username": d.username,
+				"Password": d.password,
+			},
+			"UserRequest": map[string]interface{}{
+				"Function": "Distance",
+				// TODO: Default RouteType is PcsTdyTravel, but CommercialPersonalProperty seems better. Verify.
+				"RouteType": "CommercialPersonalProperty",
```
jacquelineIO, line 96: HHGs/PPMs moves are PCS's which stand for "permanent change of station" that's also what is selected when you onboard as the service member if I recall correctly. Better to verify, but that's why I left it as default. "CommericalPersonalProperty" to me sounds like a true commercial move and not a DoD sponsored move.

jacquelineIO: TDY is also a DoD term for temporary duty (which is basically business travel)

PR #5561, opened 2020-12-30
[MB-6064] Milmove card

Ronolibert:
## Description

Adds a Milmove 101 card to storybook

## Setup
`make storybook`
View MilmoveCard component

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6064) for this change

## Screenshots

milmove_mobile


src/components/Customer/MilmoveCard/MilmoveCard.module.scss
```diff
@@ -0,0 +1,72 @@
...
+      .CounselorListHeader {
+        margin-bottom: 0px;
+      }
+    }
+  }
+}
+
+@media screen and (max-width: $mobile) {
+  .CardContainer {
+    border-width: 0px !important;
```
Ronolibert, line 49: Even with :global I couldn't specificity priority with some other classes that are applying this border

src/components/Customer/MilmoveCard/MilmoveCard.module.scss
```diff
@@ -0,0 +1,72 @@
...
+.Divider {
+  width: 100%;
+  height: 1px;
+  margin-top: 24px;
+  border: solid 1px #dcdee0;
+}
+
+.CardContainer {
+  max-width: 675px;
+  #CardBody {
```
Ronolibert, line 12: Had to use ID to get more specificity because there was a selector with a good bit more priority

src/components/Customer/MilmoveCard/MilmoveCard.module.scss
```diff
@@ -0,0 +1,72 @@
+@import 'shared/styles/_variables';
+
+.Divider {
+  width: 100%;
+  height: 1px;
+  margin-top: 24px;
+  border: solid 1px #dcdee0;
```
Ronolibert, line 7: Currently not a color that is in our system but is in the design

src/components/Customer/MilmoveCard/MilmoveCard.module.scss
suzubara: we should never apply styles using IDs unless absolutely necessary.. I think we can figure out how to select this element without. let me know if you'd like to 🍐 on it!

mariaclaratgolfo:
@Ronolibert I want to confirm that this is not 1 long card with 4 sections, but that it's 4 separate cards?

suzubara (changes requested):
I think the idea here was that a single MilMove Card would just be a generic content card that applies some minor CSS changes to the ReactUSWDS card, and then the content from the designs would be rendered inside each `<MilmoveCard>` instance on the Moving 101 page, and in Storybook, etc. Each of the sections in the design would be a single `MilmoveCard` instead of all of the content inside one. I imagine that will also result in a lot of CSS changes. let me know if you'd like to 🍐 !

Ronolibert:
> @Ronolibert I want to confirm that this is not 1 long card with 4 sections, but that it's 4 separate cards?

I misunderstood! We changed it to be a generic card that the developer who uses it will fill in the content for

src/components/Customer/MilmoveCard/MilmoveCard.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { CardGroup, Card, CardHeader, CardBody } from '@trussworks/react-uswds';
+import { node, string } from 'prop-types';
+
+import * as styles from './MilmoveCard.module.scss';
+
+const MilmoveCard = ({ headerText, children }) => (
+  <CardGroup>
```
suzubara, line 8: I think since this component represents a single `Card`, we'll want to omit `CardGroup` so that it can be used to wrap multiple Cards. For example on the Moving 101 page, we'll end up with something like:

```
<CardGroup>
  <MilmoveCard ...>
  <MilmoveCard ...>
</CardGroup>
```

src/components/Customer/MilmoveCard/MilmoveCard.module.scss
```diff
@@ -0,0 +1,23 @@
+@import 'shared/styles/_variables';
+
+.CardBody {
+  padding-top: 16px;
+}
+
+@media screen and (max-width: $mobile) {
```
suzubara, line 7: in general we should be writing mobile-first CSS, and _only_ using `min-width` media queries as much as possible. that way the mobile CSS is the default, and less work is needed to adapt the CSS as the screen size increases.

suzubara (commented):
this is way better! I left a couple of minor comments. One thing that might make this component easier to work with is if `MilmoveCard` _only_ renders `children`, and instances of it can use `CardHeader, CardMedia, CardBody`, etc. directly. so for example on the Moving 101 page:

```
<MilmoveCard>
  <CardHeader><h3>header text</h3></CardHeader>
  <CardMedia><img ... />
  // etc.
</MilmoveCard>
```

it looks to me based on the design that they should use the `exdent` style, and images should be rendered in the `CardMedia` component (see the "Exdent Card" example on https://trussworks.github.io/react-uswds/?path=/docs/card--card-examples) so custom padding doesn't need to be applied

src/components/Customer/MilmoveCard/MilmoveCard.module.scss
Ronolibert: Since using the built in props for the cards, these styles won't be needed anymore!

Ronolibert:
@suzubara Added another example for the exdented version and one for the inset version

mariaclaratgolfo (approved):

src/components/Customer/MilmoveCard/MilmoveCard.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+import * as styles from './MilmoveCard.module.scss';
+
+const MilmoveCard = ({ children, headerFirst }) => (
+  <Card containerProps={{ className: styles.CardContainer }} headerFirst={headerFirst}>
+    <>{children}</>
+    {/* <CardHeader>
+        <h3>{headerText}</h3>
+      </CardHeader>
+      <CardBody className={styles.CardBody}>{children}</CardBody> */}
```
suzubara, line 13: can we remove the commented out code? I think `{children}` doesn't have to be in a fragment either

src/components/Customer/MilmoveCard/MilmoveCard.module.scss
```diff
@@ -0,0 +1,6 @@
+@import 'shared/styles/_variables';
+
+.CardContainer {
+  border-width: 0;
+  border-bottom: 1px solid #dcdee0;
+}
```
suzubara, line 6: I think we want to add `border-radius: 0;` also, it looks like it's curved on the corners

suzubara (commented):
this is looking good! a couple minor last things. I also think we should make sure to get @transcom/truss-design approval

suzubara (approved):
oh I saw @mariaclaratgolfo already approved! I'm going to approve so that I don't block this since I'm OOO the rest of the week but if we can take care of those last few things that would be great 🎉

PR #5575, opened 2020-12-31
MB-6104 Set document viewer react-query to not refetch onFocus

donaldthai:
## Description

> Motivation is to improve the user experience of not having to refetch document data.
> 
> AC: Disable refetch onFocus for document viewer & document doesn’t reload navigating to either orders/allowances page

## Reviewer Notes

- cache/stale time is set to 15 mins
- refetchOnFocus is set to false

## Setup

```sh
make server_run
make office_client_run
```

Test on experimental:
1. Log in as TOO on `office.exp.move.mil`
2. Click on any move
3. Click on `View & edit orders`
4. Click on `View allowances` and notice that the document doesn't reload

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6104) for this change
* [this article](https://react-query.tanstack.com/reference/useQuery#_top) explains more about the approach used. `refetchOnWindowFocus` is set to disabled.

## Screenshots

### DIFF
--- a/src/hooks/queries.js
+++ b/src/hooks/queries.js
@@ -164,11 +164,15 @@ export const useOrdersDocumentQueries = (moveCode) => {
 
   // Get a document
   // TODO - "upload" instead of "uploads" is because of the schema.js entity name. Change to "uploads"
+  const staleTime = 60 * 60000; // 60 * 60000 milliseconds = 1 hour
+  const cacheTime = staleTime;
   const { data: { documents, upload } = {}, ...ordersDocumentsQuery } = useQuery(
     [ORDERS_DOCUMENTS, documentId],
     getDocument,
     {
       enabled: !!documentId,
+      staleTime,
+      cacheTime,
     },
   );
 
--- a/src/hooks/queries.js
+++ b/src/hooks/queries.js
@@ -164,15 +164,12 @@ export const useOrdersDocumentQueries = (moveCode) => {
 
   // Get a document
   // TODO - "upload" instead of "uploads" is because of the schema.js entity name. Change to "uploads"
-  const staleTime = 60 * 60000; // 60 * 60000 milliseconds = 1 hour
-  const cacheTime = staleTime;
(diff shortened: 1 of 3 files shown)


src/hooks/queries.js
```diff
@@ -164,11 +164,15 @@ export const useOrdersDocumentQueries = (moveCode) => {
 
   // Get a document
   // TODO - "upload" instead of "uploads" is because of the schema.js entity name. Change to "uploads"
+  const staleTime = 60 * 60000; // 60 * 60000 milliseconds = 1 hour
```
duncan-truss, line 167: Yea like we were discussing I think at most this could be 15 minutes because of that x-amz-expires header being set to 900 seconds.  Otherwise we'd end up getting bad results back in the document viewer, which is maybe an error we should be testing for anyway 🤔

donaldthai: 👍 yea, let me set the `refetchOnWindowFocus` to false to see if that solves the issue for us and remove the stale/cache time to test.

donaldthai: I also removed the lazy loading for the forms.

donaldthai: Hmm, so the onFocus is no longer a problem but the document still reloads after switching views for orders/allowances.  Guess I'll set the cache/stale time after all. I'll set it to 15 mins and we can track a story to ask infra to potentially increase the s3 header time.

duncan-truss: In the devtools network tab does it actually say the pdf request is cached the second time?

donaldthai: From the looks of it, it doesn't seem like it since the reponse headers look different

duncan-truss (approved):

src/hooks/queries.js
donaldthai: Ok, so setting the cache/stale time won't try to refetch for the document within the 15 min timeframe. Don't see the network call there after a second page navigation.

PR #5592, opened 2021-01-05
[MB-6089] Use ConflictError instead of InvalidInputError in updatePaymentRequestStatus

mboren:
## Description
I changed the error response you get when attempting to update the payment request status for a payment request that hasn't had all of its items approved or denied. 
We used to return an `InvalidInputError`, which was converted to an HTTP 500 response by the handler.

We only want 500s to happen when our code is broken, so I converted this to a `ConflictError`, which results in a 409 response.

## Setup

```sh
make bin/prime-api-client
make db_dev_e2e_populate
make server_run
```

Create an input file for creating a payment request. Let's call it `payment_request.json`.
*Note that, if there are any FSC or DSH service items in the payment request, the request will fail. (This is due to an issue with the DTOD API that is being investigated right now, and is unrelated to this PR.*
```json
{
  "body": {
    "isFinal": false,
    "serviceItems": [
      {
        "id": "94bc8b44-fefe-469f-83a0-39b1e31116fb"
      },
      {
      }
    ]
  }
}
```

Create the payment request, and use `jq` to generate the input file for the next step
```sh
prime-api-client --insecure create-payment-request --filename payment_request.json | jq "{paymentRequestID: .paymentServiceItems[0].paymentRequestID, ifMatch: .eTag, body: {status: \"REVIEWED\"}}" > update_payment_request_status.json
```
Attempt to update the payment request status
```
prime-api-client --insecure support-update-payment-request-status --filename update_payment_request_status.json
```
Verify that you get a response with status code 409.
Example output:
```
HTTP/1.1 409 Conflict
Content-Length: 236
Cache-Control: no-cache, no-store, must-revalidate
Content-Security-Policy: frame-ancestors 'none'
Content-Type: application/json
Date: Tue, 05 Jan 2021 22:26:27 GMT
Strict-Transport-Security: max-age=31536000; includeSubdomains; preload
X-Frame-Options: deny
X-Xss-Protection: 1; mode=block

{"detail":"id: 6260639e-b10d-4d84-bbed-b74597dfe355 is in a conflicting state All PaymentServiceItems must be approved or denied to review this PaymentRequest","instance":"ecf7627a-9fa8-4932-830d-5c508342d3a9","title":"Conflict Error"}

{"Payload":{"detail":"id: 6260639e-b10d-4d84-bbed-b74597dfe355 is in a conflicting state All PaymentServiceItems must be approved or denied to review this PaymentRequest","instance":"ecf7627a-9fa8-4932-830d-5c508342d3a9","title":"Conflict Error"}}%
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6089) for this change

### DIFF
--- a/pkg/handlers/supportapi/payment_request.go
+++ b/pkg/handlers/supportapi/payment_request.go
@@ -114,6 +114,8 @@ func (h UpdatePaymentRequestStatusHandler) Handle(params paymentrequestop.Update
 			return paymentrequestop.NewUpdatePaymentRequestStatusNotFound().WithPayload(payloads.ClientError(handlers.NotFoundMessage, err.Error(), h.GetTraceID()))
 		case services.PreconditionFailedError:
 			return paymentrequestop.NewUpdatePaymentRequestStatusPreconditionFailed().WithPayload(payloads.ClientError(handlers.PreconditionErrMessage, err.Error(), h.GetTraceID()))
+		case services.ConflictError:
+			return paymentrequestop.NewUpdatePaymentRequestStatusConflict().WithPayload(payloads.ClientError(handlers.ConflictErrMessage, err.Error(), h.GetTraceID()))
 		default:
 			logger.Error(fmt.Sprintf("Error saving payment request status for ID: %s: %s", paymentRequestID, err))
 			return paymentrequestop.NewUpdatePaymentRequestStatusInternalServerError().WithPayload(payloads.InternalServerError(handlers.FmtString(err.Error()), h.GetTraceID()))
--- a/pkg/gen/supportapi/embedded_spec.go
+++ b/pkg/gen/supportapi/embedded_spec.go
@@ -559,6 +559,9 @@ func init() {
           "404": {
             "$ref": "#/responses/NotFound"
           },
+          "409": {
+            "$ref": "#/responses/Conflict"
+          },
           "412": {
             "$ref": "#/responses/PreconditionFailed"
           },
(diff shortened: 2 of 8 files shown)


pkg/handlers/supportapi/payment_request.go
```diff
@@ -114,6 +114,8 @@ func (h UpdatePaymentRequestStatusHandler) Handle(params paymentrequestop.Update
 			return paymentrequestop.NewUpdatePaymentRequestStatusNotFound().WithPayload(payloads.ClientError(handlers.NotFoundMessage, err.Error(), h.GetTraceID()))
 		case services.PreconditionFailedError:
 			return paymentrequestop.NewUpdatePaymentRequestStatusPreconditionFailed().WithPayload(payloads.ClientError(handlers.PreconditionErrMessage, err.Error(), h.GetTraceID()))
+		case services.ConflictError:
+			return paymentrequestop.NewUpdatePaymentRequestStatusConflict().WithPayload(payloads.ClientError(handlers.ConflictErrMessage, err.Error(), h.GetTraceID()))
```
reggieriser, line 118: Should we add a unit test for this path? It looks like there are tests for the other error paths here. I think we should be able to mock the service object and force this error like those do.

reggieriser (commented):
Things look good and worked as expected when I tested it. 👍  The only improvement I noted was to add a unit test on the handler itself (similar to the other error paths).

pkg/handlers/supportapi/payment_request.go
mboren: Good idea, I added a test for this.

reggieriser (approved):
Thanks for adding the additional test.  LGTM!  🚀

PR #5595, opened 2021-01-05
MB-6113 SupportAPI: Fix updateMTOServiceItemStatus to return the proper payload

shimonacarvalho:
## Description

SORRY the ticket number in the branch name is wrong. It's mb-6113.

This ticket is to fix the support endpoint `updateMTOServiceItemStatus` that was returning a old format of payload. We need the polymorphic service item that we use everywhere else. So deleted the bad definition, pasted in the correct one from primeAPI and then fixed the model to payload functions, etc. And fixed the tests. 

I needed to fix the path of the endpoint too which was `/service-items/id` instead of `/mto-service-items/id`.
I also fixed `getMTO` to return the service items because that is a super useful support endpoint. 

## Setup

You can use postman or prime-api-client to update the status on a service item at the following endpoint.
`PATCH {{baseUrl}}/{{support}}/mto-service-items/:mtoServiceItemID/status`

3 tests were updated. So you can also test by running the tests in 
`go test ./pkg/handlers/supportapi/ -v`

## References

* [MB-6113 Support API updateMTOServiceItemStatus returns a malformed service item](https://dp3.atlassian.net/browse/MB-6113)


pkg/handlers/primeapi/payloads/model_to_payload.go
```diff
@@ -424,8 +424,8 @@ func MTOServiceItem(mtoServiceItem *models.MTOServiceItem) primemessages.MTOServ
 		if mtoServiceItem.SITDepartureDate != nil {
 			sitDepartureDate = *mtoServiceItem.SITDepartureDate
 		}
-		firstContact := getCustomerContact(mtoServiceItem.CustomerContacts, models.CustomerContactTypeFirst)
-		secondContact := getCustomerContact(mtoServiceItem.CustomerContacts, models.CustomerContactTypeSecond)
+		firstContact := GetCustomerContact(mtoServiceItem.CustomerContacts, models.CustomerContactTypeFirst)
```
shimonacarvalho, line 427: We are using these functions in the support api

pkg/handlers/supportapi/internal/payloads/model_to_payload.go
```diff
@@ -28,6 +31,7 @@ func MoveTaskOrder(moveTaskOrder *models.Move) *supportmessages.MoveTaskOrder {
 		return nil
 	}
 	mtoShipments := MTOShipments(&moveTaskOrder.MTOShipments)
+	mtoServiceItems := MTOServiceItems(&moveTaskOrder.MTOServiceItems)
```
shimonacarvalho, line 34: This adds back MTOServiceItems into the getMTO payload which was really useful for testing and debugging this pr, and will in general be useful for us.

pkg/handlers/supportapi/internal/payloads/model_to_payload.go
```diff
@@ -241,20 +247,103 @@ func MTOShipment(mtoShipment *models.MTOShipment) *supportmessages.MTOShipment {
...
-		ETag:            etag.GenerateEtag(mtoServiceItem.UpdatedAt),
-		ID:              strfmt.UUID(mtoServiceItem.ID.String()),
-		MoveTaskOrderID: strfmt.UUID(mtoServiceItem.MoveTaskOrderID.String()),
-		MtoShipmentID:   strfmt.UUID(mtoServiceItem.MTOShipmentID.String()),
-		Status:          supportmessages.MTOServiceItemStatus(mtoServiceItem.Status),
-		RejectionReason: mtoServiceItem.RejectionReason,
-	}
-
+func MTOServiceItem(mtoServiceItem *models.MTOServiceItem) supportmessages.MTOServiceItem {
+	var payload supportmessages.MTOServiceItem
```
shimonacarvalho, line 251: Basically copied over the model to payload function from the PrimeAPI side.

pkg/handlers/supportapi/internal/payloads/model_to_payload.go
```diff
@@ -241,20 +247,103 @@ func MTOShipment(mtoShipment *models.MTOShipment) *supportmessages.MTOShipment {
...
+
+	// set all relevant fields that apply to all service items
+	var shipmentIDStr string
+	if mtoServiceItem.MTOShipmentID != nil {
+		shipmentIDStr = mtoServiceItem.MTOShipmentID.String()
+	}
+
+	one := mtoServiceItem.ID.String()
+	two := strfmt.UUID(one)
+	payload.SetID(two)
```
sandy-wright, line 326: The names `one` and `two` here are kind of confusing - maybe we could rename them to something like `serviceItemIDStr` and `serviceItemID`?

swagger/support.yaml
```diff
@@ -314,7 +314,7 @@ paths:
           $ref: '#/responses/UnprocessableEntity'
         '500':
           $ref: '#/responses/ServerError'
-  '/service-items/{mtoServiceItemID}/status':
+  '/mto-service-items/{mtoServiceItemID}/status':
```
sandy-wright, line 317: Thank you for this - consistency at last 🙏

sandy-wright (approved):
LGTM! Testing in Postman and it worked like a charm. I left one small comment about variable names, but everything else was golden.

I do have a couple of notes for some future changes (NOT within the purview of this PR, for sure, just putting them here for posterity):
- `uploadedOrders` info isn't populated when I use the `getMoveTaskOrder` endpoint (although it is a required field, so we have to have data there)
- `updateMTOServiceItemStatus` returns a ConflictError when the user tries to reject without a rejection reason. This would ideally be an InvalidInputError

But also, this is a support API! So we can throw these issues to the bottom of the pile 😄

pkg/handlers/supportapi/internal/payloads/model_to_payload.go
shimonacarvalho: Haha yeah this whole function was just copy-pasted in, but when I saw the one/two in the prime payload file, I had the same thought! will fix here.

shimonacarvalho:
@sandy-wright GOOD point about that ConflictError! I think it should really be fixed in this PR - since this PR was really about getting that endpoint in correct working order. So I fixed that. I did not fix the getMTO issue, since that's a different endpoint. Thanks for the recommendation Sandy.

>     * `updateMTOServiceItemStatus` returns a ConflictError when the user tries to reject without a rejection reason. This would ideally be an InvalidInputError
>

PR #5598, opened 2021-01-06
MB-5696 Fix bugs around updating a Move's status

monfresh:
## Description
Service items can only be created on a Move whose status is either
Approved or Approvals Requested. However, this was not properly
enforced in `mto_service_item_creator`. The status update on the move
was performed outside of the transaction that created the service
items. So if service items were created on an invalid move, even
though the status update failed, the service items were still created.

I also found a bug in `mto_service_item_updater`, where if a service
item's status was set back to SUBMITTED, the move's status wouldn't
get reset to APPROVALS REQUESTED, and so the TOO wouldn't know to
review it.

To fix the first bug, I added a check at the beginning of the
service item creation function to check and fail early if the move's
status is not 'Approved' or 'Approvals Requested', because if it's
not, nothing in the rest of the function matters. I also updated the
error message to be more friendly for the Prime.

I added several missing tests around these two bugs, and I plan on
adding more, as well as doing some refactoring to make the code
easier to understand.

## Reviewer Notes

1. Using the Prime API, try to create a service item on a Move whose status is neither `Approved` nor `Approvals Requested`. You should get an error of type ConflictError (status code 409) with message `Cannot create service items before a move has been approved` and the Move's ID and current status for debugging. No service items should be created on this move.

2. Using the Prime API, create a service item on an approved Move. Verify that after the operation is successful, the Move's status has changed to `Approvals Requested`.

3. Using the Support API, update the status of an existing service item (where the Move is in Approved Status) to `SUBMITTED`. Verify that the Move's status is changed to `Approvals Requested`.

4. Using the Support API, update the status of an existing shipment (where the Move is NOT in Approved Status, such as `Submitted`) to `Approved`. Verify that you get an error saying you can't approve a shipment if the move is not approved. This is because approving a shipment automatically creates service items on the moves, but since you can't create service items on a move that isn't approved, updating the shipment status should fail.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5696) for this change

### DIFF
--- a/pkg/models/move.go
+++ b/pkg/models/move.go
@@ -172,8 +172,12 @@ func (m *Move) Approve() error {
 
 // SetApprovalsRequested sets the move to approvals requested
 func (m *Move) SetApprovalsRequested() error {
+	// Do nothing if it's already in the desired state
+	if m.Status == MoveStatusAPPROVALSREQUESTED {
+		return nil
+	}
 	if m.Status != MoveStatusAPPROVED {
-		return errors.Wrap(ErrInvalidTransition, fmt.Sprintf("Cannot move to Approvals Requested when the Move is not in an Approved state for status: %s and ID: %s", m.Status, m.ID))
+		return errors.Wrap(ErrInvalidTransition, fmt.Sprintf("The status for the Move with ID %s can only be set to 'Approvals Requested' from the 'Approved' status, but its current status is %s.", m.ID, m.Status))
 	}
 	m.Status = MoveStatusAPPROVALSREQUESTED
 	return nil
--- a/pkg/services/mto_service_item/mto_service_item_creator.go
+++ b/pkg/services/mto_service_item/mto_service_item_creator.go
@@ -187,21 +187,16 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
 			}
 		}
 
-		// Once the service item is successfuly created, the Move's status needs to
+		// Once the service item is successfully created, the Move's status needs to
 		// be updated to 'Approvals Requested' if it's not already in that state,
 		// which will let the TOO know they need to review it.
-		if move.Status != models.MoveStatusAPPROVALSREQUESTED {
-			// We're not checking for an error here because this will only error if
-			// the move's status is not 'Approved', and it's not possible for a Move
-			// that's not 'Approved' to get to this line of code because at the
-			// beginning of this function, we only allow moves with status 'Approved'
-			// or 'Approvals Requested'.
-			move.SetApprovalsRequested()
-
-			verrs, err = txBuilder.UpdateOne(&move, nil)
-			if verrs != nil || err != nil {
-				return fmt.Errorf("%#v %e", verrs, err)
-			}
+		err = move.SetApprovalsRequested()
+		if err != nil {
+			return fmt.Errorf("%e", err)
+		}
+		verrs, err = txBuilder.UpdateOne(&move, nil)
+		if verrs != nil || err != nil {
+			return fmt.Errorf("%#v %e", verrs, err)
 		}
 
 		return nil
--- a/pkg/services/mto_service_item/mto_service_item_creator.go
+++ b/pkg/services/mto_service_item/mto_service_item_creator.go
@@ -187,6 +187,23 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
 			}
 		}
 
+		// Once the service item is successfuly created, the Move's status needs to
+		// be updated to 'Approvals Requested' if it's not already in that state,
+		// which will let the TOO know they need to review it.
+		if move.Status != models.MoveStatusAPPROVALSREQUESTED {
+			// We're not checking for an error here because this will only error if
+			// the move's status is not 'Approved', and it's not possible for a Move
+			// that's not 'Approved' to get to this line of code because at the
+			// beginning of this function, we only allow moves with status 'Approved'
+			// or 'Approvals Requested'.
+			move.SetApprovalsRequested()
+
+			verrs, err = txBuilder.UpdateOne(&move, nil)
+			if verrs != nil || err != nil {
+				return fmt.Errorf("%#v %e", verrs, err)
+			}
+		}
+
 		return nil
 	})
 
@@ -194,23 +211,6 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
 		return nil, verrs, nil
 	} else if err != nil {
 		return nil, verrs, services.NewQueryError("unknown", err, "")
-	}
-
-	// Once the service item is successfuly created, the Move's status needs to
-	// be updated to 'Approvals Requested' if it's not already in that state,
-	// which will let the TOO know they need to review it.
-	if move.Status != models.MoveStatusAPPROVALSREQUESTED {
-		// We're not checking for an error here because this will only error if
-		// the move's status is not 'Approved', and it's not possible for a Move
-		// that's not 'Approved' to get to this line of code because at the
-		// beginning of this function, we only allow moves with status 'Approved'
-		// or 'Approvals Requested'.
-		move.SetApprovalsRequested()
-
-		verrs, err := o.builder.UpdateOne(&move, nil)
-		if verrs != nil || err != nil {
-			return nil, verrs, err
-		}
 	}
 
 	return &createdServiceItems, nil, nil
--- a/pkg/services/mto_service_item/mto_service_item_creator.go
+++ b/pkg/services/mto_service_item/mto_service_item_creator.go
@@ -106,7 +106,6 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
 
 	if serviceItem.ReService.Code == models.ReServiceCodeDOSHUT || serviceItem.ReService.Code == models.ReServiceCodeDDSHUT {
 		if mtoShipment.PrimeEstimatedWeight == nil {
-			fmt.Println("PrimeEstimatedWeight is nil")
 			return nil, verrs, services.NewConflictError(reService.ID, "for creating a service item. MTOShipment associated with this service item must have a valid PrimeEstimatedWeight.")
 		}
 	}
--- a/pkg/services/mto_service_item/mto_service_item_creator.go
+++ b/pkg/services/mto_service_item/mto_service_item_creator.go
@@ -45,6 +45,15 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
 		return nil, nil, services.NewNotFoundError(moveID, "in Moves")
 	}
 
+	// Service items can only be created if a Move's status is either Approved
+	// or Approvals Requested, so check and fail early.
+	if move.Status != models.MoveStatusAPPROVED && move.Status != models.MoveStatusAPPROVALSREQUESTED {
+		return nil, nil, services.NewConflictError(
+			move.ID,
+			fmt.Sprintf("Cannot create service items before a move has been approved. The current status for the move with ID %s is %s", move.ID, move.Status),
+		)
+	}
+
 	// find the re service code id
 	var reService models.ReService
 	reServiceCode := serviceItem.ReService.Code
@@ -97,6 +106,7 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
 
 	if serviceItem.ReService.Code == models.ReServiceCodeDOSHUT || serviceItem.ReService.Code == models.ReServiceCodeDDSHUT {
 		if mtoShipment.PrimeEstimatedWeight == nil {
+			fmt.Println("PrimeEstimatedWeight is nil")
 			return nil, verrs, services.NewConflictError(reService.ID, "for creating a service item. MTOShipment associated with this service item must have a valid PrimeEstimatedWeight.")
 		}
 	}
@@ -187,12 +197,17 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
 		return nil, verrs, services.NewQueryError("unknown", err, "")
 	}
 
-	// TODO: Determine if this should be in the same transaction as the service items (they get created even if this fails)
+	// Once the service item is successfuly created, the Move's status needs to
+	// be updated to 'Approvals Requested' if it's not already in that state,
+	// which will let the TOO know they need to review it.
 	if move.Status != models.MoveStatusAPPROVALSREQUESTED {
-		err := move.SetApprovalsRequested()
-		if err != nil {
-			return nil, nil, services.NewConflictError(move.ID, err.Error())
-		}
+		// We're not checking for an error here because this will only error if
+		// the move's status is not 'Approved', and it's not possible for a Move
+		// that's not 'Approved' to get to this line of code because at the
+		// beginning of this function, we only allow moves with status 'Approved'
+		// or 'Approvals Requested'.
+		move.SetApprovalsRequested()
+
 		verrs, err := o.builder.UpdateOne(&move, nil)
 		if verrs != nil || err != nil {
 			return nil, verrs, err
--- a/pkg/services/mto_service_item/mto_service_item_creator_test.go
+++ b/pkg/services/mto_service_item/mto_service_item_creator_test.go
@@ -66,7 +66,7 @@ func (suite *MTOServiceItemServiceSuite) buildValidServiceItemWithInvalidMove() 
 		Move: move,
 	})
 
-	serviceItemForUnapprovedMTO := models.MTOServiceItem{
+	serviceItemForUnapprovedMove := models.MTOServiceItem{
 		MoveTaskOrderID: move.ID,
 		MoveTaskOrder:   move,
 		ReService:       reServiceDDFSIT,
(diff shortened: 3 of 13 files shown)


pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -187,12 +196,17 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
...
-			return nil, nil, services.NewConflictError(move.ID, err.Error())
-		}
+		// We're not checking for an error here because this will only error if
+		// the move's status is not 'Approved', and it's not possible for a Move
+		// that's not 'Approved' to get to this line of code because at the
+		// beginning of this function, we only allow moves with status 'Approved'
+		// or 'Approvals Requested'.
+		move.SetApprovalsRequested()
+
 		verrs, err := o.builder.UpdateOne(&move, nil)
```
duncan-truss, line 210: This needs to use txBuilder in the same block as the service item creation to rollback on failure right?

monfresh: Yeah, good catch.

pkg/models/move.go
```diff
@@ -172,8 +172,12 @@ func (m *Move) Approve() error {
 
 // SetApprovalsRequested sets the move to approvals requested
 func (m *Move) SetApprovalsRequested() error {
+	// Do nothing if it's already in the desired state
+	if m.Status == MoveStatusAPPROVALSREQUESTED {
+		return nil
+	}
```
sandy-wright, line 178: nice, thank you!

pkg/services/mto_service_item/mto_service_item_creator.go
sandy-wright: Yeah, move this entire block before the `return nil` on line 190 and then it should be golden

pkg/services/mto_service_item/mto_service_item_creator_test.go
```diff
@@ -42,55 +42,124 @@ func (t *testCreateMTOServiceItemQueryBuilder) Transaction(fn func(tx *pop.Conne
...
+	testdatagen.MakeReService(suite.DB(), testdatagen.Assertions{
+		ReService: models.ReService{
+			Code: models.ReServiceCodeDDDSIT,
 		},
+	})
+	reServiceDDFSIT := testdatagen.MakeReService(suite.DB(), testdatagen.Assertions{
+		ReService: models.ReService{
+			Code: models.ReServiceCodeDDFSIT,
+		},
+	})
```
sandy-wright, line 64: `testdatagen.MakeDDFSITReService` will create all three of these service codes at once for you.

sandy-wright: It returns only the DDFSIT service.

pkg/services/mto_service_item/mto_service_item_creator_test.go
```diff
@@ -42,55 +42,124 @@ func (t *testCreateMTOServiceItemQueryBuilder) Transaction(fn func(tx *pop.Conne
...
+	testdatagen.MakeReService(suite.DB(), testdatagen.Assertions{
+		ReService: models.ReService{
+			Code: models.ReServiceCodeDDDSIT,
+		},
+	})
+	reServiceDDFSIT := testdatagen.MakeReService(suite.DB(), testdatagen.Assertions{
+		ReService: models.ReService{
+			Code: models.ReServiceCodeDDFSIT,
+		},
+	})
```
sandy-wright, line 105: Same dealio, you can use `reServiceDDFSIT := testdatagen.MakeDDFSITReService(suite.DB())` to simplify this process for ya 😄

pkg/services/mto_service_item/mto_service_item_creator_test.go
```diff
@@ -42,55 +42,124 @@ func (t *testCreateMTOServiceItemQueryBuilder) Transaction(fn func(tx *pop.Conne
...
+// Should return a message stating that service items can't be created if
+// the move is not in approved status.
+func (suite *MTOServiceItemServiceSuite) TestCreateMTOServiceItemWithInvalidMove() {
+	builder := query.NewQueryBuilder(suite.DB())
+	creator := NewMTOServiceItemCreator(builder)
+	serviceItemForUnapprovedMove := suite.buildValidServiceItemWithInvalidMove()
+
+	createdServiceItem, _, err := creator.CreateMTOServiceItem(&serviceItemForUnapprovedMove)
+
+	suite.DB().Find(&serviceItemForUnapprovedMove, serviceItemForUnapprovedMove.ID)
```
sandy-wright, line 131: What is the purpose of this database call? We're not going to be able to find the service item because `serviceItemForUnapprovedMove.ID` should be a nil UUID always - it's just a model struct at this point, not a DB record. I'm also surprised that this wasn't causing an error 🤔  (I'll investigate)

If we could do a find using the `move_task_order_id` and `created_at` columns to filter results instead and then check that we couldn't find anything, that would be a good test to add. Then we would have been able to see this test fail due to transaction errors.

pkg/services/mto_service_item/mto_service_item_creator_test.go
```diff
@@ -42,55 +42,124 @@ func (t *testCreateMTOServiceItemQueryBuilder) Transaction(fn func(tx *pop.Conne
...
+}
+
+// Should return a message stating that service items can't be created if
+// the move is not in approved status.
+func (suite *MTOServiceItemServiceSuite) TestCreateMTOServiceItemWithInvalidMove() {
+	builder := query.NewQueryBuilder(suite.DB())
+	creator := NewMTOServiceItemCreator(builder)
+	serviceItemForUnapprovedMove := suite.buildValidServiceItemWithInvalidMove()
+
+	createdServiceItem, _, err := creator.CreateMTOServiceItem(&serviceItemForUnapprovedMove)
```
sandy-wright, line 129: We actually return an array of service items from this function, so it should be `createdServiceItems`

pkg/services/mto_service_item/mto_service_item_creator_test.go
```diff
@@ -42,55 +42,124 @@ func (t *testCreateMTOServiceItemQueryBuilder) Transaction(fn func(tx *pop.Conne
...
+
+func (suite *MTOServiceItemServiceSuite) TestCreateMTOServiceItem() {
+	serviceItem := suite.buildValidServiceItemWithValidMove()
+	moveTaskOrder := serviceItem.MoveTaskOrder
+	builder := query.NewQueryBuilder(suite.DB())
+	creator := NewMTOServiceItemCreator(builder)
+
+	// Happy path: If the service item is created successfully it should be returned
+	suite.T().Run("success", func(t *testing.T) {
 		createdServiceItem, verrs, err := creator.CreateMTOServiceItem(&serviceItem)
```
sandy-wright, line 150: Like I mentioned above (and I see this is not at all your fault, no worries 😄 ), this should probably be named `createdServiceItems` since it's an array

pkg/services/mto_service_item/mto_service_item_creator_test.go
```diff
@@ -271,28 +323,6 @@ func (suite *MTOServiceItemServiceSuite) TestCreateMTOServiceItem() {
...
-			fakeTransaction: fakeTx,
-		}
-		fakeCreateNewBuilder := func(db *pop.Connection) createMTOServiceItemQueryBuilder {
-			return builder
-		}
-		creator := mtoServiceItemCreator{
-			builder:          builder,
-			createNewBuilder: fakeCreateNewBuilder,
-		}
-
```
sandy-wright, line 295: Thank you for cleaning all these other tests up, this is awesome!

pkg/services/mto_service_item/mto_service_item_updater.go
```diff
@@ -115,8 +116,10 @@ func (p *mtoServiceItemUpdater) UpdateMTOServiceItemStatus(mtoServiceItemID uuid
...
 	}
-	// If we didn't set to MOVE APPROVED and we aren't already at APPROVALS REQUESTED we need to get there
-	if move.Status != models.MoveStatusAPPROVALSREQUESTED && move.Status != models.MoveStatusAPPROVED {
+	// If the move should not be APPROVED (due to one or more service items
+	// being in SUBMITTED status) then it needs to be set to APPROVALS REQUESTED
+	// so the TOO can review it.
+	if !moveShouldBeMoveApproved {
 		err = move.SetApprovalsRequested()
 		if err != nil {
 			return nil, err
```
sandy-wright, line 125: Hmm, I think we also need to set up a transaction in this service object - the service item will still be updated with a new status even if this update fails.

pkg/services/mto_service_item/mto_service_item_updater_test.go
```diff
@@ -266,3 +268,77 @@ func (suite *MTOServiceItemServiceSuite) TestValidateUpdateMTOServiceItem() {
...
+		suite.DB().Find(&serviceItem, serviceItem.ID)
+
+		suite.Equal(models.MoveStatusAPPROVED, move.Status)
+		suite.Equal(models.MTOServiceItemStatusRejected, serviceItem.Status)
+		suite.Equal(rejectionReason, serviceItem.RejectionReason)
+		suite.NotNil(serviceItem.RejectedAt)
+		suite.Nil(serviceItem.ApprovedAt)
+		suite.NotNil(updatedServiceItem)
+		suite.NoError(err)
+	})
```
sandy-wright, line 343: Thanks for adding these tests! Would it be possible to add a test to check that if for some reason the Move fails an update, then the service item wasn't updated either? That way we can catch some of these transaction issue. This will involve mocking the Move's update, I think.

sandy-wright (commented):
Hi Moncef! Thanks for doing all this work - especially with cleaning up so many of these tests. I left a bunch of comments and questions, please let me know if you'd like any more clarification. I think the only thing left is shoring up some of those transaction issues, but looks good otherwise 👍

pkg/services/mto_service_item/mto_service_item_creator_test.go
monfresh: Ah, great tip!

pkg/services/mto_service_item/mto_service_item_creator_test.go
monfresh: Yeah, good catch. That's what I had at first for various queries, using`.Where` and `.First`, and then I realized it could be simplified with `.Find`, but I got overzealous in my find and replace :) It's fixed now in my latest commit.

pkg/services/mto_service_item/mto_service_item_updater.go
```diff
@@ -45,6 +45,24 @@ func (p *mtoServiceItemUpdater) UpdateMTOServiceItemStatus(mtoServiceItemID uuid
 		return nil, services.NewNotFoundError(mtoServiceItemID, "MTOServiceItemID")
 	}
 
+	var move models.Move
+	moveFilter := []services.QueryFilter{
+		query.NewQueryFilter("id", "=", mtoServiceItem.MoveTaskOrderID),
+	}
+	err = p.builder.FetchOne(&move, moveFilter)
+	if err != nil {
+		return nil, services.NewNotFoundError(mtoServiceItemID, "MTOServiceItemID")
```
duncan-truss, line 54: Should we pass the MoveTaskOrderID here instead to not confuse it with the NotFoundError above?

pkg/services/mto_service_item/mto_service_item_updater.go
```diff
@@ -45,6 +45,24 @@ func (p *mtoServiceItemUpdater) UpdateMTOServiceItemStatus(mtoServiceItemID uuid
...
+	}
+
+	// Service item status can only be updated if a Move's status is either Approved
+	// or Approvals Requested, so check and fail early.
+	if move.Status != models.MoveStatusAPPROVED && move.Status != models.MoveStatusAPPROVALSREQUESTED {
```
duncan-truss, line 59: Good idea, it's more work but should catch some unintended behavior.

pkg/services/mto_service_item/mto_service_item_updater.go
```diff
@@ -81,19 +99,17 @@ func (p *mtoServiceItemUpdater) UpdateMTOServiceItemStatus(mtoServiceItemID uuid
...
 		return nil, services.NewNotFoundError(mtoServiceItemID, "MTOServiceItemID")
 	}
 
-	// If there are no service items that are SUBMITTED then we need to change the move status to MOVE APPROVED
+	// If there are no service items that are SUBMITTED then we need to change
+	// the move status to APPROVED
 	moveShouldBeMoveApproved := true
 	for _, mtoServiceItem := range move.MTOServiceItems {
-		if mtoServiceItem.Status == models.MTOServiceItemStatusSubmitted {
+		if mtoServiceItem.Status == models.MTOServiceItemStatusSubmitted || status == models.MTOServiceItemStatusSubmitted {
```
duncan-truss, line 112: I think we could check if `status == models.MTOServiceItemStatusSubmitted` above and if true we wouldn't have to refetch the move or loop through the statuses at all.

pkg/services/mto_service_item/mto_service_item_updater_test.go
```diff
@@ -266,3 +268,115 @@ func (suite *MTOServiceItemServiceSuite) TestValidateUpdateMTOServiceItem() {
...
+		suite.Equal(models.MTOServiceItemStatusApproved, serviceItem.Status)
+		suite.NotNil(updatedServiceItem)
+		suite.NoError(err)
+	})
+
+	// Test that the move's status changes to Approvals Requested if any of its service
+	// items' status is SUBMITTED
+	suite.T().Run("When move is approved and service item is submitted", func(t *testing.T) {
+		suite.DB().TruncateAll()
+		eTag, serviceItem, move := suite.createServiceItem()
```
duncan-truss, line 324: This creates a move in the Approvals Requested status to start with I think.  Don't we want it to be in the Approved status to test that it goes back to Approvals Requested?

pkg/services/mto_service_item/mto_service_item_updater_test.go
```diff
@@ -266,3 +268,115 @@ func (suite *MTOServiceItemServiceSuite) TestValidateUpdateMTOServiceItem() {
...
+func (suite *MTOServiceItemServiceSuite) TestUpdateMTOServiceItemStatus() {
+	builder := query.NewQueryBuilder(suite.DB())
+	updater := NewMTOServiceItemUpdater(builder)
+
+	rejectionReason := swag.String("")
+
+	// Test that the move's status changes to Approved when the service item's
+	// status is no longer SUBMITTED
+	suite.T().Run("When TOO reviews move and approves service item", func(t *testing.T) {
+		suite.DB().TruncateAll()
```
duncan-truss, line 305: Can you pull this out into a SetupTest and check the return value (that's a static analysis concern) like here https://github.com/transcom/mymove/blob/4db46f0c3e15f32b776e8fdef941f3c9f881d39b/pkg/services/office_user/customer/customer_service_test.go#L15

duncan-truss (approved):
This is great work Moncef, loving the tests. This is a hairy part of the codebase.  Sandy has a good point about the other transaction but I'll approve since that wasn't part of the original story.

pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -178,6 +187,23 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
...
+		// Once the service item is successfuly created, the Move's status needs to
+		// be updated to 'Approvals Requested' if it's not already in that state,
+		// which will let the TOO know they need to review it.
+		if move.Status != models.MoveStatusAPPROVALSREQUESTED {
+			// We're not checking for an error here because this will only error if
+			// the move's status is not 'Approved', and it's not possible for a Move
+			// that's not 'Approved' to get to this line of code because at the
+			// beginning of this function, we only allow moves with status 'Approved'
+			// or 'Approvals Requested'.
+			move.SetApprovalsRequested()
```
duncan-truss, line 199: This is going to get flagged in static analysis if we ignore the error return value.

monfresh: Ah, good call. I forgot about that.

pkg/services/mto_service_item/mto_service_item_updater.go
monfresh: Yeah, I noticed that it was like that before and wondered why. I think it makes sense to use the Move ID.

pkg/services/mto_service_item/mto_service_item_updater.go
monfresh: Yes, good call.

pkg/services/mto_service_item/mto_service_item_updater_test.go
monfresh: We already have a SetupTest available. I forgot I could just call that.

pkg/services/mto_service_item/mto_service_item_updater_test.go
monfresh: Yes, good catch. Before, I wasn't clearing the DB in between tests, and the move's status would have been Approved at that point, but after I added the truncation, I forgot to update the test.

pkg/services/mto_service_item/mto_service_item_updater_test.go
monfresh: That's a great idea, but I couldn't get it to work. Is there a way to mock the move updates without having to create a new interface?

pkg/services/mto_service_item/mto_service_item_creator_test.go
monfresh: I went ahead and fixed all singular instances to plural

pkg/services/mto_service_item/mto_service_item_updater.go
monfresh: In theory, this should never fail because I added a check at the top to make sure you can only update a service item if the move is in the `Approved` or `Approvals Requested` status.

duncan-truss (approved):

Ryan-Koch (approved):

PR #5599, opened 2021-01-06
Els update cac

tinyels:
## Description

Add Erin's new cac card info to client certs


reggieriser (changes requested):
Locally, this worked fine.  But I looked at S3 exp/stg and can see the migration exists, but I get a `fatal error: An error occurred (403) when calling the HeadObject operation: Forbidden` error when I try to download it.  I tried a previous migration of mine and it downloaded fine, so I wonder if there's some strange permission thing going on.

Also, I didn't see a `prd` no-op migration, so I think this would fail when deploying to prod unless you add one.

lynzt:
we saw another 403 error w/ a CAC, so I also wonder if there is an ongoing issue: https://ustcdp3.slack.com/archives/CSGDM3NUW/p1608753738018600

PR #5602, opened 2021-01-06
MB-6159 error out if no cypress results

gidjin:
## Description

Cypress tests always generate results, whether tests pass or fail. The script running the tests should fail if none are found to let us know something is wrong.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6159) for this change
* [this article](https://ustcdp3.slack.com/archives/CP496B8DB/p1609958558007600) explains more about the approach used.

### DIFF
--- a/scripts/run-e2e-test-docker
+++ b/scripts/run-e2e-test-docker
@@ -161,8 +161,9 @@ docker logs -f ${CONTAINER_CYPRESS}
 # Stop the app container to release the DB connection
 docker stop ${CONTAINER}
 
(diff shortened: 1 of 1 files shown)


tinyels (commented):
There are some changes to tests that I don't think are meant to be part of this.

gidjin:
> There are some changes to tests that I don't think are meant to be part of this.

You are fast! I just saw those myself, they are removed.

esacteksab (approved):

PR #5616, opened 2021-01-08
[MB-6179] Trim State and City columns on addresses table

harshamurthy:
## Description

The addresses table in deployed environments contains values like ` CA`, `NJ `, or ` NC` for states, and cities like ' FORT LEONARD WOOD'. This chore is to write a migration to run a left trim and right trim on all the rows in the address table to strip extra spaces from city and state columns. 

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6179) for this change

## Screenshots


migrations/app/schema/20210108174326_fix_city_and_state_data_in_addresses.up.sql
```diff
@@ -0,0 +1,3 @@
+SELECT TRIM(city) FROM addresses;
+SELECT TRIM(state) FROM addresses;
```
reggieriser, line 2: This is just selecting the trimmed values, not updating the values in the database.  I think you'll need to do something like:
`UPDATE addresses SET city = TRIM(city);`

You may want to just update the fields where a change is needed by adding a WHERE clause.  And you may want to update the `updated_at` timestamp on any rows you change.  Something like this:
`UPDATE addresses SET city = TRIM(city), updated_at = now() WHERE city <> TRIM(city);`

Double-check me on all this, of course.

reggieriser (changes requested):
The migration isn't making the updates.  Added some notes for things to try.

migrations/app/schema/20210108174326_fix_city_and_state_data_in_addresses.up.sql
```diff
@@ -0,0 +1,5 @@
+UPDATE addresses SET street_address_1 = TRIM(street_address_1), updated_at = now() WHERE street_address_1 <> TRIM(street_address_1);
+UPDATE addresses SET street_address_2 = TRIM(street_address_2), updated_at = now() WHERE street_address_2 <> TRIM(street_address_2);
+UPDATE addresses SET street_address_3 = TRIM(street_address_3), updated_at = now() WHERE street_address_3 <> TRIM(street_address_3);
```
reggieriser, line 3: Since `street_address_2` and `street_address_3` are nullable, let's change the `WHERE` clause to include a null check for those two columns to avoid any ambiguity with the comparison.  For example:
`WHERE street_address_2 IS NOT NULL and street_address_2 <> TRIM(street_address_2)`

migrations/app/schema/20210108174326_fix_city_and_state_data_in_addresses.up.sql
```diff
@@ -1,5 +1,5 @@
 UPDATE addresses SET street_address_1 = TRIM(street_address_1), updated_at = now() WHERE street_address_1 <> TRIM(street_address_1);
-UPDATE addresses SET street_address_2 = TRIM(street_address_2), updated_at = now() WHERE street_address_2 <> TRIM(street_address_2);
-UPDATE addresses SET street_address_3 = TRIM(street_address_3), updated_at = now() WHERE street_address_3 <> TRIM(street_address_3);
+UPDATE addresses SET street_address_2 = TRIM(street_address_2), updated_at = now() WHERE street_address_2 IS NOT NULL and street_address_2 <> TRIM(street_address_2);
+UPDATE addresses SET street_address_3 = TRIM(street_address_3), updated_at = now() WHERE street_address_3 IS NOT NULL and street_address_2 <> TRIM(street_address_3);
```
reggieriser, line 3: Oops -- `street_address_2` is a typo here and would lead to more rows being updated than we want.

reggieriser (changes requested):

reggieriser (approved):
LGTM!  Thanks for the changes.  Based on the address data loaded by the migrations, I see that 12 records were changed.  That seems correct based on some queries I ran before applying your migration to clean things up.

Given you're changing data, you should probably run this against experimental before merging and make sure it runs OK and doesn't take too much time.

PR #5557, opened 2020-12-29
[MB-4530] adding aditional UT to make sure payment request number is being test…

jacquelineIO:
…ed and returns an error if there is a failure to create the payment request number. Additional testing around the handler for payment request creator from the prime API call, also for good error messaging when payment request cannot be generated due to the MTO Reference ID missing.

## Description

UT to check error message when payment request fails to generated due to missing MTO Reference ID.

## Reviewer Notes

n/a

#### Run the local server 

```sh
make  db_dev_e2e_populate && make server_run
```

#### Remove ReferenceID in dev database

Open up your favorite db editor. And locate MTO with 

```
id='9c7b255c-2981-4bf8-839f-61c7458e2b4d'
```
and remove the `reference_id` value and update the database.

#### (optional) Run fetch-mto-updates

```sh
prime-api-client --insecure fetch-mto-updates | jq 'map(select(.id == "9c7b255c-2981-4bf8-839f-61c7458e2b4d")) | .[0]'
```

#### Create payload 
filename : payload.json

```json
{                                                                                                                           
  "body": {
    "isFinal": false,
    "moveTaskOrderID": "9c7b255c-2981-4bf8-839f-61c7458e2b4d",
    "serviceItems": [
      {   
        "id": "94bc8b44-fefe-469f-83a0-39b1e31116fb"
      },  
      {   
        "id": "eee4b555-2475-4e67-a5b8-102f28d950f8"
      },  
      {   
        "id": "6431e3e2-4ee4-41b5-b226-393f9133eb6c"
      },  
      {   
        "id": "fd6741a5-a92c-44d5-8303-1d7f5e60afbf"
      }   
    ]   
  }
}
```

#### Run CLI to create payment request

```sh
prime-api-client --insecure create-payment-request --filename payload.json | jq
``` 

#### Verify output

```json
{
  "Payload": {
    "detail": "issue creating payment request unique identifier: MTO 9c7b255c-2981-4bf8-839f-61c7458e2b4d has missing ReferenceID",
    "instance": "098d5095-ffae-4fe4-85e1-57e7142a633a",
    "title": "Validation Error",
    "invalidFields": null
  }
}
```

### Staging Setup

If you can find an MTO with a missing ReferenceID, then run the create payment request CLI in staging. Otherwise, just run a regular create payment request and verify that everything works without issue.

Look at demo wiki page for running CLI to create payment requests. https://github.com/transcom/mymove/wiki/Manually-run-Prime-API-for-Slice-demo

[...]

PR #5620, opened 2021-01-09
MB-6103 make reusable test utilities

gidjin:
## Description

This PR reviewed the work in #5557 and pulled out the big blocks of setup to be put into reusable functions. 

## Reviewer Notes

Does this look like a good way to make these reusable? Are there any other chucks of setup code worth extracting? Note the new `fixtureMapOfServiceItemParams` function that `MakeRealMTOServiceItemWithAllDeps` depends on does not have all possible service items. I created the 4 I know we can price currently as examples for us going forward.

## Setup

Changes to tests only so standard server test runs

```sh
go test -failfast ./pkg/services/payment_request ./pkg/handlers/primeapi
make server_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6103) for this change


pkg/testdatagen/make_pricing_data.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/gobuffalo/pop/v5"
+	//"github.com/gofrs/uuid"
+
+	"github.com/transcom/mymove/pkg/models"
+)
+
+// SetupServiceAreaRateArea sets up contract, service area, rate area, zip3
+// returns contractYear, serviceArea, rateArea, reZip3
+func SetupServiceAreaRateArea(db *pop.Connection, serviceAreaStr string, address models.Address) (models.ReContractYear, models.ReDomesticServiceArea, models.ReRateArea, models.ReZip3) {
```
reggieriser, line 15: There are some hard-coded constants in here.  Now that this is generically in testdatagen, is there a way to pass assertions to this function (like our other makers) and have those used but fall back to defaults?  For instance, the EscalationCompounded will factor into a pricing test but wouldn't I have to assume it's set to 1.04071 always (and if it changes, the test would break).

pkg/testdatagen/make_mto_service_item.go
```diff
@@ -75,3 +77,200 @@ func MakeMTOServiceItems(db *pop.Connection) models.MTOServiceItems {
...
+		// create a service item and return it
+		mtoServiceItem := MakeMTOServiceItem(db, Assertions{
+			Move:        mto,
+			MTOShipment: mtoShipment,
+		})
+
+		return mtoServiceItem
+	}
+
+	log.Panicf("couldn't create service item service code %s not defiled", serviceCode)
```
reggieriser, line 274: Nit: typo "defiled"?

pkg/testdatagen/make_payment_request.go
```diff
@@ -127,3 +127,45 @@ func MakeMultiPaymentRequestWithItems(db *pop.Connection, assertions Assertions,
 		MakePaymentRequestWithServiceItems(db, assertions)
 	}
 }
+
+// MakeFullDLHMTOServiceItem makes a DLH type service item along with all its expected parameters returns the created move and all service items
+func MakeFullDLHMTOServiceItem(db *pop.Connection, assertions Assertions) (models.Move, models.MTOServiceItems) {
+	mtoShipment := assertions.MTOShipment
+	if mtoShipment.ID == uuid.Nil {
+		mtoShipment = MakeMTOShipment(db, assertions)
```
reggieriser, line 135: I may be missing something, but won't you have to create the MTO first, then add the shipment with an assertion that links it to the MTO?  Otherwise I think this MakeMTOShipment call is going to create a new MTO that's not related in the DB to the one you're creating below.  I see that you're passing in `MTOShipments` into the assertion, but I don't think that's going to link those together in the DB like I assume you're wanting.

pkg/testdatagen/make_mto_service_item.go
```diff
@@ -75,3 +77,200 @@ func MakeMTOServiceItems(db *pop.Connection) models.MTOServiceItems {
 	serviceItemList = append(serviceItemList, MakeDefaultMTOServiceItem(db))
 	return serviceItemList
 }
+
+type realMTOServiceParamData struct {
+	ServiceCode          models.ReServiceCode
+	ServiceItemParamKeys []models.ServiceItemParamKey
+}
+
+func fixtureMapOfServiceItemParams() map[models.ReServiceCode]realMTOServiceParamData {
```
reggieriser, line 86: I saw your note about this creating just the four service items that we're using so far.  I'm wondering if it would make more sense for us to just have a SQL fixture to load all of these as part of the setup of the test suite or something easier to digest/update rather than having to set up this parallel map in code.  I think we perhaps do something like this with the pricing template parser?

reggieriser (commented):
That's a lot of setup!  I think you did a good job pulling it out, but I had some comments for consideration.  Perhaps we can take this as a first-cut abstraction and then make chores if anything else I mentioned seems warranted.

pkg/testdatagen/make_mto_service_item.go
gidjin: I was thinking about this and I agree it would be better given how much we have. I couldn't find a good example, though the pricing template doing so sounds familiar. I did see `pkg/testdatagen/make_invoice.go` This might be a good pattern for our testdatagen that needs to create large amounts of data.

pkg/testdatagen/make_mto_service_item.go
gidjin: If I had a nickel... Thanks for catching it

pkg/testdatagen/make_pricing_data.go
gidjin: Yes you are correct. I've adjusted the method to accept overrides for everything via testdatagen.Assertions.

pkg/testdatagen/make_payment_request.go
gidjin: I think you are correct, I'll rearrange it

gidjin: I followed the pattern from `pkg/testdatagen/make_move.go` so I think it should be working as expected now. Let me know what you think.

pkg/testdatagen/make_payment_request.go
```diff
@@ -130,27 +130,33 @@ func MakeMultiPaymentRequestWithItems(db *pop.Connection, assertions Assertions,
...
+		mergeModels(&mtoShipment, assertions.MTOShipment)
+
+		mtoShipment = MakeMTOShipment(db, Assertions{
+			Move:        moveTaskOrder,
+			MTOShipment: mtoShipment,
+		})
+
+		moveTaskOrder.MTOShipments = models.MTOShipments{mtoShipment}
+	} else {
+		mtoShipment = moveTaskOrder.MTOShipments[0]
```
reggieriser, line 159: In this branch of the `if` we have a `moveTaskOrder` passed in via the assertion, but how do we know that MTO has a shipment associated with it already?  And if someone passes in a shipment via the assertion, how is it used?

pkg/testdatagen/make_pricing_data.go
```diff
@@ -0,0 +1,82 @@
...
+		ContractID: contractYear.Contract.ID,
+		IsOconus:   false,
+		Code:       "US47",
+		Name:       address.State,
+		Contract:   contractYear.Contract,
+	}
+
+	mergeModels(&rateArea, assertions.ReRateArea)
+
+	mustSave(db, &rateArea)
```
reggieriser, line 56: Wondering if we should add a "maker" for `ReRateArea` to match the other models used here in the Assertion struct.

pkg/testdatagen/make_pricing_data.go
```diff
@@ -0,0 +1,82 @@
...
+		Assertions{
+			ReDomesticServiceArea: serviceArea,
+		})
+
+	address := models.Address{
+		City:       "Birmingham",
+		State:      "AL",
+		PostalCode: "35203",
+	}
+	mergeModels(&address, assertions.Address)
```
reggieriser, line 44: Where is the address record getting saved to the DB?  Or is that irrelevant?

reggieriser (commented):
Re-reviewed -- thanks for making the changes!  Tests ran fine.  I did have a few questions/comments I'd like to get your feedback on before final approval.

pkg/testdatagen/make_pricing_data.go
gidjin: I figured it was irrelevant since all I needed were the details for the ReZip3. But Looking at it again I can remove address all together and use `assertions.ReZip3` instead, which makes more sense to me

pkg/testdatagen/make_pricing_data.go
gidjin: Probably, should be short to add I'll make one.

pkg/testdatagen/make_payment_request.go
gidjin: This line would error out if there were no shipments on the move, with an index out of range, as I understand. The shipment is needed so that we know which shipment to tie the MTOServiceItem to, which is done in `MakeRealMTOServiceItemWithAllDeps`.  If it is unclear I can take it out since for now we don't need it.

reggieriser (approved):
LGTM! 🚀

PR #5627, opened 2021-01-11
[MB-3304] F/E dependencies analysis

suzubara:
## Description

This PR sets up a new Github workflow to use the [webpack-stats-diff-action](https://github.com/chronotruck/webpack-stats-diff-action). On PRs to master that make changes in the F/E source code, the workflow will build production assets for both HEAD and the base ref, and post a comment to the PR comparing the difference in size.

I did some experimenting on this PR to test out the resulting changes but for the most part they were too small to cause a substantial diff. I would like to merge this new workflow to at least give us a baseline to measure against and maybe refine it with more detailed information (like individual chunk sizes) later on.

I also removed the following dependencies, which as far as I can tell are no longer being used in any capacity:
- @18f/us-federal-holidays
- js-yaml
- jwt-decode
- react-form
- react-numeric-input
- selenium-webdriver

Note that removing the above packages would not have an expected diff in the bundle size since they were not being used by the code.

Finally, I removed some of the `resolutions` in the package.json file that are no longer needed (they are resolving to the correct versions now due to dependency updates).

## Reviewer Notes

No code/application changes are in this PR so tests passing should be sufficient. You can run local scripts to test that everything works as expected.

## Setup

```sh
yarn install
yarn test
yarn storybook
yarn audit

make client_run
make client_build
```

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3304) for this change


suzubara:
Requesting an @transcom/truss-infrasec review on this since I am adding a Github action: https://github.com/chronotruck/webpack-stats-diff-action

esacteksab (approved):

monfresh:
Is the idea here to keep an eye on the size of our assets? Is there a specific number we don't want to go over? If so, what happens if we do? How do we enforce the maximum size?

monfresh:
In other words, what problem does this GitHub Action solve?

suzubara:
@monfresh Yes, the idea here is to start measuring asset size. I don't think we have a specific benchmark in mind, or want to start enforcing one at this point (the number here is also misleading because it is size of total assets, which are split up into chunks).

My goal here was to add an easy indicator that can be used when someone is reviewing a PR, if for example some changes result in an unexpectedly large increase in size, that might be a red flag. I'd also like to start measuring a baseline so that we can look for changes as we add code & refactor. AFAIK we are not tracking this anywhere yet, so this was a low lift way to start doing so.

jenniferlynparsons (approved):
I admit I don't know enough about what analyze-bundle.yml is doing to give a thorough review of that, but things are working as expected locally.

monfresh (approved):
LGTM. All the commands worked for me locally.

monfresh:
One thing I noticed is that my `yarn.lock` changed locally. Maybe it needs to be updated in this PR?

suzubara:
@monfresh good catch! I updated the lockfile & also resolved conflicts with master

PR #5644, opened 2021-01-12
[MB-1585] SIT Schedule lookup for dest and origin

pearl-truss:
## Description

This PR creates a lookup function that is used for both SITScheduleOrigin and SITScheduleDest

## Reviewer Notes

- I believe the sources for the pickup and destination addresses will need to be updated once those are no longer found on the MTO Service item (tickets to update this: https://dp3.atlassian.net/browse/MB-6279 and https://dp3.atlassian.net/browse/MB-6257))
- I broke out the query to find the domestic service area into a new file called `lookup_query_helper.go`. I based this off of the existing `pricer_query_helper.go` file. 
- There isn't any code set up (that I'm aware of for this ticket) but please double check that the query and unit test make sense

## Acceptance
- As noted in sprint planning, there's no acceptance for this ticket. Instead the lookup work can be accepted when the pricer is built

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-1585) for this change

### DIFF
--- a/pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go
+++ b/pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go
@@ -289,7 +289,6 @@ func ServiceParamLookupInitialize(
 
 	paramKey = models.ServiceItemParamNameSITScheduleOrigin
 	err = s.setLookup(serviceItemCode, paramKey, SITScheduleLookup{
-		// Will need to be updated after https://dp3.atlassian.net/browse/MB-6257#icft=MB-6257
 		Address: pickupAddress,
 	})
 	if err != nil {
@@ -298,7 +297,6 @@ func ServiceParamLookupInitialize(
 
 	paramKey = models.ServiceItemParamNameSITScheduleDest
 	err = s.setLookup(serviceItemCode, paramKey, SITScheduleLookup{
-		// Will need to be updated after https://dp3.atlassian.net/browse/MB-6257#icft=MB-6257
 		Address: destinationAddress,
 	})
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go
```diff
@@ -287,6 +287,24 @@ func ServiceParamLookupInitialize(
 		return nil, err
 	}
 
+	paramKey = models.ServiceItemParamNameSITScheduleOrigin
+	err = s.setLookup(serviceItemCode, paramKey, SITScheduleLookup{
+		// Will need to be updated after https://dp3.atlassian.net/browse/MB-6257#icft=MB-6257
```
jacquelineIO, line 292: According to our inputs for calculating pricing sheet https://docs.google.com/spreadsheets/d/1NRbxHmvaWV6aXQrxQ2LJhkc5tClAl3Eb1-MRxZ123Tw/edit#gid=0 
There is no change needed so what you have is good.

pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go
```diff
@@ -287,6 +287,24 @@ func ServiceParamLookupInitialize(
...
+		Address: pickupAddress,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	paramKey = models.ServiceItemParamNameSITScheduleDest
+	err = s.setLookup(serviceItemCode, paramKey, SITScheduleLookup{
+		// Will need to be updated after https://dp3.atlassian.net/browse/MB-6257#icft=MB-6257
```
jacquelineIO, line 301: Same as previous comment

jacquelineIO (approved):
lgtm
nice job adding the `fetchDomesticServiceArea` function to streamline some code.

pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go
pearl-truss: Okay cool, I've removed the comments

PR #5645, opened 2021-01-12
[MB-6255] SE Value is incorrect

harshamurthy:
## Description

The `SE` segment shows the sum of the number of header segments plus the number of service item segments plus the `L3` segment, `ST` segment, and `SE` segment.

Each service item now has seven segments, but the constant we were using to multiply was `6` causing the final SE value to be incorrect.


pearl-truss (approved):
Changes look good in the EDI I created!

PR #5567, opened 2020-12-30
MB-6045: create a low privileged postgres role for crud operations

carterjones:
## Description

This creates a new PostgreSQL role for CRUD operations (Create, Read, Update, Delete). Technically in PostgreSQL, the operations are Insert, Select, Update, Delete (ISUD), but we think CRUD is more widely known to represent these basic operations.

This role will initially be used by the main app container to reduce the privileges available to the app. However, it can be used by any other containers that also can run with fewer permissions.

## Reviewer Notes

I intentionally did not include `FOR ROLE` in the `ALTER DEFAULT PRIVILEGES` statements. In this way, the role that is running the migrations (currently `ecs_user` in RDS and `postgres` in Docker) will have its own role modified. This is a break from the previous pattern where `FOR ROLE` was used. If I should use `FOR ROLE`, I could add something like the following instead:

```sql
-- The postgres role is used by databases running in Docker containers.
DO
$do$
BEGIN
  IF NOT EXISTS (
    SELECT -- SELECT list can stay empty for this
    FROM   pg_catalog.pg_roles
    WHERE  rolname = 'postgres') THEN
    ALTER DEFAULT PRIVILEGES FOR ROLE postgres GRANT SELECT, INSERT, UPDATE, DELETE ON TABLES TO crud;
    ALTER DEFAULT PRIVILEGES FOR ROLE postgres GRANT USAGE, UPDATE ON SEQUENCES TO crud;
  END IF;
END
$do$;

-- As of 2020-12-30, we used ecs_user to perform migrations in RDS.
DO
$do$
BEGIN
  IF NOT EXISTS (
    SELECT -- SELECT list can stay empty for this
    FROM   pg_catalog.pg_roles
    WHERE  rolname = 'ecs_user') THEN
    ALTER DEFAULT PRIVILEGES FOR ROLE ecs_user GRANT SELECT, INSERT, UPDATE, DELETE ON TABLES TO crud;
    ALTER DEFAULT PRIVILEGES FOR ROLE ecs_user GRANT USAGE, UPDATE ON SEQUENCES TO crud;
  END IF;
END
$do$;
```

I am initially proposing the more simple approach, but I'm happy to use another approach if others prefer something else.

One other thing to note is that the UPDATE permission on SEQUENCES is only needed for a test. Without it, we get this error (even though other tests appear to call `setval`):

```text
--- FAIL: TestGHCInvoiceSuite (1.22s)
[...]

PR #5655, opened 2021-01-13
MB-6305: use the limited postgresql role in unit tests

carterjones:
## Description

With the introduction of a [new low privileged database role](https://github.com/transcom/mymove/pull/5567), it would be useful to run unit tests with this new role. That's what this PR does:

1. Add the ability to utilize a low-privileged database role during testing. Make this the default. Make it so that the role with elevated privileges can also be used if desired, such as when testing migrations.
1. Start using the low privileged user throughout our various docker compose files and in CircleCI.
1. Remove testing of the `orders` API since it started throwing errors. We don't develop this portion of the codebase any more, so even if we did find bugs in it, we would devote any more time to developing it. Therefore, running tests against it (and in this PR's case, fixing code to make those errors go away) is not a good use of time and effort.

## Setup

```sh
docker kill $(docker ps --quiet --filter ancestor=postgres:12.4)
docker rm $(docker ps --all --quiet --filter ancestor=postgres:12.4)
make server_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6305) for this change
* [this article](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) explains more about the approach used for the functional options I added (`PopTestSuiteOption`).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -409,7 +409,9 @@ commands:
           environment:
             APPLICATION: '<< parameters.application >>'
             DB_PASSWORD: mysecretpassword
+            DB_PASSWORD_LOW_PRIV: mysecretpassword
             DB_USER: postgres
+            DB_USER_LOW_PRIV: crud
             DB_HOST: localhost
             DB_PORT_TEST: 5433
             DB_PORT: 5432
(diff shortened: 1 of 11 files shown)


esacteksab:
The not Go stuff seems straight forward enough, but I don't feel confident enough to bless the Go changes. I'll yield to folks who know better.

monfresh (approved):
I think this makes sense, and I love that you were able to simplify it without affecting the tests or requiring us to change the way we write new tests.

PR #5657, opened 2021-01-13
Yan MB-6163 Dynamic Weight Allowance On the Move 101

YanZ777:
## Description

Added the DBAuthorizedWeight to the payload so that it could be read into MovingInfo. 
Prevented the Alert from displaying if there is no authorized Weight.

## Reviewer Notes

Since there is old data that doesn't have the DBAuthorizedWeight on the entitlements, the decision was made to not show the alert on the Move 101 page.
https://ustcdp3.slack.com/archives/CTNTFJSBA/p1610482483495400

There's existing header violations inside of aXe that I am not addressing in this PR.

## Setup

```sh
make client_run
make server_run
```
Create a new service member and go through the workflow until you hit the Move 101 page.
Also refresh the Move 101 page. 
Login as sm_no_move_type@example.com and check the Move 101 page.
Make sure the weight allowance matches the rank inside of entitlements.go

(https://github.com/transcom/mymove/wiki/migrate-the-database#secure-migrations)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6163) for this change


src/pages/MyMove/MovingInfo.jsx
```diff
@@ -1,100 +1,152 @@
-import React from 'react';
+import React, { Component } from 'react';
+import { bindActionCreators } from 'redux';
```
suzubara, line 2: We don't need to use `bindActionCreators`, you can assign `mapDispatchToProps` to be an object instead of a function ie:
```
const mapDispatchToProps = {
  fetchLatestOrders: fetchLatestOrdersAction
}
```

src/pages/MyMove/MovingInfo.jsx
```diff
@@ -1,100 +1,152 @@
...
+};
+
+MovingInfo.defaultProps = {
+  entitlementWeight: 0,
 };
 
-export default MovingInfo;
+function mapStateToProps(state, ownProps) {
+  const { moveId } = ownProps.match.params;
+  const orders = selectOrdersForMove(state, moveId);
```
suzubara, line 132: Do you need to use the `moveId` to select the orders? I think you should be able to use `selectCurrentOrders` to get the orders for the logged in service member.. then you wouldn't need the `moveId` and don't have to add `withRouter`

swagger/internal.yaml
```diff
@@ -2190,6 +2190,10 @@ definitions:
         x-nullable: true
       department_indicator:
         $ref: '#/definitions/DeptIndicator'
+      dBAuthorizedWeight:
```
suzubara, line 2193: the naming of this field (prefix with `dB`) is strange to me, is there a reason to not call it `authorizedWeight`?

pkg/testdatagen/make_entitlement.go
```diff
@@ -26,6 +26,8 @@ func MakeEntitlement(db *pop.Connection, assertions Assertions) models.Entitleme
 		StorageInTransit:      &storageInTransit,
 	}
 	entitlement.SetWeightAllotment(*grade)
+	dBAuthorizedWeight := entitlement.AuthorizedWeight()
+	entitlement.DBAuthorizedWeight = dBAuthorizedWeight
```
YanZ777, line 30: @donaldthai @kimallen 
I noticed that these lines were causing some CircleCI tests to fail. I've tested both with and without these lines of code, and the Move 101 page still gets the weights dynamically. If that is the case, are these lines ok to take out?

swagger/internal.yaml
YanZ777: I may be remembering incorrectly, but I think I was told that the internal.yaml has to match up with something else.
Maybe @kimallen can weigh in on this more?

pkg/handlers/internalapi/orders.go
```diff
@@ -29,6 +29,14 @@ func payloadForOrdersModel(storer storage.FileStorer, order models.Order) (*inte
 		moves = append(moves, payload)
 	}
 
+	var dBAuthorizedWeight *int64
+	dBAuthorizedWeight = nil
+	if order.Entitlement != nil {
+		if order.Entitlement.DBAuthorizedWeight != nil {
+			dBAuthorizedWeight = swag.Int64(int64(*order.Entitlement.DBAuthorizedWeight))
+		}
+	}
```
donaldthai, line 38: Hey, the authorized weight can be simplified by calling the `AuthorizedWeight()` function: https://github.com/transcom/mymove/blob/master/pkg/models/ghc_entitlements.go#L41

swagger/internal.yaml
donaldthai: There is a property named `DBAuthorizedWeight` that refers to the `authorized_weight` column in the `ghc_entitlements.go` file but I do think it would make more sense to rename to `authorizedWeight` since `DBAuthorizedWeight` is meant to be used to determine if the column value was set in the DB.

donaldthai: `AuthorizedWeight()` in `ghc_entitlements.go` determines the value as well

kimallen: What I remember is that when we tried to use just `authorizedWeight` in the yaml, then it didn't match the existing field on Entitlement called `DBAuthorizedWeight`, so it was erroring out.  Maybe there's another way to rename than what we were trying.

src/pages/MyMove/MovingInfo.test.jsx
```diff
@@ -3,12 +3,20 @@ import { mount } from 'enzyme';
...
+const wrapper = mount(<MovingInfo {...testProps} />);
 
 describe('MovingInfo component', () => {
   it('renders', () => {
     expect(wrapper.find('MovingInfo').length).toBe(1);
     expect(wrapper.text()).toContain('Tips for planning your shipments');
+    expect(wrapper.text()).toContain('7,000 lbs');
     expect(wrapper.find('[data-testid="shipmentsHeader"]').length).toBe(1);
     expect(wrapper.find('[data-testid="shipmentsSubHeader"]').length).toBe(4);
   });
```
suzubara, line 22: It would be good to add a test case if `entitlementWeight` is `0` or even null/undefined to make sure this page doesn't crash

suzubara (approved):
left a suggestion for adding another test case, but otherwise looks good!

suzubara (approved):
F/E changes look good to me

PR #5689, opened 2021-01-19
MB-6186-create-index-CustomerUsers-handler-AdminAPI

NamibiaLT:
## Description
Add IndexUsers Handler to a users.go in the AdminAPI

## Reviewer Notes

Since a new endpoint was added and a new definition /customer_users was created, make sure that all appropriate files have been updated to reflect the addition

## Setup
To check whether the endpoint is hitting AdminAPI and retrieving a list of customers, use Postman to check the GET for `/users`. To see whether a sort can be done by email, make sure to specify the loginGovEmail: `sort=login_gov_email`
To run tests: 

```sh
go test ./pkg/handlers/adminapi --testify.m TestIndexUsersHandler -v
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6186) for this change
* [Jira story](https://dp3.atlassian.net/browse/MB-6227) for this change


pkg/handlers/adminapi/customer_users.go
```diff
@@ -0,0 +1,65 @@
...
+	services.NewPagination
+}
+
+func payloadForCustomerUserModel(o models.User) *adminmessages.CustomerUser {
+	payload := &adminmessages.CustomerUser{
+		ID:        handlers.FmtUUID(o.ID),
+		Email:     handlers.FmtString(o.LoginGovEmail),
+		Active:    handlers.FmtBool(o.Active),
+		CreatedAt: handlers.FmtDateTime(o.CreatedAt),
+	}
```
abbyoung, line 30: When I hit the endpoint, I also get `updatedAt` and `userId` in the payload, both of which are null. Should these values be returned? Also, what's the difference between `userId` and `ID`?

swagger/admin.yaml
```diff
@@ -136,6 +136,43 @@ definitions:
...
+        example: c56a4180-65aa-42ec-a945-5fd21dec0538
+      firstName:
+        type: string
+      lastName:
+        type: string
+      email:
+        type: string
+        format: x-email
+        pattern: '^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$'
+      userId:
```
abbyoung, line 154: Follow-up to above comment, what's userId represent vs id?

swagger/admin.yaml
```diff
@@ -862,6 +899,48 @@ paths:
...
         500:
           description: server error
+  /customer_users:
+    get:
+      summary: List customers
+      description: Returns a list of customers
+      operationId: indexCustomerUsers
+      tags:
+        - customer_users
+      parameters:
```
abbyoung, line 909: I was able to use the `page` and `perPage` query params successfully! And `sort` and `order` on the id. Some questions:

* Should you be able to sort on any field? (I wasn't able to sort on email.)
* Should we add some comments to the yaml to instruct on how to use the query params? e.g. you have to use `page` if you want to use `perPage`, which properties you can sort/filter by, etc. I'm not really clear on what I can and cannot do.

NamibiaLT: I'm trying to fix the sort by email issue. Ideally that's what we would want to do.

NamibiaLT: I had to be specific in the yaml that I am using the login_gov_email. That seemed to work.

NamibiaLT: That also means when sorting that the url has to say sort=loginGovEmail... I can write another ticket to return a message to the user to make sure the sort criteria for email is specifically loginGovEmail.

pkg/handlers/adminapi/users.go
```diff
@@ -45,6 +47,47 @@ func (h GetUserHandler) Handle(params userop.GetUserParams) middleware.Responder
 	return userop.NewGetUserOK().WithPayload(payload)
 }
 
+// IndexUsersHandler returns an user via GET /users/{userID}
```
abbyoung, line 50: Maybe update this comment? It sounds like this will return just one user, when it's returning the list of users

pkg/handlers/adminapi/users.go
```diff
@@ -45,6 +47,47 @@ func (h GetUserHandler) Handle(params userop.GetUserParams) middleware.Responder
...
+
+	queriedUsersCount := len(users)
+
+	payload := make(adminmessages.Users, queriedUsersCount)
+
+	for i, s := range users {
+		payload[i] = payloadForUserModel(s)
+	}
+
+	return userop.NewIndexUsersOK().WithContentRange(fmt.Sprintf("office users %d-%d/%d", pagination.Offset(), pagination.Offset()+queriedUsersCount, totalUsersCount)).WithPayload(payload)
```
abbyoung, line 88: Change "office users" to just "users"?

pkg/handlers/adminapi/users.go
NamibiaLT: Thank you! I updated it.

pkg/handlers/adminapi/users.go
NamibiaLT: Thanks! Updated!

abbyoung (approved):
lgtm! sorting works, pagination works 🥇 🎊

PR #5700, opened 2021-01-20
MB-6300 TOO move details green checkmark

Ryan-Koch:
## Description

It makes the checkmark for the service items approved when a shipment and move is approved on the Move Details page green.

## Setup

Build the office app, run it, log in as a TOO. Approve a move or view one that is already approved. You should see a green check-mark next to the basic service items approved on that page.

## Screenshots

image


suzubara (commented):
the fix looks good, my only comment is can we use the existing color variable for success/error (see `$success` and `$error` at https://storybook.dp3.us/?path=/story/global-colors--alert-colors) instead of raw hex values? cc @amanda5iu

amanda5iu (approved):
+1 to Suz's comment. Other than that, this looks good!

Ryan-Koch:
> the fix looks good, my only comment is can we use the existing color variable for success/error (see `$success` and `$error` at https://storybook.dp3.us/?path=/story/global-colors--alert-colors) instead of raw hex values? cc @amanda5iu

Absolutely! Pushed the change switching it to use variables @suzubara.

noelledusahel (approved):
LGTM

PR #5707, opened 2021-01-21
Yan MB-6376 Make Milmove Header Storybook Dynamic

YanZ777:
## Description

Edit the MilMove Header in Storybook to be Dynamic. 
Changes made:
- edit MilMoveHeader take in the links as children
- change MilMoveHeader to take in the customer as a prop to display the first and last name
- modify the sign out link to be a button instead and take in an on click handler as a prop
- change the MilMove logo to navigate back to the office
- update the storybook to use the new features and incorporate memory router for the NavLink used by the MilMove logo

## Reviewer Notes

Please verify that it is ok Memory Router the way I am using it.
Check my css please! 

## Setup

```sh
make storybook
```
Under Components, Headers, MilMove Header, Milmove

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6376) for this change

## Screenshots
image

### DIFF
--- a/src/components/MilMoveHeader/index.jsx
+++ b/src/components/MilMoveHeader/index.jsx
@@ -1,32 +1,35 @@
 import React from 'react';
-import classNames from 'classnames/bind';
+import { node, func } from 'prop-types';
+import { Button } from '@trussworks/react-uswds';
 
 import { ReactComponent as MmLogo } from '../../shared/images/milmove-logo.svg';
 
 import styles from './index.module.scss';
 
-const cx = classNames.bind(styles);
+import { CustomerShape } from 'types/moveOrder';
 
-const MilMoveHeader = () => (
-  <div className={cx('mm-header')}>
+const MilMoveHeader = ({ children, customer, handleLogout }) => (
+  <div className={styles.mmHeader}>
     <MmLogo />
-    <div className={cx('links')}>
+    <div className={styles.links}>
+      {children}
+      <span className={styles.lineAdd}>&nbsp;</span>
       <span>
-        <a href="#">Navigation Link</a>
+        {customer.last_name}, {customer.first_name}
       </span>
       <span>
-        <a href="#">Navigation Link</a>
-      </span>
-      <span>
-        <a href="#">Navigation Link</a>
-      </span>
-      <span className={cx('line-add')}>&nbsp;</span>
-      <span>Baker, Riley</span>
-      <span>
-        <a href="#">Sign out</a>
+        <Button className={styles.signOut} disabled={false} onClick={handleLogout}>
+          <span>Sign out</span>
+        </Button>
       </span>
     </div>
   </div>
 );
+
+MilMoveHeader.propTypes = {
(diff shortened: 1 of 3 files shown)


src/components/MilMoveHeader/index.jsx
```diff
@@ -1,32 +1,42 @@
...
-        <a href="#">Navigation Link</a>
-      </span>
-      <span>
-        <a href="#">Navigation Link</a>
-      </span>
+const MilMoveHeader = ({ children, customer, handleLogout }) => (
+  <div className={styles.mmHeader}>
+    <div className="usa-logo" id="basic-logo">
+      <em className="usa-logo__text">
+        <NavLink to="/" title="office.move.mil" aria-label="office.move.mil">
```
suzubara, line 16: `NavLink` is used instead of `Link` in order to provide an "active" class when the user is on the matching page.. I don't think that is necessary for this logo link since it won't be styled differently when it's "active". Also I think it's generally accepted to use a regular `<a>` tag for logo links, which would also mean we don't need to use the mock router in stories & test files.

src/components/MilMoveHeader/index.jsx
```diff
@@ -1,32 +1,42 @@
...
+      <span className={styles.lineAdd}>&nbsp;</span>
       <span>
-        <a href="#">Navigation Link</a>
+        {customer.last_name}, {customer.first_name}
       </span>
-      <span className={cx('line-add')}>&nbsp;</span>
-      <span>Baker, Riley</span>
       <span>
-        <a href="#">Sign out</a>
+        <Button className={styles.signOut} disabled={false} onClick={handleLogout}>
```
suzubara, line 28: You can add the `unstyled` prop to this Button component in order to remove the default styling, I think that would achieve some of the CSS resets you had to add! I also don't think there's any reason to add the `disabled={false}` prop (that's the default).

suzubara: This Button should also have a `type="button"` prop

src/components/MilMoveHeader/index.jsx
```diff
@@ -1,32 +1,42 @@
...
+          <span>Sign out</span>
```
suzubara, line 29: I think we can remove this `<span>` tag now that this is a Button

src/components/MilMoveHeader/index.jsx
```diff
@@ -1,32 +1,42 @@
...
-    <div className={cx('links')}>
-      <span>
-        <a href="#">Navigation Link</a>
-      </span>
-      <span>
-        <a href="#">Navigation Link</a>
-      </span>
+const MilMoveHeader = ({ children, customer, handleLogout }) => (
+  <div className={styles.mmHeader}>
+    <div className="usa-logo" id="basic-logo">
```
suzubara, line 14: this could use the `<Title>` component from ReactUSWDS (which adds the same markup you've added)

suzubara (approved):

brooksr4 (approved):

PR #5709, opened 2021-01-21
[MB-6279] Prime API Create SIT Origin Updated to add HHG Actual Address

jacquelineIO:
## Description

This PR is to update the Create SIT Origin Flow so that when the Prime requests a SIT Origin service item, that the prime provides and address (HHG Actual Origin). This represents when the customer has a change in origin and needs to use SIT at origin. The new origin (address) comes in on the SIT @ Origin request to the TOO and the following needs to happen:

   *  Prime sends in new origin address (this is not a SIT facility, but we aren't tracking what exactly this address is)
   *  MilMove will need to take that new origin address and update the MTO shipment pickup address with the new origin address from the SIT Origin request.
   * The original pickup address from the MTO shipment and the new pickup address (actual origin address) both need to be saved on the MTO service for SIT origin. (Since linehaul has not started yet, we can't assume that SIT origin is final and that it won't change again. We have to be able to compute the distance between original origin pickup and actual origin pick for the SIT origin pickup service item)

## Reviewer Notes

The `SITOriginHHGOriginalAddress` and `SITOriginHHGActualAddress` values are **needed** on the `DOPSIT` (Domestic SIT Origin Pickup) but are also stored on the `DOFSIT` (Domestic SIT Origin First Day) service item, because the `DOFSIT` service item is used to request origin SIT and is the source of gathering the `SITOriginHHGActualAddress` from the SIT @ origin request from the Prime.

## Setup DEV local

- First, set up the baseline data: `make db_dev_e2e_populate`
- Start up the server and the client: `make server_run` and `make office_client_run`
- Log into the office client using the local login for a TOO user. (I've been using too_tio_role@office.mil.)
- Go to the TOO interface and approve one of the moves.  Make sure the orders number, TAC, and SAC are set in the orders too.  Make note of the move code so you can get the MTO ID.
- After you approve the MTO, look at it to grab the `MTO REFERENCE ID ####-####`.
- Use the Reference ID to query MTO.
```sh
prime-api-client --insecure  fetch-mto-updates | jq 'map(select(.referenceId == "1188-0410")) | .[0]'
```
- Create a `create_dofsit.json` file that looks like this:
```json  
{
  "body": {
    "moveTaskOrderID": "<insert MTO ID>",
    "mtoShipmentID": "<insert shipment ID>",
    "modelType": "MTOServiceItemOriginSIT",
    "reServiceCode": "DOFSIT",
    "sitPostalCode": "00000",
    "reason": "my address changed",
    "sitEntryDate": "2021-01-13",
    "sitHHGActualOrigin":  {
      "streetAddress1": "20245 4th St",
      "city": "Albuquerque",
      "state": "NM",
      "postalCode": "87117",
      "country": "United States"
    }
  }
}
```

- Send this payload via the prime-api-client: `prime-api-client create-mto-service-item --insecure --filename create_dofsit.json`
- You should see 3 SIT Origin service items for that move now in the TOO interface under the "Move task order" tab.  Approve it.
- If you look at the database you can also see the MTO service items have been created with address IDs for the HHG original/actual address fields. The reason field also gives a clue to which service items are the ones newly created.

## Acceptance

Very similar to developer steps.
Either create a new move through onboarding or find an existing one in staging that can be used. (Can't already have a DOFSIT attached or it will be rejected.)

Using a similar payload to the above create the new SIT origin service items.

View the new service items in the TOO app and verify the current address was updated to the address provided in the DOFSIT payload.

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6279) for this change

## Screenshots

n/a


pkg/models/address.go
```diff
@@ -168,3 +168,12 @@ func (a *Address) CountryCode() (*string, error) {
 	}
 	return nil, nil
 }
+
+// Copy returns a pointer that is a copy of the original pointer Address
+func (a *Address) Copy() *Address {
```
gidjin, line 173: Nice

pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -141,12 +140,61 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
...
+			if serviceItem.SITOriginHHGActualAddress == nil {
+				verrs = validate.NewErrors()
+				verrs.Add("reServiceCode", fmt.Sprintf("%s cannot be created", serviceItem.ReService.Code))
+				return nil, nil, services.NewInvalidInputError(serviceItem.ID, nil, verrs,
+					fmt.Sprintf("A service item with reServiceCode %s must have the sitHHGActualOrigin field set.", serviceItem.ReService.Code))
+			}
+
+			// new SIT Origin HHG Actual Address coming from the Prime API call
+			serviceItem.SITOriginHHGActualAddress.ID = uuid.Nil
+			serviceItem.SITOriginHHGActualAddressID = nil
```
gidjin, line 162: Will we never be given a service item that already has a SITOriginHHGActualAddress stored in the database? Basically do we use this service from other places than the handler?

pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -141,12 +140,61 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
...
+			originalHhgAddress := mtoShipment.PickupAddress.Copy()
+			originalHhgAddress.ID = uuid.Nil
+
+			// save the HHG (new) actual pickup address, coming from the SIT service item
+			actualHhgAddress := serviceItem.SITOriginHHGActualAddress.Copy()
+			actualHhgAddress.ID = uuid.Nil
+
+			// update the SIT service item to track/save the HHG original pickup address (that came from the
+			// MTO shipment
+			serviceItem.SITOriginHHGOriginalAddress = originalHhgAddress.Copy()
```
gidjin, line 174: Do you need to call Copy here, isn't originalHhgAddress already a Copy from above?

pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -141,12 +140,61 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
...
+			serviceItem.SITOriginHHGOriginalAddressID = nil
+
+			// update the MTO shipment the new (actual) pickup address
+			mtoShipment.PickupAddress = actualHhgAddress.Copy()
```
gidjin, line 178: Do you need to call Copy here, isn't actualHhgAddress already a Copy from above?

pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -141,12 +140,61 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(serviceItem *models.MTOServ
...
+			// changes were made to the shipment, needs to be saved to the database
+			updateShipmentPickupAddress = true
+
+			// Find the DOPSIT service item and update the SIT related address fields. These fields
+			// will be used for pricing when a payment request is created for DOPSIT
+			for itemIndex := range *extraServiceItems {
+				sitServiceItem := &(*extraServiceItems)[itemIndex]
+				if sitServiceItem.ReService.Code == models.ReServiceCodeDOPSIT {
+					sitServiceItem.SITOriginHHGActualAddress = serviceItem.SITOriginHHGActualAddress.Copy()
+					sitServiceItem.SITOriginHHGActualAddressID = nil
```
gidjin, line 190: As I understand this will create a new record with the same address as on the `DOFSIT` service item. Should they instead point to the same address record in the database instead? Same question with original address on the next couple lines

pkg/services/mto_service_item/mto_service_item_creator_test.go
```diff
@@ -417,16 +418,34 @@ func (suite *MTOServiceItemServiceSuite) TestCreateOriginSITServiceItem() {
 	reason := "lorem ipsum"
 
 	suite.T().Run("Create DOFSIT service item and auto-create DOASIT, DOPSIT", func(t *testing.T) {
+		// Customer gets new pickup address for SIT Origin Pickup (DOPSIT) which gets added when
+		// creating DOFSIT (SIT origin first day).
+
+		// Do not use testdatagen.MakeAddress, because if the information is coming from the Prime
+		// via the Prime API, the address will not have a valid database ID. And tests needs to ensure
+		// that we properly create the address coming in from the API.
+		actualPickupAddress := models.Address{
```
gidjin, line 427: Should there be a test to check that this address ends up in the db on both `DOFSIT` and `DOPSIT`? Maybe there is and I missed it?

gidjin (commented):
Looks good! I had some questions though which is why I didn't approve it, yet.

pkg/services/mto_service_item/mto_service_item_creator.go
jacquelineIO: Originally this code wasn't using `Copy`. But with the `Copy` function I can clean this up a bit. But ultimately in some places where the ID was being updated, I would have wanted to use a `Copy` (even if already copied).

pkg/services/mto_service_item/mto_service_item_creator.go
jacquelineIO: Removing this, so that if it is set, it won't get overridden. We don't verify these ID's so we are trusting that these are good.

pkg/handlers/primeapi/mto_service_item_test.go
```diff
@@ -54,13 +49,31 @@ func (suite *HandlerSuite) TestCreateMTOServiceItemHandler() {
 	sitEntryDate := time.Now()
 	sitPostalCode := "00000"
 
+	// Customer gets new pickup address for SIT Origin Pickup (DOPSIT) which gets added when
+	// creating DOFSIT (SIT origin first day).
+	//
+	// Do not use testdatagen.MakeAddress, because if the information is coming from the Prime
```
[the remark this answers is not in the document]
jacquelineIO: Didn't know about this!

pkg/services/mto_service_item/mto_service_item_creator_test.go
jacquelineIO: Yes, I'm testing this fully in the file `mto_service_item_test.go` I'm not sure why but seemed like there were other `DOPSIT`/`DOFSIT` being done in that file. But I could see having them here as well.

jacquelineIO: https://github.com/transcom/mymove/pull/5709/files#diff-6cdd433b2d5e9d4fec5a173888ea19939c70f6d44f65272f379031ab7e79b5f7R742

pkg/services/mto_service_item/mto_service_item_creator.go
jacquelineIO: I wasn't sure. But seems reasonable for the SIT items to use the same Address ID. So I've updated that they all point to the same one.  They should NOT share an ID with the MTO Shipment, so leaving that as-is.

lynzt (commented):
@jacquelineIO we should prob add an index on those 2 FK fields...

pkg/services/mto_service_item/mto_service_item_creator.go
jacquelineIO: fixed

migrations/app/schema/20210119230439_update_create_sit_origin_actual_address.up.sql
```diff
@@ -0,0 +1,14 @@
+-- Column add
+ALTER TABLE mto_service_items
+    ADD COLUMN sit_origin_hhg_original_address_id uuid,
+    ADD COLUMN sit_origin_hhg_actual_address_id uuid,
+    ADD CONSTRAINT mto_service_items_sit_origin_hhg_original_address_id_fkey FOREIGN KEY (sit_origin_hhg_original_address_id) REFERENCES addresses (id),
+    ADD CONSTRAINT mto_service_items_sit_origin_hhg_actual_address_id_fkey FOREIGN KEY (sit_origin_hhg_actual_address_id) REFERENCES addresses (id);
```
reggieriser, line 6: Nit: You can add the constraint as part of adding the column and avoid repeating the column name and to make it more succinct.  Should be same results in the end, though, so not a big deal.

reggieriser (approved):
Migration looks good, so approving it (only -- still needs full code review approval).  One small nit, but certainly not a blocker.

gidjin (approved):
Thanks for making the updated! Code works as expected. :shipit:

PR #5711, opened 2021-01-21
no ticket: re-add acceptance tests that can be run conditionally

carterjones:
## Description

As part of https://dp3.atlassian.net/browse/MB-4391 (https://github.com/transcom/mymove/pull/5443), I removed acceptance tests from CircleCI. However, acceptance tests are still useful, so I'm adding them back, but only making them run conditionally.

I worked on this while trying to get the `crud` role added. It seemed worth taking through to completion. There were a few changes I had to make in the code that appear to be very specifically designed to be run during these tests. I made a few modifications to that code. If there is a better way to make this work reliably, please feel free to let me know and I'll iterate on this.

## Setup

Replace placeholder_branch_name on the acceptance_tests filter to run acceptance tests in CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -626,6 +626,72 @@ jobs:
           destination: test-results
       - announce_failure
 
+  # `acceptance_tests` runs acceptance tests for the webserver against the local, experimental, and staging environments.
+  acceptance_tests:
+    executor: mymove_medium
+    steps:
+      - aws_vars_legacy
+      - checkout
+      - restore_cache:
+          keys:
+            - go-mod-sources-v2-{{ checksum "go.sum" }}-{{ checksum "scripts/check-go-version" }}
+      - run:
+          name: Run Local acceptance tests
+          command: |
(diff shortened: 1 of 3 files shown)


travelar (approved):
This looks ok, but more eyes would be good.

.circleci/config.yml
```diff
@@ -626,6 +626,72 @@ jobs:
           destination: test-results
       - announce_failure
 
+  # `acceptance_tests` runs acceptance tests for the webserver against the local, experimental, and staging environments.
+  acceptance_tests:
+    executor: mymove_medium
+    steps:
+      - aws_vars_legacy
```
dynamike, line 633: Do we want to be running more thing in the legacy account?

carterjones: Good point. I removed it in 9c6853f44371dd10935ef549c2a9978daa0ab613 and it still worked. This made me curious about whether or not all of those environment variables needed to be defined. So I removed them all, slowly adding them back in until the acceptance tests passed. The result is 6f268c01d5edd3c8088133778ee4c4e3fe49416a. Here is the resulting job showing that everything still passed: https://app.circleci.com/pipelines/github/transcom/mymove/26526/workflows/ff4c6287-9eaf-4f43-bc19-feb57c1766a9/jobs/408574

PR #5712, opened 2021-01-21
[MB-3239] Redux Refactor: Uploads

suzubara:
## Description

The goal of this story was to migrate / deprecate code from the `Uploader/ducks` file and resolve outstanding bugs related to uploading documents & Redux. This resulting PR looks a bit different from the other Redux Refactor ones for a couple of reasons:
- The newer `entities/modules/documents` code was being used in the customer-facing Upload Orders feature, but in a different way from other entities; this is because `documents` is not included inside any other schema, and is also never fetched on its own (in other words, there is no GET documents call), therefore `entities.documents` is never actually populated.
- Similarly, the `entities.uploads` reducer is never populated for the same reasons, except for when `ADD_ENTITIES` was dispatched after creating a new file. However, uploading a file this way also modifies the service member's `orders` object, which was not being re-fetched after uploading orders. This meant that whenever a service member uploads orders, the changes were reflected inside `entities.uploads` but _not_ `entities.orders`.

The above conditions resulted in a couple of bugs which I aimed to address in this PR. They are reproducible on `master`:
- As a service member, uploading an Orders file and then proceeding to the Home page does not show the Orders you had just uploaded, since Orders is not re-fetched after creating an upload
- As a service member, deleting an Orders file that had already been uploaded results in an error message and the UI does not get updated even though the file does get deleted on the B/E

Ultimately I made these changes:
- As far as I could see, `Uploader/ducks` was already deprecated and not being used anymore, so I deleted it outright as well as other deprecated code
- Migrated the `createUpload` and `deleteUpload` actions to use inline API calls. However, instead of storing the responses in Redux as with other API calls, I make a refetch of service member orders instead, so that orders are immediately updated to reflect the new upload.
- Added `OrdersUploader` stories
- Relocated `UploadsTable` component, fixed eslint errors, and fixed issue with the Delete button (which was using an `<a>` element instead of a button)
- I also added `file.id` to the `key` prop in `DocsUploaded` to resolve a React error if there are multiple uploads with the same filename

## Reviewer Notes

These changes will result in a slightly different UX, where uploaded orders will appear above the uploader as soon as they are uploaded (see screen capture). This work uncovered another bug, which is that clicking the delete X _inside_ the Filepond UI (directly after uploading a file) is not working. Since this bug is not related to Redux, but rather to how we are using the Filepond component, I will plan to address it as part of [this story to upgrade react-filepond](https://dp3.atlassian.net/browse/MB-4362).

Also note that the original story mentions the PPM upload expenses/weight tickets functions. I reviewed this code and found it does not use Redux and appears to function without bugs, so I didn't touch it.

## Setup

To test, run the app locally:
```
make server_run
make client_run
```

- Navigate to http://milmovelocal:3000/ and log in as a new service member
- Complete onboarding to the Upload Orders step. Verify you can upload files and they should appear in the UI after uploading. Verify you can upload and delete multiple files on this step.
- On the next step (Moving 101) click Finish Later. You should be returned to the Home page, and the orders files you've uploaded should be visible under Step 2 ("Orders uploaded").
- Click the Edit link to edit your orders, and verify that as on the initial Upload Orders step, you are able to upload and delete orders files.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3239) for this change

## Screenshots

https://user-images.githubusercontent.com/2723066/105418987-9c304b00-5c03-11eb-814b-881bf0c79f76.mp4


YanZ777 (approved):

kimallen (commented):
I'm not sure if this is in the scope of this refactor or if it's an existing bug to look at later.  The upload image that turns blue says to tap to delete.  If I try to delete the upload from there instead of the table, I get an undefined error.  If I delete it from the table, then the blue upload images remains, and I'd expect that to disappear too.
image

The request url is missing an id 
image
This happens on both the upload orders page and the edit orders page.

Otherwise, it looks good and I'm seeing the orders on the home page and the tables as expected after uploading and deleting.

suzubara:
> I'm not sure if this is in the scope of this refactor or if it's an existing bug to look at later. The upload image that turns blue says to tap to delete. If I try to delete the upload from there instead of the table, I get an undefined error. If I delete it from the table, then the blue upload images remains, and I'd expect that to disappear too.
> image
> 
> The request url is missing an id
> image
> This happens on both the upload orders page and the edit orders page.
> 
> Otherwise, it looks good and I'm seeing the orders on the home page and the tables as expected after uploading and deleting.

Yup, that is the new bug I mentioned, which I want to handle as part of the filepond upgrade since it is related to how we tell filepond about existing uploads

kimallen:
> > I'm not sure if this is in the scope of this refactor or if it's an existing bug to look at later. The upload image that turns blue says to tap to delete. If I try to delete the upload from there instead of the table, I get an undefined error. If I delete it from the table, then the blue upload images remains, and I'd expect that to disappear too.
> > image
> > The request url is missing an id
> > image
> > This happens on both the upload orders page and the edit orders page.
> > Otherwise, it looks good and I'm seeing the orders on the home page and the tables as expected after uploading and deleting.
> 
> Yup, that is the new bug I mentioned, which I want to handle as part of the filepond upgrade since it is related to how we tell filepond about existing uploads

Oh I see that now.  I missed that in your notes the first read through.  Thanks for the explanatory description.

kimallen (approved):
LGTM, as long as Design is cool with the new table view.

amanda5iu (approved):
I think the new view looks fine! Thanks for flagging design to look at this.

PR #5714, opened 2021-01-22
Added test for send to staging

shimonacarvalho:
## Description

There is a new test called `Test_SendStgNotification` that sends a real notification to the live staging environment.
You can run the test by running all webhook tests
`go test ./cmd/webhook-client/webhook -v`

But do check for the successful passing of `Test_SendStgNotification`

Since we now have a "real" test in there, I had to remove the webhook client tests from circleci. It's a bit of a catch-22, we are writing this test to help infra get a working version in staging. Currently that is not working. That also means this test won't work until we complete that effort, and therefore we're not ready to run all our webhook tests in staging. 

## Reviewer Notes

I also renamed db-webhook-notify to just webhook-notify as db was unnecessary. 

## Reference
* [MB-6193 Push Deploy: Write e2e test for webhook client - Jira](https://dp3.atlassian.net/browse/MB-6193)

### DIFF
--- a/cmd/webhook-client/utils/connection.go
+++ b/cmd/webhook-client/utils/connection.go
@@ -16,6 +16,13 @@ import (
 	"github.com/transcom/mymove/pkg/cli"
 )
 
+const (
+	// RecipientMTLSCert is the client cert to connect to the subscriber
+	RecipientMTLSCert string = "gex-mtls-client-cert"
+	// RecipientMTLSKey is the client key to connect to the subscriber
+	RecipientMTLSKey string = "gex-mtls-client-key"
+)
+
 // WebhookClientPoster is an interface that WebhookRuntime implements
 type WebhookClientPoster interface {
 	SetupClient(cert *tls.Certificate) (*WebhookRuntime, error)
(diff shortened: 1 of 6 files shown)


carterjones (approved):
Giving a ✅ since the CircleCI stuff looks good. I'll leave the Go code review to others.

cmd/webhook-client/utils/connection.go
```diff
@@ -16,6 +16,13 @@ import (
 	"github.com/transcom/mymove/pkg/cli"
 )
 
+const (
+	// RecipientMTLSCert is the client cert to connect to the subscriber
+	RecipientMTLSCert string = "gex-mtls-client-cert"
+	// RecipientMTLSKey is the client key to connect to the subscriber
+	RecipientMTLSKey string = "gex-mtls-client-key"
+)
```
abbyoung, line 24: yay 🎊

cmd/webhook-client/utils/connection.go
```diff
@@ -156,35 +162,47 @@ func CreateClient(v *viper.Viper) (*WebhookRuntime, *pksigner.Store, error) {
...
 	rc, err = runtimeClient.SetupClient(cert)
-
 	if err != nil {
 		return nil, nil, err
 	}
 
 	return rc, store, nil
 }
+
+func loadCertificate(v *viper.Viper) (*tls.Certificate, error) {
```
abbyoung, line 188: Yay for splitting this into its own function

abbyoung (approved):
LGTM! Awesome to have a test working with stg, and thank you for all the extra cleanup you did. ⭐ 🏅 🚀

cmd/webhook-client/utils/connection.go
shimonacarvalho: I had to add a new mode - getting the contents from a string, rather than a filepath, so then it got messy to have that all in the other func.

PR #5717, opened 2021-01-22
[MB-6194] Extend user endpoint to update Active status

abbyoung:
## Description

This PR updates the existing endpoint `PATCH /users/{userId}` to support changing the `Active` status of a user in the `users` table.

* The previous iteration used the update endpoint only to revoke user sessions, which was reflected in the naming.
* I've renamed all related functions, etc. to `UpdateUser`, `UserUpdatePayload`, etc. None of the previous code by @monfresh was altered except for renaming.
* This makes the code more flexible in case we need to add additional user properties to update in the future.
* Added new service object to handle updating the user (aside from session revocation, which already has its own SO). For now, this only handles the Active status, but I left a comment to direct any future engineer to add their changes to `UpdateUser`.
* Updated tests for the handler and added tests for the service object

## Reviewer Notes

See: https://github.com/transcom/mymove/compare/update-user-in-admin-app-mb-6194?expand=1#diff-1c02fbb6907330dee84754ccc54da0c2f62468c90e66e838504434b740df4279R126

Would love to know if there's a better way to do this in golang!

## Setup

Tests:

`go test ./pkg/services/user --testify.m "TestUserUpdater" -v`
`go test ./pkg/handlers/adminapi/ --testify.m "TestUpdateUserHandler" -v`

You can also test the endpoint using Postman if you would like to brave the CSRF token setup. YOU CAN DO IT. Ping me if you want to pair to set it up.

Here are some instructions I wrote down for myself:

* Make sure you have Postman Interceptor installed in your browser and you've added `adminlocal` as a domain
* Open Postman -> Make sure Interceptor is turned on in Postman and you've added `adminlocal` as a domain
* Run `make server_run` and `make admin_client_run`
* In admin app, log in as local admin user
* Open up dev tools -> Application -> Cookies -> Copy the token for `admin_session`
* In Postman, make GET request to `http://adminlocal:3000/admin/v1/users/{userId}` You'll probably get a 403 because we need to update the admin_session token.
* Click the `cookies` link in your postman app (underneath the request path) -- you should see all populated. Click the `admin_session` cookie, paste in your new value, and save.
* You can redo the GET request now and you should receive a response.
* To do a `PATCH` request, click the `cookies` link again and copy the `masked_gorilla` token value.
* Add a header for your request with the key `X-CSRF-TOKEN` and value `<token value you copied>`
* Add a body to toggle the `active` status.
```
{ "active": false}
```
* Hit it!

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6194) for this change


eamahanna (approved):
These changes make sense to me and everything worked in postman :)

pkg/handlers/adminapi/users.go
```diff
@@ -88,15 +88,16 @@ func (h IndexUsersHandler) Handle(params userop.IndexUsersParams) middleware.Res
...
+type UpdateUserHandler struct {
 	handlers.HandlerContext
 	services.UserSessionRevocation
+	services.UserUpdater
 	services.NewQueryFilter
 }
 
-// Handle revokes a user session
-func (h RevokeUserSessionHandler) Handle(params userop.RevokeUserSessionParams) middleware.Responder {
+// Handle updates a user's Active status and/or their sessions
```
shimonacarvalho, line 99: Thanks for making these comments useful and correct.

pkg/handlers/adminapi/users.go
```diff
@@ -88,15 +88,16 @@ func (h IndexUsersHandler) Handle(params userop.IndexUsersParams) middleware.Res
...
+func (h UpdateUserHandler) Handle(params userop.UpdateUserParams) middleware.Responder {
 	session, logger := h.SessionAndLoggerFromRequest(params.HTTPRequest)
 	payload := params.User
```
shimonacarvalho, line 102: Is params.User required in the payload? If not, you later do *payload which will cause a possible nil dereference.

pkg/handlers/adminapi/users_test.go
```diff
@@ -241,59 +244,101 @@ func (suite *HandlerSuite) TestRevokeUserSessionHandler() {
...
 			userservice.NewUserSessionRevocation(queryBuilder),
+			userservice.NewUserUpdater(queryBuilder),
 			newQueryFilter,
 		}
 
 		response := handler.Handle(params)
 		foundUser, _ := models.GetUser(suite.DB(), userID)
 
-		suite.IsType(&userop.RevokeUserSessionOK{}, response)
+		suite.IsType(&userop.UpdateUserOK{}, response)
```
shimonacarvalho, line 259: Great job checking the details returned here.

pkg/handlers/adminapi/users_test.go
```diff
@@ -241,59 +244,101 @@ func (suite *HandlerSuite) TestRevokeUserSessionHandler() {
...
+		).Return(&user, nil, nil).Once()
+
+		handler := UpdateUserHandler{
 			handlerContext,
+			userRevocation,
 			userUpdater,
 			newQueryFilter,
 		}
 
 		response := handler.Handle(params)
```
shimonacarvalho, line 313: Can you call the swagger check prior to each of these Handle commands? It unearths a lot of invisible bugs. 
`suite.NoError(params.Body.Validate(strfmt.Default))`

pkg/handlers/adminapi/users_test.go
```diff
@@ -241,59 +244,101 @@ func (suite *HandlerSuite) TestRevokeUserSessionHandler() {
...
 
-	userUpdater.On("RevokeUserSession",
+	userRevocation.On("RevokeUserSession",
 		mock.Anything,
 		params.User,
 		sessionManagers[0].Store,
 	).Return(nil, err, nil).Once()
 
-	handler := RevokeUserSessionHandler{
+	userUpdater.On("UpdateUser",
```
shimonacarvalho, line 328: Awesome job using mocks! Can I suggest that you add to each test a description of what is being setup and what is expected? Makes it easier for reviewer to doublecheck everything and for future updater to understand how to edit the test. 

I do something like this at the top of each test:
```
		// Under test: sendOneNotification function
		// Mocked:     Client
		// Set up:     We provide a FAILING webhook notification, and make the client fail twice to send
		//             and then succeed
		// Expected outcome:
		//             Client.Post called 3 times in total
		//             Notification would be updated as SENT
```

pkg/handlers/adminapi/users_test.go
```diff
@@ -241,59 +244,101 @@ func (suite *HandlerSuite) TestRevokeUserSessionHandler() {
...
 			handlerContext,
+			userRevocation,
 			userUpdater,
 			newQueryFilter,
 		}
 
 		response := handler.Handle(params)
 
-		suite.IsType(&userop.RevokeUserSessionOK{}, response)
+		suite.IsType(&userop.UpdateUserOK{}, response)
```
shimonacarvalho, line 315: I'm not understanding why the response should be ok if the update failed?

pkg/services/user/user_updater.go
```diff
@@ -0,0 +1,48 @@
...
+	var foundUser models.User
+
+	filters := []services.QueryFilter{query.NewQueryFilter("id", "=", id.String())}
+	err := o.builder.FetchOne(&foundUser, filters)
+
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if payload.Active != nil {
```
shimonacarvalho, line 33: payload itself is a pointer so are we sure here that it cannot be nil? If it could then payload.Active could cause a nil dereference.

shimonacarvalho (changes requested):
Great job and a lot of good work! 

I found some potential nil dereferences. In the handler, I'm not sure they will cause them because it's possible that swagger would ensure that certain things would never be nil. In the service object, I'd recommend we doublecheck, because someone else could reuse that SO.

Some updates to the tests requested.

pkg/handlers/adminapi/users_test.go
abbyoung: good question -- this was a test i updated but you're right. i'll dig in and figure out what's actually supposed to happen here. thanks!

pkg/handlers/adminapi/users.go
abbyoung: `params.User` represents the required `UpdateUserPayload` in `admin.yaml`. Should I still check that it's not nil?

pkg/handlers/adminapi/users_test.go
monfresh: Oh my! I wrote that. It looks like I meant to copy and paste what comes after that test into the test. So, lines 318-342 should go inside the test that starts on line 291, and it should replace lines 292-315.

shimonacarvalho (approved):
LGTM! Good job persevering! This one was full of booby traps!

PR #5699, opened 2021-01-20
stub PPM rate engine mb-6160

lynzt:
## Description

We [aren't loading BVS](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1609870736312200?thread_ts=1609869796.308500&cid=CP6PTUPQF) into our database anymore which [causes errors](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1609869796308500) when trying to price a PPM move.

For a temp fix, let's hard code the values returned by the function that looks at the BVS data to calculate `linehaul discount rate` and the `sit discount rate`. Fixing the PPM rate engine will be address in outcome 5.

## Reviewer Notes

Tests should pass and there should be no 500 errors when creating a PPM move in 2021 - there will be one 500 error when you first load the page which is outside the scope of this story.

The PPM data is hard coded on the review page so don't expect to see the incentive.
Added a yellow background to the `edit weight` page to indicate the shown price isn't "real"

## Setup

`make server_test`
`make e2e_test`

--------

`make server_run`
`make client_run`

Create a PPM move.

When on the weight slider page, you will see one 500 error when the page loads for the first time. After that, you should not see any 500 errors on reload or moving the weight slider.
the slider page: `http://milmovelocal:3000/moves/d16db6de-ef83-4121-89a2-46319ba8eb65/ppm-incentive`

When on the edit weight page, you shouldn't see any 500 errors when updating the weight in the textbox.
edit weight page: `http://milmovelocal:3000/moves/d16db6de-ef83-4121-89a2-46319ba8eb65/review/edit-weight`

## Code Review Verification Steps

* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6160) for this change

## Screenshots

added the yellow background to the `estimated incentive` to show this was not a "valid incentive range".

Screen Shot 2021-01-22 at 09 16 50

PR #5732, opened 2021-01-25
Adjust sometimes-failing test to account for stubbed PPM discount rates

reggieriser:
## Description

In #5699, the discount rates were hard-coded to account for the fact we aren't loading BVS rates anymore.  The test adjusted in this PR would dynamically pick one of two TSPP records (whichever's ID came first alphabetically) to use for the target of the test assertions.  But now that we've hard-coded the rates, the test would only pass if one of the two were picked.  This PR hard-codes the IDs in the test itself to make it always pick one of the two TSPP records as the target.

## Reviewer Notes

We may not ultimately need this test after the PPM rate engine is reworked in the future, but I didn't want to assume that yet.

## Setup

`make server_test`


lynzt (approved):
LGTM... thanks for finding this!

🚢

PR #5738, opened 2021-01-25
MB-6481 Sign user out after being idle

monfresh:
## Description

In May 2018, we implemented a feature that would warn the user that
they have been idle, and if they remain idle, they will be signed out.
We used `react-idle-timer` version 3.0 for this functionality, and it
worked. But then in November 2019, we [upgraded](https://github.com/transcom/mymove/commit/58e1e93604637bf7fcfa70b358d7127254182c6b) `react-idle-timer`
from 3.0.0 to 4.2.12, and because we didn't have any tests for this
feature, we didn't notice that it broke.

Version 4.0 introduced new syntax, and so the React component never
detected an idle state. Even though `react-idle-timer` was no longer
working, we should have been able to rely on the session cookie to
expire as a fallback, so that even though the user wasn't returned to
the sign in page, if they refreshed the page, they should have been
signed out. However, the original implementation added an interval
that would ping the `/internal/swagger.yaml` endpoint every 10 minutes.
This unfortunately kept the session alive, well past the required
15-minute expiration for an idle session. Back then, I don't think we
set a lifetime for the session, and so until we moved to server-side
sessions in June 2020, this would keep the session alive forever.

In this PR, I fixed the syntax issues (`activeAction` -> `onAction`,
and `idleAction` -> `onIdle`), and I also made some improvements:

- The internal endpoint is no longer pinged automatically by us. It
only gets pinged if the user becomes active again.
- To reduce the amount of requests to the internal endpoint, I limited
the events that make the user active to pressing a key or touching/
clicking on the page. By default `react-idle-timer` looks for several
more events, such as a mouse move, but making a request every time the
mouse moves is not efficient. Also, most websites that implement idle
timeouts display a modal or some other prompt that the user must click
on to remain active, so I chose events that are similar to that.
- I renamed the constants to better represent what they mean, as
opposed to naming them after specific time amounts. I also split out
the minutes, seconds, and milliseconds to make it easier to change
values for testing without having to do math in your head.
- I added a countdown that shows how much time is left before you get
signed out. Previously, the alert only said that you will be logged
out "shortly".
- I removed the `isProduction` condition to make it easier to test
locally.
- I fixed the relationship between the different time constants so
that the maximum idle time is respected. Before, the user would end
up being idle for an extra minute because the `fifteenMinutesInMilliseconds`
constant is what determined when the inactivity alert first appeared,
and then the user was given an additional minute via `oneMinuteInMilliseconds`
before being signed out.

## Reviewer Notes

Notes for improvement:

- I was hoping to be able to set the `maxIdleTime` to point to the
`SESSION_IDLE_TIMEOUT_IN_MINUTES` env var, so that if we ever change
the default, we wouldn't have to remember to change it here as well.
It looks like we don't have access to the env var as is, but we would
need to define a new `REACT_APP_SESSION_IDLE_TIMEOUT_IN_MINUTES` env
var.

- We define an alert that tells the user that they have been signed
out due to inactivity, but it only shows up right before they get
signed out. It doesn't remain visible on the sign in screen, which is
where the alert would be most useful. I couldn't figure out how to
do that.

How to test locally:

- Change `maxIdleTimeInSeconds` to `10`
- Change `maxWarningTimeBeforeTimeoutInSeconds` to `5`
- Sign in to `milmovelocal` or `officelocal` with login.gov (not devlocal) and remain idle
- [ ] Verify that after 5 seconds, the inactivity alert appears, and
starts counting down from 10 seconds
- [ ] Verify that after the countdown, you are signed out and returned
to the sign in screen
- Sign back in, and after the alert appears and the countdown begins,
click anywhere on the screen before the time is up
- [ ] Verify that the alert disappears, and that a network request is
made to `/internal/swagger.yaml` every time you click on the screen
- Remain idle again
- [ ] Verify that after 5 seconds of idle time, the inactivity alert
appears, and starts counting down from 10 seconds, and that it only
lasts 10 seconds
- Sign back in
- After the alert appears, move the mouse on the screen, but don't
click it and don't press any keyboard keys
- [ ] Verify that mouse moves do not trigger any network requests, and
that you remain idle
- Change `onActive` and `onIdle` in the React component to `activeAction`
and `idleAction` respectively. This should disable the functionality.
- Add `export SESSION_IDLE_TIMEOUT_IN_MINUTES=1` to your `.envrc.local`
- Stop the server and client
- Run `direnv allow` in both terminal tabs
- Remain idle for at least 1 minute
- Refresh the web page
- [ ] Verify that you are signed out and back on the sign in screen

## Setup

```sh
make server_run
make client_run # in another tab
```

 have been satisfied.

### References
[Jira story](https://dp3.atlassian.net/browse/MB-6481)


src/shared/User/LogoutOnInactivity.jsx
```diff
@@ -3,57 +3,69 @@ import { connect } from 'react-redux';
...
+      this.setState({ timeLeftInSeconds: this.state.timeLeftInSeconds - 1 });
+    }
+  };
+
   render() {
     const props = this.props;
     return (
       <React.Fragment>
-        {isProduction && props.isLoggedIn && (
+        {props.isLoggedIn && (
```
duncan-truss, line 55: I see you've gotten rid of this flag but I did check and we should update the host urls from commercial to govcloud:

https://github.com/transcom/mymove/blob/40191aca6c7ffee186a59c9918590e41b5c06bd1/src/shared/featureFlags.js#L73-L89

src/shared/User/LogoutOnInactivity.jsx
```diff
@@ -3,57 +3,69 @@ import { connect } from 'react-redux';
...
-    this.timeout = setTimeout(() => {
-      if (this.state.isIdle) {
-        LogoutUser();
-      }
-    }, this.props.logoutAfterWarningTimeout);
+    clearInterval(this.timer);
+    this.timer = setInterval(this.countdown, 1000);
+  };
+  onAction = () => {
+    fetch(this.props.keepAliveEndpoint);
```
duncan-truss, line 40: This seems to be getting called too frequently, every click or key press requests internal/swagger.yaml anew. It's also a response that is never cached because cookies are being set.

I wonder if internal/users/is_logged_in is a better keepAliveEndpoint and it has a much smaller response.  What purpose does the keepAliveEndpoint solve?

src/shared/User/LogoutOnInactivity.test.jsx
```diff
@@ -0,0 +1,79 @@
...
+      expect(timer).toBeDefined();
+      expect(component.find('SomethingWentWrong')).toHaveLength(0);
+    });
+
+    it('becomes idle after 14 minutes', () => {
+      const idleTimer = component.find(IdleTimer);
+
+      expect(idleTimer.state().idle).toBe(false);
+
+      clock.tick(14 * 1000 * 60);
```
duncan-truss, line 47: I don't know if it meets your needs but we have another mock date library `mockdate` that we use elsewhere to mock what now() returns.

src/shared/User/LogoutOnInactivity.jsx
monfresh: As far as I can tell, the `isProduction` that I removed came from here: https://github.com/transcom/mymove/blob/master/src/shared/constants.js#L6, and is not related to these featureFlags. That's a good catch, though. I will open a new Jira issue to look into whether we use these flags, and if so, to update them.

Also, it looks like that was the only place we used `isProduction`, so we can probably remove it from the constants file.

duncan-truss: Yea it looks like the flags are being passed to the AppContext provider whereas isProduction was read directly from what NODE_ENV is set to at build time.

src/shared/User/LogoutOnInactivity.jsx
monfresh: It's not ideal, but the purpose is to reset the cookie expiration to match their active state. If we didn't make a backend request, and if the only thing the user did was click on the screen without actually triggering a request, then after 15 minutes (default cookie expiration), if they then tried to submit a form, they would get an unauthenticated error.

The typical way to implement idle timeout is to show a modal and the user has to click OK to remain active, and that would trigger a backend request to reset the cookie. That would be the only way for them to remain active. I prefer this solution, but I wanted to preserve the existing functionality and get it working again before making UI changes.

src/shared/User/LogoutOnInactivity.test.jsx
monfresh: I tried it but it didn't work as I expected it to.

src/shared/User/LogoutOnInactivity.jsx
duncan-truss: Thanks for explaining that.  It looks like most of our api queries set the cookie on response and `logged_in` gets called quite frequently too.  I would replace `keydown` with `blur` and `focus` and maybe set the `throttle` or `debounce` property as well to reduce the number of events.

duncan-truss:
I don't know when this Alert would ever get shown, I couldn't get it to show on either app.  If we logout and redirect we would need to find another way to trigger that message maybe like a query param. https://github.com/transcom/mymove/pull/5738/files#diff-29cafd4291047f1d4c4e2e0cc48109a60f71a3ee8e8ffbc1ea114cdcdbfa8c25R76-R80

monfresh:
Yeah, I know that the logged out alert doesn't work. I mentioned that in my commit message. To make it appear, we can add `this.setState({ showLoggedOutAlert: true });` right before `LogoutUser();`, but that would only show it briefly. I wish we could use Flash messages like in Rails. I will look into the query param. I thought about it but wanted to check with frontend folks to see if there was a better way.

monfresh:
@duncan-truss I fixed the alert after signing out. Can you give it a try?

monfresh:
One bug I found with this approach is that if you have multiple tabs open, and you are only active in one of them, `react-idle-timer` will sign you out in the inactive tab, and then you'll run into errors in the active tab because you've been signed out. 

Typically, this functionality is implemented with a backend endpoint that the frontend pings in order to determine if the user is still logged in, and checking the time remaining via the backend session. So, rather than having some React plugin determine idle state based on user activity in the current browser tab or window, the frontend should periodically ping an endpoint that not only returns whether or not the user is logged in, but also when the session expires, and how much time is remaining. 

I will write up a new story for this because this requires more involved work.

src/shared/User/api.js
```diff
@@ -27,7 +27,8 @@ export async function LogoutUser() {
   try {
     // Successful logout should return a redirect url
     let resp = await Swagger.http(req);
-    window.location.href = resp.text;
+    let redirect_url = timedout ? resp.text + '#timedout=true' : resp.text;
```
duncan-truss, line 30: I think we want to add timedout as a query param so it ends up in the location.search.  A hash would indicate an anchor link.

You should be able to use the query-string library (https://www.npmjs.com/package/query-string) to build the url instead of string appending or potentially use the react-router instead of setting window.location directly.

duncan-truss:
I think it would be good to have @transcom/truss-pamplemoose get eyes on this as well so they are aware.

src/shared/User/LogoutOnInactivity.jsx
```diff
@@ -3,81 +3,78 @@ import { connect } from 'react-redux';
...
 
-import { isProduction } from 'shared/constants';
 import Alert from 'shared/Alert';
 import { selectCurrentUser } from 'shared/Data/users';
 import { LogoutUser } from 'shared/User/api.js';
 
-const fifteenMinutesInMilliseconds = 900000;
-const tenMinutesInMilliseconds = 600000;
-const oneMinuteInMilliseconds = 60000;
+const maxIdleTimeInSeconds = 15 * 60;
```
Ryan-Koch, line 10: nice change, this is a fair bit easier to read

src/shared/User/LogoutOnInactivity.test.jsx
```diff
@@ -0,0 +1,79 @@
+/* eslint-disable no-only-tests/no-only-tests */
+/* eslint-disable react/jsx-props-no-spreading */
```
Ryan-Koch, line 2: Do we need these to be disabled?

src/shared/User/api.js
monfresh: Alright, here's the deal. `resp.text` here is the response from our own `/auth/logout` endpoint, which returns the URL of the request that is supposed to be sent to login.gov so that the user is signed out of login.gov as well as MilMove. That URL looks something like this: 
```
"https://idp.int.identitysandbox.gov/openid_connect/logout?id_token_hint=eyJhbGciOiJSUzI1NiJ9.eyJzdWIiOiJmNjk0MmRhYS01NDU3LTQ3MDMtYTE3My0zOTZhYWNiNWE1NzkiLCJpc3MiOiJodHRwczovL2lkcC5pbnQuaWRlbnRpdHlzYW5kYm94Lmdvdi8iLCJlbWFpbCI6Im1vbmNlZitpbnRzYW5kYm94QHRydXNzLndvcmtzIiwiZW1haWxfdmVyaWZpZWQiOnRydWUsImFjciI6Imh0dHA6Ly9pZG1hbmFnZW1lbnQuZ292L25zL2Fzc3VyYW5jZS9pYWwvMSIsIm5vbmNlIjoiOXJBbXd1bmNGU1dndTRBaFVRcDcteFZRalFBR2VPYVdCLUhRWmtUQkhkNV92SjJHLW5SU1MyZVVMNERsNjRTXzBhbDJwSU12MTVnRnFSWkVOSGZwM3c9PSIsImF1ZCI6InVybjpnb3Y6Z3NhOm9wZW5pZGNvbm5lY3QucHJvZmlsZXM6c3A6c3NvOmRvZDpteW1vdmVtaWxsb2NhbCIsImp0aSI6ImpwWVJVNzdZZWhYQVloZDFvOFE0Y2ciLCJhdF9oYXNoIjoidkUtbTZRQ1FqNGhObjJvQW52N0t4USIsImNfaGFzaCI6IjNraWo3Q2pWOGZGSGxaelM5T1l5OXciLCJleHAiOjE2MTE4Njg0OTEsImlhdCI6MTYxMTg2NzU5MSwibmJmIjoxNjExODY3NTkxfQ.MRq2hlL2lu73s2PX_HlafnYPKAnNDj23xsQz2c38oWQ3LnYKoKpTMsPAD3eV_xF0djkoUwK6CsQ5tbipBVNmPGh-hYPaThksfId6aLkg6GEDj-KiuoP5LO-IKqmB50rAhdC_jfLGbOAf3MbNryPbz-fb7oazlDbXFFLT8nveEr0CWtPXIJK2x0Lo2GHPOa60tvbrgI5j5eIovVLZuLJGxwZKDtlmbstKTtssKsA-BEwBCQOeC_ik7_lN5GPj7R45ntofqpMobHztVS6BdgiJIKY5NJFIs5FmBEdKC2XIvn2xGGJm_Rsm-WLnDNnwCiGHPTOC-woK8G_YllhhgyJEXw&post_logout_redirect_uri=http%3A%2F%2Fmilmovelocal%3A3000%2F&state=mWJUmgABLqEw4uuq-y2BLTET3XMekEDZRK5xfOSPuxUtPdSO-Mbj6fB4oNk0WcQrpTf2vrjAguZYg1a5S0JBKg%3D%3D"
```
It includes 3 query parameters that login.gov expects:
- id_token_hint
- post_logout_redirect_uri
- state

If we manipulate `resp.text`, we are manipulating the URL of the request to login.gov, and if we try to add additional query parameters, login.gov will not send them back when redirecting back to our `post_redirect_logout_uri`, rightfully so.
However, if we append hash parameters, it works. I'm not sure if that's expected or not. So, this appears to be the best solution we have at the moment. 

Using query-string, the line would look like this:
```js
let redirect_url = timedout ? qs.stringifyUrl({ url: resp.text, fragmentIdentifier: 'timedout' }) : resp.text;
```
Do you think it's still worth it to use it instead of appending a string?

monfresh: Another thing that's strange that I haven't looked into yet is that even though our `post_logout_redirect_uri` is just `http://milmovelocal:3000`, after the redirect happens, the `?state` parameter gets added.

src/shared/User/LogoutOnInactivity.test.jsx
monfresh: Good question. I removed them and `yarn lint` didn't complain.

src/shared/User/api.js
duncan-truss: Oof that's a bit more complex. Yea I guess in that case the fragment workaround is a good one.

Did you uri encode the query param when you first added it to the resp.text?  I do think it's good practice to use URL libraries even though you could append the fragment.  Especially since we're dynamically using the response url we get back.

monfresh:
Any idea why Happo is sad?

duncan-truss:
> Any idea why Happo is sad?

I would suggest merging in master, must have been a global style adjustment merged in is my guess.

src/shared/User/api.js
monfresh: Yeah, I encoded it. It doesn't make a difference. login.gov ignores it, and it adds the `state` parameter: https://github.com/18F/identity-idp/blob/b466c937f9eb4b462664dba86f84f2cab6dec0ec/app/forms/openid_connect_logout_form.rb#L83

jenniferlynparsons:
@monfresh @duncan-truss There was a global styling PR merged this morning, hence the unhappy hippo.

src/shared/User/api.js
monfresh: The good news is that login.gov will respect any query parameters we add to the `post_logout_redirect_uri`: https://github.com/18F/identity-idp/blob/b466c937f9eb4b462664dba86f84f2cab6dec0ec/app/services/uri_service.rb#L7

So, we will need to register a new redirect uri that includes the `timedout=true` parameter (login.gov only accepts exact matches), and then in the `LogoutUser` function in `api.js`, we can set a custom header like `X-Session-Timeout` that we can send to our LogoutHandler, and if the header is present, we can add the `timedout` parameter to pass along to login.gov

monfresh:
Thanks. That was going to be my next move.

src/shared/User/api.js
monfresh: It could also be a custom URL that we register, such as "https://my.move.mil/timedout"

Ryan-Koch (approved):

src/shared/User/api.js
duncan-truss: Yea that makes sense, I know OAuth has strict callback urls like that.  Who has permission to change the whitelisted redirect urls across environments in login.gov?  I tried logging into the sandbox (https://dashboard.int.identitysandbox.gov/) but it doesn't bring anything up.

duncan-truss:
When logged in locally I'm seeing a 403 forbidden error when the auth/logout endpoint is called?

Screen Shot 2021-01-29 at 10 47 45 AM

monfresh:
I can reproduce that, but only with devlocal. Are you seeing it when signing into the office app via login.gov? If it's just devlocal, I'm not going to worry about it for now.

package.json
```diff
@@ -135,6 +135,7 @@
     "mockdate": "^3.0.2",
     "prettier": "2.2.1",
     "react-test-renderer": "^17.0.1",
+    "sinon":"^9.2.4",
```
suzubara, line 138: It looks like the `sinon` package is just being added for its fake timers feature, but Jest already provides utilities for mocking timer functions: https://jestjs.io/docs/en/timer-mocks.html

Can we use that instead of adding a new package?

suzubara (changes requested):

monfresh:
Another bug I found, which we won't address in this PR, is that the `post_logout_redirect_uri` for the office app is `http://officelocal:3000`, but then the office App redirects to `http://officelocal:3000/sign-in`, and so the office user will never see the logged out alert.

duncan-truss:
> Another bug I found, which we won't address in this PR, is that the `post_logout_redirect_uri` for the office app is `http://officelocal:3000`, but then the office App redirects to `http://officelocal:3000/sign-in`, and so the office user will never see the logged out alert.

Yea I was seeing that too, the private route redirects and you lose your hash.  That 403 forbidden only happened with devlocal login not with login.gov on the office app.

package.json
monfresh: It's not clear to me from those examples how I would use the jest fake timers. What would be the equivalent of `timerGame(callback);`?

monfresh: Nevermind. I got it.

suzubara (approved):
thanks for switching out the fake timers!

duncan-truss (approved):

PR #5737, opened 2021-01-25
MB-6460 Add light link to Storybook

Shkeating:
## Description

This branch adds the light link styling to the links page in storybook. It also adds styling to the `overrides.scss` file where the other default link styles reside.

## Reviewer Notes

If anyone has specific opinions on how the class was named for this, I would appreciate feedbac on that. Setting global styles can be risky and I want to assure we are all in alignment on how global styles are called.

## Setup

```sh
yarn storybook
```
**Global > Typography > Links**

## Code Review Verification Steps

* [x] There are no aXe warnings for UI.
* [x] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [x] User facing changes have been reviewed by design.
* [x] Request review from a member of a different team.
* [x] Have the Jira acceptance criteria been met for this change?
* [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6460?atlOrigin=eyJpIjoiNTE1MGRlNTQ5NjFiNGMxMmI0NjU4MWUzYjQ0MzNkZjUiLCJwIjoiaiJ9) for this change

## Screenshots
default state:
image

hover:
image

focus:
image

PR #5740, opened 2021-01-26
Kim MB-4729 Apply Milmove Header

YanZ777:
## Description

Put in the new MilMoveHeader for the TOO.

## Reviewer Notes

Given that this PR is out there: https://github.com/transcom/mymove/pull/5737/files
I'm not certain that all the css changes are needed. I might wait until the above PR gets merged in, and then adjust as necessary. Please let me know if that is a good idea or not. Regardless, I'd still like to get eyes on this PR early.

This PR does not cover putting in the header for the sign in page. 

I took out the padding on all the pages so that the header can fit without there being a horizontal white bar between the MilMoveHeader and the customer header. The padding will be added back to pages on a per page basis in another story.

## Setup

```sh
make office_client_run
make server_run
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4729) for this change

## Screenshots

image

image


src/components/MilMoveHeader/index.module.scss
```diff
@@ -44,20 +49,28 @@
...
   :global(.usa-logo) {
     margin: 0;
     padding-top: 5px;
   }
+
+  :global(.usa-button.usa-button--unstyled) {
+    color: $link-light;
+    &:hover,
+    &:focus {
+      color: #b7caf0;
```
kimallen, line 73: We should use a color variable from `colors.scss` for this rather than putting the code directly into the styles.   I don't see this particular one in there, but we should check with design if there is one close enough or if it's okay to add this one as well.

src/shared/User/LoginButton.jsx
```diff
@@ -37,11 +34,9 @@ const LoginButton = (props) => {
     };
 
     return (
-      <li className="usa-nav__primary-item">
-        <a className="usa-nav__link" href="#" onClick={handleLogOut}>
-          Sign Out
-        </a>
-      </li>
+      <Button unstyled className="usa-nav__link" onClick={handleLogOut} type="button">
```
kimallen, line 37: I know we removed the li's when we were trying to use this component inside the dark background of the MilmoveHeader.  And we were experimenting with using <Button> instead of the <a>.  I don't think these changes hurt anything to keep, and they also may not be necessary, especially since we are likely refactoring sign-in in a near-future story.

src/components/MilMoveHeader/index.module.scss
amanda5iu: I think it makes sense to add this color in! I imagine we'll eventually have other uses for `link--light`, so it makes sense to have all the relevant colors codified.

YanZ777: I can add that in this PR, what should it be named?

amanda5iu: In the design libraries we call it `a--light:hover` - something similar to that would be great. But I defer to whatever color naming pattern we use for regular links. Maybe `$link-light-hover`?

amanda5iu (approved):
This looks great! Thanks Yan!

src/pages/Office/index.jsx
```diff
@@ -142,7 +145,23 @@ export class OfficeApp extends Component {
       <div className={siteClasses}>
         <FOUOHeader />
         {displayChangeRole && <Link to="/select-application">Change user role</Link>}
-        {!hideHeaderPPM && <QueueHeader />}
+        {!hideHeaderPPM && (
+          <>
+            {(!userIsLoggedIn || activeRole === roleTypes.TIO) && <QueueHeader />}
```
suzubara, line 150: The `<QueueHeader />` needs to show up for office users with the PPM role also -- as-is if I log in as a PPM office user, there is no header at all and I'm not able to sign out

src/shared/User/LoginButton.jsx
suzubara: I think we might want to revert these changes because they are affecting the customer app (it's hard to see the links now)..

image

suzubara (changes requested):
This looks great for the most part, I just added a couple of comments for issues in the PPM Office app and Customer app that I think should be resolved before merging

suzubara:
@YanZ777 to answer this question:
> Given that this PR is out there: https://github.com/transcom/mymove/pull/5737/files
> I'm not certain that all the css changes are needed. I might wait until the above PR gets merged in, and then adjust as necessary. Please let me know if that is a good idea or not. Regardless, I'd still like to get eyes on this PR early.

I don't think you need to worry about it, it doesn't look like that PR will affect the work in this one

Shkeating (approved):
looks like issue in customer app has been resolved :)
image

I am seeing a `Something Went Wrong` on the office app when logged in as a TOO user, but I'm guessing that is unrelated to the header? 
image

suzubara (approved):

PR #5742, opened 2021-01-26
[MB-5370] Annotate md5 hash

Ronolibert:
## Description

Add annotations to usage of MD5 hash

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5370) for this change


kimallen (approved):
LGTM!  Glad it could be mitigated with annotation rather than working through using another package!
  
Will you be sure to add the link to the private fork in the story so that the reviewer here can check that and any progression of the ticket?  Also, ideally an engineer would review it in the fork PR in addition to Jarod.  Then once it gets here, it can quickly be approved if checks passed.

pkg/storage/storage.go
```diff
@@ -37,16 +41,21 @@ type FileStorer interface {
...
-	hash := md5.New()
+	//RA Summary: gosec - G401 - Weak cryptographic hash
+	//RA: This line was flagged because of the use of MD5 hashing
+	//RA: This line of code hashes the AWS object to be able to verify data integrity
+	//RA: Purpose of this hash is to protect against environmental risks, it does not
+	//RA: hash any sensitive user provided information such as passwords
+	//RA: AWS S3 API requires use of MD5 to validate data integrity.
+	//RA Developer Status: Mitigated
+	//RA Validator Status: Mitigated
+	//RA Validator: jneuner@mitre.org
```
YanZ777, line 56: Do the other comments need to include the RA Validator?

PR #5733, opened 2021-01-25
[MB-1578] Add DistanceZipSITDest lookup

reggieriser:
## Description

This PR adds the lookup for the `DistanceZipSITDest` param key (which we renamed from `DistanceZip5SITDest` since it could be based on a zip5 or a zip3).

## Reviewer Notes

There is nothing to do for acceptance on this PR at this point since this lookup is a precursor for pricing SIT at destination.

## Setup

`make server_test`

## Code Review Verification Steps

* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#logging)
* [ ] The requirements listed in [Querying the Database Safely](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#querying-the-database-safely) have been satisfied.
* Any new migrations/schema changes:
  * [ ] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://github.com/transcom/mymove/wiki/migrate-the-database#zero-downtime-migrations))
  * [ ] Have been communicated to #g-database
  * [ ] Secure migrations have been tested following the instructions in our [docs](https://github.com/transcom/mymove/wiki/migrate-the-database#secure-migrations)
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-1578) for this change
* [SIT Pricing Notes](https://docs.google.com/document/d/1jRl91aslMnOMt2zS1tt_XT0lD1I-xvsiKOfprAaFrAs/edit#heading=h.qvz570f9mzig)

PR #5749, opened 2021-01-26
[MB-1576] Add lookup for DistanceZipSITOriginLookup

mboren:
## Description

This PR adds the lookup for the DistanceZipSITOrigin param key.

This is extremely similar to the lookup for `DistanceZipSITDest`, which was added in this PR: https://github.com/transcom/mymove/pull/5733

Since most of the code for the actual distance lookup is identical to the `DistanceZipSITDest` lookup, I thought about pulling it out into a helper function that could be used in both, but I figure we're more likely to want to change one and not the other, than we are to add a third lookup with the same logic.

## Acceptance testing
We can't test this directly on staging yet, so no acceptance is needed.

## Setup

```sh
make server_test
```

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-1576) for this change

### DIFF
--- a/pkg/payment_request/service_param_value_lookups/distance_zip_sit_origin_lookup_test.go
+++ b/pkg/payment_request/service_param_value_lookups/distance_zip_sit_origin_lookup_test.go
@@ -1,0 +1,158 @@
+package serviceparamvaluelookups
+
+import (
+	"errors"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/mock"
+
+	"github.com/transcom/mymove/pkg/models"
+	"github.com/transcom/mymove/pkg/route/mocks"
+	"github.com/transcom/mymove/pkg/testdatagen"
+)
+
+func (suite *ServiceParamValueLookupsSuite) TestDistanceZipSITOriginLookup() {
+	key := models.ServiceItemParamNameDistanceZipSITOrigin
+
+	originZip := "30901"
+	actualOriginZipSameZip3 := "30907"
+	actualOriginZipDiffZip3 := "36106"
+
+	reService := testdatagen.FetchOrMakeReService(suite.DB(),
+		testdatagen.Assertions{
+			ReService: models.ReService{
+				Code: "DOFSIT",
+			},
+		},
+	)
+
+	originAddress := testdatagen.MakeAddress(suite.DB(),
+		testdatagen.Assertions{
+			Address: models.Address{
+				PostalCode: originZip,
+			},
+		})
+
+	actualOriginSameZip3Address := testdatagen.MakeAddress(suite.DB(),
+		testdatagen.Assertions{
+			Address: models.Address{
+				PostalCode: actualOriginZipSameZip3,
+			},
+		})
+
+	actualOriginDiffZip3Address := testdatagen.MakeAddress(suite.DB(),
+		testdatagen.Assertions{
+			Address: models.Address{
+				PostalCode: actualOriginZipDiffZip3,
+			},
+		})
+
+	move := testdatagen.MakeDefaultMove(suite.DB())
+
+	mtoShipment := testdatagen.MakeMTOShipment(suite.DB(),
+		testdatagen.Assertions{
+			PickupAddress: originAddress,
+		})
+
+	paymentRequest := testdatagen.MakePaymentRequest(suite.DB(),
+		testdatagen.Assertions{
+			Move: move,
+		})
+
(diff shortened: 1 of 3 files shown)


pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go
```diff
@@ -331,6 +345,14 @@ func ServiceParamLookupInitialize(
 		return nil, err
 	}
 
+	paramKey = models.ServiceItemParamNameDistanceZipSITOrigin
+	err = s.setLookup(serviceItemCode, paramKey, DistanceZipSITOriginLookup{
+		OriginalAddress: pickupAddress,
```
jacquelineIO, line 350: @mboren  your `OriginalAddress` should be coming from `MTOServiceItem.SITOriginHHGOriginalAddress`.

pkg/payment_request/service_param_value_lookups/distance_zip_sit_origin_lookup_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+	mtoServiceItemDiffZip3 := testdatagen.MakeMTOServiceItem(suite.DB(),
+		testdatagen.Assertions{
+			PickupAddress: originAddress,
+			ReService:     reService,
+			Move:          move,
+			MTOShipment:   mtoShipment,
+			MTOServiceItem: models.MTOServiceItem{
+				SITOriginHHGActualAddressID: &actualOriginDiffZip3Address.ID,
+				SITOriginHHGActualAddress:   &actualOriginDiffZip3Address,
```
jacquelineIO, line 84: You also need to set `MTOServiceItem.SITOriginHHGOriginalAddress`. You won't be using the address from the Shipment.

jacquelineIO (changes requested):
SIT origin service items will have both `SITOriginHHGActualAddress` and `SITOriginHHGOriginalAddress` set. Since linehaul hasn't started yet, we can't ensure that the shipment won't be updated again before the linehaul happens. So we are keeping the information needed on the SIT origin service item. This is different than SIT destination, because linehaul has been completed, and we shouldn't be changing/update the shipment and we are able to use the destination address for distance.

pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go
mboren: Ok, I'll do that

jacquelineIO (commented):
The code looks good, but I think there is confusion on when to use the case statement at the beginning of the file. I left a clarifying comment on the code, to be considered for updating the code.

pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go
```diff
@@ -101,8 +106,27 @@ func ServiceParamLookupInitialize(
 			}
 			sitDestinationFinalAddress = *mtoServiceItem.SITDestinationFinalAddress
 		}
-		fallthrough
-	default:
```
jacquelineIO, line 105: So getting rid of this `fallthrough` is why were having to use this new variable `	savePickupAndDestinationFromShipment := true`.

mboren: Yeah. The boolean feels a bit awkward, but I wasn't sure how else to do this.

pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go
```diff
@@ -101,8 +106,27 @@ func ServiceParamLookupInitialize(
...
+	case models.ReServiceCodeDOASIT, models.ReServiceCodeDOFSIT, models.ReServiceCodeDOPSIT:
+		// load original origin address from service item
+		if mtoServiceItem.SITOriginHHGOriginalAddressID != nil && *mtoServiceItem.SITOriginHHGOriginalAddressID != uuid.Nil {
+			err = db.Load(&mtoServiceItem, "SITOriginHHGOriginalAddress")
+			if err != nil {
+				return nil, err
+			}
+			sitOriginOriginalAddress = *mtoServiceItem.SITOriginHHGOriginalAddress
```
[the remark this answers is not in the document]
mboren: I took out the case that I added, removed the top level variables for SIT addresses, and moved the address loading into the `default` case. This feels a bit more sensible, but I'm still not sure i'm doing the right thing

mboren:
Just realized that I do have a DB connection available in the lookup, so I'll be changing this a bit to use that later today

jacquelineIO (approved):
lgtm

PR #5750, opened 2021-01-26
[MB-4362] Update Filepond

suzubara:
## Description

This PR updates our version of `react-filepond` to the latest major version (7.x) and does some minor clean up of the `FileUpload` component (f.k.a the OrdersUploader).
- Uses props to pass options into `FilePond` component instead of `setOptions`, removes internal `files` state from upload component
- Updates `filepond-plugin-image-preview` plugin and points back to the main distribution instead of our fork
- Use [`React.forwardRef`](https://reactjs.org/docs/forwarding-refs.html) in order to give more control of the FilePond instance to parent components
- Fixes bug with deleting upload from the FilePond UI
- Modifies the Orders UI so that FilePond resets after uploading a file, and the uploads only appear in the table
- Deletes some stale & unused code (`DocumentViewer/DocumentUploader`)

In addition to Upload Orders & Edit Orders in the Customer Onboarding flow, there are a few other places that use another legacy Uploader component (`src/shared/Uploader/index`). Originally I wanted to point those to this newer FileUpload component, but it ended up being more effort than I expected. Instead, I think we should plan on doing this work next time we are working on PPM. There were no breaking changes in the FilePond update, so these interactions should continue to work the same.

## Reviewer Notes

For each test case below, ensure you can interact with the uploader as expected based on the UI. You should be able to upload multiple files and delete files that have already been uploaded, without errors.

Upload test cases are:
- **Service member, create orders (new uploader):** log in as a new service member and upload your orders for the first time during Onboarding (`/orders/upload`) (this is also covered in a Cypress test)
- **Service member, edit orders (new uploader):** after uploading your orders, click "Edit Orders" from the home screen and upload/delete orders
- **Service member, PPM Payment Request (legacy uploader):** for a move that has been submitted and approved by a PPM office user, you should be able to upload expense documents and weight tickets during the Payment Request flow. (this is also covered in a Cypress test). **You can log in as the user ppm@requestingpayment.newflow with a fresh database seed to do this one.**
- **PPM office user (legacy uploader):** view a submitted move and click "Add Document" to upload a document

There is also a minor change in UX to the Upload Orders & Edit Orders, where the uploader UI gets reset after uploading each file. This is because uploaded files will populate in the table above the uploader, and it didn't make sense to show them in two places. I confirmed this change with @amanda5iu 

## Setup

To run the application locally:

```sh
make server_run
make client_run
```

Log in as a service member at http://milmovelocal:3000/
Log in as an office user at http://officelocal:3000/

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4362) for this change

## Screenshots

**Service member, create orders (new uploader):**

https://user-images.githubusercontent.com/2723066/106183155-7f48ca00-6165-11eb-9903-86a9ceabc12d.mp4

 **Service member, edit orders (new uploader):**

image

**Service member, PPM Payment Request (legacy uploader):**

image

image

**PPM office user (legacy uploader):**

image


YanZ777:
Should existing files still be expected to be deleted even after the user hits cancel? 
e.g. User goes to edit orders, deletes files, and then hits cancel.

suzubara:
> Should existing files still be expected to be deleted even after the user hits cancel?
> e.g. User goes to edit orders, deletes files, and then hits cancel.

good catch, I was just looking into that! In order to support this we need to be able to pass an AbortController into the swagger request function we use to make API calls. I would consider this a new bug fix, since hitting cancel on master now doesn't work either. I added a TODO comment here: https://github.com/transcom/mymove/pull/5750/commits/1f1e3628dd28eec59bfea7661b29d773d529bca3#diff-33404192fb1049d9519be45a6f1e820738158881d5ecca4ec62955a7c0346b16R35

YanZ777 (approved):

amanda5iu (commented):
Can we apply default table styling to the list of documents? So bold headers, and a border above each row? Styled headers will help folks tell them apart from the data, and the border will help tie together the info in each row. I also don't think we need the `Delete` column header.

suzubara:
> Can we apply default table styling to the list of documents? So bold headers, and a border above each row? Styled headers will help folks tell them apart from the data, and the border will help tie together the info in each row. I also don't think we need the `Delete` column header.

Yes we can but I don't think that falls in the scope of this story :( do you mind making a JIRA ticket for it (or asking a PM to)?

amanda5iu (approved):
Per our Slack convo, let's address the table styling in a separate story!

PR #5754, opened 2021-01-27
MB-6188-showUserDetail

NamibiaLT:
## Description

When viewing the admin console home page, an admin will be able to click a specific user in the ListUsers view and the app redirects them to a ShowUserDetail page where they can see the specific detail fields of a user.

The details of a customer  have the following fields, (view-only, non-editable) populated for a customer: ID, Login Gov Email, CreatedAt, UpdatedAt and a boolean Active field. The session IDs are also displayed.
## Reviewer Notes
Please ensure that the cypress tests are testing the page correctly, and that the acceptance criteria match what is seen on the page.

## Setup
Make sure that the server is running with `make server_run`, then run the admin client with `make admin_client_run`. 

Log in to the admin api using the local sign-in option. Then sign in using the `create new user` option when redirected to the sign-in page.

Finally, click on Users in the left tab, and then click on one user to be redirected tot eh show page for that one user.

To run the cypress tests:
1. Make sure the server and client are both running: `make server_run` and ` make client_run`
2. Run the cypress test with: `yarn test:e2e`
3. Inspect the cypress test by clicking on `admin/users`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6188) for this change

## Screenshots

This is what the User Show page should look like:
Screen Shot 2021-01-28 at 12 35 37 PM


src/scenes/SystemAdmin/Users/UserShow.jsx
```diff
@@ -1,14 +1,18 @@
 import React from 'react';
-import { Show, SimpleShowLayout, TextField } from 'react-admin';
+import { Show, SimpleShowLayout, TextField, BooleanField, DateField } from 'react-admin';
```
eamahanna, line 3: Can you move this file to `pages/admin/users`? I think this is the where the FE is saving page files.

eamahanna (commented):
Would you also mind updating the labels on the Details pages to have consistent casing?

eamahanna (approved):
Everything works well! Nice job. There are only two small edits I would suggest doing. 
* Make the capitalization the same on each of the labels
Screen Shot 2021-01-29 at 8 41 11 AM

* Move the UsersShow file to pages/admin/users 

These are both pretty minor nice to have changes though.

src/scenes/SystemAdmin/Users/UserShow.jsx
NamibiaLT: Yes, and I had to add an eslint next line disable because when I move the file over I start seeing a no-prop-spreading error that wasn't showing up in the file's previous location.

NamibiaLT (commented):
Left a comment to respond to request to move file to the pages folder.

eamahanna (approved):
LGTM :) thanks for making those changes

PR #5756, opened 2021-01-27
MB-3622 move details shipment mutation

Ryan-Koch:
## Description

Makes it so we're using useMutation for our updating of the MTO Shipment statuses from the Move Details page on the TOO app.

## Setup

```sh
make db_dev_e2e_populate
```

 have been satisfied.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3622) for this change


src/pages/Office/MoveDetails/MoveDetails.jsx
```diff
@@ -59,6 +61,19 @@ const MoveDetails = () => {
...
+  // use mutation calls
+  const [mutateMoveStatus] = useMutation(updateMoveStatus, {
+    onSuccess: (data) => {
+      queryCache.setQueryData([MOVES, data.locator], data);
+    },
+  });
+
+  const [mutuateMTOShipmentStatus] = useMutation(updateMTOShipmentStatus, {
+    onSuccess: (data) => {
+      queryCache.setQueryData([MTO_SHIPMENTS, data.mtoShipmentID], data);
```
donaldthai, line 73: Looks like the `MTO_SHIPMENTS` is using the keys `["mtoShipments","52718722-d4a4-4725-a8db-348c733c98e7",false]` since we're passing in normalize as false. Probably need to create some stories to clean up the normalize param.

duncan-truss: The MTO_SHIPMENTS query returns an array of shipments based on the move id parameter, it's not to fetch a single shipment.

The way this needs to work is we'd need to update this single shipment in that mtoShipments array and then set that in the cache.  And Donald's right about needing a 3rd param set to false if it's to match the hooks.

jenniferlynparsons (commented):
Reviewed and looks good barring others' comments.

src/pages/Office/MoveDetails/MoveDetails.jsx
Ryan-Koch: I think when you provide the setQueryData function an ID for an object in the collection it updates that single object.

There is also work around the normalize parameter, though I think based on the conversation at FE guild yesterday that we are likely going to be pulling the normalizer out from the swagger request generating function as the longer term move.

Ryan-Koch: I went ahead and pushed a tiny change to make the naming more clear based on what I believe should be happening. My understanding is that setQueryData is using the query key to take a peak at the queryCache and do a replacement of the object that is coming back from the mutation, using the ID as the key to know which thing to touch.

duncan-truss: There's no sense in caching what a mutation because the return value is always going to change.

We want to update the cache for the getMTOShipments query that returns an array of shipments because it will still have the stale shipment with the old status. https://github.com/transcom/mymove/blob/13f7b238949089f5f780c93f905787391a438054/src/hooks/queries.js#L268

As an example you can refer to how we handle updating the payment service item after we update it's status https://github.com/transcom/mymove/blob/13f7b238949089f5f780c93f905787391a438054/src/pages/Office/PaymentRequestReview/PaymentRequestReview.jsx#L51

Ryan-Koch: Okay after our chat I think I've made a change that does the replace within the old array and then hand over the array to setQueryData with the moveID method we talked about on zoom @duncan-truss . Does it look like it does the trick to you?

src/pages/Office/MoveDetails/MoveDetails.jsx
```diff
@@ -59,6 +61,20 @@ const MoveDetails = () => {
...
+  // use mutation calls
+  const [mutateMoveStatus] = useMutation(updateMoveStatus, {
+    onSuccess: (data) => {
+      queryCache.setQueryData([MOVES, data.locator], data);
+    },
+  });
+
+  const [mutuateMTOShipmentStatus] = useMutation(updateMTOShipmentStatus, {
+    onSuccess: (updatedMTOShipment) => {
+      mtoShipments[mtoShipments.indexOf({ id: updatedMTOShipment.id })] = updatedMTOShipment;
```
duncan-truss, line 73: I wish the indexOf method was this smart but I think it does strict equality and won't match because of all the extra keys besides `id`.  Maybe try the findIndex method instead?

src/pages/Office/MoveDetails/MoveDetails.jsx
```diff
@@ -59,6 +61,20 @@ const MoveDetails = () => {
...
+      queryCache.setQueryData([MTO_SHIPMENTS, updatedMTOShipment.moveTaskOrderID], mtoShipments);
```
duncan-truss, line 74: This looks better, we just need the 3rd query key param of false because we asked for the non normalized version in the hook so:

```
queryCache.setQueryData([MTO_SHIPMENTS, updatedMTOShipment.moveTaskOrderID, false], mtoShipments);
```

src/pages/Office/MoveDetails/MoveDetails.jsx
Ryan-Koch: oh you're right, findIndex is for primitives and findIndex is for objects. I'll make the switch.

Ryan-Koch: You're right. Switched over to findIndex.

src/pages/Office/MoveDetails/MoveDetails.jsx
Ryan-Koch: added

duncan-truss (approved):

src/pages/Office/MoveDetails/MoveDetails.jsx
```diff
@@ -68,6 +68,13 @@ const MoveDetails = () => {
     },
   });
 
+  const [mutuateMTOShipmentStatus] = useMutation(updateMTOShipmentStatus, {
```
monfresh, line 71: Is this supposed to say `mutuate` or is that a typo?

Ryan-Koch: Ack that was a typo. I'm going to quickly open a PR to fix this.

Ryan-Koch: Fixed it over here: https://github.com/transcom/mymove/pull/5773

PR #5762, opened 2021-01-28
Fix flakey edi generation test

gidjin:
## Description

Occasionally the EDI Generator unit test would fail due to the current time changing from one minute to the next during the test run. This PR makes it use the clock mock module so that the time remains constant during the run. It also updates us to use the maintained and active clock mock module vs the one that is archived and been inactive for a couple years.

## Reviewer Notes

Is there a better way to mock time in our tests to avoid failures of this sort?

## Setup

Make sure server test is still passing.

```sh
make server_tests
```

## References

* [this slack thread](https://ustcdp3.slack.com/archives/CP6F568DC/p1611700391119800) explains more about the approach used.


pkg/services/invoice/ghc_payment_request_invoice_generator_test.go
```diff
@@ -49,9 +50,17 @@ const testDateFormat = "20060102"
...
 func (suite *GHCInvoiceSuite) TestAllGenerateEdi() {
-	currentTime := time.Now()
-	generator := NewGHCPaymentRequestInvoiceGenerator(suite.DB(), suite.icnSequencer)
+	mockClock := clock.NewMock()
+	currentTime := mockClock.Now()
+	generator := &ghcPaymentRequestInvoiceGenerator{
+		db:           suite.DB(),
+		icnSequencer: suite.icnSequencer,
+		clock:        mockClock,
+	}
```
reggieriser, line 63: Related to my other comment, it looks like you had to change the code from using the constructor since a clock isn't passed in there.  If the constructor changed in the future to do something slightly different (validate one of the params, for instance), we'd miss it by directly instantiating the struct.

reggieriser (commented):
Thanks for swapping out the old clock and fixing the indeterminate tests!  Added a couple of comments about the constructor being bypassed in the new code.

pkg/services/invoice/ghc_payment_request_invoice_generator_test.go
monfresh: I agree. It's best to be able to inject our dependencies.

gidjin: I'll reply here to both comments. I agree with you two as well. I wasn't sure there was time for the larger refactor, but given the push back I think that's what I should have done. I'll update the PR

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -19,13 +19,15 @@ import (
...
 	icnSequencer sequence.Sequencer
+	clock        clock.Clock
 }
 
 // NewGHCPaymentRequestInvoiceGenerator returns an implementation of the GHCPaymentRequestInvoiceGenerator interface
 func NewGHCPaymentRequestInvoiceGenerator(db *pop.Connection, icnSequencer sequence.Sequencer) services.GHCPaymentRequestInvoiceGenerator {
 	return &ghcPaymentRequestInvoiceGenerator{
 		db:           db,
 		icnSequencer: icnSequencer,
+		clock:        clock.New(),
```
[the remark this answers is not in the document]
gidjin: See below

pkg/services/invoice/ghc_payment_request_invoice_generator_test.go
```diff
@@ -49,9 +50,13 @@ const testDateFormat = "20060102"
 const testISADateFormat = "060102"
 const testTimeFormat = "1504"
 
+func (suite *GHCInvoiceSuite) TestGeneratorConstructor() {
+}
```
monfresh, line 54: Did you mean to fill out this test?

reggieriser (approved):
Thanks for doing the additional refactoring.  This looks good to me with the exception of the empty test @monfresh already pointed out.  Going to go ahead and approve though as I assume you can address that easily. :shipit:

pkg/services/invoice/ghc_payment_request_invoice_generator_test.go
gidjin: No this was an oversight in my clean up. Will take care of it.

PR #5777, opened 2021-01-29
MB-6172 Filter hidden moves in getMoveHandler and getMoveTaskOrderHandler

abbyoung:
## Description

* Adds search parameters to `getMoveHandler` and `getMoveTaskOrderHandler` to support show/hide moves.
* Updates functionality in service objects to query based on the search params
* Updates and adds tests

## Setup

`go test ./pkg/handlers/ghcapi/ --testify.m "TestGetMoveHandler"`
`go test ./pkg/services/move --testify.m "TestMoveFetcher"`
`go test ./pkg/services/move_task_order --testify.m "TestMoveTaskOrderFetcher"`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6172) for this change


pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -163,7 +166,10 @@ func (o *moveTaskOrderUpdater) UpdatePostCounselingInfo(moveTaskOrderID uuid.UUI
 
 // ShowHide changes the value in the "Show" field for a Move. This can be either True or False and indicates if the move has been deactivated or not.
 func (o *moveTaskOrderUpdater) ShowHide(moveID uuid.UUID, show *bool) (*models.Move, error) {
```
duncan-truss, line 168: Is there any reason we can't make the show param a non-pointer boolean and in the admin api handler fail earlier if you still want to accept inputs where that isn't set?  I think that's better than waiting until the database call to error out.  I created a Jira story as well to make that show field not-null like it is in the sql schema.

duncan-truss (approved):
LGTM!

pkg/services/move_task_order/move_task_order_updater.go
sandy-wright: That reasoning makes sense to me, sounds like it would be a good idea to change it. @abbyoung - you can implement that fix if you want, or we can add a small story for it (I wrote that originally, I'm the culprit, I killed the butler!)

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -163,7 +166,10 @@ func (o *moveTaskOrderUpdater) UpdatePostCounselingInfo(moveTaskOrderID uuid.UUI
...
-	move, err := o.FetchMoveTaskOrder(moveID)
+	searchParams := services.FetchMoveTaskOrderParams{
+		IncludeHidden: true, // We need to search every move to change its status
+	}
```
sandy-wright, line 171: 💯 ✨  I would have forgotten this

pkg/handlers/internalapi/moves.go
```diff
@@ -89,6 +89,7 @@ func (h ShowMoveHandler) Handle(params moveop.ShowMoveParams) middleware.Respond
 	/* #nosec UUID is pattern matched by swagger which checks the format */
 	moveID, _ := uuid.FromString(params.MoveID.String())
 
+	//#TODO: Review this handler and changes required with Sandy
```
sandy-wright, line 92: ah yes, this is in the next PR 👍

sandy-wright (approved):
LGTM ✨ . One comment about including hidden moves in the Support API, but otherwise looks great (and I really do think the `ShowHide` update would be fine as a wee ticket for a later sprint, no pressure to do that unless you really wanna)

pkg/handlers/ghcapi/move.go
```diff
@@ -25,7 +25,11 @@ func (h GetMoveHandler) Handle(params moveop.GetMoveParams) middleware.Responder
 		return moveop.NewGetMoveBadRequest()
 	}
 
-	move, err := h.FetchMove(locator)
+	searchParams := services.MoveFetcherParams{
+		IncludeHidden: false,
+	}
```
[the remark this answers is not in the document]
abbyoung: oooh good point. so instead of setting this, would i just pass in nil?

pkg/handlers/internalapi/moves.go
abbyoung: boop i need to remove this! thanks for flagging 😄

pkg/services/move_task_order/move_task_order_updater.go
abbyoung: let's add a small story 👍

PR #5778, opened 2021-01-29
MB-6172 Add show/hide param to showMoveHandler

abbyoung:
## Description

* This PR changes the `showMoveHandler` to only return visible moves.
* Updates tests

## Reviewer Notes

~`models.FetchMove` is used in many files, so want to make sure that other teams get 👀 on this and make sure it looks okay.~

## Setup

You can hit the `showMoveHandler` with Postman using the endpoint
`http://officelocal:3000/internal/moves/<uuid>`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6172) for this change

### DIFF
--- a/pkg/models/move.go
+++ b/pkg/models/move.go
@@ -226,7 +226,7 @@ func FetchMove(db *pop.Connection, session *auth.Session, id uuid.UUID) (*Move, 
 		"SignedCertifications",
 		"Orders",
 		"MoveDocuments.Document",
-	).Find(&move, id)
+	).Where("show = TRUE").Find(&move, id)
 
 	if err != nil {
 		if errors.Cause(err).Error() == RecordNotFoundErrorString {
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "@babel/core": "~7.12.10",
     "@dump247/storybook-state": "^1.6.1",
     "@stoplight/spectral": "^5.8.0",
-    "@storybook/addon-a11y": "^6.1.14",
+    "@storybook/addon-a11y": "^6.1.15",
     "@storybook/addon-essentials": "^6.1.14",
     "@storybook/addon-knobs": "^6.1.14",
     "@storybook/addon-links": "^6.1.14",
--- a/pkg/models/move_test.go
+++ b/pkg/models/move_test.go
@@ -110,6 +110,15 @@ func (suite *ModelSuite) TestFetchMove() {
 	session.ServiceMemberID = order2.ServiceMemberID
 	fetchedMove, err = FetchMove(suite.DB(), session, move.ID)
 	suite.Equal(ErrFetchForbidden, err, "Expected to get a Forbidden back.")
+
+	suite.T().Run("Hidden move is not returned", func(t *testing.T) {
+		// Create a hidden move
+		hiddenMove := testdatagen.MakeHiddenHHGMoveWithShipment(suite.DB(), testdatagen.Assertions{})
+
+		// Attempt to fetch this move. We should receive an error.
+		_, err := FetchMove(suite.DB(), session, hiddenMove.ID)
+		suite.Equal(ErrFetchNotFound, err, "Expected to get FetchNotFound.")
+	})
 }
 
 func (suite *ModelSuite) TestMoveCancellationWithReason() {
--- a/yarn.lock
+++ b/yarn.lock
@@ -2294,19 +2294,19 @@
     "@stoplight/yaml-ast-parser" "0.0.48"
     tslib "^1.12.0"
 
-"@storybook/addon-a11y@^6.1.14":
-  version "6.1.14"
-  resolved "https://registry.yarnpkg.com/@storybook/addon-a11y/-/addon-a11y-6.1.14.tgz#17cc6864b1427e4a432727ceaa9cc72354977e66"
-  integrity sha512-FAiBMWvsLaBzfMRtlWnXeI/PeOs1akhjk3D+6a96poY0SCIuQ0RPfGqQhO00mLGBB3+DOsKzCf16y34/pdOUfQ==
+"@storybook/addon-a11y@^6.1.15":
+  version "6.1.15"
+  resolved "https://registry.yarnpkg.com/@storybook/addon-a11y/-/addon-a11y-6.1.15.tgz#771438627cab24ab948bf7117af09da39a382241"
+  integrity sha512-iD6aXdX4arhRPZVSP/6Tpl1Si7WKb0pISORpi1lt1BNwoRxIauxRCVypEZPGM4R5SDEthU0SiBAHluMmORtd+w==
   dependencies:
-    "@storybook/addons" "6.1.14"
-    "@storybook/api" "6.1.14"
-    "@storybook/channels" "6.1.14"
-    "@storybook/client-api" "6.1.14"
-    "@storybook/client-logger" "6.1.14"
-    "@storybook/components" "6.1.14"
-    "@storybook/core-events" "6.1.14"
-    "@storybook/theming" "6.1.14"
+    "@storybook/addons" "6.1.15"
+    "@storybook/api" "6.1.15"
+    "@storybook/channels" "6.1.15"
+    "@storybook/client-api" "6.1.15"
+    "@storybook/client-logger" "6.1.15"
+    "@storybook/components" "6.1.15"
+    "@storybook/core-events" "6.1.15"
+    "@storybook/theming" "6.1.15"
     axe-core "^4.0.1"
     core-js "^3.0.1"
     global "^4.3.2"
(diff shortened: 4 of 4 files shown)


pkg/models/move.go
```diff
@@ -214,19 +214,31 @@ func (m *Move) Cancel(reason string) error {
...
+}
+
+// FetchMoveParams passes search params to filter out hidden moves
+type FetchMoveParams struct {
+	IncludeHidden bool
 }
 
 // FetchMove fetches and validates a Move for this User
-func FetchMove(db *pop.Connection, session *auth.Session, id uuid.UUID) (*Move, error) {
+func FetchMove(db *pop.Connection, session *auth.Session, id uuid.UUID, searchParams *FetchMoveParams) (*Move, error) {
```
sandy-wright, line 226: This is just so weird to me still, but your changes look good

sandy-wright (approved):
Looks good and works great, even if it makes me uncomfortable!

pkg/models/move.go
abbyoung: definitely think it's worth revisiting in the future to retrofit this with a service object!

donaldthai: Have you considered using `variadic params`? It'll clean up all the `nil`'s that's being passed. https://gobyexample.com/variadic-functions

donaldthai (approved):
Works 👍

pkg/models/move.go
abbyoung: Oh that's interesting! It'd be nice to not have to pass in nil, so I looked into this and struggled on how to implement it that would simplify the code. Since they're not all the same type of args, I'd have to either pass in an interface or struct, right? I'd have to add a good deal of error checking to make sure all the correct args were passed in, unless I'm missing an easier way to implement 🙈 I'm not sure if it's worth it to save the nil arg, when it'd made the FetchMove function much more complicated.

I do think this part of the codebase needs a revision anyway because it doesn't match our current patterns, so I'd love to try and simplify the args  on the next pass.

monfresh: Adding a boolean argument to a function is a common code smell: https://www.informit.com/articles/article.aspx?p=1392524
You might also see it referred to as a "Control Couple" because the boolean argument is controlling the path the function takes. In this case, it's controlling whether or not to show hidden moves.

Instead of adding a boolean argument, it's better to have two separate methods, such as `FetchVisibleMove` and `FetchMove`. And then in the handler, you would call one or the other based on the params. This makes it much easier to read and understand, as opposed to having to look up the `FetchMove` function to figure out what the last `nil` argument represents.

The other thing that concerns me is that the showMoveHandler now allows anyone to view hidden moves. Shouldn't that functionality be restricted to admin users?

monfresh (changes requested):
I left some refactoring comments and a security question.

pkg/models/move.go
monfresh: To clarify, I realize this isn't literally a boolean argument, but the same principle applies. We want to adhere to the Single Responsibility Principle as much as possible. By having two separate, well-named functions, we separate the responsibilities into each one. One returns the move as is, and the other only returns it if it is visible. It also adheres to the "Tell Don't Ask" principle. Instead of a single function asking whether or not it should apply the `Show=true` filter, we tell it explicitly which result we want. If we want only visible moves, we call the `FetchVisibleMove` function. Otherwise, we call the `FetchMove` function. Each function only performs one thing.

sandy-wright: The default behavior with nil params is to hide hidden moves, so Abigail has changed this handler to _not_ show hidden moves. You can test this with the endpoint she mentioned in the description to verify that you get an error when you attempt to view a hidden move.

The pattern you're describing is definitely preferable, but I don't know if applying it to this function makes sense. We have 3 different "FetchMove" functions in the codebase right now that all do variations of the same thing, with no obvious need for differentiation. This one is by far the most obvious outlier because it is a function on the model and not a service object, which doesn't make sense for a fetch task. It seems like a waste of time to build on top of this function when, really, we should be removing it and consolidating all of our move fetching into one service object. Then that service object should take advantage of the pattern you mentioned.

Consolidating these functions should probably be on the radar of the tech review team, in fact! Redundant code would be really awesome to eliminate in this project

monfresh: Yes, I understand the default behavior and what has changed. I agree that we should be moving fetching to a service object. Ideally, we would take care of that in this PR. It doesn't have to be a consolidation of all the "FetchMove" functions. Any incremental improvement is worth it. I like Sarah Mei's analogy from her great [Livable Code](https://www.youtube.com/watch?v=lI77oMKr5EY) talk: one book on the shelf and five in a pile is better than six in a pile. If we wait until all 6 books are moved to the shelf at once, it might never happen, or be a while before it happens.

If we don't have time in this PR to create a new service object, then adding a new function to the model is a better marginal gain than adding an argument to the existing function because it lays the groundwork for the service object refactor, and it makes the code more readable while we're waiting for the refactor to happen.

Either way, we have to modify all those other handlers, so why not do it in a way that improves the code? Here's what I have in mind:

- Leave the current `FetchMove` function as is so that it returns the move, regardless of whether or not it's hidden
- Create a new `FetchVisibleMove` that only returns the move if its `show` field is `true`
- Rename `FetchMove` in all the places in this PR where we passed the `nil` argument to `FetchVisibleMove`
- In `showMoveHandler`, if the `IncludeHidden`params is `true`, call `FetchMove`, otherwise call `FetchVisibleMove`.

monfresh: Reading through the changes again, perhaps I'm misunderstanding something. It looks like every change here is expecting hidden moves to *not* be shown, and I don't see any changes that allow someone to pass in the `IncludeHidden` params as `true`. If it's not possible to request hidden moves via the `FetchMove` function in the move model, why do we need the params? Why can't we modify the `FetchMove` function so that it always excludes hidden moves?

At first, I thought the showMoveHandler was reading the `IncludeHidden` params, but now I see that it hardcodes it to `false`. I was confused by that because that was the only place where the params is used, and everywhere else we are passing in `nil`. It's not clear to me what the difference is.

sandy-wright: That's true, this function could be changed to only return visible moves. For the other fetch move service objects, we left it an option because the Support API and the Admin API can use them to see hidden moves. This function is only ever called by the internal API however, and that should never be working with hidden moves. So @abbyoung could just change the core functionality of `FetchMove` (and that would be the simplest option for right now)

sandy-wright: I also don't necessarily agree with the pattern you described, although I think it's worth further discussion down the road! But I think creating two functions that require repeated `if` statements to determine when they're called is an anti-pattern as well that could lead to further duplicated code. Explicitly named functions are a very good thing, but the pattern should be more like (in my opinion):

* Create a FetchMove function that takes in various parameters (in this case, the service member ID, the session, and include hidden - although I also think it's weird that the session is passed in here)
* Create FetchVisibleMove that calls FetchMove with `includeHidden=false` always and simplifies the function signature. There could be another called `FetchHiddenMove` with the opposite setting.
* Wherever the caller _always_ needs visible moves, use `FetchVisibleMove`
* _Wherever there's an option_, the caller should use the base `FetchMove` and pass in the parameter. This function should contain the logic to determine what to do, and it would be redundant to have to repeat that logic in every Admin API or Support API endpoint that might want to toggle this functionality.

This definitely isn't cut and dry - I think there's a lot of room for discussion with this. Some of the validators in the service objects are starting to do something similar, and it would be cool if we could talk about what pattern we want to stick with

abbyoung: Thank you for all this discussion and thought -- I'm going to work on the change to have FetchMove only return visible moves, and then we can remove all the nils, too.

monfresh: Yeah, I agree that if we find ourselves with repeated conditionals, it's not ideal. One way to reduce conditionals is to have separate endpoints: one for visible moves only, and one that includes hidden moves.

Another pattern I've used in Ruby is metaprogramming. For example, you might have 2 classes that implement the same method:
```ruby
class VisibleMove
  def get(id)
    Move.where(show: true).find(id)
  end
end

class Move
  def get(id)
    Move.find(id)
  end
end
```
And you decide which class needs to be called based on some parameter whose value would be one of the two class names. For example, the parameter could be called `move_kind` or `move_class`, and its value would either be `move` or `visible_move`, and the proper class would be called like this:

```ruby
move_class = params[:move_class]
move_class.classify.constantize.new.get(id)
```
Sandi Metz has a nice post about this pattern: https://sandimetz.com/blog/2014/9/9/shape-at-the-bottom-of-all-things

monfresh (approved):
Nice and simple! I like it.

pkg/models/move.go
sandy-wright: That is really cool - I wish I'd done more rails development! You can do similar things in django, but it's all just a smidge different in practice.

PR #5779, opened 2021-01-29
Fix firstAttemptedAt model due to nullable db field

shimonacarvalho:
## Description

@carterjones reported that the model did not match the db. Model-vet concurred. 
The db says `firstAttemptedAt` is nullable, but the model had a non-pointer type. 
Fixed by switching model to pointer and cleaning tests. 

## Setup

You can run the tests located here
```
go test ./cmd/webhook-client/webhook -v
```

## References

* [Retry mechanism is documented here](https://docs.google.com/document/d/1tq2jWvuLXEv30Bab2S-nSj9V1bAzLrPk7DZo3xLX9-o/edit#heading=h.fd69p5h0bm30)


LeDeep (approved):
good catch! updates to tests look good and are passing  🚀

PR #5792, opened 2021-01-29
[MB-3071] ADR to use the snapshot process on experimental to cleanup loadtesting

shimonacarvalho:
## Description

Added ADR about loadtesting deletion.


LeDeep (approved):
write up looks good.

PR #5799, opened 2021-02-01
MB-4397: Rm compare-secure-migrations cmd

kilbergr:
## Description

https://dp3.atlassian.net/browse/MB-4397

Having consulted with some app-engs (including the illustrious Erin Stanfill ), have determined that this appears to be a low-use tool used (perhaps exclusively) by Chris Gilmer. That being the case it’s not worth it to move to govcloud and instead we should tear out. 
Conversation in question here: https://ustcdp3.slack.com/archives/CP6PTUPQF/p1612194454157000

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4397) for this change


travelar (approved):
🔥 Burn it down.

PR #5788, opened 2021-01-29
MB-4729 Fix Mil Move Header Bugs

YanZ777:
## Description

The `office_user` wasn't always being populated from `ShowLoggedInUser` from the internal API handler, so the user name would not show up in the `<MilMoveHeader>`. I changed the office app to instead pull data from the user instead. The last name needed to be added to the return since it only had the first name originally. 
Also added in flexbox-shrink: 0 to the `<MilMoveHeader>` so that the header wouldn't disappear and have weird display issues.

## Setup

Add any steps or code to run in this section to help others prepare to run your code:

```sh
make office_client_run
make server_run
```

## Code Review Verification Steps

* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4729) for this change

## Screenshots

If this PR makes visible UI changes, an image of the finished UI can help reviewers and casual
observers understand the context of the changes. A before image is optional and
can be included at the submitter's discretion.

Consider using an animated image to show an entire workflow instead of using multiple images. You may want to use GIPHY CAPTURE for this! 📸

_Please frame screenshots to show enough useful context but also highlight the affected regions._

image

image

PR #5803, opened 2021-02-01
Only set ServiceMemberID in session if app is mil 

monfresh:
## Description

Recently, we added a header on the Office App to display the Office
User's name and GBLOC. When this was first implemented, the
information was only displaying for some users, but not others. This
information comes from the payload of the `/users/logged_in` endpoint.

The reason why the header wasn't working for some people, namely those
who existed as both service members and office users, is because the
`ShowLoggedInUserHandler` was only returning the Office User info if
`!session.IsServiceMember()`. Unfortunately, this only worked if the
current user had never signed into the service member application.
This is because when an office user signs in (and if they also have a
service member account), we populated the session with their Service
Member ID, which makes `session.IsServiceMember()` return `true`.

So, one reasonable solution, and the simplest one, is to only set the
ServiceMember ID in the session if the current app is the `mil` app.

This essentially reverts #5788 

## Reviewer Notes

This fix should work with existing users. If you already have a user that has both a service member and office user account, sign into the office app, and you should see the banner with your GBLOC and first and last names.

You should also make sure that brand new accounts still work, as well as signing in via devlocal-auth. 

To test with new accounts, follow the steps below:

1. Sign in via login.gov to milmovelocal:3000
- [ ] Verify that you see the OCONUS/CONUS form
2. In a separate terminal tab or window, run `make admin_client_run`. This will launch the admin app
3. Click on `Local Sign In` in the top right
4. Click on `Login` next to any admin user listed
5. Click on Office users in the left sidebar if it's not already selected
6. Click on `+ CREATE` in the top right
7. Fill out your details, making sure to use the same email as the one you used in step 1
8. Select both the `Transportation Ordering Officer` and `Transportation Invoicing Officer` roles
9. Save the user
10. Sign out of the admin app
11. In a separate terminal tab or window, run `make office_client_run`. This will launch the office app
12. Sign in via login.gov
Verify that you see the header with your name and GBLOC

## Setup

```sh
make server_run
make client_run # in a separate tab
```

### DIFF
--- a/pkg/handlers/internalapi/users.go
+++ b/pkg/handlers/internalapi/users.go
@@ -55,7 +55,6 @@ func (h ShowLoggedInUserHandler) Handle(params userop.ShowLoggedInUserParams) mi
 		userPayload := internalmessages.LoggedInUserPayload{
 			ID:         handlers.FmtUUID(session.UserID),
 			FirstName:  session.FirstName,
-			LastName:   session.LastName,
 			Email:      session.Email,
 			OfficeUser: payloads.OfficeUser(&officeUser),
 		}
--- a/src/pages/Office/index.jsx
+++ b/src/pages/Office/index.jsx
@@ -28,7 +28,7 @@ import { ConnectedSelectApplication } from 'pages/SelectApplication/SelectApplic
 import { roleTypes } from 'constants/userRoles';
 import LoadingPlaceholder from 'shared/LoadingPlaceholder';
 import { withContext } from 'shared/AppContext';
-import { LocationShape, UserRolesShape, UserShape } from 'types/index';
+import { LocationShape, UserRolesShape, OfficeUserInfoShape } from 'types/index';
 import { LogoutUser } from 'utils/api';
 
 // Lazy load these dependencies (they correspond to unique routes & only need to be loaded when that URL is accessed)
@@ -80,7 +80,7 @@ export class OfficeApp extends Component {
       activeRole,
       userIsLoggedIn,
       userRoles,
-      user,
+      officeUser,
       location: { pathname },
       logOut,
     } = this.props;
@@ -158,16 +158,16 @@ export class OfficeApp extends Component {
               <QueueHeader />
             ) : (
               <MilmoveHeader
-                lastName={user.last_name}
-                firstName={user.first_name}
+                lastName={officeUser.last_name}
+                firstName={officeUser.first_name}
                 handleLogout={() => {
                   logOut();
                   LogoutUser();
                 }}
               >
-                {user?.office_user?.transportation_office && (
+                {officeUser.transportation_office && (
                   <Link to="/">
-                    {user?.office_user?.transportation_office.gbloc} {queueText}
+                    {officeUser.transportation_office.gbloc} {queueText}
                   </Link>
                 )}
               </MilmoveHeader>
@@ -236,7 +236,7 @@ OfficeApp.propTypes = {
   userIsLoggedIn: PropTypes.bool,
   userRoles: UserRolesShape,
   activeRole: PropTypes.string,
-  user: UserShape,
+  officeUser: OfficeUserInfoShape,
   logOut: PropTypes.func.isRequired,
 };
 
@@ -245,7 +245,7 @@ OfficeApp.defaultProps = {
   userIsLoggedIn: false,
   userRoles: [],
   activeRole: null,
-  user: {},
+  officeUser: {},
 };
 
 const mapStateToProps = (state) => {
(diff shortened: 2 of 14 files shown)


pkg/handlers/internalapi/users.go
```diff
@@ -41,7 +41,7 @@ func (h ShowLoggedInUserHandler) Handle(params userop.ShowLoggedInUserParams) mi
 
 	session, logger := h.SessionAndLoggerFromRequest(params.HTTPRequest)
 
-	if !session.IsServiceMember() {
+	if session.IsOfficeUser() || session.IsAdminUser() {
```
YanZ777, line 44: If you're reverting the changes to index.jsx, should last name still be passed on line 58 (one of the the changes that I added)? I see no harm in leaving it, but at the same time, don't want to unnecessarily bloat things.

pkg/handlers/internalapi/users_test.go
```diff
@@ -13,26 +13,6 @@ import (
...
-	builder := officeuser.NewOfficeUserFetcherPop(suite.DB())
-
-	handler := ShowLoggedInUserHandler{handlers.NewHandlerContext(suite.DB(), suite.TestLogger()), builder}
-
-	response := handler.Handle(params)
-
-	okResponse, ok := response.(*userop.ShowLoggedInUserOK)
-	suite.True(ok)
-	suite.Equal(okResponse.Payload.ID.String(), unknownUser.ID.String())
-}
```
monfresh, line 34: This is an invalid test. Once a user is signed in, they can't be unknown. If we've let in an unknown user who is neither a service member, or office user, or admin user, then we've got serious problems.

src/pages/Office/index.jsx
```diff
@@ -28,7 +28,7 @@ import { ConnectedSelectApplication } from 'pages/SelectApplication/SelectApplic
 import { roleTypes } from 'constants/userRoles';
 import LoadingPlaceholder from 'shared/LoadingPlaceholder';
 import { withContext } from 'shared/AppContext';
-import { LocationShape, UserRolesShape, UserShape } from 'types/index';
```
YanZ777, line 31: If UserShape is no longer being used, maybe it should be removed from `src/types/user.js` and `src/types/index.js` as well.

src/pages/Office/index.jsx
```diff
@@ -255,7 +255,7 @@ const mapStateToProps = (state) => {
     userIsLoggedIn: user.isLoggedIn,
     userRoles: user.roles,
     activeRole: state.auth.activeRole,
-    user,
+    officeUser: user.office_user,
```
YanZ777, line 258: Should office user be used here? 
I think in the object I saw 
`user: { 
first_name: 'John'
office_user: {first_name: 'John', last_name: 'Doe'}, 
...
}`
 There seemed to be some information being duplicated between `user` and `office_user`, so I'm not quite sure which one should be used.

kimallen (approved):
I tested it out, and seemed to work fine for me.  Thanks for getting this in so quick!
Approved given the fixes to tests passes, and Yan's comments are addressed.

src/pages/Office/index.jsx
monfresh: The gbloc comes from the office user object.

pkg/handlers/internalapi/users.go
monfresh: Good question. I suppose I could revert all the changes, and if we ever need more info later, we can add it.

src/pages/Office/index.jsx
monfresh: Good point. I'll remove.

monfresh:
My change is breaking the initial sign in for new accounts because the `ServiceMemberID` doesn't get set in the session on the very first sign in, and the `/users/logged_in` endpoint is looking up the service member by its `session.ServiceMemberID`. I'll have to think more about this.

monfresh:
Ah! It's because we don't create the service member until after the authentication happens. I wonder if we could create it right after we create the user while we're still in the auth function.

monfresh:
I don't see any reason why we wouldn't create the service member right after creating the user. That's what used to happen anyways after the user was redirected. When the user lands on the home page, `initializeOnboarding` (from `onboarding.js`) gets called, and it checks if the user doesn't have any service members yet, and if not, it creates it.

In my latest commit, I moved that creation earlier, so that `session.ServiceMemberID` would be available to the `/users/logged_in` endpoint.

YanZ777 (approved):
Thanks for fixing this!

monfresh:
@YanZ777 @kimallen Could you please review this again? I added some tests and fixed devlocal as well.

YanZ777 (approved):
The changes you've made look good to me! 
Also, good idea adding Cypress tests!

PR #5821, opened 2021-02-03
[MB-5392] Fix CAT III eslint errors

Ronolibert:
## Description

Fixed most instances of CAT III eslint errors

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5392) for this change

### DIFF
--- a/src/components/DataPoint/index.jsx
+++ b/src/components/DataPoint/index.jsx
@@ -8,16 +8,16 @@ const DataPoint = ({ columnHeaders, dataRow, icon, custClass }) => (
   <table className={classnames(styles.dataPoint, 'table--data-point', custClass)}>
     <thead className="table--small">
       <tr>
-        {columnHeaders.map((header, i) => (
-          // eslint-disable-next-line react/no-array-index-key
-          <th key={i}>{header}</th>
+        {columnHeaders.map((header) => (
+          <th key={header}>{header}</th>
         ))}
       </tr>
     </thead>
     <tbody>
       <tr>
+        {/* eslint-disable react/no-array-index-key */}
+        {/* no unique identifier that can be used as a key, cell values can be duplicates (e.g. Dates) */}
         {dataRow.map((cell, i) => (
-          // eslint-disable-next-line react/no-array-index-key
           <td key={i}>
             <div className={classnames({ [`${styles.iconCellContainer}`]: !!icon && i === 0 })}>
               <span>{cell}</span>
@@ -25,6 +25,7 @@ const DataPoint = ({ columnHeaders, dataRow, icon, custClass }) => (
             </div>
           </td>
         ))}
+        {/* eslint-enable react/no-array-index-key */}
       </tr>
     </tbody>
   </table>
--- a/src/components/DataPoint/index.jsx
+++ b/src/components/DataPoint/index.jsx
@@ -18,7 +18,7 @@ const DataPoint = ({ columnHeaders, dataRow, icon, custClass }) => (
         {/*
           // RA Summary: eslint:react/no-array-index-key
           // RA: Using the index as an element key in cases where the array is reordered will result in unnecessary renders.
-          // RA: The source data is unstructured, with a potential for duplicate values.
+          // RA: The source data is unstructured, with a potential for duplicate values amongst siblings.
           // RA: A reorder function is not implemented for this array.
           // RA Developer Status: Mitigated
           // RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
--- a/src/components/DataPoint/index.jsx
+++ b/src/components/DataPoint/index.jsx
@@ -15,6 +15,16 @@ const DataPoint = ({ columnHeaders, dataRow, icon, custClass }) => (
     </thead>
     <tbody>
       <tr>
+        {/*
+          // RA Summary: eslint:react/no-array-index-key - Do not use array index in keys
+          // RA: Index of array item is being used as a part of key prop for dom element
+          // RA: The key prop is used by React to identify which items have changed, are added, or are removed and should be stable.
+          // RA: There are no other unique identifiers that can be used
+          // RA Developer Status: Mitigated
+          // RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+          // RA Validator: jneuner@mitre.org
+          // RA Modified Severity:
+        */}
         {/* eslint-disable react/no-array-index-key */}
         {/* no unique identifier that can be used as a key, cell values can be duplicates (e.g. Dates) */}
         {dataRow.map((cell, i) => (
--- a/src/components/DataPoint/index.jsx
+++ b/src/components/DataPoint/index.jsx
@@ -16,10 +16,10 @@ const DataPoint = ({ columnHeaders, dataRow, icon, custClass }) => (
     <tbody>
       <tr>
         {/*
-          // RA Summary: eslint:react/no-array-index-key - Using the index as an element key in cases where the array is reordered will result in unnecessary renders.
-          // RA: Index of array item is being used as a part of key prop for dom element
-          // RA: The key prop is used by React to identify which items have changed, are added, or are removed and should be stable.
-          // RA: There are no other unique identifiers that can be used
+          // RA Summary: eslint:react/no-array-index-key
+          // RA: Using the index as an element key in cases where the array is reordered will result in unnecessary renders.
+          // RA: The source data is unstructured, with a potential for duplicate values.
+          // RA: A reorder function is not implemented for this array.
           // RA Developer Status: Mitigated
           // RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
           // RA Validator: jneuner@mitre.org
--- a/src/components/DataPoint/index.jsx
+++ b/src/components/DataPoint/index.jsx
@@ -16,7 +16,7 @@ const DataPoint = ({ columnHeaders, dataRow, icon, custClass }) => (
     <tbody>
       <tr>
         {/*
-          // RA Summary: eslint:react/no-array-index-key - Do not use array index in keys
+          // RA Summary: eslint:react/no-array-index-key - Using the index as an element key in cases where the array is reordered will result in unnecessary renders.
           // RA: Index of array item is being used as a part of key prop for dom element
           // RA: The key prop is used by React to identify which items have changed, are added, or are removed and should be stable.
           // RA: There are no other unique identifiers that can be used
--- a/src/components/DataPoint/index.jsx
+++ b/src/components/DataPoint/index.jsx
@@ -21,9 +21,8 @@ const DataPoint = ({ columnHeaders, dataRow, icon, custClass }) => (
           // RA: The source data is unstructured, with a potential for duplicate values amongst siblings.
           // RA: A reorder function is not implemented for this array.
           // RA Developer Status: Mitigated
-          // RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
-          // RA Validator: jneuner@mitre.org
-          // RA Modified Severity:
+          // RA Validator Status: Mitigated
+          // RA Modified Severity: N/A
         */}
         {/* eslint-disable react/no-array-index-key */}
         {/* no unique identifier that can be used as a key, cell values can be duplicates (e.g. Dates) */}
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -32,6 +32,14 @@ const Table = ({
       <table {...getTableProps()}>
         <thead>
           {headerGroups.map((headerGroup, hgIndex) => (
+            // RA Summary: eslint:react/no-array-index-key - Do not use array index in keys
+            // RA: Index of array item is being used as a part of key prop for dom element
+            // RA: The key prop is used by React to identify which items have changed, are added, or are removed and should be stable.
+            // RA: There are no other unique identifiers that can be used
+            // RA Developer Status: Mitigated
+            // RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+            // RA Validator: jneuner@mitre.org
+            // RA Modified Severity:
             /* Does not have any unique properties that can be used as key */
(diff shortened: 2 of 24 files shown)


src/components/DataPoint/index.jsx
```diff
@@ -8,23 +8,24 @@ const DataPoint = ({ columnHeaders, dataRow, icon, custClass }) => (
...
-          // eslint-disable-next-line react/no-array-index-key
-          <th key={i}>{header}</th>
+        {columnHeaders.map((header) => (
+          <th key={header}>{header}</th>
         ))}
       </tr>
     </thead>
     <tbody>
       <tr>
+        {/* eslint-disable react/no-array-index-key */}
```
kimallen, line 18: In order to keep this disable, do we need annotation for it?

src/components/form/StackedTableRowForm.jsx
```diff
@@ -73,14 +73,24 @@ StackedTableRowForm.propTypes = {
   onSubmit: PropTypes.func.isRequired,
   onReset: PropTypes.func,
   // following are passed directly to formik
-  // eslint-disable-next-line react/forbid-prop-types
-  validationSchema: PropTypes.object.isRequired,
-  // eslint-disable-next-line react/forbid-prop-types
-  initialValues: PropTypes.object.isRequired,
+  validationSchema: ({ validationSchema }, propName, componentName) => {
```
kimallen, line 76: I'm confused as to what is happening here.  I thought the proptype was supposed to be definied here. How is it returning a value?

src/registerServiceWorker.js
```diff
@@ -30,8 +30,10 @@ function registerValidSW(swUrl) {
   navigator.serviceWorker
     .register(swUrl)
     .then((registration) => {
-      // eslint-disable-next-line no-param-reassign
-      registration.onupdatefound = () => {
+      // More info: https://spin.atomicobject.com/2011/04/10/javascript-don-t-reassign-your-function-arguments/
+      // This is done to avoid an edge case when mutating the argument object. Although this is not an example of the edgecase.
+      const serviceWorkerRegistration = registration;
```
kimallen, line 35: Should this be a `let` if it's being modified in the next line?
And should `registration` be deep copied here?

I'm a little confused about the explanation 'This is done to avoid an edge case when mutating the argument object.' but then it says this is not an edge case.  Why is it being done then?

src/components/Table/Table.jsx
```diff
@@ -33,6 +32,8 @@ const Table = ({
       <table {...getTableProps()}>
         <thead>
           {headerGroups.map((headerGroup, hgIndex) => (
+            /* Does not have any unique properties that can be used as key */
+            /* eslint-disable-next-line react/no-array-index-key */
```
kimallen, line 36: Does this need annotation?

src/components/Table/Table.jsx
```diff
@@ -53,9 +54,11 @@ const Table = ({
                 })}
               </tr>
               {showFilters && (
+                /* Does not have any unique properties that can be used as key */
+                /* eslint-disable-next-line react/no-array-index-key */
```
kimallen, line 58: Does this need annotation?

kimallen (changes requested):
I'm not sure which CAT IIIs are not meant to be addressed here.  I did an audit of remaining `eslint-disable`s.
Here's what I found:

- there's still 2 `no-unused-vars`: `shared/Alert/index.jsx` and `shared/Uploader/index.jsx`
- There's a `no-only-tests`, though it's not listed in the static analysis spreadsheet- we should probably check if it needs to be addressed
- there's 3 places where we do an `es-lint-disable-line` or `es-lint-disable-next-line` where we don't name a rule.  We should enable it and see what the rule is and address it if needed (`Entities/reducer.js`, `validation.test.js`, `JsonSchemaForm/index.test.js`)

I have a few questions inline to address or answer.  Otherwise it looks good!

YanZ777 (approved):

src/components/form/StackedTableRowForm.jsx
Ronolibert: This is a way to add custom validation to the proptype check. This is checking that validationSchema is an object that has at least one key/value in it. If it returns null then that means it doesn't error out. We did it this way since validationSchema is a dynamically sized object, and it also has a corresponding defaultProp to ensure that it is an object

amanda5iu (approved):

kimallen (approved):
Changes look good!  Thanks for the explanations and addressing requests.

src/testUtils.jsx
```diff
@@ -23,12 +22,13 @@ export const MockProviders = ({ children, initialState, initialEntries, history
...
-  children: PropTypes.node.isRequired,
-  // eslint-disable-next-line react/forbid-prop-types
-  initialState: PropTypes.object,
-  initialEntries: PropTypes.arrayOf(PropTypes.string),
-  // eslint-disable-next-line react/forbid-prop-types
-  history: PropTypes.object,
+  children: node.isRequired,
+  initialState: shape({}),
+  initialEntries: arrayOf(string),
+  history: shape({
```
duncan-truss, line 28: This could probably use the HistoryShape you fixed in `src/types/router.js` if we add the goBack function to it.

duncan-truss (approved):
Nice work this looked tedious 😅

YanZ777 (approved):

src/components/form/StackedTableRowForm.jsx
suzubara: this seemed like code smell to me, so I looked into it.. it doesn't look like we are actually using `StackedTableRowForm` in the app anywhere (just Storybook). the component itself is conflating a few different kinds of functionality: rendering a table row, toggling a show/edit form, rendering the form. I do think we should try to avoid using this kind of fn to validate prop types, but in this case since it's not being used, maybe adding a comment to this component to discourage usage (or at least indicate it should be refactored before using) would suffice?

src/components/DataPoint/index.jsx
jneuner-mitre: What is preventing us from creating unique ids here?  Unstructured source data?  Limited coding time?

Based on this...
  https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md
...the first line of the RA block should be...
  "Using the index as an element key in cases where the array is reordered will result in unnecessary renders."

jneuner-mitre (changes requested):
I'm hung up more on the english sentences in the RA than what we've done on the remaining items.  When we say "Mitigated" the RA: lines need to have
1) What impact/risk is identified by the linter?
2) What are the predisposing conditions?
3) How do the conditions mitigate the risk/impact?

src/components/DataPoint/index.jsx
Ronolibert: Updated the annotation! 

I believe this is an unstructured source data situation. The data being iterated over is an array of date strings which commonly has duplicates

src/shared/Entities/reducer.js
```diff
@@ -21,7 +21,7 @@ function mergeEntities(entities, newEntities) {
 // deletes all items from entities with matching key, id in deleteEntities
 function deleteEntities(entities, deleteEntities) {
   return mapValues(entities, function (value, key) {
-    // eslint-disable-next-line
+    // eslint-disable-next-line security/detect-object-injection
```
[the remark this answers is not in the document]
Ronolibert: I don't have too much extra context, but I think these type of lint errors are automatically mitigated so we don't need to annotate these? I just added the specific rule because I saw it was missing the distinction. The static analysis sheet says this finding type is mitigated for type 1 mobile code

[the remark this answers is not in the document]
jneuner-mitre: Yep, I've got the exception in the [policy](https://ustcdp3.slack.com/files/UPS881M0Q/F01EM4VNS9M/20201117_draft_management_of_source_code_weaknesses.docx) so it can be referenced and applied consistently.  But annotation still goes in the code.  The policy notes are there to make it easier to write the RA: bits.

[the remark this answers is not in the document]
Ronolibert: Ohhh I see! I'll add this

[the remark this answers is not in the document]
Ronolibert: @jneuner-mitre Added this annotation. I just annotated this one instance, but we'll have a card in the future to annotate the rest of the occurrences

src/components/DataPoint/index.jsx
```diff
@@ -8,23 +8,34 @@ const DataPoint = ({ columnHeaders, dataRow, icon, custClass }) => (
...
       <tr>
+        {/*
+          // RA Summary: eslint:react/no-array-index-key - Using the index as an element key in cases where the array is reordered will result in unnecessary renders.
+          // RA: Index of array item is being used as a part of key prop for dom element
+          // RA: The key prop is used by React to identify which items have changed, are added, or are removed and should be stable.
+          // RA: There are no other unique identifiers that can be used
+          // RA Developer Status: Mitigated
+          // RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+          // RA Validator: jneuner@mitre.org
+          // RA Modified Severity:
```
jneuner-mitre, line 26: Is this true?

```suggestion
          // RA Summary: eslint:react/no-array-index-key
          // RA: Using the index as an element key in cases where the array is reordered will result in unnecessary renders.
          // RA: The source data is unstructured, with duplicate values.
          // RA: A reorder function is not implemented for this array.
          // RA Developer Status: Mitigated
          // RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
          // RA Validator: jneuner@mitre.org
          // RA Modified Severity:
```

src/components/Table/Table.jsx
```diff
@@ -33,6 +32,16 @@ const Table = ({
...
         <thead>
           {headerGroups.map((headerGroup, hgIndex) => (
+            // RA Summary: eslint:react/no-array-index-key - Using the index as an element key in cases where the array is reordered will result in unnecessary renders.
+            // RA: Index of array item is being used as a part of key prop for dom element
+            // RA: The key prop is used by React to identify which items have changed, are added, or are removed and should be stable.
+            // RA: There are no other unique identifiers that can be used
+            // RA Developer Status: Mitigated
+            // RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+            // RA Validator: jneuner@mitre.org
+            // RA Modified Severity:
```
jneuner-mitre, line 42: Is this true?

```suggestion
          // RA Summary: eslint:react/no-array-index-key
          // RA: Using the index as an element key in cases where the array is reordered will result in unnecessary renders.
          // RA: The source data is unstructured, with duplicate values.
          // RA: A reorder function is not implemented for this array.
          // RA Developer Status: Mitigated
          // RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
          // RA Validator: jneuner@mitre.org
          // RA Modified Severity:
```

src/components/Table/Table.jsx
```diff
@@ -53,9 +62,19 @@ const Table = ({
...
               </tr>
               {showFilters && (
+                // RA Summary: eslint:react/no-array-index-key - Using the index as an element key in cases where the array is reordered will result in unnecessary renders.
+                // RA: Index of array item is being used as a part of key prop for dom element
+                // RA: The key prop is used by React to identify which items have changed, are added, or are removed and should be stable.
+                // RA: There are no other unique identifiers that can be used
+                // RA Developer Status: Mitigated
+                // RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+                // RA Validator: jneuner@mitre.org
+                // RA Modified Severity:
```
jneuner-mitre, line 72: Is this true?

```suggestion
          // RA Summary: eslint:react/no-array-index-key
          // RA: Using the index as an element key in cases where the array is reordered will result in unnecessary renders.
          // RA: The source data is unstructured, with duplicate values.
          // RA: A reorder function is not implemented for this array.
          // RA Developer Status: Mitigated
          // RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
          // RA Validator: jneuner@mitre.org
          // RA Modified Severity:
```

src/shared/Entities/reducer.js
```diff
@@ -21,7 +21,15 @@ function mergeEntities(entities, newEntities) {
 // deletes all items from entities with matching key, id in deleteEntities
 function deleteEntities(entities, deleteEntities) {
   return mapValues(entities, function (value, key) {
-    // eslint-disable-next-line
+    //RA Summary: eslint - security/detect-object-injection - Using square bracket notation with user input can lead to exploitation
+    //RA: Uses object square bracket notation
```
jneuner-mitre, line 25: Trying to keep with convention...

```suggestion
    //RA Summary: eslint - security/detect-object-injection
    //RA: Using square bracket notation with user input can lead to exploitation
    //RA: Uses object square bracket notation
```

jneuner-mitre (commented):
This really isn't convincing me that the problem is mitigated.  The comments are for future assessors that need to be convinced that predisposing conditions have been mitigated by something in the application.  Playing the devil's advocate...

> The key prop is used by React to identify which items have changed, are added, or are removed and should be stable.

Sounds like more work needs to be done.

> There are no other unique identifiers that can be used

You could make one.  Write more code!

src/components/DataPoint/index.jsx
Ronolibert: I removed this annotation. I ended up finding some unique values that the library provided for us

src/components/Table/Table.jsx
Ronolibert: I removed this annotation. I ended up finding some unique values that the library provided for us

src/components/Table/Table.jsx
Ronolibert: I removed this annotation. I ended up finding some unique values that the library provided for us

src/shared/Entities/reducer.js
Ronolibert: Ah sorry! Noted for the future

src/components/DataPoint/index.jsx
Ronolibert: I updated the annotation language. Also, on further investigation of the usage, this particular component is being used in static situations where this data never changes

jneuner-mitre (approved):

jneuner-mitre:
No more concerns from me.

PR #5824, opened 2021-02-03
[MB-6195] Update users edit page in the admin app

shimonacarvalho:
## Description

Added the active field to the UI, and some non-editable fields. Also added cypress test to check for new fields. 

## Acceptance test 

Acceptance must be done on staging. 

If you want to do these tests during review, you need to do it locally. Which you can, you just need to use a real login.gov account for the `RevokEE` not local sign in. 

LOGIN AS REVOKER
  1. Login as admin on staging, let's call that the `RevokER` account for these instructions. 
     You are going to use this account to deactivate and reactivate the test account from the apps.
     When you deactivate a user, you want to see 
- that they get automatically logged out in their window (either on refresh or they will start getting failures if they continue to use the UI) 
- AND that they can't log back in. 

TEST MILMOVE USER DEACTIVATION
  2. Create a milmove user in separate browser on my.stg.move.mil - let's call that user the `RevokEE`
  3. Find that milmove user using the `Revoker's` window and deactivate the account. 
  4. Check that `Revokee`, on refresh, gets logged out automatically
  5. Check that `Revokee`, cannot log back in
  6. Find that `Revokee` using the `Revoker's` admin window and reactivate the account. 
  7. Check that `Revokee` user CAN log back in. 

TEST OFFICE USER DEACTIVATION
8. - Use `Revoker` account to give that `Revokee` office and admin accounts. Use the create button on the `Office User` and `Admin User` page, and the email of the `Revokee`. 
9. - Login to office.stg.move.mil  and do 3-7 above to check that you can deactivate the office access of `Revokee` and kick them out. 

TEST ADMIN USER DEACTIVATION
10. - Login to admin.stg.move.mil  and do 3-7 above to check that you can deactivate the admin access of `Revokee` and kick them out. 

Note - you can sort of do these all together with one deactivate, but it's hard to keep track of when there are idle timers that will also kill the sessions.

- 

## References

* [Jira story](https://dp3.atlassian.net/browse/mb-6195) for this change

## Screenshots

New fields
image


abbyoung (approved):
lgtm! and thank you for getting that bug fix in here, too. 🎉

cypress/integration/admin/users.js
```diff
@@ -36,3 +36,30 @@ describe('Users Details Show Page', function () {
...
+      'Login gov email',
+      'Active',
+      'Revoke admin session',
+      'Revoke office session',
+      'Revoke mil session',
+    ];
+    pageContent.forEach((label) => {
+      cy.get('label').contains(label);
+    });
+  });
```
eamahanna, line 64: We should probably also test that the user can change a field and save the updated info. For the list and show pages, we only checked for content because the user couldn't perform any actions. There is an example of a user changing data in `cypress/integration/mymove/ppmCloseout.js`

cypress/integration/admin/users.js
```diff
@@ -36,3 +36,30 @@ describe('Users Details Show Page', function () {
...
+    cy.prepareAdminApp();
+  });
+
+  it('pulls up edit page for a user', function () {
+    cy.signInAsNewAdminUser();
+    failOnStatusCode: false;
+    cy.get('a[href*="system/users"]').click();
+    cy.url().should('eq', adminBaseURL + '/system/users');
+    cy.get('[data-testid="user-id"]').first().click();
+    cy.get('a').contains('Edit').click();
```
eamahanna, line 51: After this we should verify the user is on the edit page

cypress/integration/admin/users.js
shimonacarvalho: Done!

cypress/integration/admin/users.js
shimonacarvalho: Done!

cypress/integration/admin/users.js
```diff
@@ -36,3 +36,50 @@ describe('Users Details Show Page', function () {
...
+    cy.url().should('eq', adminBaseURL + '/system/users');
+    cy.get('[data-testid="user-id"]').first().click();
+
+    // ensure the page pulls up the correct user
+    cy.get('.ra-field-id > div > label')
+      .first()
+      .next()
+      .then(($userId) => {
+        cy.get('a').contains('Edit').click();
+        cy.url().should('eq', adminBaseURL + '/system/users/' + $userId.text());
```
eamahanna, line 58: NICE

eamahanna (approved):
LGTM :)

PR #5826, opened 2021-02-03
Kim mb 6518 tac validation api

kimallen:
## Description

Creates an endpoint which returns a boolean.  If an "invalid" (hardcoded stub for now) TAC is submitted, it will return {isValid: false} in the payload.

## Reviewer Notes

I was not finding how ghcapi has been checking authorization to ensure a logged in SM can't access the endpoint.  I added an authorization check in the handler (and a test), but don't see that happening in other endpoints in ghcapi, so wondering if there is somewhere else it's happening that I'm not aware of.  I [inquired with TTV](https://ustcdp3.slack.com/archives/CP4979J0G/p1612394138114100) and am awaiting a response.

The check for validity I'm using checks against a list of invalid TAC codes.  This way, until we have actual DB data to check, we can use an invalid code in the list if we need to return 'false', but users are not blocked by having to know hardcoded valid codes.

## Setup

`go test ./pkg/handlers/ghcapi -v -testify.m TestTacValidation`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6518) for this change


pkg/gen/ghcapi/configure_mymove.go
```diff
@@ -134,6 +135,11 @@ func configureAPI(api *ghcoperations.MymoveAPI) http.Handler {
 			return middleware.NotImplemented("operation mto_shipment.PatchMTOShipmentStatus has not yet been implemented")
 		})
 	}
+	if api.TacTacValidationHandler == nil {
```
lynzt, line 138: i'm ok w/ tac, but wonder if it would be more or less clear if we called it `TgetTacValidationHandler`...

pkg/handlers/ghcapi/tac.go
```diff
@@ -0,0 +1,51 @@
...
+	session, _ := h.SessionAndLoggerFromRequest(params.HTTPRequest)
+
+	if session == nil {
+		return tacop.NewTacValidationUnauthorized()
+	}
+
+	if !session.IsOfficeApp() || !session.IsOfficeUser() {
+		return tacop.NewTacValidationForbidden()
+	}
+
```
lynzt, line 27: 👍 makes sense for now...

pkg/handlers/ghcapi/tac_test.go
```diff
@@ -0,0 +1,82 @@
...
+			Tac:         tac,
+		}
+		context := handlers.NewHandlerContext(suite.DB(), suite.TestLogger())
+		handler := TacValidationHandler{context}
+		response := handler.Handle(params)
+
+		suite.Assertions.IsType(&tacop.TacValidationUnauthorized{}, response)
+	})
+
+	suite.T().Run("Unauthorized TAC validation", func(t *testing.T) {
```
lynzt, line 66: `unauthorized user tac validation`
i first read that as there are unauthorized tacs...

swagger/ghc.yaml
```diff
@@ -98,7 +98,7 @@ paths:
       operationId: getMove
   '/move-orders/{moveOrderID}':
     parameters:
-      - description: ID of customer to use
+      - description: ID of move order to use
```
lynzt, line 101: +1 good catch

lynzt (approved):
a few comments, but LGTM
🚢

swagger/ghc.yaml
```diff
@@ -1208,6 +1208,45 @@ paths:
           description: A server error occurred
           schema:
             $ref: '#/responses/ServerError'
+  '/tac/valid':
+      get:
+        summary: Validation of a TAC value
+        description: Returns a boolean based on whether a tac value is valid or not
+        operationId: tacValidation
+        tags:
+          - tac
```
duncan-truss, line 1217: I think one of the swagger lint recommendations is to declare this in the global tags block at the top of this file as well.

swagger/ghc.yaml
```diff
@@ -2554,6 +2593,12 @@ definitions:
         example: true
         type: boolean
     type: object
+  TacValid:
+    properties:
+      isValid:
+        type: boolean
+        x-nullable: false
```
duncan-truss, line 2600: I've seen us mostly use required with a list of the properties over setting x-nullable to false, is this preferable in any way?

swagger/ghc.yaml
```diff
@@ -1208,6 +1208,45 @@ paths:
...
+        summary: Validation of a TAC value
+        description: Returns a boolean based on whether a tac value is valid or not
+        operationId: tacValidation
+        tags:
+          - tac
+          - moveOrder
+        parameters:
+          - in: query
+            name: tac
+            type: string
```
duncan-truss, line 1222: I wasn't keenly following all the discussion here but were you interested in setting a min or max length validation here?

kimallen: I think for this api parameter, we won't limit the length of the string since we want to return `{isValid: true}` or `{isValid: false}` and don't want to return an error.  But that is a good point to set the value to 4 digits in the updateMoveOrder payload.  There's some questions in the air, so we'll take care of that in another story.

swagger/ghc.yaml
kimallen: No- I must have been making up my own conventions :)  I'll add it as required.

pkg/gen/ghcapi/configure_mymove.go
kimallen: I believe that would require a file name change.  Let's look at whether we want a more global naming of TGET for the file when we fill out other functionality- might make sense then.

swagger/ghc.yaml
kimallen: ok- I added it also to the global tags.  I went ahead and added the other tags used in the file that were not used there.  I also removed the `gov` tag because which was a legacy tag that was used when we originally thought we were going to have all of the customer and office apis in the same yaml.

pkg/handlers/ghcapi/tac_test.go
kimallen: Good point- I changed the subtest names to reduce confusion.

duncan-truss (approved):
Looks like a good start! Interesting that we're going to be doing inline validation on that field in the TOO orders form.

PR #5833, opened 2021-02-05
MB-4392: Delete `docker_compose_up`🗑

mdawn:
## [Delete `docker_compose_up`🗑](https://dp3.atlassian.net/browse/MB-4392)

So the original ticket said to delete the thing, but in the investigplimentory style which is in fashion nowadays, it seems like `docker-compose up` is also called here:

`run-e2e-test-docker`
- https://github.com/transcom/mymove/blob/master/scripts/run-e2e-test-docker#L52
- https://github.com/transcom/mymove/blob/master/scripts/run-e2e-test-docker#L59

Does this matter?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1030,12 +1030,6 @@ docker_compose_setup: .check_hosts.stamp ## Install requirements to use docker-c
 	brew install -f bash git docker docker-compose direnv || true
 	brew cask install -f aws-vault || true
 
-.PHONY: docker_compose_up
-docker_compose_up: ## Bring up docker-compose containers
-	aws ecr get-login-password --region "${AWS_DEFAULT_REGION}" | docker login --username AWS --password-stdin "${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.com"
-	scripts/update-docker-compose
-	docker-compose up
-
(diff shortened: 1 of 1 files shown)


carterjones (approved):
I don't have an answer regarding the `docker-compose up` calls, but I agree it's worth investigating.

esacteksab:
Oh this is interesting! Ping me in the AM, this looks fun!

esacteksab (approved):

kilbergr:
@mdawn so amusingly enough, https://github.com/transcom/mymove/blob/master/scripts/run-e2e-test-docker#L52 is `docker-compose up`, whereas you're removing `docker_compose_up`. That underscore makes a difference, because `docker_compose_up` does include `docker-compose up`, but look at all the other stuff it includes:

```
 	aws ecr get-login-password --region "${AWS_DEFAULT_REGION}" | docker login --username AWS --password-stdin "${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.com"
 	scripts/update-docker-compose
 	docker-compose up
```
Removing `docker_compose_up` therefore won't affect places this e2e script. It's a function of reuse of the same words!

kilbergr (approved):
LGTM -- See my comment about your question.

PR #5834, opened 2021-02-05
MB-6176 Add Move Edit page to Admin Interface

sandy-wright:
## Description

This PR adds the Move Edit page, which allows a user to "Show" or "Hide" a move, to the Admin Interface.

## Reviewer Notes

I didn't end up moving any files because I think these components were already in the right place, but I'm still tagging y'all, @jenniferlynparsons and @Shkeating, in case you have time and want to peek at these changes anyway ☺️ 

## Setup

To spin up the interface:

```sh
make server_run
make admin_client_run
```

Use the "Local Sign In" option, navigate to the Moves list page, click on a row to go that move's detail page, click "Edit", and then toggle the "Show" field to the other option. Save the form, and verify that the move was successfully updated (you can do this by viewing that record in your dev DB, searching for it in the TXO interface, or using the GHC API).

Then run the cypress tests:

```sh
make e2e_test
```

and select the "moves.js" tests.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6176) for this change.

## Screenshots

Screen Shot 2021-02-05 at 10 30 49 AM


cypress/integration/admin/moves.js
```diff
@@ -56,3 +57,46 @@ describe('Moves Details Show Page', function () {
...
+      'serviceMember.middleName',
+      'serviceMember.lastName',
+    ];
+    disabledFields.forEach((label) => {
+      cy.get('[id="' + label + '"]').should('be.disabled');
+    });
+
+    cy.get('#show').click();
+    cy.get('ul[aria-labelledby="show-label"] li').not('[aria-selected="true"]').click();
+    cy.get('button').contains('Save').click();
```
jenniferlynparsons, line 100: I'd like to see another expectation or two here. Beyond the "save" button, can we confirm that the content has been changed? It looks like running the `disabledFields.forEach` again but with `.should('be.enabled')` would cover that. (My syntax there is likely incorrect. 😄 )

jenniferlynparsons (approved):
The code looks good and works as expected. The comment I left about tests is optional, so I'm giving a tentative approval.

cypress/integration/admin/moves.js
sandy-wright: That's a good idea, I'll do that! This page actually redirects to the list view after the form submission, so I can find the row with the same move ID and check that the value for "Show" was saved correctly

Shkeating (commented):
Is there a way we could re-run Happo? Its showing a ton of diffs, for files you definitely did absolutely nothing to, with an error in the view, but I ran storybook on this branch and the screenshots are not accurate 🤔
image

sandy-wright:
> Is there a way we could re-run Happo? Its showing a ton of diffs, for files you definitely did absolutely nothing to, with an error in the view, but I ran storybook on this branch and the screenshots are not accurate 🤔

I'm not sure how to rerun Happo without changing anything - the only thing I can do in Happo is approve or reject the diff. There is only one difference now, at least, although it still doesn't make any sense

YanZ777:
> > Is there a way we could re-run Happo? Its showing a ton of diffs, for files you definitely did absolutely nothing to, with an error in the view, but I ran storybook on this branch and the screenshots are not accurate 🤔
> 
> I'm not sure how to rerun Happo without changing anything - the only thing I can do in Happo is approve or reject the diff. There is only one difference now, at least, although it still doesn't make any sense

You can actually rerun Happo by clicking on Happo (not the one with the Happo icon). I had a similar issue where I wanted to return Happo, but didn't know how to without doing another push, so I kind of just mucked around and found that.

sandy-wright:
> You can actually rerun Happo by clicking on Happo (not the one with the Happo icon). I had a similar issue where I wanted to return Happo, but didn't know how to without doing another push, so I kind of just mucked around and found that.

yay, thank you!! Very sneaky, Happo

PR #5861, opened 2021-02-08
MB-6656 Support API: Update getMoveTaskOrder to return the IDs needed for createMoveTaskOrder

sandy-wright:
## Description

This PR updates the Support API endpoint `getMoveTaskOrder` to return the `UploadedOrdersID`, `DestinationDutyStationID`, and the `OriginDutyStationID` fields. This is primarily so Load Testing can hit this endpoint to get all the info it needs from a move to create more moves. This information was not getting populated previously.

## Setup

To test this endpoint, first set up your local dev environment:

```sh
```

Then, using your preferred Prime API testing method (Postman or the Prime API Client), test the following endpoints in order:

* Prime API: `fetchMTOUpdates`
  * You may think, "Hmm, why didn't we change this endpoint to return these ID values?" We have to be more discerning with what information we send to the Prime, and they don't need to see things like the uploaded orders ID. They also will never need to see the contractor ID (another value we need to get from the move for load testing) because they're the contractor! So we updated the Support API instead.
  * Grab a random move's ID from this list.

* Support API: `getMoveTaskOrder`
  * Hit this endpoint using the ID you grabbed from the previous step. In the response, verify that you see values for:
    * `contractorID`
    * `destinationDutyStationID`
    * `originDutyStationID`
    * `uploadedOrdersID`
  * You can also look up this record in your dev database using a DB viewer to triple-check that the returned values are correct.

And if you want to be super thorough:

* Support API: `createMoveTaskOrder`
  * Verify that you can create a new move using the values for `contractorID`, `destinationDutyStationID`, `originDutyStationID`, `uploadedOrdersID` that you saw in `getMoveTaskOrder`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1366,7 +1366,7 @@ workflows:
             - pre_deps_golang
           filters:
             branches:
-              only: placeholder_branch_name
+              only: sw-upd-create-mto-for-load-testing
 
       - integration_tests:
           requires:
@@ -1380,7 +1380,7 @@ workflows:
           # if testing on experimental, you can disable these tests by using the commented block below.
           filters:
             branches:
-              ignore: placeholder_branch_name
+              ignore: sw-upd-create-mto-for-load-testing
 
       - integration_tests_mtls:
           requires:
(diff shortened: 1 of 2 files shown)


abbyoung (approved):
LGTM! fully tested 👍

PR #5863, opened 2021-02-08
[MB-6479] Changing region to US Gov West 1 rather than US West 1

esacteksab:
## Description

It was pointed to US West 1, this points it to US Gov West 1. To test this, a person can deploy_ecs_task which uses this code. I don't have AWS configured on this temporary computer, so I can't do this today. 

Got the region from here https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/

### DIFF
--- a/pkg/cli/aws.go
+++ b/pkg/cli/aws.go
@@ -31,7 +31,7 @@ func (e *errUnknownPartition) Error() string {
 
 // InitAWSFlags initializes AWS command line flags
 func InitAWSFlags(flag *pflag.FlagSet) {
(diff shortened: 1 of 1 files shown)


travelar (approved):

kilbergr (approved):
I think it should be fine but I would like it tested before landed--perhaps a good pairing opportunity?

esacteksab:
> I think it should be fine but I would like it tested before landed--perhaps a good pairing opportunity?

Yes, please and thank you!

kilbergr:
Tested and works!

PR #5865, opened 2021-02-09
[MB-2455] Deny payment requests if previous final payment request was already sent for the move order

mboren:
## Description

After a final payment request has been submitted for a move, we should not accept any more payment requests for that move.

In this PR, I added a check when creating a payment request to see if there are any payment requests for the same move that already have IsFinal = true. If we find any, we return `InvalidInputError`.

I also a small bug in two tests that prevented them from properly cleaning up after themselves, and broke my test when I tried to put it after them.
They both were backing up a pointer to a string (instead of its value) with the intention of restoring it at the end of the test, which did nothing.

## Reviewer Notes
- I'm never sure if I'm doing pop queries in the classiest way possible.
- I left `verrs` blank because the error doesn't really pertain to specific fields. I think this is probably OK, but I'm wondering if it means we're using the wrong flavor of error here.
 
## Setup
To test this, we have to create a `PaymentRequest` with `isFinal=true` and then create another payment request for the same MTO and verify that the second request fails.

- initial setup
```sh
make server_run
```

- create final payment request
`create_final_payment_request.json`
```json
{
  "body": {
    "isFinal": true,
    "serviceItems": [
      {
      }
    ]
  }
}
```
```sh
prime-api-client --insecure create-payment-request --filename create_final_payment_request.json
```

- Send another create-payment-request for the same MTO (this one should fail).
`create_next_payment_request.json`
```json
{
  "body": {
    "isFinal": false,
    "serviceItems": [
      {
      }
    ]
  }
}
```
```sh
prime-api-client --insecure create-payment-request --filename create_next_payment_request.json
```
Verify that this command fails with a status code of 422 and the correct error message:
```
HTTP/1.1 422 Unprocessable Entity
...
{"detail":"Cannot create PaymentRequest because a final PaymentRequest has already been submitted for MoveTaskOrder (ID: 9c7b255c-2981-4bf8-839f-61c7458e2b4d)","instance":"1fc03161-bd85-467d-8eaf-a3fb39b281f3","title":"Validation Error","invalidFields":{}}
...
```
## Acceptance
Should be essentially the same as for local, you will just need to find or create a move that has at least 2 service items that you can create payment requests for,
Then create a payment request for one of its service items with `"isFinal": true`,
and then create another payment request for the other service item, and verify that it fails.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2455) for this change

### DIFF
--- a/pkg/services/payment_request/payment_request_creator.go
+++ b/pkg/services/payment_request/payment_request_creator.go
@@ -59,6 +59,9 @@ func (p *paymentRequestCreator) CreatePaymentRequest(paymentRequestArg *models.P
 			if _, ok := err.(services.InvalidInputError); ok {
 				return err
 			}
+			if _, ok := err.(services.QueryError); ok {
+				return err
+			}
 			if errors.As(err, &badDataError) {
 				return err
 			}
@@ -254,10 +257,10 @@ func (p *paymentRequestCreator) createPaymentRequestSaveToDB(tx *pop.Connection,
 
 	// Verify that there were no previous requests that were marked as final
 	var finalPaymentRequests models.PaymentRequests
-	count, err := tx.Q().Where("move_id = $1 AND is_final = TRUE", paymentRequest.MoveTaskOrderID).Count(&finalPaymentRequests)
+	count, err := tx.Q().Where("move_i = $1 AND is_final = TRUE", paymentRequest.MoveTaskOrderID).Count(&finalPaymentRequests)
 
 	if err != nil {
-		return nil, err
+		return nil, services.NewQueryError("PaymentRequests", err, fmt.Sprintf("Error while querying final payment request for MTO %s: %s", paymentRequest.MoveTaskOrderID, err))
 	}
 
 	if count != 0 {
--- a/pkg/services/payment_request/payment_request_creator.go
+++ b/pkg/services/payment_request/payment_request_creator.go
@@ -257,7 +257,7 @@ func (p *paymentRequestCreator) createPaymentRequestSaveToDB(tx *pop.Connection,
 
 	// Verify that there were no previous requests that were marked as final
 	var finalPaymentRequests models.PaymentRequests
-	count, err := tx.Q().Where("move_i = $1 AND is_final = TRUE", paymentRequest.MoveTaskOrderID).Count(&finalPaymentRequests)
+	count, err := tx.Q().Where("move_id = $1 AND is_final = TRUE", paymentRequest.MoveTaskOrderID).Count(&finalPaymentRequests)
 
 	if err != nil {
 		return nil, services.NewQueryError("PaymentRequests", err, fmt.Sprintf("Error while querying final payment request for MTO %s: %s", paymentRequest.MoveTaskOrderID, err))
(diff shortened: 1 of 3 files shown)


pkg/services/payment_request/payment_request_creator.go
```diff
@@ -249,6 +252,14 @@ func (p *paymentRequestCreator) createPaymentRequestSaveToDB(tx *pop.Connection,
 		return nil, services.NewConflictError(moveTaskOrder.Orders.ServiceMemberID, fmt.Sprintf("ServiceMember on MoveTaskOrder (ID: %s) missing Affiliation", moveTaskOrder.ID))
 	}
 
+	// Verify that there were no previous requests that were marked as final
+	var finalPaymentRequests models.PaymentRequests
+	count, err := tx.Q().Where("move_id = $1 AND is_final = TRUE", paymentRequest.MoveTaskOrderID).Count(&finalPaymentRequests)
```
jacquelineIO, line 257: The `err` is not being being checked.

jacquelineIO (approved):
looks good. one comment about not checking the `err` response from a db query. otherwise, approving this PR.

pkg/services/payment_request/payment_request_creator.go
```diff
@@ -249,6 +255,18 @@ func (p *paymentRequestCreator) createPaymentRequestSaveToDB(tx *pop.Connection,
...
```
jacquelineIO, line 260: @mboren I do think you need to add to this query. We should also check that the payment request is not denied. The statuses are: PENDING, REVIEWED, REVIEWED_AND_ALL_SERVICE_ITEMS_REJECTED, SENT_TO_GEX, RECEIVED_BY_GEX, and PAID. 

So if one is REVIEWED_AND_ALL_SERVICE_ITEMS_REJECTED, that should not be counted as having an existing valid `final` payment request.

pkg/services/payment_request/payment_request_creator.go
```diff
@@ -249,6 +255,18 @@ func (p *paymentRequestCreator) createPaymentRequestSaveToDB(tx *pop.Connection,
...
+
+	if err != nil {
+		return nil, services.NewQueryError("PaymentRequests", err, fmt.Sprintf("Error while querying final payment request for MTO %s: %s", paymentRequest.MoveTaskOrderID, err))
```
jacquelineIO, line 263: I'm a little surprised that using formattor `%s` for `err` is not giving you an issue. I would have thought you needed `%w` or to use `err.Error()` for `%s`

jacquelineIO (commented):
Please take a look at the comment about the query.

pkg/services/payment_request/payment_request_creator.go
mboren: I copied it from somewhere else in our code. i think using `.Error()` makes more sense though, i'll tack that on.

pkg/services/payment_request/payment_request_creator.go
mboren: Thanks for catching that

mboren: I fixed this and wrote a test for it

jacquelineIO (approved):
lgtm. thanks for the updates. tested changed in local, works as expected.

PR #5869, opened 2021-02-09
Update yarn.lock

reggieriser:
## Description

It looks like our `yarn.lock` file got out of sync recently for some storybook dependencies (looks like we did some storybook updates late last week).  May just be some timing issue with the dependabot PR.  At any rate, this PR contains a regenerated `yarn.lock` that should prevent it from showing up as a local diff when building the client.

## Setup

`make clean client_deps` -- make sure there are no local diffs after this.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2604,21 +2604,6 @@
     prop-types "^15.7.2"
     regenerator-runtime "^0.13.7"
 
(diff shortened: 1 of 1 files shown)


monfresh (approved):
LGTM

PR #5872, opened 2021-02-09
use latest milmove-cypress image: MB-6645

lynzt:
## Description

We have a repo `circleci-docker` that contains a build of cypress for our integration tests. When we update any packages in `circleci-docker`, we should pull the latest image of into `mymove`.

the `circleci-docker` pr: https://github.com/transcom/circleci-docker/pull/75

## Reviewer Notes

Did I get the correct hash from Dockerhub image?

## Setup

the integration tests should run w/o an issue on circle

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6645) for this change


felipe-lee:
It looks like there's a few other places that use the hashed version of the image. Several of them seem like they're meant for only local use so idk if it'd be better to just convert them to the non-hashed version or not. 
Screen Shot 2021-02-09 at 16 47 34
The image shows the two you updated, but also has some that haven't been updated in a few months. It also shows 4 that use the non-hashed version for local use, which is what makes me think we can update the other local ones to be the non-hashed version.

gidjin:
> It looks like there's a few other places that use the hashed version of the image. Several of them seem like they're meant for only local use so idk if it'd be better to just convert them to the non-hashed version or not to avoid having to update so many, and because for local development, the hashed version matters a little less.
> 
> Screen Shot 2021-02-09 at 16 47 34
> 
> The image shows the two you updated, but also has some that haven't been updated in a few months. It also shows 4 that use the non-hashed version. Three of those seem for local use, which is what makes me think we can update the other local ones to be the non-hashed version.

This is probably something the docs need to clarify, but there are more than one image in the circleci-docker repo. You have to look at the tag names to see which is which. IE `cypress-docker:milmove-app-<hash>` is different from `cypress-docker:milmove-cypress-hash`. That said I agree with feilipe, if we change `milmove-app` hash for circle config we should update it everywhere. I wonder if @lynzt  meant to only change cypress related ones. 

As I recall we each different image gets its own tag, there isn't one `latest` for all circleci-docker images, also having the hash ensures we are not surprised by upgrades, especially in CircleCI which is where these are mostly used. But these change so infrequently it hasn't been an issue, though open to hearing otherwise.

.circleci/config.yml
```diff
@@ -13,7 +13,7 @@ version: '2.1'
 
 # References for variables shared across the file
 references:
-  circleci-docker: &circleci-docker milmove/circleci-docker:milmove-app-ff8839d671dc164855c21255c40541bb23fa1e75
+  circleci-docker: &circleci-docker milmove/circleci-docker:milmove-app-24c1fef64420e1e48de3ddb8749bad5534e272c4
```
gidjin, line 16: Did you intend to upgrade this or only references to `milmove-cypress`?

felipe-lee (approved):

PR #5873, opened 2021-02-09
[MB-5432] [CAT III] Property should not be marked as both required and readOnly

Ronolibert:
## Description

Removed createdAt and updatedAt from the required fields on the Upload definition

## Reviewer Notes
- Is this correct? From what I understand, `readOnly` is configuration for responses and `required` is a configuration for requests so I can see why this validation warning exists but are there situations where we would want both?
 
## Setup
`make server_test`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5423) for this change

### DIFF
--- a/pkg/gen/primeapi/embedded_spec.go
+++ b/pkg/gen/primeapi/embedded_spec.go
@@ -2183,9 +2183,7 @@ func init() {
       "required": [
         "filename",
         "contentType",
-        "bytes",
-        "createdAt",
-        "updatedAt"
+        "bytes"
       ],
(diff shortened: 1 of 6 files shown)


shimonacarvalho:
`readOnly` says that this value should not be in the request as it will have no effect. We use this a lot in primeAPI because it makes the field disappear from the redoc. And yeah `required` means it must be in the request. So I can see how `required` and `readOnly` wouldn't really make sense together.

shimonacarvalho (approved):
LGTM

PR #5878, opened 2021-02-10
MB-6673 Speed and clean up common Make commands

monfresh:
## Description

This PR updates `db_dev_e2e_populate` so that instead of resetting and
migrating the DB from scratch, it runs only the missing migrations, if
any, then truncates the tables. This saves 3 minutes each time you
run it. The vast majority of the time that someone runs this command,
their local dev DB is already set up, so it doesn't make sense to
destroy the DB and rebuild it from scratch when all you need is to
clear the tables and repopulate them.

And in the case where you do need to start from scratch for some reason,
this PR introduces a `db_dev_fresh` command that will reset the DB and
run all the migrations.

To follow this pattern, I updated the existing `e2e_test` script so
that it only clears the test DB tables and repopulates them, which
makes it a whole lot faster to rerun end-to-end tests with fresh data.
And in the case where you do need to start from scratch, I introduced
the `e2e_fresh` script that does what `e2e_test` used to do.

I also moved `db_dev_e2e_populate` to the `DB_DEV TARGETS` section,
where it belongs since it's used to populate the Dev DB, and not for
running automated E2E tests.

## Reviewer Notes

Run `make db_dev_e2e_populate` and `make e2e_test`. They should take about 15 seconds to run, not 3 minutes+.

Run `make db_dev_fresh` and `make e2e_test_fresh`. These should destroy the DB and run all the migrations from scratch.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6673) for this change

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -520,6 +520,37 @@ db_dev_migrate: db_dev_migrate_standalone ## Migrate Dev DB
 db_dev_psql: ## Open PostgreSQL shell for Dev DB
 	scripts/psql-dev
 
+.PHONY: db_dev_fresh
+db_dev_fresh: db_dev_reset db_dev_migrate
+	@echo "Ensure that you're running the correct APPLICATION..."
+	./scripts/ensure-application app
+	@echo "Populate the ${DB_NAME_DEV} database..."
+	go run github.com/transcom/mymove/cmd/generate-test-data --named-scenario="dev_seed" --db-env="development"
+
+.PHONY: db_dev_e2e_populate
+db_dev_e2e_populate: db_dev_migrate ## Populate Dev DB with generated e2e (end-to-end) data
+	@echo "Ensure that you're running the correct APPLICATION..."
+	./scripts/ensure-application app
+	@echo "Truncate the ${DB_NAME_DEV} database..."
+	psql postgres://postgres:$(PGPASSWORD)@localhost:$(DB_PORT_DEV)/$(DB_NAME_DEV)?sslmode=disable -c 'TRUNCATE users CASCADE; TRUNCATE uploads CASCADE;'
+	@echo "Populate the ${DB_NAME_DEV} database..."
+	go run github.com/transcom/mymove/cmd/generate-test-data --named-scenario="dev_seed" --db-env="development"
+
+## Alias for db_dev_bandwidth_up
+## We started with `db_bandwidth_up`, which some folks are already using, and
+## then renamed it to `db_dev_bandwidth_up`. To allow folks to keep using the
+## name they're familiar with, we've added this alias to the renamed command.
+.PHONY: db_bandwidth_up
+db_bandwidth_up: db_dev_bandwidth_up
+
+.PHONY: db_dev_bandwidth_up
+db_dev_bandwidth_up: bin/generate-test-data	 ## Truncate Dev DB and Generate data for bandwidth tests
+	@echo "Ensure that you're running the correct APPLICATION..."
+	./scripts/ensure-application app
+	@echo "Truncate the ${DB_NAME_DEV} database..."
+	psql postgres://postgres:$(PGPASSWORD)@localhost:$(DB_PORT_DEV)/$(DB_NAME_DEV)?sslmode=disable -c 'TRUNCATE users CASCADE; TRUNCATE uploads CASCADE;'
+	@echo "Populate the ${DB_NAME_DEV} database..."
+	DB_PORT=$(DB_PORT_DEV) go run github.com/transcom/mymove/cmd/generate-test-data --named-scenario="bandwidth" --db-env="development"
 #
 # ----- END DB_DEV TARGETS -----
 #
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -431,18 +428,32 @@ A few commands exist for starting and stopping the DB docker container:
...
-* `make db_dev_reset`: Destroys your database container. Useful if you want to start from scratch.
-* `make db_dev_migrate`: Applies database migrations against your running database container.
-* `make db_dev_migrate_standalone`: Applies database migrations against your running database container but will not check for server dependencies first.
-* `make db_dev_e2e_populate`: Populate data with data used to run e2e tests
 
 #### Test DB Commands
 
-The Dev Commands are used to talk to the dev DB.  If you were working with the test DB you would use these commands:
+These commands are available for the Test DB. You will rarely need to use these
+individually since the commands to run tests already set up the test DB properly.
```
gidjin, line 456: I run into needing to run a reset and migrate on the test database when I've deleted my docker containers for one reason or another and want to run an individual golang test. I do that rarely though as once the db is up I don't often take it down.

package.json
```diff
@@ -86,8 +86,6 @@
     "test": "react-scripts test --env=jsdom",
     "test:debug": "react-scripts --inspect-brk test --runInBand --env=jsdom",
     "test:coverage": "react-scripts test --coverage --env=jsdom --watchAll=false --coverageDirectory=coverage --runInBand",
-    "test:e2e": "cypress open --config baseUrl=http://milmovelocal:3000 --env testClientPort=3000",
```
suzubara, line 89: I use `yarn test:e2e` quite frequently to run Cypress' UI against my dev environment when debugging. It's fine with me if we want to rename this script so it's clearer that it's for debugging purposes but we shouldn't delete it.

Makefile
```diff
@@ -520,6 +520,30 @@ db_dev_migrate: db_dev_migrate_standalone ## Migrate Dev DB
...
+	go run github.com/transcom/mymove/cmd/generate-test-data --named-scenario="dev_seed" --db-env="development"
+
+.PHONY: db_bandwidth_up
+db_bandwidth_up: bin/generate-test-data	 ## Truncate Dev DB and Generate data for bandwidth tests
+	@echo "Ensure that you're running the correct APPLICATION..."
+	./scripts/ensure-application app
+	@echo "Truncate the ${DB_NAME_DEV} database..."
+	psql postgres://postgres:$(PGPASSWORD)@localhost:$(DB_PORT_DEV)/$(DB_NAME_DEV)?sslmode=disable -c 'TRUNCATE users CASCADE; TRUNCATE uploads CASCADE;'
+	@echo "Populate the ${DB_NAME_DEV} database..."
+	DB_PORT=$(DB_PORT_DEV) go run github.com/transcom/mymove/cmd/generate-test-data --named-scenario="bandwidth" --db-env="development"
```
gidjin, line 546: Thanks for cleaning this up. I am looking at these and they are fairly similar, I wonder if passing a parameter would be better vs having so many similar commands. Not that you need to change it now, just thinking of how we can improve the size of our makefile. I had never heard of `db_bandwidth_up` before this PR, which didn't add, but it looks identical to `db_dev_e2e_populate` except for the `--named-scenario` flag

gidjin: One minor nit, maybe we should rename `db_bandwidth_up` to `db_dev_bandwidth_up` since it interacts with the dev db

Makefile
```diff
@@ -661,7 +685,11 @@ db_test_psql: ## Open PostgreSQL shell for Test DB
 #
 
 .PHONY: e2e_test
-e2e_test: bin/gin server_generate server_build client_build db_e2e_init ## Run e2e (end-to-end) integration tests
+e2e_test: db_e2e_up ## Run e2e (end-to-end) integration tests
```
gidjin, line 688: I'm curious how `db_test_e2e_populate` comes into play here with `e2e_test` and `e2e_test_fresh`?

gidjin (approved):
I left some comments, but overall I think this is taking us in the right direction, thanks for making these changes! I'm approving but probably good to get more sr-eng inputs, especially those who work with the cypress tests daily

package.json
monfresh: I'd like to understand this use case more. 

- What is the advantage of `yarn test:e2e` over `make e2e_test`? 
- If `yarn test:e2e` uses the dev DB, how do you ensure the tests are valid if they depend on the DB to be in a certain state? For example, I just ran `yarn test:e2e` and several tests in `office/homepage.js` failed, but they all passed when I ran `make e2e_test`, which makes sense because my dev DB wasn't in the proper state.

README.md
monfresh: Yeah, good point. I should also mention that you'll need to run `make db_test_run` after restarting your computer.

Makefile
monfresh: `db_bandwidth_up` is brand new. I just added it last week as a helper for our bandwidth/performance tests with Puppeteer. Renaming it is a good idea.

Makefile
monfresh: Good question. It looks like it does the same thing as `db_e2e_init` (which is called by `e2e_test_fresh`), except it also adds `build_tools`. I think we can remove it.

package.json
suzubara: @monfresh purely for debugging purposes when looking into why a test failed. I agree the existing script is confusing, and I think it's fine to truncate this script to just run `cypress open`(that way the port arg can be passed in if needed) and/or rename it to something more explicit like `yarn cypress:dev`.

But sometimes if a Cypress test is failing and I don't know why, it is helpful to run against my dev instance so I can easily reproduce the test in my browser and inspect what is going on, get hot-reloading on changes, etc.

monfresh: Thanks for explaining. I noticed that we also have a `yarn test:e2e-debug`, which does the same thing as `yarn test:e2e`, and it also sets this env var: `DEBUG=cypress:server:util:process_profiler`. If `yarn test:e2e` is used for debugging, should we get rid of `yarn test:e2e-debug` and add the debug env var to `yarn test:e2e`?

suzubara: oh, huh, I never used that one before. It looks like it's used for debugging the Cypress process itself, which I don't know is used often enough to warrant its own yarn script. I can look further into these scripts and try to come up with names that make more sense..

PR #5888, opened 2021-02-11
[MB-4393] Removing transcom-ppp from .envrc and vault.go

esacteksab:
## Description

Originally tried this in #5490 but needed reverted in #5882. 

`transcom-ppp` doesn't exist, and `transcom-com-legacy` is going away. `transcom-gov-*` is the future. Because of the architecture change in AWS, the hard-coded nature of `transcom-ppp` should no longer exist. We do need `AWS_PROFILE` to be set, but it could vary depending on what task is being ran. Hopefully this ensures it is set, but allows it to be anything -- whether or not it's a valid `AWS_PROFILE` is another thing all together, and not something this takes into consideration.


pkg/cli/vault.go
```diff
@@ -88,19 +84,9 @@ func CheckVault(v *viper.Viper) error {
...
-		if len(awsProfile) > 0 && !stringSliceContains(awsProfiles, awsProfile) {
-			return errors.Wrap(&errInvalidAWSProfile{Profile: awsProfile},
-				fmt.Sprintf("%s is invalid, expected %v", VaultAWSProfileFlag, awsProfiles))
-		}
-
-		// Require both are set or neither are set
-		if (len(keychainName) != 0 && len(awsProfile) == 0) || (len(keychainName) == 0 && len(awsProfile) != 0) {
-			return errors.Wrap(&errInvalidVault{KeychainName: keychainName, Profile: awsProfile},
-				fmt.Sprintf("If either %s or %s is set the other is required", VaultAWSKeychainNameFlag, VaultAWSProfileFlag))
+		if len(awsProfile) > 0 {
```
kilbergr, line 87: Should this be 
```suggestion
		if len(awsProfile) == 0 {
```
Because you want to make sure that awsProfile DOES exists (IS > 0), right? Please correct if I'm misunderstanding.

pkg/cli/vault.go
```diff
@@ -30,10 +30,6 @@ const (
 
 	// VaultAWSKeychainNameDefault is the aws-vault default keychain name
 	VaultAWSKeychainNameDefault string = "login"
-	// VaultAWSProfileTranscomPPP is the aws-vault profile name for transcom-ppp
-	VaultAWSProfileTranscomPPP string = "transcom-ppp"
-	// VaultAWSProfileTranscomComLegacy is the aws-vault profile name for transcom-com-legacy
-	VaultAWSProfileTranscomComLegacy string = "transcom-com-legacy"
```
kilbergr, line 36: An alternate approach is to add all the other account names here and use those in the `awsProfiles` slice. That would still be hardcoding, though. I don't know if that's really better than what you have here.

kilbergr (changes requested):
I still think you want to put the line `export AWS_DEFAULT_REGION="us-gov-west-1"` (182) from .envrc outside the if block to cover the whole env as is done here: https://github.com/transcom/mymove/pull/5490/files.
Unless you're trying to avoid having the default region in the .envrc completely, which should probably be fine, it's just a change in approach. LMK if your intent was to remove as much global hardcoding as possible :)

pkg/cli/vault.go
esacteksab: Yeah, then if we add/remove an env, we have to know to update this script, and that feels icky. If we find that we want some mechanism to ensure that we use only specific environments/profiles, then I'm gonna advocate for some other mechanism rather than hard-coding those values here.

pkg/cli/vault.go
esacteksab: Off to the playground I go!

esacteksab: https://play.golang.org/p/EALIjnmAO9l I think you're right. Did I do my example right?

pkg/cli/vault.go
kilbergr: Fair fair. Thanks for addressing.

pkg/cli/vault.go
kilbergr: Yes -- I think you're understanding what I'm saying.

.envrc
```diff
@@ -47,10 +47,8 @@ check_required_variables() {
 #############################
 
 export AWS_VAULT_KEYCHAIN_NAME=login
-# The AWS_PROFILE setting is in the project configuration section
-# because now we're working with multiple accounts.
 
-export AWS_PROFILE=transcom-ppp
+export AWS_PROFILE=transcom-gov-base-id
```
travelar, line 51: I feel skeptical about this profile, but not a blocker. 🤷

travelar (approved):

.envrc
esacteksab: Admittedly I don't know which one to use. I'm happy to change it if one sticks out in your mind as better. But troubleshooting `aws-vault` this week, `AWS_PROFILE` needs to be set, in my experience, only `transcom-ppp` worked, but I'm hoping we can work past that. :-/

kilbergr: I'm really not sure what's appropriate here! I think we could either try this out and see what happens OR remove this, then alert people they may have failures, and try to address the failures where they came up. We have been following the second suggestion so far, so I'm prone to favor that. But, as I was telling Barry yesterday, there's still some aspects of aws-vault with direnv that are a bit mysterious to me. I think in those cases where something doesn't work and appears to need `AWS_PROFILE` to work, we should be noting that down and then addressing in the code to make sure it doesn't need `AWS_PROFILE` to work, or that the profile is set explicitly. What do you two think?

esacteksab: I like the idea of leaving it out. I'll merge this on Monday after notifying as many people as possible before doing so and we'll see what happens, but if there is an opportunity to remove something, we should do it!

kilbergr (approved):
LGTM...now, we wait.

kilbergr:
noice. I still approve.

kilbergr:
omg accidentally closed plz forgive so sorry!!

PR #5889, opened 2021-02-11
[MB-5526] CAT II: Remediate instances of implicit memory aliasing

Ronolibert:
## Description

Remediate instances of implicit memory aliasing in range statements that could cause unexpected side effects

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5526) for this change
* [Using loop iterator variables in go](https://github.com/transcom/mymove/wiki/Using-loop-iterator-variables-in-go)

### DIFF
--- a/pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
+++ b/pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
@@ -310,8 +310,8 @@ func MTOShipments(mtoShipments *models.MTOShipments) *ghcmessages.MTOShipments {
 	payload := make(ghcmessages.MTOShipments, len(*mtoShipments))
 
 	for i, m := range *mtoShipments {
-		// #nosec G601 TODO needs review
-		payload[i] = MTOShipment(&m)
+		copyOfMtoShipment := m // Make copy to avoid implicit memory aliasing of items from a range statement.
+		payload[i] = MTOShipment(&copyOfMtoShipment)
 	}
 	return &payload
 }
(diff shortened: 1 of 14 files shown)


pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -468,8 +468,8 @@ func MTOServiceItemDimension(d *models.MTOServiceItemDimension) *ghcmessages.MTO
 func MTOServiceItemDimensions(d models.MTOServiceItemDimensions) ghcmessages.MTOServiceItemDimensions {
 	payload := make(ghcmessages.MTOServiceItemDimensions, len(d))
 	for i, item := range d {
-		// #nosec G601 TODO needs review
-		payload[i] = MTOServiceItemDimension(&item)
+		copyOfServiceItem := item // Make copy to avoid implicit memory aliasing of items from a range statement.
```
reggieriser, line 471: Nit: These are service item dimensions, not service items, so maybe a different name here than the one above?

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -487,8 +487,8 @@ func MTOServiceItemCustomerContact(c *models.MTOServiceItemCustomerContact) *ghc
 func MTOServiceItemCustomerContacts(c models.MTOServiceItemCustomerContacts) ghcmessages.MTOServiceItemCustomerContacts {
 	payload := make(ghcmessages.MTOServiceItemCustomerContacts, len(c))
 	for i, item := range c {
-		// #nosec G601 TODO needs review
-		payload[i] = MTOServiceItemCustomerContact(&item)
+		copyOfServiceItem := item // Make copy to avoid implicit memory aliasing of items from a range statement.
```
reggieriser, line 490: Nit: Same thing here -- these are service item customer contacts, not service items.

reggieriser (approved):
This LGTM.  I added a couple of nit comments, but I think everything should work OK, so I'm going to go ahead and approve.

If you haven't already, you may want to spot check the returned JSON payload in a couple of the modified places to make sure you're getting back what you'd expect.  It may also be interesting to compare the before and after payloads to see if these changes actually fix any issues in the returned payload -- that's how we uncovered this issue a few months back with `fetch-mto-updates`.  I don't think our tests will always catch deep associations that aren't coming back with the expected JSON data.

PR #5904, opened 2021-02-12
MB-6686 Do not return service member info for logged in user if move has been hidden

abbyoung:
## Description

This PR is a part of the larger effort to not return moves if their `Show` status is marked as `False` in the db. If a move has been deactivated/hidden by an admin, we should not return it to the customer.

This bug was discovered during QA of MB-2558.

## Review

@sandy-wright, can you check out the change I made to mto_shipment? Am I correct in thinking we shouldn't use the returned move if it's hidden?

## Setup

**Test Flow**

* As a CUSTOMER: Create and submit a move.

* As an ADMIN: Edit the newly created move, and update "Show" to NO.

* As a CUSTOMER: Log in as customer and verify you can no longer see the move. 

* EXPECTED OUTCOME: The updated move should not be visible after updating "Show" to NO. When you refresh the screen as the logged-in user, you will see the logged out screen. You will not be able to log back in, but can see a 401 error in the network tab.

The lack of UI error handling is something to be addressed in future feature work.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6686) for this change

### DIFF
--- a/pkg/handlers/internalapi/mto_shipment.go
+++ b/pkg/handlers/internalapi/mto_shipment.go
@@ -169,6 +169,7 @@ func (h ListMTOShipmentsHandler) Handle(params mtoshipmentops.ListMTOShipmentsPa
 	// check if move task order exists first
 	queryFilters := []services.QueryFilter{
 		query.NewQueryFilter("id", "=", moveTaskOrderID.String()),
+		query.NewQueryFilter("show", "=", "TRUE"),
 	}
 
 	moveTaskOrder := &models.Move{}
--- a/pkg/models/service_member.go
+++ b/pkg/models/service_member.go
@@ -98,15 +98,13 @@ func FetchServiceMemberForUser(ctx context.Context, db *pop.Connection, session 
 		"BackupMailingAddress",
 		"BackupContacts",
 		"DutyStation.Address",
-		"DutyStation.TransportationOffice",
 		"DutyStation.TransportationOffice.PhoneLines",
 		"Orders.NewDutyStation.TransportationOffice",
-		"Orders.Moves",
 		"Orders.UploadedOrders.UserUploads.Upload",
-		"ResidentialAddress",
-		"Orders.Moves").
+		"Orders.Moves",
+		"ResidentialAddress").
 		Join("orders", "orders.service_member_id = service_members.id").
-		Join("moves", "moves.orders_id = orders.id"). // We only want to return a move if it is visible to the user
+		Join("moves", "moves.orders_id = orders.id").
 		Where("show = TRUE").Find(&serviceMember, id)
 
 	if err != nil {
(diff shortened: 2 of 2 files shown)


pkg/models/service_member.go
```diff
@@ -98,12 +98,15 @@ func FetchServiceMemberForUser(ctx context.Context, db *pop.Connection, session
 		"BackupMailingAddress",
 		"BackupContacts",
 		"DutyStation.Address",
-		"DutyStation.TransportationOffice",
```
abbyoung, line 101: This was removed because the next line will grab the TransportationOffice association anyway.

abbyoung: @Ronolibert I tested and it looks good to me, but since you were last in the file, tagging you 😄

pkg/models/service_member.go
```diff
@@ -98,12 +98,15 @@ func FetchServiceMemberForUser(ctx context.Context, db *pop.Connection, session
...
 		"Orders.NewDutyStation.TransportationOffice",
-		"Orders.Moves",
 		"Orders.UploadedOrders.UserUploads.Upload",
-		"ResidentialAddress").Find(&serviceMember, id)
+		"Orders.Moves",
+		"ResidentialAddress").
+		Join("orders", "orders.service_member_id = service_members.id").
+		Join("moves", "moves.orders_id = orders.id").
+		Where("show = TRUE").Find(&serviceMember, id)
+
```
eamahanna, line 109: Can you add a unit test to make sure this works? I'm not sure if that is overkill or not since this is a bug fix with a small scope.

abbyoung: no you are right -- turns out i broke a lot of tests and am working on that now. thank you!

pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -169,6 +169,7 @@ func (h ListMTOShipmentsHandler) Handle(params mtoshipmentops.ListMTOShipmentsPa
 	// check if move task order exists first
 	queryFilters := []services.QueryFilter{
 		query.NewQueryFilter("id", "=", moveTaskOrderID.String()),
+		query.NewQueryFilter("show", "=", "TRUE"),
```
sandy-wright, line 172: this looks good to me ✨

pkg/models/service_member.go
abbyoung: jk i ended up reverting this since i no longer needed to touch this file 🎉

pkg/models/service_member.go
abbyoung: reverted 🙅

sandy-wright (approved):
Definitely is wild to just get booted to the sign-on screen, what an experience, but it's doing what it should otherwise! LGTM ✨

PR #5926, opened 2021-02-17
[MB-5432] CAT III: SA6005 Use EqualFold instead of string comparison

Ronolibert:
## Description

Replace regular string equality comparison with use of EqualFold method

## Setup
`make server_run`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5432) for this change

### DIFF
--- a/pkg/migrate/SplitStatements.go
+++ b/pkg/migrate/SplitStatements.go
@@ -220,7 +220,7 @@ func SplitStatements(lines chan string, statements chan string, wait time.Durati
 
 		// Let's see if we match the last block
 		if !blocks.Empty() {
-			lastBlock := strings.ToUpper(blocks.Last())
+			lastBlock := blocks.Last()
 			str, err := in.Range(i, i+len(lastBlock))
 			if err != nil {
 				if err == ErrWait {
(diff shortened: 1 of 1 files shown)


pkg/migrate/SplitStatements.go
```diff
@@ -231,7 +231,7 @@ func SplitStatements(lines chan string, statements chan string, wait time.Durati
 				close(statements)
 				return
 			}
-			if strings.ToUpper(str) == strings.ToUpper(lastBlock) {
+			if strings.EqualFold(str, lastBlock) {
```
lynzt, line 234: wondering if we need the `strings.ToUpper` on line 223? I know that's not part of this, but uppercasing above seems not needed...

lynzt (approved):
question regarding extra uppercasing...
other than that, LGTM

pkg/migrate/SplitStatements.go
Ronolibert: Good call! I actually added the toUpper calls back in because its guarding against case sensitivity. I did remove it from 223 since it was a redundant call

PR #5927, opened 2021-02-17
MB-6218 Incorporate the EULA Modal Into the Office App

YanZ777:
## Description

Remove the code that disabled the EULA modal for the office app. 
Import the USWDS styles so that the modal overlays correctly.

## Reviewer Notes

Should I be importing the USWDS styles in that manner?
Getting this PR out early since I want some weigh in on what people think of the above question. 
I'll be adding tests and changing the Sign Out to be a button as well.

## Setup

```sh
make office_client_run
make server_run
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6218) for this change

## Screenshots

image


YanZ777:
I'll try to beef up the Login / SigninButton tests.

YanZ777:
@suzubara I'll be changing the SignOut to be a button as well in this PR.

jenniferlynparsons (approved):
Looks good!

src/containers/LoginButton/LoginButton.jsx
```diff
@@ -39,21 +39,15 @@ const LoginButton = ({ isLoggedIn, logOut, showDevlocalButton, useEula }) => {
...
-            </Button>
-          ) : (
-            <a className="usa-nav__link" data-hook="signin" href="/auth/login-gov">
-              Sign In
-            </a>
-          )}
+          <Button
+            aria-label="Sign In"
+            className={styles.signIn}
+            testid="signin"
```
suzubara, line 45: test ID attributes should be `data-testid`

src/containers/LoginButton/LoginButton.module.scss
```diff
@@ -2,18 +2,30 @@
...
+    width: 46px;
+  }
+}
+
+.signOut {
+  &:hover::after {
+    width: 56px;
+  }
+}
+
```
suzubara, line 28: I was just looking at these buttons and I _think_ if you add the `usa-nav__link` to them, you won't have to re-implement the `::after` psuedo element (but I think you would need to keep `&:hover::after { background-color: white }`) - that will probably be easier to maintain

suzubara (commented):
Just a couple of very minor comments, but overall looking good!

mariaclaratgolfo (commented):
Why is the cancel button red for this modal?

YanZ777:
> Why is the cancel button red for this modal?

@mariaclaratgolfo 
Looks like it is coming from the usa-button--secondary styling. Do you want me to change it to white with blue outline like on the customer side?

src/containers/LoginButton/LoginButton.module.scss
YanZ777: I didn't use the usa-nav__link because I'd end up getting errors about "No toggle target found with id:". Maybe I'm missing something?

mariaclaratgolfo:
I think we should treat it the same way we're treating it on the Customer side.
Red buttons should be reserved for destructive actions. Canceling in this context isn't destructive on the user's part.
@amanda5iu feel free to share if you disagree!

amanda5iu:
> I think we should treat it the same way we're treating it on the Customer side.
> Red buttons should be reserved for destructive actions. Canceling in this context isn't destructive on the user's part.

Agreed. @YanZ777 can we use the secondary button style from Storybook?

mariaclaratgolfo (approved):

PR #5932, opened 2021-02-18
Make some devseeded moves available to prime

shimonacarvalho:
## Description

Many of the devseed moves have payment requests but are not available to prime. This does not make sense, since a move would have to have been available to prime, for prime to create a payment request. 

This change sets the `availableToPrimeAt` field for every move with a payment request. 
Also in one case, a move with a payment request was in `SUBMITTED` state so moved it to `APPROVED`.

I've checked and the number of "available" moves goes from 5 to 22 in a freshly populated db. This should help with testing prime stuff. 

## Reviewer Notes

@gidjin I'd appreciate if you could check that I didn't do it to any move that SHOULDN'T be availableToPrime.

## References

* [this conversation](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1607559551003900) discussing fixing this.


pkg/testdatagen/scenario/devseed.go
```diff
@@ -194,6 +194,8 @@ func createPPMWithPaymentRequest(db *pop.Connection, userUploader *uploader.User
 	})
 	ppm2.Move.Submit(time.Now())
 	ppm2.Move.Approve()
+	ppm2.Move.AvailableToPrimeAt = swag.Time(time.Now())
```
gidjin, line 197: 🤔  this move is a PPM not an HHG, I'm not sure PPMs are ever available to the prime? Maybe not set it on the ppms? Also I see the method is called `createPPMReadyToRequestPayment` but I don't think it's referring to a PaymentRequest for us to invoice via Syncada. But again we haven't been dealing with PPMs so 🤷‍♀️

pkg/testdatagen/scenario/devseed.go
```diff
@@ -604,6 +606,8 @@ func createPPMReadyToRequestPayment(db *pop.Connection, userUploader *uploader.U
 	})
 	ppm6.Move.Submit(time.Now())
 	ppm6.Move.Approve()
+	ppm6.Move.AvailableToPrimeAt = swag.Time(time.Now())
```
gidjin, line 609: Same here, this is a ppm not an HHG and I'm not sure it makes sense for a PPM to have `AvailableToPrimeAt` set.

gidjin (approved):
Changes look good I don't think they will mess anything up for us. I did leave a note on the 2 PPMs as I'm not clear on if those should have `AvailableToPrimeAt` set or not. Feels like no since I don't think the prime helps with PPMs.

pkg/testdatagen/scenario/devseed.go
shimonacarvalho: Ah yes, that makes sense. Thanks!

monfresh:
When I first created the devseed  file, I fixed some of the data that didn’t represent reality. If I recall, part of that was making sure the moves were in SUBMITTED status instead of DRAFT, and that `AvailableToPrimeAt` was **not** set, because it can’t be available to prime until the TOO approves it.

If we want moves that have already been approved and made available to the prime, I think we should create new ones with the appropriate move status, as opposed to changing existing ones. If we have a move with SUBMITTED status and AvailableToPrimeAt set, it doesn't reflect reality.

monfresh:
Then again, as John pointed out, we have payment requests on submitted moves, which doesn't make sense either. I suppose as long as the TOO can still view the move in a submitted state and approve it, and if setting `AvailableToPrimeAt` doesn't affect that, then it should be fine.

PR #5938, opened 2021-02-18
Remove toUpper call in EqualFold

Ronolibert:
## Description

In addition to #5926, removing the toUpper call in EqualFold because EqualFold already does a case insensitive comparison


lynzt (approved):
LGTM 🚢

PR #5940, opened 2021-02-19
updateWebhookSubscription

NamibiaLT:
## Description

This is part of the Push Deploy work to add a new handler to update webhook subscriptions.

## Reviewer Notes

Please play close attention to the tests for UpdateWebhookSubscription Handler and Service Object

**The endpoint show allow updating of the following fields:**
subscriber_id - not the id of the subscription, but links to which contractor this subscription is related to status (ACTIVE, FAILING, DISABLED), event_key, callback_url, severity

## Setup
To test the endpoint in Postman
1. Add a new collection, which will be this branch `nt-CORRECTED-mb-6578-UpdateWebhookSubscription`
2. Make sure that you are making a PATCH request with url: {{baseUrl}}/webhook_subscriptions
3. Follow [these instructions] (https://github.com/transcom/mymove/wiki/Test-Admin-and-Office-APIs-with-Postman) for copying the correct X-CSRF token into your Header in Postman
4. You should be able to send the request and receive a 200 response.

To run tests

```sh
go test ./pkg/services/webhook_subscription -testify.m TestWebhookSubscriptionUpdater -failfast  -v
go test ./pkg/handlers/adminapi -testify.m TestWebhookSubscriptionUpdater -failfast  -v

```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6578) for this change


pkg/handlers/adminapi/payloads/payload_to_model.go
```diff
@@ -27,3 +29,22 @@ func UserModel(user *adminmessages.UserUpdatePayload, id uuid.UUID) (*models.Use
 
 	return model, nil
 }
+
+// WebhookSubscriptionModel represents a webhook subscription
```
shimonacarvalho, line 33: Minor nit - this function converts a webhook subscription payload to a model.

pkg/handlers/adminapi/payloads/payload_to_model.go
```diff
@@ -27,3 +29,22 @@ func UserModel(user *adminmessages.UserUpdatePayload, id uuid.UUID) (*models.Use
...
+func WebhookSubscriptionModel(sub *adminmessages.WebhookSubscription) *models.WebhookSubscription {
+	model := &models.WebhookSubscription{
+		CallbackURL:  sub.CallbackURL,
+		CreatedAt:    time.Now(),
```
shimonacarvalho, line 37: You can remove CreatedAt and UpdatedAt. Pop will automatically update them. We especially don't want createdAt to be updated everytime the subscription is updated.

pkg/handlers/adminapi/payloads/payload_to_model.go
```diff
@@ -27,3 +29,22 @@ func UserModel(user *adminmessages.UserUpdatePayload, id uuid.UUID) (*models.Use
...
+		EventKey:     sub.EventKey,
+		ID:           uuid.FromStringOrNil(sub.ID.String()),
+		Severity:     int(*sub.Severity),
```
shimonacarvalho, line 40: Don't have the Severity here because if it is nil, there will be a dereference. Your other code on line 46 takes care of this.

pkg/handlers/adminapi/webhook_subscriptions.go
```diff
@@ -91,3 +93,31 @@ func (h GetWebhookSubscriptionHandler) Handle(params webhooksubscriptionop.GetWe
...
+}
+
+// Handle updates a webhook subscription content
+func (h UpdateWebhookSubscriptionHandler) Handle(params webhooksubscriptionop.UpdateWebhookSubscriptionParams) middleware.Responder {
+	logger := h.LoggerFromRequest(params.HTTPRequest)
+	payload := params.WebhookSubscription
+
+	// Payload to model converter
+	webhookSubscription := payloads.WebhookSubscriptionModel(payload)
+	logger.Info("got here")
```
shimonacarvalho, line 111: these were for debug, you can take them out :)

pkg/handlers/adminapi/webhook_subscriptions.go
```diff
@@ -91,3 +93,31 @@ func (h GetWebhookSubscriptionHandler) Handle(params webhooksubscriptionop.GetWe
...
+	fmt.Println(webhookSubscription)
+
+	updatedWebhookSubscription, err := h.WebhookSubscriptionUpdater.UpdateWebhookSubscription(webhookSubscription, "eTag")
```
shimonacarvalho, line 114: You're passing a string "eTag" here, that seems wrong? I am not sure why that would work actually. 
I'm actually not sure if you need to check the etag on admin api endpoints, we can discuss in slack.

pkg/handlers/adminapi/webhook_subscriptions.go
```diff
@@ -91,3 +93,31 @@ func (h GetWebhookSubscriptionHandler) Handle(params webhooksubscriptionop.GetWe
...
+	// Check that the uuid provided is valid
+	if err != nil {
+		logger.Error(fmt.Sprintf("Error updating webhookSubscription %s", params.WebhookSubscriptionID.String()), zap.Error(err))
```
shimonacarvalho, line 117: If the update failed, probably you want to return some sort of error response? This way it eventually returns an ok, even if the update failed.

shimonacarvalho (changes requested):
Hi Namibia, that was a lot of work! I've put some review comments that need to be fixed. I didn't look at the tests yet because i think they will need some rework after the changes are made.

swagger/admin.yaml
```diff
@@ -1732,6 +1754,31 @@ paths:
           description: Webhook subscriptions not found
         500:
           description: Server error
+    post:
```
shimonacarvalho, line 1757: Little unsure why this is showing up here, you didn't add this one right? It should have been there already?

swagger/admin.yaml
```diff
@@ -1758,3 +1805,37 @@ paths:
           description: subscription not found
         500:
           description: server error
+    patch:
```
shimonacarvalho, line 1808: Looks good!

swagger/admin.yaml
eamahanna: You just need to merge the master branch into this one. Then you will no longer see previous changes. (If that is what this is.)

pkg/handlers/adminapi/webhook_subscriptions.go
NamibiaLT: After this err check if there was something wrong, would there be another failure that hasn't been checked? Or does this error response need to be more specific?

swagger/admin.yaml
NamibiaLT: Hmmm, I don't remember adding anything but the patch, should I remove it? It looks like I added it based on it being green.

swagger/admin.yaml
NamibiaLT: Yay!

pkg/handlers/adminapi/webhook_subscriptions.go
shimonacarvalho: No I meant this "logger.Error" is just printing something in the logs, you need to return an actual error to the client. See other admin api handlers to see how they return errors (since it's probably a bit different than primeAPI).
```
return webhooksubscriptionop.NewUpdateWebhookSubscriptionInternalServerError...
```

pkg/handlers/adminapi/webhook_subscriptions_test.go
```diff
@@ -154,3 +154,70 @@ func (suite *HandlerSuite) TestGetWebhookSubscriptionHandler() {
...
+
+	suite.T().Run("200 - OK, Successfuly updated webhook subscription", func(t *testing.T) {
+		// Testing: 			UdateWebhookSubscriptionHandler, Updater
+		// Set up: 				Provide a valid request with the id of a webhook_subscription
+		// 		   					to the updateWebhookSubscription endpoint.
+		// Expected Outcome: 	The webhookSubscription is updated and we
+		//					 		receive a 200 OK.
+		params := webhooksubscriptionop.UpdateWebhookSubscriptionParams{
+			HTTPRequest:           req,
+			WebhookSubscriptionID: strfmt.UUID(webhookSubscription.ID.String()),
```
shimonacarvalho, line 171: Better to update all fields and check that they were updated.

pkg/handlers/adminapi/webhook_subscriptions_test.go
```diff
@@ -154,3 +154,70 @@ func (suite *HandlerSuite) TestGetWebhookSubscriptionHandler() {
...
+			query.NewQueryFilter,
+		}
+
+		response := handler.Handle(params)
+
+		expectedResponse := &handlers.ErrResponse{
+			Code: http.StatusNotFound,
+			Err:  expectedError,
+		}
+		suite.Equal(expectedResponse, response)
```
shimonacarvalho, line 221: I'm confused how this would work because the handler has only one return statement as follows:
```
return webhooksubscriptionop.NewUpdateWebhookSubscriptionOK().WithPayload(payload)
```
And that returns OK not an error.

pkg/services/webhook_subscription/webhook_subscription_updater.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	verrs, err := o.builder.UpdateOne(&foundWebhookSubscription, nil)
+
+	if verrs != nil && verrs.HasAny() {
+		return nil, services.NewInvalidInputError(webhookSubscriptionID, err, verrs, "")
+	}
+
+	// First check to see if there is an error on the type and return a precondition fail error, if not return the error
+	if err != nil {
```
shimonacarvalho, line 63: You need to check the precondition before you call updateOne. If the precondition fails, you should not update the subscription.

shimonacarvalho (changes requested):
Hey there, good work, getting closer but still needs some fixes.

pkg/handlers/adminapi/webhook_subscriptions_test.go
NamibiaLT: All fields in the model?

pkg/handlers/adminapi/webhook_subscriptions_test.go
NamibiaLT: Is this a sufficient error checker?
	// Check that the uuid provided is valid
	if err != nil {
		logger.Error(fmt.Sprintf("Error updating webhookSubscription %s", params.WebhookSubscriptionID.String()), zap.Error(err))
	}

 I thought we return OK() once the model is converted back to a payload.

pkg/services/webhook_subscription/webhook_subscription_updater.go
NamibiaLT: Ok that makes sense. I'll mmove the precondition fail check before updating and update the test.

pkg/handlers/adminapi/webhook_subscriptions_test.go
shimonacarvalho: All fields that can be updated should be passed in, and then checked at the end of the test that they were updated correctly.

pkg/handlers/adminapi/webhook_subscriptions_test.go
shimonacarvalho: Please take out the mocks here. The mock is making it seem like it would return "http.StatusNotFound" but the handler is not returning that, at least that's what we saw when we paired.

pkg/services/webhook_subscription/webhook_subscription_updater.go
```diff
@@ -0,0 +1,78 @@
...
+			return nil, err
+		default:
+			return nil, err
+		}
+	}
+
+	// Update webhook subscription new status for Active
+	if &webhooksubscription.Status != nil {
+		foundWebhookSubscription.Status = webhooksubscription.Status
+	}
```
shimonacarvalho, line 38: Hmm I'm not sure there should be an & before webhookSubscription.Status (or any of the fields) here. Because that will be the address of that and would never be nil.

pkg/handlers/adminapi/webhook_subscriptions_test.go
NamibiaLT: See 8693770, it has been removed from the tests

pkg/handlers/adminapi/webhook_subscriptions_test.go
```diff
@@ -154,3 +155,131 @@ func (suite *HandlerSuite) TestGetWebhookSubscriptionHandler() {
...
+		}
+
+		response := handler.Handle(params)
+
+		expectedError := models.ErrFetchForbidden
+		expectedResponse := &handlers.ErrResponse{
+			Code: http.StatusNotFound,
+			Err:  expectedError,
+		}
+		suite.IsType(expectedResponse, response)
```
shimonacarvalho, line 280: This test looks identical to the previous test, I might be missing something. Can you explain what is different and why this test returns StatusNotFound, but the previous returns InternalServerError?

pkg/services/webhook_subscription/webhook_subscription_updater.go
shimonacarvalho: Needs update.

pkg/handlers/adminapi/webhook_subscriptions_test.go
```diff
@@ -154,3 +155,131 @@ func (suite *HandlerSuite) TestGetWebhookSubscriptionHandler() {
...
+		// 	WebhookSubscriptionID: strfmt.UUID(webhookSubscription.ID.String()),
+		// }
+		fakeID, err := uuid.NewV4()
+		suite.NoError(err)
+
+		params := webhooksubscriptionop.UpdateWebhookSubscriptionParams{
+			HTTPRequest:           req,
+			WebhookSubscriptionID: strfmt.UUID(fakeID.String()),
+			WebhookSubscription: &adminmessages.WebhookSubscription{
+				CallbackURL:  "something.com",
```
shimonacarvalho, line 216: I tested with postman and if you don't send all the values, you get a validation error. Since this is a patch, it should be okay to send some values but not all.

abbyoung:
@NamibiaLT This PR can be closed since the rebased branch and final changes were merged in via #6030

shimonacarvalho:
This PR can be closed since the rebased branch and final changes were merged in via #6030

PR #5954, opened 2021-02-22
[MB-5442] CAT III: SA1006 Printf with no dynamic arguments

Ronolibert:
## Description

Convert Printf usages to Print due to not using dynamic argument for Printf

## Setup
`make server_test`
Go to .golangci.yml
Move `staticcheck` from disabled linters to enabled linters
Run command `pre-commit run -v --all-files golangci-lint` 

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5442) for this change


pkg/handlers/supportapi/payment_request_test.go
```diff
@@ -561,7 +561,7 @@ func (suite *HandlerSuite) TestProcessReviewedPaymentRequestsHandler() {
 
 	suite.T().Run("successful update of reviewed payment requests with send to syncada true", func(t *testing.T) {
 		// Call the handler to update all reviewed payment request to a "Sent_To_Gex" status
-		req := httptest.NewRequest("PATCH", fmt.Sprintf(urlFormat), nil)
+		req := httptest.NewRequest("PATCH", fmt.Sprintln(urlFormat), nil)
```
felipe-lee, line 564: Mostly curious, why did this one turn into `Sprintln` while the rest in this file are now `Sprint`. Based on reading the docs I _think_ they're similar except one has spaces and a newline, but I want to make sure I'm understanding correctly.

pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -441,7 +441,7 @@ func (o *mtoServiceItemCreator) validateFirstDaySITServiceItem(serviceItem *mode
 		verrs := validate.NewErrors()
 		verrs.Add("reServiceCode", fmt.Sprintf("%s invalid code", serviceItem.ReService.Code))
 		return nil, services.NewInvalidInputError(serviceItem.ID, nil, verrs,
-			fmt.Sprintf(fmt.Sprintf("No additional items can be created for this service item with code %s", serviceItem.ReService.Code)))
+			fmt.Sprintf("No additional items can be created for this service item with code %s", serviceItem.ReService.Code))
```
YanZ777, line 444: Should this also be changed to a fmt.Sprint?

pkg/handlers/supportapi/payment_request_test.go
Ronolibert: Good catch! This was a mistake. It even makes the test fail because the handler is expecting this to be `Sprint`

pkg/services/mto_service_item/mto_service_item_creator.go
Ronolibert: I think we want it in this case because we're using a dynamic variable in `serviceItem.ReService.Code`

YanZ777 (approved):

PR #5966, opened 2021-02-23
MB-6315 MTO Service Item auto approval on move approval

Ryan-Koch:
## Description

This attempts to fix a bug where after one clicks on this approval button:
image

the service items in the grey box are created in an approved state, as opposed to the submitted state that's happening currently:
image

Also added a test run to `TestUpdateMTOShipmentStatus` to ensure we get a failure if we create a situation where these aren't getting created as approved again later.

## Reviewer Notes

I moved setting the status into the `CreateMTOServiceItemHandler` since to my understanding the Prime can only ask for MTO service items and can't approve them unilaterally. I also then needed to update our service object tests to ensure that the status is being provided (since that field should not be nil at that time). The other place where we're creating this is from the office side and from there we're setting an approved status before it gets into the MTO Service Item creator.  So that made the tests happy, but beyond those unit tests there are a couple of important questions:

- Are my assumptions okay?
- Have I broken something without realizing it?

## Setup

Get your office app up and running

```sh
make server_run
make office_client_run
make db_dev_e2e_populate
```

- Find a 'new move' status move in your TOO queue.
- Approve the move
- Click on the Move Task Orders tab and see if the service items that were created are approved (they should be 🤞)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6315) for this change

## Screenshots

autoapprove


eamahanna (commented):
I ran load testing against these changes and everything worked as it should. I think you may need to look at why a cypress test failed.

Ryan-Koch:
> I ran load testing against these changes and everything worked as it should. I think you may need to look at why a cypress test failed.

Ah yeah I need to fix that. The test is looking for requested service items because that's how it was before. Kind of a 'this task failed successfully' kind of moment. I'll push a change to take care of that.

duncan-truss (approved):
Works as expected for me, it looks like we are now always passing in a status to that function from the ghc and prime flows.

Should there be an error in that createMTOServiceItem function if the status is not set on the service item?  Since you've removed the default, it might do a whole lot of work before erroring out.  I could see us either setting it to submitted if the status is the empty string or failing early with some expected error message.

donaldthai (approved):
LGTM

Ryan-Koch:
> Works as expected for me, it looks like we are now always passing in a status to that function from the ghc and prime flows.
> 
> Should there be an error in that createMTOServiceItem function if the status is not set on the service item? Since you've removed the default, it might do a whole lot of work before erroring out. I could see us either setting it to submitted if the status is the empty string or failing early with some expected error message.

That's a good point. I like the idea of going with `submitted` as a default if we end up with a situation where we aren't passing it. I'll add in a check for that and a test for it to keep our bases covered.

PR #5973, opened 2021-02-24
MB-6397 Add migration for TAC codes

felipe-lee:
## Description

We're doing a secure migration so the committed file only has fake TAC codes. The secure migration looks similar, but with 35,570 rows instead of just 19.

## Reviewer Notes

Does this look like the right way to bulk add records?

I also updated the [instructions for secure migrations](https://github.com/transcom/mymove/wiki/migrate-the-database) to add a few details I learned as I was working on this. Here is [diff of the instruction changes](https://github.com/transcom/mymove/wiki/migrate-the-database/_compare/92ba0331af4b501ebd75298f4f31c5ed5fba02fb...091c8bb52ba48710ff1112fb3d0458030e4a9816). Can you review those as well to make sure it all makes sense?

## Setup

You can test this migration on the dev db by running:
```sh
make db_dev_migrate
```

You can test the secure migration by running:
```sh
make run_prd_migrations
```

  * [x] Have been communicated to #g-database
  * [x] Secure migrations have been tested following the instructions in our [docs](https://github.com/transcom/mymove/wiki/migrate-the-database#secure-migrations)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6397) for this change


lynzt:
@felipe-lee let's use `now()` vs hardcoding datetime values

felipe-lee:
> @felipe-lee let's use `now()` vs hardcoding datetime values

@lynzt Yeah Reggie and I talked earlier and I'm testing out some migrations locally that I'll push once I'm done. Looks like they're working.

migrations/app/secure/20210224203932_add_transportation_accounting_codes.up.sql
```diff
@@ -0,0 +1,51 @@
...
+SET standard_conforming_strings = on;
+SELECT pg_catalog.set_config('search_path', '', false);
+SET check_function_bodies = false;
+SET xmloption = content;
+SET client_min_messages = warning;
+SET row_security = off;
+
+-- Temporarily drop indexes to speed up load
+
+DROP INDEX public.transportation_accounting_codes_tac_idx;
```
reggieriser, line 21: I was looking at the DDL for this table and I'm wondering if the `transportation_accounting_codes_tac_idx` index is actually double-indexing this column.  I believe given that `tac` is a unique column, Postgres will automatically create an index for it [per the Postgres docs](https://www.postgresql.org/docs/12/indexes-unique.html).  But if you remove this `transportation_accounting_codes_tac_idx` index, you still may want to drop the unique constraint if you're trying to get the highest throughput.  It may not be worth it for 35k rows, but you'd have to experiment and see.  I know it did make a non-trivial difference when I was inserting 400k+ rows in another PR, but that's a lot more data than you're dealing with here.

felipe-lee: Ah I see. Good to know, I can make the index removal permanent in the regular migration then and then make the secure migration temporarily remove the unique constraint. I don't know if it'll buy us much, but it shouldn't really hurt to temporarily remove and re-add it since I know the data for sure is unique.

reggieriser (approved):
I tested the migrations both locally and for `prd` and the schema/data looked as expected.  Migrations are nicely done and efficient.  👍 

I also grabbed the source data file noted in the JIRA story and pulled out what I think are the TACs from the file with some unix shell cutting/unique sorting, then compared against a ordered list from a database query.  They matched 100% (although my list also had a blank TAC for some reason, but I assume that's just some funky source data).

lynzt (approved):
LGTM...
Ran this in `prd` and locally and looks good 🚢

PR #5982, opened 2021-02-24
[MB-5444] CAT III: staticcheck 1019 - Field or method is deprecated

Ronolibert:
## Description

Remove usages of deprecated fields or methods

## Setup
`make server_test`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5444) for this change

### DIFF
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -66,7 +66,6 @@ func (suite *serverSuite) TestParseSingleTLSCert() {
 	})
 	suite.NoError(err)
 	suite.Equal(len(httpsServer.TLSConfig.Certificates), 1)
-	suite.Contains(httpsServer.TLSConfig.NameToCertificate, "localhost")
 }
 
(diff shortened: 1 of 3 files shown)


pkg/server/server_test.go
```diff
@@ -66,7 +66,6 @@ func (suite *serverSuite) TestParseSingleTLSCert() {
 	})
 	suite.NoError(err)
 	suite.Equal(len(httpsServer.TLSConfig.Certificates), 1)
-	suite.Contains(httpsServer.TLSConfig.NameToCertificate, "localhost")
```
Ronolibert, line 69: The lint error for this is: `NameToCertificate only allows associating a single certificate with a given name. Leave this field nil to let the library select the first compatible chain from Certificates`

Is removing this and letting the library handle this business logic correct?

travelar: My Go is not well refined, but I think that this will work ok. If not, we can revert and swarm on it.

travelar (approved):

PR #5987, opened 2021-02-24
Sc prime demo update

shimonacarvalho:
## Description

Adds ability to select different types of PRs to create with a menu of options to select from.
We needed in this demo to create an empty and final pr. 

While I was in there, I also 
- moved adding accessorial service item to a function, so we can comment out one line when not needed for demo.
- check that the proof of service docs actually exist since I messed this up on the demo and we can easily check this
- broke up the wall of bash with some section comments to improve scannability
- reworded some bits of the usage statement that confused me

## Reviewer Notes

Please confirm that this still works for a-team's acceptance needs. 

## Setup

Create a new move on staging and get the moveCode

```sh
prime-api-demo MOVECD api.stg.move.mil
```

### DIFF
--- a/scripts/prime-api-demo
+++ b/scripts/prime-api-demo
@@ -29,7 +29,7 @@ function generatePRJson() {
   echo tmp/pad_create_payment_request.json
 }
 function generatePRJsonEmpty() {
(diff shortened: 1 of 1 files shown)


amanda5iu (approved):

pearl-truss (approved):
I love these changes!

PR #5995, opened 2021-02-25
MB-6571 Update relevant Office & Admin statuses when User status is updated

abbyoung:
## Description

This PR keeps the `office_users` and `admin_users` tables in sync with `users`. If a user is deactivated that is also an office or admin user, both `Active` statuses will update.

* `user_updater` service object now uses `admin_user_updater` and `office_user_updater` services objects to update as needed. By using these updaters, we can keep our error messages consistent.
* If there is an issue updating the `admin_users` or `office_users` tables, an error is logged. I decided to _not_ use a transaction, because if the code gets to this point, it means the user was successfully deactivated. Since one primary use case is security-related, we should always deactivate the user if we can.
* Two test functions added to create a new admin or office user that is active, and new tests added.

## Reviewer Notes

* Is the logging of errors around admin/office not updating sufficient?

## Setup

**Run tests**
`go test ./pkg/services/user --testify.m "TestUserUpdater" -v`

**Use UI to manuall test**
* Run the server and client, and go to http://adminlocal:3000
* Log in as an admin user. Go to the **Office Users** page and pick an active office user. Grab their email address.
* Go to the **Users** index page and search for the user by email. Edit the user and set Active to No. Save and confirm they are no longer active.
* Go back to the **Office Users** page and find the same user. Confirm their Active status is now false.
* Repeat this process for an Admin User.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6751) for this change

### DIFF
--- a/pkg/handlers/adminapi/api.go
+++ b/pkg/handlers/adminapi/api.go
@@ -42,6 +42,9 @@ func NewAdminAPIHandler(context handlers.HandlerContext) http.Handler {
 
 	adminAPI := adminops.NewMymoveAPI(adminSpec)
 	queryBuilder := query.NewQueryBuilder(context.DB())
+	officeUpdater := officeuser.NewOfficeUserUpdater(queryBuilder)
+	adminUpdater := adminuser.NewAdminUserUpdater(queryBuilder)
+
 	adminAPI.ServeError = handlers.ServeCustomError
 
 	adminAPI.OfficeUsersIndexOfficeUsersHandler = IndexOfficeUsersHandler{
@@ -129,7 +132,7 @@ func NewAdminAPIHandler(context handlers.HandlerContext) http.Handler {
 	adminAPI.UsersUpdateUserHandler = UpdateUserHandler{
 		context,
 		user.NewUserSessionRevocation(queryBuilder),
-		user.NewUserUpdater(queryBuilder),
+		user.NewUserUpdater(queryBuilder, officeUpdater, adminUpdater),
 		query.NewQueryFilter,
 	}
 
--- a/pkg/handlers/adminapi/api.go
+++ b/pkg/handlers/adminapi/api.go
@@ -70,7 +70,7 @@ func NewAdminAPIHandler(context handlers.HandlerContext) http.Handler {
 
 	adminAPI.OfficeUsersUpdateOfficeUserHandler = UpdateOfficeUserHandler{
 		context,
-		officeuser.NewOfficeUserUpdater(queryBuilder),
+		officeUpdater,
 		query.NewQueryFilter,
 		userRolesCreator,
 	}
@@ -150,7 +150,7 @@ func NewAdminAPIHandler(context handlers.HandlerContext) http.Handler {
 
 	adminAPI.AdminUsersUpdateAdminUserHandler = UpdateAdminUserHandler{
 		context,
-		adminuser.NewAdminUserUpdater(queryBuilder),
+		adminUpdater,
 		query.NewQueryFilter,
 	}
 
--- a/pkg/services/user/user_updater.go
+++ b/pkg/services/user/user_updater.go
@@ -71,9 +71,9 @@ func (o *userUpdater) UpdateUser(id uuid.UUID, user *models.User) (*models.User,
 			_, verrs, err = o.officeUserUpdater.UpdateOfficeUser(foundOfficeUser.ID, &payload)
 
 			if verrs != nil {
-				logger.Info("Could not update office user", zap.Error(verrs))
+				logger.Error("Could not update office user", zap.Error(verrs))
 			} else if err != nil {
-				logger.Info("Could not update office user", zap.Error(err))
+				logger.Error("Could not update office user", zap.Error(err))
 			}
 		}
 
@@ -87,9 +87,9 @@ func (o *userUpdater) UpdateUser(id uuid.UUID, user *models.User) (*models.User,
 			}
 			_, verrs, err = o.adminUserUpdater.UpdateAdminUser(foundAdminUser.ID, &payload)
 			if verrs != nil {
-				logger.Info("Could not update admin user", zap.Error(verrs))
+				logger.Error("Could not update admin user", zap.Error(verrs))
 			} else if err != nil {
-				logger.Info("Could not update admin user", zap.Error(err))
+				logger.Error("Could not update admin user", zap.Error(err))
 			}
 		}
 	}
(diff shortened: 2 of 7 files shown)


pkg/services/user/user_updater.go
```diff
@@ -43,6 +52,48 @@ func (o *userUpdater) UpdateUser(id uuid.UUID, user *models.User) (*models.User,
...
+	// If the update was successful and we are deactivating the user,
+	// update the office and admin statuses to match.
+
+	// Check if we are deactivating the user
+	if !user.Active {
+
+		// Check for Office User
+		foundOfficeUser := models.OfficeUser{}
+		filters = []services.QueryFilter{query.NewQueryFilter("user_id", "=", id.String())}
+		err = o.builder.FetchOne(&foundOfficeUser, filters)
```
abbyoung, line 64: I originally planned to do an `Eager` when the user is fetched at the top of the service object, but `FetchOne` can't grab associations. I decided to go ahead and make the query here, since Eager does do additional queries, anyway. So there is slightly more code, but it should work just as well.

pkg/services/user/user_updater.go
```diff
@@ -43,6 +52,48 @@ func (o *userUpdater) UpdateUser(id uuid.UUID, user *models.User) (*models.User,
...
+
+		// If we find a matching Office User, update their status
+		if err == nil {
+			payload := adminmessages.OfficeUserUpdatePayload{
+				Active: &user.Active,
+			}
+			_, verrs, err = o.officeUserUpdater.UpdateOfficeUser(foundOfficeUser.ID, &payload)
+
+			if verrs != nil {
```
abbyoung, line 73: I don't want to return these errors here, since the user has already been deactivated and that's the most important outcome of this function. Would love any input on these logs, or if there's additional logging/error handling we should do here!

pkg/testdatagen/make_admin_user.go
```diff
@@ -68,3 +68,15 @@ func MakeAdminUserWithNoUser(db *pop.Connection, assertions Assertions) models.A
 func MakeDefaultAdminUser(db *pop.Connection) models.AdminUser {
 	return MakeAdminUser(db, Assertions{})
 }
+
+// MakeActiveAdminUser makes an AdminUser that is active
+func MakeActiveAdminUser(db *pop.Connection) models.AdminUser {
```
abbyoung, line 73: Oddly enough, a default admin user is not set to active! I created this function in case that is intentional.

pkg/handlers/adminapi/api.go
```diff
@@ -129,7 +132,7 @@ func NewAdminAPIHandler(context handlers.HandlerContext) http.Handler {
 	adminAPI.UsersUpdateUserHandler = UpdateUserHandler{
 		context,
 		user.NewUserSessionRevocation(queryBuilder),
-		user.NewUserUpdater(queryBuilder),
+		user.NewUserUpdater(queryBuilder, officeUpdater, adminUpdater),
```
abbyoung, line 135: I decided to use these two service objects instead of rewriting the code/query. It ended up being slightly more code, but I do think it'll be helpful to use the same function for the same purpose, and keep our error messages aligned (once we actually fix those.)

pkg/services/user/user_updater_test.go
```diff
@@ -31,6 +37,66 @@ func (suite *UserServiceSuite) TestUserUpdater() {
...
+		// Check that there are no errors and both statuses successfully updated
+		suite.Nil(verr)
+		suite.Nil(err)
+		suite.False(updatedOfficeUser.Active)
+		suite.False(updatedUser.Active)
+
+	})
+
+	suite.T().Run("Deactivate an Admin User successfully", func(t *testing.T) {
+		// Create an active office user
```
abbyoung, line 71: boop need to update my comments here 🔍

pkg/handlers/adminapi/api.go
sandy-wright: Since you added the `officeUpdater` and `adminUpdater` variables, you could also use them in `UpdateOfficeUserHandler` (line 73) and `UpdateAdminUserHandler` (line 153) for consistency's sake

pkg/services/user/user_updater.go
```diff
@@ -3,25 +3,34 @@ package user
...
+		officeUserUpdater,
+		adminUserUpdater,
+	}
 }
 
 // UpdateUser updates any user
 func (o *userUpdater) UpdateUser(id uuid.UUID, user *models.User) (*models.User, *validate.Errors, error) {
 	filters := []services.QueryFilter{query.NewQueryFilter("id", "=", id.String())}
 	var foundUser models.User
+	var logger Logger
```
sandy-wright, line 33: _nice_

pkg/services/user/user_updater.go
sandy-wright: hmmm, I was thinking at first that all of these actions should be in the same transaction, but it really does make sense to keep them separate because these updates are just tidying up the database, they're not functional. I just wish there was a way to return a response that was like a "Yes, but..." so the user could see that the main action succeeded, but this action didn't :(

Maybe these logs should be more prominent than `Info` logs? We don't return the error, but I think putting it at the `logger.Error` level is still justified here

sandy-wright: _`logger.Error` doesn't cause any errors, right?_ 🤔

sandy-wright (approved):
LGTM! Left a couple of comments, but they're minor. This works the way we want it to!

pkg/handlers/adminapi/api.go
abbyoung: OOOH will do!!

pkg/services/user/user_updater.go
abbyoung: That was my first thought, but once I read through our logging docs, it sounds like we shouldn't use logger.Error for this, and primarily be using logger.Info? https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#logging

But I'm with you. I want to make sure these are being surfaced in a meaningful/helpful way.

abbyoung: Re-reading that section, now I'm not sure. It _would_ be a failed db update, and it wouldn't be because of a bad id at that point. Perhaps error is ok?

sandy-wright: yeah, I wasn't sure after reading that either! My interpretation is that `Error` should be used for failures that weren't caused by user error, which applies in this case. This would be a strange and unexpected system failure, so we'd probably want to know so we could find the cause before it affected something more critical. I would defend it in logging court

abbyoung: DONE! I will be your co-defense.

PR #5997, opened 2021-02-25
[MB-6805] create utility function that creates new payment service params for pricers

pearl-truss:
## Description

This PR creates the utility function that will be used by pricers to create the relevant payment service item params. This PR builds on [MB-6695](https://dp3.atlassian.net/browse/MB-6695)

## Reviewer Notes

I don't think there's anything to except yet since no pricer is using this yet. However, please make sure the function itself is correctly creating and saving the payment service item params. And that this function is adequately tested 

## Setup

```sh
make server_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6805) for this change
* [Jira story that this ticket builds on](https://dp3.atlassian.net/browse/MB-6695) for this change
* [discover doc](https://docs.google.com/document/d/1m2GApS48Vgxi6roa_do3wNBz4V8cyCD73OMkDaAPjkA/edit?ts=602fd09d#) explains more about the approach used.


jacquelineIO (approved):
lgtm

pkg/services/ghcrateengine/pricer_helpers.go
```diff
@@ -158,3 +159,33 @@ func priceDomesticPickupDeliverySIT(db *pop.Connection, pickupDeliverySITCode mo
...
+			Where("origin = ?", models.ServiceItemParamOriginPricer).
+			First(&serviceItemParamKey)
+		if err != nil {
+			return paymentServiceItemParams, fmt.Errorf("unable to find service item param key for %v", serviceItemParamKey.Key)
+		}
+
+		newParam := models.PaymentServiceItemParam{
+			PaymentServiceItemID:  paymentServiceItemID,
+			ServiceItemParamKeyID: serviceItemParamKey.ID,
+			Value:                 fmt.Sprintf("%v", param.Value),
```
reggieriser, line 178: I'm not sure that this is going to give us the desired format always depending on the underlying type of `param.Value`.  It probably works fine for basic types like integers and booleans, but what about `time.Time` values we want to treat as dates, for instance?  The current pricer param types are DECIMAL/STRING/BOOLEAN so that may not apply for now, but that may not always be the case.  May want to type assert at least on DATE and TIMESTAMP to write those out in the same format we write out other similar params.

pkg/services/ghcrateengine/pricer_helpers.go
```diff
@@ -158,3 +159,33 @@ func priceDomesticPickupDeliverySIT(db *pop.Connection, pickupDeliverySITCode mo
...
 	return totalPriceCents, nil, nil
 }
+
+func createPricerGeneratedParams(db *pop.Connection, paymentServiceItemID uuid.UUID, params services.PricingParams) (models.PaymentServiceItemParams, error) {
+	var paymentServiceItemParams models.PaymentServiceItemParams
+	for _, param := range params {
+		var serviceItemParamKey models.ServiceItemParamKey
+		err := db.Q().
+			Where("key = ?", param.Key).
+			Where("origin = ?", models.ServiceItemParamOriginPricer).
```
reggieriser, line 169: The `key` is unique in this table, so you don't necessarily need to look at the `origin` field unless you're just trying to ensure it's a pricer param.  If that's the case, maybe we have two different error conditions here to clarify what's going wrong -- one if the key can't be found, but then if we find it, check the origin to make sure it's a pricer param.

pkg/services/ghcrateengine/pricer_helpers_test.go
```diff
@@ -174,3 +177,96 @@ func (suite *GHCRateEngineServiceSuite) Test_priceDomesticPickupDeliverySIT50Mil
...
 }
+
+func (suite *GHCRateEngineServiceSuite) Test_createPricerGeneratedParams() {
+	params := services.PricingParams{
+		{
+			Key:   models.ServiceItemParamNamePriceRateOrFactor,
+			Value: 40000,
+		}, {
+			Key:   models.ServiceItemParamNameEscalationCompounded,
+			Value: 1,
```
reggieriser, line 188: Nit: Maybe make the two values above be something with a decimal point since that's what most of the real data will be like.

pkg/services/ghcrateengine/pricer_helpers_test.go
```diff
@@ -174,3 +177,96 @@ func (suite *GHCRateEngineServiceSuite) Test_priceDomesticPickupDeliverySIT50Mil
...
+
+	paymentServiceItem := testdatagen.MakePaymentServiceItem(
+		suite.DB(),
+		testdatagen.Assertions{},
+	)
+
+	suite.T().Run("payment service item params created for the pricer", func(t *testing.T) {
+		paymentServiceItemParams, err := createPricerGeneratedParams(suite.DB(), paymentServiceItem.ID, params)
+		suite.NoError(err)
+		suite.Equal(len(paymentServiceItemParams), 4)
```
reggieriser, line 239: I think we should put some assertions on the values in the four params to make sure they are translating to strings in the way we expect.

pkg/services/ghcrateengine/pricer_helpers_test.go
```diff
@@ -174,3 +177,96 @@ func (suite *GHCRateEngineServiceSuite) Test_priceDomesticPickupDeliverySIT50Mil
...
+				Type:        models.ServiceItemParamTypeString,
+				Origin:      models.ServiceItemParamOriginPrime,
+			},
+		})
+
+		_, err := createPricerGeneratedParams(suite.DB(), paymentServiceItem.ID, invalidParam)
+		suite.Error(err)
+		suite.Contains(err.Error(), "unable to find service item param key for ")
+	})
+	// TODO: test that a validation error is thrown for a bad key or value
```
reggieriser, line 271: Is that TODO covered by the test above or is it still needed?

reggieriser (commented):
Looking good!  I added a few comments -- mainly concerned about ensuring the format of the parameter's value in the database is being string-ified in a consistent way.

Also, do we need a story to wire this function into the appropriate spot (probably in `PriceServiceItem`) and ensuring pricing params come back as part of the create payment request return payload?  I had been thinking we may do that in this PR, but if that's not the case, I don't want to lose that need.

pkg/services/ghcrateengine/pricer_helpers_test.go
```diff
@@ -174,3 +177,96 @@ func (suite *GHCRateEngineServiceSuite) Test_priceDomesticPickupDeliverySIT50Mil
...
+}
```
reggieriser, line 272: One other testing thought -- do we need a test to see what happens if a nil or empty slice is passed into the function?

pkg/services/ghcrateengine/pricer_helpers.go
pearl-truss: I'm not sure I'm following. The model for PaymentServiceItemParam expects the value field to be a string. Would I still need to type assert for DATE and TIMESTAMP?

reggieriser: I made a quick [Go Playground](https://play.golang.org/p/tFrrmGMk2hy) that may help demonstrate what I'm talking about regarding the string format.  Basically, I just want to make sure that any string conversions we do here matches what's done in other places (like the lookups) for all the types.  Let's walk through it in real-time if it's still unclear.

pkg/services/ghcrateengine/pricer_helpers_test.go
pearl-truss: I think this is covered

pkg/services/ghcrateengine/pricer_helpers_test.go
pearl-truss: I missed this, working on a commit now

pkg/services/ghcrateengine/pricer_helpers.go
pearl-truss: I think this is resolved now

pkg/services/ghcrateengine/pricer_helpers.go
```diff
@@ -166,16 +166,31 @@ func createPricerGeneratedParams(db *pop.Connection, paymentServiceItemID uuid.U
...
+		if serviceItemParamKey.Origin != models.ServiceItemParamOriginPricer {
+			return paymentServiceItemParams, fmt.Errorf("Service item param key is not a pricer param. Param key: %v", serviceItemParamKey.Key)
+		}
+
+		value := param.Value
+		switch serviceItemParamKey.Type {
+		case models.ServiceItemParamTypeTimestamp:
+			if timestampValue, ok := param.Value.(time.Time); ok {
+				value = timestampValue.Format(TimestampParamFormat)
+			}
```
reggieriser, line 187: Might want to have an `else` part of this (and the other one like it below) that throws an error in case it's not a `time.Time` for some reason.  Otherwise, it falls through and writes out the value in whatever default format applies for the type.

pkg/services/ghcrateengine/pricer_helpers_test.go
```diff
@@ -236,7 +236,10 @@ func (suite *GHCRateEngineServiceSuite) Test_createPricerGeneratedParams() {
 	suite.T().Run("payment service item params created for the pricer", func(t *testing.T) {
 		paymentServiceItemParams, err := createPricerGeneratedParams(suite.DB(), paymentServiceItem.ID, params)
 		suite.NoError(err)
-		suite.Equal(len(paymentServiceItemParams), 4)
+		expectedValues := [4]string{"40000.9", "1.06", "true", "TRUSS_TEST"}
+		for i, paymentServiceItemParam := range paymentServiceItemParams {
+			suite.Equal(expectedValues[i], paymentServiceItemParam.Value)
+		}
```
reggieriser, line 242: Nit: This isn't a blocker, but just pointing out that this assumes the `paymentServiceItemParams` come back in the same order as the key/value pairs that are passed.  I believe they are now based on the implementation, but that's not necessarily guaranteed.   Could be "safer" to look in the param slice by key, then compare the value on the found row (that would also have the benefit of making sure the keys were the correct ones too).

reggieriser (approved):
LGTM! 🚢

PR #6003, opened 2021-02-26
MB-6953 Remove MacOS version check in CAC scripts

felipe-lee:
## Description

I noticed that `make prereqs` doesn't work if you have MacOS > 10.15. I checked with infra and it seems we're all on 10.15 or greater now so I removed the version checks in the CAC scripts.

## Reviewer Notes

It looks like the location of `opensc-pkcs11.so` can vary by `opensc` versions. The docs don't really explain which location to expect. Based on some testing, it _looks_ like the `/usr/local/lib/opensc-pkcs11.so` might be best to use since we're updating to use the `--cask` version. That's why I set the default for the gen certs command to use that path. I let the bash scripts use either path. Lmk if you think I should set up the gen certs command to check both paths too. 

I also moved the call to `cac-prereqs` in each of the scripts to the top so that we can have one place that checks that `opensc-pkcs11.so` can be found at all and throws an error when not. It is also now being sourced so that we can use the vars that it sets in each of the CAC scripts rather than setting them again in every file. 

The change to sourcing the `cac-prereqs` script meant I needed to add `shellcheck` directives so it knew where the scripts were, but this then led to me needing to update the `pre-commit` config to pass `-x` as an arg to `shellcheck` so it would follow the path to find the scripts. Lmk if we shouldn't do this though. Alternatively we can tell shellcheck to not worry about the path, or we can go back to a version where we don't source `cac-prereqs`.

I also updated the `aws-vault` and `chamber` versions that are checked for in the scripts called by `make prereqs` since that's the command that set off this work in the first place and the goal is to get it working again.

## Setup

Check if the scripts work for you, e.g.:

```shell
cac-prereqs
```

You can also check at a higher level:

```shell
make prereqs
```

Note that if you have the file `.prereqs.stamp` you'll have to remove it first.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6953) for this change


scripts/cac-extract-cert
```diff
@@ -6,14 +6,7 @@ set -eu -o pipefail
...
 
-OS_VERSION=$(/usr/bin/sw_vers -productVersion)
-MAC_OS="10.15"
-if [[ $OS_VERSION == *$MAC_OS* ]]; then
-  MODULE=/usr/local/lib/pkcs11/opensc-pkcs11.so
-elif [[ -f "${MODULE}" ]]; then
-  MODULE=/usr/local/lib/pkcs11/cackey.dylib
-fi
-readonly MODULE
+readonly MODULE=/usr/local/lib/pkcs11/opensc-pkcs11.so
```
travelar, line 9: Depending on the version of opensc, this path is not valid. Mine is located in /usr/local/lib/opensc-pkcs11.so.

felipe-lee: I finally got a CAC reader...so barely getting back to this. It looks like I have both, and both seem to be links to the true version installed by homebrew. I have version 0.21.0. 

I wonder if it'd be better to switch it, or have it check both paths. I guess since I don't know which direction this is going in, like if they're switching from one path to the other, or just switching to having both.

travelar (approved):

scripts/cac-extract-cert
```diff
@@ -6,18 +6,17 @@ set -eu -o pipefail
 # Get a certificate from CAC
 #
 
-OS_VERSION=$(/usr/bin/sw_vers -productVersion)
-MAC_OS="10.15"
-if [[ $OS_VERSION == *$MAC_OS* ]]; then
```
gidjin, line 11: Is it better to leave this in with the correct version for the next major upgrade of mac os? Maybe it's not worth it.

scripts/cac-extract-cert
```diff
@@ -6,18 +6,17 @@ set -eu -o pipefail
...
+# Check the CAC Pre-Requirements
+cac-prereqs
+
+if [[ -f /usr/local/lib/opensc-pkcs11.so ]]; then
+  MODULE=/usr/local/lib/opensc-pkcs11.so
+elif [[ -f /usr/local/lib/pkcs11/opensc-pkcs11.so ]]; then
   MODULE=/usr/local/lib/pkcs11/opensc-pkcs11.so
```
gidjin, line 15: instead of repeating this if statement in every cac file I wonder if it can be moved into `cac-prereqs`. You might have to export the setting of module.

gidjin (approved):
Changes look good. I made a suggestion on moving the if statement if possible. I wouldn't mind keeping the mac os version check if it can be fixed but I'm okay if that's not possible. Everything seemed to work properly for me as well.

monfresh:
When I run `brew info opensc`, I get this Caveat:

```
==> Caveats
The OpenSSH PKCS11 smartcard integration will not work from High Sierra
onwards. If you need this functionality, unlink this formula, then install
the OpenSC cask.
```
I don't know if this applies to us, but if it does, we'll want to change the instructions for installing opensc to use the cask:

```
brew install --cask opensc
```

I verified that installing the cask installs both `pkcs11-tool` and `pkcs15-tool`  in `/usr/local/bin`.

scripts/cac-extract-cert
felipe-lee: The main reason I took it out was that it seems that `cackey.dylib` isn't supported after 10.14 so I didn't think we'd need it anymore.

scripts/cac-extract-cert
felipe-lee: Ah true, I wasn't sure if it mattered if I left it around, but I guess we could also unset it if becomes an issue for some reason. I'll update it to do that and test it out. I'd really like to avoid the repeated `if` statements.

felipe-lee:
> When I run `brew info opensc`, I get this Caveat:
> 
> ```
> ==> Caveats
> The OpenSSH PKCS11 smartcard integration will not work from High Sierra
> onwards. If you need this functionality, unlink this formula, then install
> the OpenSC cask.
> ```
> 
> I don't know if this applies to us, but if it does, we'll want to change the instructions for installing opensc to use the cask:
> 
> ```
> brew install --cask opensc
> ```
> 
> I verified that installing the cask installs both `pkcs11-tool` and `pkcs15-tool` in `/usr/local/bin`.

That's a good question. Looking back at the [original issue](https://github.com/Homebrew/homebrew-core/issues/44786) that led to this message being added, it looks like it fails because of symlinks. I wonder if that's why we're doing all the work around figuring out the path to the module. I'll test a few things out to see if maybe we can switch to using the cask version and then not have to specify the path to the module each time.

.pre-commit-config.yaml
```diff
@@ -71,6 +71,7 @@ repos:
     rev: 1.0.5
     hooks:
       - id: shell-lint
+        args: [-x]
```
felipe-lee, line 74: I added this because `shellcheck` didn't like that I was sourcing the `cac-prereqs` file from the other cac scripts without telling it where the file was. But even once I added the directive telling it where to find things, it still wanted this flag set. If there's a reason not to have this flag, we can either add a directive telling `shellcheck` to ignore the lack of `cac-prereqs` source, or we can go back to the old way of each file having their own vars that they set.

travelar (approved):

PR #6020, opened 2021-03-02
MB-6917 Add PaymentServiceItemParams to GHC API

monfresh:
## Description

A TIO needs to see how weights factor into the pricing of service items
so that they can understand how the price was determined.

This information comes from the `payment_service_item_params` and
`service_item_param_keys` tables.

The GHC YAML file already included the necessary fields in the
definition of the `getPaymentRequestsForMove` endpoint via the
definitions for
PaymentRequest->PaymentServiceItem->PaymentServiceItemParam

To add these to the payload, I copied the functions from the Prime API,
and then I updated the `FetchPaymentRequestListByMove` function to
perform the necessary joins and eager loading to fetch the data from the
DB.

I used TDD to write a failing test first, and converted the existing
`TestGetPaymentRequestsForMoveHandler` test to an integration test that
hits the DB, as opposed to using mocks. To figure out how to properly
set up the DB with all the necessary data, I looked up existing usage of
`paymentServiceItemParams` in our codebase, and found one in the Prime
API's `pkg/handlers/primeapi/move_task_order_test.go`. I copied that
over and added tests for the `Key` and `Value` fields in the payload.

It's not clear from the JIRA issue what the payload needs to contain for
the client to display the information to the TIO. If it needs the full
paylod as defined in the `PaymentServiceItemParam` section of the YAML,
then I'll add more tests for the rest of the payload.

## Reviewer Notes

Note that this PR only implements the backend portion of this feature. There is a separate unscheduled story (in TTV's March Slice) for the [frontend portion](https://dp3.atlassian.net/browse/MB-4800).

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6917) for this change

## Screenshots
Screen Shot 2021-03-04 at 1 06 38 PM

### DIFF
--- a/pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
+++ b/pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
@@ -390,18 +390,24 @@ func PaymentRequest(pr *models.PaymentRequest, storer storage.FileStorer) (*ghcm
 
 // PaymentServiceItem payload
 func PaymentServiceItem(ps *models.PaymentServiceItem) *ghcmessages.PaymentServiceItem {
+	if ps == nil {
+		return nil
+	}
+	paymentServiceItemParams := PaymentServiceItemParams(&ps.PaymentServiceItemParams)
+
 	return &ghcmessages.PaymentServiceItem{
-		ID:                 *handlers.FmtUUID(ps.ID),
-		MtoServiceItemID:   *handlers.FmtUUID(ps.MTOServiceItemID),
-		MtoServiceItemName: ps.MTOServiceItem.ReService.Name,
-		MtoShipmentType:    ghcmessages.MTOShipmentType(ps.MTOServiceItem.MTOShipment.ShipmentType),
-		MtoShipmentID:      handlers.FmtUUIDPtr(ps.MTOServiceItem.MTOShipmentID),
-		CreatedAt:          strfmt.DateTime(ps.CreatedAt),
-		PriceCents:         handlers.FmtCost(ps.PriceCents),
-		RejectionReason:    ps.RejectionReason,
-		Status:             ghcmessages.PaymentServiceItemStatus(ps.Status),
-		ReferenceID:        ps.ReferenceID,
-		ETag:               etag.GenerateEtag(ps.UpdatedAt),
+		ID:                       *handlers.FmtUUID(ps.ID),
+		MtoServiceItemID:         *handlers.FmtUUID(ps.MTOServiceItemID),
+		MtoServiceItemName:       ps.MTOServiceItem.ReService.Name,
+		MtoShipmentType:          ghcmessages.MTOShipmentType(ps.MTOServiceItem.MTOShipment.ShipmentType),
+		MtoShipmentID:            handlers.FmtUUIDPtr(ps.MTOServiceItem.MTOShipmentID),
+		CreatedAt:                strfmt.DateTime(ps.CreatedAt),
+		PriceCents:               handlers.FmtCost(ps.PriceCents),
+		RejectionReason:          ps.RejectionReason,
+		Status:                   ghcmessages.PaymentServiceItemStatus(ps.Status),
+		ReferenceID:              ps.ReferenceID,
+		ETag:                     etag.GenerateEtag(ps.UpdatedAt),
+		PaymentServiceItemParams: *paymentServiceItemParams,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -415,6 +421,38 @@ func PaymentServiceItems(paymentServiceItems *models.PaymentServiceItems) *ghcme
...
+// PaymentServiceItemParams payload
+func PaymentServiceItemParams(paymentServiceItemParams *models.PaymentServiceItemParams) *ghcmessages.PaymentServiceItemParams {
+	if paymentServiceItemParams == nil {
+		return nil
+	}
+
+	payload := make(ghcmessages.PaymentServiceItemParams, len(*paymentServiceItemParams))
+
+	for i, p := range *paymentServiceItemParams {
+		copyOfP := p // Make copy to avoid implicit memory aliasing of items from a range statement.
```
jacquelineIO, line 450: If you change the signature of the function `func PaymentServiceItemParam(paymentServiceItemParam *models.PaymentServiceItemParam)` to NOT use `*` then you shouldn't have to do this `copyOfP := p`. 

This was done in places where the code was going to be a lot to refactor, but for new code the `copyOfP` shouldn't have to be used.

pkg/services/payment_request/payment_request_list_fetcher.go
```diff
@@ -137,13 +137,22 @@ func (f *paymentRequestListFetcher) FetchPaymentRequestListByMove(officeUserID u
...
 		InnerJoin("moves", "payment_requests.move_id = moves.id").
 		InnerJoin("orders", "orders.id = moves.orders_id").
 		InnerJoin("service_members", "orders.service_member_id = service_members.id").
 		InnerJoin("contractors", "contractors.id = moves.contractor_id").
 		InnerJoin("duty_stations", "duty_stations.id = orders.origin_duty_station_id").
 		InnerJoin("transportation_offices", "transportation_offices.id = duty_stations.transportation_office_id").
+		InnerJoin("payment_service_items", "payment_requests.id = payment_service_items.payment_request_id").
+		InnerJoin("mto_service_items", "payment_service_items.mto_service_item_id = mto_service_items.id").
+		InnerJoin("payment_service_item_params", "payment_service_items.id = payment_service_item_params.payment_service_item_id").
+		InnerJoin("service_item_param_keys", "payment_service_item_params.service_item_param_key_id = service_item_param_keys.id").
```
jacquelineIO, line 155: Why do you need these additional joins? My memory about this is not good. I did play around a bunch with this when I did the uploads refactor and I don't exactly remember what the answer is, but I thought that you only needed the joins if you were using the additional columns in a WHERE clause. 

I'm curious if you have tried using this w/o the additional joins?

monfresh: Great question. I had the same initial thought, and I thought it didn't work without the joins, but I'm probably misremembering because I just tried it and it's working.

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
monfresh: Good to know! I will add a comment in the Prime API to say we shouldn't copy this pattern.

monfresh: If I remove the `copyOf` stuff, golangci-lint will complain and won't let you commit.

jacquelineIO: Yes, you have to do two things, which is:
* removed the copy
* update the signature of the function accepting a `*` to not be a `*` which will force a copy 


This is referenced at the bottom of the doc
> Another option that achieves a similar goal would be to change the called function/method's signature to accept a value type rather than a reference:

monfresh: Ah. I see what you mean now by "a lot to refactor". Both are valid options, but one could potentially require a lot of changes. That document doesn't seem to have a strong opinion on which one to choose and whether or not we want to be consistent. It does say that the copy option is the most straightforward.

If we want to be consistent, we should pick one and stick with it. Otherwise, any working option should be fine, no?

monfresh: I can also see the argument that since this is new code, we could make it less verbose to begin with. I'll make the change.

jacquelineIO: Yes, leaving as-is is fine. 

`copyOfP := p` is really a fix to a bad design IMO (us not understanding how golang works). The document focuses on the fix.

pkg/handlers/ghcapi/payment_request_test.go
```diff
@@ -117,52 +117,54 @@ func (suite *HandlerSuite) TestFetchPaymentRequestHandler() {
 }
 
 func (suite *HandlerSuite) TestGetPaymentRequestsForMoveHandler() {
-	prUUID, _ := uuid.NewV4()
-	expectedServiceItemName := "Test Service Item Name"
+	expectedServiceItemName := "Test Service"
```
duncan-truss, line 120: I know this was already there but can we use a realist service item eventhough this is the one created by default?

I'm seeing more realistic test data here https://github.com/transcom/mymove/blob/08143cb0f9128ec54b4f039d8f844cf22c97e92b/pkg/services/invoice/ghc_payment_request_invoice_generator_test.go#L118

Could we put an example in the devseed too?

Ryan-Koch:
This is looking good. I think for this to show up on the request we're making on the payment request overview that we'll also need to reference the `PaymentServiceItems` that you've updated in the `PaymentRequest` payload.

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -390,18 +390,24 @@ func PaymentRequest(pr *models.PaymentRequest, storer storage.FileStorer) (*ghcm
...
+		MtoServiceItemName:       ps.MTOServiceItem.ReService.Name,
+		MtoShipmentType:          ghcmessages.MTOShipmentType(ps.MTOServiceItem.MTOShipment.ShipmentType),
+		MtoShipmentID:            handlers.FmtUUIDPtr(ps.MTOServiceItem.MTOShipmentID),
+		CreatedAt:                strfmt.DateTime(ps.CreatedAt),
+		PriceCents:               handlers.FmtCost(ps.PriceCents),
+		RejectionReason:          ps.RejectionReason,
+		Status:                   ghcmessages.PaymentServiceItemStatus(ps.Status),
+		ReferenceID:              ps.ReferenceID,
+		ETag:                     etag.GenerateEtag(ps.UpdatedAt),
+		PaymentServiceItemParams: *paymentServiceItemParams,
```
duncan-truss, line 410: Will basic service items have payment service item params?  This looks like it could be a nil pointer error if we try to get the value of a nil result.

pkg/handlers/ghcapi/payment_request_test.go
monfresh: Done. Devseed already had a payment request with item params. It's the one with the Move ID "7cbe57ba-fd3a-45a7-aa9a-1970f1908ae7". I changed the Move's status to submitted so that it appears in the TOO queue.

monfresh:
@Ryan-Koch I think this is the only Move so far in Devseed that has a payment request with service item params: Move ID "7cbe57ba-fd3a-45a7-aa9a-1970f1908ae7"

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
monfresh: This seems to be working fine. I clicked through all payment requests in Dev until I found one with 2 basic service items (Move Management and Counseling) and it loaded fine. No errors in the response or the server logs.

monfresh: The Prime API is doing the same thing.

Ryan-Koch:
> @Ryan-Koch I think this is the only Move so far in Devseed that has a payment request with service item params: Move ID "7cbe57ba-fd3a-45a7-aa9a-1970f1908ae7"

Thanks for the ID! I see the things included in the payload now for that move.

Ryan-Koch (approved):

PR #6024, opened 2021-03-02
[MB-6844] Use Redirect for SignIn

suzubara:
## Description

This PR makes some changes to the Customer app entry point (`MyMove/index.jsx`) in order to follow a closer pattern to the Office app (`Office/index.jsx`) with regards to user gating (i.e., checking for a logged in user before rendering a route, and redirecting to the Sign In page if not logged in).

- **Moved (and renamed) `src/shared/User/ValidatedPrivateRoute` to `src/containers/CustomerPrivateRoute`:** I had to relocate in order to add a test file. Other than returning a `<Redirect />` for the SignIn/AccessCode pages and displaying the loader if the user data is still loading, there are not really any substantial code changes to this component.
  - I named it `CustomerPrivateRoute` to distinguish from the `PrivateRoute` component used by the Office. I did consider whether both apps could use the same `PrivateRoute` component, but opted to keep them separate for now, since the Office app does some additional user role validation, and the Customer app currently checks for an access code to prevent just anyone from using it.
- **Removed user gating from the Customer app Home page & PPM landing pages:** This is no longer necessary, since both of these routes (`/, /ppm`) now use the `CustomerPrivateRoute` component, and will never be rendered until a user has already been loaded.

## Reviewer Notes

Primary things to test here are:
- you can log into the customer app as expected and access logged-in routes such as `/` and `/ppm`
- if you are not logged into the customer app, trying to access any logged in routes such as `/` or `/ppm` redirects you to the Sign In page (`/sign-in`)

## Setup

To run the app locally:
```
make server_run
make client_run
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6844) for this change


src/App/index.jsx
```diff
@@ -76,9 +76,11 @@ const App = () => {
   return (
     <Provider store={store}>
       <AppContext.Provider value={myMoveContext}>
-        <Suspense fallback={<LoadingPlaceholder />}>
-          <MyMove />
-        </Suspense>
+        <ConnectedRouter history={history}>
```
suzubara, line 79: This is just moving the `<ConnectedRouter>` from `MyMove/index` to `App/index` so effectively should result in no change (this is the same pattern the Office app uses).

src/scenes/MyMove/index.jsx
```diff
@@ -1,18 +1,36 @@
...
+import { loadUser as loadUserAction } from 'store/auth/actions';
+import { initOnboarding as initOnboardingAction } from 'store/onboarding/actions';
+import { selectConusStatus } from 'store/onboarding/selectors';
+import {
+  selectServiceMemberFromLoggedInUser,
+  selectCurrentMove,
+  selectHasCanceledMove,
+  selectMoveType,
+} from 'store/entities/selectors';
+/** Pages */
```
suzubara, line 33: I just did some reordering/reorg of these imports, to split out all of the "page" components (ie, anything rendered at a specific route) - the idea being that these can all be lazy loaded (see SignIn & AccessCode examples below)

src/scenes/MyMove/index.jsx
```diff
@@ -90,16 +97,18 @@ export class AppWrapper extends Component {
 
   render() {
     const props = this.props;
-    const Tag = detectIE11() ? 'div' : 'main';
```
kimallen, line 93: Assuming this was written because it was necessary for IE11 originally, is this removed because `main` now supported by IE11?

src/scenes/MyMove/index.jsx
kimallen: Thx!  This note is helpful for understanding what change happened here.

YanZ777 (approved):

src/scenes/MyMove/index.jsx
suzubara: technically `main` isn't supported by IE11, however using it as an HTML tag will not break anything ([see MDN's docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main#browser_compatibility)). We made the same change on the Office app, but basically IE11 will default to `main` acting as a `div` so this check was not necessary.

kimallen (approved):
Looks good!

mariaclaratgolfo (approved):

PR #6030, opened 2021-03-03
[MB-6578] UpdateWebhookSubscription endpoint

shimonacarvalho:
## Description

This is part of the Push Deploy work to add a new handler to update webhook subscriptions.

## Reviewer Notes

Please play close attention to the tests for UpdateWebhookSubscription Handler and Service Object

**The endpoint show allow updating of the following fields:**
subscriber_id - not the id of the subscription, but links to which contractor this subscription is related to status (ACTIVE, FAILING, DISABLED), event_key, callback_url, severity

## Setup
To test the endpoint in Postman
1. Add a new endpoint in postman for updateWebhookSubscription
2. Make sure that you are making a PATCH request with url: {{baseUrl}}/webhook_subscriptions/:webhookSubscriptionID
3. Construct a body similar to this example (this was generated by redoc)
```
{
  "subscriberId": "d494f114-05a2-4b39-840c-3d33243b7e29",
  "status": "ACTIVE",
  "callbackUrl": "string",
  "severity": 0,
  "eventKey": "string",
}
```

3. Follow [these instructions] (https://github.com/transcom/mymove/wiki/Test-Admin-and-Office-APIs-with-Postman) for copying the correct X-CSRF token into your Header in Postman
4. You should be able to send the request and receive a 200 response. Try some other variants to make sure they work.

To run tests

```sh
go test ./pkg/services/webhook_subscription -testify.m TestWebhookSubscriptionUpdater -failfast  -v
go test ./pkg/handlers/adminapi -testify.m TestWebhookSubscriptionUpdater -failfast  -v

```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6578) for this change


pkg/handlers/adminapi/webhook_subscriptions.go
```diff
@@ -147,3 +152,45 @@ func (h CreateWebhookSubscriptionHandler) Handle(params webhooksubscriptionop.Cr
 	returnPayload := payloadForWebhookSubscriptionModel(*createdWebhookSubscription)
 	return webhooksubscriptionop.NewCreateWebhookSubscriptionCreated().WithPayload(returnPayload)
 }
+
+// UpdateWebhookSubscriptionHandler returns an updated webhook subscription via PUT
```
abbyoung, line 156: small thing, but should this say PATCH?

abbyoung: or is this referencing something else?

pkg/services/webhook_subscription/webhook_subscription_updater.go
```diff
@@ -0,0 +1,72 @@
...
+	verrs, err := o.builder.UpdateOne(&foundSub, nil)
+
+	if verrs != nil && verrs.HasAny() {
+		return nil, services.NewInvalidInputError(webhookSubscriptionID, err, verrs, "")
+	}
+
+	// First check to see if there is an error on the type and return a precondition fail error, if not return the error
+	if err != nil {
+		switch err.(type) {
+		case query.StaleIdentifierError:
```
abbyoung, line 60: Do we need this if we are not using etags? I suppose if we're going to implement them in a future story it's ok?

pkg/handlers/adminapi/webhook_subscriptions.go
```diff
@@ -147,3 +152,45 @@ func (h CreateWebhookSubscriptionHandler) Handle(params webhooksubscriptionop.Cr
...
+	services.WebhookSubscriptionUpdater
+	services.NewQueryFilter
+}
+
+// Handle updates a webhook subscription
+func (h UpdateWebhookSubscriptionHandler) Handle(params webhooksubscriptionop.UpdateWebhookSubscriptionParams) middleware.Responder {
+	logger := h.LoggerFromRequest(params.HTTPRequest)
+	payload := params.WebhookSubscription
+
+	// Checks that ID in body matches ID in query
```
abbyoung, line 168: I'm still unclear why sometimes we pass the id in the body AND the path, and sometimes we don't. In this case my read is that because we're reusing the WebhookSubscription definition in the yaml, and it has the id as one of its properties. Still feels weird though!

pkg/handlers/adminapi/webhook_subscriptions.go
```diff
@@ -147,3 +152,45 @@ func (h CreateWebhookSubscriptionHandler) Handle(params webhooksubscriptionop.Cr
...
+	payloadID := uuid.FromStringOrNil(payload.ID.String())
+	if payloadID != uuid.Nil && params.WebhookSubscriptionID != payload.ID {
+		return webhooksubscriptionop.NewUpdateWebhookSubscriptionUnprocessableEntity()
+	}
+
+	// If no ID in body, use query ID
```
abbyoung, line 174: Oh perhaps I should've KEPT READING.

pkg/services/webhook_subscription/webhook_subscription_updater.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/transcom/mymove/pkg/services/query"
+)
+
+type webhookSubscriptionUpdater struct {
+	builder webhookSubscriptionQueryBuilder
+}
+
+// UpdateWebhookSubscription updates a webhook subscription
+// It uses the id in the passed in model to find the subscription.
+// For the severity field, it uses severity from the parameter not the model. If nil, severity will not be updated.
```
abbyoung, line 17: Thank you for these very useful comments!

pkg/services/webhook_subscription/webhook_subscription_updater.go
abbyoung: Never mind -- it's just the admin api that doesn't have etags yet, and this is the service object. Plus, we'll be adding them.

abbyoung (approved):
lgtm! tested and works 🎊 left a few small questions, mostly for my own knowledge. it looks like you'll need to recheck in the gen files, but other that, :shipit:

pkg/handlers/adminapi/webhook_subscriptions.go
shimonacarvalho: Nope that should be PATCH. Will fix.

pkg/services/webhook_subscription/webhook_subscription_updater.go
shimonacarvalho: Yeah since this is the service object, we left the check there so it could be used by another endpoint. We put a comment in the handler saying etags were NOT being checked.

pkg/handlers/adminapi/webhook_subscriptions.go
shimonacarvalho: Yeah I don't know either. I didn't want to create a whole new definition (that has a risk of diverging) so we just handled it in the code. We only use the query id, and return an error if they happen to provide a non-matching body id.

NamibiaLT:
Thanks for the feedback and reviews everyone!

PR #6031, opened 2021-03-03
MB-6935 Set access code flag in auth context

monfresh:
## Description

When the service member is first created upon their very first sign in,
we set their `requires_access_code` field to match the value of the
`FEATURE_FLAG_ACCESS_CODE` env var. This env var was not being properly
read in `auth.go`, and so it was always false, which meant customers
in prod could bypass the access code requirement.

The fix was to set the env var in the `context` that `auth.go` uses,
which is different from the `HandlerContext` that handlers use.

I looked into writing a Cypress test for this, but if I'm reading the [docs](https://docs.cypress.io/guides/guides/environment-variables.html#Option-3-CYPRESS)
correctly, this would have to be a separate standalone test that gets run with the env var.
For example:
```
FEATURE_FLAG_ACCESS_CODE=true yarn cypress run --spec cypress/integration/path/to/file.jsx
```
I don't think there's a way to set the env var within a test while running the entire test suite.

In Rails, you can easily set env vars on a per-test basis with Capybara and RSpec. For example:
```ruby
allow(ENV).to receive(:[]).with("FEATURE_FLAG_ACCESS_CODE").and_return(true)
```

## Reviewer Notes

## Setup

```sh
echo "Code goes here"
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6935) for this change

## Screenshots

Before, a new user would go straight to the first onboarding page where they choose CONUS/OCONUS. Now, they see this if the feature flag is turned on:

Screen Shot 2021-03-02 at 4 12 46 PM

Note that the field to enter the access code is missing. That's probably because the screenshot was taken before an access code was created for that service member. However, after I created a new access code, and made sure not to set the `claimed_at` field, the next time the user signed in, they weren't prompted for the code and went straight to the CONUS/OCONUS page.

### DIFF
--- a/cmd/milmove/serve.go
+++ b/cmd/milmove/serve.go
@@ -943,6 +943,12 @@ func serveFunction(cmd *cobra.Command, args []string) error {
 	}
 
 	authContext := authentication.NewAuthContext(logger, loginGovProvider, loginGovCallbackProtocol, loginGovCallbackPort, sessionManagers)
+	authContext.SetFeatureFlag(
+		authentication.FeatureFlag{
+			Name:   cli.FeatureFlagAccessCode,
+			Active: v.GetBool(cli.FeatureFlagAccessCode),
+		},
+	)
 	authMux := goji.SubMux()
 	root.Handle(pat.New("/auth/*"), authMux)
 	authMux.Handle(pat.Get("/login-gov"), authentication.RedirectHandler{Context: authContext})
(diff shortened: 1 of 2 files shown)


suzubara (approved):
I tested & confirmed I was able to create a new service member user that required an access code using the ENV flag. The access code UX is indeed broken but that's another issue, which I've flagged to the product team 🙌

PR #6032, opened 2021-03-03
Use inmemory filesystem for postgres test db

gidjin:
## Description

Changes the `server_test` setup in the Makefile to use an in memory file system for the postgres db. This is fine since our tests do not preserve any data anyway, and in memory will gain us a speed improvement. I ran this locally and in CircleCI a couple times and looks like at least a 1-2 minute savings on each run of the server tests. 

## Reviewer Notes

What speed difference is this for you vs master?

## Setup

```sh
make server_tests
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -636,9 +636,8 @@ ifndef CIRCLECI
 			POSTGRES_PASSWORD=$(PGPASSWORD) \
 			-d \
 			-p $(DB_PORT_TEST):$(DB_PORT_DOCKER)\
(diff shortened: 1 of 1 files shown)


reggieriser (approved):
LGTM!  I ran `time make clean server_test` a couple of times on `master` and this branch and saw a speed up of around 15-20% on average on the branch.  I also tried `make schemaspy` since that also depends on the test DB and it seemed to work fine as well.

PR #6034, opened 2021-03-03
[MB-6909] Create AK1 EDI Segment for the 997 EDI Response

pearl-truss:
## Description

This PR creates the AK1 segment and its validations as defined in this [mapping doc](https://docs.google.com/spreadsheets/d/1OYYgAp6GD0wG6SKx61De2VRZFkAPHBHZV2BpyvD0nMc/edit#gid=0)

## Reviewer Notes

 GroupControlNumber has a min set to 1. When I [tested the invalid min case](https://github.com/transcom/mymove/pull/6034/files#diff-6c8ef3d6f7e71f9d6ec5a720eb3ca3f81e6b45ff779cae5118e10001fd359aa1R21), I initially set the value to 0. However, this resulted in a "required" error instead of a min error. I'm not sure why that is

## Setup

## References

### DIFF
--- a/pkg/edi/segment/ak1.go
+++ b/pkg/edi/segment/ak1.go
@@ -7,8 +7,8 @@ import (
 
 // AK1 represents the AK1 EDI segment
 type AK1 struct {
-	FunctionalIdentifierCode string `validate:"required,eq=SI"`
-	GroupControlNumber       int64  `validate:"required,min=1,max=999999999"`
+	FunctionalIdentifierCode string `validate:"eq=SI"`
+	GroupControlNumber       int64  `validate:"min=1,max=999999999"`
 }
 
 // StringArray converts AK1 to an array of strings
@@ -30,5 +30,8 @@ func (s *AK1) Parse(elements []string) error {
 	var err error
 	s.FunctionalIdentifierCode = elements[0]
 	s.GroupControlNumber, err = strconv.ParseInt(elements[1], 10, 64)
-	return err
+	if err != nil {
+		return err
+	}
+	return nil
 }
--- a/pkg/edi/segment/ak1.go
+++ b/pkg/edi/segment/ak1.go
@@ -1,0 +1,34 @@
+package edisegment
+
+import (
+	"fmt"
+	"strconv"
+)
+
+// AK1 represents the AK1 EDI segment
+type AK1 struct {
+	FunctionalIdentifierCode string `validate:"required,eq=SI"`
+	GroupControlNumber       int64  `validate:"required,min=1,max=999999999"`
+}
+
+// StringArray converts AK1 to an array of strings
+func (s *AK1) StringArray() []string {
+	return []string{
+		"AK1",
+		s.FunctionalIdentifierCode,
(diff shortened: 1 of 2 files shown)


pkg/edi/segment/ak1.go
```diff
@@ -0,0 +1,34 @@
...
+func (s *AK1) Parse(elements []string) error {
+	expectedNumElements := 2
+	if len(elements) != expectedNumElements {
+		return fmt.Errorf("AK1: Wrong number of elements, expected %d, got %d", expectedNumElements, len(elements))
+	}
+
+	var err error
+	s.FunctionalIdentifierCode = elements[0]
+	s.GroupControlNumber, err = strconv.ParseInt(elements[1], 10, 64)
+	return err
```
reggieriser, line 33: I think you can just return `nil` here and not even bother declaring `var err error`.

pkg/edi/segment/ak1.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"fmt"
+	"strconv"
+)
+
+// AK1 represents the AK1 EDI segment
+type AK1 struct {
+	FunctionalIdentifierCode string `validate:"required,eq=SI"`
+	GroupControlNumber       int64  `validate:"required,min=1,max=999999999"`
```
reggieriser, line 11: Looking at how we set up some of the other validations, you may not have to do the `required` validation since you're already requiring these two values to meet some other criteria.  I don't know that it necessarily hurts, although I think the issue you mentioned in the reviewer notes about getting a required error when passing in a zero may be related.  The `required` validator for an `int64` is just checking that it's not set to the zero value, which is zero for a non-pointer `int64`.  If you take the required validator out, it'll probably tell you it's not in the range instead as you were probably expecting.

reggieriser (commented):
Looking good!  Just a few small things to tweak.

pkg/edi/segment/ak1.go
pearl-truss: I think I have to declare var because line 32 throws an `err undefined` without it and if I switch it to := there's an error for `err error
non-name s.GroupControlNumber on left side of :=` which I found is because go needs err to have an explicit type

reggieriser: You're right.  I totally missed the `err` being returned on line 32.  Sorry for the false alarm.

pkg/edi/segment/ak1.go
```diff
@@ -0,0 +1,37 @@
...
+		return fmt.Errorf("AK1: Wrong number of elements, expected %d, got %d", expectedNumElements, len(elements))
+	}
+
+	var err error
+	s.FunctionalIdentifierCode = elements[0]
+	s.GroupControlNumber, err = strconv.ParseInt(elements[1], 10, 64)
+	if err != nil {
+		return err
+	}
+	return nil
```
reggieriser, line 36: Your previous code was fine here (my mistake), so just doing a `return err` is OK and should achieve the same results if you wanted to go back to it (but it's fine like this too).

reggieriser (approved):
LGTM!  Thanks for making the adjustments. 👍

PR #6048, opened 2021-03-04
[MB-5366] [CAT II] errcheck: Error return value of `function` is not checked.

Ronolibert:
## Description

Error return values from certain functions were not being captured. Add requisite logging in these different scenarios. Instances in test files are annotated and have been approved.

## Reviewer Notes

- A couple tests started to fail after I checked for the return value of a transaction. May be an issue with the data setup. I commented these tests out for now and they will be fixed in a [this story](https://dp3.atlassian.net/browse/MB-6988)

## Setup
`make server_test`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5366) for this change

### DIFF
--- a/cmd/big-cat/main.go
+++ b/cmd/big-cat/main.go
@@ -65,10 +65,11 @@ func main() {
 			panic(err)
 		}
 
-		err = f.Close()
-		if err != nil {
-			logger.Debug("Failed to close filepath", zap.Error(err))
-		}
+		defer func() {
+			if closeErr := f.Close(); closeErr != nil {
+				logger.Debug("Failed to close filepath", zap.Error(closeErr))
+			}
+		}()
 
 		count++
 		if limit >= 0 && count == limit {
--- a/cmd/big-cat/main.go
+++ b/cmd/big-cat/main.go
@@ -18,14 +18,6 @@ import (
 	"github.com/transcom/mymove/pkg/logging"
 )
 
-type logger interface {
-	Debug(msg string, fields ...zap.Field)
-	Info(msg string, fields ...zap.Field)
-	Error(msg string, fields ...zap.Field)
-	Warn(msg string, fields ...zap.Field)
-	Fatal(msg string, fields ...zap.Field)
-}
-
 func main() {
 	flag := pflag.CommandLine
 	cli.InitLoggingFlags(flag)
@@ -69,7 +61,7 @@ func main() {
 		if err != nil {
 			panic(err)
 		}
-		if _, err := io.Copy(os.Stdout, bufio.NewReader(f)); err != nil {
+		if _, err = io.Copy(os.Stdout, bufio.NewReader(f)); err != nil {
 			panic(err)
 		}
 
--- a/cmd/big-cat/main.go
+++ b/cmd/big-cat/main.go
@@ -4,12 +4,49 @@ import (
 	"bufio"
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
+
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+	"github.com/transcom/mymove/pkg/cli"
+	"github.com/transcom/mymove/pkg/logging"
+	"go.uber.org/zap"
 )
 
+type logger interface {
+	Debug(msg string, fields ...zap.Field)
+	Info(msg string, fields ...zap.Field)
+	Error(msg string, fields ...zap.Field)
+	Warn(msg string, fields ...zap.Field)
+	Fatal(msg string, fields ...zap.Field)
+}
+
 func main() {
+	flag := pflag.CommandLine
+	cli.InitLoggingFlags(flag)
+
+	v := viper.New()
+	bindErr := v.BindPFlags(flag)
+	if bindErr != nil {
+		log.Fatal("failed to bind flags", zap.Error(bindErr))
+	}
+
+	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
+	v.AutomaticEnv()
+
+	dbEnv := v.GetString(cli.DbEnvFlag)
+	dbEnv := v.GetString(cli.DbEnvFlag)
+
+	logger, err := logging.Config(logging.WithEnvironment(dbEnv), logging.WithLoggingLevel(v.GetString(cli.LoggingLevelFlag)))
+
+	if err != nil {
+		log.Fatalf("Failed to initialize Zap logging due to %v", err)
+	}
+
 	if len(os.Args) != 2 && len(os.Args) != 3 {
 		fmt.Println("Usage: big-cat <path> [limit]")
 		os.Exit(1)
@@ -35,7 +72,12 @@ func main() {
 		if _, err := io.Copy(os.Stdout, bufio.NewReader(f)); err != nil {
 			panic(err)
 		}
-		f.Close()
+
+		err = f.Close()
+		if err != nil {
+			logger.Debug("Failed to close filepath", zap.Error(err))
+		}
+
 		count++
 		if limit >= 0 && count == limit {
 			break
--- a/cmd/big-cat/main.go
+++ b/cmd/big-cat/main.go
@@ -12,9 +12,10 @@ import (
 
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
+	"go.uber.org/zap"
+
 	"github.com/transcom/mymove/pkg/cli"
 	"github.com/transcom/mymove/pkg/logging"
-	"go.uber.org/zap"
 )
 
 type logger interface {
@@ -38,7 +39,6 @@ func main() {
 	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 	v.AutomaticEnv()
 
-	dbEnv := v.GetString(cli.DbEnvFlag)
 	dbEnv := v.GetString(cli.DbEnvFlag)
 
 	logger, err := logging.Config(logging.WithEnvironment(dbEnv), logging.WithLoggingLevel(v.GetString(cli.LoggingLevelFlag)))
--- a/cmd/big-cat/main.go
+++ b/cmd/big-cat/main.go
@@ -61,15 +61,15 @@ func main() {
 		if err != nil {
 			panic(err)
 		}
-		if _, err = io.Copy(os.Stdout, bufio.NewReader(f)); err != nil {
-			panic(err)
-		}
-
 		defer func() {
 			if closeErr := f.Close(); closeErr != nil {
 				logger.Debug("Failed to close filepath", zap.Error(closeErr))
 			}
 		}()
+
+		if _, err = io.Copy(os.Stdout, bufio.NewReader(f)); err != nil {
+			panic(err)
+		}
 
 		count++
 		if limit >= 0 && count == limit {
--- a/cmd/milmove/gen_certs_migration.go
+++ b/cmd/milmove/gen_certs_migration.go
@@ -202,7 +202,12 @@ func genCertsMigration(cmd *cobra.Command, args []string) error {
 		if errStore != nil {
 			return fmt.Errorf("Ensure CAC reader and card inserted: %w", errStore)
 		}
-		defer store.Close()
+
+		defer func() {
+			if closeErr := store.Close(); closeErr != nil {
+				fmt.Println(fmt.Errorf("Failed to close CAC store").Error())
+			}
+		}()
 
 		cert, errTLSCert := store.TLSCertificate()
 		if errTLSCert != nil {
--- a/cmd/milmove/gen_certs_migration.go
+++ b/cmd/milmove/gen_certs_migration.go
@@ -205,7 +205,7 @@ func genCertsMigration(cmd *cobra.Command, args []string) error {
 
 		defer func() {
 			if closeErr := store.Close(); closeErr != nil {
-				fmt.Println(fmt.Errorf("Failed to close CAC store").Error())
+				fmt.Println("Failed to close CAC store")
 			}
 		}()
 
--- a/cmd/milmove/main_test.go
+++ b/cmd/milmove/main_test.go
@@ -31,7 +31,10 @@ func TestWebServerSuite(t *testing.T) {
 	flag.Parse([]string{})
 
 	v := viper.New()
-	v.BindPFlags(flag)
+	bindErr := v.BindPFlags(flag)
+	if bindErr != nil {
+		log.Fatal("failed to bind flags", zap.Error(bindErr))
+	}
 	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 	v.AutomaticEnv()
 
--- a/cmd/milmove/main_test.go
+++ b/cmd/milmove/main_test.go
@@ -1,3 +1,11 @@
+//RA Summary: gosec - errcheck - Unchecked return value
+//RA: Linter flags errcheck error: Ignoring a method's return value can cause the program to overlook unexpected states and conditions.
+//RA: Functions with unchecked return values in the file are used set up environment variables
+//RA: Given the functions causing the lint errors are used to set environment variables for testing purposes, it does not present a risk
+//RA Developer Status: Mitigated
+//RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+//RA Modified Severity: N/A
+// nolint:errcheck
 package main
 
 import (
--- a/cmd/milmove/main_test.go
+++ b/cmd/milmove/main_test.go
@@ -1,11 +1,3 @@
-//RA Summary: gosec - errcheck - Unchecked return value
-//RA: Linter flags errcheck error: Ignoring a method's return value can cause the program to overlook unexpected states and conditions.
-//RA: Functions with unchecked return values in the file are used set up environment variables
-//RA: Given the functions causing the lint errors are used to set environment variables for testing purposes, it does not present a risk
-//RA Developer Status: Mitigated
-//RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
-//RA Modified Severity: N/A
-// nolint:errcheck
 package main
 
 import (
@@ -36,7 +28,14 @@ func TestWebServerSuite(t *testing.T) {
 
 	flag := pflag.CommandLine
 	initServeFlags(flag)
-	flag.Parse([]string{})
+	//RA Summary: gosec - errcheck - Unchecked return value
+	//RA: Linter flags errcheck error: Ignoring a method's return value can cause the program to overlook unexpected states and conditions.
+	//RA: Functions with unchecked return values in the file are used set up environment variables
+	//RA: Given the functions causing the lint errors are used to set environment variables for testing purposes, it does not present a risk
+	//RA Developer Status: Mitigated
+	//RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+	//RA Modified Severity: N/A
+	flag.Parse([]string{}) // nolint:errcheck
 
 	v := viper.New()
 	bindErr := v.BindPFlags(flag)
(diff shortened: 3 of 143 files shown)


cmd/webhook-client/webhook/webhook_test.go
```diff
@@ -1,3 +1,11 @@
+//RA Summary: gosec - errcheck - Unchecked return value
```
shimonacarvalho, line 1: I'm not sure I understand what is happening with this file. Can you explain?

cmd/webhook-client/webhook_notify.go
```diff
@@ -56,7 +56,11 @@ func webhookNotify(cmd *cobra.Command, args []string) error {
 
 	// Defer closing the store until after the api call has completed
 	if cacStore != nil {
-		defer cacStore.Close()
+		defer func() {
```
shimonacarvalho, line 59: Did you run the webhook-client tests for these changes? They don't run automatically on server_test since this is a different application. Might be a good idea to check. 

I think it's just
```
go test ./cmd/webhook-client/webhook
```

cmd/webhook-client/webhook/webhook_test.go
Ronolibert: Yeah, this is just justifying that it is fine to ignore the lint error in this instance because it is a test file. Jarod, our information system security officer, goes through each of these annotations and either approves it or pushes back saying it shouldn't actually be disabled because of this specific case

cmd/webhook-client/webhook_notify.go
Ronolibert: Not yet! Will re-request review once I finish fixing all the tests that have broke. Adding some of these log errors unearthed some tests that need fixing

Ronolibert: I ran them and everything passed except for `Test_SendStgNotification` which I assume is fine? Just because its also failing on master

pkg/auth/authentication/devlocal.go
```diff
@@ -412,7 +415,10 @@ func createUser(h devlocalAuthHandler, w http.ResponseWriter, r *http.Request) (
 		}
 
 		role := roles.Role{}
-		h.db.Where("role_type = $1", "ppm_office_users").First(&role)
+		err = h.db.Where("role_type = $1", "ppm_office_users").First(&role)
+		if err != nil {
+			h.logger.Error("could not fetch role ppm_office_users", zap.Error(err))
```
Ronolibert, line 420: short circuit

Ronolibert: @reggieriser For some reason, nowhere in the file does it short circuit if it runs into an error. This is something only local but not sure what I should do in this situation

cmd/webhook-client/webhook_notify.go
shimonacarvalho: Hmm just passed for me... 
```
    --- PASS: TestWebhookClientTestingSuite/Test_SendStgNotification (0.89s)
        --- PASS: TestWebhookClientTestingSuite/Test_SendStgNotification/Successful_post_to_staging (0.77s)
PASS
ok  	github.com/transcom/mymove/cmd/webhook-client/webhook	5.790s
```

cmd/milmove/gen_certs_migration.go
```diff
@@ -202,7 +202,12 @@ func genCertsMigration(cmd *cobra.Command, args []string) error {
 		if errStore != nil {
 			return fmt.Errorf("Ensure CAC reader and card inserted: %w", errStore)
 		}
-		defer store.Close()
+
+		defer func() {
+			if closeErr := store.Close(); closeErr != nil {
+				fmt.Println(fmt.Errorf("Failed to close CAC store").Error())
```
reggieriser, line 208: I'm confused about why we need to create an error here just to print out the message via `Error()`.  Is there any reason it couldn't just be `fmt.Println("Failed to close CAC store")`?  Also, is it OK here to just print to stdout as opposed to logging it like we do for others?

cmd/milmove/main_test.go
```diff
@@ -1,3 +1,11 @@
+//RA Summary: gosec - errcheck - Unchecked return value
+//RA: Linter flags errcheck error: Ignoring a method's return value can cause the program to overlook unexpected states and conditions.
+//RA: Functions with unchecked return values in the file are used set up environment variables
+//RA: Given the functions causing the lint errors are used to set environment variables for testing purposes, it does not present a risk
+//RA Developer Status: Mitigated
+//RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+//RA Modified Severity: N/A
+// nolint:errcheck
```
reggieriser, line 8: I think I see just one place in this file (line 39) where we're not checking an error.  I'm not sure I follow why that's a particularly hard one to check.  Am I missing something?

Also, I'm wondering why we turn on `nolint` for the entire file (which I think this does) instead of targeting it to a specific line.  If this file is modified in the future, wouldn't we want any new cases where errors aren't checked to be flagged?

pkg/testdatagen/scenario/devseed.go
```diff
@@ -2292,8 +2302,10 @@ func createMoveWithServiceItems(db *pop.Connection, userUploader *uploader.UserU
 
 	move9 := testdatagen.MakeMove(db, testdatagen.Assertions{
 		Move: models.Move{
-			ID:                 uuid.FromStringOrNil("7cbe57ba-fd3a-45a7-aa9a-1970f1908ae7"),
-			OrdersID:           orders9.ID,
+			ID:       uuid.FromStringOrNil("7cbe57ba-fd3a-45a7-aa9a-1970f1908ae7"),
+			OrdersID: orders9.ID,
+			// SelectedMoveType:   &hhgMoveType,
+			Status:             models.MoveStatusSUBMITTED,
```
reggieriser, line 2308: What's this change about?  Is it somehow related to the error checking?

pkg/testdatagen/make_admin_user.go
```diff
@@ -75,7 +76,11 @@ func MakeActiveAdminUser(db *pop.Connection) models.AdminUser {
 
 	adminUser.Active = true
 
-	db.Update(&adminUser)
+	err := db.Update(&adminUser)
```
reggieriser, line 79: We have a `mustSave` that may be helpful here.  But looking at this function, I wonder why we're not just passing this in via an assertion to the `MakeAdminUser` function instead.

pkg/testdatagen/make_office_user.go
```diff
@@ -113,7 +114,11 @@ func MakeActiveOfficeUser(db *pop.Connection) models.OfficeUser {
 
 	officeUser.Active = true
 
-	db.Update(&officeUser)
+	err := db.Update(&officeUser)
```
reggieriser, line 117: Same thing here as I mentioned in `make_admin_user.go`.

pkg/storage/memory.go
```diff
@@ -78,7 +78,12 @@ func (fs *Memory) Store(key string, data io.ReadSeeker, checksum string, tags *s
 	if err != nil {
 		return nil, errors.Wrap(err, "could not open file")
 	}
-	defer file.Close()
+
+	defer func() {
+		if closeErr := file.Close(); closeErr != nil {
+			fmt.Println(fmt.Errorf("Failed to close file").Error())
```
reggieriser, line 84: Another case of creating an error just to turn it into a string for printing (rather than just printing the string directly).

reggieriser (commented):
Things generally look good.  I added some comments for the files assigned to me on the [Slack thread](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1615477922361800).

cmd/big-cat/main.go
```diff
@@ -32,10 +61,15 @@ func main() {
 		if err != nil {
 			panic(err)
 		}
-		if _, err := io.Copy(os.Stdout, bufio.NewReader(f)); err != nil {
+		if _, err = io.Copy(os.Stdout, bufio.NewReader(f)); err != nil {
 			panic(err)
 		}
-		f.Close()
+
+		err = f.Close()
```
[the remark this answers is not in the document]
Ronolibert: That makes sense, it'd also align more with how all the other command line tools are structured

cmd/milmove/gen_certs_migration.go
Ronolibert: I'll change it to a simpler Println. I think it was probably a convention from a different file. I think its ok to print. I don't think any of the other cmd line utilities do any extensive logging besides regular prints

pkg/testdatagen/make_admin_user.go
Ronolibert: I think we should leave this the way it is. I just want to keep the scope of this story to capturing the errors

cmd/milmove/main_test.go
Ronolibert: In all the test files, it's less about the effort and more about us already getting sign off that it's unnecessary to change it for the ATO for this specific error

I moved the annotation lower to the relevant line

pkg/testdatagen/scenario/devseed.go
Ronolibert: That's odd. My git blame doesn't even show me making this change. In any case, I'll uncomment it because it seems like this one was an accident

pkg/middleware/recovery.go
```diff
@@ -44,7 +44,11 @@ func Recovery(logger Logger) func(inner http.Handler) http.Handler {
 					}{handlers.InternalServerErrMessage, traceID, "An unexpected server error has occurred."})
 					w.Header().Set("Content-Type", "application/json")
 					w.WriteHeader(http.StatusInternalServerError)
-					w.Write(jsonBody)
+
+					_, err := w.Write(jsonBody)
+					if err != nil {
+						logger.Error("Failed to write data to the connection", zap.Error(err))
```
duncan-truss, line 50: This is minor but connection seems like a weird phrase here, I might just say failed to write json error to the response body.

pkg/migrate/SplitStatements_test.go
```diff
@@ -61,8 +69,8 @@ func TestSplitStatementsCopyFromStdinMultiple(t *testing.T) {
 	// Load the fixture with the sql example
 	fixture := "./fixtures/copyFromStdinMultiple.sql"
 	f, err := os.Open(fixture)
-	// #nosec G307 TODO needs review
-	defer f.Close()
+
+	defer f.Close() //lint:ignore SA5001
```
duncan-truss, line 73: Why does this need a different ignore rule than the nosec G307 that is used below?

pkg/models/move.go
```diff
@@ -414,9 +421,12 @@ func (m Move) CreateMovingExpenseDocument(
 		}
 
 		return nil
-
 	})
 
+	if transactionErr != nil {
+		return nil, responseVErrors, transactionErr
```
duncan-truss, line 427: I think this should be changed to `return nil, responseVErrors, responseError`

pkg/models/move.go
```diff
@@ -649,6 +659,10 @@ func SaveMoveDependencies(db *pop.Connection, move *Move) (*validate.Errors, err
 		return nil
 	})
 
+	if transactionErr != nil {
+		return responseVErrors, transactionErr
```
duncan-truss, line 663: similar change here `return responseVErrors, responseError`

pkg/models/move_document.go
```diff
@@ -392,5 +392,9 @@ func SaveMoveDocument(db *pop.Connection, moveDocument *MoveDocument, saveExpens
 		return nil
 	})
 
+	if err != nil {
+		return responseVErrors, err
```
duncan-truss, line 396: This should be `return responseVErrors, responseError`

pkg/models/order.go
```diff
@@ -121,6 +121,11 @@ func SaveOrder(db *pop.Connection, order *Order) (*validate.Errors, error) {
 		}
 		return nil
 	})
+
+	if transactionErr != nil {
+		return responseVErrors, transactionErr
```
duncan-truss, line 126: This should be `return responseVErrors, responseError`

pkg/models/order.go
```diff
@@ -99,7 +99,7 @@ func SaveOrder(db *pop.Connection, order *Order) (*validate.Errors, error) {
 	responseVErrors := validate.NewErrors()
 	var responseError error
 
-	db.Transaction(func(dbConnection *pop.Connection) error {
+	transactionErr := db.Transaction(func(dbConnection *pop.Connection) error {
 		transactionError := errors.New("Rollback The transaction")
 
 		ppm, err := FetchPersonallyProcuredMoveByOrderID(db, order.ID)
```
duncan-truss, line 105: In this first error block we should set responseError = err.  We should really just be returning responseError in these blocks directly instead of the transactionError.

pkg/models/personally_procured_move.go
```diff
@@ -237,7 +237,7 @@ func SavePersonallyProcuredMove(db *pop.Connection, ppm *PersonallyProcuredMove)
 	responseVErrors := validate.NewErrors()
 	var responseError error
 
-	db.Transaction(func(db *pop.Connection) error {
+	transactionErr := db.Transaction(func(db *pop.Connection) error {
 		transactionError := errors.New("Rollback The transaction")
 
 		if ppm.HasRequestedAdvance {
```
duncan-truss, line 243: We need to set a value for responseError in the `if ppm.Advance.MethodOfReceipt == MethodOfReceiptGTCC {` block.

pkg/models/personally_procured_move.go
```diff
@@ -291,5 +291,9 @@ func SavePersonallyProcuredMove(db *pop.Connection, ppm *PersonallyProcuredMove)
 
 	})
 
+	if transactionErr != nil {
+		return responseVErrors, transactionErr
```
duncan-truss, line 295: This should become `return responseVErrors, responseError`

pkg/models/service_member.go
```diff
@@ -181,6 +182,10 @@ func SaveServiceMember(dbConnection *pop.Connection, serviceMember *ServiceMembe
 		return nil
 	})
 
+	if transactionErr != nil {
+		return responseVErrors, transactionErr
```
duncan-truss, line 186: Change to `return responseVErrors, responseError`

pkg/models/service_member.go
```diff
@@ -268,6 +273,10 @@ func (s ServiceMember) CreateOrder(db *pop.Connection,
 		return nil
 	})
 
+	if transactionErr != nil {
+		return newOrders, responseVErrors, transactionErr
```
duncan-truss, line 277: Change to `newOrders, responseVErrors, responseError`

pkg/models/weight_ticket_set_document.go
```diff
@@ -145,8 +145,11 @@ func (m Move) CreateWeightTicketSetDocument(
 		}
 
 		return nil
-
 	})
 
+	if transactionErr != nil {
+		return weightTicketSetDocument, responseVErrors, transactionErr
```
duncan-truss, line 151: We need to set a value for response error in the above block on line 132 for `if responseVErrors.HasAny() || responseError != nil {`

Then change this to `weightTicketSetDocument, responseVErrors, responseError`

pkg/paperwork/forms.go
```diff
@@ -152,7 +152,10 @@ func (f *FormFiller) AppendPage(templateImage io.ReadSeeker, fields map[string]F
 	if err != nil {
 		return errors.Wrap(err, "could not decode image config")
 	}
-	templateImage.Seek(0, io.SeekStart)
+	_, err = templateImage.Seek(0, io.SeekStart)
+	if err != nil {
+		return errors.Wrap(err, "could not find image config")
```
duncan-truss, line 157: I don't know what config means here maybe could not read the image data

pkg/migrate/SplitStatements_test.go
Ronolibert: The SA5001 error code is a staticcheck while the G307 is gosec. This actually doesn't even work. There's currently a bug that staticcheck ignore behavior doesn't behave as expected.

pkg/payment_request/service_param_value_lookups/distance_zip3_lookup.go
```diff
@@ -58,7 +58,10 @@ func (r DistanceZip3Lookup) lookup(keyData *ServiceItemParamKeyData) (string, er
 	if distanceMiles >= 50 {
 		miles := unit.Miles(distanceMiles)
 		mtoShipment.Distance = &miles
-		db.Save(&mtoShipment)
+		err := db.Save(&mtoShipment)
```
duncan-truss, line 61: I think err is already defined in the outer scope not sure if you'll get another linter error to that effect.

pkg/services/move_order/move_order_fetcher.go
```diff
@@ -161,7 +161,10 @@ func (f moveOrderFetcher) FetchMoveOrder(moveOrderID uuid.UUID) (*models.Order,
 	// cannot eager load the address as "OriginDutyStation.Address" because
 	// OriginDutyStation is a pointer.
 	if moveOrder.OriginDutyStation != nil {
-		f.db.Load(moveOrder.OriginDutyStation, "Address")
+		err := f.db.Load(moveOrder.OriginDutyStation, "Address")
```
duncan-truss, line 164: err is already defined in the outer scope

pkg/services/payment_request/payment_request_creator.go
```diff
@@ -236,7 +241,11 @@ func (p *paymentRequestCreator) createPaymentRequestSaveToDB(tx *pop.Connection,
 		return nil, services.NewConflictError(moveTaskOrder.OrdersID, fmt.Sprintf("Orders on MoveTaskOrder (ID: %s) missing OriginDutyStation", moveTaskOrder.ID))
 	}
 	// Verify that ServiceMember is Valid
-	tx.Load(&moveTaskOrder.Orders, "ServiceMember")
+	err = tx.Load(&moveTaskOrder.Orders, "ServiceMember")
+	if err != nil {
+		return nil, services.NewNotFoundError(moveTaskOrder.OrdersID, fmt.Sprintf("ServiceMember on MoveTaskOrder (ID: %s) not valid", moveTaskOrder.ID))
```
duncan-truss, line 246: Following the other patterns I think you want moveTaskOrder.ServiceMemberID as the first not found arg.

duncan-truss (approved):
Thanks Ron, I think you addressed all the feedback from the packages I reviewed.

cmd/webhook-client/webhook_notify.go
Ronolibert: That's odd, I must be missing something locally? I still get the same failing on master

abbyoung: Just ran on this branch and it's also passing for me

abbyoung (approved):
LGTM! Thank you!

cmd/big-cat/main.go
```diff
@@ -65,10 +65,11 @@ func main() {
 			panic(err)
 		}
 
-		err = f.Close()
-		if err != nil {
-			logger.Debug("Failed to close filepath", zap.Error(err))
-		}
+		defer func() {
```
reggieriser, line 68: Thanks for making this a defer -- I think I was also suggesting moving it to above the `Copy` statement so that it will close the file if the copy fails.  If you feel that's changing too much for this PR, that's fine.

pkg/services/payment_request/upload_creator_test.go
```diff
@@ -70,8 +78,9 @@ func (suite *PaymentRequestServiceSuite) TestCreateUploadFailure() {
 	suite.T().Run("invalid payment request ID", func(t *testing.T) {
 		testFile, err := os.Open("../../testdatagen/testdata/test.pdf")
 		suite.NoError(err)
-		// #nosec G307 TODO needs review
-		defer testFile.Close()
+
+		// TODO
+		defer testFile.Close() // #nosec G307
```
reggieriser, line 83: Is this supposed to be fixed like the others?  Or is this covered by the nolint directive at the top of the file?  I'm not clear why this one is different or what's left TODO.  There are three cases like this in this file.

reggieriser (approved):
I looked at the changes you made in response to my comments.  Also looked at tests in `pkg/services`.  I feel good enough to approve my portion.  I did leave a few notes about a couple of things.  Also, we seem to often turn off the error check linter globally for a file (at least in test files I looked at).  That may be fine, but just wanted to mention it since it's going to stop any future missed error checks from being flagged in that file.

pkg/services/payment_request/upload_creator_test.go
Ronolibert: I'll remove the TODO, this is covered by the annotation but 2 different linters gave 2 separate error codes so I think thats why there's also a disable inline

cmd/big-cat/main.go
Ronolibert: Ohhh, I understand. I'll move it above the copy statement. Makes total sense you want to initialize the defer before failure can occur

PR #6049, opened 2021-03-04
MB-4738 TOO can see tag showing shipments needing approval on Move Details tab

jenniferlynparsons:
## Description

The tag next to the "Move Details" tab in the navbar now shows the unapproved shipments count.

## Reviewer Notes

Adding the extra query to the `useMovePaymentRequestsQueries` seemed like the most efficient way to bring that value in, but I'd be interested to hear of a better approach.

## Setup

```sh
make server_run
make office_client_run
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4738) for this change
* [Discussion of the approach](https://dp3.atlassian.net/browse/MB-6859)

## Screenshots
Screen Shot 2021-03-04 at 2 00 29 PM


src/hooks/queries.js
```diff
@@ -231,21 +231,37 @@ export const usePaymentRequestQueueQueries = ({ sort, order, filters = [], curre
...
-
   const { data: mtoShipments = [], ...mtoShipmentQuery } = useQuery([MTO_SHIPMENTS, mtoID, false], getMTOShipments, {
     enabled: !!mtoID,
   });
 
-  const { isLoading, isError, isSuccess } = getQueriesStatus([movePaymentRequestsQuery, mtoShipmentQuery]);
+  // This queries for the unapproved shipments count used in the Navbar
+  const { data: move = {} } = useQuery([MOVES, moveCode], getMove);
+  const moveId = move?.id;
+  const { data: unapprovedShipments = [], ...unapprovedShipmentsQuery } = useQuery(
```
duncan-truss, line 247: I think we're already doing this exact query above right? https://github.com/transcom/mymove/pull/6049/files#diff-2ebe9b4126d348862abb61996ac199b2623d79a2606d2c2d3ab4c63400039f89R240

jenniferlynparsons: Nope! I thought so too, but the first query runs `getMovePaymentRequests` which returns an empty object for the `mto_shipments`.

jenniferlynparsons: If there are no payment requests yet, `data` from the first query is empty. So it's possible to have the `unapprovedShipments` be empty as well, even if there are shipments. Meaning the tag will be hidden on Move Details when viewing this tab.

duncan-truss: Ok I think I see what you're getting at.  You're saying we don't run the second query if there are no payment requests yet for the move because the return value is empty.

I'm not sure how likely it is the user has reached the payment requests tab in that case unless they've navigated from the Move Details or MTO tags and the state will already be accurate?

I might suggest you move the getMove query up to run after the first.  Then we could make the getMTOShipments query have an enabled of something like:
```
const { data: mtoShipments = [], ...mtoShipmentQuery } = useQuery([MTO_SHIPMENTS, mtoID, false], getMTOShipments, {
    enabled: !!mtoID || !!moveId,
  });
```

Then we wouldn't have multiple of the same query.

src/pages/Office/MoveDetails/MoveDetails.jsx
```diff
@@ -75,6 +76,11 @@ const MoveDetails = () => {
     },
   });
 
+  useEffect(() => {
+    const shipmentCount = mtoShipments.filter((shipment) => shipment.status === 'SUBMITTED').length;
+    setUnapprovedShipmentCount(shipmentCount);
+  });
```
duncan-truss, line 82: Could we make this only run if mtoShipments has changed with:

```javascript
  useEffect(() => {
    const shipmentCount = mtoShipments.filter((shipment) => shipment.status === 'SUBMITTED').length;
    setUnapprovedShipmentCount(shipmentCount);
  }, [mtoShipments]);
```

jenniferlynparsons: Done! I had to add `setUnapprovedShipmentCount` to the dependency array as well or eslint threw a fit.

src/pages/Office/MoveDetails/MoveDetails.test.jsx
```diff
@@ -233,7 +235,7 @@ describe('MoveDetails page', () => {
 
     const wrapper = mount(
       <MockProviders initialEntries={[`/moves/${mockRequestedMoveCode}/details`]}>
-        <MoveDetails />
+        <MoveDetails setUnapprovedShipmentCount={setUnapprovedShipmentCount} />
       </MockProviders>,
     );
```
duncan-truss, line 241: Did you want a test that asserts that your mocked function is called once with the value of 1?

jenniferlynparsons: I thought about that but didn't because it felt like testing implementation details since this is mostly dependent on `useEffect`, `useState` and the query. Do you feel it's worth it for that?

yarn.lock
```diff
@@ -2580,21 +2580,6 @@
     prop-types "^15.7.2"
     regenerator-runtime "^0.13.7"
 
-"@storybook/addons@6.1.20":
```
Ryan-Koch, line 2583: Do we know what's driving the changes to yarn.lock? Wasn't sure if this is purposeful or a side effect we're seeing here.

Ryan-Koch (approved):
This does the thing! It looks like y'all worked through the query ordering/duplication topic I saw some chatter about. I asked a question, but don't want to block over it.

yarn.lock
jenniferlynparsons: I'm going to ping the engineering channel and get some clarification on this. It looks like we've upgraded to 6.1.21 so I'm not sure what's up there.

Ryan-Koch: Ah I gotcha. I wonder if this just needs a merge from master. I think I recall seeing a bump up for this in one of the PRs from the AM deploy.

jenniferlynparsons: Ah, that makes sense. I'll pull that in and see if it resolves the issue. I rebased this morning but the issue was already there when I did that.

src/pages/Office/MoveDetails/MoveDetails.test.jsx
duncan-truss: I think it's sufficiently limited to testing this component, and we'll know if someone breaks it.

Something simple like this:
```javascript
it('updates the unapproved shipments tag', () => {
      expect(setUnapprovedShipmentCount.mock.calls[0][0]).toBe(1);
    });
```

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -101,6 +103,13 @@ export const MoveTaskOrder = ({ match }) => {
     });
   };
 
+  useEffect(() => {
+    const shipmentCount = mtoShipments
+      ? Object.values(mtoShipments).filter((shipment) => shipment.status === 'SUBMITTED').length
```
duncan-truss, line 108: I think mtoShipments here always has a default value of an empty object so may not need the conditional. We are also calling Object.values later in this component again on mtoShipments if we wanted to consolidate that.

It looks like mtoShipments is not normalized in MoveDetails and PaymentRequest useQuery hooks but we are normalizing it here and getting an object back.  Not a must have here but maybe on next tag pass we can make sure they align.

duncan-truss (approved):
SOLID! 🎉

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
jenniferlynparsons: I cleaned this up a little. Normalizing looked like it added another layer of work (things exploded a bit) so maybe we want to make that a separate chore.

PR #6050, opened 2021-03-04
MB-6620 make iampostpostgres tests less flaky

YanZ777:
## Description

Extracted out the anonymous function into its own function.
Created a channel to pass errors up.
Created a channel to terminate EnableIAM.
Rewrote the tests such that it no longer checks that different passwords are being cycled through, but instead checks that after kicking off EnableIAM, that the current password does change from the initial password to whatever password the EnableIAM is using.

## Reviewer Notes
Are these tests sufficient?
Should I be doing logger.Fatal or logger.Error?
Should I have removed the entropy?
Anything else I should be doing?
Am I doing Go correctly?s
Anything special I need to do to further test?

## Setup

No need to change the timer since the very nature of the tests have changed.
Run the tests and verify that it passes.
 
```sh
go test -count=1 ./pkg/iampostgres/... -run TestEnableIAMNormal
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6620) for this change


lynzt (commented):
LGTM, but let's get a review from Infra to double check...

pkg/iampostgres/iampostgres.go
```diff
@@ -78,40 +76,29 @@ func updateDSN(dsn string) (string, error) {
...
-		maxDur := 5000
-		randInt, err := random.GetRandomIntAddend(minDur, maxDur)
-		if err != nil {
-			logger.Error("Error building auth token", zap.Error(err))
+// Refreshes the RDS IAM on the given interval.
+func refreshRDSIAM(host string, port string, region string, user string, creds *credentials.Credentials, rus RDSUtilService, ticker *time.Ticker, logger Logger, errorMessagesChan chan error, shouldQuitChan chan bool) {
+	// This for loop immediately runs the first tick then on interval
+	// This for loop will run indefinitely until it either errors or true is
+	// passed to the should quit channel.
+	for {
```
mdawn, line 84: this looks much improved to me

pkg/iampostgres/iampostgres_test.go
```diff
@@ -80,68 +87,81 @@ func TestGetCurrentPasswordFail(t *testing.T) {
...
+		shouldQuitChan)
+
+	// The sleep time should be greater than how often the password will cycle
+	// so that the next time the password is fetched, it will have changed.
+	time.Sleep(2 * time.Second)
+
+	// Confirm that the password has changed (it's no longer the initial
+	// password) to the 1 password being cycled through.
+	pass = GetCurrentPass()
+	assert.Equal("abc", pass)
```
mdawn, line 162: Since we're channel-silly 😄, I wonder if you could've used some flavor of `for val := range c` to auto-break the loop and plug in these 3 values. [This](https://medium.com/rungo/anatomy-of-channels-in-go-concurrency-in-go-1ec336086adb) explains it better than I'm doing...

Not sure what the Postgres implications would be, if any.

mdawn (approved):
And so begins Yan's meteoric rise 🧠 

lgtm!

YanZ777:
Force push was for updating the config.yml, I missed a spot to put in my branch name and wanted just 1 commit to revert.

pkg/iampostgres/iampostgres_test.go
YanZ777: I think using `for val := range c` would make it more complex, since I would have to split generation of a new authToken into its own function so that it can write out to a channel, and then have `refreshRDSIAM` read from said channel, but I don't necessarily see the benefit of doing that?

lynzt (approved):
Based on 👍 from infra, 🚢

PR #6057, opened 2021-03-05
[MB-6910] Created AK2 EDI segment 

pearl-truss:
## Description

This PR creates the AK2 segment and its validations as defined in this [mapping doc](https://docs.google.com/spreadsheets/d/1OYYgAp6GD0wG6SKx61De2VRZFkAPHBHZV2BpyvD0nMc/edit#gid=0)

## Reviewer Notes

The code and test for this segment is essentially identical to the ST segment. Because of the way our code is set up I didn't see a low-to-moderate way to make this more DRY. However, if I'm missing something please let me know!

## Setup

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6910) for this change

### DIFF
--- a/pkg/edi/segment/ak2.go
+++ b/pkg/edi/segment/ak2.go
@@ -19,7 +19,7 @@ func (s *AK2) StringArray() []string {
 	}
 }
 
-// Parse parses an AK2 string that's split into an array into the AK2 struct
+// Parse parses an X12 string that's split into an array into the AK2 struct
 func (s *AK2) Parse(elements []string) error {
 	expectedNumElements := 2
 	if len(elements) != expectedNumElements {
--- a/pkg/edi/segment/ak2.go
+++ b/pkg/edi/segment/ak2.go
@@ -1,0 +1,31 @@
+package edisegment
+
+import (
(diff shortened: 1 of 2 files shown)


pkg/edi/segment/ak2.go
```diff
@@ -0,0 +1,31 @@
...
+// StringArray converts AK2 to an array of strings
+func (s *AK2) StringArray() []string {
+	return []string{
+		"AK2",
+		s.TransactionSetIdentifierCode,
+		s.TransactionSetControlNumber,
+	}
+}
+
+// Parse parses an AK2 string that's split into an array into the AK2 struct
```
reggieriser, line 22: Nit: I think this comment needed to stay "parses an X12 string" like the other segments rather than "parses an AK2 string".  I think they all say `X12` since that's the standard.

reggieriser (approved):
Looks good!  One small nit that's definitely not a blocker.

As far as your comment about keeping things DRY ... you could do some struct embedding and perhaps pull something off, but I don't think it's probably worth that given there are just 2 fields here.

PR #6058, opened 2021-03-05
[MB-6926] Update node to 12.21.0 to match version in docker image

reggieriser:
## Description

Our docker container currently installs the [latest version of node 12.x](https://github.com/transcom/circleci-docker/blob/30242a3c2300c60a7d1a61d09769cdb746c0f4b9/milmove-app/Dockerfile#L58-L60) as part of its setup.  That's version `12.21.0` at the time of this writing.  However, the local version we have in `.node-version` is `12.16.3` and appears to have been that way since we introduced `nodenv` to manage our node versions locally.  In addition, we appear to use `12.16.3` in some GitHub actions and in pre-commit too.  This PR attempts to sync all the versions up (at least for now).

For a future PR, we may want to consider pinning our image to a specific node version similar to what we do for golang versions.  That way, we always can be in sync (in theory), but that does mean we have to manually ensure our containers are updated with new versions (and new node releases happen quite often).  I don't think dependabot can help us with this as it currently stands because node is brought into the image via `curl`.  But this PR at least fixes the sync issue for now.

## Reviewer Notes

Just try some node-related things out (building, running the client, running pre-commit checks, etc.) and make sure you don't see any obvious problems.  The containers have been running with `12.21.0`, so hopefully it's not a big leap to make the change in the other places.

## Setup

You'll need to install node 12.21.0:
`nodenv install 12.21.0`

Once you've done that, check that the node version used when in your repo's directory is `12.21.0`:
`nodenv version` (to show current version)
`nodenv versions` (to show all versions with current version noted)

If you run into problems doing something with pre-commit (like a `pre-commit run -a`), then you may want to try clearing out your pre-commit cache (should be in `~/.cache/pre-commit`) and then installing hooks again (`pre-commit install-hooks`).

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6926) for this change


pearl-truss (approved):
Looks good! I tested building and running the client and storybook

kimallen (approved):
changes to config look fine! (Pamplemoose is triggered to make sure we aren't making changes that impact static analysis)

suzubara (approved):

PR #6067, opened 2021-03-08
[MB-6911] create AK3 EDI segment

pearl-truss:
## Description

This PR creates the AK3 segment and its validations as defined in this [mapping doc](https://docs.google.com/spreadsheets/d/1OYYgAp6GD0wG6SKx61De2VRZFkAPHBHZV2BpyvD0nMc/edit#gid=0)

## Setup

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6911) for this change

### DIFF
--- a/pkg/edi/segment/ak3.go
+++ b/pkg/edi/segment/ak3.go
@@ -1,0 +1,44 @@
+package edisegment
+
+import (
+	"fmt"
+	"strconv"
+)
+
+// AK3 represents the AK3 EDI segment
+type AK3 struct {
+	SegmentIDCode                   string `validate:"min=2,max=3"`
+	SegmentPositionInTransactionSet int    `validate:"min=1,max=999999"`
+	LoopIdentifierCode              string `validate:"min=1,max=6"`
+	SegmentSyntaxErrorCode          string `validate:"min=1,max=3"`
+}
+
+// StringArray converts AK3 to an array of strings
+func (s *AK3) StringArray() []string {
+	return []string{
+		"AK3",
+		s.SegmentIDCode,
+		strconv.Itoa(s.SegmentPositionInTransactionSet),
+		s.LoopIdentifierCode,
+		s.SegmentSyntaxErrorCode,
+	}
+}
+
+// Parse parses an X12 string that's split into an array into the AK3 struct
+func (s *AK3) Parse(elements []string) error {
+	expectedNumElements := 4
+	if len(elements) != expectedNumElements {
+		return fmt.Errorf("AK3: Wrong number of elements, expected %d, got %d", expectedNumElements, len(elements))
+	}
+
+	var err error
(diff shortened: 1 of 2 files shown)


pkg/edi/segment/ak3.go
```diff
@@ -0,0 +1,44 @@
...
+	"fmt"
+	"strconv"
+)
+
+// AK3 represents the AK3 EDI segment
+type AK3 struct {
+	SegmentIDCode                   string `validate:"min=2,max=3"`
+	SegmentPositionInTransactionSet int    `validate:"min=1,max=999999"`
+	LoopIdentifierCode              string `validate:"min=1,max=6"`
+	SegmentSyntaxErrorCode          string `validate:"min=1,max=3"`
```
reggieriser, line 13: Actually, I just noticed something as I was looking at the `AK4` segment.  These are optional fields per the spec.  So I don't think an empty string should trigger a validation failure?

pearl-truss: @reggieriser I made these updates. I'll hold off on merging in case you want to re-review

pkg/edi/segment/ak3_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	altValidAK3 := AK3{
+		SegmentIDCode:                   "ID",
+		SegmentPositionInTransactionSet: 12345,
+	}
+
+	suite.T().Run("validate success with only required fields", func(t *testing.T) {
+		err := suite.validator.Struct(altValidAK3)
+		suite.NoError(err)
+		suite.Equal([]string{"AK3", "ID", "12345", "", ""}, altValidAK3.StringArray())
```
reggieriser, line 28: Nit: I think you can take this out from here since you test that in the new `TestStringArrayAK3` test method.

pkg/edi/segment/ak3_test.go
```diff
@@ -0,0 +1,123 @@
...
+	})
+
+	suite.T().Run("wrong number of elements", func(t *testing.T) {
+		badArrayAK3 := []string{"11", "hello"}
+		var badAK3 AK3
+		err := badAK3.Parse(badArrayAK3)
+		if suite.Error(err) {
+			suite.Contains(err.Error(), "Wrong number of elements")
+		}
+	})
```
reggieriser, line 122: I ran test coverage and we're really close at 91.7% -- there's one more path we're not checking that should be pretty easy to test if we want to get to 100%: when a conversion of a `string` to an `int` (`s.SegmentPositionInTransactionSet`) fails because the string isn't a valid int.

reggieriser (approved):
Thanks for the changes!  LGTM, although I did note a couple of nits in the tests.  They are not blockers so it's up to you if you want to make further changes.  No need for me to re-review if you do address.

PR #6080, opened 2021-03-09
MB-6838 disable fields on edit profile after move submitted

YanZ777:
## Description

After the customer has submitted their move, the branch, rank, and duty station become disabled. 

## Reviewer Notes

I completely forgot that we said we were just going to disable the fields.
So I added a new ServiceInfoTable component so that the disabled fields show up in a table format.
The storybook component is under Customer Components / ServiceInfoTable.

Did I selector correctly?
Is there a better way for me to CSS the things?

## Setup

```sh
make storybook
make client_run
make server_run
```

Good cases to test:
Creating a new user
Editing and saving the profile information before uploading orders
Editing and saving the profile information before submitting the move
Editing the profile information after the move has been submitted (the profile information should be read only and displayed in a table format)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6838) for this change

## Screenshots

Before submitting: 
image

After submitting: 
image


Shkeating (approved):
looks good to me ~

pkg/handlers/internalapi/service_members.go
```diff
@@ -185,8 +185,17 @@ func (h PatchServiceMemberHandler) Handle(params servicememberop.PatchServiceMem
 		return handlers.ResponseForError(logger, err)
 	}
 
+	var move *models.Move
+	if len(serviceMember.Orders) != 0 {
+		moveID, _ := uuid.FromString(params.PatchServiceMemberPayload.MoveID.String())
```
suzubara, line 190: Is there a reason to not load the service member's move on the B/E instead of relying on the move ID parameter?

suzubara (commented):
Most of the changes look good, I left one question about the API handler and also we should remove the eslint ignore line.

pkg/handlers/internalapi/service_members.go
YanZ777: I didn't realize that was an option. Is there an example that I could take a look at somewhere?

pkg/handlers/internalapi/service_members.go
```diff
@@ -197,15 +206,15 @@ func (h PatchServiceMemberHandler) Handle(params servicememberop.PatchServiceMem
...
 }
 
-func (h PatchServiceMemberHandler) patchServiceMemberWithPayload(serviceMember *models.ServiceMember, payload *internalmessages.PatchServiceMemberPayload) (*validate.Errors, error) {
+func (h PatchServiceMemberHandler) patchServiceMemberWithPayload(serviceMember *models.ServiceMember, payload *internalmessages.PatchServiceMemberPayload, move *models.Move) (*validate.Errors, error) {
 
 	if payload.Edipi != nil {
 		serviceMember.Edipi = payload.Edipi
 	}
-	if payload.Affiliation != nil {
+	if payload.Affiliation != nil && (move == nil || move.Status != models.MoveStatusSUBMITTED) {
```
kimallen, line 214: With this conditional, if the move status was `APPROVED`, `CANCELED`, or `APPROVALS_REQUESTED`, they would be able to edit the field.  I think you just want to check that `move.Status == models.MoveStatusDRAFT`

monfresh (changes requested):
Can we add some handler tests please?

YanZ777:
> Can we add some handler tests please?

Certainly. I am working on those as we speak (and seeing what's wrong with the current server tests).

pkg/handlers/internalapi/service_members_test.go
```diff
@@ -223,6 +224,152 @@ func (suite *HandlerSuite) TestPatchServiceMemberHandler() {
...
+	origDutyStation := testdatagen.FetchOrMakeDefaultCurrentDutyStation(suite.DB())
+	newDutyStation := testdatagen.FetchOrMakeDefaultNewOrdersDutyStation(suite.DB())
+	newDutyStationID := strfmt.UUID(newDutyStation.ID.String())
+
+	newServiceMember := models.ServiceMember{
+		UserID:      user.ID,
+		Edipi:       &edipi,
+		Rank:        &origRank,
+		Affiliation: &origAffiliation,
+		DutyStation: origDutyStation,
```
monfresh, line 250: For this to save properly, you also need the Duty Station ID:
```golang
DutyStationID: &origDutyStation.ID,
```

pkg/handlers/internalapi/service_members_test.go
```diff
@@ -223,6 +224,152 @@ func (suite *HandlerSuite) TestPatchServiceMemberHandler() {
...
+	response := handler.Handle(params)
+
+	suite.Assertions.IsType(&servicememberop.PatchServiceMemberOK{}, response)
+	okResponse := response.(*servicememberop.PatchServiceMemberOK)
+
+	serviceMemberPayload := okResponse.Payload
+
+	suite.Assertions.Equal(*serviceMemberPayload.Affiliation, newAffiliation)
+	suite.Assertions.Equal(*serviceMemberPayload.Rank, newRank)
+	suite.Assertions.Equal(*serviceMemberPayload.CurrentStation, newDutyStation)
```
monfresh, line 293: Try comparing the values of the ID strings:
```golang
suite.Assertions.EqualValues(newDutyStation.ID.String(), *serviceMemberPayload.CurrentStation.ID)
```
Also, I know this is confusing, but all of these assertions need to be reversed. The first item in the assertion is the expected value, and the second item is the actual value, so all the payload attributes should be on the right.

pkg/handlers/internalapi/service_members_test.go
```diff
@@ -223,6 +224,152 @@ func (suite *HandlerSuite) TestPatchServiceMemberHandler() {
...
+	suite.Assertions.Equal(*serviceMemberPayload.BackupMailingAddress.StreetAddress1, *backupAddress.StreetAddress1)
+
+	// Then: we expect addresses to have been created
+	addresses := []models.Address{}
+	suite.DB().All(&addresses)
+	suite.Assertions.Len(addresses, 2)
+
+	// If there are orders and the move has been submitted, then the
+	// affiliation rank, and duty station should not be editable.
+
```
monfresh, line 304: Can we add a new separate test for this since it's two different scenarios we're testing?

pkg/handlers/internalapi/service_members_test.go
```diff
@@ -223,6 +224,152 @@ func (suite *HandlerSuite) TestPatchServiceMemberHandler() {
...
+	suite.Assertions.Equal(*serviceMemberPayload.Affiliation, newAffiliation)
+	suite.Assertions.Equal(*serviceMemberPayload.Rank, newRank)
+	suite.Assertions.Equal(*serviceMemberPayload.CurrentStation, newDutyStation)
+	suite.Assertions.Equal(*serviceMemberPayload.ResidentialAddress.StreetAddress1, *resAddress.StreetAddress1)
+	suite.Assertions.Equal(*serviceMemberPayload.BackupMailingAddress.StreetAddress1, *backupAddress.StreetAddress1)
+
+	// Then: we expect addresses to have been created
+	addresses := []models.Address{}
+	suite.DB().All(&addresses)
+	suite.Assertions.Len(addresses, 2)
```
monfresh, line 300: Can we use this syntax instead? 
```golang
suite.Equal(2, len(addresses))
```
It makes it easier to debug. When the actual and expected values don't match, using `suite.Len` means you have to scroll a long time to the right to see the error (at least that's how it is in GoLand). With the `Equal` syntax, the error is immediately obvious:
```
Error:    Not equal: 
          expected: 2
          actual  : 6
```

pkg/handlers/internalapi/service_members_test.go
```diff
@@ -223,6 +224,152 @@ func (suite *HandlerSuite) TestPatchServiceMemberHandler() {
...
+	params := servicememberop.PatchServiceMemberParams{
+		HTTPRequest:               req,
+		ServiceMemberID:           strfmt.UUID(newServiceMember.ID.String()),
+		PatchServiceMemberPayload: &patchPayload,
+	}
+
+	handler := PatchServiceMemberHandler{handlers.NewHandlerContext(suite.DB(), suite.TestLogger())}
+	response := handler.Handle(params)
+
+	suite.Assertions.IsType(&servicememberop.PatchServiceMemberOK{}, response)
```
monfresh, line 286: The `.Assertions` is not necessary. You can shorten all of these to `suite.IsType` or `suite.Equal`, etc.

pkg/handlers/internalapi/service_members_test.go
monfresh: I think the error is right in this case. We expect 6 addresses to be created:
1. Origin Duty Station
2. Transportation Office for Origin DS
3. New Duty Station
4. TO for New DS
5. Backup Address
6. Residential Address

pkg/handlers/internalapi/service_members_test.go
```diff
@@ -223,6 +224,152 @@ func (suite *HandlerSuite) TestPatchServiceMemberHandler() {
 	suite.Assertions.Len(addresses, 2)
 }
 
+func (suite *HandlerSuite) TestPatchServiceMemberHandlerSubmittedMove() {
```
monfresh, line 227: I just realized that we already have a test right above this one to test the patch functionality. Instead of repeating that test in this new one, why not update the existing test with all the possible fields that can be updated (which will get rid of the `TODO: add more fields to change`), and then add assertions for all of those fields to make sure they can all be updated.

And then this new test right here would deal with the scenario where the move is already submitted, and it would basically be the same test except we would check that certain fields can't be updated.

pkg/handlers/internalapi/service_members_test.go
monfresh: Scratch that. See my comment above about updating the existing Patch test above this one and making this one only deal with the scenario where the move is submitted.

monfresh:
@YanZ777 In case you're stuck, here is what the test should look like:

```golang
func (suite *HandlerSuite) TestPatchServiceMemberHandlerSubmittedMove() {
	// Given: a logged in user
	user := testdatagen.MakeDefaultUser(suite.DB())

	edipi := "2342342344"

	// If there are no orders, then the rank, duty station, and affiliation
	// are still editable.
	origRank := models.ServiceMemberRankE1
	newRank := internalmessages.ServiceMemberRankE2

	origAffiliation := models.AffiliationAIRFORCE
	newAffiliation := internalmessages.AffiliationARMY

	origDutyStation := testdatagen.FetchOrMakeDefaultCurrentDutyStation(suite.DB())
	newDutyStation := testdatagen.FetchOrMakeDefaultNewOrdersDutyStation(suite.DB())
	newDutyStationID := strfmt.UUID(newDutyStation.ID.String())

	newServiceMember := models.ServiceMember{
		UserID:        user.ID,
		Edipi:         &edipi,
		Rank:          &origRank,
		Affiliation:   &origAffiliation,
		DutyStation:   origDutyStation,
		DutyStationID: &origDutyStation.ID,
	}
	suite.MustSave(&newServiceMember)

	move := testdatagen.MakeMove(suite.DB(), testdatagen.Assertions{
		Order: models.Order{
			ServiceMember:   newServiceMember,
			ServiceMemberID: newServiceMember.ID,
		},
	})
	move.Submit(time.Now())
	suite.MustSave(&move)

	resAddress := fakeAddressPayload()
	backupAddress := fakeAddressPayload()

	// Check that the branch, rank, or duty station have not changed.
	// These fields should not be changed once a move is submitted
	patchPayload := internalmessages.PatchServiceMemberPayload{
		Edipi:                &edipi,
		BackupMailingAddress: backupAddress,
		ResidentialAddress:   resAddress,
		Affiliation:          &newAffiliation,
		EmailIsPreferred:     swag.Bool(true),
		FirstName:            swag.String("Firstname"),
		LastName:             swag.String("Lastname"),
		MiddleName:           swag.String("Middlename"),
		PersonalEmail:        swag.String("name@domain.com"),
		PhoneIsPreferred:     swag.Bool(true),
		Rank:                 &newRank,
		SecondaryTelephone:   swag.String("555555555"),
		Suffix:               swag.String("Sr."),
		Telephone:            swag.String("555555555"),
		CurrentStationID:     &newDutyStationID,
	}

	req := httptest.NewRequest("PATCH", "/service_members/some_id", nil)
	req = suite.AuthenticateRequest(req, newServiceMember)

	params := servicememberop.PatchServiceMemberParams{
		HTTPRequest:               req,
		ServiceMemberID:           strfmt.UUID(newServiceMember.ID.String()),
		PatchServiceMemberPayload: &patchPayload,
	}

	context := handlers.NewHandlerContext(suite.DB(), suite.TestLogger())
	fakeS3 := storageTest.NewFakeS3Storage(true)
	context.SetFileStorer(fakeS3)
	handler := PatchServiceMemberHandler{context}

	suite.Assertions.IsType(&servicememberop.PatchServiceMemberOK{}, response)
	okResponse := response.(*servicememberop.PatchServiceMemberOK)

	serviceMemberPayload := okResponse.Payload

	suite.Assertions.EqualValues(origAffiliation, *serviceMemberPayload.Affiliation)
	suite.Assertions.EqualValues(origRank, *serviceMemberPayload.Rank)
	suite.Assertions.EqualValues(origDutyStation.ID.String(), *serviceMemberPayload.CurrentStation.ID)
}
```
It was missing these important setup steps:

* The file storer needed to be set in the handler context
* The Move had to be saved again in the DB after submitting it (calling `move.Submit` on its own doesn't update the DB)
* DutyStationID was missing from serviceMember model

pkg/handlers/internalapi/service_members_test.go
YanZ777: That makes perfect sense! Thanks for pointing that out.

pkg/handlers/internalapi/service_members_test.go
YanZ777: Do you think the tests that @Ronolibert is adding are sufficient or should I add more? 
https://github.com/transcom/mymove/pull/6103/files#diff-36ec86e09647281f2fd614e111966653133a1fe887279fc485aab510f2e2b8e1

pkg/handlers/internalapi/service_members.go
```diff
@@ -185,8 +185,26 @@ func (h PatchServiceMemberHandler) Handle(params servicememberop.PatchServiceMem
...
+			}
+		}
+
+		// Don't allow the customer to edit certain fields if there are no
+		// moves in draft status.
+		if len(draftMoves) <= 0 {
+			allowEditIfMoveInDraft = false
+		}
+	}
+
```
monfresh, line 205: A general rule of thumb I like to follow is to keep functions as short as possible. Instead of adding more code to an existing function, consider creating a new function for this new piece of code, and call it from the existing function.

Here's an example that wouldn't require modifying the signature of the `patchServiceMemberWithPayload` function:

```golang
func (h PatchServiceMemberHandler) isDraftMove(serviceMember *models.ServiceMember) bool {
	if serviceMember.Orders == nil {
		return true
	}

	order := serviceMember.Orders[0]
	moves := order.Moves
	var draftMoves []*models.Move
	for i := range moves {
		if moves[i].Status == models.MoveStatusDRAFT {
			draftMoves = append(draftMoves, &moves[i])
		}
	}

	// Don't allow the customer to edit certain fields if there are no
	// moves in draft status.
	if len(draftMoves) <= 0 {
		return false
	}

	return true
}
```

And then this can be called from within `patchServiceMemberWithPayload` by grouping together all the fields that can only be edited if the move is in draft status:

```golang
func (h PatchServiceMemberHandler) patchServiceMemberWithPayload(serviceMember *models.ServiceMember, payload *internalmessages.PatchServiceMemberPayload) (*validate.Errors, error) {
	if h.isDraftMove(serviceMember) {
		if payload.CurrentStationID != nil {
			stationID, err := uuid.FromString(payload.CurrentStationID.String())
			if err != nil {
				return validate.NewErrors(), err
			}
			// Fetch the model partially as a validation on the ID
			station, err := models.FetchDutyStation(h.DB(), stationID)
			if err != nil {
				return validate.NewErrors(), err
			}
			serviceMember.DutyStation = station
			serviceMember.DutyStationID = &stationID
		}

		if payload.Affiliation != nil {
			serviceMember.Affiliation = (*models.ServiceMemberAffiliation)(payload.Affiliation)
		}

		if payload.Rank != nil {
			serviceMember.Rank = (*models.ServiceMemberRank)(payload.Rank)
		}
	}

	if payload.Edipi != nil {
		serviceMember.Edipi = payload.Edipi
	}

	if payload.FirstName != nil {
		serviceMember.FirstName = payload.FirstName
	}
	if payload.MiddleName != nil {
		serviceMember.MiddleName = payload.MiddleName
	}
	if payload.LastName != nil {
		serviceMember.LastName = payload.LastName
	}
	if payload.Suffix != nil {
		serviceMember.Suffix = payload.Suffix
	}
	if payload.Telephone != nil {
		serviceMember.Telephone = payload.Telephone
	}
	if payload.SecondaryTelephone != nil {
		serviceMember.SecondaryTelephone = payload.SecondaryTelephone
	}
	if payload.PersonalEmail != nil {
		serviceMember.PersonalEmail = payload.PersonalEmail
	}
	if payload.PhoneIsPreferred != nil {
		serviceMember.PhoneIsPreferred = payload.PhoneIsPreferred
	}
	if payload.EmailIsPreferred != nil {
		serviceMember.EmailIsPreferred = payload.EmailIsPreferred
	}

	if payload.ResidentialAddress != nil {
		if serviceMember.ResidentialAddress == nil {
			serviceMember.ResidentialAddress = addressModelFromPayload(payload.ResidentialAddress)
		} else {
			updateAddressWithPayload(serviceMember.ResidentialAddress, payload.ResidentialAddress)
		}
	}
	if payload.BackupMailingAddress != nil {
		if serviceMember.BackupMailingAddress == nil {
			serviceMember.BackupMailingAddress = addressModelFromPayload(payload.BackupMailingAddress)
		} else {
			updateAddressWithPayload(serviceMember.BackupMailingAddress, payload.BackupMailingAddress)
		}
	}

	return validate.NewErrors(), nil
}
```

pkg/handlers/internalapi/service_members_test.go
monfresh: I think we still need more. We want to test every possible field that can be patched as listed in `PatchServiceMemberPayload`.

pkg/handlers/internalapi/service_members_test.go
YanZ777: I'm getting 8 instead of 6.

pkg/handlers/internalapi/service_members.go
```diff
@@ -185,8 +185,26 @@ func (h PatchServiceMemberHandler) Handle(params servicememberop.PatchServiceMem
 		return handlers.ResponseForError(logger, err)
 	}
 
+	allowEditIfMoveInDraft := true
+	if serviceMember.Orders != nil && len(serviceMember.Orders) > 0 {
+		orders := serviceMember.Orders[0]
```
monfresh, line 190: Another thing that feels strange here is that we are using the first Order that gets returned, which might not necessarily be the one that is currently being updated. At some point, we will have service members with multiple orders, but not at the same time I think. 

Once an order is complete, if they have to move again, they will get a new set of orders. So let's say a service member has already moved once a few years ago. They now have one Order in the database. Then, a few years later, they start a new Move, but they haven't yet uploaded the Orders, and they want to edit their Profile. Using this logic, we will look up their first Order, and then the Move for that order, which was completed years ago, and is therefore not a Draft anymore, and they won't be able to update their Profile.

We should make sure to test this scenario as well (where a service member has multiple orders).

pkg/handlers/internalapi/service_members.go
```diff
@@ -185,8 +185,26 @@ func (h PatchServiceMemberHandler) Handle(params servicememberop.PatchServiceMem
...
+		moves := orders.Moves
+		var draftMoves []*models.Move
+		for i := range moves {
```
monfresh, line 193: Is it possible for a service member's order to have multiple moves? When a service member uploads their orders documents, that creates the Order, and then the Move. Can a service member have multiple Moves within the same Order?

YanZ777: From what I understand based off the following slack thread, a service member can only have 1 move within an order. 
https://ustcdp3.slack.com/archives/CNTDW9Q1Z/p1615499703025500

pkg/handlers/internalapi/service_members_test.go
YanZ777: Once @Ronolibert merges his, I'll add more tests to the `PatchServiceMemberHandler`.

amanda5iu (commented):
Hey Yan, this is looking great! I like the table treatment. 

One small tweak: can we make sure there's 3 units of padding between `Service info` and the top of the table? Other than that this looks good to me 👍🏼

pkg/handlers/internalapi/service_members.go
lynzt: @monfresh I think this is out of scope for this change. Supporting multiple is something we should def. keep an eye on, but I don't want to block based on supporting multiple moves.

pkg/handlers/internalapi/service_members_test.go
YanZ777: I'll actually be adding it to his PR if that's ok with you @monfresh.

monfresh: Sure, that works.

pkg/handlers/internalapi/service_members.go
monfresh: I agree that we can wait to support multiple *orders*. My question about the moves was to see if we can simplify this code. If an Order can only ever have a single Move, then we don't need to iterate through the moves. As far as I can tell, creating the order also creates the move at the same time, so I think we can simplify the `isDraftMove` function to this:

```golang
func (h PatchServiceMemberHandler) isDraftMove(serviceMember *models.ServiceMember) bool {
	if serviceMember.Orders == nil || len(serviceMember.Orders) == 0 {
		return true
	}

	move := serviceMember.Orders[0].Moves[0]

	return move.Status == models.MoveStatusDRAFT
}
```

YanZ777: Thanks for the clarification! I've updated that method and run through the manual tests / the tests I've added to make sure it all still works.

amanda5iu (approved):

monfresh (approved):
LGTM. Thanks for making all the changes! Can we please double check that we are indeed expecting 8 addresses to be created?

YanZ777:
> LGTM. Thanks for making all the changes! Can we please double check that we are indeed expecting 8 addresses to be created?

Did some println sleuthing. 
Looks like the extra 2 addresses are coming from:

move.Orders.OriginDutyStation.Address
move.Orders.OriginDutyStation.TransportationOffice.Address 

The rest are coming from where you would expect them as you mentioned earlier (quoted here for ease of reference):
1. Origin Duty Station
2. Transportation Office for Origin DS
3. New Duty Station
4. TO for New DS
5. Backup Address
6. Residential Address

monfresh:
Hmm. Why would it create 2 additional addresses. Shouldn't the move.Orders.OriginDutyStation.Address point to the same address as the service member's Origin Duty Station Address, and the same for the TO address?

pkg/handlers/internalapi/service_members.go
```diff
@@ -198,16 +210,33 @@ func (h PatchServiceMemberHandler) Handle(params servicememberop.PatchServiceMem
...
+			// Fetch the model partially as a validation on the ID
+			station, err := models.FetchDutyStation(h.DB(), stationID)
+			if err != nil {
+				return validate.NewErrors(), err
+			}
+			serviceMember.DutyStation = station
+			serviceMember.DutyStationID = &stationID
+		}
+
+		if payload.Affiliation != nil {
```
lynzt, line 228: I wonder if it would be easier to reason about in the future if you pull `payload.Affiliation` and  `payload.Rank != nil` out of this is draft and put them in their own `isDraftMove` check...

PR #6089, opened 2021-03-10
[MB-6947] EDI Error Tables Database Migration

jacquelineIO:
## Description

This PR updates `payment_requests` table and `PaymentRequest` model to have EDI errors added on to payment requests.

The table is used to record errors with/from EDIs is `edi_errors`. This table should be used to store any errors while sending EDI 858s and from recieving 997s and 824s.

The original design for this had 3 different tables to record errors and a new table to collect all of these errors. After reviewing the code this seemed more complex than needed and now those tables have been collapsed into one table called `edi_errors`. Before collapsing this, 2 of the tables had all of the same fields and the 3rd was very close to the same minus one field. Missing the one field could potentially be a valid entry in the first 2 identical tables so even more reason to make them all the same. The different records will then be identified by the `edi_type` field to know what kind of EDI this error came from.

You will be able to access all errors for a given `PaymentRequest` by accessing `PaymentRequest.EdiErrors`.

### Notes about the older design

To see the original models before condensing tables:
* [edi_error.go](https://github.com/transcom/mymove/blob/7a338242942dfad5d08a30ed27d95b6b44cda555/pkg/models/edi_error.go)
* [edi_errors_acknowledgement_code_error.go](https://github.com/transcom/mymove/blob/7a338242942dfad5d08a30ed27d95b6b44cda555/pkg/models/edi_errors_acknowledgement_code_error.go)
* [edi_errors_send_to_syncada_error.go](https://github.com/transcom/mymove/blob/7a338242942dfad5d08a30ed27d95b6b44cda555/pkg/models/edi_errors_send_to_syncada_error.go)
* [edi_errors_technical_error_description.go](https://github.com/transcom/mymove/blob/7a338242942dfad5d08a30ed27d95b6b44cda555/pkg/models/edi_errors_technical_error_description.go)
* [create-edi-errors-tables.up.sql](https://github.com/transcom/mymove/blob/dfab464f65b52ceb0ffc2c91adb1139d288232c8/migrations/app/schema/20210330000000_create-edi-errors-tables.up.sql)

In using this old design, the additional code that would be needed to support this is
* checking if a payment request has and an existing `edi_errors` record, find it and use it
* if it doesn't, then create it, then find it and use it

This would allow the `PaymentRequest` model not to get clutter for EDI errors info. However, since we are using `go.Pop`, there is not extra code needed. And the additional creation and checking for and `EdiError` record is not needed. With that in mind, I've dropped the `edi_errors` table, and have moved the same into the `payment_requests` table and `PaymentRequest` model. 

### This affects JIRA Stories:
* [Write migration and model for TED table](https://dp3.atlassian.net/browse/MB-6947)
* [Create a new payment request status to indicate the PR has an unresolved issue](https://dp3.atlassian.net/browse/MB-6948)

### Original design doc for the database tables is obsolete:
Putting here for reference.
* [Recording EDI errors in the database for the 997, 824, and 858](https://docs.google.com/document/d/11W0HTPYy9fAjPVAgh5M1cunLaFT0QZhLHRhFpnQoEes/edit#) <-- THIS PR
* The overall design for processing the 997 and 824 can be found [here](https://docs.google.com/document/d/1e8Zq2dkqZ1nWxJNTI5V-w1iZo0ccVTunmWEAr61F4MA/edit#). <-- additional reference info 

## Reviewer Notes

## Setup

```sh
echo "Code goes here"
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6947) for this change

## Screenshots

n/a


migrations/app/schema/20210310173244_create-edi-errors-tables.up.sql
```diff
@@ -0,0 +1,48 @@
+create type edi_response_type as enum (
```
jacquelineIO, line 1: This can change to just `edi_types`. Because it might be good to add the `EDI858` here too. After reviewing the stories in the backlog, I can see where it might be used for recording start/stop times and potentially recording EDI send to Syncada errors as well.

migrations/app/schema/20210330000000_create-edi-errors-tables.up.sql
```diff
@@ -0,0 +1,69 @@
...
+    description varchar,
+    edi_type varchar,
+    created_at timestamp not null,
+    updated_at timestamp not null
+);
+
+create table edi_errors_technical_error_descriptions (
+     id uuid not null primary key,
+     edi_error_id uuid not null
+         constraint edi_errors_technical_error_descriptions_edi_errors_fkey
```
pearl-truss, line 33: I might be misunderstanding this because fkeys are still a bit confusing to me.[ In my branch](https://github.com/transcom/mymove/compare/pm-mb-6944-process-824-file?expand=1#diff-29b62994ec2bb4bcb70f80082f915a18289eacbde732051c57868ff8d1f97a8fR68) I'm getting a violating foreign key constraint error for this fkey because the EDI_errors table doesn't have the TED yet when I'm trying to save the TED but I don't think I can add the EDI_errors first without having saved the TED. It seems like something circular is happening

pearl-truss (commented):
Thanks for getting this out! I left a comment about fkeys below and I think the payment request model also needs to be updated to hold the new status

migrations/app/schema/20210330000000_create-edi-errors-tables.up.sql
jacquelineIO: @pearl-truss  I have the payment request status in my branch locally, I had a failing test, should be getting that pushed up very soon.

As far as the EDI Errors table, yes, you should be able to create the EDI Errors table then create the TED table.

pearl-truss: Ooooh you know what, I think I was looking at a previous version of the EDI Errors table that included a field for a slice of TEDs

pearl-truss (approved):
This looks good to me and works in my branch that uses it

lynzt (approved):
do we need indexes on those foreign keys? Postgres doesn't create them automatically...

PR #6092, opened 2021-03-10
[MB-6904] lock rows for job runner

pearl-truss:
## Description

The job runner will be calling the `ProcessReviewedPaymentRequest` service as a re-occurring scheduled task. This PR updates the code so that the payment requests that are being processed by the job runner are "locked" to avoid double submissions and other errors if the next instance of the scheduled task starts before the previous run is finished.

### Selecting Appropriate [Lock Mode](https://www.postgresql.org/docs/current/explicit-locking.html) : 
Postgres offers 4 row level lock modes. To varying degrees these modes affect the ability for other transactions to implement locks and write operations on rows that have been locked by the transaction that invokes the given lock mode. These lock modes do not prevent other transactions from reading/plain selecting locked rows.

**Listed from most strict to least strict**

- **For Update**: FOR UPDATE causes the rows retrieved by the SELECT statement to be locked as though for update. This prevents them from being locked, modified or deleted by other transactions until the current transaction ends. That is, other transactions that attempt UPDATE, DELETE, SELECT FOR UPDATE, SELECT FOR NO KEY UPDATE, SELECT FOR SHARE or SELECT FOR KEY SHARE of these rows will be blocked until the current transaction ends;
- **For No Key Update**: Behaves similarly to FOR UPDATE, except that the lock acquired is weaker: this lock will not block SELECT FOR KEY SHARE commands that attempt to acquire a lock on the same rows. 
- **For Share**: Behaves similarly to FOR NO KEY UPDATE, except that it acquires a shared lock rather than exclusive lock on each retrieved row. A shared lock blocks other transactions from performing UPDATE, DELETE, SELECT FOR UPDATE or SELECT FOR NO KEY UPDATE on these rows, but it does not prevent them from performing SELECT FOR SHARE or SELECT FOR KEY SHARE.
- **For Key Share**: Behaves similarly to FOR SHARE, except that the lock is weaker: SELECT FOR UPDATE is blocked, but not SELECT FOR NO KEY UPDATE. A key-shared lock blocks other transactions from performing DELETE or any UPDATE that changes the key values, but not other UPDATE, and neither does it prevent SELECT FOR NO KEY UPDATE, SELECT FOR SHARE, or SELECT FOR KEY SHARE.

### [Locking Clauses](https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE)
Adding a locking clause to the lock mode ^^ can control what happens when the current lock encounters a row that has been locked by another process. There are 2 locking clauses: skip locked and nowait

- Without a locking clause: it will wait

- `Skip locked`: will skip it

- `No wait`: will throw an error

#### I propose that we use `skip locked`. There is a note in the docs:

> Skipping locked rows provides an inconsistent view of the data, so this is not suitable for general purpose work, but can be used to avoid lock contention with multiple consumers accessing a queue-like table

However, I believe this may (or may not --please let me know!) be okay since we'll be displaying the data in the TIO/TXO view which is basically a queue table

### Testing

- ~~I've been researching ways to test the locking portion and haven't found any examples. I did find [an opinion](https://stackoverflow.com/a/19196799) that locking may be untestable and to use goroutines instead. I would appreciate any thoughts on this~~ No longer relevant, Jacquie pointed out that we're not using locks on shared memory which is what the discussion in this link was specifically referring to
- ~~There's a failing server test for a nil pointer being dereferenced in `payment_request_processor_test` that I'm currently debugging~~ I've resolved the server test failures mentioned here
- I'm currently working on figuring out how to make the tests that cover case of multiple transactions trying to process the same payment request...

## References

* [Postgres Locking Documentation](https://www.postgresql.org/docs/current/explicit-locking.html) 
* [slack thread with TTV and design](https://ustcdp3.slack.com/archives/CP6F568DC/p1615475356124600)


pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -72,76 +72,76 @@ func InitNewPaymentRequestReviewedProcessor(db *pop.Connection, logger Logger, s
...
-	// Fetch all payment request that have been reviewed
-	reviewedPaymentRequests, err := p.reviewedPaymentRequestFetcher.FetchReviewedPaymentRequest()
-	if err != nil {
-		return fmt.Errorf("function ProcessReviewedPaymentRequest failed call to FetchReviewedPaymentRequest: %w", err)
-	}
+	transactionError = p.db.Transaction(func(tx *pop.Connection) error {
+		// Fetch all payment request that have been reviewed
+		reviewedPaymentRequests, err := p.reviewedPaymentRequestFetcher.FetchAndLockReviewedPaymentRequest()
+		if err != nil {
+			return fmt.Errorf("function ProcessReviewedPaymentRequest failed call to FetchReviewedPaymentRequest: %w", err)
```
jacquelineIO, line 81: So if there are records that are NOT locked but are in the reviewed state, those records will be retrieved and locked, meanwhile ignoring any previously locked records in the extremely rare case we have this happen. I don't expect this scenario, just curious if that's what you think would happen?

Or is this all or nothing? Meaning, it will attempt to retrieve all REVIEWED records and lock them and fail if it can't?

jacquelineIO: This might be where `skip lock` comes in to play (thinking about our Slack convo)

pearl-truss: If it encounters a reviewed payment request that is already locked by another transaction:
Currently: it will wait

`Skip lock`:  will skip it

`No wait`: throw an error 

[docs for lock clauses](When it comes to the [locking clauses](https://www.postgresql.org/docs/9.4/sql-select.html#SQL-FOR-UPDATE-SHARE))

pearl-truss: @jacquelineIO I added to the description, that I think we would want to use `no wait` since `skip locked` has a note in the docs that it can lead to data inconsistency and isn't for general purpose uses

jacquelineIO: @pearl-truss  looks like you went with `skip locked` instead? https://github.com/transcom/mymove/pull/6092/files#diff-b6d11133bebe048babb9e9d70df8c8c838de92b425f4ab7356da539e683f62bbR49

pearl-truss: @jacquelineIO yeah sorry, I updated the description and code after reading some more

pearl-truss: I propose that we use skip locked. There is a note in the docs:
> Skipping locked rows provides an inconsistent view of the data, so this is not suitable for general purpose work, but can be used to avoid lock contention with multiple consumers accessing a queue-like table

However, I believe this may (or may not --please let me know!) be okay since we'll be displaying the data in the TIO/TXO view which is basically a queue table

pearl-truss: Apparently @gidjin specified in the [job runner design doc](https://docs.google.com/document/d/1CTexWoGYPQnmBZuO6LTphp9zS1KKcWJjFunStK0pPbw/edit#) to use FOR UPDATE SKIP LOCKED so I feel good about the choices here

pkg/services/payment_request/payment_request_reviewed_fetcher.go
```diff
@@ -29,3 +29,30 @@ func (p *paymentRequestReviewedFetcher) FetchReviewedPaymentRequest() (models.Pa
 	}
 	return reviewedPaymentRequests, err
 }
+
+const limitOfPRsToProcess int = 100
+
+// const lockTimeout string = "1s"
+
+//FetchAndLockReviewedPaymentRequest finds all payment request with status 'reviewed'
+func (p *paymentRequestReviewedFetcher) FetchAndLockReviewedPaymentRequest() (models.PaymentRequests, error) {
```
reggieriser, line 38: I think the expectation is that this method is called within the context of a transaction.  If it's not, I don't think anything would be locked since it's in autocommit mode.  Is there a way to check at the beginning of the function to ensure a transaction is active and fail with an error if not?

pkg/services/payment_request/payment_request_reviewed_fetcher.go
```diff
@@ -29,3 +29,30 @@ func (p *paymentRequestReviewedFetcher) FetchReviewedPaymentRequest() (models.Pa
...
+	var reviewedPaymentRequests models.PaymentRequests
+	var err error
+	err = p.db.RawQuery(`SET LOCAL lock_timeout = '1s';`).Exec()
```
reggieriser, line 41: How do you anticipate this timeout being utilized?  Given the `SKIP LOCKED`, I'm wondering where we might see a query get stuck in this transaction.

pkg/services/payment_request/payment_request_reviewed_fetcher_test.go
```diff
@@ -33,3 +33,41 @@ func (suite *PaymentRequestServiceSuite) TestFetchReviewedPaymentRequest() {
...
+	})
+
+	// suite.T().Run("throw an error if a locked payment request is updated", func(t *testing.T) {
+	// 	_ = suite.createPaymentRequest(100)
+
+	// 	suite.DB().Transaction(func(tx *pop.Connection) error {
+	// 		_, err := reviewedPaymentRequestFetcher.FetchAndLockReviewedPaymentRequest()
+	// 		suite.NoError(err)
+	// 		return err
+	// 	})
```
reggieriser, line 55: I believe the way this is set up will be two back-to-back transactions and you won't really run into contention (which may be why you commented it out).  You'll have to leave the first transaction open -- maybe a goroutine with a sleep? -- while you try to do the second transaction and figure out if it blocked.  Sounds tricky.

pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -72,76 +72,65 @@ func InitNewPaymentRequestReviewedProcessor(db *pop.Connection, logger Logger, s
...
 	// records for PRs that failed to send
 	var failed []string
-
-	// Send all reviewed payment request to Syncada
-	for _, pr := range reviewedPaymentRequests {
-
-		// generate EDI file
-		var edi858c ediinvoice.Invoice858C
-		edi858c, err = p.ediGenerator.Generate(pr, false)
+	transactionError = p.db.Transaction(func(tx *pop.Connection) error {
```
reggieriser, line 80: I believe the way this is currently written with a transaction over the entire process means that any single failure when generating EDI is going to roll back *all* processed payment requests in the transaction.  I don't think that's what we want because we may have already sent some to Syncada and now the statuses will be wrong.  Or am I misunderstanding the flow?

I wonder if the scope of the transaction should instead be around a single payment request being processed.  In other words, we get a list of all the reviewed payment requests, then create a new transaction for each individual PR being processed (either sequentially or in a goroutine).  This also minimizes contention because we're locking the smallest number of rows for the smallest length of time.

reggieriser (commented):
Good research on our transaction options!  Concurrency can be a tough nut to crack.  I have some comments about how we're processing all of the reviewed PRs here in one big transaction.  Let me know what you think.  Happy to brainstorm if it helps.

pkg/services/payment_request/payment_request_reviewed_processor.go
pearl-truss: I thought about this re: processing a single payment request at a time but from my understanding locks can only be added to a SELECT query. So, I was thinking in this case the PR would already be found so we wouldn't need to reselect it. But as I'm writing this out SELECT and SELECT FOR UPDATE are different so maybe it'll be fine

pkg/services/payment_request/payment_request_reviewed_fetcher.go
pearl-truss: I agree, I think we may not need it now that skip locked has been added

pkg/services/payment_request/payment_request_reviewed_processor_test.go
```diff
@@ -451,3 +454,60 @@ func (suite *PaymentRequestServiceSuite) TestProcessReviewedPaymentRequest() {
...
+		generator,
+		sendToSyncada,
+		gexSender,
+		SFTPSession)
+
+	suite.T().Run("successfully process prs even when a locked row has a delay", func(t *testing.T) {
+		reviewedPaymentRequests := suite.createPaymentRequest(2)
+		for i, pr := range reviewedPaymentRequests {
+			if i == 0 {
+				err := suite.delayedProcessPR(paymentRequestReviewedProcessor, pr)
```
reggieriser, line 495: I'm not sure if this is doing what I think you may want here.  If I'm following it correctly, the "delayed" PR is being processed (lock obtained and released), then it waits for 2 seconds, then the next iteration of the loop happens and the second one processes without any artificial delay.  What scenario are you trying to test here?

pearl-truss: I'm still working on figuring out what to test here. What I really want is to be able to test that a locked row is being skipped if grabbed by a subsequent SELECT FOR UPDATE

pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -70,9 +70,63 @@ func InitNewPaymentRequestReviewedProcessor(db *pop.Connection, logger Logger, s
...
 
+	transactionError = p.db.Transaction(func(tx *pop.Connection) error {
+		var lockedPR models.PaymentRequest
+
+		query := `
+			SELECT * FROM payment_requests
+			WHERE id = $1 FOR UPDATE SKIP LOCKED;
+		`
+		err := p.db.RawQuery(query, pr.ID).First(&lockedPR)
+		if err != nil {
```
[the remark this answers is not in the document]
pearl-truss: Yeah I think it makes sense to silently return w/o an error in that case

pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -80,13 +79,11 @@ func (p *paymentRequestReviewedProcessor) ProcessAndLockReviewedPR(pr models.Pay
 			SELECT * FROM payment_requests
 			WHERE id = $1 FOR UPDATE SKIP LOCKED;
 		`
-		err := p.db.RawQuery(query, pr.ID).First(&lockedPR)
-		if err != nil {
-			return fmt.Errorf("failure retrieving and locking payment request: %w", err)
-		}
+		p.db.RawQuery(query, pr.ID).First(&lockedPR)
```
reggieriser: I don't think we want to entirely ignore the error entirely here -- that can be a red flag to some of the static analysis tools (Ron had a big PR I reviewed recently to clear a lot of this up across the project).  Instead, we want to catch the `sql.ErrNoRows` and return on that, but do flag any other errors (like a database issue).

pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -102,17 +99,12 @@ func (p *paymentRequestReviewedProcessor) ProcessAndLockReviewedPR(pr models.Pay
...
-			UPDATE payment_requests
-			SET
-				status = $1,
-				sent_to_gex_at = $2
-			WHERE id = $3;`
-		err = tx.RawQuery(q, models.PaymentRequestStatusSentToGex.String(), sentToGexAt, lockedPR.ID.String()).Exec()
+		sentToGexAt := time.Now()
+		lockedPR.SentToGexAt = &sentToGexAt
+		lockedPR.Status = models.PaymentRequestStatusSentToGex
+		err = p.db.Update(&lockedPR)
```
reggieriser, line 108: Thanks for converting from a raw query!

reggieriser (commented):
We're close -- just need to not ignore any error when selecting the record.

reggieriser (approved):

PR #6094, opened 2021-03-10
[MB-6949] Add BGN EDI segment and tests

reggieriser:
## Description

This PR adds the BGN segment and its validations as defined in this [mapping doc](https://docs.google.com/spreadsheets/d/1OYYgAp6GD0wG6SKx61De2VRZFkAPHBHZV2BpyvD0nMc/edit#gid=367426795).

## Reviewer Notes

- I added tests for `Parse` and `StringArray` that we don't seem to have in most other segments.  I'm getting 100% coverage (at least according to GoLand) where most of the other segments show 0% coverage.  That's perhaps a little misleading since we do test validations in all segments, but we aren't getting credit for those since those are are defined via struct tags.
- The BGN segment has a Date field, but I'm storing it as a `string` rather than a `time.Time` to match what the other date/time fields do in other segments.  I'm not sure it matters much for this since it's got to be converted to a string anyway to create the EDI.

## Setup

There's no way to use this in the app yet, but `make server_test` will exercise the code.  Also, please verify the fields/validations against the mapping doc.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6949) for this change
* [Spreadsheet of mapped values](https://docs.google.com/spreadsheets/d/1OYYgAp6GD0wG6SKx61De2VRZFkAPHBHZV2BpyvD0nMc/edit#gid=367426795)
* [824 specification](https://drive.google.com/file/d/144rKMw_aFFPgHwLfOhMKqvmkFO4SzJ2M/view?usp=sharing)

### DIFF
--- a/pkg/edi/segment/bgn.go
+++ b/pkg/edi/segment/bgn.go
@@ -1,0 +1,36 @@
+package edisegment
+
+import (
+	"fmt"
+)
+
+// BGN represents the BGN EDI segment
+type BGN struct {
+	TransactionSetPurposeCode string `validate:"eq=11"`
+	ReferenceIdentification   string `validate:"min=1,max=30"`
+	Date                      string `validate:"datetime=20060102"`
+}
+
+// StringArray converts BGN to an array of strings
+func (s *BGN) StringArray() []string {
+	return []string{
+		"BGN",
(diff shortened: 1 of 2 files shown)


jacquelineIO (approved):
lgtm

PR #6093, opened 2021-03-10
[DRAFT for review] AK5 Description Helper

jacquelineIO:
## Description

This is a draft to consider if we want to have data elements defined when we know we'll be using them for certain segments. 

This is using the 717 Data Element Dictionary definition from the EDI 997 spec. 

https://drive.google.com/file/d/13cbg99nrrQP1PwkFYvBTlJjqD_b5ieGB/view

## Reviewer Notes

Is there anything you would like reviewers to give additional scrutiny?

## Setup

Add any steps or code to run in this section to help others prepare to run your code:

```sh
echo "Code goes here"
```

## Code Review Verification Steps

* [ ] If the change is risky, it has been tested in experimental before merging.
* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#logging)
* [ ] The requirements listed in [Querying the Database Safely](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#querying-the-database-safely) have been satisfied.
* Any new migrations/schema changes:
  * [ ] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://github.com/transcom/mymove/wiki/migrate-the-database#zero-downtime-migrations))
  * [ ] Have been communicated to #g-database
  * [ ] Secure migrations have been tested following the instructions in our [docs](https://github.com/transcom/mymove/wiki/migrate-the-database#secure-migrations)
* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](tbd) for this change
* [this article](tbd) explains more about the approach used.

## Screenshots

If this PR makes visible UI changes, an image of the finished UI can help reviewers and casual
observers understand the context of the changes. A before image is optional and
can be included at the submitter's discretion.

Consider using an animated image to show an entire workflow instead of using multiple images. You may want to use GIPHY CAPTURE for this! 📸

_Please frame screenshots to show enough useful context but also highlight the affected regions._

PR #6100, opened 2021-03-11
[MB-6951] Add TED segment and tests

reggieriser:
## Description

This PR adds the TED segment and its validations as defined in this [mapping doc](https://docs.google.com/spreadsheets/d/1OYYgAp6GD0wG6SKx61De2VRZFkAPHBHZV2BpyvD0nMc/edit#gid=367426795).

## Reviewer Notes

- I added tests for `Parse` and `StringArray` that we don't seem to have in most other segments.  I'm getting 100% coverage (at least according to GoLand) where most of the other segments show 0% coverage.  That's perhaps a little misleading since we do test validations in all segments, but we aren't getting credit for those since those are are defined via struct tags.

## Setup

There's no way to use this in the app yet, but `make server_test` will exercise the code.  Also, please verify the fields/validations against the mapping doc.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6951) for this change
* [Spreadsheet of mapped values](https://docs.google.com/spreadsheets/d/1OYYgAp6GD0wG6SKx61De2VRZFkAPHBHZV2BpyvD0nMc/edit#gid=367426795)
* [824 specification](https://drive.google.com/file/d/144rKMw_aFFPgHwLfOhMKqvmkFO4SzJ2M/view?usp=sharing)

### DIFF
--- a/pkg/edi/segment/ted.go
+++ b/pkg/edi/segment/ted.go
@@ -1,0 +1,33 @@
+package edisegment
+
+import (
+	"fmt"
+)
+
+// TED represents the TED EDI segment
+type TED struct {
+	ApplicationErrorConditionCode string `validate:"oneof=007 812 832 DUP IID INC K MJ PPD T ZZZ"`
+	FreeFormMessage               string `validate:"omitempty,max=60"`
+}
+
+// StringArray converts TED to an array of strings
+func (s *TED) StringArray() []string {
+	return []string{
+		"TED",
+		s.ApplicationErrorConditionCode,
+		s.FreeFormMessage,
+	}
+}
+
+// Parse parses an X12 string that's split into an array into the TED struct
+func (s *TED) Parse(elements []string) error {
+	expectedNumElements := 2
+	if len(elements) != expectedNumElements {
+		return fmt.Errorf("TED: Wrong number of fields, expected %d, got %d", expectedNumElements, len(elements))
+	}
+
+	s.ApplicationErrorConditionCode = elements[0]
+	s.FreeFormMessage = elements[1]
+
+	return nil
+}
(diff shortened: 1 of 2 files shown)


mboren (approved):
The definition and validations look like they match the spec, tests pass locally, and I don't see any issues with the code.

pkg/edi/segment/ted.go
```diff
@@ -0,0 +1,33 @@
+package edisegment
+
+import (
+	"fmt"
+)
+
+// TED represents the TED EDI segment
+type TED struct {
+	ApplicationErrorConditionCode string `validate:"oneof=007 812 832 DUP IID INC K MJ PPD T ZZZ"`
```
jacquelineIO, line 9: I'm not sure we want to reject a message that we received. I realize these are the valid fields, and I'm thinking (this could be wrong) that we want to be able to map the messages to the descriptions, which is what I'm playing around with here https://github.com/transcom/mymove/pull/6093. I think ultimately we need to know that we got this message and store into the errors table that we don't know what this field means. Or else, we'll have a dangling EDI with no apparent 824. So the EDI is not getting paid and we have no clue that something is wrong.

reggieriser: We can relax this if needed, but we control when/if the validations are applied (by calling the `validator.Struct` method in the validator library).  On their own, the struct tags don't prevent you from assigning any values to the struct (assuming you adhere to the Go types).  For segments like these that we're getting from elsewhere as opposed to creating, we might just view the validations as more like documentation than anything we actively validate against.

jacquelineIO: good to know thanks! We'll have to make sure this is understand when we do the parsing.

PR #6101, opened 2021-03-11
[MB-6914] Create AK9 EDI segment

mboren:
## Description
This PR creates the AK9 EDI segment with validations and tests.
I plagiarized a lot of this code from #6084 (thanks Jacquie!)

## Reviewer Notes

I couldn't find the potential values for the optional fields in this segment, AK905, AK906, AK907, AK908, or AK909,
They are supposed to be defined in Data Element 716, but that is missing from the data element dictionary section of the [997 spec](https://drive.google.com/file/d/13cbg99nrrQP1PwkFYvBTlJjqD_b5ieGB/view?usp=sharing).
I think this is fine, but if we ever find these definitions, we might be able to have tighter validations for these fields.

## Setup
Nothing in the app uses this segment yet, so we can only run its tests for now:
```sh
make server_test
```

Please also double check the fields and validations against [this spreadsheet](https://docs.google.com/spreadsheets/d/1OYYgAp6GD0wG6SKx61De2VRZFkAPHBHZV2BpyvD0nMc/edit#gid=0)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6914) for this change
* [Spreadsheet of mapped values](https://docs.google.com/spreadsheets/d/1OYYgAp6GD0wG6SKx61De2VRZFkAPHBHZV2BpyvD0nMc/edit#gid=0) explains more about the approach used.
* [997 specification](https://drive.google.com/file/d/13cbg99nrrQP1PwkFYvBTlJjqD_b5ieGB/view?usp=sharing)

### DIFF
--- a/pkg/edi/segment/ak9_test.go
+++ b/pkg/edi/segment/ak9_test.go
@@ -196,4 +196,29 @@ func (suite *SegmentSuite) TestParseAK9() {
 			suite.Contains(err.Error(), "Wrong number of elements")
 		}
 	})
+
+	suite.T().Run("parse fails for invalid ints", func(t *testing.T) {
+		var validOptionalAK9 AK9
+		arrayInvalidIntsAK9 := []string{"A", "g", "2", "3", "", "", "", "", ""}
+
+		err := validOptionalAK9.Parse(arrayInvalidIntsAK9)
+		if suite.Error(err) {
+			suite.Contains(err.Error(), "invalid syntax")
+		}
+
+		arrayInvalidIntsAK9[1] = "1"
+		arrayInvalidIntsAK9[2] = "AAA"
+
+		err = validOptionalAK9.Parse(arrayInvalidIntsAK9)
+		if suite.Error(err) {
+			suite.Contains(err.Error(), "invalid syntax")
+		}
+		arrayInvalidIntsAK9[2] = "2"
+		arrayInvalidIntsAK9[3] = "3.0"
+
+		err = validOptionalAK9.Parse(arrayInvalidIntsAK9)
+		if suite.Error(err) {
+			suite.Contains(err.Error(), "invalid syntax")
+		}
+	})
 }
--- a/pkg/edi/segment/ak9_test.go
+++ b/pkg/edi/segment/ak9_test.go
@@ -80,9 +80,9 @@ func (suite *SegmentSuite) TestValidateAK9() {
 		// length of characters are more than max
 		ak9 := AK9{
 			FunctionalGroupAcknowledgeCode:      "AA",
-			NumberOfTransactionSetsIncluded:     1,
-			NumberOfReceivedTransactionSets:     2,
-			NumberOfAcceptedTransactionSets:     3,
+			NumberOfTransactionSetsIncluded:     1000000,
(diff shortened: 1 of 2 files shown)


pkg/edi/segment/ak9_test.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+	})
+
+	suite.T().Run("validate failure for invalid FunctionalGroupAcknowledgeCode", func(t *testing.T) {
+		validAK9 := AK9{
+			FunctionalGroupAcknowledgeCode:  "B",
+			NumberOfTransactionSetsIncluded: 1,
+			NumberOfReceivedTransactionSets: 2,
+			NumberOfAcceptedTransactionSets: 3,
+		}
```
reggieriser, line 57: Nothing to change here, but just a note ... if I have a struct of value fields that's mostly the same across several tests, to save some repetition I sometimes create a baseline one outside of any subtest and then just use a simple assignment (which copies all the value fields) and then just change the field(s) I'm interested in for the specific subtest.

pkg/edi/segment/ak9_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+		err := suite.validator.Struct(ak9)
+		suite.ValidateError(err, "FunctionalGroupAcknowledgeCode", "oneof")
+		suite.ValidateError(err, "NumberOfTransactionSetsIncluded", "min")
+		suite.ValidateError(err, "NumberOfReceivedTransactionSets", "min")
+		suite.ValidateError(err, "NumberOfAcceptedTransactionSets", "min")
+		suite.ValidateErrorLen(err, 4)
+	})
+}
```
reggieriser, line 119: I think we may be missing the `max` checks on the `int` fields?

pkg/edi/segment/ak9_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+	suite.T().Run("wrong number of elements", func(t *testing.T) {
+		badArrayAK9 := []string{"A", "abc"}
+		var badAK9 AK9
+		err := badAK9.Parse(badArrayAK9)
+		if suite.Error(err) {
+			suite.Contains(err.Error(), "Wrong number of elements")
+		}
+	})
+}
```
reggieriser, line 196: If we wanted to go for 100% coverage, we could write some tests that check the error path for the `int` fields when we pass a `string` that can't be converted to an int.

reggieriser (approved):
Great job!  Lots of nice tests.  I did see a couple of places where we could test a few more spots to get even better coverage, but that's not a blocker, so going ahead and approving.  No need for me to re-review if you end up adding more tests.

pkg/edi/segment/ak9_test.go
mboren: Oops! Thanks for catching that.

pkg/edi/segment/ak9_test.go
mboren: I'll add a test for that

pkg/edi/segment/ak9_test.go
mboren: I was thinking about that after reviewing one of your recent PRs, I'll keep in mind for the future

PR #6105, opened 2021-03-12
MB-4667 Update SM duty station when TOO updates order origin duty station

felipe-lee:
## Description

This makes it so that when the TOO updates the origin duty station on an order, the SM duty station also gets updated. Before this they would get out of sync.

## Reviewer Notes

We considered doing this via 2 api calls as mentioned on the jira ticket, but didn't in the end because there's not an office endpoint to edit the customer data. Also this way gives us the added benefit of having both updates in a single transaction.

## Setup

1. Create a move order from the customer side.
2. Log in as a TOO.
3. Open the move that you created and click on "Edit Orders"
4. Change the "Current duty station" and save (you also likely need to set a "SAC" value).
5. As the customer, go to "Review your request" and check that the "Current duty station" has been updated under the "Profile" section.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4667) for this change


suzubara (approved):
I tested this and it looks good!

PR #6117, opened 2021-03-12
[MB-6915] Parse EDI 997

jacquelineIO:
## Description

EDI 997 is a Functional Acknowledgement message that is sent to the MilMove system to simply acknowledge that the corresponding EDI 858 message was received.

This story was to test parsing the 997 which is basically having to write a 997 `Parse` function and test it. I tested with with 3 version of 997s.
1. This is the sample that we did received via SFTP when @gidjin  was doing the USBank testing.
2. Is a mod of the the first one, adding in the optional segments that were not in the first `AK3` and `AK4`.
3. This is another mod of the first, seeing if the `Parse` function can handle loops for different segments. E.g., the functional group, transaction sets, and data segments, etc allow for multiple loops (arrays) of data.

The 997 spec can be found [here](https://drive.google.com/file/d/13cbg99nrrQP1PwkFYvBTlJjqD_b5ieGB/view). 

A-Team detailed mapping out the fields [here](https://docs.google.com/spreadsheets/d/1OYYgAp6GD0wG6SKx61De2VRZFkAPHBHZV2BpyvD0nMc/edit#gid=0). 

There is a [EDI Responses design document](https://docs.google.com/document/d/1e8Zq2dkqZ1nWxJNTI5V-w1iZo0ccVTunmWEAr61F4MA/edit#heading=h.pydiodiv30kf) which talks about how the 997 EDI is used in the MilMove system.

## Reviewer Notes

About mid-way through the final UT case, I changed styles from using long lines of arrays with index to trying to use some shortened variables. I'm curious if there is one that is better to follow, because I'm going to have to change this one test case to be consistent and wanted to get some feedback. 

```go
		// FunctionalGroup 1 > TransactionSet 2 > FunctionalGroupResponse > TransactionSetResponse 2 > Data Segment 1
		ak4String = "AK4*1*2*3*4*MM*bad data goes here 91"
		ak4 = edi997.InterchangeControlEnvelope.FunctionalGroups[0].TransactionSets[1].FunctionalGroupResponse.TransactionSetResponses[1].dataSegments[0].AK4
		suite.validateAK4(ak4String, ak4)

		// FunctionalGroup 1 > TransactionSet 2 > FunctionalGroupResponse > TransactionSetResponse 2 END
		fgIndex := 0
		tsIndex := 1
		tsrIndex := 1

		fgInfo := edi997.InterchangeControlEnvelope.FunctionalGroups[fgIndex]
		tsInfo := fgInfo.TransactionSets[tsIndex]
		fgr := tsInfo.FunctionalGroupResponse
		tsr := fgr.TransactionSetResponses[tsrIndex]
```

## Setup

Nothing to test here. Running UT will test the `Parse` function. When the [final story is implemented to process the 997](https://dp3.atlassian.net/browse/MB-6906), we'll be able to run dev/acceptance testing.

```sh
make server_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6915) for this change

## Screenshots
n/a


pkg/edi/edi997/parser.go
```diff
@@ -0,0 +1,172 @@
...
+			}
+		case "IEA": // trailer to ISA
+			err = e.InterchangeControlEnvelope.IEA.Parse(record[1:])
+			if err != nil {
+				return err
+			}
+		}
+	} // end of scanner loop
+
+	return nil
```
gidjin, line 171: I wonder if we need to validate the data coming back beyond just trying to parse it? Like should we run the validations against what we parsed? I'm thinking probably not at least not until we get to doing something with the messages.

gidjin (approved):
Changes look good to me. I left a couple comments but they don't need to hold things up.

pkg/edi/edi997/parser.go
jacquelineIO: @gidjin there is a lot of validation that we don't do. We also do not check that the segments are in the expected order. But to the point of validation of the segment itself. I kind of think that we don't want to do that, because we'll end up not receiving a response that we were waiting on. I sort of commented on this in another PR https://github.com/transcom/mymove/pull/6100#discussion_r592749753

So if the validation does fails, we still want to get the message. We aren't handling this scenario so some more thought/discovery could be done for this.

pkg/edi/edi997/parser_test.go
```diff
@@ -0,0 +1,710 @@
...
+
+		// FunctionalGroup 1 > TransactionSet 2 > FunctionalGroupResponse > TransactionSetResponse 2 > Data Segment 1
+		ak4String = "AK4*1*2*3*4*MM*bad data goes here 91"
+		ak4 = edi997.InterchangeControlEnvelope.FunctionalGroups[0].TransactionSets[1].FunctionalGroupResponse.TransactionSetResponses[1].dataSegments[0].AK4
+		suite.validateAK4(ak4String, ak4)
+
+		// FunctionalGroup 1 > TransactionSet 2 > FunctionalGroupResponse > TransactionSetResponse 2 END
+		fgIndex := 0
+		tsIndex := 1
+		tsrIndex := 1
```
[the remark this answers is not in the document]
jacquelineIO: OK, I decided to go with the shorter variables in that run case. The shorter run cases still have the long form indexes.

pkg/edi/edi997/parser.go
jacquelineIO: I will write a discovery story to see if we want to explore this some more.

PR #6125, opened 2021-03-15
Hoist SubmitMoveForm initial values up to page level

suzubara:
## Description

This PR moves the `initialValues` passed into the `SubmitMoveForm` up to the Agreement page level. This should prevent Happo tests from failing all the time, since one of those initial values defaults to the current date. This is also an architectural improvement, since most of the time `initialValues` data will come from the API / Redux / some other external source and will need to be mocked out in Storybook anyways.


src/components/Customer/SubmitMoveForm/SubmitMoveForm.stories.jsx
```diff
@@ -15,6 +15,7 @@ export default {
 
 export const DefaultState = (argTypes) => (
   <SubmitMoveForm
+    initialValues={{ signature: '', date: '2021-01-20' }}
```
monfresh, line 18: I see what you did there :)

monfresh (approved):
LGTM.  Thanks for fixing this!

amanda5iu (approved):

PR #6131, opened 2021-03-16
MB-6840 Update SM Affiliation/Rank when Order Branch/Grade are updated

felipe-lee:
## Description

This makes it so that when the TOO updates the rank (a.k.a. grade) on an order, the SM rank also gets updated. Before this they would get out of sync. 

The story was also to make it so the branch/affiliation stay in sync, but this only seems to be saved on the service member model. The code for that was already in place so I just added a test around that logic. 

## Reviewer Notes

In the `pkg/handlers/ghcapi/orders.go` file, there's an `Order` func that seems to be massaging some of the data into an `Order` model of sorts. I didn't use it since this logic is only a stopgap until we've updated everything to look at the orders model as appropriate.

## Setup

1. Create a move order from the customer side.
2. Log in as a TOO.
3. Open the move that you created and click on "Edit Allowances"
4. Change the "Branch" and/or "Rank" and save.
5. As the customer, go to "Review your request" and check that the "Branch" and/or "Rank" has(have) been updated under the "Profile" section.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6840) for this change


Ronolibert (approved):
Works as expected! Yeah I don't think you need to use that Order func in this case, just adds more extra steps than needed

PR #6145, opened 2021-03-16
MB-7019 create storybook component for contact info fieldset

YanZ777:
## Description

Add new storybook component that holds the phone number, alternate phone number, email, and option for selecting preferred contact method.

## Reviewer Notes

This is currently being merged to `sr-7020-residential-address-fieldset` since I wanted to grab some changes from there while working on this story.

CSS! 
Did I put the component in the correct location? It's currently under: `components/Fieldsets/ServiceMemberContactInfoFields`

## Setup

```sh
make storybook
make client_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7019) for this change

## Screenshots

image
image


src/components/form/ServiceMemberContactInfoFields/ServiceMemberContactInfoFields.module.scss
```diff
@@ -0,0 +1,23 @@
...
+    @include at-media('tablet') {
+      width: 61%;
+    }
+  }
+
+  :global(.usa-checkbox) {
+    display: inline-block;
+    margin-right: 40px;
+  }
+}
```
suzubara, line 23: I was kind of hoping no CSS would be needed for this story! are you able to get the desired layout using the existing [generic form CSS](https://github.com/transcom/mymove/blob/master/src/styles/form.module.scss)? We might need to expand on the `.radioGroup` so it's usable for inline checkboxes also.

We also shouldn't need to set any `width` on elements, that can be controlled using the `Grid` components provided by ReactUSWDS. There's a bunch of Storybook examples [here](https://storybook.dp3.us/?path=/story/design-system_components-grid--default-container) and the AddressFields also uses this to put the state & zip fields side-by-side

src/components/form/ServiceMemberContactInfoFields/index.jsx
```diff
@@ -0,0 +1,76 @@
...
+            id={`preferPhone_${ServiceMemberContactInfoFieldsUUID}`}
+            label="Phone"
+            name={`${name}.preferPhone`}
+            onChange={onChangePreferPhone}
+          />
+          <Checkbox
+            id={`prefer_email${ServiceMemberContactInfoFieldsUUID}`}
+            label="Email"
+            name={`${name}.preferEmail`}
+            onChange={onChangePreferEmail}
```
suzubara, line 53: For the Checkboxes, they should be hooked up to Formik like the `TextField` component is, and don't need to be passed an explicit `onChange` handler.

Since there's no corresponding `CheckboxField` component, you could use Formik's `<Field />` ([docs](https://formik.org/docs/api/field)), with the `as={Checkbox}` prop so it renders ReactUSWDS's Checkbox, or you could also set up a CheckboxField component if you want (although I imagine it wouldn't include extra markup like the TextField does).

Formik also provides an example of using `<Field />` with checkbox inputs: https://formik.org/docs/examples/checkboxes

src/components/form/ServiceMemberContactInfoFields/index.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+export const ServiceMemberContactInfoFields = ({
+  legend,
+  className,
+  onChangePreferEmail,
+  onChangePreferPhone,
+  name,
+  render,
+}) => {
+  const ServiceMemberContactInfoFieldsUUID = uuidv4();
```
suzubara, line 18: I just made this update to AddressFields, but this value should be stored in a `useRef`, otherwise the uuid will change on every render (ie, whenever you type into the input). you can see my example here: https://github.com/transcom/mymove/pull/6139/files#diff-b110391453e762afa87747ad47ef60e098df61390d930a6cc049d02170c04b11R64

suzubara (changes requested):
this is mostly there, I made some suggestions for improvements! the most important thing is to get the Formik integration because that will ensure the inputs are hooked up to whatever form they're rendered in correctly

src/components/form/ServiceMemberContactInfoFields/ServiceMemberContactInfoFields.module.scss
```diff
@@ -0,0 +1,23 @@
...
+    width: 100%;
+    @include at-media('tablet') {
+      width: 60%;
+    }
+  }
+
+  :global(.usa-label) {
+    width: 100%;
+    @include at-media('tablet') {
+      width: 61%;
```
Shkeating, line 15: this value needs to be equal to the one you set at line 8 in order for the optional tag be aligned with the end of the field

image

Shkeating: but also listen to suz if u are ditching the css file altogether ~

YanZ777: Does the optional tag need to be aligned with the end of the field? The [Zeplin](https://app.zeplin.io/project/5f80b1b3ec80a85f303afe20/screen/5fca9f5d2c500143eddd9a81) made it look like it shouldn't be, but it does make my life easier if it _does_ need to be aligned.

src/components/form/ServiceMemberContactInfoFields/index.jsx
YanZ777: I'm still working on this part. I think I'll create a CheckboxField component so that way it can be used in other forms as well.

src/components/form/ServiceMemberContactInfoFields/ServiceMemberContactInfoFields.module.scss
suzubara: I'm guessing the Zeplin is mistaken.. this should probably use the same "Optional" treatment as demonstrated in the [kitchen sink form example](https://storybook.dp3.us/?path=/story/components-forms--kitchen-sink-form)

src/components/form/ServiceMemberContactInfoFields/ServiceMemberContactInfoFields.stories.jsx
```diff
@@ -0,0 +1,35 @@
...
+      },
+    }}
+  >
+    {() => (
+      <Form className={formStyles.form}>
+        <ServiceMemberContactInfoFields
+          name="contact"
+          legend="Your contact info"
+          onChangePreferPhone={action('clicked')}
+          onChangePreferEmail={action('clicked')}
```
suzubara, line 30: can these be removed since these props don't exist anymore in this component?

suzubara (approved):
looks great to me!

src/components/form/ServiceMemberContactInfoFields/ServiceMemberContactInfoFields.stories.jsx
YanZ777: Yes!

Shkeating (approved):
lookin goood

amanda5iu (approved):
Looking good ✨

PR #6152, opened 2021-03-17
[MB-7065] DangerJS allow eslint-disable with rules

Ronolibert:
## Description

Don't raise a warning when using eslint-disable as long as it has rules attached

## Reviewer Notes
- My code change removes the check for `eslint-disable` because dangerjs was linting itself and breaking 'eslint-disable' was being used in a context that was not in a comment. Normal circumstances, it would be a syntax error but in this file we're using `eslint-disable` as a string.

## Setup
1. Change variable to have a dangling `_` (eg. serviceMember_)
2. Add a valid eslint-disable to a file (eg. /* eslint-disable no-underscore-dangle */
3. Commit change
4. Run `npm run dangerjs:local`
Expected result: You should not get an error

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7065) for this change

### DIFF
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -100,15 +100,15 @@ function doesLineHaveProhibitedOverride(disablingString) {
     .split(/[\s,]+/)
     .map((item) => item.trim());
   // disablingStringParts format: ['eslint-disable-next-line', 'no-jsx', 'no-default']
-  if (disablingStringParts[0] === 'eslint-disable') {
-    // fail because don't disable whole file please!
-    prohibitedOverrideMsg =
-      'Found `eslint-disable`. This disables the whole file, which is bad practice because it can allow security issues to slip in unnoticed. Please specify exact rules on a line by line basis, eg `eslint-disable-next-line no-underscore-dangle`.';
-  }
 
   if (disablingStringParts.length === 1) {
-    // fail because rule should be specified
-    prohibitedOverrideMsg = 'Must specify the rule you are disabling';
+    if (disablingStringParts[0] === 'eslint-disable') {
+      prohibitedOverrideMsg =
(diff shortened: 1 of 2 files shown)


Ronolibert:
@kimallen I opted to just remove the block because dangerjs was linting itself and breaking because there was an `eslint-disable` that was not a comment. In regular code situations, a non-comment eslint-disable would be a syntax error

suzubara (commented):
I am still getting this error after adding `/* eslint-disable no-underscore-dangle */` to a file:
```
## Warnings
-
It looks like you are attempting to bypass a linter rule, which is not within
      security compliance rules.
** Contains a rule that is not in the permitted eslint list. You are free to disable only: (
no-underscore-dangle
,prefer-object-spread
,object-shorthand
,camelcase
,react/jsx-props-no-spreading
,react/destructuring-assignment
,react/forbid-prop-types
,react/prefer-stateless-function
,react/sort-comp
,import/no-extraneous-dependencies
,import/order
,import/prefer-default-export
,import/no-named-as-default
) **
 Please remove the bypass code and address the underlying issue. cc: @transcom/Truss-Pamplemoose
```

I debugged a little bit and it looks like the `*/` part of the string is being passed into the `doesLineHaveProhibitedOverride` function and also being checked against the array of allowed rules, so I'm guessing that's the problem. I also tried adding just `/* eslint-disable */` to a file, which fails with the same issue as above, but I think we still want to make sure to prevent that, right?

Ronolibert:
@suzubara Ahh, I think when I tested it it was before I committed! Fixed that edge case where the closing comment was being included in that rules array. Yeah, we still want that behavior where it has a warning if there's a bare `eslint-disable` or `eslint-disable-next-line`

PR #6153, opened 2021-03-17
[MB-6946] Create new model and table for EDI response processing info

pearl-truss:
## Description

This PR creates a new table and model for capturing data related to processing edi response messages. It was based off of this [doc](https://docs.google.com/document/d/1e8Zq2dkqZ1nWxJNTI5V-w1iZo0ccVTunmWEAr61F4MA/edit#heading=h.bhd1p9ls5hrb)

The new table and model holds the following info: 
Message type (997, 824, 858), start time of processing, end time of processing, number of messages processed

## Reviewer Notes
I currently have the model to require that all fields are present. I'm not sure if we actually want this for validation? 

## Setup

```sh
make db_dev_migrate
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6946) for this change
* [Job Runner EDI Response Design Doc](https://docs.google.com/document/d/1e8Zq2dkqZ1nWxJNTI5V-w1iZo0ccVTunmWEAr61F4MA/edit#heading=h.bhd1p9ls5hrb)


migrations/app/schema/20210317151448_create_edi_response_processing_table.up.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TYPE edi_response_message_type AS ENUM (
```
jacquelineIO, line 1: Can you change this from `edi_response_message_type` to `edi_type` and add in the `858` as well? After thinking about this is in another PR, I realized we should just add all of the EDIs, we have another scenario where we'll be needed the `858`.

migrations/app/schema/20210317151448_create_edi_response_processing_table.up.sql
```diff
@@ -0,0 +1,14 @@
...
+);
+
+CREATE TABLE edi_response_processing
+(
+	id uuid PRIMARY KEY NOT NULL,
+	message_type edi_response_message_type NOT NULL,
+    num_messages_processed INT NOT NULL,
+	process_started_at timestamp NOT NULL,
+	process_ended_at timestamp NOT NULL
+)
```
jacquelineIO, line 14: I think you also have to add these to all tables when creating them? But I'm not completely sure, going to poke at some other migrations and see
```sql
    created_at timestamp not null,
    updated_at timestamp not null,
    deleted_at timestamp with time zone
```

migrations/app/schema/20210317151448_create_edi_response_processing_table.up.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TYPE edi_response_message_type AS ENUM (
+    '997',
+    '824',
+	'810'
+);
+
+CREATE TABLE edi_response_processing
```
jacquelineIO, line 7: oh and change `edi_response_processing` to `edi_processing`. going to also add the `858` to this table.

migrations/app/schema/20210317151448_create_edi_response_processing_table.up.sql
pearl-truss: I checked and I think we might not need `deleted_at` because I'm seeing examples like 20191023200010_create_mto_required_fields_tables.up.sql that don't have that one

jacquelineIO: great @pearl-truss  sounds good!

reggieriser: That's correct -- if you're doing a `CREATE TABLE` (as opposed to using the fizz DSL that we used to use), you have to include `created_at` and `updated_at` as above because Pop will attempt to write to those fields automatically.  The `deleted_at` is only added in cases where we want to do soft deletes (and Pop doesn't do any magic with it on its own).

migrations/app/schema/20210317151448_create_edi_response_processing_table.up.sql
```diff
@@ -0,0 +1,14 @@
...
+(
+	id uuid PRIMARY KEY NOT NULL,
```
reggieriser, line 9: I tend to give constraint names to keys (primary and foreign) and constraints so it's clear what we need to reference if we want to remove/update/etc. them later.  Here's an [example](https://github.com/transcom/mymove/blob/master/migrations/app/schema/20191010165703_re_create_misc_pricing_tables.up.sql).  If we don't name them, we have to assume the name based on whatever Postgres assigns (which hopefully is predictable).

jacquelineIO (approved):
lgtm

migrations/app/schema/20210318162232_create_edi_processing_table.up.sql
```diff
@@ -0,0 +1,18 @@
...
+(
+	id uuid
+	    constraint edi_processing_pkey primary key,
+	edi_type edi_type NOT NULL,
+    num_edis_processed INT NOT NULL,
+	process_started_at timestamp NOT NULL,
+	process_ended_at timestamp NOT NULL,
+	created_at timestamp not null,
+    updated_at timestamp not null
+);
```
reggieriser, line 18: Nit: it looks like you've got some tabs on some lines and spaces on others given the odd indenting shown above.  Also, you use all caps `NOT NULL` in some places and `not null` in others.

reggieriser (approved):
I added a small formatting/casing nit, but not a blocker.  I didn't see any indexes, but I'm not sure what makes the most sense for this table yet (if any).  We can always add those if applicable when we write code that queries this table.  Otherwise, LGTM!

migrations/app/schema/20210318162232_create_edi_processing_table.up.sql
lynzt: does this need any indexes?

pearl-truss: @lynzt I'm sorry I missed this comment! I'm not sure if it needs indexes. @reggieriser do you know if it does? If it does I can make a quick chore to address

reggieriser: I hesitate to add any indexes until we know the specific queries we may be executing against this table so we can create the appropriate type of index (if we were dealing with foreign keys, it may be easier to just add indexes for them at table creation time).  I'm guessing we may be showing some of these records by a date range and/or sorted by date, but not sure yet.

PR #6154, opened 2021-03-17
MB-6676 Increase tap area on shipment type selector

Shkeating:
## Description
This branch uses a css trick to increase the tappable area of a **SelectableCard** using the` ::after` pseudo element on the `label` in the component.

## Reviewer Notes
This is not a perfect solution in that I cannot get the `::after` that is increasing the tap area (which has to be affiliated with the input and label element) can only be set to a fixed height. This unfortunately cannot dynamically resize if there are multiple lines of text in a **SelectableCard**, or when the screen gets really narrow, causing the words per line to decrease and make the card taller. I also left some pace on the side so our target doesn't overlap when there is an info icon on a selctable card instance. It is still a much larger, easier to use touch target than what is there currently.

## Setup

In app:
```sh
make client_run
make server_run
```
Create a new milmove user
Take a look at the conus/oconus page, see that the clickable area is much larger than the label and radio button.

In Storybook:
```sh
yarn storybook
```
[localhost:6006/?path=/story/customer-components-selectablecard](localhost:6006/?path=/story/customer-components-selectablecard)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6676) for this change

## Screenshots

area shown in red is the tappable area
image

image


suzubara (commented):
the code changes here seem ok, but using a pseudo element like this doesn't seem ideal 🙁 . If this isn't too urgent, I would recommend waiting for the ReactUSWDS implementation of this style, so we don't need to write more custom CSS (https://github.com/trussworks/react-uswds/issues/982).

However, that component may also not be compatible with our version where we embed a help icon (https://storybook.dp3.us/?path=/story/customer-components-selectablecard--with-help-button). I don't know if we might want to consider an alternative design, since generally I think it's best to avoid a UI where a clickable element is contained within another clickable element.

amanda5iu (approved):
Nice big target area 🤩

Shkeating:
going to close - would rather open new branch when this is unblocked

PR #6159, opened 2021-03-17
MB-6581-add-webhook-subscription-page

NamibiaTorres:
## Description

Add an edit webhook subscription page that should enable the following fields to update: subscriptionId, EventKey, Severity, CallbackURL, Status.

Add one cypress test to test that the new edit page works as expected and updates the appropriate fields.

## Reviewer Notes

When writing my test I noticed that the test associated with the create page isn't working. I updated the tests so that they are now passing. A second pair of eyes on these fixes are appreciated. 

## Setup

To check the new page exists run `make server_run` and `make admin_client_run`. Do a `local sign-in` and then select an existing admin-user to log in. Go to the WebhookSubscription tab and click on one subscription. Then click on "Edit" in the top right hand corner to inspect the Edit page.

To run the tests: `yarn test:e2e`. Then select the "Webhook Subscription" file in the  cypress browser. You should be directed to the tests for all Webhook Subscription pages. 

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6581) for this change

### DIFF
--- a/cypress/integration/admin/webhookSubscriptions.js
+++ b/cypress/integration/admin/webhookSubscriptions.js
@@ -3,6 +3,7 @@ import { adminBaseURL } from '../../support/constants';
 describe('Webhook Subscriptions', function () {
   before(() => {
     cy.prepareAdminApp();
+    cy.clearAllCookies();
   });
 
   it('successfully navigates to the webhook subscriptions list page', function () {
@@ -21,6 +22,7 @@ describe('Webhook Subscriptions', function () {
 describe('WebhookSubscriptions Details Show Page', function () {
   before(() => {
     cy.prepareAdminApp();
+    cy.clearAllCookies();
   });
 
   it('pulls up details page for a webhook subscription', function () {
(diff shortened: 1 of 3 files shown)


cypress/integration/admin/webhookSubscriptions.js
```diff
@@ -43,6 +43,43 @@ describe('WebhookSubscriptions Details Show Page', function () {
...
+        cy.get('a').contains('Edit').click();
+        cy.url().should('eq', adminBaseURL + '/system/webhook_subscriptions/' + $id.text());
+      });
+
+    // check labels on edit page
+    const labels = ['Id', 'Subscriber', 'Status', 'Event key', 'Callback url', 'Created at', 'Updated at', 'Severity'];
+    labels.forEach((label) => {
+      cy.get('label').contains(label);
+    });
+
```
eamahanna, line 71: Would it make sense to check which fields are read-only and which are editable?

eamahanna (approved):
Just a small comment, but everything looks good :)

cypress/integration/admin/webhookSubscriptions.js
NamibiaTorres: Hmm, was reading through this [old cypress PR](https://github.com/cypress-io/cypress/pull/4729), I think that depends if the field has hidden or disabled inputs? I was looking at using `.type()` to check readonly fields.

PR #6161, opened 2021-03-18
[MB-7104] add notification creation endpoint

shimonacarvalho:
## Description

This ticket was to create a new endpoint for webhook notifications but I took the opportunity to clean up the various representations of notifications to all match. 

- Created a POST /webhook-notifications/ endpoint to create test notifications
- Noticed notifications model did not match the db using model-vet, fixed that (Payload should not be a pointer)
- Created a WebhookNotification definition in supportapi to make usage consistent.
- Made webhook-notify endpoint and webhook-client use the same definition
- Added tests for new endpoint
- Make logging useful but not expose payloads 

## Reviewer Notes

Tests that cover this functionality are
`go test ./pkg/handlers/supportapi/ --testify.m TestCreateWebhookNotification -v`
`go test ./cmd/webhook-client/webhook -v`

## Setup

```sh
echo "Code goes here"
```

## References
- [setup postman to make mutual tls api calls · transcom/mymove Wiki](https://github.com/transcom/mymove/wiki/setup-postman-to-make-mutual-tls-api-calls)
- [WebHooks for Milmove - Google Docs](https://docs.google.com/document/d/1tq2jWvuLXEv30Bab2S-nSj9V1bAzLrPk7DZo3xLX9-o/edit#heading=h.5x0d5h95i329)


cmd/webhook-client/webhook/payloads.go
```diff
@@ -0,0 +1,27 @@
...
+		UpdatedAt:       *handlers.FmtDateTime(model.UpdatedAt),
+		ObjectID:        handlers.FmtUUIDPtr(model.ObjectID),
+		MoveTaskOrderID: handlers.FmtUUIDPtr(model.MoveTaskOrderID),
+	}
+
+	if model.TraceID != nil {
+		payload.TraceID = *handlers.FmtUUIDPtr(model.TraceID)
+	}
+	return &payload
+}
```
sandy-wright, line 27: Maybe we could use the `WebhookSubscriptionPayload` function from the Admin API to do this? Only difference could be the `TraceID`, which I'm not sure we have in the Admin API's definition. But there's really no reason not to add it to that definition too. This is a very soft suggestion, for the record!

sandy-wright: oh wait, this is to connect to the support endpoint in particular, so you can disregard this! If you can import a model-to-payload function from the Support API, that might be nice. Otherwise, this is totally fine

pkg/handlers/supportapi/webhook_notification.go
```diff
@@ -0,0 +1,84 @@
...
+
+	logger.Info("Received Webhook Notification: ",
+		zap.String("id", notif.ID.String()),
+		zap.String("eventKey", notif.EventKey),
+		zap.String("createdAt", notif.CreatedAt.String()),
+		zap.String("traceID", notif.TraceID.String()),
+		zap.String("moveID", mtoID),
+		zap.String("objectID", objectID))
+	return webhookops.NewReceiveWebhookNotificationOK().WithPayload(notif)
+}
```
sandy-wright, line 43: So this endpoint just returns whatever it received without making any updates?

pkg/handlers/supportapi/webhook_notification.go
```diff
@@ -0,0 +1,84 @@
...
+	if payload == nil {
+		// create a default notification payload
+		message := "{ \"message\": \"This is a test notification\" }"
+		payload = &supportmessages.WebhookNotification{
+			EventKey: string(event.TestCreateEventKey),
+			TraceID:  *handlers.FmtUUID(h.GetTraceID()),
+			Object:   swag.String(message),
+			Status:   supportmessages.WebhookNotificationStatusPENDING,
+		}
+	}
```
sandy-wright, line 65: nice, I like that it has a default!

sandy-wright (approved):
LGTM! Thanks for cleaning up all this code and all these references, this is really great ✨

cmd/webhook-client/webhook/payloads.go
shimonacarvalho: That one in AdminAPI is subscription, this one is notification, so we can't use the same definition. 
But also, I really really wanted to import the SupportAPI but it's in `internal/` so i couldn't, without moving all the payload/model files. This PR was enough of a catchall that I decided not to go so far. That's a good little pr to do by itself though.

pkg/handlers/supportapi/webhook_notification.go
shimonacarvalho: Yeah it actually just functions as a fake prime server. 

The real value is that it allows us to test the "Send notification using MTLS functionality" on the webhook-client. We don't want the server to take any action though, since in the real world we are not the recipients of this notification.

PR #6163, opened 2021-03-18
[MB-6905, MB-6945] Add service to read files from Syncada via SFTP

mboren:
## Description
We created a service in `pkg/services/invoice/syncada_sftp_reader.go` to read files from an SFTP directory, process them, and delete files that were successfully processed.

This service will attempt to process every file in the remote directory that it can. It does not walk the directory tree, so subdirectories will be ignored. Failures will be logged, and will not stop processing of other files.

Processing is done by passing something that implements `SyncadaFileProcessor` to `syncadareaderSFTPSession.FetchAndProcessFiles`.
The processor should take a file's contents as a string and do whatever needs to be done with it, returning an error if something failed. For testing purposes, we have a simple processor that prints the file path and contents to `stdout`

The processing function should be able to deal with duplicates across different runs because it will be run in a scheduled task, and one execution could overlap with another if it takes longer to run than the time between runs.

`FetchAndProcessFiles` takes a time argument and returns the modified time of the newest file it processed. This can be used to skip over files that we were unable to read or process in previous runs, so we don't log the same error a million times. This has been useful for testing, but I'm not sure it's a feature we will actually want in production, since errors will hopefully be rare and we delete files after processing them.

Currently on the testing server, we have all response files going to the same directory. We're hoping that we will get different kinds of responses (997s and 824s) in different directories.

We also added a command line utility for manual testing:

```sh
# Fetch and process files from syncada:
go run ./cmd/fetch-from-syncada-via-sftp  --directory /CSFTP866/Pickup/

# Fetch, process and delete files from syncada:
go run ./cmd/fetch-from-syncada-via-sftp  --directory /CSFTP866/Pickup/ --delete-after-processing

# Omit files created before a particular time with `--last-read-time`:
go run ./cmd/fetch-from-syncada-via-sftp  --directory /CSFTP866/Pickup/ --last-read-time 2021-03-19T18:25:36Z
```

### Sending EDIs to Syncada
**This has been a bit finicky for me. Sometimes it works and sometimes it doesn't, and i haven't figured out yet whether there's just a long delay, or the files aren't being uploaded. Since sending the files isn't part of this PR, you may want to skip over this and try downloading files that have already been uploaded.**
The devseed data has some payment requests that aren't set up quite right, so the most foolproof way to go about this is to start with a fresh DB and make a move from scratch.

If you do want to use `make db_e2e_populate`, you'll have to go into the `payment_requests` table and update the statuses of any moves that you don't want to send to `SENT_TO_GEX`. 

```sh
make server_run

# in another terminal
make client_run
```

Create a move with a payment request, and approve the payment request as a TIO.

Send EDIs for all reviewed payment requests to Syncada with:
```json
{
  "body": {
    "sendToSyncada": true
  }
}
```
```sh
```

### Download files from Syncada
```
go run ./cmd/fetch-from-syncada-via-sftp  --directory /CSFTP866/Pickup/
```
You should see some EDI files logged.

If you delete the files with:
```
go run ./cmd/fetch-from-syncada-via-sftp  --directory /CSFTP866/Pickup/ --delete-after-processing
```
and then run again, you shouldn't see any more files logged.

## References
We ended up doing both of these stories together:
* [jira story for reading files](https://dp3.atlassian.net/browse/MB-6905)
* [jira story for deletion](https://dp3.atlassian.net/browse/MB-6945)


pkg/cli/syncada_sftp.go
```diff
@@ -0,0 +1,98 @@
...
+// TODO: Figure out what to do in terms of static analysis on the false positive for G101
+const (
+	// SyncadaSFTPPortFlag is the ENV var for the Syncada SFTP port
+	SyncadaSFTPPortFlag string = "syncada-sftp-port"
+	// SyncadaSFTPUserIDFlag is the ENV var for the Syncada SFTP user ID
+	SyncadaSFTPUserIDFlag string = "syncada-sftp-user-id"
+	// SyncadaSFTPIPAddressFlag is the ENV var for the Syncada SFTP IP address
+	SyncadaSFTPIPAddressFlag string = "syncada-sftp-ip-address"
+	// SyncadaSFTPPasswordFlag is the ENV var for the Syncada SFTP password
+	SyncadaSFTPPasswordFlag string = "syncada-sftp-password" // #nosec G101
```
reggieriser, line 22: I will be filing a separate story to get an annotation (and approval) on this false positive per our latest guidance.

pearl-truss:
@mboren @reggieriser I ran into this issue when running the command to read files.

`FATAL   fetch-from-syncada-via-sftp/main.go:109 couldn't initialize SSH client {"error": "failed to parse host key ssh: no key found"`

I'm guessing, I might be missing some environment variables locally?

pearl-truss:
> > I'm guessing, I might be missing some environment variables locally?
> 
> Yes, sounds like it. You may need to try resetting your `.envrc.local` from chamber so you pick up the needed SFTP variables. I'm not sure how you have yours set up, but see if [these instructions help](https://github.com/transcom/mymove#setup-direnv). If you're still having problems, we can try pairing.

Thanks! I was able to download/read the files with `go run ./cmd/fetch-from-syncada-via-sftp  --directory /CSFTP866/Pickup/`. I then ran `go run ./cmd/fetch-from-syncada-via-sftp  --directory /CSFTP866/Pickup/ --delete-files-after-processing` to delete them and I think that worked but when I check by re-running the first command, I got back: 

```sh
 INFO    invoice/syncada_sftp_reader.go:50       Error while downloading Syncada file    {"path": "/CSFTP866/Pickup/.working", "error": "failed to open file over SFTP: permission denied"}
```

mboren:
> > Thanks! I was able to download/read the files with `go run ./cmd/fetch-from-syncada-via-sftp --directory /CSFTP866/Pickup/`. I then ran `go run ./cmd/fetch-from-syncada-via-sftp --directory /CSFTP866/Pickup/ --delete-files-after-processing` to delete them and I think that worked but when I check by re-running the first command, I got back:
> > ```shell
> >  INFO    invoice/syncada_sftp_reader.go:50       Error while downloading Syncada file    {"path": "/CSFTP866/Pickup/.working", "error": "failed to open file over SFTP: permission denied"}
> > ```
> 
> There are some files out there like this `.working` file that are in the directory on the SFTP server but we don't have access to open/read them. Since we don't know the pattern yet for the filenames we should be reading, it's tough to know what to automatically skip. This is just an informational log message. @mboren might have more context.

Yep, Reggie's got it. I think I'll update the log message to be a bit less alarming and add a comment with some context.

jacquelineIO (approved):
lgtm

PR #6171, opened 2021-03-18
MB-7015 Prevent Move approval if Order is invalid

monfresh:
## Description

A move cannot be approved unless the following fields are present
on the Order:

* TAC
* OrdersNumber
* DepartmentIndicator
* OrdersTypeDetail

There are other required Order fields, but those are required for the 
initial creation of the Order, so for this PR, we are only validating the 
fields that are initially optional, but need to be filled in before the TOO 
can approve the Move.

We also want to validate that the TAC must be exactly four 
alphanumeric characters.

To prevent the Move approval, we add an early check in 
`MakeAvailableToPrime` to see if the Order is valid.

To make sure the Order is valid, we add new model validations that
only run if the Move's status is no longer `DRAFT`.

Since I was touching code related to orders and moves, and in the
spirit of leaving code better than I found it, I also did some more
renaming of "move orders" to "orders" and "move task order" to
"move"

## Reviewer Notes

Please see if I missed any test cases.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7015) for this change

### DIFF
--- a/pkg/handlers/ghcapi/move_task_order.go
+++ b/pkg/handlers/ghcapi/move_task_order.go
@@ -1,6 +1,7 @@
 package ghcapi
 
 import (
+	"github.com/gobuffalo/validate/v3"
 	"github.com/gofrs/uuid"
 	"go.uber.org/zap"
 
@@ -77,7 +78,8 @@ func (h UpdateMoveTaskOrderStatusHandlerFunc) Handle(params movetaskorderops.Upd
 		case services.NotFoundError:
 			return movetaskorderops.NewUpdateMoveTaskOrderStatusNotFound()
 		case services.InvalidInputError:
-			return movetaskorderops.NewUpdateMoveTaskOrderStatusBadRequest()
+			payload := payloadForValidationError("Unable to complete request", err.Error(), h.GetTraceID(), validate.NewErrors())
+			return movetaskorderops.NewUpdateMoveTaskOrderStatusUnprocessableEntity().WithPayload(payload)
 		case services.PreconditionFailedError:
 			return movetaskorderops.NewUpdateMoveTaskOrderStatusPreconditionFailed().WithPayload(&ghcmessages.Error{Message: handlers.FmtString(err.Error())})
 		case services.ConflictError:
--- a/pkg/handlers/ghcapi/order_test.go
+++ b/pkg/handlers/ghcapi/order_test.go
@@ -21,9 +21,9 @@ import (
 )
 
 func (suite *HandlerSuite) TestGetOrderHandlerIntegration() {
-	moveTaskOrder := testdatagen.MakeDefaultMove(suite.DB())
-	order := moveTaskOrder.Orders
-	request := httptest.NewRequest("GET", "/move-orders/{orderID}", nil)
+	move := testdatagen.MakeDefaultMove(suite.DB())
+	order := move.Orders
+	request := httptest.NewRequest("GET", "/orders/{orderID}", nil)
 	params := orderop.GetOrderParams{
 		HTTPRequest: request,
 		OrderID:     strfmt.UUID(order.ID.String()),
@@ -41,7 +41,7 @@ func (suite *HandlerSuite) TestGetOrderHandlerIntegration() {
 
 	suite.Assertions.IsType(&orderop.GetOrderOK{}, response)
 	suite.Equal(order.ID.String(), ordersPayload.ID.String())
-	suite.Equal(moveTaskOrder.Locator, ordersPayload.MoveCode)
+	suite.Equal(move.Locator, ordersPayload.MoveCode)
 	suite.Equal(order.ServiceMemberID.String(), ordersPayload.Customer.ID.String())
 	suite.Equal(order.NewDutyStationID.String(), ordersPayload.DestinationDutyStation.ID.String())
 	suite.NotNil(order.NewDutyStation)
@@ -67,7 +67,7 @@ func (suite *HandlerSuite) TestWeightAllowances() {
 				DependentsAuthorized: swag.Bool(false),
 			},
 		})
-		request := httptest.NewRequest("GET", "/move-orders/{orderID}", nil)
+		request := httptest.NewRequest("GET", "/orders/{orderID}", nil)
 		params := orderop.GetOrderParams{
 			HTTPRequest: request,
 			OrderID:     strfmt.UUID(order.ID.String()),
@@ -105,7 +105,7 @@ func (suite *HandlerSuite) TestWeightAllowances() {
 			},
 		})
 
-		request := httptest.NewRequest("GET", "/move-orders/{orderID}", nil)
+		request := httptest.NewRequest("GET", "/orders/{orderID}", nil)
 		params := orderop.GetOrderParams{
 			HTTPRequest: request,
 			OrderID:     strfmt.UUID(order.ID.String()),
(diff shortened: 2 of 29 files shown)


pkg/models/order.go
```diff
@@ -79,6 +79,8 @@ func (o *Order) Validate(tx *pop.Connection) (*validate.Errors, error) {
 		&CannotBeTrueIfFalse{Field1: o.SpouseHasProGear, Name1: "SpouseHasProGear", Field2: o.HasDependents, Name2: "HasDependents"},
 		&OptionalUUIDIsPresent{Field: o.EntitlementID, Name: "EntitlementID"},
 		&OptionalUUIDIsPresent{Field: o.OriginDutyStationID, Name: "OriginDutyStationID"},
+		&StringIsPresentAfterSubmission{Name: "TransportationAccountingCode", Field: o.TAC, Order: *o, DB: tx, Message: "TAC cannot be empty."},
+		&OptionalRegexMatch{Name: "TransportationAccountingCode", Field: o.TAC, Expr: `\A([A-Za-z0-9]){4}\z`, Message: "TAC must be exactly 4 alphanumeric characters."},
```
monfresh, line 83: This will allow lowercase TACs for a better user experience. When we display the TAC back in the UI, we should capitalize it.

monfresh: @amanda5iu Can you please confirm that the correct format for the TAC is 4 alphanumeric characters?

amanda5iu: That's correct!

pkg/testdatagen/scenario/e2ebasic.go
```diff
@@ -350,7 +350,7 @@ func (e e2eBasicScenario) Run(db *pop.Connection, userUploader *uploader.UserUpl
 			OrdersNumber:        models.StringPointer("12345"),
 			OrdersTypeDetail:    &typeDetail,
 			DepartmentIndicator: models.StringPointer("AIR_FORCE"),
-			TAC:                 models.StringPointer("99"),
+			TAC:                 models.StringPointer("ABC1"),
```
duncan-truss, line 353: I see you're making these alphanumeric but shouldn't they also be 4 characters to not be invalid?

pkg/models/order.go
```diff
@@ -79,6 +79,8 @@ func (o *Order) Validate(tx *pop.Connection) (*validate.Errors, error) {
 		&CannotBeTrueIfFalse{Field1: o.SpouseHasProGear, Name1: "SpouseHasProGear", Field2: o.HasDependents, Name2: "HasDependents"},
 		&OptionalUUIDIsPresent{Field: o.EntitlementID, Name: "EntitlementID"},
 		&OptionalUUIDIsPresent{Field: o.OriginDutyStationID, Name: "OriginDutyStationID"},
+		&StringIsPresentAfterSubmission{Name: "TransportationAccountingCode", Field: o.TAC, Order: *o, DB: tx, Message: "TAC cannot be empty."},
```
duncan-truss, line 82: Is this only about the TAC? I think we're also concerned with the orders number, orders type detail, and department indicator as well.

pkg/testdatagen/scenario/e2ebasic.go
duncan-truss, line 353: We do have valid TACs in the `transportation_accounting_codes` table that we should probably use unless these are meant to be invalid.

pkg/handlers/ghcapi/order_test.go
```diff
@@ -379,5 +383,134 @@ func (suite *HandlerSuite) TestUpdateOrderHandlerBadRequest() {
...
+			desc string
+			tac  string
+		}{
+			{"TestOneCharacter", "A"},
+			{"TestTwoCharacters", "AB"},
+			{"TestThreeCharacters", "ABC"},
+			{"TestGreaterThanFourChars", "ABCD1"},
+			{"TestNonAlphaNumChars", "AB-C"},
+		}
+		for _, invalidCase := range invalidCases {
```
duncan-truss, line 494: Good 'ole Go table tests 👍

pkg/testdatagen/scenario/e2ebasic.go
monfresh: From what I understand, any 4-character alphanumeric string is valid. I wasn't aware of a specific set of valid codes beyond the 4-character alphanumeric format. How does a code get added to the `transportation_accounting_codes` table? I just noticed that we have a `TacValidationHandler` that checks to see if the code exists in the table. I don't know when that gets triggered. I will ask Pamplemoose. Looks like they added this table a month ago.

duncan-truss: In the edit orders form they call the isValid endpoint when the text field changes to display a warning if the TAC is not in the database.  It won't stop them from entering a TAC not in the db but it will show a warning in the UI.

pkg/models/order.go
monfresh: This story is only for the TAC.

duncan-truss: I read the Jira story and I think it gave TAC as an example but mentions the broader orders info.

duncan-truss (approved):

pkg/models/order.go
```diff
@@ -94,6 +100,48 @@ func (o *Order) ValidateUpdate(tx *pop.Connection) (*validate.Errors, error) {
...
+	DB    *pop.Connection
+}
+
+// IsValid adds an error if the field is blank
+func (v *StringIsPresentAfterSubmission) IsValid(errors *validate.Errors) {
+	order := v.Order
+
+	if len(order.Moves) <= 0 || order.Moves[0].Status == MoveStatusDRAFT {
+		return
+	}
```
monfresh, line 117: Ideally, we wouldn't have to check the Move's status, and instead, we should update the Order status after the customer submits their move, and check for it here without having to deal with associations, which is causing some issues with Pop.

The only statuses we are using for Order are Draft and Canceled. I'm not sure why that is, so for now I decided to stick with the existing behavior. If it turns out there's no special reason we aren't updating the Order status after move submission, then we should make that change.

pkg/models/move.go
```diff
@@ -645,7 +645,9 @@ func SaveMoveDependencies(db *pop.Connection, move *Move) (*validate.Errors, err
 			}
 		}
 
-		if verrs, err := db.ValidateAndSave(&move.Orders); verrs.HasAny() || err != nil {
+		order := &move.Orders
+		db.Load(&order, "Moves")
```
monfresh, line 649: Before, I was loading the Moves association from within the new Order validations I added in this PR. However, this was causing an obscure `driver: bad connection` from Pop. So now it means that before we validate Orders outside of the model file, we always have to remember to first load the Moves association, which is not intuitive.

The only reason we have to load the Moves association is because that's how we can tell whether the Order is still in draft or if it has already been submitted. If we made use of the Status fields and changed the Order status to Submitted, we wouldn't have to deal with associations in the validation. We need to investigate why we aren't making use of the Submitted status, and if there's no objection to setting it, let's do it and simplify this code!

pkg/handlers/ghcapi/move_task_order.go
```diff
@@ -77,7 +78,8 @@ func (h UpdateMoveTaskOrderStatusHandlerFunc) Handle(params movetaskorderops.Upd
 		case services.NotFoundError:
 			return movetaskorderops.NewUpdateMoveTaskOrderStatusNotFound()
 		case services.InvalidInputError:
-			return movetaskorderops.NewUpdateMoveTaskOrderStatusBadRequest()
+			payload := payloadForValidationError("Unable to complete request", err.Error(), h.GetTraceID(), validate.NewErrors())
+			return movetaskorderops.NewUpdateMoveTaskOrderStatusUnprocessableEntity().WithPayload(payload)
```
monfresh, line 82: For validation errors, our convention is to return a 422 (Unprocessable Entity), not a 400 (Bad Request)

amanda5iu (approved):
I'm getting a `Required` error when there's no SAC; it should be optional since not all moves will have this code. Approving because this isn't technically in the story's scope, but wanted to note for future work.

monfresh:
The SAC showing up as required is a pre-existing issue. I can reproduce it in staging.

PR #6201, opened 2021-03-23
[MB-5307] Formik Migration: DOD Info

suzubara:
## Description

This PR completes the Formik & Wizard migration for the DOD Info profile step.
- Created a new `DodInfoForm` component that replicates the existing DOD info step, with added tests and Storybook page
  - There is a minor difference from the current functionality of the DOD ID number input. Currently the input does not accept any values that are non-numeric, which means if a user types in non-numeric characters, nothing happens. This can be a confusing UX that might feel broken. The new functionality is that the input will accept up to 10 of any type of character, but if the value is not 10 digits an error message will display. See the screenshot below and [related Slack thread](https://ustcdp3.slack.com/archives/CNTDW9Q1Z/p1616509988028800).
- Created a new `DodInfo` page component that renders the form, connects to Redux, and passes along initial values, the submit handler, and routing functions.
- Delete the old `DodInfo` component and use the new one in `getWorkflowRoutes`
- I also turned off the `react/jsx-props-no-spreading` eslint rule for stories & tests files

## Reviewer Notes

There should be no apparent changes to the user. Verify you're able to complete the DOD Info step of profile creation as before.

You can also review the new DOD Info form in Storybook at **Customer Components / Forms / DOD Info Form**.

Design should review the DOD Info form in Storybook & the DOD Info profile step at various screen sizes to verify it matches the design.

## Setup

To run the app locally:
```sh
make server_run
make client_run
```

To view Storybook locally:
```
yarn storybook
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5307) for this change

## Screenshots

DOD Info page:
image

image

DOD ID validation error:
image

image

With a server error:
image

image


suzubara:
@transcom/truss-design The screenshots show an extra margin above the Branch label, that should be fixed in the code

src/components/Customer/DodInfoForm/DodInfoForm.jsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
```
YanZ777, line 1: Should this file be named index.jsx if it is in its own, similarly named folder?

suzubara: We don't have an established pattern for this, it feels like a toss up to me. It might be good to (separately) determine a standard naming convention for component directory/file structure.

YanZ777: I wasn't sure either and I have no particularly strong feelings either way.  
I do think files are a bit easier to find when they're not all named index.jsx. 😅
Shall we continue this discussion elsewhere, maybe in frontend checkin tomorrow?

Shkeating (approved):
image

src/pages/MyMove/Profile/DodInfo.jsx
```diff
@@ -0,0 +1,88 @@
...
+export const DodInfo = ({ updateServiceMember, serviceMember, push }) => {
+  const [serverError, setServerError] = useState(null);
+
+  const initialValues = {
+    affiliation: serviceMember?.affiliation,
+    edipi: serviceMember?.edipi,
+    rank: serviceMember?.rank,
+  };
+
+  const handleBack = () => {
```
kimallen, line 25: My understanding is that we want to save any changes `onBack`, which I believe is the current behavior.  Is there new guidance from design on this?

kimallen: Or maybe that's for the pages after this first page of the wizard?

src/types/customerShapes.js
```diff
@@ -3,6 +3,13 @@ import { arrayOf, bool, func, string, shape, object, number } from 'prop-types';
 import { AddressShape } from 'types/address';
 import { DutyStationShape } from 'types/dutyStation';
 
+export const ServiceMemberShape = shape({
```
YanZ777, line 6: Does this need to be added to the export default at the bottom of the file?

src/components/Customer/DodInfoForm/DodInfoForm.test.jsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
```
YanZ777, line 1: I'm seeing this warning when I run these tests:
Warning: An update to Formik inside a test was not wrapped in act(...).

src/pages/MyMove/Profile/DodInfo.jsx
suzubara: @kimallen OH yes good call, I forgot about this! I'm asking @transcom/truss-design for input

src/types/customerShapes.js
suzubara: nope, the type files are all using named exports instead of defaults

src/components/Customer/DodInfoForm/DodInfoForm.test.jsx
suzubara: thanks for the catch! this should be fixed now, by making the test callback an `async` function. I believe this is because the form uses Formik's `validateOnMount` prop, which results in a state change on render. Adding `async` and `waitFor` to the assertions seems to resolve this warning.

src/pages/MyMove/Profile/DodInfo.jsx
suzubara: confirmed that we want to preserve the save on Back click behavior, I'll modify this now!

src/scenes/MyMove/getWorkflowRoutes.jsx
```diff
@@ -95,7 +95,7 @@ const pages = {
   [customerRoutes.DOD_INFO_PATH]: {
     isInFlow: myFirstRodeo,
     isComplete: ({ sm }) => sm.is_profile_complete || every([sm.rank, sm.edipi, sm.affiliation]),
-    render: (key, pages) => ({ match }) => <DodInfo pages={pages} pageKey={key} match={match} />,
+    render: (key, pages) => ({ history }) => <DodInfo push={history.push} />,
```
kimallen, line 98: Do we still need `key` and `pages` here?

suzubara: nope! Since `WizardPage` is removed, they are no longer used and can be removed from the component

src/components/Customer/DodInfoForm/DodInfoForm.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+    onSubmit: jest.fn(),
+    onBack: jest.fn(),
+    initialValues: { affiliation: '', edipi: '', rank: '' },
+  };
+
+  it('renders the form inputs', async () => {
+    const { getByLabelText } = render(<DodInfoForm {...testProps} />);
+
+    await waitFor(() => {
+      expect(getByLabelText('Branch of service')).toBeInstanceOf(HTMLSelectElement);
```
kimallen, line 18: When is it preferable to use `data-testid` to find the element?  Would this be a case for it since the label could change?

kimallen: Ooh, I and I just learned that there are existing queries that look for `data-testid`: `getByTestId`, `queryByTestId` etc.

suzubara: yes! so Testing Library recommends using queries as closely as how a user or device would be interacting with the page (see [docs about choosing queries](https://github.com/testing-library/testing-library-docs/blob/master/docs/queries/about.mdx#priority)) as possible. even though the label can change, in this case we would want to make sure to change the test with that, since the label text is probably how a user would identify a form input. They recommend testId as kind of a last resort when none of the other queries will work.

kimallen (approved):
Looks good!  Nice work figuring out the back button issue!

src/pages/MyMove/Profile/DodInfo.jsx
```diff
@@ -0,0 +1,98 @@
...
+            <Alert type="error" heading="An error occurred">
+              {serverError}
+            </Alert>
+          </Grid>
+        </Grid>
+      )}
+
+      <Grid row>
+        <Grid col desktop={{ col: 8, offset: 2 }}>
+          <DodInfoForm initialValues={initialValues} onSubmit={handleSubmit} onNext={handleNext} onBack={handleBack} />
```
felipe-lee, line 77: Does this still need the `onNext={handleNext} onBack={handleBack}` bits? Those are now being called inside the `handleSubmit` now right?

kimallen, line 77: Do onNext and onBack need to be passed to the `DodInfoForm`?

src/components/Customer/DodInfoForm/DodInfoForm.test.jsx
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
+import { render, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import DodInfoForm from './DodInfoForm';
+
+describe('DodInfoForm component', () => {
+  const testProps = {
+    onSubmit: jest.fn().mockImplementation(() => Promise.resolve()),
+    onNext: jest.fn(),
```
kimallen, line 10: I think these `onNext` and `onBack` can be removed too

suzubara:
@kimallen @felipe-lee thanks for catching those!! I have removed them

PR #6218, opened 2021-03-24
[MB-7053] Add domestic shorthaul support to service items calculator

hannaliebl:
## Description

Add pricing details for Domestic shorthaul.

## Reviewer Notes

I'm fairly sure `PriceRateOrFactor` and `EscalationCompounded` params aren't implemented yet, so the baseline shorthaul price and price escalation factors are blank for now.

## Setup

`make db_dev_e2e_populate`
`make server_run`
`make client_run`

TIO role -> PARAMS move locator.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7053) for this change

## Screenshots

Screen Shot 2021-03-24 at 7 36 29 PM

### DIFF
--- a/src/components/Office/ServiceItemCalculations/helpers.js
+++ b/src/components/Office/ServiceItemCalculations/helpers.js
@@ -58,7 +58,7 @@ const mileageZip5 = (params) => {
   const value = getParamValue(SERVICE_ITEM_PARAM_KEYS.DistanceZip5, params);
   const label = SERVICE_ITEM_CALCULATION_LABELS.Mileage;
   const detail = `${SERVICE_ITEM_CALCULATION_LABELS[SERVICE_ITEM_PARAM_KEYS.ZipPickupAddress]} ${getParamValue(
-    SERVICE_ITEM_PARAM_KEYS.ZipPickupAddress, // take the zip 3
+    SERVICE_ITEM_PARAM_KEYS.ZipPickupAddress,
     params,
   )} to ${SERVICE_ITEM_CALCULATION_LABELS[SERVICE_ITEM_PARAM_KEYS.ZipDestAddress]} ${getParamValue(
     SERVICE_ITEM_PARAM_KEYS.ZipDestAddress,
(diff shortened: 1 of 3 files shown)


src/components/Office/ServiceItemCalculations/helpers.js
```diff
@@ -54,6 +54,20 @@ const mileage = (params) => {
   return calculation(value, label, detail);
 };
 
+const mileageZip5 = (params) => {
```
hannaliebl, line 57: I can also see a world where the mileage function just handles both cases. Thoughts?

src/components/Office/ServiceItemCalculations/helpers.js
```diff
@@ -72,6 +86,24 @@ const baselineLinehaulPrice = (params) => {
   return calculation(value, label, detail1, detail2, detail3);
 };
 
+const baselineShorthaulPrice = (params) => {
```
hannaliebl, line 89: This is essentially the same as the function above it, and I could see an argument for just having one...uh, "haul" function to handle both short and linehaul? For now I kept them separate, also to hopefully ease potential merge conflicts, but that might be premature, and I'm happy to just make one that accepts some parameters for short or line hauls.

donaldthai (approved):
Looks good to me. I think we just need to add the test case in the `helper.test.js` file.

donaldthai:
Let's add a storybook thing too for this service item. I did it this way: https://github.com/transcom/mymove/pull/6220/files#diff-fe115b2dcdec206cdac17f841c58e17b8366dddfb04dbd1a2bbedfcdaf787f03R32

hannaliebl:
> Looks good to me. I think we just need to add the test case in the `helper.test.js` file.

I just realized JP added it already 🎉

src/components/Office/ServiceItemCalculations/helpers.js
jenniferlynparsons: I can see that, too. They're really not very different. We can make that a refinement for next week.

jenniferlynparsons (approved):
This looks good. Merging in the main base pricing branch should resolve the failing tests and then this is good to go. Keep your ZIP5 function, I removed it from mine. 😄

PR #6221, opened 2021-03-25
MB-5317 refactor sm name page to use formik

YanZ777:
## Description

- Created a new component `NameForm` that uses Formik. Added it into Storybook and added tests for it. 
- The `NameForm` component will be replacing the existing Name step.
- Moved `Name.jsx` into `pages/MyMove/Profile/`
- - Ripped out a lot of the plumbing inside of `Name.jsx` so that it would use the `NameForm` and handle the routing. 
- Update `getWorkflowRoutes` to use the updated `Name.jsx`
- Update Name.test.jsx to verify that the routing is correct

## Reviewer Notes

Is it ok for optional to appear on the right rather than the left?

## Setup

```sh
make client_run
make server_run
make storybook
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5317) for this change

## Screenshots

image
image


src/components/Customer/NameForm/NameForm.stories.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const DefaultState = (argTypes) => (
+  <NameForm initialValues={{}} onSubmit={argTypes.onSubmit} onBack={argTypes.onBack} />
+);
+
+export const WithInitialValues = (argTypes) => (
+  <NameForm
+    initialValues={{
+      first_name: 'Leo',
+      middle_name: 'Star',
```
kimallen, line 22: Nice!

src/pages/MyMove/Profile/Name.test.jsx
```diff
@@ -69,6 +185,8 @@ describe('requireCustomerState Name', () => {
             rank: 'test rank',
             edipi: '1234567890',
             affiliation: 'ARMY',
+            first_name: 'Tester',
```
kimallen, line 188: Adding these names would make it the "NAME COMPLETE" state, and looks like we want the current state to be DOD INFO COMPLETE

src/pages/MyMove/Profile/Name.test.jsx
```diff
@@ -38,6 +152,8 @@ describe('requireCustomerState Name', () => {
         serviceMembers: {
           testServiceMemberId: {
             id: 'testServiceMemberId',
+            first_name: 'Tester',
```
kimallen, line 155: This is still DOD INFO COMPLETE state with this added since the dod info is missing, but I don't think they would ever have the name and not the dod info.  I think you can take this out.

src/scenes/MyMove/getWorkflowRoutes.jsx
```diff
@@ -100,7 +100,7 @@ const pages = {
   [customerRoutes.NAME_PATH]: {
     isInFlow: myFirstRodeo,
     isComplete: ({ sm }) => sm.is_profile_complete || every([sm.first_name, sm.last_name]),
-    render: (key, pages) => ({ match }) => <SMName pages={pages} pageKey={key} match={match} />,
+    render: (key, pages) => ({ history }) => <SMName push={history.push} />,
```
kimallen, line 103: You could take those `key` and `pages` args out here, I believe.

kimallen (commented):
Works as expected! I had just a couple nitpiks.  Not sure why the test `does not redirect if the current state equals the "DOD INFO COMPLETE" state` is not failing with your change (I'm assuming you added it because it _was_ failing)

src/scenes/MyMove/getWorkflowRoutes.jsx
YanZ777: > Works as expected! I had just a couple nitpiks. Not sure why the test `does not redirect if the current state equals the "DOD INFO COMPLETE" state` is not failing with your change (I'm assuming you added it because it _was_ failing)

That's kind of surprising to me as well that those tests didn't fail. I actually added them in because I misinterpreted the tests and was getting prop marked as required, but its value was undefined errors.

src/components/Customer/NameForm/NameForm.jsx
```diff
@@ -0,0 +1,72 @@
...
+        const handleNext = (e) => {
+          setFieldValue('nextPage', 'next');
+          handleSubmit(e);
+        };
+
+        return (
+          <Form className={formStyles.form}>
+            <h1>Name</h1>
+            <SectionWrapper className={formStyles.formSection}>
+              <TextField label="First name" name="first_name" id="firstName" required inputMode="text" />
```
suzubara, line 37: `inputMode="text"` is the default so no need to add it

src/components/Customer/NameForm/NameForm.stories.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+
+import NameForm from './NameForm';
+
+export default {
+  title: 'Customer Components / Forms / Name Form',
+  component: NameForm,
+  argTypes: {
+    onSubmit: { action: 'submit form' },
+    onBack: { action: 'go back' },
```
suzubara, line 10: I think `onBack` can be deleted from here & props

src/types/customerShapes.js
```diff
@@ -3,6 +3,14 @@ import { arrayOf, bool, func, string, shape, object, number } from 'prop-types';
...
 import { DutyStationShape } from 'types/dutyStation';
 
+export const ServiceMemberShape = shape({
+  id: string.isRequired,
+  first_name: string.isRequired,
+  middle_name: string,
+  last_name: string.isRequired,
+  suffix: string,
+});
+
```
suzubara, line 13: Since I merged the DOD Info branch, this will need to get merged in with that, and I don't think `first_name/last_name` can be required because they won't always exist (`id` is the only property that will exist as soon as the SM is created)

suzubara (approved):
there's a couple conflicts to resolve since I merged the DOD branch but this looks great!!

suzubara:
@YanZ777 I wasn't sure what you meant by this question:
> Is it ok for optional to appear on the left rather than the right?

the "Optional" label seems to be on the right? Also, I think the text should be uppercase with no parentheses according to the designs:
```
- (optional)
+ Optional
```

YanZ777:
> @YanZ777 I wasn't sure what you meant by this question:
> 
> > Is it ok for optional to appear on the left rather than the right?
> 
> the "Optional" label seems to be on the right? Also, I think the text should be uppercase with no parentheses according to the designs:
> 
> ```
> - (optional)
> + Optional
> ```

Oops, I got my directions mixed up.
I'll make those changes you noted!

src/components/Customer/NameForm/NameForm.stories.jsx
YanZ777: Aren't we adding back in the onBack since we're not doing the save on back thing though?

src/components/Customer/NameForm/NameForm.jsx
```diff
@@ -0,0 +1,72 @@
...
+              <TextField
+                label="Middle name"
+                name="middle_name"
+                id="middleName"
+                inputMode="text"
+                labelHint="(optional)"
```
Shkeating, line 43: change to `Optional` which is how this is usually treated 👍

src/components/Customer/NameForm/NameForm.jsx
```diff
@@ -0,0 +1,72 @@
...
+              />
+              <TextField label="Last name" name="last_name" id="lastName" required inputMode="text" />
+              <TextField label="Suffix" name="suffix" id="suffix" inputMode="text" labelHint="(optional)" />
```
Shkeating, line 46: same here: `Optional`, no parentheses.

Shkeating (changes requested):
small change, but other than that it looks good!

Shkeating (approved):
thumbs up

src/components/Customer/NameForm/NameForm.stories.jsx
suzubara: yes now we are 😅

PR #6223, opened 2021-03-25
MB-6665 migration model for saving ICN

gidjin:
## Description

The response EDI files (997, 824, & 810) use the Interchange Control Number (ICN) we initially sent in the original EDI 858. This number is used to correlate the response file to the original invoice order. This PR saves the generated ICN to a new table so that it can be referenced later when processing the response files. 

## Reviewer Notes

There was discussion of just adding a field to the payment request model, but this option leaves us with a little more flexibility if we need to change the relationship. For example if we ever put more than one payment request in a single 858, or if we have to send the same payment request again in a new 858. The flexibility seemed worth while and the difference was minimal to the implementation.

See [this slack thread](https://ustcdp3.slack.com/archives/CP6F568DC/p1616770353096400)

## Setup

You will need to create a payment request and generate it's associated EDI 858.

### Local Dev

Setup the server and start it
```sh
make db_dev_reset db_dev_migrate db_dev_e2e_populate
make server_run
make office_client_run
```

Create a payment request using the following `payload.json` file
```json
{
  "body": {
    "isFinal": false,
    "serviceItems": [
      {
      },
      {
      },
      {
      }
    ]
  }
}
```

And this command to create the payment request
```sh
prime-api-client --insecure create-payment-request --filename payload.json | jq '.id, .paymentRequestNumber'

"fda063aa-7e0e-4c45-9fa5-ea24de548048"
"5768-1974-1"
```

Approve the payment request in the http://officelocal:3000/ ui. Look for Move Code `RDY4PY`

Generate the EDI using the payment request number from the previous command
```sh
rm -f bin/generate-payment-request-edi
make bin/generate-payment-request-edi
bin/generate-payment-request-edi --payment-request-number 5768-1974-1

ISA*00*0084182369*00*0000000000*ZZ*MILMOVE        *12*8004171844     *210325*0604*U*00401*763222484*0*T*|
GS*SI*MILMOVE*8004171844*20210325*0604*763222484*X*004010
...
```

Verify that the ICN, which can be found near the end of the `ISA` segment eg `763222484`, was saved with the payment request id in the table as below:
```sql
select * from payment_request_to_interchange_control_numbers where payment_request_id = 'fda063aa-7e0e-4c45-9fa5-ea24de548048';
```
```sh
                  id                  |          payment_request_id          | interchange_control_number 
--------------------------------------+--------------------------------------+----------------------------
 39338278-812b-47ef-930c-de3a7b1ce10a | fda063aa-7e0e-4c45-9fa5-ea24de548048 |                  763222484
```

### Acceptance

In staging you can trigger the process payment requests to ensure that it still functions, however there is no way to look in the database to see if the record was saved.

To trigger the code that is changed on staging an EDI needs to be sent to Syncada.

Create or find a move that is ready for TIO to approve a payment request. You can use `prime-api-demo <move id>` to help get a new move ready.

Mark other edis as sent using this `payload.json`
```json
{
  "body": {
    "sendToSyncada": false
  }
}
```

and this command
```sh
prime-api-client --cac --hostname api.stg.move.mil --port 443 support-reviewed-payment-requests --filename payload.json | jq .
```

Approve the payment request you prepared earlier in the office ui. Then run the following payload to trigger the code
Mark other edis as sent using this `payload.json`
```json
{
  "body": {
    "sendToSyncada": true
  }
}
```

and this command
```sh
prime-api-client --cac --hostname api.stg.move.mil --port 443 support-reviewed-payment-requests --filename payload.json | jq .
```

Should get a similar response like below
```json
[
  {
    "eTag": "MjAyMS0wMy0yNVQwNToxOTowNi41ODkzMjda",
    "id": "fda063aa-7e0e-4c45-9fa5-ea24de548048",
    "isFinal": false,
    "paymentRequestNumber": "5768-1974-1",
    "status": "SENT_TO_GEX"
  }
]
```

* [X] The requirements listed in [Querying the Database Safely](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#querying-the-database-safely) have been satisfied.
  * [X] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://github.com/transcom/mymove/wiki/migrate-the-database#zero-downtime-migrations))
  * [X] Have been communicated to #g-database

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6665) for this change
* [this slack thread](https://ustcdp3.slack.com/archives/CP6F568DC/p1616770353096400)

### DIFF
--- a/migrations/app/schema/20210323204715_add_payment_request_to_interchange_control_number.up.sql
+++ b/migrations/app/schema/20210323204715_add_payment_request_to_interchange_control_number.up.sql
@@ -1,0 +1,11 @@
+CREATE TABLE payment_request_to_interchange_control_numbers
+(
+    id uuid primary key,
+    payment_request_id uuid NOT NULL,
+    interchange_control_number int NOT NULL,
+    constraint payment_request_id_interchange_control_number_unique_key unique (payment_request_id, interchange_control_number)
+)
+
+COMMENT ON COLUMN payment_request_to_interchange_control_numbers.id is 'The id of this record';
+COMMENT ON COLUMN payment_request_to_interchange_control_numbers.payment_request_id is 'The id of the associated payment request';
+COMMENT ON COLUMN payment_request_to_interchange_control_numbers.interchange_control_number is 'The interchange control number generated in the out going EDI 858 invoice';
--- a/migrations/app/schema/20210323204715_add_payment_request_to_interchange_control_number.up.sql
+++ b/migrations/app/schema/20210323204715_add_payment_request_to_interchange_control_number.up.sql
@@ -1,9 +1,10 @@
 CREATE TABLE payment_request_to_interchange_control_numbers
 (
     id uuid primary key,
-    payment_request_id uuid NOT NULL,
+    payment_request_id uuid NOT NULL
+        constraint payment_request_to_interchange_control_numbers_payment_request_id_fkey references payment_requests,
     interchange_control_number int NOT NULL,
-    constraint payment_request_id_interchange_control_number_unique_key unique (payment_request_id, interchange_control_number)
+    constraint payment_request_id_interchange_control_number_unique_key unique (interchange_control_number, payment_request_id)
 );
 
 COMMENT ON COLUMN payment_request_to_interchange_control_numbers.id IS 'The id of this record';
--- a/migrations/app/schema/20210323204715_add_payment_request_to_interchange_control_number.up.sql
+++ b/migrations/app/schema/20210323204715_add_payment_request_to_interchange_control_number.up.sql
@@ -4,8 +4,8 @@ CREATE TABLE payment_request_to_interchange_control_numbers
     payment_request_id uuid NOT NULL,
     interchange_control_number int NOT NULL,
     constraint payment_request_id_interchange_control_number_unique_key unique (payment_request_id, interchange_control_number)
-)
+);
 
-COMMENT ON COLUMN payment_request_to_interchange_control_numbers.id is 'The id of this record';
-COMMENT ON COLUMN payment_request_to_interchange_control_numbers.payment_request_id is 'The id of the associated payment request';
-COMMENT ON COLUMN payment_request_to_interchange_control_numbers.interchange_control_number is 'The interchange control number generated in the out going EDI 858 invoice';
+COMMENT ON COLUMN payment_request_to_interchange_control_numbers.id IS 'The id of this record';
+COMMENT ON COLUMN payment_request_to_interchange_control_numbers.payment_request_id IS 'The id of the associated payment request';
+COMMENT ON COLUMN payment_request_to_interchange_control_numbers.interchange_control_number IS 'The interchange control number generated in the out going EDI 858 invoice';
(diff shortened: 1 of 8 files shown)


mboren (approved):
LGTM and works locally.

reggieriser (approved):
Looked through the code and it seems straightforward and complete.  Ran it locally per instructions and everything checked out.  LGTM!  🚢

reggieriser (commented):
One thing I just thought of ... do we need an index on the FK?

reggieriser (changes requested):
Actually, it doesn't look like we have a foreign key in the schema here.  We should add one before this lands.

gidjin:
🤔  I think an index would probably make sense, I will add one

gidjin:
Added the FK constraint and corrected the unique constraint since we will be looking up by ICN

migrations/app/schema/20210323204715_add_payment_request_to_interchange_control_number.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE payment_request_to_interchange_control_numbers
+(
+    id uuid primary key,
+    payment_request_id uuid NOT NULL
+        constraint payment_request_to_interchange_control_numbers_payment_request_id_fkey references payment_requests,
+    interchange_control_number int NOT NULL,
+    constraint payment_request_id_interchange_control_number_unique_key unique (interchange_control_number, payment_request_id)
```
reggieriser, line 7: Nit: The name of the constraint could be reversed too.

migrations/app/schema/20210323204715_add_payment_request_to_interchange_control_number.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE payment_request_to_interchange_control_numbers
+(
+    id uuid primary key,
+    payment_request_id uuid NOT NULL
+        constraint payment_request_to_interchange_control_numbers_payment_request_id_fkey references payment_requests,
```
reggieriser, line 5: Nit: It looks like postgres silently chops off the long name here.  When I introspect the database, I see `payment_request_to_interchange_control_numbers_payment_request_`.  Maybe we should shorten the name so what we see in the migration matches the reality in the DB.

reggieriser (approved):
Thanks for the changes.  I noted a couple of small things, but don't consider them blockers, so I'm going ahead and approving.  Feel free to adjust those if you want -- no need for a re-review.

migrations/app/schema/20210323204715_add_payment_request_to_interchange_control_number.up.sql
gidjin: I abbreviated `interchange_control_numbers` to `icns` and added the abbreviation to the comment. Now it fits in the db.

migrations/app/schema/20210323204715_add_payment_request_to_interchange_control_number.up.sql
gidjin: good catch fixed!

PR #6227, opened 2021-03-25
[MB-6906] 997 EDI Response Processing

pearl-truss:
## Description

This PR creates a new service that validates an incoming EDI string and parses it into EDI segments. It also finds and updates the associated payment request status to received_by_gex

## Reviewer Notes

`ValidateEDIHeader` is in this EDI 997 service, I think once we have a higher level handler or service it may make sense to move that there since this function can be reused by the 824.

I think John's PR should be merged in first and then I'll sync with master in this PR

## Setup

Nothing is yet calling this code

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6906) for this change
* [design doc](https://docs.google.com/document/d/1e8Zq2dkqZ1nWxJNTI5V-w1iZo0ccVTunmWEAr61F4MA/edit#heading=h.ncpvvr4dkfkz) explains more about the approach used.


gidjin:
FYI the db changes in this PR are from #6223 which this PR is dependent on.

pkg/edi/invoice/generator.go
```diff
@@ -57,6 +57,13 @@ type InvoiceHeader struct {
 	OriginPhone              *edisegment.PER
 }
 
+// InvoiceResponseHeader holds all the segments used in the headers of the 997, 824 and 810 response types
+type InvoiceResponseHeader struct {
```
gidjin, line 61: Possibly outside the scope of this PR, but I wonder if we should do this for the InvoiceResponseFooter as well since I think it is a similar set of 3 SE, GE, and IEA.

pkg/services/invoice/process_edi997.go
```diff
@@ -0,0 +1,148 @@
...
+	logger Logger) services.EDI997Processor {
+
+	return &edi997Processor{
+		db:     db,
+		logger: logger,
+	}
+}
+
+//ProcessEDI997 parses an EDI 997 response and updates the payment request status
+func (e *edi997Processor) ProcessEDI997(stringEDI997 string) (ediResponse997.EDI, error) {
```
gidjin, line 33: Will this service need to implement the `SyncadaFileProcessor` interface from #6163? If so should this wait until that is merged in too?

pkg/services/invoice/process_edi997.go
```diff
@@ -0,0 +1,148 @@
...
+				if transactionSetAcknowledgmentCode != "A" {
+					msg := fmt.Sprintf("Invalid TransactionSetAcknowledgmentCode in AK5 segment: %s for PaymentRequestID: %s", transactionSetAcknowledgmentCode, paymentRequest.ID)
+					errString += msg + "\n"
+				}
+			}
+		}
+	}
+	if errString != "" {
+		e.logger.Error(errString)
+		return edi997, fmt.Errorf(errString)
```
gidjin, line 92: At some point in the design we want to save the errors to the database. It might be helpful to mark where that would optimally happen with `// TODO` comments. I'm not sure if it should happen here or in each statement above that appends to `errString`.

pkg/services/invoice/process_edi997.go
```diff
@@ -0,0 +1,148 @@
...
+
+	if transactionError != nil {
+		return edi997, transactionError
+	}
+
+	return edi997, nil
+}
+
+// ValidateEDIHeader validates an EDI header for a 997 or 824
+func ValidateEDIHeader(ediHeader edi.InvoiceResponseHeader, paymentRequestID uuid.UUID) error {
```
gidjin, line 114: The validation rules for these segments are in the segment go files as I recall. Could we make use of the `github.com/go-playground/validator/v10` package like we do in [generating an 858](https://github.com/transcom/mymove/blob/23bec4557aa8809b114a22406739088f751226ca/pkg/edi/invoice/generator.go#L176) to check and report any errors to avoid duplicating the rules?

gidjin (commented):
This is looking good! I left a couple comments for feedback though.

pkg/services/invoice/process_edi997.go
pearl-truss: From my understanding of this story, this service wouldn't need to implement SyncadaFileProcessor. Because I've been assuming that whatever calls this service will pass it and EDI string and so this service wouldn't need to read files from syncada directly. If that's not what we want, I can make updates

pkg/services/invoice/process_edi997.go
pearl-truss: I think if we're saving errors to the db it might make sense to add that in each statement

pkg/edi/invoice/generator.go
pearl-truss: I can add that here

pkg/services/invoice/process_edi997.go
gidjin: I haven't had a chance to fully review their PR but looks like `SyncadaFileProcessor`'s `ProcessFile` method takes the path and the file text. But maybe @reggieriser or @mboren can confirm if that is the case. If so I'd say it should just implement it, but that can be part of either PR, which ever lands last.

pkg/services/invoice/process_edi997.go
pearl-truss: @gidjin I've been working on implementing this and one thing I'm noticing is that I think the ISA validations we have may be 858 specific, like there's a validator on ISA.InterchangeSender that expects it to equal: 'MILMOVE        ' but in the 997 examples we're getting '8004171844     '

jacquelineIO: @gidjin @pearl-truss  I'm not sure what happens if a validation fails. But we don't want to discard receiving a response, we'll need to log that there is a parse error in the db and that it needs further investigation. This might be an instance where we would want to store the malformed EDI to be looked at later.

pkg/services/invoice/process_edi997.go
mboren: Yeah if you have a function with the signature `func (e *edi997Processor) ProcessFile(path string,  stringEDI997 string) error`, then it'll integrate nicely with the stuff reggie and I were working on. I don't think you need to change your implementation at all, just the signature (or leave it as is and wrap it).

jacquelineIO (approved):
lgtm

PR #6228, opened 2021-03-25
Mb 5319 fl update address fields to accept field validators

felipe-lee:
## Description

This updates the `AddressFields` component to optionally accept validators for each field. It's needed to enable the rest of the work for the linked story.

## Reviewer Notes

Is there more I should do on the tests?

## Setup

1. Run storybook

    ```sh
    make storybook
    ```

2. Open [the sample residential address with validators](http://localhost:6006/?path=/story/components-fieldsets-addressfields--current-residential-address-with-custom-validators)
3. You can test putting any zip code that isn't 99999. It will show an error message.
    1. You can also test with the city by putting "Nowhere"

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5319) for this change


src/components/form/AddressFields/AddressFields.jsx
```diff
@@ -109,12 +118,20 @@ AddressFields.propTypes = {
   className: PropTypes.string,
   name: PropTypes.string.isRequired,
   render: PropTypes.func,
+  validators: PropTypes.shape({
+    streetAddress1: PropTypes.func,
+    streetAddress2: PropTypes.func,
+    city: PropTypes.func,
+    state: PropTypes.func,
+    postalCode: PropTypes.func,
+  }),
```
felipe-lee, line 127: I wasn't sure if to make these keys have camel case names or underscore names. The underscore names would match the `initialValues` object keys.

suzubara: I think this is fine! the API keys are inconsistent right now so encouraging things to be camel case is 👍

suzubara (approved):
this looks awesome!!

Shkeating (approved):
lookin good! this is probably out of scope for this ticket, but I did want to point out that when we have two fields next to each other, and one has an error they become misaligned

PR #6231, opened 2021-03-25
Refactor contact info page to use formik

kimallen:
## Description

This converts the Contact Info page in the customer profile to use Formik instead of Redux forms.  It includes validation of telephone numbers, email, and requires a 'preferred contact method' to be checked.

## Reviewer Notes

• The telephone field `Required` validation message isn't showing as expected upon blur the way email is.  It does show up after some clicking in and out or removing the value.  This may be related to the `MaskedInputField`.  It does function as needed for the validation (which Suz thought was acceptable)

 I'm getting the this warning `Warning: An update to Formik inside a test was not wrapped in act(...).` in the `ContactInfo.test.jsx`, although I'm using `async` and `await` which looks like it's related to [this](https://ustcdp3.slack.com/archives/CTNTFJSBA/p1616693573251600), so is acceptable.

The back button functionality is complicated, and because the Formik workaround to allow saving on the back button didn't allow the user to go back if there was a blank form, this now does not save on 'Back'.  Saving can happen in future work.

BUG: I discovered an existing bug, for which I will write up a separate ticket.  If customer has an alt. phone # saved and then removes it, and clicks next, an error is thrown and it does not save or progress (staging behavior is that it moves forward silently and does not save).  This bug was uncovered due to Formik validation.

## Setup

- Login as a new customer user.  Fill the profile in until you get to the Contact info page.  Enter phone numbers.  Check that Best contact phone and personal email show 'Required' on blur when empty (or when removing the value in the case of the telephone field).
- Check that an invalid or incomplete telephone or email cannot be created and you see a validation message when invalid.
- Check that the back and next buttons do indeed save all field values when valid.
- Check that the back and next buttons do not save and do show the validation errors when clicked.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5318) for this change
* [this Yup documentation](git push --set-upstream origin kim-MB-5318-refactor-contact-info) explains more about the approach used for the preferred contact method validation.


src/components/Customer/ContactInfoForm/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+import { CustomerContactInfoFields } from 'components/form/CustomerContactInfoFields';
+import SectionWrapper from 'components/Customer/SectionWrapper';
+import { Form } from 'components/form/Form';
+import WizardNavigation from 'components/Customer/WizardNavigation/WizardNavigation';
+import formStyles from 'styles/form.module.scss';
+
+const ContactInfoForm = ({ initialValues, onSubmit }) => {
+  // Note: This method cannot use an arrow function due this particular use of 'this' (see Yup mixed.test documentation)
+  function validatePreferredContactMethod() {
+    return !!(this.parent.phone_is_preferred || this.parent.email_is_preferred); // eslint-disable-line react/no-this-in-sfc
```
kimallen, line 15: Based on the documentation for Yup, this `this` is necessary.  Not sure if there is a workaround.  This is not in our acceptable disabling list (yet?), so might need annotation.

src/components/form/CustomerContactInfoFields/index.jsx
```diff
@@ -5,50 +5,61 @@ import { Label, Fieldset } from '@trussworks/react-uswds';
...
       {render(
         <>
           <div className="grid-row grid-gap">
             <div className="mobile-lg:grid-col-7">
-              <TextField
+              <MaskedTextField
                 label="Best contact phone"
-                id={`phone_${ServiceMemberContactInfoFieldsUUID}`}
-                name={`${name}.phone`}
+                id={`telephone_${CustomerContactInfoFieldsUUID}`}
```
suzubara, line 22: Since `CustomerContactInfoFieldsUUID` is a [React ref](https://reactjs.org/docs/hooks-reference.html#useref) any references to it need to use `CustomerContactInfoFieldsUUID.current` instead. I think this is why the id & label association is not working (and probably causing test failures)

src/components/form/CustomerContactInfoFields/CustomerContactInfoFields.test.jsx
```diff
@@ -22,23 +22,21 @@ describe('ContactInfoFields component', () => {
...
+        secondary_telephone: '555-890-1234',
+        personal_email: 'test@example.com',
+        phone_is_preferred: true,
+        email_is_preferred: true,
       };
 
       const { getByLabelText } = render(
         <Formik initialValues={initialValues}>
-          <ServiceMemberContactInfoFields legend="Your contact info" name="contact" />
+          <CustomerContactInfoFields legend="Your contact info" name="contact" />
```
suzubara, line 34: Since the `name` prop was removed from this component we can remove it here also

src/pages/MyMove/Profile/ContactInfo.jsx
```diff
@@ -0,0 +1,107 @@
...
+  };
+
+  const handleBack = () => {
+    return push(customerRoutes.NAME_PATH);
+  };
+
+  const handleSubmit = (values) => {
+    const { nextPage } = values;
+
+    if (values) {
```
suzubara, line 37: We can remove this `if` block (and the `return Promise.resolve(); line) because there will always be values

src/scenes/MyMove/getWorkflowRoutes.jsx
```diff
@@ -10,7 +10,8 @@ import { no_op } from 'shared/utils';
 import { NULL_UUID, SHIPMENT_OPTIONS, CONUS_STATUS } from 'shared/constants';
 import DodInfo from 'scenes/ServiceMembers/DodInfo';
 import SMName from 'scenes/ServiceMembers/Name';
-import ContactInfo from 'scenes/ServiceMembers/ContactInfo';
+// import ContactInfo from 'scenes/ServiceMembers/ContactInfo';
```
suzubara, line 13: can we delete this line altogether?

src/components/Customer/ContactInfoForm/ContactInfoForm.stories.jsx
```diff
@@ -0,0 +1,27 @@
...
+import ContactInfoForm from './index';
+
+export default {
+  title: 'Customer Components / Forms/ Contact Info Form',
+  component: ContactInfoForm,
+};
+
+const onSubmit = () => console.log('submitted on Next');
+const onBack = () => console.log('Saved on Back');
+export const DefaultState = () => <ContactInfoForm initialValues={{}} onSubmit={() => {}} onBack={() => {}} />;
```
suzubara, line 13: Can we add an object with empty string values for the default state? Otherwise this error shows up when entering a value: 
image

src/components/Customer/ContactInfoForm/ContactInfoForm.stories.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+import ContactInfoForm from './index';
+
+export default {
+  title: 'Customer Components / Forms/ Contact Info Form',
+  component: ContactInfoForm,
+};
+
+const onSubmit = () => console.log('submitted on Next');
+const onBack = () => console.log('Saved on Back');
```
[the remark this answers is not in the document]
kimallen: oops, missed this one- will work on that now!

cosburn (approved):
Tested locally, everything looks and behaves as expected. Noting the decision was made to not save data when clicking the back button.

src/scenes/ServiceMembers/ContactInfo.jsx
```diff
@@ -1,141 +0,0 @@
...
-  };
-
-  render() {
-    const { pages, pageKey, currentServiceMember, userEmail, schema } = this.props;
-    const { errorMessage } = this.state;
-
-    // initialValues has to be null until there are values from the action since only the first values are taken
-    const initialValues = currentServiceMember ? pick(currentServiceMember, subsetOfFields) : null;
-    if (initialValues && !initialValues.personal_email) {
-      initialValues.personal_email = userEmail;
```
suzubara, line 85: Do we want to retain this functionality (where the service member email address is pre-filled from the logged in user)?

suzubara (approved):
I left minor comments, it looks like the secondary phone number bug is breaking the Cypress tests... if we delete the key from the payload if it's empty I think that will fix the issue for now. otherwise looks great!!

src/scenes/ServiceMembers/ContactInfo.jsx
kimallen: oops, I missed that- yes.  I'll add that back in.  Good catch!

PR #6240, opened 2021-03-26
MB-5319 fl refactor res address page to use formik

felipe-lee:
## Description

This PR completes the Formik & Wizard migration for the Residential Address profile step.
- Created a new `ResidentialAddressForm` component that replicates the existing Residential Address step, with added tests and Storybook examples.
- There are a few differences with current behavior:
    - Back button: Will no longer save the data when clicking the back button. [Slack thread with more info and decision re: back button](https://ustcdp3.slack.com/archives/CTNTFJSBA/p1616771085274900?thread_ts=1616736269.269800&cid=CTNTFJSBA)
    - ZIP code validation: It used to happen async on zip code field blur. Formik doesn't let you only validate a field when it blurs, it validates when any field blurs. By default it also validates when any field changes but I turned that bit off so now it only validates on blur. This shouldn't affect users too much, but does mean we are likely to hit the api endpoint more often. 
        - This is a known issue with Formik: [Validate only one field](https://github.com/formium/formik/issues/512) and [Validate by field](https://github.com/formium/formik/issues/860). The first one is where they're tracking all of it, but it talks about both `Yup` validation as well as custom validation functions. The second one focuses more on custom validation functions, but was closed in favor of the first.
- There are also a few visual changes:
    - Label changes:
        - "Street address 1" -> "Address 1"
        - "Street address 2" -> "Address 2"
        - "Optional" part of "Address 2" is now on the right instead of the left.
    - "City" now takes up a full row on its own.
    - "State" and "ZIP" now share a row in desktop view, while each now has its own row on mobile.
- Created a new `ResidentialAddress` page component that renders the form, connects to Redux, and passes along initial values, the submit handler, and routing functions.
- Delete the old `ResidentialAddress` scene component and use the new one in `getWorkflowRoutes`

## Reviewer Notes

Verify you're able to complete the Residential Address step of profile creation as before.

You can also review the new Residential Address form in Storybook at Customer Components / Forms / ResidentialAddressForm.

Design should review the Residential Address form in Storybook & the Residential Address profile step at various screen sizes to verify it matches the design.

## Setup

To run the app locally:
```sh
make server_run
make client_run
```

To view Storybook locally:
```
yarn storybook
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5319) for this change

## Screenshots

Old Residential Address page:
Screen Shot 2021-03-26 at 14 47 58
Screen Shot 2021-03-26 at 14 48 37

New Residential Address page:
Screen Shot 2021-03-26 at 14 50 12
Screen Shot 2021-03-26 at 14 49 38

With missing data:
Screen Shot 2021-03-26 at 15 00 40
Screen Shot 2021-03-26 at 15 01 14

With invalid zip:
Screen Shot 2021-03-26 at 15 01 46
Screen Shot 2021-03-26 at 15 01 57


src/pages/MyMove/Profile/ResidentialAddress.jsx
```diff
@@ -0,0 +1,119 @@
...
+  if (!value) {
+    return undefined;
+  }
+
+  let responseBody;
+  try {
+    responseBody = await ValidateZipRateData(value, 'origin');
+  } catch (e) {
+    return 'Error checking ZIP';
+  }
```
felipe-lee, line 30: I'm not sure if this is the best way to handle this. This wasn't handled before, but can be triggered if a zip + 4 is input (rather than only a 5 digit zip)

src/components/Customer/ResidentialAddressForm/ResidentialAddressForm.stories.jsx
```diff
@@ -0,0 +1,52 @@
...
+  argTypes: {
+    onSubmit: {
+      action: 'submit form',
+    },
+    onBack: {
+      action: 'go back',
+    },
+  },
+};
+
```
suzubara, line 17: These stories are missing the `formFieldsName` prop which looks like it's required by the component. TBH I don't know that this prop is really necessary (we could probably refactor `AddressFields` so it doesn't always require a `name` prop & can work for both nested & flat address forms) but for now it should probably be passed into the stories since it's required

src/components/Customer/ResidentialAddressForm/ResidentialAddressForm.test.jsx
```diff
@@ -0,0 +1,117 @@
...
+    });
+  });
+
+  it('shows an error message if trying to submit an invalid form', async () => {
+    const { getByRole, findAllByRole } = render(<ResidentialAddressForm {...testProps} />);
+    const submitBtn = getByRole('button', { name: 'Next' });
+
+    userEvent.click(submitBtn);
+
+    const alerts = await findAllByRole('alert');
```
suzubara, line 74: nice!

src/pages/MyMove/Profile/ResidentialAddress.jsx
suzubara: I think this is fine for now.. this needs a change in the API rather than in the frontend if we want to support 9 digit zips

src/pages/MyMove/Profile/ResidentialAddress.jsx
```diff
@@ -0,0 +1,121 @@
...
+
+  return responseBody.valid ? undefined : UnsupportedZipCodeErrorMsg;
+};
+
+export const ResidentialAddress = ({ serviceMember, updateServiceMember, push }) => {
+  const [serverError, setServerError] = useState(null);
+
+  const formFieldsName = 'current_residence';
+
+  // initialValues has to be null until there are values from the action since only the first values are taken
```
suzubara, line 41: what does this comment refer to?

src/pages/MyMove/Profile/ResidentialAddress.jsx
```diff
@@ -0,0 +1,121 @@
...
+  const initialValues = {
+    [formFieldsName]: get(serviceMember, 'residential_address'),
```
suzubara, line 43: We're trying to deprecate lodash from this project (which is where the `get` method comes from). This should also default to an object in the same shape as the address object but with empty string values if there are none. otherwise React will log this error:
```
index.js:1 Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://reactjs.org/link/controlled-components
```

so I would recommend changing this to:
```
const initialValues = {
  [formFieldsName]: {
    street_address_1: serviceMember.residential_address?.street_address_1 || '',
    street_address_2: serviceMember.residential_address?.street_address_2 || '',
    city: serviceMember.residential_address?.city || '',
    state: serviceMember.residential_address?.state || '',
    postal_code: serviceMember.residential_address?.postal_code || '',
  }
}
```

src/components/Customer/ResidentialAddressForm/ResidentialAddressForm.stories.jsx
```diff
@@ -0,0 +1,52 @@
...
+        state: '',
+        postal_code: '',
+      },
+    }}
+    onBack={argTypes.onBack}
+    onSubmit={argTypes.onSubmit}
+  />
+);
+
+export const WithCustomValidators = (argTypes) => (
```
suzubara, line 34: Can we add a story that shows this form being pre-filled with initial values?

suzubara (commented):
left some comments, I'm not requesting changes since I don't want this PR to be blocked by me, but I think it would be good to at least make the `initialValues` initialization change

src/components/Customer/ResidentialAddressForm/ResidentialAddressForm.stories.jsx
felipe-lee: Sounds good, I'll add it for now then.

src/pages/MyMove/Profile/ResidentialAddress.jsx
felipe-lee: TBH, I don't know. It was on the original `ResidentialAddress` component that was in scenes. I left it since I hadn't really modified that part of it much, other than how the key value is set. I can take it out though, since I think one of your other comments effectively negates this comment.

src/pages/MyMove/Profile/ResidentialAddress.jsx
felipe-lee: Sounds good, I'll set it up with the defaults and get rid of the lodash usage

lynzt (approved):
The zip field allows for 10 alpha numeric chars, is this what we want?

otherwise, LGTM 🚢

felipe-lee:
> The zip field allows for 10 alpha numeric chars, is this what we want?
> 
> otherwise, LGTM 🚢

Yes and no haha. So the API that saves this data does accept a 10 char zip (5 digit zip, dash, +4 digits). But the validator that we use async to validate the zip only accepts a 5 digit zip.

cosburn:
In the non-error states, I'm seeing that the spacing is a bit too tight between the `State` label and the field above it, and the `ZIP` label when the fields stack on mobile. Attached screenshots for reference -- the areas marked in black should be equal to the area marked in green.
Screen Shot 2021-03-29 at 1 49 33 PM
Screen Shot 2021-03-29 at 1 49 46 PM

felipe-lee:
> In the non-error states, I'm seeing that the spacing is a bit too tight between the `State` label and the field above it, and the `ZIP` label when the fields stack on mobile. Attached screenshots for reference -- the areas marked in black should be equal to the area marked in green.
> Screen Shot 2021-03-29 at 1 49 33 PM
> Screen Shot 2021-03-29 at 1 49 46 PM

Oh, nice catch. I'll look into that.

felipe-lee:
> > In the non-error states, I'm seeing that the spacing is a bit too tight between the `State` label and the field above it, and the `ZIP` label when the fields stack on mobile. Attached screenshots for reference -- the areas marked in black should be equal to the area marked in green.
> > Screen Shot 2021-03-29 at 1 49 33 PM
> > Screen Shot 2021-03-29 at 1 49 46 PM
> 
> Oh, nice catch. I'll look into that.

After looking into it for a bit, we decided it should become its own issue because it's due to pre-existing CSS and will affect multiple forms.

cosburn (approved):
Noting the spacing issue is captured in a separate story, the rest looks great to me!

lynzt:
@felipe-lee can you please add a link to the jira story where we'll deal w/ the css spacing in this PR?

felipe-lee:
Story where we'll address CSS issue: https://dp3.atlassian.net/browse/MB-7515

jj-crypto (approved):

PR #6241, opened 2021-03-26
Fixed copy/paste error in Management Services pricer test

reggieriser:
## Description

This PR just fixes a copy/paste error in a variable name that I noticed while discussing pricers with Team Roci.

## Setup

This variable only appears in a test, so we just need to ensure tests still pass fine.


mboren (approved):
lgtm

PR #6243, opened 2021-03-26
MB-5320 refactor sm backup address to use formik

felipe-lee:
## Description

This PR completes the Formik & Wizard migration for the Backup Mailing Address profile step.
- Created a new `BackupMailingAddressForm` component that replicates the existing Backup Mailing Address step, with added tests and Storybook examples.
- There is one difference with current behavior:
    - Back button: Will no longer save the data when clicking the back button. [Slack thread with more info and decision re: back button](https://ustcdp3.slack.com/archives/CTNTFJSBA/p1616771085274900?thread_ts=1616736269.269800&cid=CTNTFJSBA)
- There are also a few visual changes:
    - Label changes:
        - "Street address 1" -> "Address 1"
        - "Street address 2" -> "Address 2"
        - "Optional" part of "Address 2" is now on the right instead of the left.
    - "City" now takes up a full row on its own.
    - "State" and "ZIP" now share a row in desktop view, while each now has its own row on mobile.
- Created a new `BackupMailingAddress` page component that renders the form, connects to Redux, and passes along initial values, the submit handler, and routing functions.
- Delete the old `BackupMailingAddress` scene component and use the new one in `getWorkflowRoutes`

Currently set `mb-5319-fl-refactor-res-address-page-to-use-formik` as the base branch to merge to because I'm using a shape defined in there. Once it merges (#6240), then this should look good against `master`.

One thing to note, the `BackupMailingAddressForm` component is almost exactly the same as the `ResidentialAddressForm` component. I don't know if it's preferable to have one for each since they're different pages, or to just have a single one that takes props as needed to set the unique bits. Right now the main difference between the two is the header used on the form. The other differences are already handled by props (e.g. `validators` and `formFieldsName`)

## Reviewer Notes

Verify you're able to complete the Backup Mailing Address step of profile creation as before.

You can also review the new Backup Mailing Address form in Storybook at Customer Components / Forms / BackupMailingAddressForm.

Design should review the Backup Mailing Address form in Storybook & the Backup Mailing Address profile step at various screen sizes to verify it matches the design.

## Setup

To run the app locally:
```sh
make server_run
make client_run
```

To view Storybook locally:
```
yarn storybook
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5320) for this change

## Screenshots

Old Backup Mailing Address page: 
Screen Shot 2021-03-26 at 18 40 43
Screen Shot 2021-03-26 at 18 40 24

New Backup Mailing Address page:
Screen Shot 2021-03-26 at 18 38 51
Screen Shot 2021-03-26 at 18 39 17

With missing data:
Screen Shot 2021-03-26 at 18 39 48
Screen Shot 2021-03-26 at 18 39 59


src/components/Customer/BackupMailingAddressForm/BackupMailingAddressForm.test.jsx
```diff
@@ -0,0 +1,117 @@
...
+
+    expect(testProps.onSubmit).not.toHaveBeenCalled();
+  });
+
+  it('submits the form when its valid', async () => {
+    const { getByRole, getByLabelText } = render(<BackupMailingAddressForm {...testProps} />);
+    const submitBtn = getByRole('button', { name: 'Next' });
+
+    userEvent.type(getByLabelText('Address 1'), fakeAddress.street_address_1);
+    userEvent.type(getByLabelText('City'), fakeAddress.city);
```
kimallen, line 90: I wonder if we should pass an `Address 2` value in here as well to make sure that is working too.

kimallen (changes requested):
When I click 'Next', it is hitting the patch handler, but does not update with the values.  The payload only has the sm id in it.  It also stays on the BackupAddress page instead of progressing to the next page.

src/pages/MyMove/Profile/BackupMailingAddress.jsx
```diff
@@ -0,0 +1,106 @@
...
+  };
+
+  const handleNext = () => {
+    push(customerRoutes.BACKUP_CONTACTS_PATH);
+  };
+
+  const handleSubmit = (values) => {
+    const payload = {
+      id: serviceMember.id,
+      residential_address: values.current_residence,
```
kimallen, line 42: Looks like this payload should be `backup_mailing_address`

kimallen (commented):
I'm not sure if this is in the scope of this PR, but I think it would be helpful to use a mask for the zip code.  The user should be able to input either a 5 digit OR a 9 digit zip, I believe.  It's not clear as a user whether I can enter the `-` as part of the zip.  It would be better if the form did it automatically.  Check out how `MaskedTextField` is used in `CustomerContactInfoFields/index.jsx` (in master).  We use react-imask.  My guess is this would use pattern mask https://imask.js.org/guide.html

src/scenes/MyMove/getWorkflowRoutes.jsx
```diff
@@ -123,7 +123,7 @@ const pages = {
   [customerRoutes.BACKUP_ADDRESS_PATH]: {
     isInFlow: myFirstRodeo,
     isComplete: ({ sm }) => sm.is_profile_complete || Boolean(sm.backup_mailing_address),
-    render: (key, pages) => ({ match }) => <BackupMailingAddress pages={pages} pageKey={key} match={match} />,
+    render: (key, pages) => ({ history }) => <BackupMailingAddress push={history.push} />,
```
YanZ777, line 126: I think you should be able to take out key, pages from the parameters here since they are unused.

src/pages/MyMove/Profile/BackupMailingAddress.jsx
felipe-lee: Ah good catch. hmm now i'm wondering how the test passed. I'll check the tests and update this.

src/components/Customer/BackupMailingAddressForm/BackupMailingAddressForm.test.jsx
felipe-lee: I'll add that

src/pages/MyMove/Profile/BackupMailingAddress.jsx
felipe-lee: The way the test had been set up it would pass without actually doing what we wanted. I updated it to be correct and verified it failed before I fixed the code, then passed afterwards.

kimallen:
> > I'm not sure if this is in the scope of this PR, but I think it would be helpful to use a mask for the zip code. The user should be able to input either a 5 digit OR a 9 digit zip, I believe. It's not clear as a user whether I can enter the `-` as part of the zip. It would be better if the form did it automatically. Check out how `MaskedTextField` is used in `CustomerContactInfoFields/index.jsx` (in master). We use react-imask. My guess is this would use pattern mask https://imask.js.org/guide.html
> 
> I think this might be for a later task because the underlying component that lays out the fields is used in a few places and I know for sure Residential and Backup accept different things. Like residential can only accept 5 digit zip codes because of the async validation that is used (the save endpoint can accept both types), while backup can accept either one. So it'd be in part about potentially modifying one endpoint and making sure everywhere `AddressFields` is used can accept either type of zip. Or possibly having some prop that determines which zip to use in `AddressFields`?

Yeah, that makes a lot of sense to wait.  I didn't realize it was different for different addresses. I know there's other places I've encountered validation around 5 or 7, but I'm not sure where.

src/components/Customer/BackupMailingAddressForm/BackupMailingAddressForm.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+import { AddressFields } from 'components/form/AddressFields/AddressFields';
+import SectionWrapper from 'components/Customer/SectionWrapper';
+import WizardNavigation from 'components/Customer/WizardNavigation/WizardNavigation';
+import { Form } from 'components/form/Form';
+import formStyles from 'styles/form.module.scss';
+import { requiredAddressSchema } from 'utils/validation';
+import { ResidentialAddressShape } from 'types/address';
+
+const BackupMailingAddressForm = ({ formFieldsName, initialValues, onSubmit, onBack, validators }) => {
```
kimallen, line 14: I see `validators` being passed in as a prop, but I don't see `BackupMailingAddress` passing anything in.  I'm assuming there just aren't any validators for this form- does it need to be passed in?

kimallen (approved):
Looks good!  Nice work on your first React ticket!

I added one last comment, but it's not blocking.

src/components/Customer/BackupMailingAddressForm/BackupMailingAddressForm.jsx
felipe-lee: Ah, yeah I just passed those along because `AddressFields` can accept it, and I had done it in the `ResidentialAddressForm` component. I can take it off though, because like you said, it isn't being used, and if we decide to use it later, we can easily add it.

felipe-lee: Though I guess the stories do set it up as something that can be used 🤔  idk what's better haha.

felipe-lee: I decided to remove it. YAGNI!

Shkeating (approved):

PR #6263, opened 2021-03-30
[MB-7055] Domestic destination pricing details

hannaliebl:
## Description

Add service item calculations display to domestic destination pricing details.

## Reviewer Notes

I refactored the `originPrice` function into one that can handle both origin and destination prices in an attempt to make things a little DRYer.

## Setup

```sh
make db_dev_e2e_populate
make server_run
make client_run
```

TIO role -> PARAMS move locator.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7055) for this change

## Screenshots

Screen Shot 2021-03-30 at 2 45 42 PM


src/components/Office/ServiceItemCalculations/helpers.js
```diff
@@ -78,19 +78,19 @@ const baselineLinehaulPrice = (params) => {
...
+// There is no param representing the orgin price or destination price as available in the re_domestic_service_area_prices table
 // A param to return the service schedule is also not being created
-const originPrice = (params) => {
-  const value = getParamValue(SERVICE_ITEM_PARAM_KEYS.OriginPrice, params)
-    ? getParamValue(SERVICE_ITEM_PARAM_KEYS.OriginPrice, params)
-    : '';
-  const serviceAreaVal = getParamValue(SERVICE_ITEM_PARAM_KEYS.ServiceAreaOrigin, params)
-    ? getParamValue(SERVICE_ITEM_PARAM_KEYS.ServiceAreaOrigin, params)
-    : '';
+const originAndDestinationPrice = (params, isOrigin = true) => {
```
jenniferlynparsons, line 83: I wonder if `originAndDestinationPrice` makes it sound like there are multiple values expected. Not a blocker, but would `originOrDestinationPrice` make more semantic sense?

jenniferlynparsons (approved):
Left a comment that's not a blocker. Looks like the integration tests failed on flakiness but otherwise this looks good to go.

amanda5iu (approved):
Looks good!

src/components/Office/ServiceItemCalculations/helpers.js
```diff
@@ -78,19 +78,19 @@ const baselineLinehaulPrice = (params) => {
...
+// There is no param representing the orgin price or destination price as available in the re_domestic_service_area_prices table
 // A param to return the service schedule is also not being created
-const originPrice = (params) => {
-  const value = getParamValue(SERVICE_ITEM_PARAM_KEYS.OriginPrice, params)
-    ? getParamValue(SERVICE_ITEM_PARAM_KEYS.OriginPrice, params)
-    : '';
-  const serviceAreaVal = getParamValue(SERVICE_ITEM_PARAM_KEYS.ServiceAreaOrigin, params)
-    ? getParamValue(SERVICE_ITEM_PARAM_KEYS.ServiceAreaOrigin, params)
-    : '';
+const originOrDestinationPrice = (params, isOrigin = true) => {
```
duncan-truss, line 83: This would only be a minor difference, but I wonder if passing the keys would simplify things instead of a boolean.

src/components/Office/ServiceItemCalculations/helpers.js
```diff
@@ -78,19 +78,19 @@ const baselineLinehaulPrice = (params) => {
...
+  const valueKey = isOrigin ? SERVICE_ITEM_PARAM_KEYS.OriginPrice : SERVICE_ITEM_PARAM_KEYS.DestinationPrice;
+  const value = getParamValue(valueKey, params) ? getParamValue(valueKey, params) : '';
+  const serviceAreaKey = isOrigin ? SERVICE_ITEM_PARAM_KEYS.ServiceAreaOrigin : SERVICE_ITEM_PARAM_KEYS.ServiceAreaDest;
+  const serviceAreaVal = getParamValue(serviceAreaKey, params) ? getParamValue(serviceAreaKey, params) : '';
```
duncan-truss, line 87: Some of these ternary operations can just become:

`const value = getParamValue(valueKey, params) || '';`
and
` const serviceAreaVal = getParamValue(serviceAreaKey, params) || '';`

src/components/Office/ServiceItemCalculations/ServiceItemCalculations.stories.jsx
```diff
@@ -78,3 +78,12 @@ export const DOP = (data) => (
...
 );
+
+export const DDP = (data) => (
+  <ServiceItemCalculations
+    serviceItemParams={testParams.DomesticDestinationPrice}
+    totalAmountRequested={642}
+    itemCode="DDP"
+    tableSize={data.tableSize}
+  />
+);
```
duncan-truss, line 89: Thanks for adding this!

src/components/Office/ServiceItemCalculations/helpers.test.js
```diff
@@ -92,7 +92,28 @@ describe('makeCalculations', () => {
...
     const result = makeCalculations('DDP', 99999, testParams.DomesticDestinationPrice);
-    expect(result).toEqual([]);
+    expect(result).toEqual([
+      {
+        value: '85 cwt',
+        label: 'Billable weight (cwt)',
+        details: ['Shipment weight: 8,500 lbs', 'Estimated: 8,000 lbs'],
+      },
+      {
+        value: '',
```
duncan-truss, line 102: I think this is going to end up being the `PriceRateorFactor` service item param instead of OriginPrice or DetsinationPrice but I know it's missing.  I think we should still add it and test it will populate if it's available.

src/components/Office/ServiceItemCalculations/helpers.js
```diff
@@ -78,19 +78,19 @@ const baselineLinehaulPrice = (params) => {
...
 // A param to return the service schedule is also not being created
-const originPrice = (params) => {
-  const value = getParamValue(SERVICE_ITEM_PARAM_KEYS.OriginPrice, params)
-    ? getParamValue(SERVICE_ITEM_PARAM_KEYS.OriginPrice, params)
-    : '';
-  const serviceAreaVal = getParamValue(SERVICE_ITEM_PARAM_KEYS.ServiceAreaOrigin, params)
-    ? getParamValue(SERVICE_ITEM_PARAM_KEYS.ServiceAreaOrigin, params)
-    : '';
+const originOrDestinationPrice = (params, isOrigin = true) => {
+  const valueKey = isOrigin ? SERVICE_ITEM_PARAM_KEYS.OriginPrice : SERVICE_ITEM_PARAM_KEYS.DestinationPrice;
```
duncan-truss, line 84: Similar comment as below I think this param when it's not missing will have the key `PriceRateOrFactor`.  Have you seen otherwise?

hannaliebl: Ah, ok. I see what you mean that the google sheet only references `PriceRateOrFactor` for this and domestic origin price (and DOFSIT, which uses this as well!) I think for me the confusion came from the Zeplin sheet showing different values (origin price, destination price), but I guess that all is ultimately coming from the PriceRateOrFactor param in all those cases, yes?

duncan-truss: Yea that's exactly what I thought the first time too.  Reggie had to explain it to me in this thread that origin and destination price will come back in that missing `PriceRateOrFactor` param https://ustcdp3.slack.com/archives/CP6F568DC/p1616707661088800?thread_ts=1616619213.060600&cid=CP6F568DC

src/components/Office/ServiceItemCalculations/helpers.js
hannaliebl: I can see an argument for this, but I feel like there are already so many references to constants in different spots that to add another place for them wherever this function is called seems less clean than a boolean. I think in order for that to work, this would have to be called with both the service area keys and the label keys. Or did you mean passing something like "dest" or "origin" rather than a boolean to make it more explicit?

hannaliebl:
@duncan-truss thanks for pointing out the `PriceRateOrFactor` param issue - I've updated that function to look at those and updated the tests as well to use those params. Let me know if you think there is still some work to be done with clarity, etc.

src/components/Office/ServiceItemCalculations/helpers.js
```diff
@@ -15,6 +15,10 @@ const getParamValue = (key, params) => {
   return params?.find((param) => param?.key === key)?.value;
 };
 
+const getPriceRateOrFactor = (params) => {
```
hannaliebl, line 18: Maybe this is over-optimization...for me, though, it's easier to parse in the functions that call it. If this is too annoying because other PRs don't have this work yet, though, I can remove this.

src/components/Office/ServiceItemCalculations/helpers.js
duncan-truss: I was thinking the multiple service area and label keys as args, I'm good with the way you have it.

One drawback of the current way we're doing things is re-searching the params array.  Maybe we can just convert it into a hashmap once for each service item, then the lookups won't need to loop back through params so much.  Could be worth tackling once everything is implemented.

duncan-truss (approved):
Nice work! I think if you merge in latest from master that CI error will go away.

src/components/Office/ServiceItemCalculations/helpers.js
hannaliebl: Ah, yeah. I think doing some optimizations in this file once we are done with the tickets would be good.

src/components/Office/ServiceItemCalculations/helpers.js
```diff
@@ -266,6 +267,15 @@ const makeCalculations = (itemCode, totalAmount, params) => {
         totalAmountRequested(totalAmount),
       ];
       break;
+    // Domestic destination
+    case SERVICE_ITEM_CODES.DDP:
+      result = [
+        billableWeight(params),
+        originOrDestinationPrice(params, false),
```
monfresh, line 274: I generally treat boolean arguments as a code smell, especially if they're not named, because  you don't know what they represent unless you look up the function, and you have to remember to do that every time you need to use it.

In this case, I think a little duplication is better, and having two distinct functions to call is easier to reason about. For Origin items, it's clear you should call `originPrice`, and for destination ones, you would call `destinationPrice`.

Also, having a boolean argument is an open invitation to add some more in case more duplication is found, and the problem gets worse. Whereas if you only have separate functions, then that's a signal that a new one should be created. As Sandi Metz says, for better or for worse, the patterns you establish today will be replicated forever. We should think about what others who come after us will do when they see our code.

PR #3342, opened 2020-01-15
Clear session when attempting to load the devlocal user list

jim:
## Description

I have been tripping on this for a long time. When the session and the database get out of sync, the user gets in a bad state where they can't log out because the app thinks they are logged in, except the front end think you're logged out and so it doesn't render the "logout" links.

The solution is generally to manually clear out the cookies in your browser and start fresh. But why put ourselves through that ritual?

## Reviewer Notes

This PR destroys the current session when a user visits `/devlocal-auth/login`. Yes, this is a GET request, but this change fixes the situation described above without needing to modify any of the other auth logic in the app. Since this is purely a local development tool, I think this is OK.

## Setup (_new and improved!_)

**Checkout the master branch.**

```
$ make db_dev_e2e_populate
$ make server_run
$ make client_run
```

Go to `http://milmovelocal:3000` and create a new MilMove user. Once you see the first profile form, kill the server and:

```
$ make db_dev_e2e_populate
$ make server_run
```

Reload the browser window containing the profile form. It should show you the logged-out homepage. You should now be stuck in this loop:

can't log in ugh

**Now, check out this branch and restart the server.** You should be able to click `Local Sign In` and actually log in.

PR #6267, opened 2021-03-30
Fix flakey admin Cypress tests

monfresh:
## Description

I believe the flakey tests we have been seeing in `adminUsers.js` and
`webhookSubscriptions.js` are due to them always being together in the
same parallel run. From what I can tell, all builds are running the
exact same groups of tests in the same order and in the same parallel
runs.

It just happens that the only parallel run with 2 tests that are both
from the admin folder is Parallel Run 0, which runs both `adminUsers.js`
and `webhookSubscriptions.js` one after the other. And because we had
backend code that would redirect to the app's home page if a session
already existed, and because Cypress did not automatically clear the
session in between tests, these were the tests that failed often.

We then added calls to clear cookies before the webhook subscription
tests, but the order in which the call to clear the cookies was made
got botched during a merge conflict. In the second webhook subscription
test (Details Show Page), cookies were cleared before preparing the
admin app, which cause this test to start failing again.

To hopefully solve this once and for all, I've removed the backend
redirect because it causes more confusion than it does anything useful.
I also replaced `clearAllCookies` with `cy.logout()` and I added
calls to `logout` in other admin tests for good measure. With the 
backend change, it shouldn't be necessary to logout but I added it as 
an extra safety measure. The `clearAllCookies` command we added uses an 
undocumented and unsupported Cypress option, so I thought it's best to
stick with our custom `logout` command.

## Reviewer Notes

## Setup

```sh
echo "Code goes here"
```

## References

## Screenshots


pkg/auth/authentication/devlocal.go
```diff
@@ -53,19 +53,6 @@ func NewUserListHandler(ac Context, db *pop.Connection) UserListHandler {
...
-		// User is already authenticated, so clear out their current session and have
-		// them try again. This the issue where a developer will get stuck with a stale
-		// session and have to manually clear cookies to get back to the login page.
-		err := h.sessionManager(session).Destroy(r.Context())
-		if err != nil {
-			h.logger.Error("Could not destroy session", zap.Error(err))
-		}
-		auth.DeleteCSRFCookies(w)
-
-		http.Redirect(w, r, h.landingURL(session), http.StatusTemporaryRedirect)
```
monfresh, line 66: This redirect here is what caused the original issue that resulted in adding code to clear the session in this PR: https://github.com/transcom/mymove/pull/3342. Removing the redirect altogether prevents the original issue from happening. In development, there's no special reason to redirect if the user is already logged in. Most people don't attempt to visit `devlocal-auth/login` when they're already authenticated, and if they do, they probably want to sign in with a different user, so it's fine to let them go there even if they're already authenticated.

suzubara (approved):
this looks good to me! nice investigating 🕵️

PR #6276, opened 2021-03-31
adr for MB-7112

tinyels:
## Description

This is an adr for the state changes we need for service counseling. 
There is some markdown linting stuff I need to fix, but wanted to get 👀 on content ASAP so a team can pick this up next week.

## References
* [Create new move "status" for Services Counseling](https://dp3.atlassian.net/browse/MB-7112) 
* [Miro](https://miro.com/app/board/o9J_lNMizs4=/) with a graphical view of it.


docs/adr/0060-move-state-for-service-counseling.md
```diff
@@ -0,0 +1,24 @@
...
+**User Story:** [Create new move "status" for Services Counseling](https://dp3.atlassian.net/browse/MB-7112)
+
+_We need to define states for when moves go to service counseling and when service counseling is complete. it would be nice to minimize the impact of this to downstream systems._
+
+## Considered Alternatives
+
+- _Create two states and one new timestamp_
+  - When a customer submits a move, the submitted timestamp is always set.
+  - If the move is routed to service counseling, the move state is set to `needs_service_counseling`. Otherwise it is set to `submitted`.
+  - When the service counselor is done, the move state is set to `service_counseling_complete` and the `service_counseling_timestamp` is set to the current time.
```
monfresh, line 12: Can we rename the timestamp field to `service_counseling_completed_at`?

docs/adr/0060-move-state-for-service-counseling.md
```diff
@@ -0,0 +1,24 @@
...
+  - When a customer submits a move, the submitted timestamp is always set.
+  - If the move is routed to service counseling, the move state is set to `needs_service_counseling`. Otherwise it is set to `submitted`.
+  - When the service counselor is done, the move state is set to `submitted` and the `service_counseling_timestamp` is set to the current time.
+  - The TOO queue excludes items in `draft`, `needs_service_counseling` and shows those in `submitted` (along with other states it currently shows).
+
+## Decision Outcome
+
+- Chosen Alternative: _Create two states and one new timestamp_
+  - This more closely matches the mental model of the user and is not significantly more work than the alternative.
+  - We don't get into a confusing situation where moving to submitted updates different timestamps.
```
monfresh, line 24: This looks good overall. 

I'm not sure if this needs to be captured in this ADR, but I thought I would document some of the other things we'll need to change. For example, in order to submit a move, you have to call `move.Submit()`. Currently, this will only set the status to `Submitted` if the previous status was `Draft`, which is still valid, but it assumes that the previous Draft status was the correct one.

Now that we are introducing the `needs_service_counseling` status, we need some kind of assurance that a move that was meant for services counseling cannot have a `draft` status.  Since the routing to services counseling is supposed to happen automatically based on origin duty station, we should have a validation that verifies that when the move gets created, it should have the appropriate status. This would prevent us from creating invalid data in our tests and seeds.

This assumes that we would determine if a move needs service counseling during the initial Move creation, which happens at the same time the Order is first created. Presumably we have all the information we need at this point to determine if it needs service counseling or not. Pamplemoose might know better.

If it can't happen until the user submits their move, then we might need a new state function such as `move.SendToServicesCounseling()`, and a new service object that the `SubmitMoveHandler` would call to check if it needs service counseling, and call `move.SendToServicesCounseling()` if it does, otherwise `move.Submit()`, and the requirement that the previous status be `draft` would remain.

Similarly, the `move.Approve()` function will need to be updated to add the new `service_counseling_complete` state as a valid state from which you can transition into approval.

felipe-lee: Up until a customer submits the move, they're able to change their origin duty station, so it'll have to be checked on submission. Unless the action to change origin duty station is changed to account for this instead. I think it's less juggling though if checked on submission rather than potentially changing other things because the origin duty station changed before the move was submitted.

tinyels: Is the service counselor able to change the duty station?

docs/adr/0060-move-state-for-service-counseling.md
tinyels: That is a good idea!

docs/adr/0060-move-state-for-service-counseling.md
tinyels: I also think that we don't want to exclude the service counselor from interacting with draft moves since they may create moves for situations such as Blue Bark. I'm not sure how we are going to distinguish those drafts from moves customers have not completed. I also don't think that is something we need to decide just yet.

Ryan-Koch (approved):

PR #6277, opened 2021-03-31
[MB-7060] Domestic origin SIT pickup payment service item calculations

duncan-truss:
## Description

This PR adds the calculations to the DOPSIT payment service item in the TIO interface.  DOPSIT is one of the origin SIT service items created through the Prime API when a new DOFSIT service item is created.

The SIT Pickup Price is missing because the `PriceRateOrFactor` param is not yet provided by the pricer.  Similarly the price escalation factor and base year are also still outstanding values.

I've modified the devseed for the `PARAMS` move to update the DOPSIT service item with a departure date similar to how it might happen in the Prime API.

## Setup

```sh
make db_dev_e2e_populate
make server_run
make client_run
```

1. Login as the TIO user
2. Search for the PARAMS move locator code in the queue and select that payment request
3. Review all of the service items on the payment request
4. Find the Domestic Origin SIT Pickup service item in the HHG longhaul shipment and expand to reveal the calculations table.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7060) for this change

## Screenshots

Screen Shot 2021-03-31 at 12 41 30 PM


pkg/testdatagen/scenario/devseed.go
```diff
@@ -692,8 +692,13 @@ func createDefaultHHGMoveWithPaymentRequest(db *pop.Connection, userUploader *up
 	createHHGMoveWithPaymentRequest(db, userUploader, primeUploader, logger, affiliation, testdatagen.Assertions{})
 }
 
+// Creates a payment request with domestic longhaul and shorthaul shipments with
+// service item pricing params for displaying cost calculations
 func createHHGWithPaymentServiceItems(db *pop.Connection, userUploader *uploader.UserUploader, primeUploader *uploader.PrimeUploader, routePlanner route.Planner, logger Logger, affiliation models.ServiceMemberAffiliation, assertions testdatagen.Assertions) {
-	actualPickupDate := time.Now().Add(-24 * time.Hour)
+
+	issueDate := time.Date(testdatagen.GHCTestYear, 3, 15, 0, 0, 0, 0, time.UTC)
+	reportByDate := time.Date(testdatagen.GHCTestYear, 8, 1, 0, 0, 0, 0, time.UTC)
```
duncan-truss, line 700: I noticed in the test factory for Orders that they seemed to use these set year values so I tried to make the move and order dates line up.

pkg/testdatagen/scenario/devseed.go
```diff
@@ -705,16 +710,27 @@ func createHHGWithPaymentServiceItems(db *pop.Connection, userUploader *uploader
...
+		},
 	})
 
 	move := longhaulShipment.MoveTaskOrder
 
+	shorthaulDestinationAddress := testdatagen.MakeAddress(db, testdatagen.Assertions{
+		Address: models.Address{
+			PostalCode: "90211",
+		},
+	})
```
duncan-truss, line 725: I think a shorthaul is less than 50 miles and I also wanted to have a zips that had the first 3 digits since that's used in some of the lookups.

pkg/testdatagen/scenario/devseed.go
```diff
@@ -740,19 +756,21 @@ func createHHGWithPaymentServiceItems(db *pop.Connection, userUploader *uploader
...
+	planner.On("TransitDistance", mock.Anything, mock.Anything).Return(3, nil).Once()
+
+	// called for domestic linehaul service item
+	planner.On("Zip3TransitDistance", "94535", "94535").Return(348, nil).Once()
+
+	// called for domestic shorthaul service item
+	planner.On("Zip5TransitDistance", "90210", "90211").Return(3, nil).Once()
+
+	// called for domestic origin SIT pickup service item
+	planner.On("Zip3TransitDistance", "90210", "94535").Return(348, nil).Once()
```
duncan-truss, line 773: I think it's easier to understand how each calculation is computed if we don't mock out the planner functions for any argument.  Hopefully we can get more realistic values this way too so they don't always return the same value for different args.

pkg/testdatagen/scenario/devseed.go
```diff
@@ -790,7 +808,32 @@ func createHHGWithPaymentServiceItems(db *pop.Connection, userUploader *uploader
...
+	}
+
+	originDepartureDate := originEntryDate.Add(15 * 24 * time.Hour)
+	serviceItemDOPSIT.SITDepartureDate = &originDepartureDate
+
+	updatedDOPSIT, updateErr := serviceItemUpdator.UpdateMTOServiceItemPrime(db, &serviceItemDOPSIT, etag.GenerateEtag(serviceItemDOPSIT.UpdatedAt))
+
+	if updateErr != nil {
+		logger.Fatal("Error updating DOPSIT with departure date")
+	}
```
duncan-truss, line 832: I believe this is roughly what happens in the Prime that they later come back and update the Origin SIT with a departure date when it becomes known.

src/components/Office/ExpandableServiceItemRow/ExpandableServiceItemRow.jsx
```diff
@@ -40,6 +40,7 @@ const ExpandableServiceItemRow = ({ serviceItem, index }) => {
         data-groupid={index}
         className={tableRowClasses}
         onClick={canClickToExpandContent(canShowExpandableContent, serviceItem) ? handleExpandClick : undefined}
+        aria-expanded={isExpanded}
```
duncan-truss, line 43: Good for accessibility any time we have a toggle reveal/hide action.

pkg/testdatagen/scenario/devseed.go
```diff
@@ -740,19 +756,21 @@ func createHHGWithPaymentServiceItems(db *pop.Connection, userUploader *uploader
...
-		mock.Anything,
-		mock.Anything,
-	).Return(90210, nil)
-	planner.On("TransitDistance", mock.Anything, mock.Anything).Return(100, nil)
+
+	// called using the addresses with origin zip of 90210 and destination zip of 94535
+	planner.On("TransitDistance", mock.Anything, mock.Anything).Return(348, nil).Once()
+
+	// called using the addresses with origin zip of 90210 and destination zip of 90211
+	planner.On("TransitDistance", mock.Anything, mock.Anything).Return(3, nil).Once()
```
hannaliebl, line 764: How does this work if it's the exact same mock as the one above, but returns a different value here vs there?

duncan-truss: Adding the `Once()` call to the end of the mock allows us to return different values depending for the first call versus the second time it's called.  I think they are always called in the same order but I could be wrong.  I would add the exact args instead mock.Anything but they are Address models and I think that sometimes fails equality even if the values are the same.

pkg/testdatagen/scenario/devseed.go
monfresh: While we're making changes here, I think this is the perfect time to refactor this super long function into smaller ones, so that it's easier to see the different steps involved, and for potential reusability. See the bandwidth seed as an example:

https://github.com/transcom/mymove/blob/master/pkg/testdatagen/scenario/bandwidth.go#L31-L39

src/components/Office/ServiceItemCalculations/ServiceItemCalculations.stories.jsx
```diff
@@ -78,3 +78,12 @@ export const DOP = (data) => (
     tableSize={data.tableSize}
   />
 );
+
+export const DOPSIT = (data) => (
+  <ServiceItemCalculations
+    serviceItemParams={testParams.DomesticOriginSITDelivery}
```
monfresh, line 84: The service code DOPSIT stands for Domestic origin SIT Pickup, so either the code is wrong here or the params name `DomesticOriginSITDelivery` is wrong. Unless I'm missing something.

src/components/Office/ServiceItemCalculations/helpers.test.js
```diff
@@ -163,8 +163,34 @@ describe('makeCalculations', () => {
   });
 
   it('returns correct data for DomesticOriginSITDelivery', () => {
```
monfresh, line 165: I think this should say `DomesticOriginSITPickup`, not delivery.

src/components/Office/ServiceItemCalculations/serviceItemTestParams.js
```diff
@@ -464,25 +485,28 @@ const testParams = {
   ],
   DomesticOriginSITDelivery: [
```
monfresh, line 486: Same here. It should be `DomesticOriginSITPickup`

pkg/testdatagen/scenario/devseed.go
duncan-truss: That's true I think we can pull out the different parts with clearer naming.  I don't know if it's so much we're doing variations of the same thing as in the bandwidth example.

src/components/Office/ServiceItemCalculations/ServiceItemCalculations.stories.jsx
duncan-truss: Yep that threw me off too, I think it's because destination SIT uses Delivery but origin SIT uses Pickup.  I'll stick with pickup which matches the code and service item name.

pkg/testdatagen/scenario/devseed.go
duncan-truss: I do want to be conscious of inflight work from JP that is also adding to this devseed function.  I think she is also developing a chore to do a larger cleanup.

pkg/testdatagen/scenario/devseed.go
```diff
@@ -790,7 +808,32 @@ func createHHGWithPaymentServiceItems(db *pop.Connection, userUploader *uploader
 
 	serviceItemUpdator := mtoserviceitem.NewMTOServiceItemUpdater(queryBuilder)
 
+	var serviceItemDOFSIT models.MTOServiceItem
+	var serviceItemDOASIT models.MTOServiceItem
+	var serviceItemDOPSIT models.MTOServiceItem
```
Ryan-Koch, line 813: Not a blocking thing and I think it's just the nature of these codes, but these names only being one letter off from each other messed with me a bit when I was reading through this section. I wonder if there's a reasonable way we can create more differentiation.

Ryan-Koch (approved):

jenniferlynparsons (approved):
Looks good!

donaldthai (approved):

monfresh (approved):
Changes look good. Thanks!

amanda5iu (commented):
Can we change `Zip` to `ZIP` (all caps)? Other than that this is looking good.

amanda5iu (commented):
Also noticed the content is spilling out of the container on Safari. Not a deal-breaker since it's still legible, but maybe something to throw in the front-end backlog.

Screen Shot 2021-04-02 at 10 46 58 AM

duncan-truss:
> Also noticed the content is spilling out of the container on Safari. Not a deal-breaker since it's still legible, but maybe something to throw in the front-end backlog.
> 
> Screen Shot 2021-04-02 at 10 46 58 AM

@amanda5iu I see that behavior too when I have the page zoomed in on Firefox as well.  I don't have the time to make this responsive on this PR. I can either set it to wrap the flexbox row so the total amount appears underneath the content or I can make it overflow into a scrollable container.

Screen Shot 2021-04-02 at 12 05 31 PM
Screen Shot 2021-04-02 at 12 06 08 PM

amanda5iu:
> > Also noticed the content is spilling out of the container on Safari. Not a deal-breaker since it's still legible, but maybe something to throw in the front-end backlog.
> > Screen Shot 2021-04-02 at 10 46 58 AM
> 
> @amanda5iu I see that behavior too when I have the page zoomed in on Firefox as well. I don't have the time to make this responsive on this PR. I can either set it to wrap the flexbox row so the total amount appears underneath the content or I can make it overflow into a scrollable container.
> 
> Screen Shot 2021-04-02 at 12 05 31 PM
> 
> Screen Shot 2021-04-02 at 12 06 08 PM

If it's a low lift to implement, let's go with the scrollable option 🙂

amanda5iu (approved):
Thanks for making the front-end tweaks! This looks great 🤩

PR #6279, opened 2021-03-31
[MB-6944] process an EDI824 file

pearl-truss:
## Description

This PR creates a function to parse a string that represents an EDI 824 response type and saves the technical error descriptions to the database

## Reviewer Notes
 
- It'll be clearer to see what's new to this PR after the [migration PR](https://github.com/transcom/mymove/pull/6089) lands
- This function isn't being called anywhere yet so a good look through the code would be appreciated!
- To match the incoming changes from the [EDI metrics PR](https://github.com/transcom/mymove/compare/rr-MB-7406-edi-metrics#diff-5feb2a1250d847cb404e09869d5b019711e08bfdaf6a4e1c65156b7b38359ee7) I added an EDIType() method to the 824 processor. However, I didn't change the model type from EDI824 to EDIType824 like in the metrics PR since there a quite a few files touched by that change. If that PR lands first I will update here, otherwise that PR will need to update the model name in the 824 processor

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6944) for this change
* [design doc](https://docs.google.com/document/d/1e8Zq2dkqZ1nWxJNTI5V-w1iZo0ccVTunmWEAr61F4MA/edit#heading=h.kybzt7s2xqqh) explains more about the approach used.


pkg/services/invoice/process_edi824.go
```diff
@@ -0,0 +1,119 @@
...
+	teds := fetchTEDSegments(edi824)
+
+	if errString != "" {
+		e.logger.Error(errString)
+		return fmt.Errorf(errString)
+	}
+
+	var transactionError error
+	transactionError = e.db.Transaction(func(tx *pop.Connection) error {
+		for _, ted := range teds {
```
jacquelineIO, line 66: I like this. There is also a way to build up a string/slice and do this in one pass but can't find PR where I used this. Will update if I find it.

jacquelineIO (commented):
I need to re-review, but sending over these first comments because one does need to be looked at concerning the ICN check.

pkg/services/invoice/process_edi824.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// Find the PaymentRequestID that matches the ICN
+	icn := edi824.InterchangeControlEnvelope.ISA.InterchangeControlNumber
+
+	var prToICN models.PaymentRequestToInterchangeControlNumber
+	err = e.db.Q().
+		Where("interchange_control_number = ?", int(icn)).
+		First(&prToICN)
+	if err != nil {
+		errString += fmt.Sprintf("unable to find PaymentRequestTOInterchangeControlNumber with ICN: %s, %d", err.Error(), int(icn)) + "\n"
```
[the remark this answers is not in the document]
pearl-truss: I was thinking I didn't need to explicitly validate any of the segments since I'm calling the validator.Validate on the entire edi. But it sounds like additional validation is needed to check that it equals a mto reference id?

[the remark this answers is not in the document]
jacquelineIO: @pearl-truss  that `BGN02` segment is what you'll need to find what PR this goes with. That's the only way to tie the two together.

pkg/services/invoice.go
```diff
@@ -21,6 +21,13 @@ type GHCPaymentRequestInvoiceGenerator interface {
 	Generate(paymentRequest models.PaymentRequest, sendProductionInvoice bool) (ediinvoice.Invoice858C, error)
 }
 
+// EDI824Processor is the exported interface for generating an invoice
+//go:generate mockery -name EDI824Processor
+type EDI824Processor interface {
+	ProcessFile(path string, ediString string) error
+	EDIType() models.EDIType
+}
+
```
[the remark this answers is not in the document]
pearl-truss: There's definitely a lot of shared code between the 997 and 824 processors but there are a few specific error handling  for each and the call to validate expects different structs so I think we would still need both. But if we wanted to we could probably separate out some of the shared code into its own function. 

We talked about adding the EDIType at office hours yesterday, I'll ping @mboren and @reggieriser in case there's more context to add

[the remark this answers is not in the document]
mboren: I think you can add `EDIType()` to the `SyncadaFileProcessor` interface, and then you won't need `EDI824Processor` or `EDI997Processor` as separate interfaces.

pkg/services/invoice/process_edi824.go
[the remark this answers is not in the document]
pearl-truss: Okay we can pair on this because I thought models.PaymentRequestToICN is what allowed us to find a matching PR

pkg/services/invoice.go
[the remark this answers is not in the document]
mboren: I don't think I explained that well so i'm taking another stab at it: I think we still need the separate `EDIType` implementations for the 997 and 824 processors, but we don't need to define an `EDI824Processor` interface. The 824 processor stuff you wrote implements the `SyncadaFileProcessor` interface, so it doesn't need a separate interface to be defined.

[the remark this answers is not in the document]
jacquelineIO: @mboren yes, Reggie has an example of the 997 processing doing that in this PR https://github.com/transcom/mymove/pull/6283.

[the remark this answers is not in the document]
mboren: Thanks! That's what I was getting at. And sorry about the confusing last minute change to an interface used by multiple stories.

pkg/services/invoice/process_edi824.go
```diff
@@ -0,0 +1,135 @@
...
+		Where("payment_request_to_interchange_control_numbers.interchange_control_number = ?", int(otiGCN)).
+		First(&paymentRequest)
+	if err != nil {
+		e.logger.Error("unable to find PaymentRequest with GCN", zap.Error(err))
+		return fmt.Errorf("unable to find PaymentRequest with GCN: %s, %d", err.Error(), int(otiGCN))
+	}
+
+	var move models.Move
+	err = e.db.Q().
+		Find(&move, paymentRequest.MoveTaskOrderID)
```
pearl-truss, line 59: Am I doing something wrong here? When I tried testing a case of a move that didn't have a corresponding paymentRequest.MoveTaskOrderID, I found that instead of returning error this was creating an empty move record

jacquelineIO: @pearl-truss  it's possible there is bad data in the test database. There are some ways to view this. I think you use VSCode but in GoLand you could use a break point in the debugger. Then look at the database. 

Also you can comment out the teardown function and that would leave the test databse in tact then you could review it that way. Comment out `ts.PopTestSuite.TearDown()` in the test suite for your test case. These take a little time to fumble with but sometimes it's useful when you feel like you are going crazy.

pkg/services/invoice/process_edi824.go
```diff
@@ -0,0 +1,135 @@
...
+	fmt.Printf(path)
+
+	edi824 := ediResponse824.EDI{}
+	err := edi824.Parse(stringEDI824)
+	if err != nil {
+		e.logger.Error("unable to parse EDI824", zap.Error(err))
+		return fmt.Errorf("unable to parse EDI824")
+	}
+
+	transactionSet := edi824.InterchangeControlEnvelope.FunctionalGroups[0].TransactionSets[0]
```
jacquelineIO, line 42: The indexes/len here and in the OTI will need to be checked. If you get a bad array for some reason this will bomb out. Unless you call validation sooner, that could posssible take care of it.

pkg/services/invoice/process_edi824.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	bgnRefIdentification := bgn.ReferenceIdentification
+	mtoRefID := move.ReferenceID
+	if mtoRefID == nil {
+		e.logger.Error("unable to find PaymentRequest with GCN", zap.Error(err))
+		return fmt.Errorf("The BGN02 Reference Identification field: %s doesn't match the Reference ID %s of the associated move", bgnRefIdentification, *mtoRefID)
+	}
+	if bgnRefIdentification != *mtoRefID {
+		e.logger.Error("unable to find PaymentRequest with GCN", zap.Error(err))
```
jacquelineIO, line 72: Update log comment here.

pkg/services/invoice/process_edi824.go
```diff
@@ -0,0 +1,135 @@
...
+
+	transactionSet := edi824.InterchangeControlEnvelope.FunctionalGroups[0].TransactionSets[0]
+	icn := edi824.InterchangeControlEnvelope.ISA.InterchangeControlNumber
+	otiGCN := transactionSet.OTIs[0].GroupControlNumber
+	bgn := transactionSet.BGN
+
+	var paymentRequest models.PaymentRequest
+	err = e.db.Q().
+		Join("payment_request_to_interchange_control_numbers", "payment_request_to_interchange_control_numbers.payment_request_id = payment_requests.id").
+		Where("payment_request_to_interchange_control_numbers.interchange_control_number = ?", int(otiGCN)).
```
jacquelineIO, line 50: Might need to add some comments to query. 

Something like. 

```
// In the 858, the EDI only has 1 group, and the ICN and the GCN are the same. Therefore, we'll query the PR to ICN table 
// to find the associated payment request using the reported GCN from the 997.
```

pkg/services/invoice/process_edi824.go
```diff
@@ -0,0 +1,135 @@
...
+
+	var move models.Move
+	err = e.db.Q().
+		Find(&move, paymentRequest.MoveTaskOrderID)
+	if err != nil {
+		e.logger.Error("unable to find move with associated payment request", zap.Error(err))
+		return fmt.Errorf("unable to find move with associated payment request: %w", err)
+	}
+
+	bgnRefIdentification := bgn.ReferenceIdentification
```
jacquelineIO, line 65: Some comments here might be useful too 

```
// The BGN02 Reference Identification field from the 824 stores the reference identification used in the 858.
// For MilMove we use the MTO Reference ID in the 858 (which used to the field for the GBLOC, but is not relevant for GHC MilMove).
```

pkg/services/invoice/process_edi824.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	if bgnRefIdentification != *mtoRefID {
+		e.logger.Error("unable to find PaymentRequest with GCN", zap.Error(err))
+		return fmt.Errorf("The BGN02 Reference Identification field: %s doesn't match the Reference ID %v of the associated move", bgnRefIdentification, *mtoRefID)
+	}
+
+	err = edi824.Validate()
+	if err != nil {
+		e.logger.Error("Validation error(s) detected with the EDI824", zap.Error(err))
+		return fmt.Errorf("Validation error(s) detected with the EDI824: %w", err)
```
jacquelineIO, line 79: @gidjin  if we error here, we lose everything and don't know that something was received. This is sort of the conversation we were having during pairing. 

@pearl-truss  maybe we can log to the `edi_errors` table here and just say the validation failed for the EDI 824 along with the err.Error() message.

pkg/services/invoice/process_edi824.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	return nil
+}
+
+func fetchTEDSegments(edi ediResponse824.EDI) []edisegment.TED {
+	var teds []edisegment.TED
+	for _, functionalGroup := range edi.InterchangeControlEnvelope.FunctionalGroups {
+		for _, transactionSet := range functionalGroup.TransactionSets {
+			for _, ted := range transactionSet.TEDs {
```
jacquelineIO, line 129: I think you could also just do 

```
teds = append(teds,  transactionSet.TEDs)
```
and not have to loop through them.

pkg/services/invoice/process_edi824.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		e.logger.Error("Validation error(s) detected with the EDI824", zap.Error(err))
+		return fmt.Errorf("Validation error(s) detected with the EDI824: %w", err)
+	}
+
+	teds := fetchTEDSegments(edi824)
+
+	var transactionError error
+	transactionError = e.db.Transaction(func(tx *pop.Connection) error {
+		prToICN := models.PaymentRequestToInterchangeControlNumber{
```
jacquelineIO, line 86: Can you add a comment here, something like

```
// Store the 824 ICN along with the Payment Request ID
```

jacquelineIO (commented):
Good updates. Some minor comments and checks needed still for the PR.

pkg/services/invoice/process_edi824.go
jacquelineIO: Looks like you'd need `...` to make what I said work. https://stackoverflow.com/a/16252034

pkg/services/invoice/process_edi824.go
gidjin: I'm not sure what we can do with invalid 824s, other than work with our partners at Syncada to sort out what's wrong. I think logging to the errors table is a good plan for now. I wonder if we can tie the error to the file name in our error saving.

pkg/services/invoice/process_edi824.go
```diff
@@ -0,0 +1,167 @@
...
+
+	var transactionError error
+	var otiGCN int64
+	var bgn edisegment.BGN
+	transactionError = e.db.Transaction(func(tx *pop.Connection) error {
+		icn := edi824.InterchangeControlEnvelope.ISA.InterchangeControlNumber
+		if edi824.InterchangeControlEnvelope.FunctionalGroups != nil {
+			if edi824.InterchangeControlEnvelope.FunctionalGroups[0].TransactionSets != nil {
+				transactionSet := edi824.InterchangeControlEnvelope.FunctionalGroups[0].TransactionSets[0]
+				otiGCN = transactionSet.OTIs[0].GroupControlNumber
```
jacquelineIO, line 50: Thank you for the updates here. You will still have to check the index on the OTIs too. If we get a file/format with 0 len OTIs this will cause a failure.

pkg/services/invoice/process_edi824.go
```diff
@@ -0,0 +1,167 @@
...
+			First(&paymentRequest)
+		if err != nil {
+			e.logger.Error("unable to find PaymentRequest with GCN", zap.Error(err))
+			return fmt.Errorf("unable to find PaymentRequest with GCN: %s, %d", err.Error(), int(otiGCN))
+		}
+
+		prToICN := models.PaymentRequestToInterchangeControlNumber{
+			InterchangeControlNumber: int(icn),
+			PaymentRequestID:         paymentRequest.ID,
+		}
```
jacquelineIO, line 73: Now seeing how saving to the PR to ICN table works. I'm going to write 1 pt story to add the EDI Type column to the table. Which means I can probably drop it from the `edi_errors` table. Nothing needed of this PR.

pkg/services/invoice/process_edi824.go
jacquelineIO: Something to consider, might be a good test to pass in the file with no OTIs or no FGs, TSs

pkg/services/invoice/process_edi824_test.go
```diff
@@ -0,0 +1,350 @@
...
+		})
+		paymentRequest := testdatagen.MakePaymentRequest(suite.DB(), testdatagen.Assertions{
+			PaymentRequest: models.PaymentRequest{
+				MoveTaskOrderID: move.ID,
+			},
+		})
+		testdatagen.MakePaymentRequestToInterchangeControlNumber(suite.DB(), testdatagen.Assertions{
+			PaymentRequestToInterchangeControlNumber: models.PaymentRequestToInterchangeControlNumber{
+				PaymentRequestID:         paymentRequest.ID,
+				InterchangeControlNumber: 100001251,
```
jacquelineIO, line 70: You are correctly using the ICN here. Did you update your test or what this failing before you changed the code?

pkg/services/invoice/process_edi997.go
```diff
@@ -29,54 +29,88 @@ func NewEDI997Processor(db *pop.Connection,
...
+	// Find the PaymentRequestID that matches the GCN
 	icn := edi997.InterchangeControlEnvelope.ISA.InterchangeControlNumber
+	var gcn int64
+	if edi997.InterchangeControlEnvelope.FunctionalGroups != nil {
+		if edi997.InterchangeControlEnvelope.FunctionalGroups[0].TransactionSets != nil {
+			transactionSet := edi997.InterchangeControlEnvelope.FunctionalGroups[0].TransactionSets[0].FunctionalGroupResponse.AK1
+			gcn = transactionSet.GroupControlNumber
+		}
+	} else {
+		e.logger.Error("Validation error(s) detected with the EDI824. EDI Errors could not be saved", zap.Error(err))
```
jacquelineIO, line 49: The comments here say 824

pkg/services/invoice/process_edi997.go
```diff
@@ -29,54 +29,88 @@ func NewEDI997Processor(db *pop.Connection,
...
+		return fmt.Errorf("unable to parse EDI997")
 	}
 
-	// Find the PaymentRequestID that matches the ICN
+	// Find the PaymentRequestID that matches the GCN
 	icn := edi997.InterchangeControlEnvelope.ISA.InterchangeControlNumber
+	var gcn int64
+	if edi997.InterchangeControlEnvelope.FunctionalGroups != nil {
+		if edi997.InterchangeControlEnvelope.FunctionalGroups[0].TransactionSets != nil {
+			transactionSet := edi997.InterchangeControlEnvelope.FunctionalGroups[0].TransactionSets[0].FunctionalGroupResponse.AK1
```
jacquelineIO, line 45: `transactionSet` variable name should be `ak1`

pkg/services/invoice/process_edi824_test.go
pearl-truss: I'm not sure what you mean? This test is for when everything works and wasn't failing previously

jacquelineIO (approved):
thank you Pearl. lgtm. 

This will be tested with another PR that will trigger the CLI or async task call for this code.

PR #6288, opened 2021-03-31
MB-7400 Add newly required fields to createMoveTaskOrder in Support API

sandy-wright:
## Description

This PR updates the Support API's `createMoveTaskOrder` endpoint to include two new fields that are required for approving the move later on. It also adds an integration test to the tests for this endpoint so that we can be sure that any future changes that break our ability to approve a freshly created move are caught with the original PR.

## Setup

First, run your server:

```sh 
```

Then, using the testing method of your choice (Postman or the Prime API Client), hit the Support API's `createMoveTaskOrder` endpoint with a valid body. You'll need to set:
 - the move's `status` to `"SUBMITTED"`,
 - the order's `status` to `"APPROVED"`,
 - and fill in valid values for the `ordersTypeDetail` and `departmentIndicator` fields in particular.

All the other fields can be filled in however you like - just remember to include everything that's marked as required in `support.yaml`.

Once you have successfully created a new move, grab the new ID and the eTag body from the response. Using these values, submit a request to the `makeMoveTaskOrderAvailable` endpoint.

Verify that you were successfully able to approve and mark this move as available to the Prime. Yay!

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7400) for this change.


pkg/handlers/supportapi/move_task_order_test.go
```diff
@@ -231,11 +230,14 @@ func (suite *HandlerSuite) TestCreateMoveTaskOrderRequestHandler() {
 	// Create the mto payload we will be requesting to create
 	issueDate := swag.Time(time.Now())
 	reportByDate := swag.Time(time.Now().AddDate(0, 0, -1))
+	ordersTypedetail := supportmessages.OrdersTypeDetailHHGPERMITTED
+	deptIndicator := supportmessages.DeptIndicatorAIRFORCE
+	selectedMoveType := supportmessages.SelectedMoveTypeHHG
 
 	mtoPayload := &supportmessages.MoveTaskOrder{
-		AvailableToPrimeAt: handlers.FmtDateTime(time.Now()),
```
eamahanna, line 236: Why is this field no longer needed?

eamahanna (approved):
LGTM :)

pkg/handlers/supportapi/move_task_order_test.go
sandy-wright: Oh good question. This field is updated when we use the `makeAvailableToPrime` endpoint, and if this value is set before we hit that endpoint, it causes an error and the update won't take. This endpoint also marks the move as "APPROVED," and it _has_ to be approved for just about every other action down the line.

So basically, in load testing, we can't set this field when we're creating the MTO or everything else breaks. I removed it here as well because I want these tests to stay as close as possible to the actual use case, with the hope that it's easier to catch when something changes that's going to break load testing.

PR #6297, opened 2021-04-01
MB-7211 update create mto service item error handling

eamahanna:
## Description

This PR edits the error messaging for createMTOServiceItem. When an address payload has an ID that is non-null, it should fail and return a 422. If the service item is a SITStandaloneServiceItem, then the address ID can be non-null or match the existing service item's address ID.

## Setup

To run the test, execute `make server_test`

To test via postman use the Prime API to create a Destination SIT service item. Use the service item's ID as the `sitDestinationFinalAddress.ID`, and a 422 failure should be returned. Use a null uuid as the `sitDestinationFinalAddress.ID`, and it should succeed.  

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7211) for this change


pkg/testdatagen/scenario/devseed.go
```diff
@@ -766,7 +766,9 @@ func createHHGWithPaymentServiceItems(db *pop.Connection, userUploader *uploader
 	// have a departure date for the payment request param lookup to not encounter an error
 	entryDate := time.Now().Add(-29 * 24 * time.Hour)
 
-	originSITAddress := testdatagen.MakeAddress(db, testdatagen.Assertions{})
+	originSITAddress := testdatagen.MakeAddress(db, testdatagen.Assertions{Stub: true})
+	originSITAddress.ID = uuid.Nil
```
sandy-wright, line 770: Thanks for updating this!

pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -417,6 +417,7 @@ func validateTimeMilitaryField(timeMilitary string) error {
 	return nil
 }
 
+// Check if and address has and ID, if it does, it needs to match OG SIT
```
sandy-wright, line 420: Small typo - should be "an" address, I think 😄

pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -447,6 +448,21 @@ func (o *mtoServiceItemCreator) validateSITStandaloneServiceItem(serviceItem *mo
...
+			fmt.Sprintf("SITOriginHHGOriginalAddressID must be nil or match existing ID and it is %s", serviceItem.SITOriginHHGOriginalAddress.ID))
+	}
+
+	if serviceItem.SITOriginHHGActualAddress != nil && serviceItem.SITOriginHHGActualAddress.ID != mtoServiceItem.SITOriginHHGActualAddress.ID {
+		verrs := validate.NewErrors()
+		verrs.Add("SITOriginHHGActualAddress", fmt.Sprintf("%s invalid SITOriginHHGActualAddressID", serviceItem.ReService.Code))
+		return nil, services.NewInvalidInputError(serviceItem.ID, nil, verrs,
+			fmt.Sprintf("SITOriginHHGActualAddressID must be nil or match existing ID and it is %s", serviceItem.SITOriginHHGActualAddress.ID))
+	}
+
```
sandy-wright, line 465: I want to be clear that this is NOT REQUIRED, but you could refactor this slightly to compile the validation errors and then make the `InvalidInputError` only once. For example:

```go
verrs := validate.NewErrors()

if serviceItem.SITOriginHHGOriginalAddress // the ID isn't empty {
    verrs.Add( // error stuff )
}

if serviceItem.SITOriginHHGActualAddress // same dealio {
    verrs.Add( // error stuff )
}

if verrs.HasAny() {
    return nil, services.NewInvalidInputError("There was invalid input in the standalone service item", // other bits)
}
```

The advantage here is that if, for some reason, both IDs were incorrectly set, the user would see the validation errors for both with one response. I think we're moving towards that model in general, so it might be nice to implement it here (and in the following function). But again, NOT a required change.

sandy-wright (approved):
LGTM! I left some comments for a suggested refactor, but it works as expected now 👍

pkg/services/mto_service_item/mto_service_item_creator.go
eamahanna: oooooo good call

PR #6298, opened 2021-04-01
Remove constant condition of false

kimallen:
## Description

A condition that could only be false was found in the dangerfile.ts.  Inquiry showed there did not seem to be a reason for it, and it likely was a forgotten debugging statement.

## Reviewer Notes

Discussion: https://ustcdp3.slack.com/archives/CP6PTUPQF/p1617312210173800


gidjin (approved):
Thanks for fixing this! 🚀

PR #6299, opened 2021-04-02
Updated pricerDisplayParam to string

shimonacarvalho:
## Description

`PricingDisplayParam.Value` will be stored in the db table `PaymentServiceItemParams` as a string and returned to UI, audit records or Prime. Therefore it needs to be formatted as a string that is human readable/comprehensible, so we will be changing the format of the field to a string in the `PricingDisplayParam` type as well. 

Type updated, tests fixed. 

## Setup

Run the tests for the `ghcrateengine`
```
go test ./pkg/services/ghcrateengine/ -v
```

## References

* [Jira story](https://dp3.atlassian.net/browse/mb-7541) for this change
* [this slack conversation](https://ustcdp3.slack.com/archives/CP497TGAU/p1617233952146100) explains more about the approach used.

### DIFF
--- a/pkg/services/ghcrateengine/pricer_helpers.go
+++ b/pkg/services/ghcrateengine/pricer_helpers.go
@@ -160,8 +160,8 @@ func priceDomesticPickupDeliverySIT(db *pop.Connection, pickupDeliverySITCode mo
 	return totalPriceCents, nil, nil
 }
 
-// createPricerGeneratedParams stores params, whose origin is the PRICER, into the database
-// It also returns the
+// createPricerGeneratedParams stores PaymentServiceItemParams, whose origin is the PRICER, into the database
+// It also returns the newly created PaymentServiceItemParams.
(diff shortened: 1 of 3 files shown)


pkg/services/ghcrateengine/pricer_helpers.go
```diff
@@ -168,37 +170,24 @@ func createPricerGeneratedParams(db *pop.Connection, paymentServiceItemID uuid.U
...
 	for _, param := range params {
+
+		// Find the paymentServiceItemParam associated with this PricingDisplayParam
 		var serviceItemParamKey models.ServiceItemParamKey
 		err := db.Q().
 			Where("key = ?", param.Key).
 			First(&serviceItemParamKey)
 		if err != nil {
-			return paymentServiceItemParams, fmt.Errorf("Unable to find service item param key for %v", serviceItemParamKey.Key)
+			return paymentServiceItemParams, fmt.Errorf("Unable to find service item param key for %v", param.Key)
```
gidjin, line 180: Nice catch of a subtle issue!

gidjin (approved):
This seems simpler, changes look good to me. Left a comment with one minor nit but no need for a re-review on my part.

PR #6322, opened 2021-04-05
MB-7112 Add service counseling Move statuses

monfresh:
## Description

This PR adds two new Move statuses to support Service Counseling:
- MoveStatusNeedsServiceCounseling
- MoveStatusServiceCounselingCompleted

The Move status column comment was updated to explain how these new
statuses work as part of the end to end workflow.

Note that to make the long status names readable, we made them Pascal
case as opposed to uppercase like the existing ones. If we like this
new convention, we can update the case of the older ones.

For readability, we formatted the column comment with newlines and
paragraphs. It doesn't seem to have any adverse effects in TablePlus,
but in `psql` it seems harder to read. We are looking for feedback
from folks with more experience with comments.

## Reviewer Notes

Run the migrate command in the setup steps below, and verify you can see the new comments.

## Setup

```sh
make db_dev_migrate
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7112) for this change


migrations/app/schema/20210405200247_update_move_status_comment.up.sql
```diff
@@ -0,0 +1,27 @@
+COMMENT ON COLUMN moves.status IS 'The current status of the Move. Allowed values are:
+DRAFT,
+SUBMITTED,
+APPROVED,
+APPROVALS REQUESTED,
+CANCELED,
+NEEDS SERVICE COUNSELING,
+SERVICE COUNSELING COMPLETED.
+
+A Move starts out in DRAFT status upon creation. When the customer submits the
```
tinyels, line 10: Could we put a link to the miro on statuses here (possibly instead of this incomplete description)?

pkg/models/move.go
```diff
@@ -34,6 +34,10 @@ const (
 	MoveStatusCANCELED MoveStatus = "CANCELED"
 	// MoveStatusAPPROVALSREQUESTED captures enum value "APPROVALS REQUESTED"
 	MoveStatusAPPROVALSREQUESTED MoveStatus = "APPROVALS REQUESTED"
+	// MoveStatusNeedsServiceCounseling captures enum value "NEEDS SERVICE COUNSELING"
+	MoveStatusNeedsServiceCounseling MoveStatus = "NEEDS SERVICE COUNSELING"
```
tinyels, line 38: Do you want to update
https://github.com/transcom/mymove/blob/7010fa6c7f3c9c07b02b80808e15b3da71d3774f/src/shared/constants.js#L22
and https://github.com/transcom/mymove/blob/7010fa6c7f3c9c07b02b80808e15b3da71d3774f/src/constants/moves.js#L2

migrations/app/schema/20210405200247_update_move_status_comment.up.sql
monfresh: That's a good idea.

pkg/models/move.go
monfresh: Yes, will do. I wonder why we define the constant in two places, and in the first place, it's missing the Approvals Requested status. I'll check to see if we're not using that first one.

monfresh: Hmm. The constant in `mymove/src/constants/moves.js` only seems to be imported in `src/utils/shipmentInfo.js` and the only status that's queried there is `DRAFT`. 

The other constant in `mymove/src/shared/constants.js` is used in several places, so I added them there.

We probably also want to update is `src/constants/queues.js`, but now that I think about it, that is probably out of scope for this PR. We typically separate frontend and backend work in different stories. I think it makes more sense to update the queue statuses in the other stories.

duncan-truss: We've slowly been moving things out of src/shared and into different folders since this ADR https://github.com/transcom/mymove/blob/master/docs/adr/0048-frontend-file-org.md and layout structure https://github.com/transcom/mymove/wiki/frontend#file-layout--naming

`shared` is the legacy folder and `constants` is the preferred new location.  The customer app likely still has a number of uses of shared and we still need to replace it in a few office files too.  Seems like the service counseling work will be a good time to make that happen at some point in this slice.

duncan-truss:
The readability in schemaspy could be better maybe if we didn't have the hard character line limit.  If you look at orders.orders_type we have a longer description there that just wraps fwiw.

Screen Shot 2021-04-05 at 5 36 24 PM

monfresh:
I've replaced that long description with a link to the Miro board per Erin's suggestion. Here's what it looks like now in schemaspy:

Screen Shot 2021-04-05 at 5 49 22 PM

pkg/models/move.go
tinyels: Perhaps create a new story to do that cleanup?

tinyels (approved):

PR #6334, opened 2021-04-06
[MB-7220] Turn flag on for deleting files from Syncada

mboren:
## Description

I had this flag set to false so it wouldn't clear out the files I was using for testing, and forgot to flip it back.

## Setup

- Create a move
- approve with TOO
- use prime demo script and TIO interface to create a payment request and send it
- Run the following command a couple of times to send the request to syncada and process it when it comes back.`prime-api-client --insecure support-reviewed-payment-requests --filename readfromsyncada.json`
It should eventually change your payment request status to RECEIVED_BY_GEX, and delete the file from syncada


pearl-truss (approved):
LGTM

mboren:
Replaced by https://github.com/transcom/mymove/pull/6359. We decided to turn this into a parameter instead of hardcoding it.

PR #6339, opened 2021-04-06
Kim mb 5882 profile icon link

kimallen:
## Description

This adds a visible Profile icon to the navigation bar for the Customer.

## Reviewer Notes

The Customer's milmove header needs to be refactored such that it uses the MilmoveHeader component.  The existing header uses a `LoginButton` component that holds a lot of logic that made it not an easy swap-out, so I just made the icon visible and clickable for now.  It's a bit odd to be added to the `LoginButton` component semantically, but that's what makes sense for how the code is constructed around the Menu/buttons setup right now.  Note that the 'Menu' item will be removed in the refactor.

The screenshots at the bottom of this summary are 3x size for the icon.  Does this work?  below is 2x, which Claire thought could be bigger.  I don't think there is an in-between size:
image

## Setup

log in as a Customer.
Navigate past the profile wizard.
See the profile icon.  It should click to a non-existent page (which is being created in a concurrent story this sprint)
When in mobile or narrow view, you will see the 'Menu' button.  Clicking on this, you will see the Profile icon and the signout option.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5882) for this change
* [Story that best captures the needed header refactor](https://dp3.atlassian.net/browse/MB-247)
*
## Screenshots

desktop view
image

Phone and narrow view
image


src/containers/LoginButton/LoginButton.jsx
```diff
@@ -58,17 +60,24 @@ const LoginButton = ({ isLoggedIn, logOut, showDevlocalButton }) => {
...
-        data-testid="signout"
-        onClick={handleLogOut}
-        type="button"
-      >
-        Sign Out
-      </Button>
-    </li>
+    <>
+      <li className="usa-nav__primary-item">
+        <Link to="/moves/review/profile" title="profile-link" aria-label="profile-link" className={styles.profileLink}>
```
suzubara, line 65: this brings me to the 404 page, is it supposed to be hooked up to something? Also if possible can we use a value defined in `src/constants/routes.js` for the URL (we can add there as needed)

src/shared/Header/MyMove/index.scss
```diff
@@ -1,3 +1,5 @@
+@import 'shared/styles/colors';
```
suzubara, line 1: it doesn't look like this file is using any values from this import, can it be removed?

src/shared/Header/MyMove/index.scss
```diff
@@ -29,14 +30,12 @@
   background-color: #252f3e;
 }
 
-.my-move-header .usa-nav {
+.my-move-header .usa-nav__primary {
   padding: 0;
```
suzubara, line 34: whoops, I think the `.usa-nav` element still needs `padding: 0;`, so this change should be a new selector instead of changing the existing one

src/containers/LoginButton/LoginButton.jsx
```diff
@@ -58,17 +60,24 @@ const LoginButton = ({ isLoggedIn, logOut, showDevlocalButton }) => {
...
+          <FontAwesomeIcon className="fa-3x" icon={['far', 'user']} />
```
suzubara, line 66: this looks really big to me, is it supposed to be 3x or 2x?

src/containers/LoginButton/LoginButton.jsx
kimallen: I will create a const.  This page doesn't exist yet.  That story is currently being worked on this sprint: https://dp3.atlassian.net/browse/MB-5881.

src/containers/LoginButton/LoginButton.jsx
kimallen: When I put a 2x, Claire thought it should be bigger.  I'm waiting for review from @cosburn to decide the size

kimallen: Confirmed changing it back to 2x

suzubara (approved):

cosburn (approved):

PR #6344, opened 2021-04-06
[MB-7407] Record EDI 858 generation errors in database

mboren:
## Description
When we run into an error while generating a EDI 858, the error will now be recorded in the `edi_errors` table.

If we have new errors while trying to save the error, I chose to log the new errors and return the original error.

I made `EdiError.InterchangeControlNumberID` nullable so that we can add errors for 858s without looking up their ICN. The ICN is not meaningful for 858s because the errors are happening on our end, and the ICN is used by Syncada.

## Reviewer Notes
I'm not sure if my manual test case is due to a payment request in our e2e data being set up wrong, or just being incomplete, and I'm not sure how to set up a PR that would break 858 generation from scratch. I'm open to any suggestions about other ways to test this.

## Setup
```sh
make db_dev_e2e_populate
make server_run
```

sendtosyncada.json:
```json
{
  "body": {
    "sendToSyncada": true,
    "readFromSyncada": false
  }
}
```
```sh
```
This should crash while generating an 858. This response will look something like:
```
{"detail":"function ProcessReviewedPaymentRequest failed call to generator.Generate: id: ebf89e73-06e9-4150-a9e0-fde56a651a22 not found for paymentServiceItemParam","instance":"f4bb3f32-8782-4a97-9891-107ef89edb2d","title":"Internal Server Error"}
```

Then, check the `edi_errors` table:
```sql
select * from edi_errors;
```
You should see that a single record has been created.
The `description` column in that record should look something like:
```function ProcessReviewedPaymentRequest failed call to generator.Generate: id: ebf89e73-06e9-4150-a9e0-fde56a651a22 not found for paymentServiceItemParam```

## Acceptance
Since the changes are only visible in the DB, I don't think there is acceptance to do.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7407) for this change


gidjin (approved):
Changes look good. ICNs are from a sequence so I don't expect us to not be able to get one but how you deal with that should be good. As for manual test case, trying to generate an 858 with out a TAC will fail for sure. I think there are a few other places too if you look through the validation rules in the segments filled in to create the 858. However, I'm not sure there is a way to set it up without messing with the database or code.

jacquelineIO:
> doesn't let you leave the ICN field blank
The intention was the allow blank ICN fields, so either you aren't on the latest or I goofed the in the sql

mboren:
> Do you think you can modify that column to be allow null?
Done. I made the column nullable and removed the query to try to add the icn to 858 errors.

lynzt (approved):
I looked at the migration only but LGTM 🚢

jacquelineIO (commented):
LGTM. One thing to add is that once we log the error we should change the payment request status to EDI_ERROR.

jacquelineIO (approved):
looks good! thanks for the updates. was able to re-run CLI to see 2 payment requests update to `EDI_ERROR`

PR #6354, opened 2021-04-07
[MB-7422] Send all moves to service counseling

hannaliebl:
## Description

Send all moves to service counseling by setting their status to the new `MoveStatusNeedsServiceCounseling` status (from: https://github.com/transcom/mymove/pull/6322). We added a service object to handle this logic (anticipating that later we will need to determine which moves get this new status and which get the old "submitted" status), and put this behind a feature flag called `FEATURE_FLAG_SERVICE_COUNSELING`.

## Reviewer Notes

Because this is adding a feature flag and will (likely) affect other teams, is there some place I should be documenting this?

## Setup

```sh
make db_dev_e2e_populate
make server_run
make client_run
```

1. This work is behind a feature flag called `FEATURE_FLAG_SERVICE_COUNSELING`. It is set to `true` by default in `.envrc`. When the flag is enabled, any new move that you submit will still show up in the TOO queue, but it should have the status `NEEDS SERVICE COUNSELING` and you won't be able to approve it because approval requires completion of Service Counseling, which isn't implemented yet. 
2. Run the MyMove app, and login as `hhg@only.unsubmitted` test user
3. Go to Review and Submit, click Next
4. Enter a signature, open the network tab, and click on Complete
5. The submit POST should respond with an object that includes `status: "NEEDS SERVICE COUNSELING"`

To verify that moves are not sent to service counseling when the flag is off:
1. Change the feature flag in `.envrc.local` to `export FEATURE_FLAG_SERVICE_COUNSELING=false`, run `direnv allow` after you make the change.
2. Run `make db_dev_e2e_populate` to reset the test data, then run the server/client
3. In the MyMove app, login to `hhg@only.unsubmitted` test user
4. Go to Review and Submit, click Next
5. Enter a signature, open the network tab, and click on Complete
6. The submit POST should respond with an object that includes `status: "SUBMITTED"`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7422) for this change


pkg/services/move/move_status_router.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/spf13/viper"
+	"go.uber.org/zap"
+
+	"github.com/transcom/mymove/pkg/cli"
+
+	"github.com/transcom/mymove/pkg/models"
+	"github.com/transcom/mymove/pkg/services"
+)
+
+type moveStatusRouter struct {
```
monfresh, line 17: @hannaliebl I'm thinking of renaming this to `moveRouter`, and have it be the place where all move routing happens. So, we would move the `Submit` and `SendToServicesCounseling` functions out of the Move model and into here, and we would rename `Submit` to `SendToOfficeUser` or something like that, and instead of `RouteMove` down below, it could become `Submit` to match the action that the customer takes. The move gets submitted by the customer, and then the MoveRouter either sends to service counseling, or to the office user.

hannaliebl: That makes sense to me. Are there other ways a move can "move," though, that would then need to be handled in there? I'm still not familiar enough with the whole app to know if that is the case.

Ryan-Koch: I can see that idea working out. I suspect that'll take some doing, so I wonder if we should tackle it when we come back and readdress this later when we're adding in the logic around how the split is going to work.

pkg/models/move.go
```diff
@@ -164,16 +164,56 @@ func (m *Move) Submit(submittedDate time.Time) error {
 	return nil
 }
 
+// SendToServiceCounseling sends the move to needs service counseling
+func (m *Move) SendToServiceCounseling() error {
+	if m.Status != MoveStatusDRAFT {
+		return errors.Wrap(ErrInvalidTransition, fmt.Sprintf("Cannot move to NeedsServiceCounseling state when the Move is not in Draft status: %s", m.Status))
```
Ryan-Koch, line 170: It looks like the status this error is reporting back is the status of the move that came in, which is helpful. Could we adjust the wording a bit to make it clear that's what it's sending back? Maybe something like: `Cannot move to NeedsServiceCounseling state when the Move is not in Draft status. Its current status is %s` (the approve() version of this uses a similar phrasing).

Ryan-Koch (approved):
Left a minor note about a comment. Was able to verify this does the thing so LGTM!

pkg/services/move/move_status_router.go
monfresh: Oh yeah, I didn't mean in this PR. Here's the JIRA ticket I created for this: https://dp3.atlassian.net/browse/MB-7635. I'd love if we could work on this next sprint while it's still fresh in my mind.

PR #6356, opened 2021-04-07
Kim mb 7561 fix errcheck violations

kimallen:
## Description

Remediates all remaining errcheck violations (that are not already annotated) and enables the errcheck linter.

## Reviewer Notes

There was some question as to whether we can disable `errcheck` on all tests or not.  For now, we will not disable and work to make sure all errs are handled.

## Setup

`pre-commit run -v --all-files golangci-lint`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7561) for this change


pkg/models/order_test.go
```diff
@@ -36,7 +39,10 @@ func (suite *ModelSuite) TestTacNotNilAfterSubmission() {
 		suite.T().Fatal("Should transition.")
 	}
 	suite.MustSave(&move)
-	suite.DB().Load(&order, "Moves")
+	err = suite.DB().Load(&order, "Moves")
+	if err != nil {
+		log.Fatal("issue loading order from db", zap.Error(err))
```
felipe-lee, line 44: Should these log or fail the test? (same for this whole file)

pkg/services/invoice/process_edi824_test.go
```diff
@@ -218,11 +218,16 @@ IEA*1*00000005
...
+		if err != nil {
+			suite.NoError(err)
+		}
 
 		var updatedPR models.PaymentRequest
-		err := suite.DB().Where("id = ?", paymentRequest.ID).First(&updatedPR)
-		suite.NoError(err)
+		err = suite.DB().Where("id = ?", paymentRequest.ID).First(&updatedPR)
+		if err != nil {
+			log.Fatal("issue getting payment request", zap.Error(err))
```
felipe-lee, line 229: should this one fail the test?

felipe-lee (commented):
Question about logging vs failing for tests, but if logging is good, then LGTM

pkg/models/order_test.go
monfresh: They should fail the test. I would assert that there is no error: `suite.NoError(err)`.

monfresh: The test should fail anyways if the Moves aren't loaded, but maybe it's cleaner to assert that there's no error, which avoids adding the log and zap dependencies.

pkg/models/payment_request_to_interchange_control_number_test.go
```diff
@@ -15,7 +18,10 @@ func (suite *ModelSuite) TestPaymentRequestToInterchangeControlNumber() {
 		validPR2ICN := testdatagen.MakePaymentRequestToInterchangeControlNumber(suite.DB(), testdatagen.Assertions{
 			PaymentRequestToInterchangeControlNumber: models.PaymentRequestToInterchangeControlNumber{PaymentRequestID: paymentRequest.ID}})
 		suite.Equal(paymentRequest.ID, validPR2ICN.PaymentRequestID)
-		suite.DB().Load(&validPR2ICN, "PaymentRequest")
+		err := suite.DB().Load(&validPR2ICN, "PaymentRequest")
+		if err != nil {
+			log.Fatal("could not fetch payment request", zap.Error(err))
+		}
```
monfresh, line 24: Same comment here about using `suite.NoError(err)` instead of adding dependencies.

pkg/services/invoice/process_edi824_test.go
```diff
@@ -218,11 +218,16 @@ IEA*1*00000005
 `
 		paymentRequest := testdatagen.MakePaymentRequest(suite.DB(), testdatagen.Assertions{})
 
-		edi824Processor.ProcessFile("", sample824EDIString)
+		err := edi824Processor.ProcessFile("", sample824EDIString)
+		if err != nil {
```
monfresh, line 222: I don't think this conditional is needed. You can assert directly with `suite.NoError(err)`.

pkg/models/order_test.go
kimallen: I've tended to add a `log.Fatal` when the failure is related to something about setting up the test and the asserts that are testing what I'm checking in the test, and only using the assert for what I'm checking in the test.  This is the pattern I've seen used before. I'll put a question out in #prac-engineering and see what the thoughts are.

pkg/services/invoice/process_edi824_test.go
kimallen: changed it to suite.NoError

pkg/models/order_test.go
kimallen: Ok, I switched it to suite.NoError.  Thanks for the additional context and info in Slack, @monfresh !

felipe-lee (approved):

PR #6357, opened 2021-04-07
[MB-6830] update to shorthaul pricer to return rate engine params

jacquelineIO:
## Description

Update the DSH pricer to return pricing rate engine params so that they are saved to the database. There are 4 params added:

- contract year name
- escalation compounded
- is peak
- price/rate/factor

## Reviewer Notes

LH pricer was also recently updated merged in this PR https://github.com/transcom/mymove/pull/6341

## Setup

Currently, can't test DSH. There is a good amount of UT around it.

Need to update setup here, but setup from PR https://github.com/transcom/mymove/pull/6341 can be followed making sure to use DSH (make move within same zip3). But not sure DSH can be setup in our manual testing yet.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6830) for this change

## Screenshots


gidjin (commented):
Changes look good, I left one nit that you can fix if you like. I'm not approving it because I couldn't run it locally, for some reason I was hitting an issue completing new move submission. I don't think it's related but wanted to let someone else walk through it tomorrow before approving.

pkg/services/ghcrateengine/ghc_rate_engine_service_test.go
```diff
@@ -185,3 +185,20 @@ func (suite *GHCRateEngineServiceSuite) HasDisplayParam(displayParams services.P
...
+
+	var foundAParamKey bool
+	for _, aValue := range actualValues {
+		foundAParamKey = false
+		for _, eValue := range expectedValues {
+			if aValue.Key == eValue.Key {
+				suite.Equal(eValue.Value, aValue.Value, "Matching expected value for actual value ServiceItemParamKey %s", eValue.Key)
+				foundAParamKey = true
+			}
+		}
```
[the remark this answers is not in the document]
jacquelineIO: It's not really doing the same thing, because I check all the values at once and not one at a time. I would have to add more code, I think, to run what Reggie has.

@gidjin I don't think you can run with Shorthaul. I'm going to continue to work on https://github.com/transcom/mymove/pull/6368 because even if we don't get buy-in from the front end and Prime API POV, I can make linehaul work for both shorthaul and regular linehaul and we don't care what the shipment HHG tag is.

pkg/services/ghcrateengine/domestic_shorthaul_pricer.go
```diff
@@ -64,24 +67,25 @@ func (p domesticShorthaulPricer) Price(contractCode string, requestedPickupDate
...
-	//)
-	return totalCost, nil, nil
+	var pricingRateEngineParams = services.PricingDisplayParams{
+		{
+			Key:   models.ServiceItemParamNameContractYearName,
+			Value: contractYear.Name,
+		},
+		{
+			Key:   models.ServiceItemParamNamePriceRateOrFactor,
+			Value: domServiceAreaPrice.PriceCents.String(),
```
reggieriser, line 77: This is going to leave the price in cents, but other pricers are using the `FormatCents` helper to return it in dollars.  This [Slack thread](https://ustcdp3.slack.com/archives/CP4979J0G/p1617724802047300) has some context.

pkg/services/ghcrateengine/ghc_rate_engine_service_test.go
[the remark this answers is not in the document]
reggieriser: I'm wondering if @gidjin is suggesting that your function could stay, but just call `HasDisplayParam` as it tries to confirm that each parameter is there from your `expectedValues` slice.  Alternatively, we could have a story to consolidate pricers to use one or the other helpers.

pkg/services/ghcrateengine/domestic_shorthaul_pricer_test.go
```diff
@@ -291,3 +321,42 @@ func (suite *GHCRateEngineServiceSuite) setUpDomesticShorthaulData() {
 	domesticShorthaulNonpeakPrice.PriceCents = 127
 	suite.MustSave(&domesticShorthaulNonpeakPrice)
 }
+
+func (suite *GHCRateEngineServiceSuite) getExpectedDSHPricerCreatedParamsFromDBGivenParams(serviceArea string, requestedPickUp string) services.PricingDisplayParams {
```
reggieriser, line 325: The other pricers so far haven't been going to the DB to read the expected parameter values since the same test is writing to the DB with the test values (stored in constants).  In theory, we should know what the expected values are without looking in the DB (not sure if shorthaul is different though).  Just noting -- going to the DB should still be fine.

pkg/services/ghcrateengine/domestic_shorthaul_pricer_test.go
```diff
@@ -148,10 +170,11 @@ func (suite *GHCRateEngineServiceSuite) TestPriceDomesticShorthaul() {
 
 		suite.Error(err)
 		suite.Equal("Could not lookup Domestic Service Area Price: "+models.RecordNotFoundErrorString, err.Error())
+		suite.validatePricerCreatedParams(nil, rateEngineParams)
```
reggieriser, line 173: Nit: I see above that you use a `suite.Nil` to verify that a nil is returned, but here (and below) you're using the `validatePricerCreatedParams` instead.  Any particular reason to not use the slightly simpler `suite.Nil` here?

pkg/services/ghcrateengine/domestic_shorthaul_pricer.go
jacquelineIO: @reggieriser  OK, I will change it. I did see that in your PR and didn't realize that I didn't have it here.

pkg/services/ghcrateengine/domestic_shorthaul_pricer_test.go
jacquelineIO: oversight, I will update it.

pkg/services/ghcrateengine/domestic_shorthaul_pricer_test.go
jacquelineIO: @reggieriser  I considered that but also realized that the database saves were previously tested so felt comfortable doing it, but I do agree this isn't the normal case.

pkg/services/ghcrateengine/ghc_rate_engine_service_test.go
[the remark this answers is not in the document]
jacquelineIO: @reggieriser  I'm not sure what benefit that would give me. I'm setting up my expected values to have all of the needed keys. So "has key" would be redundant I think.

pkg/services/ghcrateengine/domestic_shorthaul_pricer.go
```diff
@@ -74,15 +74,15 @@ func (p domesticShorthaulPricer) Price(contractCode string,
 		},
 		{
 			Key:   models.ServiceItemParamNamePriceRateOrFactor,
-			Value: domServiceAreaPrice.PriceCents.String(),
+			Value: FormatFloat(domServiceAreaPrice.PriceCents.ToDollarFloatNoRound(), 3),
```
reggieriser, line 77: According to the pricing template, we only record up to 2 decimal points here (no smaller than a cent) for DSH, so I would think this could just use `FormatCents`.  The DLH one was a bit different since it allows up to tenths of a cent (I think it's the only one like that).

pkg/services/ghcrateengine/ghc_rate_engine_service_test.go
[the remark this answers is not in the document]
reggieriser: I haven't tested this at all, but I was thinking of something like this which I think ends up doing mostly the same thing (but consider this optional since it works as is):
```go
func (suite *GHCRateEngineServiceSuite) validatePricerCreatedParams(expectedValues services.PricingDisplayParams, actualValues services.PricingDisplayParams) {
	suite.Equal(len(expectedValues), len(actualValues))

	for _, value := range expectedValues {
		suite.HasDisplayParam(actualValues, value.Key, value.Value)
	}
}
```

reggieriser (commented):
Thanks for the changes.  I think there's one spot where the number of decimal places needs to be adjusted, but otherwise things look good.

I did add a comment with a bit of untested sample code around the `validatePricerCreatedParams` discussion, but consider that optional.

pkg/unit/cents.go
```diff
@@ -60,8 +60,8 @@ func (c Cents) ToDollarString() string {
 	return s
 }
 
-// ToDollarFloat returns a dollar float representation of this value
-func (c Cents) ToDollarFloat() float64 {
+// ToDollarFloatNoRound returns a dollar float representation of this value
+func (c Cents) ToDollarFloatNoRound() float64 {
```
jacquelineIO, line 64: Changed this to make consistent with how `Milicents.ToDollarFloatNoRound` is named. But now not using the function in my PR. But still leaving this change.

pkg/services/ghcrateengine/pricer_formatters.go
```diff
@@ -33,3 +33,8 @@ func FormatBool(value bool) string {
 func FormatFloat(value float64, precision int) string {
 	return strconv.FormatFloat(value, 'f', precision, 64)
 }
+
+// FormatEscalation returns a formatted pricing escalation value to display to the TXO
+func FormatEscalation(value float64) string {
+	return strconv.FormatFloat(value, 'f', 5, 64)
+}
```
reggieriser, line 40: Nice!  I mentioned the possibility of doing this on a recent PR review, so glad to see it implemented and the changes made to other pricers.  I noticed we don't have any tests for it like the other formatters, so might be nice to add a quick table-test driven one similar to `TestFormatFloat`.  That said, this is a simple one-liner function, so not going to block on that.

reggieriser (approved):
LGTM!  I noted a place for a possible simple test, but that's not a blocker.

I didn't test this beyond the unit tests as I couldn't seem to create a DSH even if I made a move between the same zip3s.  If there's something I should try, let me know.

PR #6359, opened 2021-04-07
[MB-7607] Add deleteFromSyncada param to support-reviewed-payment-requests

mboren:
## Description
I was originally hardcoding this value when passing to `NewSyncadaSFTPReaderSession`, but we decided to add a parameter for it so it doesn't require a code change to modify.

## Setup

```sh

```
deletefromsyncada.json
```json
{
  "body": {
    "sendToSyncada": false,
    "readFromSyncada": true,
    "deleteFromSyncada": true
  }
}
```

```sh
prime-api-client --insecure support-reviewed-payment-requests --filename deletefromsyncada.json
```
If you have any payment requests locally that match a response file in syncada, then they should be deleted from Syncada. However, since we can't get response files from syncada right now, we can't test that the deletion is really happening.
We've tested deleting from syncada in previous work though, so I think the most important thing to test is that I didn't break anything and the command still runs.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7607) for this change

### DIFF
--- a/pkg/gen/supportapi/embedded_spec.go
+++ b/pkg/gen/supportapi/embedded_spec.go
@@ -2060,9 +2060,15 @@ func init() {
       "type": "object",
       "required": [
         "sendToSyncada",
-        "readFromSyncada"
+        "readFromSyncada",
+        "deleteFromSyncada"
       ],
       "properties": {
+        "deleteFromSyncada": {
+          "type": "boolean",
+          "x-nullable": true,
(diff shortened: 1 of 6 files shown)


jacquelineIO (approved):
lgtm. Ran locally, looks to run fine. Can't really test but this follows a patter and we've seen deletion work when the test server was running OK. The delete function was not updated, only how you activate it.

PR #6361, opened 2021-04-08
[MB-6828] Adding Pricer Params to Domestic Destination Pricer

shimonacarvalho:
## Description

Updates the Domestic Destination pricer to return a display parameter for 4 param keys.

## Setup

Please follow [A-Team's testing instructions](https://github.com/transcom/mymove/wiki/Acceptance-Testing-Payment-Requests#move-task-order-actions) for creating a new MTO. When you use the TOO interface (or the Support API endpoint) to approve a shipment associated with this move, the default service items should be created. One is DDP or the domestic destination price. 

Note that this change is NOT yet visible in the payment request.
This change is visible in the TIO payment request but only after you approve the payment request. 
This change is also visible in fetch-mto-updates

There should be 4 new params for 
*    ContractYearName
*    PriceRateOrFactor
*    IsPeak
*    EscalationCompounded

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6828) for this change.

### DIFF
--- a/pkg/services/ghcrateengine/ghc_rate_engine_service_test.go
+++ b/pkg/services/ghcrateengine/ghc_rate_engine_service_test.go
@@ -177,7 +177,7 @@ func (suite *GHCRateEngineServiceSuite) setupDomesticLinehaulPrice(serviceAreaCo
 func (suite *GHCRateEngineServiceSuite) HasDisplayParam(displayParams services.PricingDisplayParams, key models.ServiceItemParamName, value string) bool {
 	for _, displayParam := range displayParams {
 		if displayParam.Key == key {
-			return suite.Equal(value, displayParam.Value)
+			return suite.Equal(value, displayParam.Value, "%s param actual value did not match expected", key.String())
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/services/ghcrateengine/ghc_rate_engine_service_test.go
```diff
@@ -177,9 +177,7 @@ func (suite *GHCRateEngineServiceSuite) setupDomesticLinehaulPrice(serviceAreaCo
 func (suite *GHCRateEngineServiceSuite) HasDisplayParam(displayParams services.PricingDisplayParams, key models.ServiceItemParamName, value string) bool {
 	for _, displayParam := range displayParams {
 		if displayParam.Key == key {
-			if displayParam.Value == value {
-				return true
-			}
+			return suite.Equal(value, displayParam.Value)
```
shimonacarvalho, line 180: @reggieriser I changed this because if the key existed but the value didn't match, i was getting the error "Could not find display param". Now it shows the "Not Equal" error with expected and actual which is a bit easier to understand.

reggieriser: Sounds good.  Could we also add a message to the `suite.Equal` (or `suite.Equalf`) that notes the param key name for the value that's not matching?

shimonacarvalho: Good idea, I'll do that.

eamahanna (approved):
Works wells and I was able to see everything in the UI

abbyoung (approved):
LGTM! 🚀

PR #6342, opened 2021-04-06
MB-7465 Create a ContactInfoDisplay component to be used on the new Profile page

felipe-lee:
## Description

Create a `ContactInfo` component that can be used on the new Profile page.

## Reviewer Notes

A few questions:
1. Is the placement in storybook ok?
2. Does the CSS look ok. It's somewhat based on the CSS in `src/components/Customer/Review/ShipmentCard/ShipmentCard.module.scss` and `src/components/Customer/Review/Review.module.scss`

## Setup

1. Run storybook:

    ```sh
    make storybook
    ```

2. Navigate to [Customer Components / Profile / ContactInfo](http://localhost:6006/?path=/story/customer-components-profile-contactinfo--default-state)

## Code Review Verification Steps

* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7465) for this change

#### Without Alt Phone
Screen Shot 2021-04-08 at 16 57 59

#### With Alt Phone
Screen Shot 2021-04-08 at 16 58 50

#### Without Alt Phone
Screen Shot 2021-04-08 at 16 59 49

#### With Alt Phone
Screen Shot 2021-04-08 at 17 00 02

PR #6362, opened 2021-04-08
MB-5881 create profile page

YanZ777:
## Description

Create a Profile page and add it as a route.
Expanded the ServiceInfoTable component so that it would have an edit button and an edit vs no edit mode. The ServiceInfoTable needs to have a no edit mode once the move has been submitted.

## Reviewer Notes
- Should there be something written up to convert the Service Info Table to a definition list? Something has been written! 
[https://dp3.atlassian.net/browse/MB-7732](https://dp3.atlassian.net/browse/MB-7732)

## Setup

```sh
make client_run
make server_run
```

Go to this link [http://milmovelocal:3000/service-member/profile](http://milmovelocal:3000/service-member/profile)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-5881) for this change

## Screenshots

milmovelocal_3000_service-member_profile

milmovelocal_3000_service-member_profile(iPhone 6_7_8 Plus)


src/components/Customer/Review/ServiceInfoTable/index.jsx
```diff
@@ -21,15 +24,28 @@ const ServiceInfoTable = ({
     serviceInfoTableStyles.ServiceInfoTable,
   );
   const tableClassNames = classnames('table--stacked', reviewStyles['review-table']);
+  const editProfilePath = '/moves/review/edit-profile';
```
kimallen, line 27: You can add this path into `constants/routes`

src/pages/MyMove/Profile/Profile.jsx
```diff
@@ -0,0 +1,62 @@
...
+import { connect } from 'react-redux';
+import { push } from 'connected-react-router';
+
+import { OrdersShape, ServiceMemberShape } from 'types/customerShapes';
+import { selectServiceMemberFromLoggedInUser, selectCurrentOrders, selectCurrentMove } from 'store/entities/selectors';
+import SectionWrapper from 'components/Customer/SectionWrapper';
+import ServiceInfoTable from 'components/Customer/Review/ServiceInfoTable';
+
+const Profile = ({ currentOrders, serviceMember }) => {
+  const rank = currentOrders ? currentOrders.grade : serviceMember.rank;
```
kimallen, line 12: Would we expect grade to ever not be on `orders` at this point?  Eventually we will be removing rank from `serviceMember`, so this would break when that happens, though not sure if we are fully there yet

src/scenes/MyMove/index.jsx
```diff
@@ -146,6 +147,7 @@ export class CustomerApp extends Component {
                     path={customerRoutes.SHIPMENT_CREATE_PATH}
                     component={ConnectedCreateOrEditMtoShipment}
                   />
+                  <CustomerPrivateRoute exact path="/service-member/profile" component={Profile} />
```
kimallen, line 150: You can get this path from `constants/routes` as well

kimallen: I have that constant in my Profile link PR, but if it doesn't get merged before yours you can just add it

YanZ777: Sounds good to me!

src/pages/MyMove/Profile/Profile.jsx
YanZ777: I'm not sure. However, if we are going to be removing stuff later, is there a place where this gets kept track?

src/components/Customer/Review/ServiceInfoTable/index.jsx
```diff
@@ -25,11 +30,23 @@ const ServiceInfoTable = ({
     <div className={containerClassNames}>
       <div className={classnames(reviewStyles['review-header'], serviceInfoTableStyles.ReviewHeader)}>
         <h2>Service info</h2>
+        {isEditable && (
+          <Button
+            unstyled
+            className={reviewStyles['edit-btn']}
+            data-testid="edit-profile-table"
+            onClick={() => onEditClick(customerRoutes.EDIT_PROFILE_PATH)}
```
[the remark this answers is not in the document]
suzubara: actually I take that back, this is a `Button` component but should probably be a `Link` (because all it does is navigate to another page)

suzubara (commented):
this looks like a good start! to answer your questions:

> What sort of tests should I be adding?
- ServiceInfoTable doesn't have any tests, IMO it would be worth adding at least one that tests the isEditable states
- Profile page should also have some basic tests that the correct content is rendering, and this way you can build on them to test what happens if a service member in various states of completion visits this page

> The edit button on the service info table isn't working, I think I set this up correctly, but I might have missed something. Maybe this is a problem for these stories?

For the Contact Info edit button, the component is just calling the handler without passing in a path as the parameter.

But also, the handler functions are calling `push` which is the import from `connected-react-router` when it needs to be the `push` prop that gets passed into this component in `mapDispatchToProps`. You may have seen this pattern in other components but tldr:

```
import { push as pushAction } from 'connected-react-router'

const mapDispatchToProps = {
  push: pushAction,
}

const Profile = ({ serviceMember, currentOrders, currentBackupContacts, push }) => {
...
```

then `push` will refer to the function that has been connected via Redux and passed in as a prop, and that should fix the navigation.

> Should there be something written up to convert the Service Info Table to a definition list?

IMO yes :) I think there could be one shared CSS module file that defines the styles for these definition lists and then all these components can be converted to `dl` and use the same CSS. see the `dt, dd, .row` styling in [this file](https://github.com/transcom/mymove/blob/master/src/components/Customer/Review/ShipmentCard/ShipmentCard.module.scss) for a good starting point

> How do I not make things crash when the user tries to access the profile page when they haven't finished filling out their profile?

There should be conditional operator checks on any prop objects that might not exist (ie, `currentStation, currentOrders, currentBackupContacts`) - just to be safe. This is a good thing to add tests for, ie how does this page render at various states of completion. Another option if we want to prevent a user from accessing this page based on their state is to use the [requireCustomerState HOC](https://github.com/transcom/mymove/blob/master/src/containers/requireCustomerState/requireCustomerState.jsx)

src/components/Customer/Review/ServiceInfoTable/index.jsx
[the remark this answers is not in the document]
felipe-lee: Yup, I agree and I'll update mine. So in this case we'll update the lower components to just take an `editUrl" parameter instead of an `onEditClick`  or something like that.

src/components/Customer/Review/ServiceInfoDisplay/ServiceInfoDisplay.jsx
```diff
@@ -0,0 +1,83 @@
...
+  isEditable,
+  lastName,
+  editURL,
+  rank,
+}) => {
+  return (
+    <div className={serviceInfoDisplayStyles.serviceInfoContainer}>
+      <div className={serviceInfoDisplayStyles.header}>
+        <h2>Service info</h2>
+        {isEditable && <a href={editURL}>Edit</a>}
```
felipe-lee, line 23: I think this might need to use the `Link` component instead. @suzubara ? If so, it might require a few tweaks to stories/tests like I did in https://github.com/transcom/mymove/pull/6342 to account for the routing aspect of it.

src/components/Customer/Review/ServiceInfoDisplay/ServiceInfoDisplay.jsx
```diff
@@ -0,0 +1,83 @@
...
+  isEditable: bool,
+  lastName: string.isRequired,
+  editURL: string,
+  rank: string.isRequired,
+};
+
+ServiceInfoDisplay.defaultProps = {
+  currentDutyStationPhone: '',
+  editURL: '',
+  isEditable: true,
```
felipe-lee, line 80: I think either `isEditable` should default to `false`. If `isEditable` defaults to `true`, and `editURL` isn't required, it makes it easy to miss that no url was passed.

src/components/Customer/Review/ServiceInfoDisplay/ServiceInfoDisplay.stories.jsx
```diff
@@ -14,11 +14,17 @@ const defaultProps = {
...
 export default {
-  title: 'Customer Components / ServiceInfoTable',
+  title: 'Customer Components / ServiceInfoDisplay',
 };
 
-export const Basic = () => (
+export const Editable = () => (
   <div style={{ padding: 40 }}>
-    <ServiceInfoTable {...defaultProps} />
+    <ServiceInfoDisplay {...defaultProps} />
```
felipe-lee, line 22: Should this one show an `editURL` being passed?

felipe-lee (commented):
A few comments and a question that may require a few changes depending on the answer.

src/components/Customer/Review/ServiceInfoDisplay/ServiceInfoDisplay.stories.jsx
YanZ777: I left one off since it defaults to nothing. 
I don't know if that was the right move with regards to either leaving it off or defaulting to nothing.

felipe-lee: I think it's fine to default to an empty string since it could possibly be not editable, but I think for the story example that shows it as editable, it should probably have a url passed in rather than using the default of blank. More so like someone looking at the story sees what it would look like code-wise when making it editable.

YanZ777: That makes sense! I'll update it.

src/components/Customer/Review/ServiceInfoDisplay/ServiceInfoDisplay.jsx
suzubara: Yup this should use `<Link>`

src/scenes/MyMove/index.jsx
```diff
@@ -36,6 +36,7 @@ import InfectedUpload from 'shared/Uploader/InfectedUpload';
 import ProcessingUpload from 'shared/Uploader/ProcessingUpload';
 import PpmLanding from 'scenes/PpmLanding';
 import Edit from 'scenes/Review/Edit';
+import Profile from 'pages/MyMove/Profile/Profile';
```
suzubara, line 39: We should start using the `lazy` import pattern for these page components, then they won't be loaded until the user navigates to that route (see lines ~60 of this file):
```
const Profile = lazy(() => import('pages/MyMove/Profile/Profile'));
```

suzubara (approved):
looks good! Nice job updating the `ServiceInfoDisplay` component :D I think the main thing is to change the Edit links to use the `<Link>` component, but otherwise 👍

cosburn:
Looks great on desktop, noticing the spacing is a bit too tight in a couple spots on mobile, screenshots attached. Let me know if you need more info on what that spacing should be. 
IMG_5071
IMG_5072

YanZ777:
> Looks great on desktop, noticing the spacing is a bit too tight in a couple spots on mobile, screenshots attached. Let me know if you need more info on what that spacing should be.

Added some spacing!

amanda5iu (commented):
This is looking good! Just got a couple visual tweaks:

- label should be `DoD ID` - lowercase o
- Looks like there's an extra 16px of bottom margin on `descriptionList` that shouldn't be there.
- There should be 24px of padding between the h2s (Contact info and Service info) and the top of each list. Right now it's 16px.

YanZ777:
> This is looking good! Just got a couple visual tweaks:
> 
> * label should be `DoD ID` - lowercase o
> * Looks like there's an extra 16px of bottom margin on `descriptionList` that shouldn't be there.
> * There should be 24px of padding between the h2s (Contact info and Service info) and the top of each list. Right now it's 16px.

Thanks for reviewing! I've made changes!

amanda5iu:
Hmm I'm still getting extra space on desktop. 
Screen Shot 2021-04-15 at 4 10 41 PM

Padding below `Contact info` and `Service info` should be 24px. Same goes for padding above `Backup contact`. Though the spacing looks fine at smaller breakpoints.

YanZ777:
> Hmm I'm still getting extra space on desktop.
> 
> Padding below `Contact info` and `Service info` should be 24px. Same goes for padding above `Backup contact`. Though the spacing looks fine at smaller breakpoints.

I think I got it right this time! 🤞

YanZ777:
@amanda5iu I also meant to ask, how you measuring these spaces? Is there a browser extension that I can use or are you just eyeballing it?

amanda5iu:
> @amanda5iu I also meant to ask, how you measuring these spaces? Is there a browser extension that I can use or are you just eyeballing it?

I'm using my browser's inspector (Firefox)!

amanda5iu (approved):
Padding on the headers look great! I'm still seeing extra space beneath the contact info and service info tables, but don't wanna hold up the merge on account of 16px. Approving!

YanZ777:
> Padding on the headers look great! I'm still seeing extra space beneath the contact info and service info tables, but don't wanna hold up the merge on account of 16px. Approving!

Thank you! I will write something up for this!

PR #6367, opened 2021-04-08
[MB-3245] Deprecate Review & UI reducers

suzubara:
## Description

This PR removes two Redux reducers that were not effectively being used anymore (or could be replaced with the flash message reducer) - Review & UI. The UI reducer exports & state values were not being used anywhere so that was a straightforward search & delete. The Review state was being used to handle post-submission messaging when editing profile / orders / PPM pages, which is currently being refactored. I replaced those instances with flash messages as well. The affected flows are:
- Service member edits backup contact & sees the "Your changes have been saved." message on the next page
- Service member edits their contact info & sees the "Your changes have been saved." message on the next page
- Service member edits their PPM date & location & sees the "Your changes have been saved." message on the next page
- Service member edits their PPM weight & sees the "Your changes have been saved." message on the next page
- Service member edits their orders & sees the "Your changes have been saved." message on the next page if they didn't change their dependents/spouse has pro gear values
- Service member edits their orders & sees the "Your changes have been saved. Note that the entitlement has also changed." message on the next page if they did change their dependents/spouse has pro gear values
- Service member edits their profile & sees the "Your changes have been saved." message on the next page if they didn't change their rank
- Service member edits their profile & sees the "Your changes have been saved. Note that the entitlement has also changed." message on the next page if they did change their rank

Other changes:
- Display the flash message component on the Review page
- Review Summary checks for the entitlement error (status 409) on mount instead of after form submissions & displays error from state instead of from Redux
- Moved & added tests for `selectEntitlementsForLoggedInUser` selector function. Note there's still a similar function being used by the Office PPM app so I didn't delete the existing one (located in `src/shared/entitlements.js`).

## Reviewer Notes

The catch with this PR is that most of the flows mentioned above are not testable, either because of existing bugs or because they're not currently fully implemented in the app (ie, links to these pages are missing).

**For backup contact (`/moves/review/edit-backup-contact`), contact info (`/moves/review/edit-contact-info`), and PPM weight (`/moves/<move ID>/review/edit-weight`):**
You can access these pages by going directly to the URL, but you won't see the flash message after submitting because the page does a real refresh when it goes back (instead of a browser history push).

**For PPM date & location:**
You can access this page by going to the Review page and clicking "Edit" on the PPM card. However, there's a server error that occurs when saving changes to the PPM which prevents the form from completing the submission flow so you won't be directed back to the Review page to see the flash message.

**For Orders:**
You can access this page by going to the Review page and clicking "Edit" in the Orders section. There are some bugs with this form & saving changes to Orders currently, but you can view the flash message if you _only_ change the "Has dependents?" question and click save.

**For Profile:**
This is the only step that is currently working 100%. You can access this page by going to the Review page and clicking "Edit" in the Profile section. If you make any changes to these values _other than Rank_ you should be brought back to the Review page & see the success flash message. If you make changes to the Rank value, you should be brought back to the Review page & see the entitlement changed message.

> **Why do this work if it's not testable?** Since most of the steps above are about to be refactored, and the PPM flow has existing bugs & needs to be updated, I thought it was still worth making these changes so that the older Redux pattern is not present anywhere. That way when these components are updated, it should be clear what the desired UX is after submitting the forms, and that UX can be included as part of the refactor instead of having to interpret old code to figure out what is supposed to happen.

Finally, you can also test that the **"Your estimated weight is above your entitlement."** warning shows up on the Review page. An easy way to do this is to have a high ranked service member or include dependents in your orders, set your PPM estimated weight to the maximum, and then change your rank to a lower rank or don't include dependents anymore. You should see the warning message at the top of the Review page (this is not a flash message and will show up at all times until you change the weight - see the last screenshot below).

## Setup

To run the app locally:

```sh
make server_run
make client_run
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3245) for this change

## Screenshots

image

image

image


lynzt (approved):
LGTM 🚢 
random question - how do you feel about object destructuring as way to get something similar to named parameters? i'm thinking about this in relation to `setFlashMessage` - just something i thought about when looking at the number of arguments for that function.

suzubara:
> random question - how do you feel about object destructuring as way to get something similar to named parameters? i'm thinking about this in relation to `setFlashMessage` - just something i thought about when looking at the number of arguments for that function.

yeah I agree, `setFlashMessage` started off simple but it's now at the point where I think passing in an object param would simplify things

PR #6368, opened 2021-04-08
[Prosposal] [MB-7627] Remove shorthaul service item

jacquelineIO:
## Invoicing Updates
This PR has been updated to make the linehaul pricer for invoicing be able to handle both shorthaul and regular linehaul pricing. 

Both `ReServiceCodeDSH` and `ReServiceCodeDLH` point to `NewDomesticLinehaulPricer`. Removing `NewDomesticShorthaulPricer`. Tests written for `NewDomesticShorthaulPricer` should simply point to `NewDomesticLinehaulPricer` as the testing will still be valid.

Both `ServiceItemParamNameDistanceZip3` and `ServiceItemParamNameDistanceZip3` point to `DistanceZipLookup`. Removing both `DistanceZip5Lookup` and `DistanceZip3Lookup`. I kept test functions for both and consolidated them to one test file `distance_zip_lookup_test.go` because the existing tests were still valid.

This code also clears up an issue where mileage for HHG was using `< 50` to determine if it was shorthaul or linehaul and this is incorrect. Shorthaul is determined by having the same ZIP3. If it is the same ZIP3, then the ZIP 5 distance function is used. There is a [Jira issue to address this](https://dp3.atlassian.net/browse/MB-3669). However, the `50` mileage indicator is used in SIT distance caluclation.

Having the code in pricing written this way allows invoicing to test shorthaul moves without having to add a `DSH` service item to the move. The pricing will happen based on ZIPs, which is the correct way to determine pricing. The frontend will able continue with the plan to add in shorthaul or they will be able to remove shorthaul and also allow invoicing to remove the shorthaul service item. The backend to opperate the same.

## Possible Furture Updates
If there is a decision made to remove the `DSH`/Domestic Shorthual service item then it can be safely removed from invoicing and the existing functions will still be in use.

With these changes, invoicing could move forward with consolidating service item param `DistanceZip3` and `DistanceZip5` to just be `DistanceZip` or they can stay as-is. Frontend work has already been done to support these params.

To support shorthaul on the frontend, the frontend will have to call out to determine the distance of the move and then determine if it a new HHG shorthaul move or a new HHG linehaul move. This can be combersum if the address changes, then this status could also change. However, invoicing will still price correctly, because invoicing will look at the ZIPs used at the time payment is requested. Invoicing will not be relying on the `HHG_SHORTHAUL_DOMESTIC` vs `HHG_LINEHAUL_DOMESTIC` shipment status.

Screen Shot 2021-04-08 at 12 41 30 PM

## Snapshot of the DSH & DLH service item code and supporting params
Currently we have Domestic Linehaul and Domestic Shorthaul service items. I think using only Domestic Linehual and using a indicator for whether or not this is a shorthaul is sufficient. Shorthaul can be seen as a type of linehaul. While they are both linehauls. And invoicing is able to support both as well. 

Without the changes in this PR, we aren't able to test shorthaul in the current system because it is viewed as a different type of service. Outside of understanding how to do pricing I'm not sure what the `HHG_SHORTHAUL_DOMESTIC` vs `HHG_LINEHAUL_DOMESTIC` status was needed for. There could be a need for the TOO UI to know this information upfront but again, if the address changes then this status could change as well. If the TIO UI needs this information invoicing would be able to set the status for the shipment.

## Reviewer Notes

Additional review for A-Team. [MB-3669](https://dp3.atlassian.net/browse/MB-3669) is also addressed in this PR. 

## Setup

```sh
echo "Code goes here"
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7627) for this change

## Screenshots


pkg/services/ghcrateengine/domestic_linehaul_pricer.go
```diff
@@ -49,27 +51,60 @@ func (p domesticLinehaulPricer) Price(contractCode string, requestedPickupDate t
...
-	domesticLinehaulPrice, err := fetchDomesticLinehaulPrice(p.db, contractCode, isPeakPeriod, distance, weight, serviceArea)
-	if err != nil {
-		return unit.Cents(0), nil, fmt.Errorf("could not fetch domestic linehaul rate: %w", err)
+	var escalatedTotalPrice float64
+	var baseTotalPrice float64
+	var contractYear models.ReContractYear
+	var domesticLinehaulRatePriceMillicents *unit.Millicents
+	var domesticShortHaulRatePriceCents *unit.Cents
+
+	if isShortHaul {
```
jacquelineIO, line 60: These sections could further break down into functions, but this is just a draft proposal so didn't do that here.

tinyels:
Will this have an impact on the prime? (e.g. will they expect a short haul service item). @LeDeep

jacquelineIO:
With the changes in this PR, I'm finding bad `devseed` data. `server_test` are running fine. I think the changes are making more restrictions for regular linehaul to have the right zip codes in use. And not allowing calls to the wrong planner. e.g., if using the same ZIP 3 in a shipment you won't be able to call to the zip3 distance, it will force the call to zip5 distance which is failing because it is not mocked out.

Still need to run test with `make server_run` and `make client_run`. 

Also cleaned up some short haul that were not updated with new code changes.

pkg/payment_request/service_param_value_lookups/distance_zip5_lookup.go
```diff
@@ -50,18 +50,16 @@ func (r DistanceZip5Lookup) lookup(keyData *ServiceItemParamKeyData) (string, er
 	// Now calculate the distance between zip5s
 	pickupZip := r.PickupAddress.PostalCode
 	destinationZip := r.DestinationAddress.PostalCode
-	distanceMiles, err := planner.Zip5TransitDistance(pickupZip, destinationZip)
+	distanceMiles, err := distanceZip(planner, pickupZip, destinationZip)
```
gidjin, line 53: Should we move the lookup code for zip5 into the file same one as distance zip3 since that file has a more generic name now?

pkg/payment_request/service_param_value_lookups/distance_zip_lookup_test.go
```diff
@@ -114,3 +118,54 @@ func (suite *ServiceParamValueLookupsSuite) TestDistanceZip3Lookup() {
...
+		var mtoShipment models.MTOShipment
+		//RA Summary: gosec - errcheck - Unchecked return value
+		//RA: Linter flags errcheck error: Ignoring a method's return value can cause the program to overlook unexpected states and conditions.
+		//RA: Functions with unchecked return values in the file are used fetch data and assign data to a variable that is checked later on
+		//RA: Given the return value is being checked in a different line and the functions that are flagged by the linter are being used to assign variables
+		//RA: in a unit test, then there is no risk
+		//RA Developer Status: Mitigated
+		//RA Validator Status: Mitigated
+		//RA Modified Severity: N/A
+		suite.DB().Find(&mtoShipment, mtoServiceItem.MTOShipmentID) // nolint:errcheck
```
gidjin, line 167: Minor nit, I think we can drop the annotation if we do a `suite.NoError(err)` check after this line and remove the lint exclusion.

pkg/payment_request/service_param_value_lookups/distance_zip5_lookup.go
jacquelineIO: @gidjin I would actually delete this file `distance_zip5_lookup.go`. It's not in use in this PR. I'm chipping away at this, but the next pass would delete this file and delete the short haul pricer file.

gidjin (commented):
I like this, I think we should go for it. That said I'm running into an issue running the migrations and devseed data

```
2021-04-12T21:23:44.189Z        FATAL   scenario/devseed.go:952 Error creating payment request  {"error": "failure pricing service DLH for MTO service item ID 0058aaaf-6d35-4831-9b8a-cf744ba70a4b: could not find param with key DistanceZip5"}
github.com/transcom/mymove/pkg/testdatagen/scenario.createHHGWithPaymentServiceItems
        /Users/john/projects/dod/mymove/pkg/testdatagen/scenario/devseed.go:952
github.com/transcom/mymove/pkg/testdatagen/scenario.devSeedScenario.Run
        /Users/john/projects/dod/mymove/pkg/testdatagen/scenario/devseed.go:3351
main.main
        /Users/john/projects/dod/mymove/cmd/generate-test-data/main.go:209
runtime.main
        /nix/store/v66s6cg9yl0gr1c9iym3ig1pli3g4ngj-go-1.15.10/share/go/src/runtime/proc.go:204
exit status 1
make: *** [db_dev_fresh] Error 1
```

jacquelineIO:
@gidjin yea devseed is failing. We use bad zips in devseed and it is now being caught. Some minor tweaks are needed to the pricer function while we have both ZIP3 and ZIP3 param keys. Then I'll see how devseed is working. ZIP3 and ZIP5 will be used based on which service item is used until we get rid of having both (unless we don't get rid  of having both). So that's one drawback. But as long as the right service item is used it's not a problem. But the pricing will be correct either way.

PR #6370, opened 2021-04-08
[MB-6902] save pricer params in paymentrequestcreator

pearl-truss:
## Description

This PR updates the paymentRequestCreator service so that pricer generated params are included in the output of create-payment-request

## Setup

```sh
prime-api-demo RDY4PY
```
If you go through the `prime-api-demo` flow to create a payment request, you should see 4 params with origin: "PRICER" for DLH service item

image

## Acceptance
In staging find or create a move, ensure that the DLH service item is present and approved. You can use the `prime-api-demo` to create a payment request for the move. Ensure that it has the pricer generated params (see image above) for DLH. Also note, that by the time this is tested, it's possible that more service items other than DLH will populate the pricer generated params. Any service items that've been updated should output their pricer generated params with the cli/prime-api-demo


pkg/services/payment_request/payment_request_creator.go
```diff
@@ -435,6 +438,9 @@ func (p *paymentRequestCreator) createServiceItemParamFromLookup(tx *pop.Connect
 		Value:                 value,
 	}
 
+	if paymentServiceItemParam.ServiceItemParamKey.Origin == models.ServiceItemParamOriginPricer {
```
jacquelineIO, line 441: This check should probably be happening either at the top of this function or before calling this function. The lookup function `ServiceParamValue` does not need to be called for params created by the pricer. I'm also thinking the code won't reach this point, but if it were to this check should be sooner.

jacquelineIO (commented):
Changes look good. I need to run locally before I can approve.

mboren (approved):
LGTM

PR #6373, opened 2021-04-09
[MB-7526] Custom go linter for ATO

Ronolibert:
## Description

Implement custom go linter that checks for disabling of gosec and checks if there is a reason and an approved annotation

## Reviewer Notes
- currently there is a bug that prevents being able to add a custom linter to .golangci.yml but we are using the cmd line utility in the github action so it shouldn't affect us for now
- there are 7 false positives that the linter catches because the #nosec is used in the same line of the code it is disabling. To standardize this, we'd just need the #nosec to be connected to the annotation above it. It's difficult to account for the inline case because inline comments are treated differently than comment blocks. We could fix these in a future card once we are ready to block others

## Setup
```
go install ./cmd/...
ato-linter ./pkg/...
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7526) for this change
* [this article](https://disaev.me/p/writing-useful-go-analysis-linter/) explains go-analysis setup
* [this repo](https://github.com/dbraley/example-linter) is an example linter
* [official linter docs](https://golangci-lint.run/contributing/new-linters/)
* [another helpful tutorial](https://arslan.io/2019/06/13/using-go-analysis-to-write-a-custom-linter/)


testdata/src/linter_tests/ato.go
```diff
@@ -0,0 +1,15 @@
+package linter_tests
+
+// #nosec // want "Disabling of gosec must have an annotation associated with it. Please visit https://docs.google.com/document/d/1qiBNHlctSby0RZeaPzb-afVxAdA9vlrrQgce00zjDww/edit#heading=h.b2vss780hqfi"
+func nosecShouldHaveAnnotation() {}
```
kimallen, line 4: I thought I'd seen a bunch more test cases in here when you reviewed this with me - did they somehow get deleted?  Or am I understanding this wrong?  It seems there are missing test cases

kimallen (commented):
A few questions, otherwise looks good!

pkg/ato-linter/ato.go
```diff
@@ -0,0 +1,128 @@
...
+
+func containsNosec(comments []*ast.Comment) bool {
+	for _, comment := range comments {
+		if strings.Contains(comment.Text, disableNoSec) {
+			individualCommentArr := strings.Split(comment.Text, " ")
+			for _, str := range individualCommentArr {
+				if str == disableNoSec {
+					return true
+				}
+			}
```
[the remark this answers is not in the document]
Ronolibert: You're right! This inner loop for this function is not needed since the first conditional already determines that the noSec is there

pkg/ato-linter/ato.go
```diff
@@ -0,0 +1,128 @@
...
+				if index > 0 && !validatorStatuses[strings.ToUpper(str)] {
+					return true
+				}
+			}
+		}
+	}
+	return false
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
```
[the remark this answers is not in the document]
Ronolibert: From what I've seen, it just seems like convention at this point. I think the interface and error return values here are used for further use in other tooling if we were to integrate it in more complex systems or with golangci-lint. For our purposes, the `pass.Reportf` will also be considered an "error" that the github action uses.

testdata/src/linter_tests/ato.go
Ronolibert: It was just these two! One other case I wanted to test but can't is #nosec with no rule, but there is the limitation of the assertion comment being parsed along with the actual comment that prevents this test case. I think when we were pairing I was just demonstrating how the API worked but these 2 are the main test cases

pkg/ato-linter/ato_test.go
```diff
@@ -0,0 +1,19 @@
...
+	"path/filepath"
+	"testing"
+
+	"golang.org/x/tools/go/analysis/analysistest"
+)
+
+func TestAll(t *testing.T) {
+	wd, err := os.Getwd()
+	if err != nil {
+		t.Fatalf("Failed to get wd: %s", err)
```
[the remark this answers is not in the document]
Ronolibert: Good call! Changing this to errorf since Errorf errors while Fatalf is just logging

pkg/ato-linter/ato.go
```diff
@@ -0,0 +1,123 @@
...
+
+		containsDisablingGosecNoAnnotation := containsGosecNoAnnotation(comments.List)
+		if containsDisablingGosecNoAnnotation {
+			pass.Reportf(node.Pos(), "Disabling of gosec must have an annotation associated with it. Please visit https://docs.google.com/document/d/1qiBNHlctSby0RZeaPzb-afVxAdA9vlrrQgce00zjDww/edit#heading=h.b2vss780hqfi")
+			return
+		}
+
+		containsAnnotationNotApproved := containsAnnotationNotApproved(comments.List)
+		if containsAnnotationNotApproved {
+			pass.Reportf(node.Pos(), "Annotation needs approval from an ISSO")
```
kimallen, line 115: Can we add to the message, "Please add @truss-is3 as reviewers for this PR and ping the ISSO in Slack"

kimallen (approved):
LGTM! Changes look good.
I added a comment for an addon to the "needs approval" annotation message, otherwise 🚢

pkg/ato-linter/ato.go
```diff
@@ -0,0 +1,123 @@
...
+		if strings.Contains(comment.Text, validatorStatusLabel) {
+			// example str: //RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+			individualCommentArr := strings.Split(comment.Text, ": ")
+			// Has validator status label but no value ex. //RA Validator Status:
+			if len(individualCommentArr) == 1 {
+				return true
+			}
+			for index, str := range individualCommentArr {
+				str = strings.Trim(str, " ")
+				if index > 0 && !validatorStatuses[strings.ToUpper(str)] {
```
gidjin, line 76: Won't `str` still have the commas after calling split? I'm not sure I see how this would match since the `validatorStatuses` keys don't have commas. But I might be missing something.

gidjin (approved):
Left a couple questions. Does the bug preventing it from being called in `.golangci.yml` mean that developers wouldn't see violations until the PR builds? If that is the case I would recommend looking to see if there is a newer version of golangci-lint that we could use? Or if we could add it directly to pre-commit so developers would know earlier. Otherwise this looks good to me.

pkg/ato-linter/ato.go
```diff
@@ -0,0 +1,123 @@
...
+
+// ATOAnalyzer describes an analysis function and its options.
+var ATOAnalyzer = &analysis.Analyzer{
+	Name:     "atolint",
+	Doc:      "Checks that disabling of gosec is accompanied by annotations",
+	Run:      run,
+	Requires: []*analysis.Analyzer{inspect.Analyzer},
+}
+
+const disableNoSec = "#nosec"
```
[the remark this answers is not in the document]
kimallen: Yes- we also need to write a custom linter (or amend this one to accommodate) for `staticcheck` and `errcheck`.  We have a story for it.

pkg/ato-linter/ato.go
kimallen: What I'm seeing is that if it doesn't match a `validatorStatus` (ex. the `str` is `{RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}` (but capitalized) instead of `KNOWN ISSUE`, then it will return `true`, that it contains annotation but it is not approved.

testdata/src/linter_tests/ato.go
```diff
@@ -0,0 +1,15 @@
+package linter_tests
```
kimallen, line 1: Is it possible to write a success test (if it does have approved annotation?)

pkg/ato-linter/ato.go
```diff
@@ -0,0 +1,123 @@
...
+
+	inspector.Preorder(nodeFilter, func(node ast.Node) {
+		comments := node.(*ast.CommentGroup)
+		commentsContainNosec := containsNosec(comments.List)
+
+		if !commentsContainNosec {
+			return
+		}
+
+		containsDisablingGosecWithNoReason := containsGosecDisableNoRule(comments.List)
```
kimallen, line 100: This var needs to be changed to `Rule` too

cmd/ato-linter/main.go
```diff
@@ -0,0 +1,11 @@
+package main
+
+import (
+	"golang.org/x/tools/go/analysis/singlechecker"
+
+	atolinter "github.com/transcom/mymove/pkg/ato-linter"
```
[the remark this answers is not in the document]
felipe-lee: Yeah I think that'd be great if we do that in a future step.

PR #6381, opened 2021-04-10
MB-7635 Initial refactoring of MoveRouter

monfresh:
This renames `MoveStatusRouter` to a more generic and meaningful
`MoveRouter`, and renames `RouteMove` to `Submit`, which is what the
`SubmitMoveHandler` now calls, since this happens when the customer
submits their move.

The `Submit` function in the `MoveRouter` then calls internal functions
that determine whether to send the move to the Service Counselor or
TOO. These are internal functions because there aren't any other places
in the app where we would want to directly send a move to a SC or TOO.

I also copied over the other Move state machine functions from the Move
model, but didn't delete them from the model yet, to keep things
working, and to make this initial commit easier to review.

## Description

## Reviewer Notes

## Setup

```sh
echo "Code goes here"
```

## References

## Screenshots


pkg/services/move/move_router.go
```diff
@@ -0,0 +1,211 @@
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// Approve approves the Move
+func (router moveRouter) Approve() error {
```
hannaliebl, line 122: Is it worth noting that this and functions below here aren't being used yet, or not relevant?

pkg/services/move.go
```diff
@@ -23,8 +23,9 @@ type MoveFetcherParams struct {
 	IncludeHidden bool // indicates if a hidden/disabled move can be returned
 }
 
-// MoveStatusRouter is the exported interface for routing moves after customer submission
-//go:generate mockery -name MoveStatusRouter
-type MoveStatusRouter interface {
-	RouteMove(move *models.Move) error
+// MoveRouter is the exported interface for routing moves after customer submission
+//go:generate mockery -name MoveRouter
+type MoveRouter interface {
```
hannaliebl, line 28: I don't know enough about Go to know whether the other methods (Cancel, etc) should be added here.

monfresh: You are right! I was working on that earlier. I just pushed some more changes.

pkg/services/move/move_router.go
monfresh: Good question. I think that might not be necessary because the fact that they're here and not used is not breaking anything. The idea is to make small incremental refactorings that are easy to review. I wrote a bunch of follow-up stories that will start using these functions here instead of the ones in the Move model.

hannaliebl: Makes sense, thanks!

hannaliebl (approved):
LGTM!

pkg/services/move/move_router.go
```diff
@@ -0,0 +1,235 @@
...
+// Service Items unless the Move is approved.
+func (router moveRouter) Approve() error {
+	move := router.move
+
+	if router.approvable() {
+		move.Status = models.MoveStatusAPPROVED
+		return nil
+	}
+	if router.alreadyApproved() {
+		return nil
```
duncan-truss, line 130: I think we might want to not silently do nothing if we try to approve a move that is already approved.  Is this actively happening somewhere?

pkg/services/move/move_router.go
```diff
@@ -0,0 +1,235 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/transcom/mymove/pkg/cli"
+	"github.com/transcom/mymove/pkg/models"
+	"github.com/transcom/mymove/pkg/services"
+)
+
+type moveRouter struct {
+	db   *pop.Connection
+	move *models.Move
```
duncan-truss, line 23: I think it might be better to have the move passed as arguments to the different transitions.  Then this router can be added to the different handlers as a dependency prior to knowing what move it will be operating on like what we do in `pkg/ghcapi/api.go`.  Each handler won't have to ask for a new instance for every move just to call one method.

pkg/services/move/move_router_test.go
```diff
@@ -0,0 +1,53 @@
...
+			{"Canceled", models.MoveStatusCANCELED},
+			{"Needs Service Counseling", models.MoveStatusNeedsServiceCounseling},
+		}
+		for _, invalidStatus := range invalidStatuses {
+			move.Status = invalidStatus.status
+
+			err := moveRouter.Approve()
+
+			suite.Error(err)
+			suite.Contains(err.Error(), "A move can only be approved if it's in one of these states")
```
duncan-truss, line 49: You could add the full expected error message to your struct that holds the invalidStatuses struct instead of looking for substrings.

pkg/handlers/internalapi/moves_test.go
```diff
@@ -192,6 +191,7 @@ func (suite *HandlerSuite) TestShowMoveWrongUser() {
 
 func (suite *HandlerSuite) TestSubmitMoveForApprovalHandler() {
 	suite.Run("Submits ppm success", func() {
+		os.Setenv("FEATURE_FLAG_SERVICE_COUNSELING", "false")
```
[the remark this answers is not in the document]
duncan-truss: Is the env only changed for this test or could it have side effects on other tests?  I can't remember if this persists between tests I vaguely remember encountering this in the past.

pkg/services/move/move_router.go
monfresh: Good question. This is existing behavior. We do the same thing when setting Approvals Requested. It looks like the only places where to try to approve a move are:
- when we make it available to the prime, 
- when we create service items
- when we update service items

And for Approvals Requested, it's when we create or update service items.

pkg/services/move/move_router.go
monfresh: That's what we had originally. You can see the diff where I updated `api.go`. I'm fine either way. Coming from Ruby, passing in the same argument to all the methods in a Class is a smell that the argument should be passed in when initializing an instance of the Class. I'll ask what other Gophers think.

duncan-truss: I'm not sure the same router will ever call more than 1 transition? I think each transition is ultimately followed by a database save, which would update the move's eTag then you would have stale move data in the router.  That's sort of why it feels not quite right to me.

Ryan-Koch (approved):

pkg/handlers/internalapi/moves_test.go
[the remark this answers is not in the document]
monfresh: Yes it's only per test. I confirmed it by moving the test that sets the env var to true above the one that needs it to be false, and it passed. I also set the env var to false in my `.envrc`.

pkg/services/move/move_router_test.go
monfresh: Yeah, but I think that might make it hard to read and quite long, and it would be repeated three times.

pkg/services/move/move_router.go
monfresh: OK, I changed it back. PTAL!

duncan-truss (approved):
Changes look good to me.  May want to include @transcom/truss-pamplemoose  as reviewers as well.

PR #6401, opened 2021-04-13
[MB-7603] Changes to support review app ephemeral deployments

ahobson:
## Description

To support ephemeral deployments, a new environment called `review` has been added. The changes are
* Allow configuring a base domain used to calculate the hostname in the review environment
* Add  `Dockerfile.reviewapp` and `docker-compose.reviewapp.yml` and for building and running the review app locally
* Add `docker-compose.ecs.yml` and `ecs-params.yml.in` for running in AWS ECS via [ecs-cli](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ECS_CLI.html)
* When devlocal auth is enabled, allow non-tls connections to default to the `devlocal` cert for the prime endpoint

In addition, because `Dockerfile.reviewapp` uses multi stage builds and the `dockerfilelint` pre-commit does not support that, switch to `hadolint`

## Reviewer Notes

The changes for the new environment should not introduce any code path changes for the production environment

## Setup

You may need to re-run pre-commit to get the new config.

## References

* [MB-7603](https://dp3.atlassian.net/browse/MB-7603) is the ticket for this change

## Screenshots


travelar (approved):
There's a lot going on here. It looks ok for Infra, but more eyes on this would definitely be good.

docker-compose.ecs.yml
```diff
@@ -0,0 +1,121 @@
...
+      - '4000:4000'
+      - '9443:9443'
+    environment:
+      - CLIENT_AUTH_SECRET_KEY
+      - CSRF_AUTH_KEY
+      - DB_DEBUG=1
+      - DB_ENV=development
+      - DB_HOST=127.0.0.1
+      - DB_NAME=dev_db
+      - DB_PASSWORD=mysecretpassword
```
tinyels, line 64: Is this really ok for a deployed env? Even just for throwaway?

tinyels (commented):
There are a couple of passwords in plain text. I'm not sure how big a deal that is for this sort of work but worth making sure this isn't leaking inadvertently.

docker-compose.ecs.yml
ahobson: I think so because the Postgres db is not accessible except from the application. AFAIK, there's no way to actually use this password unless you have access to deploy something in AWS and at that point, we have bigger fish to fry.

ahobson:
> There are a couple of passwords in plain text. I'm not sure how big a deal that is for this sort of work but worth making sure this isn't leaking inadvertently.

Yeah, this is a good point. I'm just using the same plaintext password used in the dev environment, so those passwords are already in `Makefile` and various existing `docker-compose.*.yml` files

monfresh:
From the app side, I believe we'll need a file in `/config/env/review.app.env` so we can define env vars there. Take a look at https://github.com/transcom/mymove/blob/master/config/env/stg.app.env as an example.

ahobson:
> From the app side, I believe we'll need a file in `/config/env/review.app.env` so we can define env vars there. Take a look at https://github.com/transcom/mymove/blob/master/config/env/stg.app.env as an example.

Fascinating. According to [config/README.md](https://github.com/transcom/mymove/blob/master/config/README.md) those are used by the `ecs-deploy task-def` tool, but the ephemeral environments aren't deployed that way. Instead, all the environment variables are in the `docker-compose.ecs.yml` file.

Is there a better place for me to explain that?

monfresh:
Ah, got it. Can we add `FEATURE_FLAG_SERVICE_COUNSELING=true` to the Docker env vars? I'm assuming we want that turned on right now to test the Service Counseling feature.

ahobson:
> Ah, got it. Can we add `FEATURE_FLAG_SERVICE_COUNSELING=true` to the Docker env vars? I'm assuming we want that turned on right now to test the Service Counseling feature.

Yeah! I'll update this PR to include that.

ahobson:
> > Ah, got it. Can we add `FEATURE_FLAG_SERVICE_COUNSELING=true` to the Docker env vars? I'm assuming we want that turned on right now to test the Service Counseling feature.
> 
> Yeah! I'll update this PR to include that.

Actually, I figured out how to use a file when deploying the ephemeral environments and I hope that's easier now that all the environment vars are in a single dir for folks to manage. Please take a look.

docker-compose.reviewapp.yml
```diff
@@ -0,0 +1,84 @@
...
+      context: .
+      dockerfile: Dockerfile.reviewapp
+      target: migrate
+      args:
+        GIT_BRANCH: ${GIT_BRANCH:-unknown}
+        GIT_COMMIT: ${GIT_COMMIT:-unknown}
+    links:
+      - database_review
+    env_file:
+      - config/env/review.app.env
```
monfresh, line 32: 🎉

monfresh (approved):
Looks great. Thanks!

cmd/milmove/serve.go
```diff
@@ -819,7 +819,12 @@ func serveFunction(cmd *cobra.Command, args []string) error {
 		primeMux := goji.SubMux()
 		primeDetectionMiddleware := auth.HostnameDetectorMiddleware(logger, appnames.PrimeServername)
 		primeMux.Use(primeDetectionMiddleware)
-		primeMux.Use(clientCertMiddleware)
+		if v.GetBool(cli.DevlocalAuthFlag) {
```
shimonacarvalho, line 822: Hey Drew, can you check that this setting is actively off in all other environments? I know it obviously should be but we've run into some issues in the past where the env flags were inconsistent or relying on defaults.

shimonacarvalho (approved):
Looks good, please just ensure that flag is not accidentally or defaults to on in other environments.

cmd/milmove/serve.go
ahobson: If I go to `https://my.stg.move.mil/sign-in` and `https://my.move.mil/sign-in` I don't see the devlocal login link. I don't have an account on any environment, so I don't know if you had some other check in mind.

YanZ777 (approved):

PR #6420, opened 2021-04-14
MB-7361 Services counseling move allowances details

donaldthai:
## Description

> As a services counselor, I want to be able to view the allowance details for a move, so I can make validate the information with the service member.   
> 
> ### Acceptance criteria:
> 
> GIVEN I am a services counselor
> 
> When I am on the move details page
> 
> THEN I am able to view the following allowance details:
> 
> - Branch, rank
> 
> - Weight allowance
> 
> - Authorized weight
> 
> - Storage in transit
> 
> - Dependents
> 
> - Pro-gear
> 
> - Spouse pro-gear
> 
> - RME (required medical equipment) - NEW
> 
> - OCIE (organizational clothing and individual equipment) - NEW

## Reviewer Notes

- Changes to `AllowancesTable.jsx` also affects the TXO side. Fields were re-arranged and new fields added.
- New columns added to the `entitlements` table: `required_medical_equipment_weight` and `organizational_clothing_and_individual_equipment`.

## Setup

```sh
make db_dev_migrate
make db_dev_e2e_populate
make server_run
make office_client_run
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7361) for this change
* [Zeplin](https://app.zeplin.io/project/6058ba7eb92cf7391a455943/screen/60678a9e1886e094520d7187) design mockup

## Screenshots

image


migrations/app/schema/20210412213134_add_entitlments_columns.up.sql
```diff
@@ -0,0 +1,15 @@
+-- add two new columns to entitlements table
+ALTER TABLE entitlements
+    ADD COLUMN required_medical_equipment_weight int4;
```
Ryan-Koch, line 3: With these we have a default value that we're going to want to have until the services counselor wants to change them. There's a couple ways we could approach this, but I did notice that postgres does have the ability to define default values: https://www.postgresql.org/docs/9.3/ddl-default.html. If that's an approach you think makes sense then we might be able to do something like this:
`ADD COLUMN required_medical_equipment_weight in4 DEFAULT 0` (it's possible I'm off a bit on the syntax, but maybe worth playing with).

migrations/app/schema/20210412213134_add_entitlments_columns.up.sql
```diff
@@ -0,0 +1,15 @@
...
+COMMENT ON COLUMN entitlements.required_medical_equipment_weight IS 'The RME (required medical equipment) weight in pounds. A Service member or a dependent who is entitled to, and
+receiving, medical care authorized by 10 U.S.C. §1071-§1110. may ship medical equipment necessary for such care. The medical equipment may be shipped in the same way as HHG, but has no weight limit.
+The weight of authorized medical equipment is not included in the maximum authorized HHG weight
+allowance.
+1. Required medical equipment does not include a modified personally owned vehicle.
+2. For medical equipment to qualify for shipment under this paragraph, an appropriate.
+Uniformed Services healthcare provider must certify that the equipment is necessary for medical
+treatment of the Service member or the dependent who is authorized medical care under.';
+ALTER TABLE entitlements
+    ADD COLUMN organizational_clothing_and_individual_equipment bool;
```
Ryan-Koch, line 13: Similar idea here as my above comment, though it'd need to be for the boolean value. So maybe something like `ADD COLUMN organizational_clothing_and_individual_equipment bool DEFAULT false NOT NULL`

migrations/app/schema/20210412213134_add_entitlments_columns.up.sql
monfresh: Agreed. Here's an example from one of our migrations:
```
ADD COLUMN severity INT NOT NULL DEFAULT 0;
```

migrations/app/schema/20210412213134_add_entitlments_columns.up.sql
```diff
@@ -0,0 +1,15 @@
...
+COMMENT ON COLUMN entitlements.organizational_clothing_and_individual_equipment IS 'A yes/no field reflecting whether the customer has OCIE (organizational clothing and individual equipment) that will need to be shipped as part of their move. Government property issued to the Service
+member or employee by an Agency or Service for official use. A term specific to the Army and not other services.';
```
monfresh, line 15: It looks like this file is missing an empty newline at the end. Not sure why our `.editorconfig` is not catching these. Maybe Goland is not supported? Is that what you're using? I don't see it listed here: https://editorconfig.org

In any case, it looks like there is a setting for this in Goland, and it's turned off by default:
Preferences -> Editor -> General -> On Save -> Ensure every saved file ends with a line break

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -195,18 +195,24 @@ func Entitlement(entitlement *models.Entitlement) *ghcmessages.Entitlements {
 	if entitlement.TotalDependents != nil {
 		totalDependents = int64(*entitlement.TotalDependents)
 	}
+	var requiredMedicalEquipmentWeight int64
+	if entitlement.RequiredMedicalEquipmentWeight != nil {
```
monfresh, line 199: Given these can't be nil, we won't need this anymore, right?

pkg/models/ghc_entitlements.go
```diff
@@ -14,11 +14,13 @@ type Entitlement struct {
...
-	DBAuthorizedWeight *int             `db:"authorized_weight"`
-	weightAllotment    *WeightAllotment `db:"-"`
-	StorageInTransit   *int             `db:"storage_in_transit"`
-	CreatedAt          time.Time        `db:"created_at"`
-	UpdatedAt          time.Time        `db:"updated_at"`
+	DBAuthorizedWeight                           *int             `db:"authorized_weight"`
+	weightAllotment                              *WeightAllotment `db:"-"`
+	StorageInTransit                             *int             `db:"storage_in_transit"`
+	RequiredMedicalEquipmentWeight               *int             `db:"required_medical_equipment_weight"`
+	OrganizationalClothingAndIndividualEquipment *bool            `db:"organizational_clothing_and_individual_equipment"`
```
monfresh, line 21: These last 2 should not be pointers anymore once we set the default values in the DB.

migrations/app/schema/20210412213134_add_entitlments_columns.up.sql
duncan-truss: I thought the mocks want us to highlight in the allowances table when these values aren't filled in.  It looked like we might display default values in the UI if they are NULL but how would we know they need updated with those database defaults? https://zpl.io/annlDeY

migrations/app/schema/20210412213134_add_entitlments_columns.up.sql
donaldthai: Yea, I'm using Goland. I'll see if I can turn on that default.

migrations/app/schema/20210412213134_add_entitlments_columns.up.sql
donaldthai: So essentially, we want to use the zero values to determine if it's not set or not?

Ryan-Koch: Eventually we'll want to highlight when we're on the default values. I think we could infer that state since we know what these default values are the default values just as we could by assuming null means default. I'm of the inclination that it'd be prudent to represent the real values in the DB in case later processes wind up needing them later for things. If that is indeed the case we'd have to carry that inference through the rest of the way to keep things consistent, which might be more complicated to keep track of.

duncan-truss: Once we have an edit form though, I'm not sure how you'd tell the difference between a default integer and default boolean value and the ones that get updated to those same unless the validations don't allow those values.

Ryan-Koch: I suspect that the SC wouldn't update the values to the default values. I suppose if they do I think for now I'd be okay with us having the visual cue there still. At that point the SC would be accepting that the default value works for that case and they would move on and ignore the cue.

migrations/app/schema/20210412213134_add_entitlments_columns.up.sql
```diff
@@ -0,0 +1,17 @@
...
+allowance.
+1. Required medical equipment does not include a modified personally owned vehicle.
+2. For medical equipment to qualify for shipment under this paragraph, an appropriate.
+Uniformed Services healthcare provider must certify that the equipment is necessary for medical
+treatment of the Service member or the dependent who is authorized medical care under.';
+ALTER TABLE entitlements
+    ADD COLUMN organizational_clothing_and_individual_equipment bool
+        DEFAULT FALSE NOT NULL;
+COMMENT ON COLUMN entitlements.organizational_clothing_and_individual_equipment IS 'A yes/no field reflecting whether the customer has OCIE (organizational clothing and individual equipment) that will need to be shipped as part of their move. Government property issued to the Service
+member or employee by an Agency or Service for official use. A term specific to the Army and not other services.';
```
monfresh, line 17: I'm going through the PR review checklist and was wondering if you pinged #g-database for a review.

pkg/handlers/ghcapi/orders_test.go
```diff
@@ -49,6 +49,8 @@ func (suite *HandlerSuite) TestGetOrderHandlerIntegration() {
 	suite.Equal((*order.EntitlementID).String(), payloadEntitlement.ID.String())
 	orderEntitlement := order.Entitlement
 	suite.NotNil(orderEntitlement)
+	suite.Equal(1000, orderEntitlement.RequiredMedicalEquipmentWeight)
+	suite.Equal(true, orderEntitlement.OrganizationalClothingAndIndividualEquipment)
```
monfresh, line 53: Did you mean to compare the payload value with the order's value?
```golang
suite.Equal(orderEntitlement.RequiredMedicalEquipmentWeight, payloadEntitlement.RequiredMedicalEquipmentWeight)

suite.Equal(orderEntitlement.OrganizationalClothingAndIndividualEquipment, payloadEntitlement.OrganizationalClothingAndIndividualEquipment)
```

src/pages/Office/ServicesCounselingMoveDetails/ServicesCounselingMoveDetails.jsx
```diff
@@ -41,6 +42,21 @@ const ServicesCounselingMoveDetails = () => {
...
+    weightAllowance: allowances.totalWeight,
+    authorizedWeight: allowances.authorizedWeight,
+    progear: allowances.proGearWeight,
+    spouseProgear: allowances.proGearWeightSpouse,
+    storageInTransit: allowances.storageInTransit,
+    dependents: allowances.dependentsAuthorized,
+    requiredMedicalEquipmentWeight: allowances.requiredMedicalEquipmentWeight,
+    organizationalClothingAndIndividualEquipment: allowances.organizationalClothingAndIndividualEquipment,
+  };
+
```
monfresh, line 59: I noticed we have an `allowancesInfo` constant in `src/pages/Office/MoveDetails/MoveDetails.jsx`, but I'm not sure if it's being used because it wasn't changed in this PR, yet I can see the RME and OCIE info on the Move details page for a TOO (not a service counselor). How does that work? Is it picking up the constant defined here?

lynzt (approved):
looked at the migration only...
LGTM. 🚢 
Will we need to filter or search on either of these fields? If so, we might want an index

src/pages/Office/ServicesCounselingMoveDetails/ServicesCounselingMoveDetails.jsx
duncan-truss: Yea I think we need to map the keys in the MoveDetails allowancesInfo object as well, right now it shows the defaults 0 lbs and No because it's always null in allowancesInfo even though they have values in the payload

donaldthai: oops, totally missed that. Will map the values now.

pkg/handlers/ghcapi/orders_test.go
donaldthai: No, I meant to do the actual values but maybe it does make more sense to compare with the payload value.

amanda5iu (commented):
I got some copy feedback for the OCIE value - can we change the values to `Authorized`/`Unauthorized`? It's more explicit than a simple yes/no.

The container padding looks off - should be 4 units/32px top and bottom.

Also, I believe @hannaliebl 's PR changes the `Allowances` header size from h4 to h2. Just a heads in case that impacts styling changes!

pkg/handlers/ghcapi/orders_test.go
monfresh: Yeah, this is a handler test that makes sure the payload includes those new fields. If we just checked that the values exist on the Order, we wouldn't be testing the payload. The test would still pass if we remove the values from the payload.

amanda5iu (approved):
Looking good 🤩

monfresh (approved):
LGTM. Thanks for making the changes.

PR #6421, opened 2021-04-14
MB-413-admin-can-add-sc-role-to-officeuser

NamibiaTorres:
## Description

Add services counselor role for office users (in the model) and run a migration with an update to the data base to reflect the new role.

Note this ticket says to update the handler and handler tests if needed. It was not needed for the office user handler.

## Reviewer Notes
Check that the migration was successful

## Setup
In postman, got to the Admin API folder and under the Office User folder, go to PATCH update office user. In the body, make sure to put  the values for the following fields, including the role that you want updated:

```json
{
    "firstName": "<string>",
    "middleInitials": "<string>",
    "lastName": "<string>",
    "telephone": "<telephone>",
    "active": "<boolean>",
    "roles": [
        {
            "name": "<string>",
            "roleType": "<string>"
        },
        {
            "name": "<string>",
            "roleType": "<string>"
        }
    ]
}
```
In the response you should see the updated role Service Counselor if the request was successful.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4136) for this change


migrations/app/schema/20210407223734_add_service_counselor_role.up.sql
```diff
@@ -0,0 +1,6 @@
+-- Inserting a new role type and role name into the roles table
+COMMENT ON COLUMN roles.role_type IS 'These are the names of the roles in snake case.';
+
+INSERT INTO roles (id, role_type, created_at, updated_at, role_name)
+VALUES
+(uuid_generate_v4(), 'services_counselor', now(),now(),'Services Counselor');
```
shimonacarvalho, line 6: @sandy-wright Did you recently write up some guidance on not using uuid_generate_v4?

sandy-wright: yeah, the recommendation is to generate a constant UUID instead: https://github.com/transcom/mymove/wiki/Database-Migrations#writing-migrations

NamibiaTorres: Oh no, looks like my push didn't go through! I changed this value in my most recent commit.

NamibiaTorres: Just pushed it up @shimonacarvalho

NamibiaTorres: > yeah, the recommendation is to generate a constant UUID instead: https://github.com/transcom/mymove/wiki/Database-Migrations#writing-migrations

I made the change in my last commit, but hadn't pushed it up. The update should be reflected now.

migrations/app/migrations_manifest.txt
```diff
@@ -621,4 +621,5 @@
 20210331135106_rename_edi_processing_table.up.sql
 20210405214216_update_move_status_comment_with_miro_link.up.sql
 20210405224456_add_service_counseling_completed_at_to_moves.up.sql
+20210407223734_add_service_counselor_role.up.sql
 20210409170358_make_icn_column_nullable.up.sql
```
sandy-wright, line 625: This is a question for @transcom/truss-db - is it good practice to make sure your new migration is listed last in the manifest? I have it in my head that there can be issues with applying a new migration that comes before a previously applied migration, but I'm not sure that's actually true. I tested this locally and there doesn't seem to be a problem with it, so do we not have to worry about the migration order in the manifest?

migrations/app/schema/20210407223734_add_service_counselor_role.up.sql
```diff
@@ -0,0 +1,6 @@
+-- Inserting a new role type and role name into the roles table
+COMMENT ON COLUMN roles.role_type IS 'These are the names of the roles in snake case.';
+
+INSERT INTO roles (id, role_type, created_at, updated_at, role_name)
+VALUES
+('010bdae1-8ebe-44c9-b8ee-8c4477fae2a6', 'services_counselor', now(),now(),'Services Counselor');
```
sandy-wright, line 6: This is one heck of a nitpick, but would you mind adding spaces after the last two commas?

sandy-wright (approved):
LGTM! I was able to add the service counselor role using the Admin API, huzzah! I left one suggestion and one question for the DB reviewers, but other than that, it looks good to go

migrations/app/migrations_manifest.txt
NamibiaTorres: Also, more info for context. The migration was placed in this position after resolving a merge conflict with master.

reggieriser: Our convention has been that migrations should be placed in chronological order based on the timestamp in the filename.  There's a pre-commit hook that should ensure this.

However, I think @sandy-wright is correct in that we could have this scenario: 1) person A creates a migration on their branch that has a timestamp of 3 days ago, 2) person B creates a migration on their branch that has a timestamp of 2 days ago, 3) person B merges their branch to master, 4) everything is fine, 5) person A merges their branch to master and it appears in the manifest before person B's since it has an earlier timestamp, 6) the two migrations conflict (say, they're both trying to add or remove the same column).  That said, I'm not sure I recall ever seeing this in practice on the project, though.  One reason is that a merge conflict on the manifest would likely occur since these are both added toward the end of the file, so person A may be forced to resolve the conflict in their branch before merging.  Unless we force a branch to always be 100% up-to-date with master before merging, I think this has the possibility of happening (just like a code change could cause something to break even though there are no merge conflicts).

Also keep in mind that an existing database will only have migrations applied that it doesn't already have.  So, the order that migrations are applied in an existing database might be slightly different than how they're applied in a fresh database where all migrations are being executed in the order they appear in the manifest.

Other DB folks may have some additional context, but my take is the manifest in this PR is acceptable as long as we're careful.  Alternatively, you could always generate a new migration with a later timestamp (before merging) and use that instead -- but that still doesn't eliminate the possibility that something else has slipped in right before your merge.

reggieriser (approved):
The migration LGTM (note that I only looked at the migration on behalf of the DB reviewers).

migrations/app/migrations_manifest.txt
sandy-wright: Yay, thanks @reggieriser! That context is super helpful, and I forgot that migrations might have already been applied in a different order than they're listed in the manifest. I'll keep this info in mind for the future 😄

NamibiaTorres: Thanks @reggieriser and @sandy-wright ! Definitely good information and context to have before merging. Appreciate the thorough discussion.

NamibiaTorres:
Create a migration to add the Service Counselor role for Office Users.

PR #6422, opened 2021-04-14
[MB-247] Refactor Header components

suzubara:
## Description

This PR does some refactoring to the Header components used by the Customer (MyMove) app & Office app. The new component structure is:

- All Header components render the `<MilMoveHeader />` component (which renders the ReactUSWDS header with some significant CSS customization). Navigation and other instance-specific contents can be passed in using `props.children`.
- I created "subcomponents" for the `UserInfo` block in each header type -- this is the contents on the far-right of the header (everything to the right of the border) that is user-specific, and contains the login/logout links and user info.
  - There are 3 subcomponents for this content: `CustomerUserInfo, OfficeUserInfo, LoggedOutUserInfo`. Each one is viewable in the MilMove Header stories
- I also created container components for each application instance of the header: `CustomerLoggedInHeader, OfficeLoggedInHeader, LoggedOutHeader`
  - I made these as containers because they each connect to Redux in order to select relevant data such as the user data to render the correct header contents, and to handle the logout action
 
Changes in the application code are:
- Both apps (`MyMove/index, Office/index`) render the same `<LoggedOutHeader />` component if the user is not logged in
  - This header renders the login button & handles the EULA modal behavior. It does _not_ render the dev local login button, which has moved to the Sign In page.
- The Customer app renders the `<CustomerLoggedInHeader />` component on all pages if the user is logged in
  - This header renders the logout button and the profile link if the user's profile is complete
- Office app renders the `<OfficeLoggedInHeader />` component on pages that display a header if the user is logged in (regardless of their role)
  - This header renders the queue link based on the office user's role & gbloc info, the office user's name, and the logout button
  - The only material change here is that PPM Office users will see the new header now instead of the old header, but that seems ok since functionally they aren't different. But I would like to confirm this is ok with @transcom/truss-teen-vogue & @transcom/truss-design 

## Reviewer Notes

Test coverage!

## Setup

```sh
make client_run
make server_run
make storybook
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-247) for this change

## Screenshots

Customer: 
milmovelocal_3000_sign-in
milmovelocal_3000_sign-in(iPhone 6_7_8 Plus)

image
image

Office:
officelocal_3000_sign-in
officelocal_3000_sign-in(iPhone 6_7_8 Plus)

image
image

Admin:
adminlocal_3000_system_webhook_subscriptions_46511663-5f18-4c34-9b1f-2d69b52490a3


Shkeating (approved):
approving, but there is some work needed to create the correct amount of space between the profile icon the link that follows it. I also am not sure what design's intended hover style for these links is, but from what I recall this might not be what was intended. I will create a separate ticket for this and sort it out on my own though, as it seems out of scope for what this branch is for, and based off my initial assessment of the work necessary to do this, would be too involved to tack onto this branch.

kimallen (commented):
I noticed in the office app inidividual movepage there's a white line (border or gap) between the Customer Header and the Milmove Header
image

In stg, it does not have that line:
image

Not sure if that's something to align here or in the PR that @Shkeating might write

src/components/MilMoveHeader/LoggedOutUserInfo.jsx
```diff
@@ -0,0 +1,25 @@
...
+        <li className="usa-nav__primary-item">
+          <Button unstyled aria-label="Sign In" onClick={handleLogin} data-testid="signin" type="button">
+            Sign in
+          </Button>
+        </li>
+      </ul>
+    </div>
+  );
+};
+
```
eamahanna, line 20: Does this component have a corresponding test file?

eamahanna: I saw we are testing it some in milmoveHeader test file. Is that enough or do we try and test each component individually?

src/components/MilMoveHeader/OfficeUserInfo.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { string, func } from 'prop-types';
+import { Button } from '@trussworks/react-uswds';
+
+import styles from './index.module.scss';
+
+const OfficeUserInfo = ({ handleLogout, firstName, lastName }) => {
+  return (
+    <div className={styles.userInfo}>
```
eamahanna, line 9: Similar question to above, should this have it's own test file?

src/components/MilMoveHeader/index.jsx
```diff
@@ -1,44 +1,32 @@
...
-      <div className={styles.verticalLine} />
-      {lastName !== '' && firstName !== '' && (
-        <span>
-          {lastName}, {firstName}
-        </span>
-      )}
-      <Button unstyled className={styles.signOut} onClick={handleLogout} type="button">
-        Sign out
-      </Button>
+const MilMoveHeader = ({ children }) => (
```
eamahanna, line 9: 🎉🎉🎉🎉🎉🎉

eamahanna (commented):
LGTM, I just had a couple of questions about tests.

YanZ777:
> I noticed in the office app inidividual movepage there's a white line (border or gap) between the Customer Header and the Milmove Header

Ok I fixed it, but I feel a little gross doing so. I had to put in an `!important` because for the life for me I could not figure out the selectors for it. I suspect I can not because the style is coming from `headers.scss`, which is not something that we control. It looks like those styles get put in last as well, so anything I do put gets overridden, hence the `!important`.

src/scenes/Office/office.scss
```diff
@@ -502,3 +502,7 @@ head .panel-subhead {
 .editable-panel-column .missing-info-alert {
   margin-bottom: 20px;
 }
+
+header.nav-header {
+  border-bottom: 1px solid #dfe1e2;
```
jenniferlynparsons, line 507: Is there a variable for this hex color?

jenniferlynparsons (approved):
Tentative approval with a need for follow up work to refine the tests.

I agree with @eamahanna that the unit tests should be broken out into separate test files. Currently, the tests on MilMoveHeader, OfficeUserInfo, etc. read like integration tests rather than unit test.

My suggestion is to create individual unit test files and also create a cypress test for the header sign in/sign out flow. This might mean making the homepage cypress test more robust or creating a whole new test.

src/scenes/Office/office.scss
YanZ777: Yes, there is! And I changed it to use that instead. Thanks for catching this.

YanZ777:
> Tentative approval with a need for follow up work to refine the tests.
> 
> I agree with @eamahanna that the unit tests should be broken out into separate test files. Currently, the tests on MilMoveHeader, OfficeUserInfo, etc. read like integration tests rather than unit test.
> 
> My suggestion is to create individual unit test files and also create a cypress test for the header sign in/sign out flow. This might mean making the homepage cypress test more robust or creating a whole new test.

@jenniferlynparsons I will write up a new ticket to put in more robust testing for the various headers

If everything else is ok, I will be hitting merge.

Ronolibert (approved):
Everything looks good! Looks like a lot easier of an API to use and more flexible. I do agree that we can use more testing on it for the slightly different header types

Shkeating (approved):
I am going to open a ticket for myself to look at that gap Kim pointed out and the spacing between the items on the right, I think this is good to go as is for the scope of its ticket!

PR #6424, opened 2021-04-14
MB-7585 Use PaymentRequestNumber instead of ReferenceID for looking up 824

gidjin:
## Description

Originally we were using the MTO Reference IDs in our 858s when sending them to Syncada. This switched to the PaymentRequestNumber as it was more unique but in turn the 824 processor needed to also do that as well. This PR fixes the 824 processor to use the Payment Request Number.

## Reviewer Notes

Do the tests look good. Especially `doesn't update a payment request status after processing an invalid EDI824` on Lines 185 - 207

## Setup

This is called by the job runner but until there is a connection to GEX we can't test this in staging. Locally you can ensure the tests run, but similarly calling the job runner will not work.

```sh
make server_test

# call the job runner if you want
go run ./cmd/milmove-tasks process-edis
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7585) for this change

### DIFF
--- a/pkg/services/invoice/process_edi824.go
+++ b/pkg/services/invoice/process_edi824.go
@@ -114,17 +114,13 @@ func (e *edi824Processor) ProcessFile(path string, stringEDI824 string) error {
 			return fmt.Errorf("unable to find move with associated payment request: %w", err)
 		}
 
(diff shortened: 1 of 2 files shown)


pearl-truss (approved):
LGTM

PR #6427, opened 2021-04-15
[MB-7623] Switch the connection for the Job Runner to use HTTPS to GEX

mboren:
## Description

I took the GEX connection code from `post_file_to_gex.go`, updated it a bit, and used it to change places where we were using SFTP to send files to Syncada to use HTTP to send to GEX instead. 
I left in the ability to switch back to using SFTP to send to Syncada. If we determine that we will not need this in the future, we should add a story to remove it.

I also added an ICN to the filenames that we send, to avoid having a file name collision if we sent multiple files at the same second. The new format is: `<timestamp>_<ICN>_edi858.txt`.

## Setup
I haven't had any luck running this locally yet, so we may have to wait until it lands in staging to actually test it.

To test this, we need to get a move with a payment request in the REVIEWED state.
This can be done by creating a move, approving as TIO, creating a payment request, and then approving as the TIO.
In staging, the job runner should automatically run every 5-15 minutes (not 100% sure).
You should be able to use fetch-mto-updates to see the payment request status. It should be REVIEWED after the TIO approves it, and should change to SENT_TO_GEX after this job runs.

You can also trigger the job manually with:

```sh
go run ./cmd/milmove-tasks process-edis --send-to-syncada true
```

Or you can call the support api endpoint instead:
sendtosyncada.json
```json
{
  "body": {
    "sendToSyncada": true,
    "readFromSyncada": false,
    "deleteFromSyncada": false
  }
}
```

```sh
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7623) for this change


gidjin (changes requested):
Had a few suggestions, marking this as changes requested because the gex sender isn't respecting the `SEND_TO_SYNCADA` flag (aka `sendEDIFile` parameter)

pkg/services/payment_request/payment_request_reviewed_processor_test.go
```diff
@@ -526,6 +535,140 @@ func (suite *PaymentRequestServiceSuite) TestProcessReviewedPaymentRequest() {
...
+			gexSender,
+			sftpSender)
+
+		err = paymentRequestReviewedProcessor.ProcessReviewedPaymentRequest()
+		suite.Contains(err.Error(), "error sending the following EDI")
+
+		// Ensure that sent_to_gex_at is Nil on unsuccessful call to processReviewedPaymentRequest service
+		fetcher := NewPaymentRequestFetcher(suite.DB())
+		paymentRequest, _ := fetcher.FetchPaymentRequest(pr.ID)
+		suite.Nil(paymentRequest.SentToGexAt)
```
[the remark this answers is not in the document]
mboren: Ok, I added some checks to make sure it's set to EDI_ERROR on failures and SENT_TO_GEX on successes.

jacquelineIO (approved):
Changes look good to me. There are valid UT, just need to run it to see what happens.

pkg/services/invoice/gex_sender_http_test.go
```diff
@@ -42,7 +42,7 @@ func (suite *GexSuite) TestSendToGexHTTP_Call() {
 	mockServer = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusInternalServerError)
 	}))
-	resp, err = NewGexSenderHTTP(mockServer.URL, false, nil, "", "").
+	resp, err = NewGexSenderHTTP(mockServer.URL, "", false, nil, "", "").
```
gidjin, line 45: Small nit, could use a test that verifies the channel and file name end up in the url. But only if it's easy.

gidjin (approved):
Approving was able to connect to GEX via HTTPS locally

pkg/services/invoice/gex_sender_http_test.go
mboren: OK i added a test for this.

PR #6406, opened 2021-04-13
[MB-7624] update process reviewed payment requests to use real GEX HTTP Sender

jacquelineIO:
## Description

This PR updates the MilMove Invoice ECS scheduled tasks (AKA jobrunner) for running the `ProcessReviewedPaymentRequests` service object to call the GEX HTTP sender service.

## Reviewer Notes

The code for sending to Syncada via SFTP (using the Syncada test server) is still present. 

UT is using mocks for the GEX HTTP sender because we don't want to send to GEX from UT.

## Setup

Add any steps or code to run in this section to help others prepare to run your code:

```sh
echo "Code goes here"
```

#### Run targeted UT with 

```sh
go test ./pkg/services/payment_request -testify.m ReviewedPaymentRequest -v
```

## Code Review Verification Steps

* [ ] If the change is risky, it has been tested in experimental before merging.
* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#logging)
* [ ] The requirements listed in [Querying the Database Safely](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#querying-the-database-safely) have been satisfied.
* Any new migrations/schema changes:
  * [ ] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://github.com/transcom/mymove/wiki/migrate-the-database#zero-downtime-migrations))
  * [ ] Have been communicated to #g-database
  * [ ] Secure migrations have been tested following the instructions in our [docs](https://github.com/transcom/mymove/wiki/migrate-the-database#secure-migrations)
* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7624) for this change

## Screenshots

If this PR makes visible UI changes, an image of the finished UI can help reviewers and casual
observers understand the context of the changes. A before image is optional and
can be included at the submitter's discretion.

Consider using an animated image to show an entire workflow instead of using multiple images. You may want to use GIPHY CAPTURE for this! 📸

[...]

PR #6434, opened 2021-04-15
[MB-7721] [Bug] fix to ProcessAndLockReviewedPR to use the tx inside the transaction …

jacquelineIO:
…instead of using the regular db call

# Description

Update to transaction used in ProcessAndLockReviewedPR to use the transaction and not the receiver's db connection. When the update was made this caused a panic in UT due to a deadlock cause by the locked payment request UT. (Great test, because it caught an issue.) 

Reggie found an issue with how we were doing [explicit locking](https://www.postgresql.org/docs/12/explicit-locking.html#LOCKING-ROWS) in the database. Due to an invoicing process that will be run async, our code locks the payment request that is being processed in the async task. And noted that we should use `FOR NO KEY UPDATE` instead of the `FOR UPDATE` as describe in this [SO thread](https://dba.stackexchange.com/questions/188169/select-for-update-locking-other-tables-in-postgresql).

`NO KEY UPDATE` is also used here https://github.com/transcom/mymove/blob/390cf490fb1ef84d85f20095807eb72a979f4501/pkg/services/payment_request/payment_request_creator.go#L213[…]L221. 

Also discovered in ths PR were a couple of places that should also be using the transaction in their process instead of using the regular db connection. Which were the invoice generator code and the param cache code. So those updates were made here too.

## Reviewer Notes

### Create a new move  or use RDY4PY in local
To test locally, run your server and create a move. 

```
make db_dev_migrate db_dev_e2e_populate
make server_run
make client_run
```

Create a move, approve it with the TOO UI.  If using `RDY4PY`, then just 
run the `prime-api-demo` until you create the payment request. Then run the 
```sh
prime-api-demo RDY4PY local
```
through until you have created a payment request. 

Have the TIO UI approve the payment request then generate an EDI. 

```sh
go run ./cmd/generate-payment-request-edi/ --payment-request-number 3983-9154-1
```

### Test support API
Save this in tmp/local_get_edi.json
```json
{
  "paymentRequestID": "9a4b881b-c8fd-40d0-864f-5a0290ff9073"
}
```

```sh
prime-api-client --insecure support-get-payment-request-edi --filename tmp/payloads/stg_get_edi.json| jq -r .edi  
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7721) for this change

## Screenshots

n/a


gidjin (approved):
This looks good to me! Left one question.

pkg/services/invoice/ghc_payment_request_invoice_generator_test.go
```diff
@@ -846,3 +850,47 @@ func (suite *GHCInvoiceSuite) TestTruncateStrFunc() {
...
+	testdatagen.MakePaymentServiceItemWithParams(
+		suite.DB(),
+		models.ReServiceCodeCS,
+		basicPaymentServiceItemParams,
+		assertions,
+	)
+
+	_, err := generator.Generate(paymentRequest, false)
+	suite.Error(err, "DB pointer is nil")
+}
```
monfresh, line 896: Since the DB check is performed at the very beginning of the `Generate` method, it doesn't depend on any data in the DB. So, we can avoid using the DB completely and can simplify this test to this:
```golang
generator := NewGHCPaymentRequestInvoiceGenerator(suite.icnSequencer, clock.NewMock())
_, err := generator.Generate(models.PaymentRequest{}, false)

suite.Error(err, "DB pointer is nil")
```

pkg/services/invoice/ghc_payment_request_invoice_generator.go
```diff
@@ -42,8 +42,16 @@ const isaDateFormat = "060102"
...
 
+// InitDB stores a database pop connection to the receiver instance
+func (g *ghcPaymentRequestInvoiceGenerator) InitDB(db *pop.Connection) {
+	g.db = db
+}
+
 // Generate method takes a payment request and returns an Invoice858C
 func (g ghcPaymentRequestInvoiceGenerator) Generate(paymentRequest models.PaymentRequest, sendProductionInvoice bool) (ediinvoice.Invoice858C, error) {
+	if g.db == nil {
+		return ediinvoice.Invoice858C{}, services.NewQueryError("pop.Connection", errors.New("database not initialized"), "DB pointer is nil, call ghcPaymentRequestInvoiceGenerator.InitDB()")
```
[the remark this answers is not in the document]
jacquelineIO: @gidjin  I didn't have any formatting so didn't use `Errorf` but other than that, not that I know of.

pkg/services/invoice/ghc_payment_request_invoice_generator_test.go
jacquelineIO: @monfresh  that is true. I will fix that in https://github.com/transcom/mymove/pull/6406 because it is waiting on the fix from this PR. And I'm working both and would like to get this one done. Thank you for the review. I'm still doing UT updates in 6406.

PR #6436, opened 2021-04-15
[MB-7664] Add Sort by GBLOC to the ListOrders for Service Counseling

shimonacarvalho:
## Description

This ticket is to ensure that sorting by GBLOC works. 

If you've accessed the Office UI as a TOO, you'll notice we only see one GBLOC, the one that matches the office user's transportation office. So you can't really sort GBLOCs if you're only getting one GBLOC. 

The GBLOC sorting comes into play for USMC office users. We have a separate transportation office for USMC office users. If you sign in as office user from that GBLOC, you should see all moves associated with US Marine service users. In that case, there could be multiple GBLOCs so you can test the sort. 

To test
- Go to `officelocal:3000`
- Sign in as a USMC office user. In devlocal local sign-in, you'll one available called `too_tio_role_usmc@office.mil`
- Go to the services counseling page `officelocal:3000/counseling/queue` and sort the 2 users you should see using the gbloc.

Also there's a new test in pkg/services/order

## References

* [Jira story](https://dp3.atlassian.net/browse/mb-7664) for this change

### DIFF
--- a/pkg/services/order/order_fetcher.go
+++ b/pkg/services/order/order_fetcher.go
@@ -45,6 +45,7 @@ func (f orderFetcher) ListOrders(officeUserID uuid.UUID, params *services.ListOr
 	// We also only want to do the gbloc filtering thing if we aren't a USMC user, which we cover with the else.
 	var gblocQuery QueryOption
 	if gbloc == "USMC" {
+		fmt.Println("gbloc is usmc")
 		branchQuery = branchFilter(swag.String(string(models.AffiliationMARINES)))
 	} else {
 		gblocQuery = gblocFilter(gbloc)
@@ -256,7 +257,6 @@ func sortOrder(sort *string, order *string) QueryOption {
 		"submittedAt":            "moves.submitted_at",
 		"destinationDutyStation": "dest_ds.name",
 		"originGBLOC":            "origin_to.gbloc",
-		"requestedMoveDate":      "min(mto_shipments.requested_pickup_date)",
 	}
 
 	return func(query *pop.Query) {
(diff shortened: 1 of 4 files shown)


pkg/services/order/order_fetcher.go
```diff
@@ -45,6 +45,7 @@ func (f orderFetcher) ListOrders(officeUserID uuid.UUID, params *services.ListOr
 	// We also only want to do the gbloc filtering thing if we aren't a USMC user, which we cover with the else.
 	var gblocQuery QueryOption
 	if gbloc == "USMC" {
+		fmt.Println("gbloc is usmc")
```
eamahanna, line 48: You may want to remove this line :)

pkg/services/order/order_fetcher.go
```diff
@@ -252,7 +256,7 @@ func sortOrder(sort *string, order *string) QueryOption {
 		"status":                 "moves.status",
 		"submittedAt":            "moves.submitted_at",
 		"destinationDutyStation": "dest_ds.name",
-		"requestedMoveDate":      "min(mto_shipments.requested_pickup_date)",
```
eamahanna, line 255: Was this removed on purpose? I think it needs to stay.

pkg/services/order/order_fetcher.go
shimonacarvalho: You are quick on the draw, I hadn't added any reviewers yet because I was dealing with a messy merge! Will fix, and thanks!

pkg/services/order/order_fetcher.go
shimonacarvalho: Returned!

eamahanna (approved):

PR #6437, opened 2021-04-15
MB-7633 Update fetchMTOUpdates to return origin duty station

eamahanna:
## Description

This PR fixes fetchMtoUpdates to return the origin duty stating information.

## Reviewer Notes

Should FetchMoveTaskOrder in the move_task_order_fetcher also return the origin duty station?

## Setup

To verify `fetchMTOUpdates` return the originDutyStation information, use postman and in the response you will see 
```
            "originDutyStation": {
                "address": {
                    "city": "Des Moines",
                    "country": "US",
                    "eTag": "MjAyMS0wNC0xNVQyMDoyMjo0MC4yODY0ODVa",
                    "id": "b2e4c180-1b9c-41f4-b041-a5321f012e69",
                    "postalCode": "50309",
                    "state": "IA",
                    "streetAddress1": "987 Other Avenue",
                    "streetAddress2": "P.O. Box 1234",
                    "streetAddress3": "c/o Another Person"
                },
                "addressID": "b2e4c180-1b9c-41f4-b041-a5321f012e69",
                "id": "38bc33e0-0c57-48cc-991a-9ba3ea9be235",
                "name": "DLs6QYVS4w"
            },
```

## References

## Screenshots


pkg/services/move_task_order/move_task_order_fetcher_test.go
```diff
@@ -62,10 +62,10 @@ func (suite *MoveTaskOrderServiceSuite) TestListMoveTaskOrdersFetcher() {
...
 		suite.Equal(expectedMTO.Orders.ID, actualMTO.Orders.ID)
 		suite.NotZero(actualMTO.Orders)
-		suite.NotNil(expectedMTO.Locator)
-		suite.NotNil(expectedMTO.ReferenceID)
-		suite.Nil(expectedMTO.AvailableToPrimeAt)
-		suite.NotEqual(expectedMTO.Status, models.MoveStatusCANCELED)
+		suite.NotNil(actualMTO.Locator)
+		suite.NotNil(actualMTO.ReferenceID)
+		suite.Nil(actualMTO.AvailableToPrimeAt)
+		suite.NotEqual(actualMTO.Status, models.MoveStatusCANCELED)
```
sandy-wright, line 68: ooh, nice catch!!

pkg/services/move_task_order/move_task_order_fetcher.go
```diff
@@ -54,6 +54,7 @@ func (f moveTaskOrderFetcher) ListAllMoveTaskOrders(searchParams *services.ListM
 		"Orders.ServiceMember",
 		"Orders.Entitlement",
 		"Orders.NewDutyStation.Address",
+		"Orders.OriginDutyStation.Address",
```
sandy-wright, line 57: I think this might cause issues because `origin_duty_station` can be null. So if any Orders don't have an Origin Duty Station, trying to get the address from a nil object will cause a panic. 

You can see this based on the definition on the model: 
```go
type Order struct {
    ...
	OriginDutyStation   *DutyStation                       `belongs_to:"duty_stations" fk_id:"origin_duty_station_id"`
	OriginDutyStationID *uuid.UUID                         `json:"origin_duty_station_id" db:"origin_duty_station_id"`
	NewDutyStationID    uuid.UUID                          `json:"new_duty_station_id" db:"new_duty_station_id"`
	NewDutyStation      DutyStation                        `belongs_to:"duty_stations" fk_id:"new_duty_station_id"`
    ...
}
```

And on the table definition for `orders`:
```sql
create table orders
(
new_duty_station_id    uuid                                            not null
        constraint orders_new_duty_station_id_fkey
            references duty_stations
            on delete cascade,
...
origin_duty_station_id uuid
);
```

One way to test this would be to borrow experimental and try `fetchMTOUpdates` there. Experimental has moves that are missing a LOT of data, so if it fails anywhere, I could see it failing there.

sandy-wright (commented):
I think this update might introduce a potential panic, so we might want to revisit it. I'm testing atm and looking for some good examples, so I'll keep you updated :)

pkg/services/move_task_order/move_task_order_fetcher.go
sandy-wright: Although I see this is also live in the `ListOrders` service object used to display the frontend queues, so maybe it's fine? Would be good to test in Experimental either way though, because I suspect it's an unhandled error

shimonacarvalho: I'm not sure it would Sandy, since those params are strings that just tell pop what tables to look for to populate data, not so much actual variable dereferences. For e.g. you could have no Dimensions on MTOServiceItems and that would be ok. 

I think it's a good idea to test though. A simple way would be to find a move in devlocal, delete the associated originDutyStation id so it doesn't have one, and then fetch-mto-updates on it.

sandy-wright: There is also a special case in this service object for another ref in `OriginDutyStation`:
```go
	for i := range moves {
		// There appears to be a bug in Pop for EagerPreload when you have two or more eager paths with 3+ levels
		// where the first 2 levels match.  For example:
		//   "Orders.OriginDutyStation.Address" and "Orders.OriginDutyStation.TransportationOffice"
		// In those cases, only the last relationship is loaded in the results.  So, we can only do one of the paths
		// in the EagerPreload above and request the second one explicitly with a separate Load call.
		//
		// Note that we also had a problem before with Eager as well.  Here's what we found with it:
		//   Due to a bug in pop (https://github.com/gobuffalo/pop/issues/578), we
		//   cannot eager load the address as "OriginDutyStation.Address" because
		//   OriginDutyStation is a pointer.
		if moves[i].Orders.OriginDutyStation != nil {
			loadErr := f.db.Load(moves[i].Orders.OriginDutyStation, "TransportationOffice")
			if loadErr != nil {
			}
		}
	}
```
I think we need to be doing this with Address too. It even mentions Address, but doesn't include it in this chunk? 🤔 @duncan-truss @monfresh @reggieriser would any of y'all have insight on this? I see y'all were recently in this file (`order_fetcher.go`)

sandy-wright: Oh WHOOPS, I didn't see your comment, @shimonacarvalho!!! I did just test it locally, and it seemed to be fine! I could've sworn there was an error from Pop with this in the past, but maybe I made it up 😄. Like it was trying to load all of the data even when the parent was nil

shimonacarvalho: I tested and it worked.

pkg/services/move_task_order/move_task_order_fetcher_test.go
shimonacarvalho: In the suite.NotEqual, can you swap the params, Emily? The first should be expected, the second actual. Makes the errors easier to understand.

pkg/services/move_task_order/move_task_order_fetcher_test.go
```diff
@@ -134,6 +136,9 @@ func (suite *MoveTaskOrderServiceSuite) TestListAllMoveTaskOrdersFetcher() {
 		}
 		suite.True(found)
 		suite.Equal(len(moveTaskOrders), 4)
```
shimonacarvalho, line 138: Same here, swap the expected, and actual.

pkg/services/move_task_order/move_task_order_fetcher.go
shimonacarvalho: Oh weird, I didn't see YOUR comment, hehe. Weird. I mean, there are def some bugs in Pop, so it's always worth a check.

sandy-wright (approved):
Okay, this LGTM! Sorry about the confusion. I do have one suggestion though: We might want to add this line to `FetchMoveTaskOrder` as well, line 109: https://github.com/transcom/mymove/blob/01170c19e9d5abaed7bb14461b6bd74ff0e2a67e/pkg/services/move_task_order/move_task_order_fetcher.go#L109

Only if it's simple! Totally optional

pkg/services/move_task_order/move_task_order_fetcher.go
sandy-wright: I also tested with resounding success, so I think we're fine! But I was secretly kind of hoping this was the cause of the some of the office errors I get almost every time on Exp haha

pkg/services/move_task_order/move_task_order_fetcher_test.go
sandy-wright: oh that's right, I always forget that too

pkg/services/move_task_order/move_task_order_fetcher.go
monfresh: I've noticed that GitHub sometimes doesn't show updates, even after a refresh, so I will miss some comments as well (or Circle CI updates). Comments usually come through right away via email, but not on the website.

pkg/services/move_task_order/move_task_order_fetcher_test.go
monfresh: @eamahanna It looks like the change Shimona suggested didn't make it before merging. Perhaps you forgot to push the changes back up? Were you planning to do a follow-up PR?

PR #6439, opened 2021-04-16
Refactor PaymentRequestReviewedProcessor and tests

monfresh:
## Description

This speeds up the tests by more than half: from over 30 seconds down to
about 12 seconds.

Here are the principles I used in this refactor:

- Functions should be as short as possible
- Extract each logical component of a function into a separate function
- Where it makes sense, create an interface for that logical component
so it can be mocked easily in tests
- Tests should be as fast as possible
- Tests should use the DB as little as possible
- Don't test the same things in every test. Each variation should focus
on what's unique and mock everything else that was already tested.
- If there are more than 2 subtests that write to the same DB table,
use `suite.TruncateAll()` before each subtest to avoid confusion around
what data is expected to be present in each subtest. By not truncating,
it prevents running a single subtest and forces the developer to run
the entire set of subtests, which slows them down. This can be avoided
by using the DB as little as possible. See the previous principles.

I added more comments inline.

There is further improvement that can speed up the tests even more, by
being able to mock more of this flow.

There's also a missing test that I wasn't sure how to write off the top
of my head. Since this workflow runs while iterating through payment
requests, we want to test that if there is more than one payment
request, if the first one succeeds, but the second one does not, the
first one should pass, but the second one should not. I think that's
what we want, but I'm not 100% sure. I don't know if testify supports
this kind of mocking, where you tell it to return an error only on the
second time.

## Reviewer Notes

## Setup

```sh
echo "Code goes here"
```

## References

## Screenshots

### DIFF
--- a/pkg/payment_request/send_to_syncada.go
+++ b/pkg/payment_request/send_to_syncada.go
@@ -14,9 +14,6 @@ import (
 func SendToSyncada(edi string, gexSender services.GexSender, sftpSender services.SyncadaSFTPSender, sendEDIFile bool, logger Logger) error {
 	var err error
 
-	if (gexSender == nil) && (sftpSender == nil) {
-		return fmt.Errorf("cannot send to Syncada, SendToSyncada() senders are nil")
-	}
 	if gexSender != nil {
 		//TODO: Send to Syncada via GEX needs to be implemented
 		logger.Warn("func SendToSyncada() -- GEX Sender NOT IMPLEMENTED")
--- a/pkg/services/payment_request/payment_request_reviewed_processor.go
+++ b/pkg/services/payment_request/payment_request_reviewed_processor.go
@@ -26,6 +26,8 @@ type paymentRequestReviewedProcessor struct {
 	runSendToSyncada              bool // if false, do not send to Syncada, e.g. UT shouldn't send to Syncada
 	gexSender                     services.GexSender
 	sftpSender                    services.SyncadaSFTPSender
+	numProcessed                  int
+	startedAt                     time.Time
 }
 
 // NewPaymentRequestReviewedProcessor returns a new payment request reviewed processor
@@ -73,14 +75,13 @@ func InitNewPaymentRequestReviewedProcessor(db *pop.Connection, logger Logger, s
 
 func (p *paymentRequestReviewedProcessor) ProcessAndLockReviewedPR(pr models.PaymentRequest) error {
 	var transactionError error
-
 	transactionError = p.db.Transaction(func(tx *pop.Connection) error {
 		var lockedPR models.PaymentRequest
 
 		query := `
-			SELECT * FROM payment_requests
-			WHERE id = $1 FOR UPDATE SKIP LOCKED;
-		`
+      		SELECT * FROM payment_requests
+      		WHERE id = $1 FOR UPDATE SKIP LOCKED;
+    	`
 		err := p.db.RawQuery(query, pr.ID).First(&lockedPR)
 		if err != nil {
 			if err == sql.ErrNoRows {
@@ -174,28 +175,41 @@ func (p *paymentRequestReviewedProcessor) ProcessAndLockReviewedPR(pr models.Pay
 }
 
 func (p *paymentRequestReviewedProcessor) ProcessReviewedPaymentRequest() error {
+	// fail early if both senders are nil
+	if (p.gexSender == nil) && (p.sftpSender == nil) {
+		return fmt.Errorf("cannot send to Syncada, SendToSyncada() senders are nil")
+	}
+
+	p.numProcessed = 0
+	p.startedAt = time.Now()
+
 	// Store/log metrics about EDI processing upon exiting this method.
-	numProcessed := 0
-	start := time.Now()
-	defer func() {
-		ediProcessing := models.EDIProcessing{
-			EDIType:          models.EDIType858,
-			ProcessStartedAt: start,
-			ProcessEndedAt:   time.Now(),
-			NumEDIsProcessed: numProcessed,
-		}
-		p.logger.Info("EDIs processed", zap.Object("EDIs processed", &ediProcessing))
+	defer p.saveEDIProcessing()
 
-		verrs, err := p.db.ValidateAndCreate(&ediProcessing)
-		if err != nil {
-			p.logger.Error("failed to create EDIProcessing record", zap.Error(err))
-		}
-		if verrs.HasAny() {
-			p.logger.Error("failed to validate EDIProcessing record", zap.Error(err))
-		}
-	}()
+	return p.ProcessAllReviewedPaymentRequests()
+}
 
-	// Fetch all payment request that have been reviewed
+func (p *paymentRequestReviewedProcessor) saveEDIProcessing() {
+	ediProcessing := models.EDIProcessing{
+		EDIType:          models.EDIType858,
+		ProcessStartedAt: p.startedAt,
+		ProcessEndedAt:   time.Now(),
+		NumEDIsProcessed: p.numProcessed,
+	}
+
+	p.logger.Info("EDIs processed", zap.Object("EDIs processed", &ediProcessing))
+
+	verrs, err := p.db.ValidateAndCreate(&ediProcessing)
+	if err != nil {
+		p.logger.Error("failed to create EDIProcessing record", zap.Error(err))
+	}
+	if verrs.HasAny() {
+		p.logger.Error("failed to validate EDIProcessing record", zap.Error(err))
+	}
+}
+
+func (p *paymentRequestReviewedProcessor) ProcessAllReviewedPaymentRequests() error {
+	// Fetch all payment requests that have been reviewed
 	reviewedPaymentRequests, err := p.reviewedPaymentRequestFetcher.FetchReviewedPaymentRequest()
 	if err != nil {
 		return fmt.Errorf("function ProcessReviewedPaymentRequest failed call to FetchReviewedPaymentRequest: %w", err)
@@ -212,8 +226,7 @@ func (p *paymentRequestReviewedProcessor) ProcessReviewedPaymentRequest() error 
 		if err != nil {
 			return err
 		}
-		numProcessed++
+		p.numProcessed++
 	}
-
 	return nil
 }
--- a/pkg/services/payment_request/payment_request_reviewed_processor_test.go
+++ b/pkg/services/payment_request/payment_request_reviewed_processor_test.go
@@ -14,7 +14,6 @@ import (
 	"time"
 
 	"github.com/benbjohnson/clock"
-	"github.com/gofrs/uuid"
 	"github.com/pkg/errors"
 
 	ediinvoice "github.com/transcom/mymove/pkg/edi/invoice"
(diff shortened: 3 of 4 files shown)


pkg/payment_request/send_to_syncada.go
```diff
@@ -14,9 +14,6 @@ import (
 func SendToSyncada(edi string, gexSender services.GexSender, sftpSender services.SyncadaSFTPSender, sendEDIFile bool, logger Logger) error {
 	var err error
 
-	if (gexSender == nil) && (sftpSender == nil) {
-		return fmt.Errorf("cannot send to Syncada, SendToSyncada() senders are nil")
-	}
```
monfresh, line 19: I moved this to the very top of `ProcessReviewedPaymentRequest` since it seems like if this fails, the whole workflow should fail. I might be missing something, but it didn't seem to make sense to attempt any work if this fails.

pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -73,14 +75,13 @@ func InitNewPaymentRequestReviewedProcessor(db *pop.Connection, logger Logger, s
...
 	transactionError = p.db.Transaction(func(tx *pop.Connection) error {
 		var lockedPR models.PaymentRequest
 
 		query := `
-			SELECT * FROM payment_requests
-			WHERE id = $1 FOR UPDATE SKIP LOCKED;
-		`
+      		SELECT * FROM payment_requests
+      		WHERE id = $1 FOR UPDATE SKIP LOCKED;
+    	`
```
monfresh, line 84: I think this is my editor doing automatic formatting. Is this right?

pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -174,28 +175,41 @@ func (p *paymentRequestReviewedProcessor) ProcessAndLockReviewedPR(pr models.Pay
...
-	start := time.Now()
-	defer func() {
-		ediProcessing := models.EDIProcessing{
-			EDIType:          models.EDIType858,
-			ProcessStartedAt: start,
-			ProcessEndedAt:   time.Now(),
-			NumEDIsProcessed: numProcessed,
-		}
-		p.logger.Info("EDIs processed", zap.Object("EDIs processed", &ediProcessing))
+	defer p.saveEDIProcessing()
```
monfresh, line 187: I'm not sure if this is the best name for this function. Would `logEDIProcessingMetrics` be better?

pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -174,28 +175,41 @@ func (p *paymentRequestReviewedProcessor) ProcessAndLockReviewedPR(pr models.Pay
...
 
-		verrs, err := p.db.ValidateAndCreate(&ediProcessing)
-		if err != nil {
-			p.logger.Error("failed to create EDIProcessing record", zap.Error(err))
-		}
-		if verrs.HasAny() {
-			p.logger.Error("failed to validate EDIProcessing record", zap.Error(err))
-		}
-	}()
+	return p.ProcessAllReviewedPaymentRequests()
```
monfresh, line 189: By extracting these 2 separate events into their own functions, it makes this function more readable and now you can easily see what it does. It's no longer necessary to read through several lines of code to figure out what's happening at a high level. If you want more details, you can look at the functions.

pkg/services/payment_request/payment_request_reviewed_processor_test.go
```diff
@@ -175,57 +174,10 @@ func (suite *PaymentRequestServiceSuite) TestProcessReviewedPaymentRequest() {
 	//   ssh-keygen -q -N "" -t ecdsa -f /tmp/ssh_host_ecdsa_key
 	os.Setenv("SYNCADA_SFTP_HOST_KEY", "ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBI+M4xIGU6D4On+Wxz9k/QT12TieNvaXA0lvosnW135MRQzwZp5VDThQ6Vx7yhp18shgjEIxFHFTLxpmUc6JdMc= fake@localhost")
 
-	suite.T().Run("process reviewed payment request successfully (0 Payments to review)", func(t *testing.T) {
-		var ediProcessingBefore models.EDIProcessing
-		countProcessingRecordsBefore, err := suite.DB().Where("edi_type = ?", models.EDIType858).Count(&ediProcessingBefore)
-		suite.NoError(err, "Get count of EDIProcessing")
```
monfresh, line 181: It's not necessary to check the before and after count when the DB is truncated in between tests. We can check at the end of test that there was only 1 EDI Processing created since we started at 0.

pkg/services/payment_request/payment_request_reviewed_processor_test.go
```diff
@@ -175,57 +174,10 @@ func (suite *PaymentRequestServiceSuite) TestProcessReviewedPaymentRequest() {
...
-		suite.NoError(err, "Get number of processed files")
-		suite.Equal(0, ediProcessing.NumEDIsProcessed)
-
-		newCount, err := suite.DB().Where("edi_type = ?", models.EDIType858).Count(&ediProcessing)
-		suite.NoError(err, "Get count of EDIProcessing")
-		suite.Greater(newCount, countProcessingRecordsBefore)
-		suite.Equal(countProcessingRecordsBefore+1, newCount)
-	})
-
-	suite.T().Run("process reviewed payment request successfully (1 payment request reviewed all rejected excluded)", func(t *testing.T) {
```
monfresh, line 214: This test is not necessary because `FetchReviewedPaymentRequest` is already unit tested in `/pkg/services/payment_request/payment_request_reviewed_fetcher_test.go`, so we know that it only returns reviewed payment requests. Therefore, we can always mock it in all of these tests, which saves us some DB calls.

pkg/services/payment_request/payment_request_reviewed_processor_test.go
```diff
@@ -175,57 +174,10 @@ func (suite *PaymentRequestServiceSuite) TestProcessReviewedPaymentRequest() {
...
-		rejectedPaymentRequest := testdatagen.MakePaymentRequest(suite.DB(), testdatagen.Assertions{
-			PaymentRequest: models.PaymentRequest{
-				IsFinal:         false,
-				Status:          models.PaymentRequestStatusReviewedAllRejected,
-				RejectionReason: &rejectionReason,
-			},
-		})
+	suite.T().Run("EDIProcessing record is created even when there are 0 payment requests to process", func(t *testing.T) {
+		reviewedPaymentRequestFetcher := &mocks.PaymentRequestReviewedFetcher{}
+		reviewedPaymentRequestFetcher.On("FetchReviewedPaymentRequest", mock.Anything, mock.Anything).Return(models.PaymentRequests{}, nil)
```
monfresh, line 179: `FetchReviewedPaymentRequest` is already unit tested in `/pkg/services/payment_request/payment_request_reviewed_fetcher_test.go`, so we know that it only returns reviewed payment requests. Therefore, we can always mock it in all of these tests, which saves us some DB calls.

pkg/services/payment_request/payment_request_reviewed_processor_test.go
```diff
@@ -242,39 +194,23 @@ func (suite *PaymentRequestServiceSuite) TestProcessReviewedPaymentRequest() {
...
-		prs := suite.createPaymentRequest(4)
-
-		_ = testdatagen.MakeExtendedServiceMember(suite.DB(), testdatagen.Assertions{
-			ServiceMember: models.ServiceMember{
-				ID: uuid.FromStringOrNil("d66d2f35-218c-4b85-b9d1-631949b9d984"),
-			},
-		})
+		suite.TruncateAll()
 
+		prs := suite.createPaymentRequest(2)
```
monfresh, line 213: Unless I'm missing something, we should never need more than 2 payment requests in any of these tests. The only reason we are creating 2 in this test is because we can't mock the call to `saveEDIProcessing` and assert that it only gets called once. We could though once we convert it to an interface. AFAIK, testify mocks only work with interfaces, right?

pkg/services/payment_request/payment_request_reviewed_processor_test.go
```diff
@@ -371,21 +308,11 @@ func (suite *PaymentRequestServiceSuite) TestProcessReviewedPaymentRequest() {
 		// we only expect the first payment request with an error to be recorded.
 		suite.Len(ediErrors, 1)
 		suite.Contains(*(ediErrors[0].Description), "test error")
-		suite.Equal(ediErrors[0].PaymentRequestID, prs[0].ID)
-
-		// Make sure that PR status is updated
-		var updatedPaymentRequest models.PaymentRequest
-		err = suite.DB().Where("id = ?", prs[0].ID).First(&updatedPaymentRequest)
-		suite.NoError(err)
-		suite.Equal(updatedPaymentRequest.Status, models.PaymentRequestStatusEDIError)
```
monfresh, line 380: The PR was already fetched above on line 352. I could be missing something, but it seems like we can check both `SentToGexAt` and the `Status` together up above, no?

pkg/services/payment_request/payment_request_reviewed_processor_test.go
```diff
@@ -459,36 +370,23 @@ func (suite *PaymentRequestServiceSuite) TestProcessReviewedPaymentRequest() {
...
 		fetcher := NewPaymentRequestFetcher(suite.DB())
-		for _, pr := range prs {
-			paymentRequest, _ := fetcher.FetchPaymentRequest(pr.ID)
-			suite.Nil(paymentRequest.SentToGexAt)
-		}
-
-		var ediProcessing models.EDIProcessing
-		err = suite.DB().Where("edi_type = ?", models.EDIType858).Order("process_ended_at desc").First(&ediProcessing)
-		suite.NoError(err, "Get number of processed files")
-		suite.Equal(0, ediProcessing.NumEDIsProcessed)
```
monfresh, line 475: We don't need to test this in every test. If it works in one test, it has to work in the others.

pkg/services/payment_request/payment_request_reviewed_processor_test.go
```diff
@@ -565,67 +449,3 @@ func (suite *PaymentRequestServiceSuite) TestProcessReviewedPaymentRequest() {
...
-		SET
-			status = $2,
-		WHERE id = $1;
-	`
-	suite.DB().RawQuery(query, prID, models.PaymentRequestStatusPaid).Exec()
-	time.Sleep(1 * time.Second)
-	suite.DB().RawQuery(`COMMIT;`).Exec()
-}
-
-func (suite *PaymentRequestServiceSuite) TestProcessLockedReviewedPaymentRequest() {
```
monfresh, line 584: I read the PR that introduced this test and it was fascinating. I admire the work that went into testing this, so I feel bad removing it. Unfortunately, this test is not testing our code. AFAICT, the only part of our code that I could see that would affect this test is the query with `UPDATE SKIP LOCKED` at the top of `ProcessAndLockReviewedPR`. If I remove the `UPDATE SKIP LOCKED`, the test still passes. 

I believe that's what we were originally trying to test, and if we remove it and the test still passes, the test is not serving us. We are essentially testing Postgres. If we wanted to make sure we are using the skip locked feature of Postgres, we could look into asserting that a certain SQL query was made. This might require further refactoring to make this easier to test.

pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -174,28 +175,41 @@ func (p *paymentRequestReviewedProcessor) ProcessAndLockReviewedPR(pr models.Pay
 }
 
 func (p *paymentRequestReviewedProcessor) ProcessReviewedPaymentRequest() error {
+	// fail early if both senders are nil
+	if (p.gexSender == nil) && (p.sftpSender == nil) {
+		return fmt.Errorf("cannot send to Syncada, SendToSyncada() senders are nil")
+	}
+
+	p.numProcessed = 0
```
gidjin, line 183: Is it necessary to set this? Isn't the default value for an int already 0?

pkg/services/payment_request/payment_request_reviewed_processor.go
gidjin: This needs to still be wrapped in an anonymous function as I understand so that the changing values are updated accordingly. But @reggieriser might know better. The name seems fine since it is saving to the database not just logging. Maybe `saveAndLogEDIProcessing`

```golang
defer func() {
    p.saveEDIProcessing()
}()
```

pkg/services/payment_request/payment_request_reviewed_processor_test.go
gidjin: I believe that the mocks only work with interfaces too. Not sure that we need more than 2 either

pkg/services/payment_request/payment_request_reviewed_processor_test.go
gidjin: I disagree here, this test I believe, is what lead @jacquelineIO and @reggieriser to catch that we were not passing our database transaction connection down correctly. It also caught that we should have been doing `FOR NO KEY UPDATE` vs `FOR UPDATE`. See PR #6439 6434

gidjin (commented):
Thanks Moncef for tackling improving this. I left a few comments, that said we are actively working on this code for Outcome 1 stuff that will definitely cause conflicts. See #6427 and #6434. In light of that I'd rather this PR go in after those are done since we need to verify our connection to sftp GEX asap.

gidjin:
@mboren and @jacquelineIO I believe this PR would affect both of yours if you have a moment please take a look, though I will reiterate I think this should go in after your PRs are in and verified working in Staging.

pkg/services/payment_request/payment_request_reviewed_processor_test.go
monfresh: I counter disagree 😄 . I just pulled down #6434, and I removed the `NO KEY` and reverted the code back to `FOR UPDATE SKIP LOCKED`, and the test still passed. Is it failing for you if you change the query in the actual `ProcessAndLockReviewedPR` function in `services/payment_request/payment_request_reviewed_processor.go`?

If you think about it, the test is hardcoding the query that we want, so it's not testing our code.

monfresh: It also still passes if I revert the change from `p.db` to `tx`.

PR #6448, opened 2021-04-19
Revert "Revert "[MB-7623] Switch the connection for the Job Runner to use HTTPS to GEX""

mboren:
Reverts transcom/mymove#6447

The original changes broke master due to a bad automerge. I ended up with a duplicate import for pop, and some tests that I added used a function that had a different signature on master, so that broke too.


gidjin (approved):
LGTM

PR #6451, opened 2021-04-20
Add annotation regarding use of child-process

kimallen:
## Description

We are disabling `eslint-plugin-security` linter rule `detect-child-process` in the `dangerfile.ts`.  This PR adds annotation explaining why we are using the child-process and how this usage does not involve user input, which is the concern.

## Reviewer Notes

This article explains the concern and how to avoid it https://github.com/nodesecurity/eslint-plugin-security/blob/master/docs/avoid-command-injection-node.md.

```
Avoid using child_process.exec, and never use it if the command contains any input that changes based on user input.
Try to avoid letting users pass in options to commands if possible. Typically values are okay when using spawn or execfile, but selecting options via a user controlled string is a bad idea.
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7722) for this change
* [this article](https://github.com/nodesecurity/eslint-plugin-security/blob/master/docs/avoid-command-injection-node.md.) explains more about the approach used.

### DIFF
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -6,9 +6,9 @@ import { danger, warn, fail } from 'danger';
 //RA: This usage checks for any critical or high vulnerabilities and upgrades in our dependencies to alert the Github user.
 //RA: This usage does not utilize any user input and there is no opening for command injection.
 //RA Developer Status: Mitigated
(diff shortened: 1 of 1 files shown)


Leofscott (approved):
LGTM

jacquelineIO (approved):
Looks good to me. Also +1 to not using `child.exec` and using `child.spawnSync` instead.

PR #6461, opened 2021-04-20
MB-3164 service counselor login

eamahanna:
## Description

This PR creates a login process for the service counselor role, adds seed services counselor users, and updates tests.

## Reviewer Notes

Currently in the app, when a user is logged in with a role and attempts to access a page they do not have the correct role for, they are redirected to the page that they are authorized to view. This is how the services counselor page works as well. The ticket for this work says the user should be redirected to a 403. Is this required/in scope?

## Setup

To run the tests execute `make test`

To test the login functionality in the UI complete the following steps:

Execute `make db_dev_e2e_populate`

DevLocal:
- Start the office app
- Select one of the existing users for services counselor 
- You should be redirected to the services counselor queue
- Logout
- Click local signin again and select one of the users that has multiple roles including the services counselor
- When redirected, click on the Change User Roles in the upper left corner
- Select services counselor
- You should be redirected to the counseling queue

Login.gov
- Login as a services counselor
- You should be redirected to the services counseling queue

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3164) for this change

## Screenshots
Screen Shot 2021-04-20 at 12 55 54 PM
Screen Shot 2021-04-20 at 12 57 47 PM


pkg/auth/authentication/devlocal.go
```diff
@@ -589,6 +601,76 @@ func createUser(h devlocalAuthHandler, w http.ResponseWriter, r *http.Request) (
...
+
+		verrs, err = h.db.ValidateAndSave(&usersRole)
+		if err != nil {
+			h.logger.Error("could not create user role", zap.Error(err))
+		}
+		if verrs.HasAny() {
+			h.logger.Error("validation errors creating user role", zap.Stringer("errors", verrs))
+		}
+
+		office := models.TransportationOffice{
```
duncan-truss, line 646: I'm actually working on a [PR this week to fix these devlocal transportation offices](https://dp3.atlassian.net/browse/MB-7667) to include a value for Gbloc because it is currently missing.  Do you mind adding `Gbloc: "LKNQ"` here?

eamahanna: Sure!

eamahanna: That fixes an issue I was having lol

pkg/testdatagen/make_office_user.go
```diff
@@ -203,3 +226,32 @@ func MakeOfficeUserWithUSMCGBLOC(db *pop.Connection) models.OfficeUser {
 		},
 	})
 }
+
+// MakeServicesCounselorOfficeUserWithUSMCGBLOC makes an OfficeUser tied to the USMC GBLOC
```
shimonacarvalho, line 230: Minor change, could you say makes a Services Counselor tied to the USMC GBLOC? There is another function that makes a USMC office user with the identical comment.

pkg/testdatagen/scenario/devseed.go
```diff
@@ -2756,6 +2756,36 @@ func createTIO(db *pop.Connection) {
 	})
 }
 
+func createServicesCounselor(db *pop.Connection) {
+	/* A user with tio role */
```
shimonacarvalho, line 2760: This comment should likely say "services counselor" role.

sandy-wright (commented):
Nice, LGTM! The redirect seems to be a lil wonky when I create a fresh Service Counselor, but I think that's just the nature of the beast. I did leave some minor suggestions, but you could argue against them, so I'll leave it up to you to apply them or not ☺️

pkg/testdatagen/scenario/devseed.go
```diff
@@ -2826,6 +2856,82 @@ func createTXO(db *pop.Connection) {
...
+	tooTioServicesUUID := uuid.Must(uuid.FromString("8d78c849-0853-4eb8-a7a7-73055db7a6a8"))
+	loginGovUUID := uuid.Must(uuid.NewV4())
+	user := testdatagen.MakeUser(db, testdatagen.Assertions{
+		User: models.User{
+			ID:            tooTioServicesUUID,
+			LoginGovUUID:  &loginGovUUID,
+			LoginGovEmail: email,
+			Active:        true,
+			Roles:         []roles.Role{tooRole, tioRole},
+		},
```
shimonacarvalho, line 2889: A little confused by this function. Can you break it up with comments explaining what is happening? 
I'm assuming it would create one office user with all three roles. 

However, this user here gets created with only two roles (too, tio) on line 2882.
And then later the same userID gets used to create a service member on line 2899? I don't think an office user needs to be a service member?

And then another user gets created on line 2915, this time with all three roles and usmc. 

Would be good to clarify what is happening and why, how many users get created, and possibly break down into smaller creation functions.

pkg/testdatagen/scenario/e2ebasic.go
```diff
@@ -2026,6 +2054,34 @@ func (e e2eBasicScenario) Run(db *pop.Connection, userUploader *uploader.UserUpl
...
+	})
+	testdatagen.MakeOfficeUser(db, testdatagen.Assertions{
+		OfficeUser: models.OfficeUser{
+			ID:     uuid.FromStringOrNil("f3503012-e17a-4136-aa3c-508ee3b1962f"),
+			Email:  email,
+			Active: true,
+			UserID: &ttooTioServicesUUID,
+		},
+	})
+	testdatagen.MakeServiceMember(db, testdatagen.Assertions{
```
shimonacarvalho, line 2078: Here again, unsure why a servicemember is being created for the office user.

pkg/testdatagen/scenario/devseed.go
shimonacarvalho: Maybe this link will be helpful:
https://github.com/transcom/mymove/wiki/Understanding-Testdatagen-Functions

shimonacarvalho (changes requested):
Works perfectly! Tried with a devlocal and a login.gov user.
Some of the testdatagen functions are confusing and possibly need refactoring.

pkg/testdatagen/scenario/devseed.go
eamahanna: So it doesn't look like an office user has to be a service member, so I removed that part. This function creates two offices users with the 3 office user roles. However, one is associated with usmc. 

The reason we have to create a user instead of just having a default user created in `MakeOfficeUser` is because we need to have a hard coded userID. 	

I added comments, so hopefully is should make it clearer.

shimonacarvalho: Okay, looking better. I still think it would be good to split it up. The function name is `createTXOServicesCounselor` which sounds like one user is being created. Perhaps split it up into `createTXOServicesCounselor` and `createTXOServicesCounselorUSMC`.

pkg/testdatagen/scenario/devseed.go
```diff
@@ -2826,6 +2856,84 @@ func createTXO(db *pop.Connection) {
 	})
 }
 
+func createTXOServicesCounselor(db *pop.Connection) {
+	/* A user with both too, tio, and services counselor roles */
+	email := "too_tio_services_counselor_role@office.mil"
+	tooRole := roles.Role{}
+	err := db.Where("role_type = $1", roles.RoleTypeTOO).First(&tooRole)
+	if err != nil {
```
shimonacarvalho, line 2864: Is there any reason the three roles are queried one at a time? Seems like all the role ids could be queried at once, which would simplify this function further, and speed up the devseed code.

eamahanna: I think it makes it easier to panic and send a specific error message.

pkg/testdatagen/scenario/devseed.go
eamahanna: Okay I split it, but having them separate functions will create additional calls to the DB and likely slow it down a bit.

pkg/testdatagen/scenario/devseed.go
eamahanna: Can you provide an example of what you think I should change it to?

shimonacarvalho: Something like...
```
	addRoleTypes := []roles.RoleType{roles.RoleTypeTOO, roles.RoleTypeTIO, roles.RoleTypeServicesCounselor}
	var addRoles []roles.Role
	err := db.Where("role_type in (?)", addRoleTypes).All(&addRoles)
```
Basically here are the three roles i want to add to my user, get me their ids.

Sure the error will be more generic, but this is devseed. No customer will be reading the errors and the state of the system is pretty fixed, not dynamic, as it runs on an empty database. If you have no errors on checkin, almost guaranteed to work exactly the same each time. 

The thing with devseed is that it's gotten slower and slower and harder to understand. So any attempts to keep it clean and lean help a lot with tech debt.

pkg/testdatagen/scenario/devseed.go
shimonacarvalho: Are the additional calls the roles queries? That can be simplified down to one call per function, so overall we should have fewer I think. 

What happens is that people reuse these helper functions and if they think it creates one counselor but actually creates two, this doubling will happen each time. Over time, this gets impossible to clean up because it's hard to tell if they used the function because they wanted a regular Services Counselor or a USMC Services Counselor.

pkg/testdatagen/scenario/devseed.go
eamahanna: ahhhh, that makes sense. Changes pushed! Thanks for the feed back :)

shimonacarvalho (approved):
Looks good, thanks!

pkg/testdatagen/scenario/devseed.go
monfresh: @eamahanna What is the reason for hardcoding a user ID? Why can't we use the helper functions that we added in `make_move.go`, such as `MakeServicesCounselorOfficeUserWithUSMCGBLOC`?

pkg/testdatagen/scenario/e2ebasic.go
```diff
@@ -2026,6 +2054,28 @@ func (e e2eBasicScenario) Run(db *pop.Connection, userUploader *uploader.UserUpl
...
+	})
+	testdatagen.MakeOfficeUser(db, testdatagen.Assertions{
+		OfficeUser: models.OfficeUser{
+			ID:     uuid.FromStringOrNil("f3503012-e17a-4136-aa3c-508ee3b1962f"),
+			Email:  email,
+			Active: true,
+			UserID: &ttooTioServicesUUID,
+		},
+	})
+
```
monfresh, line 2078: `e2ebasic` is only used for the Cypress tests. If we add data to it without using it, we are unnecessarily slowing down the Cypress tests. Maybe I missed it, but I don't see any Cypress tests in this PR.

pkg/testdatagen/scenario/devseed.go
monfresh: Also, now we have both `createTXO` and `createTXOServicesCounselor`. It seems like having just one TXO with all 3 roles would be enough, as opposed to one that has 2 roles, and another that has 3 roles. Thoughts?

PR #6462, opened 2021-04-20
[MB-7546] Update PRToInterchangeControlNumber to have an EDIType Field

pearl-truss:
## Description

This PR updates the PaymentRequestToInterchangeControlNumber table and model to have an EDIType Field

## Reviewer Notes

This PR makes it EDIType mandatory for `PaymentRequestToInterchangeControlNumber` , do I need to do anything to address any existing `PaymentRequestToInterchangeControlNumber` records that do not have this field in the database?

## Setup

Ensure that the build still runs and server test are passing

```sh
make db_dev_migrate 
make server_run
make server_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7546) for this change


pkg/services/invoice/process_edi997.go
```diff
@@ -72,6 +72,7 @@ func (e *edi997Processor) ProcessFile(path string, stringEDI997 string) error {
 	prToICN := models.PaymentRequestToInterchangeControlNumber{
 		InterchangeControlNumber: int(icn),
```
jacquelineIO, line 73: On line 44, is that comment correct? I don't think that it is.

jacquelineIO: I don't think that line 45 is used until line 73. Can you move that line closer to where it is used?

pkg/services/invoice/process_edi824_test.go
```diff
@@ -171,6 +173,7 @@ IEA*1*000000996
 				PaymentRequestID:         paymentRequest.ID,
 				InterchangeControlNumber: 100001255,
 				PaymentRequest:           paymentRequest,
+				EDIType:                  models.EDIType824,
```
jacquelineIO, line 176: Why is there a `PaymentRequestToInterchangeControlNumber` entry happening here for the 824? This is being created in the process function https://github.com/transcom/mymove/pull/6462/files#diff-29b62994ec2bb4bcb70f80082f915a18289eacbde732051c57868ff8d1f97a8fR84 

Curious what the one here is doing?

migrations/app/schema/20210419195440_add_editype_to_paymentRequestToInterchange_table.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE payment_request_to_interchange_control_numbers
+    ADD COLUMN edi_type varchar not null;
```
reggieriser, line 2: I noticed you mention this possibility in the reviewer notes, but you'll have to consider any existing records or I think this will fail when run.  Options include:
- Add a default to the column, even if temporary.  But can we simply default the existing records (e.g., are they all 858s, for instance?) or will that cause some records to be marked incorrectly?
- Delete all existing `payment_request_to_interchange_control_numbers` records, but there could be consequences of doing so if the code expects them to always be present.
- Delete all existing `payment_request_to_interchange_control_numbers` records and cascade through associated payment requests and such so that you don't end up in an inconsistent state.  You'd need to make sure this is OK with the team.
- Make the column nullable, but that's relaxing a constraint that could be useful going forward.
- Make the column nullable, then adjust any existing records (is there a way to inspect the data and know how this field should be set?), then make the column non-nullable.  You could do this in a single migration since each migration is run in its own transaction.

reggieriser: Also, I noticed that we already have an `edi_type` enum that we're using in the `edi_processings` table.  Is this `edi_type` here different or should we be using the `edi_type` type instead of a `varchar`?  BTW, I noticed that the `edi_errors` table also has an `edi_type` column which is a `varchar` -- wonder if we need a story to change that as well.

jacquelineIO: Good call, I would say it's mostly informational. And defaulting them all to 858 or anything would be fine. All of the data is test data for now. And I don't think anything would care.

pkg/models/edi_errors_test.go
```diff
@@ -17,6 +17,7 @@ func (suite *ModelSuite) TestEdiErrors() {
 		ID:                       uuid.Must(uuid.NewV4()),
 		PaymentRequestID:         pr.ID,
 		InterchangeControlNumber: 5,
+		EDIType:                  models.EDIType997,
```
reggieriser, line 20: I was wondering why you had to add this line for the `edi_errors` table's tests that you didn't touch, then I saw that this test is creating a tree of records just to test the validations (so you had to put something in the non-nullable field).  I don't think we usually create records to test vadliations, and I don't think it's necessary since the validations usually only care that we have a UUID, not that it references a real UUID in the DB.  This is probably outside the scope of this PR, but we could remove that record creation and simplify things plus speed the test up.

pkg/models/payment_request_to_interchange_control_number.go
```diff
@@ -20,6 +21,7 @@ type PaymentRequestToInterchangeControlNumber struct {
 // Validate gets run every time you call a "pop.Validate*" (pop.ValidateAndSave, pop.ValidateAndCreate, pop.ValidateAndUpdate) method.
 func (p *PaymentRequestToInterchangeControlNumber) Validate(tx *pop.Connection) (*validate.Errors, error) {
 	return validate.Validate(
+		&validators.StringInclusion{Field: string(p.EDIType), Name: "EDIType", List: allowedEDITypes},
```
reggieriser, line 24: FYI, we do have a `String()` method on the `EDIType`, so you could use `p.EDIType.String()` as well (but no big deal either way).

pkg/models/payment_request_to_interchange_control_number_test.go
```diff
@@ -13,7 +13,10 @@ func (suite *ModelSuite) TestPaymentRequestToInterchangeControlNumber() {
 	suite.T().Run("test PaymentRequest association", func(t *testing.T) {
 		paymentRequest := testdatagen.MakePaymentRequest(suite.DB(), testdatagen.Assertions{})
 		validPR2ICN := testdatagen.MakePaymentRequestToInterchangeControlNumber(suite.DB(), testdatagen.Assertions{
-			PaymentRequestToInterchangeControlNumber: models.PaymentRequestToInterchangeControlNumber{PaymentRequestID: paymentRequest.ID}})
+			PaymentRequestToInterchangeControlNumber: models.PaymentRequestToInterchangeControlNumber{
+				PaymentRequestID: paymentRequest.ID,
+				EDIType:          models.EDIType824,
```
reggieriser, line 18: I think it probably makes more sense to adjust the `MakePaymentRequestToInterchangeControlNumber` to set a default `EDIType` that can be overridden if needed through an assertion.  Then, you wouldn't be forced to set it like you are here (and I don't think you care what it's set to in this case, as long as it's a legal value).

pkg/services/invoice/process_edi824_test.go
pearl-truss: The [processor](https://github.com/transcom/mymove/pull/6462/files#diff-29b62994ec2bb4bcb70f80082f915a18289eacbde732051c57868ff8d1f97a8fL74) is looking at the `PaymentRequestToInterchangeControlNumber` table to find the payment request that has an ICN field that matches the GCN in the OTI segment, so these lines in the test are creating that record so that the right payment request is found

jacquelineIO: ah OK, that would be an EDI 858 though, but that's fine for the test.

pearl-truss: That's a good point! I'll update to use 858s so it's clearer

migrations/app/schema/20210419195440_add_editype_to_paymentRequestToInterchange_table.up.sql
```diff
@@ -1,4 +1,4 @@
 ALTER TABLE payment_request_to_interchange_control_numbers
-    ADD COLUMN edi_type varchar not null;
+    ADD COLUMN edi_type edi_type DEFAULT '858';
```
reggieriser, line 2: Do we want to keep the default permanently?  If there's no sense that we should default to `858` for new records, we should drop the default and make it `NOT NULL` again after converting existing records.

pkg/models/edi_errors_test.go
```diff
@@ -19,7 +19,6 @@ func (suite *ModelSuite) TestEdiErrors() {
 		InterchangeControlNumber: 5,
 		EDIType:                  models.EDIType997,
 	}
-	suite.MustCreate(suite.DB(), &prICN)
```
reggieriser, line 21: Hmm, I was thinking more that we didn't need to create any of the DB records above, not just the `models.PaymentRequestToInterchangeControlNumber` one above.  Just have the places below that reference the `pr` or `prICN` struct instead reference dummy UUIDs created with `uuid.Must(uuid.NewV4())`.  We don't care what it actually is or that it's in the DB as long as it looks like a real UUID for validation purposes.

pkg/testdatagen/make_payment_request_to_interchange_control_number.go
```diff
@@ -29,9 +29,15 @@ func MakePaymentRequestToInterchangeControlNumber(db *pop.Connection, assertions
 		log.Panic(fmt.Errorf("Errors encountered getting random interchange control number: %v", err))
 	}
 
+	ediType := assertions.PaymentRequestToInterchangeControlNumber.EDIType
+	if len(ediType.String()) == 0 {
+		ediType = models.EDIType858
+	}
```
reggieriser, line 35: Nit: I think `mergeModels` below will handle this the right way without the code above.  It should overwrite any values in `pr2icn` with any non-zero values in the assertion struct (essentially what this is doing).

reggieriser (commented):
Thanks for making the changes.  Had a few more questions/comments.

migrations/app/schema/20210419195440_add_editype_to_paymentRequestToInterchange_table.up.sql
pearl-truss: I've been having a hard time tracking down how to this, I've tried variations of this:

```
ALTER TABLE payment_request_to_interchange_control_numbers 
    ALTER COLUMN edi_type DROP DEFAULT;

ALTER TABLE payment_request_to_interchange_control_numbers 
    ALTER COLUMN edi_type edi_type NOT NULL;
```

Should I be creating this in a separate migration?

reggieriser: I think this should work:
```sql
ALTER TABLE payment_request_to_interchange_control_numbers
    ALTER COLUMN edi_type DROP DEFAULT,
    ALTER COLUMN edi_type SET NOT NULL;
```
I tried running all the migrations but yours (via `db_dev_e2e_populate`), then adding a record to this table, then running your migration with the above added to it.  It seemed to work and the new record and table structure looked good, but double check it.

pearl-truss: Thanks! I was able to confirm this worked!

reggieriser (approved):
Additional changes LGTM!  🚢

PR #6463, opened 2021-04-20
[MB-7725] Colocate all #nosec instances to be with annotation

Ronolibert:
## Description

Colocated all #nosec instances to be with annotation to ensure that ato-linter does not flag it as a false positive

## Reviewer Notes
- there is one instance at `mymove/pkg/cli/gex_sftp.go:22:51` which will need annotation. Annotation will be done in separate pr
- having a comment above a group of constant declarations was needed for #nosec to have expected behavior. Without it, it was expecting the required comment for a constant to be above the annotation block

## Setup
`go install ./cmd/...`
`ato-linter ./pkg/...`
Expected results: It should have 1 lint error located at `mymove/pkg/cli/gex_sftp.go:22:51`
Annotation for this will be in a separate PR

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7725) for this change

### DIFF
--- a/pkg/auth/cookie.go
+++ b/pkg/auth/cookie.go
@@ -36,6 +36,7 @@ func (e *errInvalidHostname) Error() string {
 	return fmt.Sprintf("invalid hostname %s, must be one of %s, %s, or %s", e.Hostname, e.MilApp, e.OfficeApp, e.AdminApp)
 }
 
+// GorillaCSRFToken is the name of the base CSRF token
 //RA Summary: gosec - G101 - Password Management: Hardcoded Password
 //RA: This line was flagged because it detected use of the word "token"
 //RA: This line is used to identify the name of the token. GorillaCSRFToken is the name of the base CSRF token.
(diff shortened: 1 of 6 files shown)


pkg/auth/cookie.go
```diff
@@ -36,6 +36,7 @@ func (e *errInvalidHostname) Error() string {
 	return fmt.Sprintf("invalid hostname %s, must be one of %s, %s, or %s", e.Hostname, e.MilApp, e.OfficeApp, e.AdminApp)
 }
 
+// GorillaCSRFToken is the name of the base CSRF token
```
Ronolibert, line 39: `golint` expects a comment in the format of {VariableName}...

This is an edge case where the linter is interpreting the entire comment block and only looking at the first line of it. I had to move the comment to the top to satisfy this linter.

Another option is to move the annotation below the variable but we've standardized annotations always being above the line of code it is annotating

pkg/cli/auth.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/spf13/viper"
 )
 
+// Set of flags used for authentication
```
Ronolibert, line 13: The comment for this group of constant declarations were needed in order for #nosec to have expected behavior of disabling the rule for the rule violation below it. Without this comment, `golint` would error about `// ClientAuthSecretKeyFlag is the Client Auth Secret Key Flag` not existing

Ronolibert: Also to note, having #nosec in this position does not disable the rule for the entire file but only for its nearest rule violation. I tested this by adding a violation a few lines below with no #nosec disabling and it correctly found that lint error

felipe-lee (approved):
LGTM. Ran in master and in branch to see the diff in issues that were caught and I saw what was expected.

PR #6466, opened 2021-04-20
[MB-7656] Filter by gbloc for the Services Counseling Queue

shimonacarvalho:
## Description

This ticket is to ensure that filtering by GBLOC works. 

If you've accessed the Office UI as a TOO, you'll notice we only see one GBLOC, the one that matches the office user's transportation office. So you can't really sort GBLOCs if you're only getting one GBLOC. 

The GBLOC sorting comes into play for USMC office users. We have a separate transportation office for USMC office users. If you sign in as office user from that GBLOC, you should see all moves associated with US Marine service users. In that case, there could be multiple GBLOCs so you can test the filter. 

To test
- Go to `officelocal:3000`
- Sign in as a USMC office user. In devlocal local sign-in, you'll one available called `too_tio_role_usmc@office.mil`
- Go to the services counseling page `officelocal:3000/counseling/queue` and filter the 2 users you should see using the gbloc.

Also there's a new test in pkg/services/order.

## References

* [Jira story](https://dp3.atlassian.net/browse/mb-7656) for this change

### DIFF
--- a/pkg/handlers/ghcapi/queues.go
+++ b/pkg/handlers/ghcapi/queues.go
@@ -169,6 +169,7 @@ func (h GetServicesCounselingQueueHandler) Handle(params queues.GetServicesCouns
 		DodID:                  params.DodID,
 		LastName:               params.LastName,
 		DestinationDutyStation: params.DestinationDutyStation,
+		OriginGBLOC:            params.OriginGBLOC,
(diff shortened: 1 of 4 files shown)


eamahanna (approved):
This works well!

PR #6475, opened 2021-04-21
MB-7431 Add `getMoveTaskOrder` endpoint to Prime API

sandy-wright:
## Description

This PR adds the `getMoveTaskOrder` endpoint to the Prime API and Prime API Client (the CLI tool). It also adds some minor refactors along the way. Note that the Prime API version of the endpoint can only see enabled Prime-available moves, whereas the Support API endpoint can see all moves.

## Setup

Populate your DB and start your server:

```sh
```

Now you'll need to find some moves. You can do this either using a [DB viewer](https://github.com/transcom/mymove/wiki/Configure-Postico-or-TablePlus-to-connect-to-mymove-DB) or you can run your client (`make client_run`) and search for moves using the Admin and Office UIs. Try to find (or make) three moves:

1. One that is enabled and Prime-available
2. One that is enabled but not available to the Prime
3. One that is hidden/disabled (should only be visible in the Admin client or in the DB directly)

Once you have the IDs for the moves you're going to test with, set up either [Postman](https://github.com/transcom/prime_api_deliverable/wiki/Using-Postman) or the [Prime API Client](https://github.com/transcom/prime_api_deliverable/wiki/Prime-API-Client) for testing. The PAC commands for these endpoints are:

```sh
prime-api-client get-move-task-order --insecure --id <uuid> | jq
prime-api-client support-get-move-task-order --insecure --id <uuid> | jq
```

_(Remember to `rm -rf bin/prime-api-client && make bin/prime-api-client` first.)_

Test that:

* The Prime API `getMoveTaskOrder` endpoint can only see the first move, and that it shows the full details for this move.
* The Prime API `getMoveTaskOrder` returns 404 for the two remaining moves.
* The Support API `getMoveTaskOrder` endpoint can see full details for all three moves.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7431) for this change.


cmd/prime-api-client/main.go
```diff
@@ -273,26 +291,21 @@ func main() {
 	support.InitUpdatePaymentRequestStatusFlags(updatePaymentRequestStatusCommand.Flags())
 	root.AddCommand(updatePaymentRequestStatusCommand)
 
-	getMoveTaskOrder := &cobra.Command{
+	supportGetMoveTaskOrder := &cobra.Command{
```
eamahanna, line 294: good call updating this name

cmd/prime-api-client/prime/get_move_task_order.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func checkGetMTOConfig(v *viper.Viper, args []string, logger *log.Logger) error {
+	err := utils.CheckRootConfig(v)
+	if err != nil {
+		logger.Fatal(err)
+	}
+
+	if v.GetString(utils.IDFlag) == "" && (len(args) < 1 || len(args) > 0 && !utils.ContainsDash(args)) {
+		logger.Fatal(errors.New("get-move-task-order expects an ID to be passed in"))
```
eamahanna, line 34: should this check the length or format of the id to make sure it is a uuid?

cmd/prime-api-client/prime/get_move_task_order.go
```diff
@@ -0,0 +1,98 @@
...
+	err := checkGetMTOConfig(v, args, logger)
+	if err != nil {
+		logger.Fatal(err)
+	}
+
+	// Get the UUID that was passed in
+	moveID := v.GetString(utils.IDFlag)
+	var getMTOParams mto.GetMoveTaskOrderParams
+	getMTOParams.MoveTaskOrderID = moveID
+	getMTOParams.SetTimeout(time.Second * 30)
```
eamahanna, line 63: Why 30 seconds?

cmd/prime-api-client/support/get_move_task_order.go
```diff
@@ -30,8 +30,8 @@ func checkGetMTOConfig(v *viper.Viper, args []string, logger *log.Logger) error
 		logger.Fatal(err)
 	}
 
-	if v.GetString(utils.FilenameFlag) == "" && (len(args) < 1 || len(args) > 0 && !utils.ContainsDash(args)) {
```
eamahanna, line 33: Nice update!

cmd/prime-api-client/support/get_move_task_order.go
```diff
@@ -30,8 +30,8 @@ func checkGetMTOConfig(v *viper.Viper, args []string, logger *log.Logger) error
...
-		logger.Fatal(errors.New("get-mto expects a file to be passed in"))
+	if v.GetString(utils.IDFlag) == "" && (len(args) < 1 || len(args) > 0 && !utils.ContainsDash(args)) {
```
eamahanna, line 33: Same comment above about checking the length of the ID

swagger/prime.yaml
```diff
@@ -60,6 +60,40 @@ paths:
...
+          name: moveTaskOrderID
+          required: true
+          type: string
+      responses:
+        '200':
+          description: Successfully retrieve an individual move task order.
+          schema:
+            $ref: '#/definitions/MoveTaskOrder'
+        '400':
+          $ref: '#/responses/InvalidRequest'
```
eamahanna, line 88: I don't think this error is ever returned by the handler. Maybe I missed it though. Where is this returned?

cmd/prime-api-client/prime/get_move_task_order.go
sandy-wright: ooh, that's a really good idea. Will do!

cmd/prime-api-client/prime/get_move_task_order.go
sandy-wright: I'm not sure exactly, it seems to be the standard timeout for these commands. All I do know is that it breaks if you change it -- maybe @shimonacarvalho knows why?

swagger/prime.yaml
sandy-wright: Well, Swagger itself can return a 400, so even if it's not returned from the handler, the user can still get this response code from the endpoint if they mess around with it. I'm not sure if having it defined here makes a difference in that case though... I'll poke around at it and see if there's any benefit to having the defined response or not

cmd/prime-api-client/prime/get_move_task_order.go
shimonacarvalho: Not sure either, it's what was set before when we inherited this client.

pkg/handlers/primeapi/move_task_order_test.go
```diff
@@ -405,6 +405,45 @@ func (suite *HandlerSuite) TestFetchMTOUpdatesHandlerLoopIteratorPointer() {
...
+			MoveTaskOrderID: successMove.ID.String(),
+		}
+		response := handler.Handle(params)
+		suite.IsNotErrResponse(response)
+		suite.IsType(&movetaskorderops.GetMoveTaskOrderOK{}, response)
+
+		moveResponse := response.(*movetaskorderops.GetMoveTaskOrderOK)
+		movePayload := moveResponse.Payload
+		suite.Equal(movePayload.ID.String(), successMove.ID.String())
+		suite.NotNil(movePayload.AvailableToPrimeAt)
```
shimonacarvalho, line 428: Since we have so many situations where the date returns a non-nil date that is still not populated, can you make a stronger check on this date? To ensure it isn't 01/01/0001?

sandy-wright: good point, will do!

swagger/prime.yaml
sandy-wright: I don't think it makes a difference, so I went ahead and removed it!

NamibiaTorres (approved):
Ok, I followed the instructions and was able to return what was expected using the prime-api-client and the support api. I recommend having someone else review with 3 more moves (I found my moves in the DB) to double check. Code looks good and comments are descriptive. Great job!

pkg/services/move_task_order.go
```diff
@@ -53,14 +53,10 @@ type MoveTaskOrderChecker interface {
 	MTOAvailableToPrime(moveTaskOrderID uuid.UUID) (bool, error)
 }
 
-// ListMoveTaskOrderParams is a public struct that's used to pass filter arguments to the ListMoveTaskOrders and ListAllMoveTaskOrders queries
-type ListMoveTaskOrderParams struct {
-	IsAvailableToPrime bool   // indicates if all MTOs returned must be Prime-available (only used in ListAllMoveTaskOrders)
+// MoveTaskOrderFetcherParams is a public struct that's used to pass filter arguments to
+// ListMoveTaskOrders, ListAllMoveTaskOrders, and FetchMoveTaskOrder queries
+type MoveTaskOrderFetcherParams struct {
```
shimonacarvalho, line 58: Thanks for combining the struct and using one!

eamahanna (approved):
LGTM :) Thanks for making those changes

shimonacarvalho:
I noticed that in the service items - crating dimensions are empty where they should exist. Might not be in your PR, possibly inherited from the support getMTO but can you file a ticket if so?

```
            "crate": {
                "height": 0,
                "id": "00000000-0000-0000-0000-000000000000",
                "length": 0,
                "width": 0
            },
```

shimonacarvalho (approved):
LGTM - I mentioned two things which could be handled in another ticket since I don't think they were introduced here.

sandy-wright:
> I noticed that in the service items - crating dimensions are empty where they should exist. Might not be in your PR, possibly inherited from the support getMTO but can you file a ticket if so?
> 
> ```
>             "crate": {
>                 "height": 0,
>                 "id": "00000000-0000-0000-0000-000000000000",
>                 "length": 0,
>                 "width": 0
>             },
> ```
> 
> Similarly the dates show the 0001-01-01 pattern
> 
> ```
>             "primeEstimatedWeightRecordedDate": "0001-01-01",
>             "requiredDeliveryDate": "0001-01-01",
> ```

I looked into these, and it looked like they were empty because there weren't any values in the database (for me, at least). It was especially weird for the crating service items, but looks like some Domestic Packing/Unpacking service items were set up either without an item or a crate, or with just one or the other (there should always be an item AND a crate). So I _think_ these are mostly devseed issues - I'll make a ticket to fix it!

PR #6480, opened 2021-04-22
MB-7649 Edit allowances backend

donaldthai:
## Description

> Given I’m the milmove GHC API  
> When an office user sends a request to edit special allowances fields including: Pro-gear, Spouse Pro-gear, RME, and OCIE  
> Then I am able to process the requested changes to the customer's allowances

> AC
> 
> The GHC API endpoint for patching allowances is able to accept changes to pro-gear, spouse pro-gear, RME and OCIE fields

## Reviewer Notes

- ~~Been thinking about how we want to represent the pro-gear fields to be unset but not sure if we wanted to tackle that in this PR. I ended up returning the default max from the weight allotment if the fields aren't set so we don't break current implementation on the TXO side.~~ Looks like we will just being setting the values as default 0.

## Setup

```sh
make db_dev_e2e_populate
make server_run
make office_client_run
```

Test with Postman
- Setup here: https://github.com/transcom/mymove/wiki/Test-Admin-and-Office-APIs-with-Postman
- Endpoint you're testing is `/orders/:orderID` patch
- JSON, make sure to change the `order id` to the one you're testing
```json
{
    "issueDate": "2018-03-15",
    "reportByDate": "2018-08-01",
    "ordersType": "PERMANENT_CHANGE_OF_STATION",
    "newDutyStationId": "2d5ada83-e09a-47f8-8de6-83ec51694a86",
    "originDutyStationId": "7c065ad6-cb1f-4c2f-9835-05f259f3f459",
    "agency": "ARMY",
    "grade": "E_1",
    "proGearWeight": 2000,
    "proGearWeightSpouse": 500,
    "requiredMedicalEquipmentWeight": 4444,
    "organizationalClothingAndIndividualEquipment": true
}
```

Test in the browser
- Ask me, there's a trick using the HTTP fetch api

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7649) for this change


pkg/services/order/order_updater_test.go
```diff
@@ -188,6 +192,50 @@ func (suite *OrderServiceSuite) TestOrderUpdater() {
...
+			OrdersType:          defaultOrder.OrdersType,
+			Entitlement: &models.Entitlement{
+				ProGearWeight: swag.Int(2001),
+			},
+		}
+
+		expectedETag := etag.GenerateEtag(defaultOrder.UpdatedAt)
+		_, err := orderUpdater.UpdateOrder(expectedETag, updatedOrder)
+
+		suite.Error(err)
```
duncan-truss, line 218: Is this the error type you're expecting?  When testing the api there were other required fields necessary on the orders like orders number, tac, orders type detail, and department indicator.  I guess these are already filled in on your defaultOrder but that's unlikely for a services counselor to have these values present.

donaldthai: It's expected with the way how the code works. I could refactor a bit to make sure that we get a validation error instead. I can also clean up the unit test data a bit.

migrations/app/schema/20210420195709_adding_progear_and_spouse_entitlments.up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE entitlements
+    ADD COLUMN pro_gear_weight int4;
```
reggieriser, line 2: Nit: Any particular reason we're using the `int4` alias for these two new columns instead of the SQL standard `integer`?

migrations/app/schema/20210420195709_adding_progear_and_spouse_entitlments.up.sql
```diff
@@ -0,0 +1,6 @@
...
+COMMENT ON COLUMN entitlements.pro_gear_weight IS 'This is equipment a member needs for the performance of official duties at the next or a later destination. Members are given a weight allowance for progear that is separate from their normal weight allowance.';
+ALTER TABLE entitlements
+    ADD COLUMN pro_gear_weight_spouse int4;
```
reggieriser, line 5: There seems to be a couple of ways we handle the weight-related fields and nullability in this table.  I see the existing `authorized_weight` column is nullable with no default (like these two fields), but `required_medical_equipment_weight` is non-nullable with a default of 0 (and therefore one is a pointer and one is a value type).  Given that, I just want to make sure leaving this as nullable as opposed to defaulting to zero makes sense in this case.

pkg/services/order/order_updater_test.go
donaldthai: I can add in an assertion to check the actual type of the error too

reggieriser (commented):
Looked at this from the DB perspective and just had a couple of questions before approval.

migrations/app/schema/20210420195709_adding_progear_and_spouse_entitlments.up.sql
donaldthai: Nothing is particular. I was following the current pattern established but I can switch over to `integer`.

migrations/app/schema/20210420195709_adding_progear_and_spouse_entitlments.up.sql
donaldthai: Maybe @Ryan-Koch can pitch in here? I just went with nullable since we still show the default to the max value for pro-gear based on the weight allotment. Maybe there's a better way?

pkg/services/order/order_updater_test.go
duncan-truss: I guess my bigger point is the update would fail without those order fields present.  Is that coming in a future PR when the allowances form changes are happening?  That may surprise someone to learn they have to either split out the handlers or change the validations somehow for the services counselor to do their work.

pkg/services/order/order_updater.go
```diff
@@ -52,6 +60,30 @@ func (s *orderUpdater) UpdateOrder(eTag string, order models.Order) (*models.Ord
...
+			}
+			if entitlement.ProGearWeightSpouse != nil {
+				if weightAllotment == nil {
+					return services.NewInvalidInputError(order.ID, nil, nil, "error updating ProGearWeightSpouse: missing service member grade on Orders")
+				}
+				if *entitlement.ProGearWeightSpouse > weightAllotment.ProGearWeightSpouse {
+					return services.NewInvalidInputError(order.ID, nil, nil, fmt.Sprintf("error updating ProGearWeightSpouse: value cannot be greater than %d", weightAllotment.ProGearWeightSpouse))
+				}
+				existingOrder.Entitlement.ProGearWeightSpouse = entitlement.ProGearWeightSpouse
+			}
```
monfresh, line 80: What do you think about having these max values as validations in `ghc_entitlements.go` instead of putting the logic in the service? I find this hard to read.

pkg/services/order/order_updater_test.go
donaldthai: Oh I see what you're saying. Hmm....that means that we may have to create a new handler for this then.

pkg/services/order/order_updater.go
donaldthai: Let me get confirmation about the max values for all the ranks. Seems like everything has the same max pro-gear and spouse except for `midshipment` rank.

donaldthai: Model validations make sense to me

migrations/app/schema/20210420195709_adding_progear_and_spouse_entitlments.up.sql
```diff
@@ -0,0 +1,6 @@
...
+COMMENT ON COLUMN entitlements.pro_gear_weight_spouse IS 'This is equipment a member''s spouse needs for the performance of official duties at the next or a later destination. Members are given a weight allowance for progear that is separate from their normal weight allowance.';
```
monfresh, line 6: Should we make these default to 0 and NOT NULL? And also add max constraints of 2000 and 500? I think the syntax is `CHECK (entitlements.pro_gear_weight < 2000)`.

migrations/app/schema/20210420195709_adding_progear_and_spouse_entitlments.up.sql
Ryan-Koch: @donaldthai I think based on the conversation with Amanda in slack that we should go use the postgres default value method and set it to 0 like we did with the required medical equipment value last sprint.

pkg/services/order/order_updater.go
donaldthai: Sweet, got confirmation we can assume the max values for all ranks. So 2000 and 500 for pro-gear and spouse. 

Convo here: https://dp3.atlassian.net/browse/MB-7649?focusedCommentId=15905

migrations/app/schema/20210420195709_adding_progear_and_spouse_entitlments.up.sql
donaldthai: we want it to be inclusive though right? So `CHECK (entitlements.pro_gear_weight <= 2000)`?

donaldthai: not null and default to 0 makes sense to me

pkg/services/order/order_updater_test.go
donaldthai: @duncan-truss so I just verified on my local machine that service counseling will have at least the required information after submitting the move to ,`Needs Service Counseling`, in order to use the current handler `/orders/:orderID patch`. So I think we should be okay.

Is there something else I'm missing?

migrations/app/schema/20210420195709_adding_progear_and_spouse_entitlments.up.sql
donaldthai: script updated

jenniferlynparsons (approved):
As long as it's true that we don't need a new handler, this is looking good.

pkg/services/order/order_updater_test.go
duncan-truss: Yes there is an issue where even the TOO can't edit the allowances when missing orders information is not filled in.  You can see this with the `REQINF` move in the devseed if you try to edit the allowances you will get a 422 error.  So if you tried to update those fields without supplying the orders information too on the `SRVCSL` move you would get the same error.

Looking at the upcoming mocks we will need to be able to validate subsets of payloads of orders and allowances data depending on the user or endpoint used.  I don't think we've ever said you must edit orders before allowances https://zpl.io/am3O8z6

Screen Shot 2021-04-23 at 9 35 17 AM

reggieriser (approved):
Changes to migration LGTM.  Ran it successfully and verified checks/defaults are working.  Note that this approval is for the migration only, not the entire PR.

pkg/services/order/order_updater_test.go
monfresh: The fact that you can't edit allowances before filling in the required Orders field is existing behavior that is not caused by this PR. This behavior was introduced when I added the Orders validation in #6171. At the time, the context was that a move can't be approved if Orders are missing required info. I don't think we ever brought up the ability to edit other parts of the order, and it looks like we are just now discovering this issue.

I think this is something we should discuss next week, but I don't think it should hold up this PR

duncan-truss: I agree it doesn't have to be this PR but let's not set next week up for failure if our sprint goal is around the services counselor editing allowances.  Maybe it means a companion Jira story to the frontend form being proposed.  As currently laid out that's the risk I see in how things are sequenced, it may have assumed this PR would do those improvements.

monfresh: Ah, yes. To allow the services counselor to edit allowances, I believe a quick fix would be to add `|| order.Moves[0].Status == MoveStatusNeedsServiceCounseling` to the two validations in the Order model: `OrdersTypeDetailIsPresentAfterSubmission` and `StringIsPresentAfterSubmission`.

donaldthai: Okay, I can add that logic in real quick

donaldthai: and just to confirm, this only will let the service counselor edit the allowances but the TOO will still have trouble editing allowances. Probably need some kind of error notification to let the TOO know.

donaldthai: Correction: The allowances are only editable in the `Needs Service Counseling` status. Once that move status is `Service Counseling Completed` then we get that problem where we can't edit the allowances anymore until we sent those required fields.

monfresh: Yes, which matches the existing behavior. I don't think it would be trivial to implement that, which is why I suggested we talk about it next week with the whole team and design to see what makes sense to do.

donaldthai: Cool, if everyone is okay with it, I'll push up that quick fix get the allowances to be editable in the `Needs Service Counseling` status

pkg/models/ghc_entitlements_test.go
```diff
@@ -31,3 +31,36 @@ func (suite *ModelSuite) TestAuthorizedWeightWhenNotInDBAndHaveWeightAllotment()
...
+		entitlement := models.Entitlement{
+			ProGearWeight:       -1,
+			ProGearWeightSpouse: -1,
+		}
+		verrs, _ := entitlement.Validate(suite.DB())
+		suite.True(verrs.HasAny())
+		suite.NotNil(verrs.Get("pro_gear_weight"))
+		suite.NotNil(verrs.Get("pro_gear_weight_spouse"))
+	})
+}
```
monfresh, line 66: Great job not using the DB for these tests :)

pkg/services/order/order_updater_test.go
```diff
@@ -192,12 +192,56 @@ func (suite *OrderServiceSuite) TestOrderUpdater() {
...
 		suite.Equal(swag.Bool(true), actualOrder.Entitlement.DependentsAuthorized)
-		suite.Equal(swag.Int(1234), actualOrder.Entitlement.ProGearWeight)
-		suite.Equal(swag.Int(321), actualOrder.Entitlement.ProGearWeightSpouse)
+		suite.Equal(1234, actualOrder.Entitlement.ProGearWeight)
+		suite.Equal(321, actualOrder.Entitlement.ProGearWeightSpouse)
 		suite.Equal(2000, actualOrder.Entitlement.RequiredMedicalEquipmentWeight)
 		suite.Equal(true, actualOrder.Entitlement.OrganizationalClothingAndIndividualEquipment)
 	})
 
+	suite.T().Run("Entitlement is updated with move status Needs Service Counseling and missing submission fields", func(t *testing.T) {
```
duncan-truss, line 201: Thank you!

duncan-truss (approved):

PR #6484, opened 2021-04-22
[MB-7724] Eslint-plugin-ato checks for eslint-disable-line usage

Ronolibert:
## Description

Eslint-plugin-ato checks for eslint-disable-line usage and asks the developer to use `eslint-disable-next-line` instead

## Setup
yarn install (if nothing happens, `rm -rf node_modulesn then yarn install`
In .eslintrc.js
Change 'ato/no-unapproved-annotation': 'off', to 'ato/no-unapproved-annotation': 'error',
Add `const test = 'test'; // eslint-disable-line no-unused-vars` to any file not in the scenes or shared directory`
yarn run lint
Expected result: There should be a lint error

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7724) for this change

### DIFF
--- a/eslint-plugin-ato/no-unapproved-annotation.js
+++ b/eslint-plugin-ato/no-unapproved-annotation.js
@@ -1,9 +1,11 @@
 const REQUIRES_APPROVAL_MESSAGE_ID = 'no-unapproved-annotation';
 const NO_ANNOTATION_MESSAGE_ID = 'no-annotation';
+const NO_INLINE_DISABLE = 'no-inline-disable';
 const messages = {
   [REQUIRES_APPROVAL_MESSAGE_ID]: 'Requires annotation approval from an ISSO',
   [NO_ANNOTATION_MESSAGE_ID]:
     'Disabling of this rule requires an annotation. Please visit https://docs.google.com/document/d/1qiBNHlctSby0RZeaPzb-afVxAdA9vlrrQgce00zjDww/edit?usp=sharing',
+  [NO_INLINE_DISABLE]: 'Please use eslint-disable-next-line instead of eslint-disable-line',
 };
 
 // eslint-disable-next-line security/detect-unsafe-regex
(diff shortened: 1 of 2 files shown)


eslint-plugin-ato/no-unapproved-annotation.js
```diff
@@ -1,9 +1,11 @@
 const REQUIRES_APPROVAL_MESSAGE_ID = 'no-unapproved-annotation';
 const NO_ANNOTATION_MESSAGE_ID = 'no-annotation';
+const NO_INLINE_DISABLE = 'no-inline-disable';
 const messages = {
   [REQUIRES_APPROVAL_MESSAGE_ID]: 'Requires annotation approval from an ISSO',
```
kimallen, line 5: Is this the place to extend the message with more detail:
“Requires annotation approval from an ISSO. Please add the truss-is3 team as reviewers, ping the ISSO in #static-code-review Slack, and add the label ‘needs-is3-review’ to this PR”

Ronolibert: I believe changing that message is in https://dp3.atlassian.net/browse/MB-5827

eslint-plugin-ato/no-unapproved-annotation.js
```diff
@@ -106,7 +108,21 @@ const create = (context) => ({
         result && // It's a eslint-disable comment
         result.groups.ruleId // disabling a specific rule
       ) {
-        const [, rule] = result.input.split(' ');
+        const [disableComment, rule] = result.input.split(' ');
+        if (disableComment === 'eslint-disable-line') {
```
kimallen, line 112: So is the intention/ result that if there is an `eslint-disable-line` AND there's no annotation, both of these reports will happen?  Or would that never happen since this disable and the annotation would be considered different comment blocks?

Ronolibert: Yeah both lint errors would happen. In that scenario the dev would get the no-inline-disable message and the no-annotation message and I think we want that so that they don't have to fix lint error just to see another one

kimallen (approved):
Looks good!

eslint-plugin-ato/no-unapproved-annotation.js
kimallen: When I run it, I'm only seeing the "needs annotation" error when I try to create both errors

kimallen (changes requested):
I'm only seeing the "this needs annotation" error and not the inline disable rule.  I might be doing something wrong. . .

kimallen:
> I'm only seeing the "this needs annotation" error and not the inline disable rule. I might be doing something wrong. . .

Yup, I was doing something wrong.  I needed to remove `node_modules` and then `yarn install` to register the changes to the code.  Everything checks out and I see both errors

kimallen (approved):
All checks out - LGTM!

PR #6485, opened 2021-04-22
Kim mb 7533 enable backend linters

kimallen:
## Description

This removes two linters from the `disable` in the `golangci.yaml` config file and fixes all associated violations of those linters.  

## Reviewer Notes

Although `unused` and `gosimple` are both deprecated because they are now included in `staticcheck` linter, it appears that they made them backwards compatible to disable the equivalent rules in `staticcheck` if either of them were disabled.  

There are changes in 73 files, but fixes cover just a few issues.
The most common issues were related to removing `Sprint.f` when unnecessary and declaring a variable inline with a value instead of explicitly by type alone. Also removing boolean comparison when the bool value or !bool value was sufficient. Besides the config file, all changes are related to specific linter findings.

## Setup

Run the linter and there should be no violations:
`pre-commit run -v --all-files golangci-lint`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7533) for this change


pkg/parser/pricing/parse.go
```diff
@@ -148,7 +148,7 @@ type ParamConfig struct {
 // The index MUST match the sheet that is being processed. Refer to file comments or XLSX to
 // determine the correct index to add.
 func InitDataSheetInfo() []XlsxDataSheetInfo {
-	xlsxDataSheets := make([]XlsxDataSheetInfo, xlsxSheetsCountMax, xlsxSheetsCountMax)
+	xlsxDataSheets := make([]XlsxDataSheetInfo, xlsxSheetsCountMax)
```
lynzt, line 151: do you know why `xlsxSheetsCountMax` was listed twice?

lynzt (approved):
LGTM 🚢 

I think these fixes makes the code better and more readable so I'm glad we're addressing these and preventing them in the future

pkg/parser/pricing/parse.go
kimallen: The third argument is the capacity of the slice, so I think someone added it and it's unnecessary or redundant when it equals the second argument (the length

PR #6492, opened 2021-04-23
Change all the Truss offices in experimental to use AGFM gbloc

Ryan-Koch:
## Description

AGFM all the Truss entries!

### DIFF
--- a/migrations/app/secure/20210423140812_update-gbloc-truss-offices-experimental.up.sql
+++ b/migrations/app/secure/20210423140812_update-gbloc-truss-offices-experimental.up.sql
@@ -1,0 +1,5 @@
+-- This is a migration that will ultimately impact experimental. It seeks to change the gbloc of a commonly used
+-- test transportation office.
+UPDATE transportation_offices
+SET gbloc = 'AGFM'
+WHERE name = 'Truss'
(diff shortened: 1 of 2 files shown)


migrations/app/secure/20210423140812_update-gbloc-truss-offices-experimental.up.sql
```diff
@@ -0,0 +1,5 @@
+-- This is a migration that will ultimately impact experimental. It seeks to change the gbloc of a commonly used
+-- test transportation office.
+UPDATE transportation_offices
+SET gbloc = 'AGFM'
+WHERE name = 'Truss'
```
tinyels, line 5: Ultimately, I'd like us to only have one Truss office. And perhaps a uniqueness constraint on the name? I don't want to block MilMob today, but can we add some clean up work for next sprint?

Ryan-Koch: I agree, this is very much a quick an dirty thing for milmob. I think it'd definitely be worth while to clean it up next sprint.

tinyels (approved):

reggieriser (approved):
LGTM.  I ran it against local and it worked fine.  I checked stg/prd and saw it was a no-op and that exp matched this one.

PR #6500, opened 2021-04-26
MB-7366 Review Service Items aria labelling refactor

Shkeating:
## Description

this branch resolves some critical a11y testing errors in the Review Service Items component(s). It moves aria labels to the Button component and adds empty alt text to icons in unstyled buttons to ensure their purpose is not read out twice by a screen reader. It also adjusts the form labeling on the rejection fields.

## Reviewer Notes

cypress seems to dislike the htmlFor attribute, and reads it as a duplicated id, so I'll have to change the approach here if i see an integration test failure.

## Setup

```sh
yarn storybook
```

[http://localhost:6006/?path=/story/office-components-reviewserviceitems--basic ](http://localhost:6006/?path=/story/office-components-reviewserviceitems--basic )

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7366?atlOrigin=eyJpIjoiYWM5ZjY0Yzc4N2M5NGFhNGEyMmNhMDU3MzVjMmVkYmIiLCJwIjoiaiJ9) for this change
* [Buttons must have discernible text article](https://dequeuniversity.com/rules/axe/4.1/button-name?application=axeAPI) explains more about the approach used.

## Screenshots
no visual changes ~

### DIFF
--- a/src/components/Office/ReviewServiceItems/ServiceItemCard.jsx
+++ b/src/components/Office/ReviewServiceItems/ServiceItemCard.jsx
@@ -196,7 +196,7 @@ const ServiceItemCard = ({
                       aria-label="Clear status"
                     >
                       <span className="icon">
-                        <FontAwesomeIcon icon="times" title="Clear status" />
+                        <FontAwesomeIcon icon="times" title="Clear status" alt=" " />
                       </span>
                       <span aria-hidden="true">Clear selection</span>
                     </Button>
--- a/src/components/Office/ReviewServiceItems/ServiceItemCard.jsx
+++ b/src/components/Office/ReviewServiceItems/ServiceItemCard.jsx
@@ -160,7 +160,7 @@ const ServiceItemCard = ({
 
                     {values.status === DENIED && (
                       <FormGroup>
-                        <Label htmlFor="rejectReason">Reason for rejection</Label>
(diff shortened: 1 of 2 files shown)


Ronolibert (approved):

src/components/Office/ReviewServiceItems/ServiceItemCard.jsx
```diff
@@ -193,11 +193,12 @@ const ServiceItemCard = ({
                       data-testid="clearStatusButton"
                       className={styles.clearStatus}
                       onClick={handleFormReset}
+                      aria-label="Clear status"
                     >
                       <span className="icon">
-                        <FontAwesomeIcon icon="times" title="Clear status" aria-label="Clear status" />
+                        <FontAwesomeIcon icon="times" title="Clear status" />
```
duncan-truss, line 199: You had empty alt attributes in the other FontAwesomeIcon components, is there guidance on when one should be specified and not?

Shkeating: This one should actually have it too, thanks for catching this. As it stands, It will read the aria label for the button, and then read out the file path in the icon. which is no good! The reason why the label is relocated in this instance to the button is because the way our a111y tests run, it was looking for a label inside the button element, and it was failing because it couldn't find one. this solution works similarly, but also gets the test to pass 😎

PR #6511, opened 2021-04-26
…

kimallen:
Fixes service test to check for not found error and returns a NotFound response from the handler

## Description

A bug was reported that the api response for fetching an access code for a service member should return an error.  This ensures that the `FetchAccessCodeHandler` returns a `NotFound` error if there is no record found and creates a test for that.  It also adjusts the service test to check for that error.

## Reviewer Notes

The second reported case says that the `ValidateAccessCodeHandler` should return an error.  As a GET, and based on how it is handled on the frontend, this is meant to return an empty `AccessCode.code` if it is not valid and an `AccessCode` with values if valid.  It should only return an error if there's an authorization or authentication issue hitting that endpoint, or an internal server error.
Some discussion [here](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1619470549187900)

If there should be a redesign of how we check for a valid access code, that should be its own story.

## Setup

`make client_test`
or
`go test ./pkg/services/accesscode -v -testify.m TestFetchAccessCode_FetchNotFound`
`go test ./pkg/handlers/internalapi -v -testify.m TestFetchAccessCodeHandler_Failure`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-6918) for this change

### DIFF
--- a/pkg/services/accesscode/access_code_fetcher_test.go
+++ b/pkg/services/accesscode/access_code_fetcher_test.go
@@ -30,5 +30,5 @@ func (suite *AccessCodeServiceSuite) TestFetchAccessCode_FetchNotFound() {
 	fetchAccessCode := NewAccessCodeFetcher(suite.DB())
 	_, err := fetchAccessCode.FetchAccessCode(*serviceMemberID)
 	suite.Error(err)
-	suite.Equal(sql.ErrNoRows, err.Error())
+	suite.Equal(sql.ErrNoRows, err)
 }
--- a/pkg/services/accesscode/access_code_fetcher_test.go
+++ b/pkg/services/accesscode/access_code_fetcher_test.go
@@ -1,6 +1,8 @@
 package accesscode
 
 import (
+	"database/sql"
+
 	"github.com/transcom/mymove/pkg/models"
 	"github.com/transcom/mymove/pkg/testdatagen"
 )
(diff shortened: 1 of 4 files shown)


monfresh (approved):
I think this is probably fine for now. This whole access codes feature is implemented in an unusual way to me. It doesn't seem to be fully working anymore. The form to enter the code doesn't show up, so we'll need to revisit this at some point.

pkg/services/accesscode/access_code_fetcher_test.go
```diff
@@ -22,10 +22,11 @@ func (suite *AccessCodeServiceSuite) TestFetchAccessCode_FetchAccessCode() {
...
-func (suite *AccessCodeServiceSuite) TestFetchAccessCode_FetchEmptyAccessCode() {
+func (suite *AccessCodeServiceSuite) TestFetchAccessCode_FetchNotFound() {
 	user := testdatagen.MakeDefaultServiceMember(suite.DB())
 	serviceMemberID := &user.ID
 	fetchAccessCode := NewAccessCodeFetcher(suite.DB())
-	ac, _ := fetchAccessCode.FetchAccessCode(*serviceMemberID)
-	suite.Equal(ac.Code, "")
+	_, err := fetchAccessCode.FetchAccessCode(*serviceMemberID)
+	suite.Error(err)
+	suite.Equal("sql: no rows in result set", err.Error())
```
sandy-wright, line 31: You can use the `sql.ErrNoRows` constant to capture this message instead of hardcoding the string (I just recently learned about this and it's amazing)

felipe-lee (approved):
Yeah I think this looks good and fixes the jira issue, but agree that if there's parts that aren't working on the UI, that'll have to be reviewed later since it's out of scope for this.

pkg/services/accesscode/access_code_fetcher_test.go
kimallen: Ooh, thank you- I didn't know about that!

PR #6529, opened 2021-04-28
MB-7947 Allow Admins to update an office user's Transportation Office

sandy-wright:
## Description

This PR adds the "Transportation Office" field to the Office User edit page in the Admin UI.

## Reviewer Notes

I had to burn down and rewrite some of these tests. There was some funky behavior going on with Roles in testdatagen, so you'll notice that I'm not really testing those updates. I started poking at it, but that work really spiraled so I decided it was out of scope for this PR.

## Setup

* Run your server and admin client:

```sh
make server_run
make admin_client_run
```

* Navigate to http://adminlocal:3000/ and log in.
* Select an Office user from the list and click on their row.
    * If you have no Office users, you'll need to create one first.
* Once you're in the details page, click on the "Edit" button in the top right.
* Scroll down to the Transportation Office field. Verify that you can type in and enter a new office.
* Click "Save."
* Verify that you got a success message, and that you see the new office when you look at the user in the Office Users list.
* Click on that user's role again.
* Verify that you still see the new Transportation Office in the details page.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7947) for this change.

## Screenshots

Screen Shot 2021-04-28 at 3 30 15 PM


cypress/integration/admin/officeUsers.js
```diff
@@ -144,6 +144,13 @@ describe('Office Users Edit Page', function () {
 
     cy.get('input[id="firstName"]').clear().type('Edit');
     cy.get('input[id="lastName"]').clear().type('CypressUser');
+    cy.get('input[id="transportationOfficeId"]').type('Camp');
+    cy.wait(5000); // we have to wait for the autocomplete to give us something
```
monfresh, line 148: If adding this test means adding 5 seconds to our test suite, I don't think it's worth adding. I think it would be sufficient to add a JS unit test that verifies the page has this field.

sandy-wright: oh, that's a good point! This is one of the inputs that can be flaky too (since it has to wait for a DB call). I'll change the check to just make sure it's enabled instead.

pkg/handlers/adminapi/office_users_test.go
```diff
@@ -279,151 +277,73 @@ func (suite *HandlerSuite) TestCreateOfficeUserHandler() {
...
-	params := officeuserop.UpdateOfficeUserParams{
-		HTTPRequest: req,
-		OfficeUser: &adminmessages.OfficeUserUpdatePayload{
-			FirstName:      &officeUser.FirstName,
-			MiddleInitials: officeUser.MiddleInitials,
-			LastName:       &officeUser.LastName,
-			Telephone:      &officeUser.Telephone,
-		},
+	queryBuilder := query.NewQueryBuilder(suite.DB())
+	updater := officeuser.NewOfficeUserUpdater(queryBuilder)
```
monfresh, line 281: Thanks for improving these tests. One suggestion is to avoid using the DB in these tests because the OfficeUserUpdater is already tested separately. Testing the DB interaction again in these tests is redundant. We should be able to mock the updater to return a stubbed office user, and then we can verify that the payload is returning the proper fields and values.

eamahanna: Does milmove have guidelines on when we use mocks in tests vs hitting the db?

pkg/services/office_user/office_user_updater.go
```diff
@@ -45,6 +45,22 @@ func (o *officeUserUpdater) UpdateOfficeUser(id uuid.UUID, payload *adminmessage
...
+		}
+		// Use FetchOne to see if we have a transportation office that matches the provided id
+		var transportationOffice models.TransportationOffice
+		fetchErr := o.builder.FetchOne(&transportationOffice, transportationIDFilter)
+
+		if fetchErr != nil {
+			return nil, nil, fetchErr
+		}
+
+		foundUser.TransportationOfficeID = uuid.FromStringOrNil(transportationOfficeID)
```
eamahanna, line 61: 🎉

src/pages/Admin/OfficeUsers/OfficeUserEdit.jsx
```diff
@@ -28,6 +38,15 @@ const OfficeUserEdit = (props) => (
         ]}
       />
       <RolesCheckboxInput source="roles" />
+      <ReferenceInput
+        label="Transportation Office"
+        reference="offices"
+        source="transportationOfficeId"
+        perPage={500}
```
eamahanna, line 45: Do you know if 500 is what we have been using in other places?

pkg/handlers/adminapi/office_users_test.go
monfresh: We will soon! I am giving a testing talk at our huddle tomorrow.

sandy-wright: Hmmm, the handlers are so tightly connected with the DB calls even thought they're one degree removed. I have also seen times where mocking the DB calls obscures errors in the handler.

For example, handlers are often converting input and reshaping it into the model type so it can call the service object. There are times where this process has bugs, but if I mock the updater and fetcher to return the object I expect, then I'm obfuscating the fact that data isn't actually getting through to the DB the way I wanted. 

I _think_ this is a sign that we may want to refactor our relationship between handlers and service objects - or at least find a way to break things down for more modular tests. But with the current state of things, I feel a little uncomfortable putting in a mock here just because I've had to fix bugs that were hidden by mocks like this before. 

All that said, if it's possible to mock and say "have this function send back the same input it got, whatever that was" as opposed to saying "send back this exact object I have set up right now," that would quell my worries 😄 . It's not that I need to hit the DB, I just need to know that I'm getting back the result of the handler's data manipulation and not my own

src/pages/Admin/OfficeUsers/OfficeUserEdit.jsx
sandy-wright: Yeah, I copy and pasted this directly from the OfficeUserCreate page. I don't think it's used anywhere else though

pkg/handlers/adminapi/office_users_test.go
sandy-wright: Okay, I played around with it a little more! In order for it to work the way I want, I need to be veery careful about input references so I don't accidentally change my mocked expectations. 

For example, the Office user updater takes in a pointer for the payload, so I can't just use the same payload I'm passing in as the _expected_ input. I need to make a copy first so that I can ensure that any unexpected changes are caught. It's a little delicate.

pkg/handlers/adminapi/office_users_test.go
```diff
@@ -279,151 +278,85 @@ func (suite *HandlerSuite) TestCreateOfficeUserHandler() {
...
-			Name:     &payloadRoleName1,
-			RoleType: &payloadRoleType1,
-		}
-		payloadRoleName2 := "name2"
-		payloadRoleType2 := string(role2.RoleType)
-		payloadRole2 := &adminmessages.OfficeUserRole{
-			Name:     &payloadRoleName2,
-			RoleType: &payloadRoleType2,
-		}
+	officeUser := testdatagen.MakeOfficeUser(suite.DB(), testdatagen.Assertions{
```
monfresh, line 289: Did you mean to stub this office user? What is the value of storing it in the DB if we're going to mock the updater service?

sandy-wright: I didn't stub this part because the UpdateRoles service object isn't mocked, so it'll expect a real User. I have that in another ticket because there's a bunch of other wonky stuff going on with User Roles here, and messing with it was taking me down a deep rabbit hole 😅

eamahanna (approved):
LGTM :)

pkg/handlers/adminapi/office_users_test.go
monfresh: Hmm. If I add `Stub: true`, all of these tests still pass.

PR #6530, opened 2021-04-28
MB-7736 failed gex connection should not mark payment reqeust status as edi error

gidjin:
## Description

Currently when the job runner processes a Payment Request and fails to send it to GEX due to connection issue the status of the request is changed. This should not be the case as the only option for us is to retry later. This PR fixes that so the status stays `REVIEWED` so that it will be retried later.

## Reviewer Notes

I moved some tests from the payment request reviewed processor into a file for send to syncada helper, since that's what was  really being tested.

## Setup

Make sure `GEX_URL` is not set in your environment
change line 601 of `cmd/milmove/serve.go` to return a non 200 status

```golang
			w.WriteHeader(http.StatusForbidden)
```

setup the server

```sh
make db_dev_e2e_populate
make server_run
make office_client_run
```

Mark existing payment requests as sent
```sql
update payment_requests SET status = 'SENT_TO_GEX';
```

Create a payment request
```sh
rm -f bin/prime-api-client
make bin/prime-api-client
prime-api-demo RDY4PY
```

Approve the payment request

Run the job runner and see that the status does not change

```sh
go run ./cmd/milmove-tasks process-edis --send-to-syncada
```

```sql
select * from payment_requests where id = '<id of pr>';
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7736) for this change

### DIFF
--- a/pkg/services/payment_request/payment_request_reviewed_processor.go
+++ b/pkg/services/payment_request/payment_request_reviewed_processor.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/benbjohnson/clock"
 	"github.com/gobuffalo/pop/v5"
+	"github.com/gofrs/uuid"
 	"go.uber.org/zap"
 
 	"github.com/transcom/mymove/pkg/services/invoice"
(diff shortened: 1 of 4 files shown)


pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -18,6 +19,16 @@ import (
...
 )
 
+//GexSendError is returned when there is an error sending an EDI to GEX
+type GexSendError struct {
+	paymentRequestID uuid.UUID
+	err              error
+}
+
+func (e GexSendError) Error() string {
+	return fmt.Sprintf("error sending the following EDI (PaymentRequest.ID: %s) to Syncada: %s", e.paymentRequestID, e.err.Error())
```
mboren, line 29: would it make more sense to replace Syncada with GEX in this message?

mboren (approved):
Looks good to me. I added one comment about an error message I wasn't sure about.

pkg/services/payment_request/payment_request_reviewed_processor.go
gidjin: Good call

PR #6531, opened 2021-04-28
Add migration for Marty's CAC

mboren:
## Description

I followed the instructions here: https://github.com/transcom/mymove/wiki/use-mtls-with-cac


reggieriser (approved):
LGTM! I verified the exp/stg migrations looked right along with a no-op on prod. Local looks good too. Ran both local and stg migrations and they were successful. The `client_certs` table was populated in both cases as expected. 🚀

PR #6533, opened 2021-04-29
MB-7957 fix bug that bails if one payment request send fails

gidjin:
## Description

Currently the `PaymentRequestReviewedProcessor` will bail out if a single payment request fails to process for any reason. However, the intention is that it should log errors and move on to the next payment request for processing.

## Reviewer Notes

The test `TestProcessLockedReviewedPaymentRequest` seems to have become flakey for me locally. Would appreciate more 👀  on it.

## Setup

Make sure `GEX_URL` is not set in your environment
change line 601 of `cmd/milmove/serve.go` to return a non 200 status

```golang
			w.WriteHeader(http.StatusForbidden)
```

setup the server

```sh
make db_dev_e2e_populate
make server_run
```

```sh
go run ./cmd/milmove-tasks process-edis --send-to-syncada
```

```sql
select * from payment_requests where status = 'EDI_ERROR';
```

There should be 2 marked as `EDI_ERROR` instead of just one in the e2e data.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7957) for this change


mboren (approved):
Looks good to me. The annotation removal and simplifications to test logic are great.

monfresh:
The locked PR test is also failing for me intermittently. We should not knowingly commit a flakey test, so I would vote to remove it if we can't get it to pass reliably. I remain skeptical of the value of this test. It feels like we're testing Postgres functionality.

pkg/services/payment_request/payment_request_reviewed_processor_test.go
```diff
@@ -755,7 +730,7 @@ func (suite *PaymentRequestServiceSuite) TestProcessLockedReviewedPaymentRequest
 			paymentRequest, err := fetcher.FetchPaymentRequest(pr.ID)
 			suite.NoError(err)
 			if i == 0 {
-				suite.Equal(models.PaymentRequestStatusSentToGex, paymentRequest.Status)
+				suite.Equal(models.PaymentRequestStatusPaid, paymentRequest.Status)
```
monfresh, line 733: This change is confusing to me. Could you walk me through why we expect the status to be paid? From what I understand, any PR that we pass to `ProcessAndLockReviewedPR` should end up with the `PaymentRequestStatusSentToGex` status when it's successfully processed.

gidjin:
> The locked PR test is also failing for me intermittently. We should not knowingly commit a flakey test, so I would vote to remove it if we can't get it to pass reliably. I remain skeptical of the value of this test. It feels like we're testing Postgres functionality.

I agree with not committing knowingly intermittent tests. At this point I'm going to take it out and file a story to review our concurrent handling

pkg/services/payment_request/payment_request_reviewed_processor_test.go
gidjin: This is because one of the payment requests is updated by the go routine `lockPR` which sets the status to paid. So the first one should be paid and the second should be sent. The idea was to test concurrency locking. But I'm not sure we even need it at this stage of the tests.

mboren (approved):

PR #6543, opened 2021-04-29
[MB-3539] Remove references to "IsCanceled" as related to Moves/Move Task Orders

harshamurthy:
## Description

During the `Move` and `MoveTaskOrder` [consolidation](https://github.com/transcom/mymove/pull/4533/), we did not port over the "is_canceled" DB field from the `MoveTaskOrder` table to the `Move` table since the `Move` table already has a "status" DB field with `CANCELED` as a value option.

When that consolidation happened, we did not change the API, so there were still references to `IsCanceled` and the ability to set a `Move`'s `status` to `CANCELED` via the `IsCanceled` payload attribute.

This PR removes references to `IsCanceled` from the prime, ghc, and support APIs.

## Reviewer Notes

Although reference to `IsCancelled` has been removed from the API, there is no way to update a `Move`'s status. Does that need to be added?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3539) for this change


pkg/handlers/ghcapi/move_task_order_test.go
```diff
@@ -74,7 +74,8 @@ func (suite *HandlerSuite) TestGetMoveTaskOrderHandlerIntegration() {
 	suite.Assertions.IsType(&move_task_order.GetMoveTaskOrderOK{}, response)
 	suite.Equal(strfmt.UUID(moveTaskOrder.ID.String()), moveTaskOrderPayload.ID)
 	suite.Nil(moveTaskOrderPayload.AvailableToPrimeAt)
-	suite.False(*moveTaskOrderPayload.IsCanceled)
+	// TODO: Check that the *moveTaskOrderPayload.Status is not "canceled"
```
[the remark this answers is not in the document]
harshamurthy: @pearl-truss, I thought about it, but I think it would require us to add Status to the API as it's not currently there. When @gidjin and I spoke briefly about it, it sounded like that would be a follow on story.

cmd/prime-api-dev/scripts/payment_requests.go
```diff
@@ -221,10 +221,6 @@ func (pr *paymentRequestsData) displaySelectedMTO() {
 		fmt.Printf("AvailableToPrime: %s\n", mto.AvailableToPrimeAt.String())
 	}
 
-	if mto.IsCanceled != nil {
-		fmt.Printf("Is Canceled: %s\n", strconv.FormatBool(*mto.IsCanceled))
-	}
```
monfresh, line 226: Have you checked with Roci if they need this? If they do need it, did you document somewhere that we'll need to add this back in the follow-up story?

harshamurthy: @monfresh, this is something Jacquie was experimenting with that John gave me the green light to remove.

PR #6567, opened 2021-05-03
MB-8003-migration-cancel-and-diverted-shipment

NamibiaTorres:
## Description

Create a migration for the new added statuses "Cancelation" and "Diversion Requested" and the new column "Diversion" in the mto_shipments table.

Update the model with the new statuses and the new column name.

Update the model test to reflect the new statuses
## Setup

Make sure that the migration was added to the list of migrations: `20210503211133_cancelled_and_diverted_statuses.up.sql`

Run the model tests to make sure tests pass: 
`go test ./pkg/models/   -v`
## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8003) for this change
* [this article](https://github.com/transcom/mymove/wiki/Database-Migrations#Creating-Migrations) explains more about the approach used.

### DIFF
--- a/migrations/app/migrations_manifest.txt
+++ b/migrations/app/migrations_manifest.txt
@@ -633,3 +633,4 @@
 20210421175513_exp_truss_office_gbloc_change.up.sql
 20210423140812_update-gbloc-truss-offices-experimental.up.sql
 20210428225942_marty_cac.up.sql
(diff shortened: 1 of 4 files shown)


LeDeep (approved):
looks good. 🚀

PR #6570, opened 2021-05-04
[MB-7549] Add a way for the database to store information about which duty stations require services counseling

harshamurthy:
## Description
The MilMove system needs a way to check if an origin `duty_station` provides `services_counseling`. Service members moving from duty stations in the ["Government Counseling Locations" tab](https://docs.google.com/spreadsheets/d/1HQuQCeWQ99YmKwTjSVHMwnMQCqhQ9uYt/edit#gid=1151834353) will receive services counseling.

This PR adds the `provides_services_counseling` column to the `duty_stations` table. It also adds validations and updates tests.

## Reviewer Notes

1. Given `provides_services_counseling` is a boolean and there's a `NOT NULL` requirement on the column, does there also need to be a model validation?
2. Is it okay to default the value of `provides_services_counseling` to `false`?

## Setup

```sh
rake db_dev_migrate
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7549) for this change

### DIFF
--- a/migrations/app/schema/20210504162959_add_requires_services_counseling.up.sql
+++ b/migrations/app/schema/20210504162959_add_requires_services_counseling.up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE duty_stations ADD COLUMN requires_services_counseling boolean DEFAULT false NOT NULL;
--- a/migrations/app/schema/20210504162959_add_requires_services_counseling.up.sql
+++ b/migrations/app/schema/20210504162959_add_requires_services_counseling.up.sql
@@ -1,1 +1,0 @@
-ALTER TABLE duty_stations ADD COLUMN requires_services_counseling boolean DEFAULT false NOT NULL;
--- a/pkg/models/duty_station.go
+++ b/pkg/models/duty_station.go
@@ -23,7 +23,7 @@ type DutyStation struct {
 	Address                    Address                      `belongs_to:"address"`
 	TransportationOfficeID     *uuid.UUID                   `json:"transportation_office_id" db:"transportation_office_id"`
 	TransportationOffice       TransportationOffice         `belongs_to:"transportation_offices"`
-	RequiresServicesCounseling bool                         `json:"requires_services_counseling" db:"requires_services_counseling"`
+	ProvidesServicesCounseling bool                         `json:"provides_services_counseling" db:"provides_services_counseling"`
 }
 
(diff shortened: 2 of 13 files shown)


migrations/app/schema/20210504162959_add_requires_services_counseling.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE duty_stations ADD COLUMN requires_services_counseling boolean DEFAULT false NOT NULL;
```
pearl-truss, line 1: I think we're trying to add comments to columns for database documentation. Like what you see here https://github.com/transcom/mymove/blob/master/migrations/app/schema/20210419195440_add_editype_to_paymentRequestToInterchange_table.up.sql#L8

pkg/models/duty_station.go
```diff
@@ -14,15 +14,16 @@ import (
...
+	ID                         uuid.UUID                    `json:"id" db:"id"`
+	CreatedAt                  time.Time                    `json:"created_at" db:"created_at"`
+	UpdatedAt                  time.Time                    `json:"updated_at" db:"updated_at"`
+	Name                       string                       `json:"name" db:"name"`
+	Affiliation                internalmessages.Affiliation `json:"affiliation" db:"affiliation"`
+	AddressID                  uuid.UUID                    `json:"address_id" db:"address_id"`
+	Address                    Address                      `belongs_to:"address"`
+	TransportationOfficeID     *uuid.UUID                   `json:"transportation_office_id" db:"transportation_office_id"`
+	TransportationOffice       TransportationOffice         `belongs_to:"transportation_offices"`
+	RequiresServicesCounseling bool                         `json:"requires_services_counseling" db:"requires_services_counseling"`
```
pearl-truss, line 26: Nit: I'm not sure about the word 'required' here because duty stations don't require services counseling (a move with a certain duty station might require services counseling but the duty station itself wouldn't, if that makes sense). I was thinking this field would be called 'ProvidesServicesCounseling'.

harshamurthy: That makes sense to me @pearl-truss, thanks for the additional context. Will make the change.

pearl-truss (approved):
Thanks for the changes, they look good!

reggieriser (approved):
The migration LGTM.  Ran it fine -- I see the new boolean column and all existing duty stations default to false as expected.

PR #6576, opened 2021-05-05
MB-7684-diverted-shipments

NamibiaTorres:
## Description

The purpose of this PR is to:
1. Provide user (Prime) with an action that notifies the government that the shipment has been diverted successfully
2. Provide user (Prime) with an action to create a new shipment that is part of a diversion.

To do this, several updates and checks needed to be made:
1. The prime and support yaml were updated with the new DIVERSION_REQUESTED status, and the new Diversions column (bool)
2. The model to payload was updated.
3.  The handler was updated with a check on the diversions column.

## Setup
* NOTE: we really only need to check that diversion gets updated, so the you can just pass in diversion: `bool value` into the body. 

Sample body for updateMTOShipment:
```
{
   "diversion": true,
   "scheduledPickupDate": "2021-05-05"
}
```

Sample body for createMTOShipment:

```
{
    "moveTaskOrderID": "99783f4d-ee83-4fc9-8e0c-ffffffffffff",
    "pickupAddress": {
        "streetAddress1": "123 Main Ave",
        "city": "Anytown",
        "state": "MN",
        "postalCode": "90210",
        "streetAddress2": "Apartment 9000",
        "streetAddress3": "Montmârtre",
        "country": "USA",
        "eTag": "dolor mollit ad"
    },
    "destinationAddress": {
        "streetAddress1": "123 Main Ave",
        "city": "Anytown",
        "state": "NH",
        "postalCode": "90210",
        "streetAddress2": "Apartment 9000",
        "streetAddress3": "Montmârtre",
        "country": "USA",
        "eTag": "ut veniam dolor"
    },
    "shipmentType": "HHG",
    "requestedPickupDate": "1982-04-23",
    "primeEstimatedWeight": 4500,
    "customerRemarks": "handle with care",
    "agents": [],
    "pointOfContact": "magna sed",
    "mtoServiceItems": [],
    "diversion": true
}
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7684) for this change


pkg/handlers/primeapi/mto_shipment.go
```diff
@@ -140,6 +140,9 @@ func UpdateMTOShipmentModel(mtoShipmentID strfmt.UUID, payload *primemessages.MT
 	if payload.Status != "" {
 		fieldsInError.Add("status", "cannot be updated")
 	}
+	if payload.Diversion {
+		fieldsInError.Add("diversion", "cannot be updated")
+	}
```
sandy-wright, line 145: I thought we were supposed to be able to update this field?

pkg/handlers/primeapi/payloads/model_to_payload.go
```diff
@@ -341,6 +341,7 @@ func MTOShipment(mtoShipment *models.MTOShipment) *primemessages.MTOShipment {
 		CustomerRemarks:          mtoShipment.CustomerRemarks,
 		PickupAddress:            Address(mtoShipment.PickupAddress),
 		Status:                   string(mtoShipment.Status),
+		Diversion:                bool(mtoShipment.Diversion),
```
sandy-wright, line 344: I think you'll need to add this in the `payload_to_model` function too so that way we can actually pass a new value to the updater.

pkg/handlers/primeapi/mto_shipment.go
NamibiaTorres: This is what I had a question on, if someone selects True, can they change it back to false later on?

sandy-wright: Oh, that's a very good question and probably something you should ask design/product about. If you wanted to do that check though, you would need to do it in the service object. Right here, you're checking if the _input_ was True, and that should be allowed.

crwallace: I would say that it is fine that someone can change it to false later on. I think we don't want to put too many restrictions in place before the Prime gets here and we know their user behavior.

NamibiaTorres: Thanks everyone!

eamahanna (commented):
When the prime creates a shipment, it is marked as not available to prime. Does that make sense in this situation, when the prime creates the shipment? @sandy-wright @abbyoung thoughts?

eamahanna (approved):
Everything works. I just had the one question. If that's not an issue, feel free to merge :)

NamibiaTorres:
Merging PR, there is a general question from @eamahanna about creating shipments  that needs an answer but is not blocking this PR.

PR #6577, opened 2021-05-05
MB-7358 Add orders details for Services Counselor

monfresh:
## Description

This adds the Orders details definition list to the Services Counseling
view. A Services Counselor needs less info than a TOO, so I duplicated
the `OrdersList` that we use on the TOO's view, and removed the info
that a Services Counselor does not need to see.

I also added the `View and edit orders` button, but it doesn't do
anything currently, as that will be implemented in a separate story.

## Reviewer Notes

1. Using Local Sign In, sign in with a user that has the services counselor role
2. You should only see one move that needs service counseling in your queue
3. Click on it
4. You should see the Orders table as shown in the screenshot below

## Setup

```sh
make server_run
make db_dev_e2e_populate (or `make db_dev_fresh` if that fails)
make office_client_run
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7358) for this change

## Screenshots

Screen Shot 2021-05-05 at 12 43 52 PM

### DIFF
--- a/src/components/Office/DefinitionLists/ServicesCounselingOrdersList.test.jsx
+++ b/src/components/Office/DefinitionLists/ServicesCounselingOrdersList.test.jsx
@@ -1,0 +1,47 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import ServicesCounselingOrdersList from './ServicesCounselingOrdersList';
+
+const ordersInfo = {
+  currentDutyStation: { name: 'JBSA Lackland' },
+  newDutyStation: { name: 'JB Lewis-McChord' },
+  issuedDate: '2020-03-08',
+  reportByDate: '2020-04-01',
+  ordersType: 'PERMANENT_CHANGE_OF_STATION',
+};
+
+// what ordersInfo from above should be rendered as
+const expectedRenderedOrdersInfo = {
+  currentDutyStation: 'JBSA Lackland',
+  newDutyStation: 'JB Lewis-McChord',
+  issuedDate: '08 Mar 2020',
+  reportByDate: '01 Apr 2020',
+  ordersType: 'Permanent Change Of Station (PCS)',
+};
+
+const ordersInfoOnlyForTOO = {
+  departmentIndicator: '17 Navy and Marine Corps',
+  ordersNumber: '999999999',
+  ordersTypeDetail: 'Shipment of HHG Permitted',
+  tacMDC: '9999',
+  sacSDN: '999 999999 999',
+};
+
+describe('ServicesCounselingOrdersList', () => {
+  it('renders formatted orders info', () => {
+    render(<ServicesCounselingOrdersList ordersInfo={ordersInfo} />);
+
+    Object.keys(expectedRenderedOrdersInfo).forEach((key) => {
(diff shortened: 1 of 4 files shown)


src/pages/Office/ServicesCounselingMoveDetails/ServicesCounselingMoveDetails.test.jsx
```diff
@@ -166,6 +166,10 @@ describe('MoveDetails page', () => {
       expect(wrapper.find('button[data-testid="submitMoveDetailsBtn"]').length).toBe(1);
     });
 
+    it('renders the Orders Definition List', () => {
+      expect(wrapper.find('#orders h2').text()).toEqual('Orders');
```
jenniferlynparsons, line 170: Not a blocker, but on both this and the Allowances test below, we're not really checking to see if the list is rendering, just the header. Another `expect` or two would be helpful.

jenniferlynparsons (approved):

src/components/Office/DefinitionLists/ServicesCounselingOrdersList.test.jsx
```diff
@@ -0,0 +1,47 @@
...
+describe('ServicesCounselingOrdersList', () => {
+  it('renders formatted orders info', () => {
+    render(<ServicesCounselingOrdersList ordersInfo={ordersInfo} />);
+
+    Object.keys(expectedRenderedOrdersInfo).forEach((key) => {
+      expect(screen.getByText(expectedRenderedOrdersInfo[key])).toBeInTheDocument();
+    });
+  });
+
+  it('does not render unexpected info', () => {
```
duncan-truss, line 40: I'm not sure what this test is accomplishing exactly, I could understand if we updated information maybe then we might do something like this.

Is there a way with the testing library to maybe check the ordering and number of children expected in the component as a better way to know only what we wanted rendered and nothing else?

duncan-truss: Ahh I see now you're talking about the excluded other orders fields that are shown in the TOO version of this component.

monfresh: Yeah, but I think you make a good point. I was thinking about that when I wrote this test. If there's a way to check that only those specific fields are present and nothing else, that would be best.

amanda5iu:
Looking good! Styling fix if it's easy: bottom padding should be 4 units/32px on all the containers.

Screen Shot 2021-05-06 at 4 58 44 PM

src/pages/Office/ServicesCounselingMoveDetails/ServicesCounselingMoveDetails.test.jsx
monfresh: It looks like we're doing the same thing in the MoveDetails test. Is there a way to verify that a component is present? That should be enough since the component is tested separately, right?

monfresh:
If the bottom padding is wrong here, I believe it's also wrong in all the other info boxes. I'm not very familiar with how we manage our CSS, so I'll open a new ticket to fix this everywhere.

PR #6586, opened 2021-05-06
[MB-7551] Populate the database with the duty stations that require services counseling

mboren:
## Description

When the `ProvidesServicesCounseling` flag was added to `DutyStation`, it was defaulted to false. This PR sets the flag to true for duty stations that provide services counseling.
I also changed the moves created in devseed to make all the moves that are used to test services counseling have an origin duty station that provides services counseling.

The `ProvidesServicesCounseling` field is not hooked up to any logic yet, that will be done in a future PR.

## Reviewer Notes
The list of duty stations that provide services counseling is not finalized yet, so we may have to modify it later. 80% of the counseling offices mapped easily to duty stations, but there are a few left that were not obvious, so we're waiting to hear back from USTC before adding those.

The current state of the mapping is in [this spreadsheet at](https://docs.google.com/spreadsheets/d/1YqtkwuuclDyQO0QrwahdNozSmO8uLo7Q/edit#gid=1445947263) in column `I`.

Duty stations that provide services counseling and are in the default GBLOC, LKNQ, are:
- Fort Irwin
- Fort Huachuca
- Camp Pendleton
- Presidio of Monterey

`primemessages.DutyStation` does not include `ProvidesServicesCounseling`, so you can't check these with `fetch-mto-updates`. The field is also not shown in the UI yet. So I think we're stuck with DB queries for checking this PR.

### Check migration
Run migrations and make sure we've updated the right number of records:
```sh
make db_dev_reset
make db_dev_migrate
```
```
dev_db=# select count(*) from duty_stations where provides_services_counseling = TRUE;
 count
-------
    82
(1 row)
```
You should see 82 duty stations that provide services counseling.

You could also spot check the duty stations that provide counseling against the spreadsheet.
### Check devseed data
```
make db_dev_e2e_populate
```

Devseed creates 3 moves that need services counseling and 1 with services counseling completed.
The move locators for these moves are:
- SRVCSL
- USMCSS
- USMCSC
- CSLCMP

We can check that all of those are associated with duty stations that provide services counseling with the following query:
```sql
select moves.locator, ds.name, ds.provides_services_counseling from moves
    join orders o on moves.orders_id = o.id
    join duty_stations ds on o.origin_duty_station_id = ds.id
    where ds.provides_services_counseling = TRUE;
```
expected output (name column is randomly generated so it will vary):
```
 locator |    name    | provides_services_counseling
---------+------------+------------------------------
 SRVCSL  | O3Yp35MRb6 | t
 USMCSS  | nt37ShVBn8 | t
 USMCSC  | cRNpINkpLN | t
 CSLCMP  | EAip3TqKrh | t
(4 rows)
```

## Acceptance
This shouldn't result in any changes that are visible with the UI or API so I don't think there is anything we can do for acceptance.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7551) for this change


pearl-truss (approved):
Changes look good! I confirmed I see 82 duty stations with provides_services_counseling that are true and the 4 populated from devseed

reggieriser (approved):
Migration looks good to me.  I ran it fine locally and got 82 duty stations with counseling as expected.  I also populated the E2E data and saw the four moves with service counseling too.  I spot-checked a few duty stations with the spreadsheet and things matched. 🚀

PR #6593, opened 2021-05-07
Fixed lint errors and enabled ato-linter in cmd directory 

Ronolibert:
## Description

Fixed and enabled ato-linter in cmd directory

## Setup
`go run ./cmd/ato-linter/main.go -- ./cmd/...`
You should see no errors


.github/workflows/be-linter.yml
```diff
@@ -21,3 +21,6 @@ jobs:
 
       - name: Run custom go linter on pkg
         run: go run ./cmd/ato-linter/main.go -- ./pkg/...
+
+      - name: Run custom go linter on cmd
+        run: go run ./cmd/ato-linter/main.go -- ./cmd/...
```
felipe-lee, line 26: Should we just run it on everything? Like instead of `./pkg/...` do `./...`. I guess idk how efficiently it skips all non-go files.

jacquelineIO (approved):
I ran this and get the same error that is flagged inline in the PR. And I see that Ron has already pinged Leo to clear this up.

.github/workflows/be-linter.yml
Ronolibert: Good call, it does increase the runtime noticeably but at least we've covered all our bases

felipe-lee (approved):

PR #6594, opened 2021-05-07
[MB-8026] Optimizing FetchMany in query builder

reggieriser:
## Description

This PR begins to address two performance problems with our current implementation of `FetchMany` in the query builder:
- `FetchMany` uses Pop's `Eager` not `EagerPreload` so it is making way more requests than it should.
- `FetchMany` does not have a way to tell it you want to load zero associations. So it loads ALL associations. 

## Reviewer Notes

- There are a lot of files changed in this PR, but the changes break down as follows:
  - `NewQueryAssociation` was remapped to the new `NewQueryAssociationPreload` in some places where preloading made sense and seemed to work fine.
  - Changed some (accidental) default `Eager` calls (which loaded all associations) to `nil` to load no associations.  This required changes to the query builder to support the `nil`.
  - A good bit of the file changes are due to adding an explicit `fk_id` to `has_many` and `belongs_to` relationships so we don't have to rely on Pop's guesses for the field names (it was sometimes guessing wrong when preloading).
  - Added ability to specify preloading as part of setting up query associations in the query build (see pkg/services/query/)
- I tried to change`FetchOne` as well to preload associations, but I got a lot of failures when doing so; we'll have to work through those if we want to make that change.  `FetchOne` probably needs a refactoring anyway as it is loading all associations every time and there's no way to change that behavior on a case-by-case basis.  The good news is that since `FetchOne` is fetching a single base record, default eager loading of associations doesn't cause an explosion of additional queries like it can be with a big list of base records (like `FetchMany` might have).
- The most obvious performance improvements are found in the admin interface list views.  In particular, the TSPP listing should be a lot faster.  You can see the difference in SQL queries generated on dev by comparing the admin interface on master with this branch.
- I ran into a bug in Pop when trying to preload a `has_many` association with a pointer-based foreign key (like we have when preloading `MTOServiceItems.ReService` for a shipment).  I made a [PR for Pop](https://github.com/gobuffalo/pop/pull/647) and it's been merged and is in Pop 5.3.4, but [we have other issues](https://github.com/transcom/mymove/pull/6589#issuecomment-834774756) it looks like we have to solve first before moving to that version.

## Setup

- The tests should hopefully catch any major issue here; I had to avoid a preload or deleting the default eager in a few places when tests started failing.
- Most of the changes were around the admin interface, so try it and make sure you don't see anything new that's broken or data that's missing.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8026) for this change
* [Relevant Slack thread](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1619811517304100)

### DIFF
--- a/pkg/handlers/internalapi/mto_shipment.go
+++ b/pkg/handlers/internalapi/mto_shipment.go
@@ -183,11 +183,11 @@ func (h ListMTOShipmentsHandler) Handle(params mtoshipmentops.ListMTOShipmentsPa
 		query.NewQueryFilter("move_id", "=", moveTaskOrderID.String()),
 	}
 
-	// In some places, we used this unbound eager call accidentally and loaded all associations when the
-	// intention was to load no associations. In this instance, we get E2E failures if we change this to load
-	// no associations, so we'll keep it as is and can revisit later if we want to optimize further.  This is
-	// just loading shipments for a specific move (likely only 1 or 2 in most cases), so the impact of the
-	// additional loading shouldn't be too dramatic.
+	// TODO: In some places, we used this unbound eager call accidentally and loaded all associations when the
+	//   intention was to load no associations. In this instance, we get E2E failures if we change this to load
+	//   no associations, so we'll keep it as is and can revisit later if we want to optimize further.  This is
+	//   just loading shipments for a specific move (likely only 1 or 2 in most cases), so the impact of the
+	//   additional loading shouldn't be too dramatic.
 	queryAssociations := query.NewQueryAssociations([]services.QueryAssociation{})
 
 	queryOrder := query.NewQueryOrder(swag.String("created_at"), swag.Bool(true))
--- a/pkg/handlers/internalapi/mto_shipment.go
+++ b/pkg/handlers/internalapi/mto_shipment.go
@@ -183,10 +183,17 @@ func (h ListMTOShipmentsHandler) Handle(params mtoshipmentops.ListMTOShipmentsPa
 		query.NewQueryFilter("move_id", "=", moveTaskOrderID.String()),
 	}
 
+	// In some places, we used this unbound eager call accidentally and loaded all associations when the
+	// intention was to load no associations. In this instance, we get E2E failures if we change this to load
+	// no associations, so we'll keep it as is and can revisit later if we want to optimize further.  This is
+	// just loading shipments for a specific move (likely only 1 or 2 in most cases), so the impact of the
+	// additional loading shouldn't be too dramatic.
+	queryAssociations := query.NewQueryAssociations([]services.QueryAssociation{})
+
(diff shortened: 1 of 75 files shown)


pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -182,6 +182,12 @@ func (h ListMTOShipmentsHandler) Handle(params mtoshipmentops.ListMTOShipmentsPa
 	queryFilters = []services.QueryFilter{
 		query.NewQueryFilter("move_id", "=", moveTaskOrderID.String()),
 	}
+
+	// In some places, we used this unbound eager call accidentally and loaded all associations when the
```
gidjin, line 186: Should we prefix this with TODO? Just wondering if that would help us find it later, which may not be true.

pkg/models/re_domestic_linehaul_price.go
```diff
@@ -26,8 +26,8 @@ type ReDomesticLinehaulPrice struct {
 	UpdatedAt             time.Time       `json:"updated_at" db:"updated_at"`
 
 	// Associations
-	Contract            ReContract            `belongs_to:"re_contract"`
-	DomesticServiceArea ReDomesticServiceArea `belongs_to:"re_domestic_service_area"`
+	Contract            ReContract            `belongs_to:"re_contract" fk_id:"contract_id"`
```
gidjin, line 29: Looking at all these `fk_id` additions and I can't help but wonder if there are guidelines on when to add `fk_id` to a model column. Or maybe we should always do it?

pkg/services/query/query_builder.go
```diff
@@ -459,6 +459,13 @@ func (p *Builder) QueryForAssociations(model interface{}, associations services.
 }
 
 func associatedQuery(query *pop.Query, associations services.QueryAssociations, model interface{}) *pop.Query {
+	if associations == nil {
+		return query
+	}
+
+	if associations.Preload() {
+		return query.EagerPreload(associations.StringGetAssociations()...)
+	}
```
gidjin, line 468: Is there a test suite for the query builder? If so is it feasible to add a test or is this covered enough by other testing?

gidjin (approved):
Changes look good to me! Admin interface seemed to be working as expected. TSPP data page is much faster and a lot less queries. Thanks for the clear Reviewer notes that really helped. I left a couple questions but I don't think it needs to hold up approval form my end. Might be good to get some other eyes on this if possible.

pkg/handlers/internalapi/mto_shipment.go
reggieriser: Sure, I just added it.

pkg/services/query/query_builder.go
reggieriser: I added some tests that get to this via `FetchMany`.

PR #6595, opened 2021-05-07
MB-3538 Switch from `Count` to `Exists` for ref ID uniqueness check

felipe-lee:
## Description

We need to check if we have generated a unique reference ID for a move. Our current implementation uses `Count` to see if any exist with that ID. This changes to using `Exists` since we don't care about the actual count beyond whether it's >0, which `Exists` handles for us.

## Reviewer Notes

Not sure if there's a way to test for efficiency. I've used things before in django to kind of keep an eye on queries and timing, but not quite at this level, usually at an integration test level (like making sure a single handler doesn't perform more than x queries).

## Setup

```sh
make server_test
```

or

```sh
go test ./pkg/models -v -testify.m TestGenerateReferenceID
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3538) for this change

### DIFF
--- a/pkg/models/move.go
+++ b/pkg/models/move.go
@@ -654,10 +654,11 @@ func generateReferenceIDHelper(db *pop.Connection) (string, error) {
 
 	newReferenceID := fmt.Sprintf("%04d-%04d", firstNum, secondNum)
 
-	count, err := db.Where(`reference_id= $1`, newReferenceID).Count(&Move{})
+	exists, err := db.Where(`reference_id= $1`, newReferenceID).Exists(&Move{})
+
 	if err != nil {
 		return "", err
-	} else if count > 0 {
+	} else if exists {
 		return "", errors.New("move: reference_id already exists")
 	}
 


YanZ777 (approved):

monfresh:
Could you please set `export DB_DEBUG=1` in your `.envrc.local` if it's not already, and then check the logs when you create a move to see which query gets generated?

felipe-lee:
> Could you please set `export DB_DEBUG=1` in your `.envrc.local` if it's not already, and then check the logs when you create a move to see which query gets generated?

Yeah, here's the query that is generated for checking if the ID is already being used:
```sql
SELECT EXISTS (SELECT moves.available_to_prime_at, moves.cancel_reason, moves.contractor_id, moves.created_at, moves.id, moves.locator, moves.orders_id, moves.ppm_estimated_weight, moves.ppm_type, moves.reference_id, moves.selected_move_type, moves.service_counseling_completed_at, moves.show, moves.status, moves.submitted_at, moves.updated_at FROM moves AS moves WHERE reference_id= $1) | ["7689-5905"]
```

lynzt (approved):
this looks good to me 🚢 

here's the `explain analyze` from the orig query to the new one:

```**exists (new)**
Result  (cost=8.15..8.16 rows=1 width=1) (actual time=0.033..0.067 rows=1 loops=1)
  InitPlan 1 (returns $0)
    ->  Index Only Scan using moves_reference_id_key on moves  (cost=0.14..8.15 rows=1 width=0) (actual time=0.014..0.022 rows=0 loops=1)
"          Index Cond: (reference_id = '0050-2655'::text)"
          Heap Fetches: 0
Planning Time: 0.235 ms
Execution Time: 0.128 ms

**count (orig)**
Aggregate  (cost=8.16..8.17 rows=1 width=8) (actual time=0.104..0.137 rows=1 loops=1)
  ->  Index Only Scan using moves_reference_id_key on moves  (cost=0.14..8.15 rows=1 width=0) (actual time=0.084..0.093 rows=0 loops=1)
"        Index Cond: (reference_id = '3345-7486'::text)"
        Heap Fetches: 0
Planning Time: 0.189 ms
Execution Time: 0.239 ms```

PR #6609, opened 2021-05-11
[MB-7926] Services Counselor can't update Authorized weight (backend)

jenniferlynparsons:
## Description

PATCH requests sent from the services counselor cannot change the authorized weight

## Reviewer Notes
I'm unsure of the solution here and the test is definitely not working. Before I work to figure that out, I wanted to get feedback on the solution and ensure the test is going in the right direction.

- Is the code solution acceptable?
- Is the test for this solution set up similarly to what would be expected?

## Setup

```sh
make server_run
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7926) for this change

### DIFF
--- a/pkg/handlers/ghcapi/orders.go
+++ b/pkg/handlers/ghcapi/orders.go
@@ -85,8 +85,8 @@ type UpdateOrderHandler struct {
 func (h UpdateOrderHandler) Handle(params orderop.UpdateOrderParams) middleware.Responder {
 	session, logger := h.SessionAndLoggerFromRequest(params.HTTPRequest)
 
-	if session.Roles.HasRole(roles.RoleTypeServicesCounselor) && params.Body.AuthorizedWeight != nil {
-		return orderop.NewUpdateOrderBadRequest()
+	if session.IsOfficeUser() || session.Roles.HasRole(roles.RoleTypeServicesCounselor) && params.Body.AuthorizedWeight != nil {
+		return orderop.NewUpdateOrderUnauthorized()
 	}
 
 	orderID, err := uuid.FromString(params.OrderID.String())
--- a/pkg/handlers/ghcapi/orders.go
+++ b/pkg/handlers/ghcapi/orders.go
@@ -10,6 +10,7 @@ import (
 	"github.com/transcom/mymove/pkg/gen/ghcmessages"
 	"github.com/transcom/mymove/pkg/gen/internalmessages"
 	"github.com/transcom/mymove/pkg/models"
+	"github.com/transcom/mymove/pkg/models/roles"
 	"github.com/transcom/mymove/pkg/services"
 	"github.com/transcom/mymove/pkg/services/event"
 
@@ -82,7 +83,11 @@ type UpdateOrderHandler struct {
 
 // Handle ... updates an order from a request payload
 func (h UpdateOrderHandler) Handle(params orderop.UpdateOrderParams) middleware.Responder {
-	_, logger := h.SessionAndLoggerFromRequest(params.HTTPRequest)
+	session, logger := h.SessionAndLoggerFromRequest(params.HTTPRequest)
+
+	if session.Roles.HasRole(roles.RoleTypeServicesCounselor) && params.Body.AuthorizedWeight != nil {
+		return orderop.NewUpdateOrderBadRequest()
+	}
 
 	orderID, err := uuid.FromString(params.OrderID.String())
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/handlers/ghcapi/orders_test.go
```diff
@@ -209,6 +210,63 @@ func (suite *HandlerSuite) TestUpdateOrderHandlerIntegration() {
...
+
+	issueDate, _ := time.Parse("2006-01-02", "2020-08-01")
+	reportByDate, _ := time.Parse("2006-01-02", "2020-10-31")
+
+	newAuthorizedWeight := int64(10000)
+	deptIndicator := ghcmessages.DeptIndicator("COAST_GUARD")
+	affiliation := ghcmessages.BranchAIRFORCE
+	grade := ghcmessages.GradeO5
+	ordersTypeDetail := ghcmessages.OrdersTypeDetail("INSTRUCTION_20_WEEKS")
+	body := &ghcmessages.UpdateOrderPayload{
```
duncan-truss, line 233: You may be able to consolidate this into one testdatagen.MakeOrders() call and then call the model to payload function if it sets all the fields you need.

pkg/handlers/ghcapi/orders.go
```diff
@@ -82,7 +83,11 @@ type UpdateOrderHandler struct {
 
 // Handle ... updates an order from a request payload
 func (h UpdateOrderHandler) Handle(params orderop.UpdateOrderParams) middleware.Responder {
-	_, logger := h.SessionAndLoggerFromRequest(params.HTTPRequest)
+	session, logger := h.SessionAndLoggerFromRequest(params.HTTPRequest)
+
+	if session.IsOfficeUser() || session.Roles.HasRole(roles.RoleTypeServicesCounselor) && params.Body.AuthorizedWeight != nil {
```
hannaliebl, line 88: Is the `||` correct here -- is this for all office users, or just for service counselors?

monfresh: Good catch! Yeah, that should be `&&`.

jenniferlynparsons: Okay! I found this as well but it revealed some weirdness.
- With `||` as you might expect all office users could not update the weight allowance
- With `&&` the test is failing. The appearance is that the user on that test is only a SC and not an office user
- When I remove `session.IsOfficeUser()` all the tests pass locally.

I need to do more digging to see why the test data doesn't know the user is an office user as well as having a role of SC.

duncan-truss: We would still want to return unauthorized if a customer or admin made a request to this endpoint though.

jenniferlynparsons: I think the `&&` logic is correct at the end of the day. I think it's the test that has an issue.

pkg/handlers/ghcapi/orders_test.go
```diff
@@ -209,14 +210,54 @@ func (suite *HandlerSuite) TestUpdateOrderHandlerIntegration() {
 	suite.Equal(body.DependentsAuthorized, ordersPayload.Entitlement.DependentsAuthorized)
 }
 
+func (suite *HandlerSuite) TestUpdateServicesCounselorOrderHandlerIntegration() {
+	officeUser := testdatagen.MakeOfficeUser(suite.DB(), testdatagen.Assertions{Stub: true})
```
donaldthai, line 214: There's a `MakeServicesCounselorOfficeUser` function in `testdatagen` that you could use to create. Make sure that it isn't stubbed since the user needs to be in the DB.

jenniferlynparsons: YES! That worked! I also tried it with `testdatagen.Assertions{Stub: true}` and that works as well!

pkg/handlers/ghcapi/orders_test.go
```diff
@@ -400,6 +470,11 @@ func (suite *HandlerSuite) TestUpdateOrderHandlerValidationError() {
 }
 
 func (suite *HandlerSuite) TestUpdateOrderHandlerWithoutTac() {
+	officeUser := testdatagen.MakeOfficeUser(suite.DB(), testdatagen.Assertions{Stub: true})
```
monfresh, line 473: We also have `MakeTOOOfficeUser`, so if you want, you could save a few lines in each of these tests. Not a blocker.

monfresh (approved):
LGTM!

pkg/handlers/ghcapi/orders_test.go
jenniferlynparsons: Awesome. Just pushed that up. Might as well clean it up now. 😄

PR #6611, opened 2021-05-12
MB-7387 services counselor is able to edit shipment details

YanZ777:
## Description

Add a new page for editing shipments for service counselors.

## Reviewer Notes

Would like feedback on or advice:
- Did I stub the backend stuff correctly? 

## Setup

```sh
make office_client_run
make server_run
```
I think you should be able to hit this website and see the form here: [http://officelocal:3000/counseling/moves/SRVCSL/3cc2d807-2905-4ff5-9ce6-147c101652b4/edit](http://officelocal:3000/counseling/moves/SRVCSL/3cc2d807-2905-4ff5-9ce6-147c101652b4/edit)
You'll probably have to logon to the customer side and make a move that the services counselor will be able to see. 
From there, you'll need to grab the move id (I got mine from the service call) and use that to enter in `/moves/:moveId/shipmentId` on the office side. Let me know if you have trouble with this and I can demonstrate.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7387) for this change

## Screenshots

officelocal_3000_counseling_moves_6MRPHM_880d6719-5893-4fb8-b7f2-110cbf89bb29_edit (2)


src/pages/Office/ServicesCounselingEditShipmentDetails/ServicesCounselingEditShipmentDetails.jsx
```diff
@@ -0,0 +1,46 @@
...
+import ServicesCounselingShipmentForm from 'components/Office/ServicesCounselingShipmentForm/ServicesCounselingShipmentForm';
+import { SHIPMENT_OPTIONS } from 'shared/constants';
+import { useMoveDetailsQueries } from 'hooks/queries';
+import LoadingPlaceholder from 'shared/LoadingPlaceholder';
+import SomethingWentWrong from 'shared/SomethingWentWrong';
+
+const ServicesCounselingEditShipmentDetails = () => {
+  const { moveCode, shipmentId } = useParams();
+  const history = useHistory();
+  const { order, mtoShipments, isLoading, isError } = useMoveDetailsQueries(moveCode);
```
duncan-truss, line 14: This probably returns more data than you need honestly, there will likely never be any service items on the move or shipment at this stage.  I think if you wrote a new hook that didn't query for the mto service items that would be a start even if it returned all the shipments instead of only the one you want for the edit form.

duncan-truss:
@YanZ777 To answer your question about the move Id vs. the move code, the office generally uses the move code also known as the 6 alphanumeric character move locator because that's what's in most of the URLs.  The react components only have that param from the URL to start fetching data for so that's why you see move code instead of the uuid move id.  There are other endpoints for shipment and service items that still use the move id (also known as the move task order id).  So they should be distinct things unless we've slipped up in our renaming efforts.

src/pages/Office/ServicesCounselingEditShipmentDetails/ServicesCounselingEditShipmentDetails.jsx
```diff
@@ -0,0 +1,46 @@
...
+          path: '/moves/:moveId/:shipmentId',
+        })}
+        history={history}
+        updateMTOShipment={updateMtoShipmentHandler}
+        isCreatePage={false}
+        currentResidence={customer.current_address}
+        newDutyStationAddress={order.destinationDutyStation}
+        selectedMoveType={SHIPMENT_OPTIONS.HHG}
+        mtoShipment={matchingShipment}
+        serviceMember={{ ...customer, weight_allotment: weightAllotment }}
```
duncan-truss, line 40: It looks like you may not need the spread of the other customer props only the weight allotment from your propType shape or will it be needed in the future?

src/pages/Office/ServicesCounselingEditShipmentDetails/ServicesCounselingEditShipmentDetails.jsx
```diff
@@ -0,0 +1,46 @@
...
+  const { pathname } = useLocation();
+
+  if (isLoading) return <LoadingPlaceholder />;
+  if (isError) return <SomethingWentWrong />;
+
+  const { customer, entitlement: allowances } = order;
+  const matchingShipment = mtoShipments.filter((shipment) => shipment.id === shipmentId)[0];
+  const weightAllotment = { ...allowances, total_weight_self: allowances.authorizedWeight };
+
+  const updateMtoShipmentHandler = () => {};
```
duncan-truss, line 24: This is a new handler that Marjo/Felipe were working on I think yea? We've been using react-query useMutation hooks for api requests like this. You can see an example where we update the mto shipment status to approved from the TOO move details page

https://github.com/transcom/mymove/blob/36e82322ad3e87c2f9d5f32b54f2fe3233d817c9/src/pages/Office/MoveDetails/MoveDetails.jsx#L79

To your question about the key I don't know if we have constants the same way we do for the query hooks that impact caching but you'll need to define the new query here https://github.com/transcom/mymove/blob/36e82322ad3e87c2f9d5f32b54f2fe3233d817c9/src/services/ghcApi.js#L148 .  That's more about matching the swagger tag and operationID though, it looks like your schema key will be `mtoShipment`.

src/pages/Office/ServicesCounselingEditShipmentDetails/ServicesCounselingEditShipmentDetails.jsx
YanZ777: Yea, you're right. I'll take out the spread!

YanZ777:
> @YanZ777 To answer your question about the move Id vs. the move code, the office generally uses the move code also known as the 6 alphanumeric character move locator because that's what's in most of the URLs. The react components only have that param from the URL to start fetching data for so that's why you see move code instead of the uuid move id. There are other endpoints for shipment and service items that still use the move id (also known as the move task order id). So they should be distinct things unless we've slipped up in our renaming efforts.

Thanks for the clarification on this! 
I'll make a new Match Shape for the office side then since the move code and move id are different things.

src/pages/Office/ServicesCounselingEditShipmentDetails/ServicesCounselingEditShipmentDetails.jsx
YanZ777: Yea, this is related to the handler that Marjo / Felipe were working on. 
It sounds like I need to rework `ServicesCounselingShipmentForm` to match how `MoveDetails` is doing service calls, is that correct?
Thanks for pointing out the file where I'll need to add the new query!

src/components/Office/ServicesCounselingShipmentForm/ServicesCounselingShipmentForm.jsx
```diff
@@ -133,7 +139,7 @@ const ServicesCounselingShipmentForm = ({
                 address: currentResidence,
               },
             });
-          } else if (match.params.moveId === mtoShipment?.moveTaskOrderId) {
+          } else if (match.params.moveCode === mtoShipment?.moveTaskOrderId) {
```
duncan-truss, line 142: This will never match comparing the move locator value to the move uuid.  Is this just supposed to check if the mtoShipment has been saved before?

src/components/Office/ServicesCounselingShipmentForm/ServicesCounselingShipmentForm.jsx
```diff
@@ -162,190 +168,186 @@ const ServicesCounselingShipmentForm = ({
...
+
+            <div className={styles.ServicesCounselingShipmentForm}>
+              <ShipmentTag shipmentType={shipmentType} shipmentNumber={shipmentNumber} />
+
+              <h1>Edit shipment details</h1>
+
+              <SectionWrapper className={styles.weightAllowance}>
+                <p>
+                  <strong>Weight Allowance: </strong>
+                  {serviceMember.weight_allotment.total_weight_self} lbs
```
duncan-truss, line 186: I think we have a shared weight formatter if you want to use that.

src/pages/Office/ServicesCounselingEditShipmentDetails/ServicesCounselingEditShipmentDetails.test.jsx
```diff
@@ -0,0 +1,295 @@
...
+    expect(saveButton).not.toBeDisabled();
+
+    userEvent.click(saveButton);
+
+    await waitFor(() => {
+      expect(updateMTOShipment).toHaveBeenCalledWith(expectedPayload);
+    });
+  });
+
+  it('shows an error if the patchServiceMember API returns an error', async () => {
```
duncan-truss, line 252: 👏

src/pages/Office/ServicesCounselingEditShipmentDetails/ServicesCounselingEditShipmentDetails.test.jsx
```diff
@@ -0,0 +1,295 @@
...
+        isSuccess: false,
+      };
+    })
+    .mockImplementationOnce(() => {
+      return {
+        isLoading: false,
+        isError: true,
+        isSuccess: false,
+      };
+    }),
```
duncan-truss, line 156: I haven't seen it done this way exactly, you are mimicking the way react query resolves the statuses?  I have seen this tested in the src/hooks/queries.test.js mainly where waitForNextUpdate() is used.

src/pages/Office/ServicesCounselingEditShipmentDetails/ServicesCounselingEditShipmentDetails.jsx
```diff
@@ -0,0 +1,59 @@
...
+      queryCache.setQueryData([MTO_SHIPMENTS, updatedMTOShipment.moveTaskOrderID, false], mtoShipments);
+      queryCache.invalidateQueries([MTO_SHIPMENTS, updatedMTOShipment.moveTaskOrderID]);
+    },
+  });
+
+  if (isLoading) return <LoadingPlaceholder />;
+  if (isError) return <SomethingWentWrong />;
+
+  const { customer, entitlement: allowances } = order;
+  const matchingShipment = mtoShipments.filter((shipment) => shipment.id === shipmentId)[0];
```
duncan-truss, line 31: I think mtoShipments could be undefined here since I don't see a default value being set in useEditShipmentQueries.  Either set it to default to `[]` by default or use the optional chaining here with mtoShipments?.filter

src/pages/Office/ServicesCounselingEditShipmentDetails/ServicesCounselingEditShipmentDetails.test.jsx
YanZ777: Yea it is kind of similar. I mostly wanted to test the error and loading states in isolation without having to do a wait for next update in each test.

src/components/Office/ServicesCounselingShipmentForm/ServicesCounselingShipmentForm.jsx
YanZ777: Yea, I think so, I'll probably take it out since the TODO isn't sure either.

src/components/Office/ServicesCounselingShipmentForm/ServicesCounselingShipmentForm.jsx
```diff
@@ -162,190 +159,186 @@ const ServicesCounselingShipmentForm = ({
...
+
+            <div className={styles.ServicesCounselingShipmentForm}>
+              <ShipmentTag shipmentType={shipmentType} shipmentNumber={shipmentNumber} />
+
+              <h1>Edit shipment details</h1>
+
+              <SectionWrapper className={styles.weightAllowance}>
+                <p>
+                  <strong>Weight Allowance: </strong>
+                  {formatWeight(serviceMember.weight_allotment.total_weight_self)} lbs
```
duncan-truss, line 177: formatWeight should already be adding the "lbs" suffix

src/pages/Office/ServicesCounselingEditShipmentDetails/ServicesCounselingEditShipmentDetails.jsx
```diff
@@ -0,0 +1,59 @@
...
+          <ServicesCounselingShipmentForm
+            match={match}
+            history={history}
+            updateMTOShipment={mutateMTOShipment}
+            isCreatePage={false}
+            currentResidence={customer.current_address}
+            newDutyStationAddress={order.destinationDutyStation?.address}
+            selectedMoveType={SHIPMENT_OPTIONS.HHG}
+            mtoShipment={matchingShipment}
+            serviceMember={{ weight_allotment: weightAllotment }}
```
duncan-truss, line 47: I know we still have these legacy snake case payloads from the api but you should be able to control the props here to be camel case passing down.

src/constants/routes.js
```diff
@@ -30,5 +30,6 @@ export const customerRoutes = {
 };
 
 export const servicesCounselingRoutes = {
-  EDIT_SHIPMENT_INFO_PATH: '/counseling/moves/:moveCode/:shipmentId',
+  EDIT_SHIPMENT_INFO_PATH: '/counseling/moves/:moveCode/:shipmentId/edit',
```
duncan-truss, line 33: Our URL for editing a shipment feels like it should have a `shipments` resource name in the path to me what do you think? I know that's not what's listed here https://docs.google.com/spreadsheets/d/1VWtagZWTcmgqfWaLdQlUHKBJJVtqBXlEJ_KuMMOSZMs/edit#gid=1627844212&range=8:8

cosburn:
It's looking good! A few notes:

1. The space between the header and the HHG shipment tag looks a bit tight. That should be 6 units / 48px.
2. The HHG shipment tag should have dark text color instead of white.
3. The page background color should be the light gray of all desktop pages (let me know if you need that color).
4. The content on the page looks a bit wide. I can find out the exact column width or I'd suggest just taking it down by two to keep it centered and see how that looks.
5. In the mockup, the `Save` and `Cancel` options are both buttons. Let me know if there's a reason that need to change, otherwise let's make it match the design.
6. According to the mockup, this page should include the same header as the move details view, with the customer's name and move info in it. Not sure if that's intended to be part of this story or not, let me know if it's work captured somewhere else!

Screen Shot 2021-05-19 at 9 49 26 AM
Screen Shot 2021-05-19 at 10 05 36 AM
Screen Shot 2021-05-19 at 10 05 40 AM

YanZ777:
> It's looking good! A few notes:
> 
> 1. The space between the header and the HHG shipment tag looks a bit tight. That should be 6 units / 48px.
> 2. The HHG shipment tag should have dark text color instead of white.
> 3. The page background color should be the light gray of all desktop pages (let me know if you need that color).
> 4. The content on the page looks a bit wide. I can find out the exact column width or I'd suggest just taking it down by two to keep it centered and see how that looks.
> 5. In the mockup, the `Save` and `Cancel` options are both buttons. Let me know if there's a reason that need to change, otherwise let's make it match the design.
> 6. According to the mockup, this page should include the same header as the move details view, with the customer's name and move info in it. Not sure if that's intended to be part of this story or not, let me know if it's work captured somewhere else!

Thanks for the review! 
1. I've added a customer header and a tab container so hopefully that fixes the spacing
2. Should this dark text color be happening for all shipment tags?
3. This should be fixed.
4. Hopefully I got the size right.
5. Wow, I can't believe I missed this. The cancel button is normally unstyled on the customer side and I'm reusing a component from there, should I just put in specific styling for this for this page? 
6. I also can't believe I missed this as well. I've put the header in.

src/components/Office/ServicesCounselingShipmentForm/ServicesCounselingShipmentForm.jsx
```diff
@@ -2,29 +2,21 @@ import React from 'react';
...
-} from '@trussworks/react-uswds';
+import { Alert, Checkbox, Fieldset, FormGroup, Label, Radio, Textarea } from '@trussworks/react-uswds';
 
 import getShipmentOptions from '../../Customer/MtoShipmentForm/getShipmentOptions';
 
 import styles from './ServicesCounselingShipmentForm.module.scss';
 
 import formStyles from 'styles/form.module.scss';
-import { customerRoutes } from 'constants/routes';
+import { servicesCounselingRoutes } from 'constants/routes';
```
felipe-lee, line 12: Can you sort these imports by the import path? I just had a pr where I had to sort a bunch of imports to make sure components/styles are imported in a consistent order. You don't have that problem rn, but I'd like to avoid it if possible.

src/pages/Office/ServicesCounselingEditShipmentDetails/ServicesCounselingEditShipmentDetails.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+import { MTO_SHIPMENTS } from 'constants/queryKeys';
+import { MatchShape } from 'types/officeShapes';
+import ServicesCounselingShipmentForm from 'components/Office/ServicesCounselingShipmentForm/ServicesCounselingShipmentForm';
+import { SHIPMENT_OPTIONS } from 'shared/constants';
+import { useEditShipmentQueries } from 'hooks/queries';
+import LoadingPlaceholder from 'shared/LoadingPlaceholder';
+import SomethingWentWrong from 'shared/SomethingWentWrong';
+import { updateMTOShipment } from 'services/ghcApi';
+import CustomerHeader from 'components/CustomerHeader';
```
felipe-lee, line 16: Can you sort these too?

felipe-lee (commented):
I haven't had a chance to review it all, but just I have a little feedback from pulling it up:
1. It seems that the edit button doesn't go to the page for some reason. Maybe I set it up incorrectly, but this story seems the time to get it linked up properly.
2. For some reason, when I try to load the edit page for a shipment, the expected delivery date isn't being filled with the current data.

src/components/Office/ServicesCounselingShipmentForm/ServicesCounselingShipmentForm.jsx
YanZ777: I did a type of sorting. Could you be more specific about sorting by the import path?

YanZ777:
> I haven't had a chance to review it all, but just I have a little feedback from pulling it up:
> 
> 1. It seems that the edit button doesn't go to the page for some reason. Maybe I set it up incorrectly, but this story seems the time to get it linked up properly.
> 2. For some reason, when I try to load the edit page for a shipment, the expected delivery date isn't being filled with the current data.

1. Wrong props was being used, should be fixed now.
2. Looks like the expected delivery date is missing from the mto_shipment call. I haven't dug into this get, but did we want to address  this in this story? (I am worried about scope.)

src/components/Office/ShipmentDisplay/ShipmentDisplay.jsx
```diff
@@ -58,9 +58,11 @@ const ShipmentDisplay = ({ shipmentType, displayInfo, onChange, shipmentId, isSu
         </dl>
         {editURL && (
           <EditButton
-            to={editURL}
+            onClick={() => {
+              window.location.href = editURL;
+            }}
             className={styles.editButton}
-            data-testid="editButton"
+            data-testid={editURL}
```
YanZ777, line 65: @felipe-lee I changed the `data-testid` to be the `editURL`, that way in the tests for the `ServicesCounselingMoveDetails` can still check if the URL is being generated correctly. In those tests, I retrieve the `data-testid` attribute rather than the `to` since I took out the `to` attribute here.

I'm not fully comfy with doing things this way, so if you have a better idea, please let me know!

cosburn (approved):
Looks great!

src/constants/routes.js
felipe-lee: @duncan-truss James had a question about examples, but I'm not sure where to point him: https://ustcdp3.slack.com/archives/C01QCQ6R3BQ/p1621546764002200?thread_ts=1621537350.001800&cid=C01QCQ6R3BQ

src/pages/Office/ServicesCounselingEditShipmentDetails/ServicesCounselingEditShipmentDetails.test.jsx
```diff
@@ -0,0 +1,207 @@
...
+    const saveButton = screen.getByRole('button', { name: 'Save' });
+
+    expect(saveButton).not.toBeDisabled();
+
+    userEvent.click(saveButton);
+
+    await waitFor(() => {
+      expect(mockPush).toHaveBeenCalledWith('/counseling/moves/move123/details');
+    });
+  });
```
felipe-lee, line 206: Should we also have one test for going back?

src/services/ghcApi.js
```diff
@@ -172,6 +172,27 @@ export function updateMTOShipmentStatus({
...
+    {
+      moveTaskOrderID,
+      shipmentID,
+      'If-Match': ifMatchETag,
+      body,
+    },
+    { schemaKey, normalize },
+  );
+}
+
```
felipe-lee, line 195: This is more a question of me double checking. So right now we are actually trying to make the request right? So like, I should be able to create a branch that has both yours and Marjo's changes, and it would all work?

felipe-lee (commented):
Overall LGTM! Just a few questions and comments

src/constants/routes.js
YanZ777: I was running into a case where the route was matching with other things first because Switch picks whichever one matches first. I think there was also something else with the tests as well, but I forget.
e.g. if your route is named /counseling/moves/:moveCode/:shipmentId, the Switch inside the of the index.jsx for the office app will first match /counseling/moves/:moveCode if you put your route after that one.
And react router does mention that because of this, more specific routes should be placed at the top.

src/services/ghcApi.js
YanZ777: Yes, it should be trying to make the request. If it's hooked up properly (am I calling the correct things) is a different story.

src/services/ghcApi.js
```diff
@@ -172,6 +172,27 @@ export function updateMTOShipmentStatus({
   );
 }
 
+export function updateMTOShipment({
```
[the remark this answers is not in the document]
YanZ777: I think only the gets need to be async since they are fetching data and we don't want the FE to be blocked until the request returns. Since it's an update, I don't think it needs to be async.

src/components/Office/ShipmentDisplay/ShipmentDisplay.jsx
YanZ777: Can we have this be a separate task? I'd rather not keep adding to this PR since it's getting kind of giant.

src/hooks/queries.js
```diff
@@ -101,6 +101,34 @@ export const usePaymentRequestQueries = (paymentRequestId) => {
...
+  const { data: move = {}, ...moveQuery } = useQuery([MOVES, moveCode], getMove);
+
+  const moveId = move?.id;
+  const orderId = move?.ordersId;
+
+  const { data: { orders } = {}, ...orderQuery } = useQuery([ORDERS, orderId], getOrder, {
+    enabled: !!orderId,
+  });
+
+  const order = Object.values(orders || {})?.[0];
```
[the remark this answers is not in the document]
YanZ777: Probably, I'd like to do this in a separate task though.

src/services/ghcApi.js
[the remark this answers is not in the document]
felipe-lee: Ah ok, sounds good

src/services/ghcApi.js
felipe-lee: Awesome, I'll do a merge locally for the demo and see if it works.

src/constants/routes.js
felipe-lee: Ah ok. Well I guess we'll see what design says about the URL for this page given Duncan's comment and deal with it then. Possibly in a separate ticket at this point.

felipe-lee (approved):

PR #6621, opened 2021-05-13
MB-8077 Speed up models tests

monfresh:
**Description**

The main purpose of this PR is to speed up the models tests by 4x to 5x:
from about 100s down to about 20-30s. The main reason our server tests are 
slow is because we truncate the DB in between each test. In this PR, we 
remove the call to suite.TruncateAll from the SetupTest function, and 
only call it in tests that require it.

This PR also refactors models tests to remove unnecessary DB calls by 
calling `Validate` instead of `ValidateAndCreate` or `ValidateAndSave`, 
and by using the stubbed versions of testdatagen objects.

## Reviewer Notes

1. On the `master` branch, run `make db_test_reset db_test_migrate`
1. On the `master` branch, run `go test ./pkg/models -count=1`. Note how long it takes.
2. Run `go test ./pkg/models -count=1` on this branch. Verify that it runs much faster.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8077) for this change

### DIFF
--- a/pkg/models/move_document_test.go
+++ b/pkg/models/move_document_test.go
@@ -99,7 +99,7 @@ func (suite *ModelSuite) TestFetchApprovedMovingExpenseDocuments() {
 	}
 
 	// When: the logged in user is an office user
-	officeUser := testdatagen.MakeDefaultOfficeUser(suite.DB())
+	officeUser := testdatagen.MakeStubbedOfficeUser(suite.DB())
 	session.UserID = *officeUser.UserID
 	session.OfficeUserID = officeUser.ID
 	session.ApplicationName = auth.OfficeApp
@@ -217,7 +217,7 @@ func (suite *ModelSuite) TestFetchMovingExpenseDocuments() {
 func (suite *ModelSuite) TestFetchMovingExpenseDocumentsAuth() {
 	ppm := testdatagen.MakeDefaultPPM(suite.DB())
 	sm := ppm.Move.Orders.ServiceMember
-	officeUser := testdatagen.MakeDefaultOfficeUser(suite.DB())
+	officeUser := testdatagen.MakeStubbedOfficeUser(suite.DB())
 	authorizedSession := &auth.Session{
 		ApplicationName: auth.MilApp,
 		UserID:          sm.UserID,
@@ -278,7 +278,7 @@ func (suite *ModelSuite) TestFetchMoveDocument() {
 	}
 
 	// When: the logged in user is an office user
-	officeUser := testdatagen.MakeDefaultOfficeUser(suite.DB())
+	officeUser := testdatagen.MakeStubbedOfficeUser(suite.DB())
 	session.UserID = *officeUser.UserID
 	session.OfficeUserID = officeUser.ID
 	session.ApplicationName = auth.OfficeApp
--- a/pkg/models/move_test.go
+++ b/pkg/models/move_test.go
@@ -129,122 +129,87 @@ func (suite *ModelSuite) TestFetchMove() {
 	})
 }
 
-func (suite *ModelSuite) TestMoveCancellationWithReason() {
-	orders := testdatagen.MakeDefaultOrder(suite.DB())
-	orders.Status = OrderStatusSUBMITTED // NEVER do this outside of a test.
-	suite.MustSave(&orders)
-	testdatagen.MakeDefaultContractor(suite.DB())
+func (suite *ModelSuite) TestMoveCancellation() {
+	suite.Run("defaults to nil reason if empty string provided", func() {
+		move := testdatagen.MakeMove(suite.DB(), testdatagen.Assertions{Stub: true})
+		err := move.Cancel("")
 
-	selectedMoveType := SelectedMoveTypeHHGPPM
+		suite.NoError(err)
+		suite.Equal(MoveStatusCANCELED, move.Status, "expected Canceled")
+		suite.Nil(move.CancelReason)
+	})
 
-	moveOptions := MoveOptions{
-		SelectedType: &selectedMoveType,
-		Show:         swag.Bool(true),
-	}
-	move, verrs, err := orders.CreateNewMove(suite.DB(), moveOptions)
-	suite.NoError(err)
-	suite.False(verrs.HasAny(), "failed to validate move")
-	move.Orders = orders
-	reason := "SM's orders revoked"
+	suite.Run("adds reason if provided", func() {
+		move := testdatagen.MakeMove(suite.DB(), testdatagen.Assertions{Stub: true})
 
-	// Check to ensure move shows SUBMITTED before Cancel()
-	err = move.Submit()
-	suite.NoError(err)
-	suite.Equal(MoveStatusSUBMITTED, move.Status, "expected Submitted")
+		reason := "SM's orders revoked"
+		err := move.Cancel(reason)
 
-	// Can cancel move, and status changes as expected
-	err = move.Cancel(reason)
-	suite.NoError(err)
-	suite.Equal(MoveStatusCANCELED, move.Status, "expected Canceled")
-	suite.Equal(&reason, move.CancelReason, "expected 'SM's orders revoked'")
+		suite.NoError(err)
+		suite.Equal(MoveStatusCANCELED, move.Status, "expected Canceled")
+		suite.Equal(&reason, move.CancelReason, "expected 'SM's orders revoked'")
+	})
 
+	suite.Run("does not cancel a move that is already canceled", func() {
+		move := testdatagen.MakeMove(suite.DB(), testdatagen.Assertions{Stub: true})
+		move.Status = MoveStatusCANCELED
+
+		err := move.Cancel("")
+
+		suite.Error(err)
+		suite.Contains(err.Error(), "Cannot cancel a move that is already canceled.")
+	})
+
+	suite.Run("cancels PPM and Order when move is canceled", func() {
+		move := testdatagen.MakeMove(suite.DB(), testdatagen.Assertions{Stub: true})
+
+		// Create PPM on this move
+		advance := BuildDraftReimbursement(1000, MethodOfReceiptMILPAY)
+		ppm := testdatagen.MakePPM(suite.DB(), testdatagen.Assertions{
+			PersonallyProcuredMove: PersonallyProcuredMove{
+				Move:      move,
+				MoveID:    move.ID,
+				Status:    PPMStatusDRAFT,
+				Advance:   &advance,
+				AdvanceID: &advance.ID,
+			},
+			Stub: true,
+		})
+		move.PersonallyProcuredMoves = append(move.PersonallyProcuredMoves, ppm)
+
+		err := move.Cancel("")
+
+		suite.NoError(err)
+		suite.Equal(MoveStatusCANCELED, move.Status, "expected Canceled")
+		suite.Equal(PPMStatusCANCELED, move.PersonallyProcuredMoves[0].Status, "expected Canceled")
+		suite.Equal(OrderStatusCANCELED, move.Orders.Status, "expected Canceled")
+	})
 }
 
-func (suite *ModelSuite) TestMoveStateMachine() {
-	orders := testdatagen.MakeDefaultOrder(suite.DB())
-	orders.Status = OrderStatusSUBMITTED // NEVER do this outside of a test.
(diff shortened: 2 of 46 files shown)


pkg/models/order_test.go
```diff
@@ -297,64 +311,9 @@ func (suite *ModelSuite) TestOrderStateMachine() {
...
-	suite.NoError(err)
-
-	reason := "Mistaken identity"
-	err = move.Cancel(reason)
-	suite.NoError(err)
-	suite.Equal(MoveStatusCANCELED, move.Status, "expected Canceled")
-	suite.Equal(OrderStatusCANCELED, move.Orders.Status, "expected Canceled")
-
-}
-
```
monfresh, line 356: This is a duplicate test that is already covered in the Move model test.

pkg/models/mto_agents_test.go
```diff
@@ -29,21 +27,3 @@ func (suite *ModelSuite) TestMTOAgentValidation() {
...
-	suite.Equal(MTOAgent.MTOAgentType, models.MTOAgentReleasing)
-	// test with assertions
-	someFirstName := "Some other agent name"
-	otherMTOAgent := testdatagen.MakeMTOAgent(suite.DB(), testdatagen.Assertions{
-		MTOAgent: models.MTOAgent{
-			FirstName: &someFirstName,
-		},
-	})
-	suite.Equal(otherMTOAgent.FirstName, swag.String("Some other agent name"))
-}
```
monfresh, line 49: This is testing the testdatagen function itself and nothing else. Our testdatagen functions are already implicitly tested by the tests that call them.

pkg/models/payment_request_to_interchange_control_number_test.go
```diff
@@ -6,22 +6,9 @@ import (
...
-		validPR2ICN := testdatagen.MakePaymentRequestToInterchangeControlNumber(suite.DB(), testdatagen.Assertions{
-			PaymentRequestToInterchangeControlNumber: models.PaymentRequestToInterchangeControlNumber{
-				PaymentRequestID: paymentRequest.ID,
-			}})
-		suite.Equal(paymentRequest.ID, validPR2ICN.PaymentRequestID)
-		err := suite.DB().Load(&validPR2ICN, "PaymentRequest")
-		suite.NoError(err)
-		suite.Equal(paymentRequest.ID, validPR2ICN.PaymentRequest.ID)
-	})
-
```
monfresh, line 24: This is testing the testdatagen function itself and nothing else. Our testdatagen functions are already implicitly tested by the tests that call them.

pkg/models/prime_upload_test.go
```diff
@@ -24,76 +19,22 @@ func (suite *ModelSuite) Test_PrimeUploadCreate() {
...
-	if err != nil {
-		t.Fatalf("could not save PrimeUpload: %v", err)
-	}
-
-	if verrs.Count() != 0 {
-		t.Errorf("did not expect PrimeUpload validation errors: %v", verrs)
-	}
-}
-
-func (suite *ModelSuite) Test_PrimeUploadCreateWithID() {
```
monfresh, line 53: I combined this test with the previous one by adding the `ID` field. There's nothing special that we are doing to allow the `ID` field to be set. The fact that you can specify an ID and that it will get saved to the DB is handled by Pop. We shouldn't be writing tests for Pop functionality.

pkg/models/upload_test.go
```diff
@@ -1,64 +1,26 @@
...
-	}
-
-	if verrs.Count() != 0 {
-		t.Errorf("did not expect UserUpload validation errors: %v", verrs)
-	}
+	var expErrors = map[string][]string{}
+	suite.verifyValidationErrors(&upload, expErrors)
 }
 
-func (suite *ModelSuite) Test_UploadCreateWithID() {
```
monfresh, line 33: Same comment here about this test being unnecessary (as above with the prime upload test).

pkg/models/upload_test.go
```diff
@@ -71,97 +33,3 @@ func (suite *ModelSuite) Test_UploadValidations() {
...
-	upUser, _ := models.FetchUserUpload(suite.DB(), &session, uploadUser.ID)
-	suite.Equal(upUser.UploadID, upload.ID)
-	suite.Equal(upUser.Upload.ID, upload.ID)
-	suite.Equal(upUser.ID, uploadUser.ID)
-
-	upUser, _ = models.FetchUserUploadFromUploadID(suite.DB(), &session, upload.ID)
-	suite.Equal(upUser.UploadID, upload.ID)
-	suite.Equal(upUser.Upload.ID, upload.ID)
-	suite.Equal(upUser.ID, uploadUser.ID)
-}
```
monfresh, line 125: This test is testing a method on the `UserUpload` model, but this file is for the `Upload` model. Both of these functions are already tested in `user_upload_test.go` below.

pkg/models/upload_test.go
```diff
@@ -71,97 +33,3 @@ func (suite *ModelSuite) Test_UploadValidations() {
...
-	//RA: in a unit test, then there is no risk
-	//RA Developer Status: Mitigated
-	//RA Validator Status: Mitigated
-	//RA Modified Severity: N/A
-	models.DeleteUpload(suite.DB(), &upload) // nolint:errcheck
-	up, _ := models.FetchUserUpload(suite.DB(), &session, upload.ID)
-
-	// fetches a nil upload
-	suite.Equal(up.ID, uuid.Nil)
-}
```
monfresh, line 167: This is a duplicate of a test in `user_upload_test.go`

pkg/models/user_upload_test.go
```diff
@@ -24,75 +22,16 @@ func (suite *ModelSuite) Test_UserUploadCreate() {
...
-	if err != nil {
-		t.Fatalf("could not save UserUpload: %v", err)
-	}
-
-	if verrs.Count() != 0 {
-		t.Errorf("did not expect UserUpload validation errors: %v", verrs)
-	}
-}
-
-func (suite *ModelSuite) Test_UserUploadCreateWithID() {
```
monfresh, line 53: Same comment here as with the other unnecessary tests that test that ID can be set.

pkg/services/mto_shipment/mto_shipment_updater_test.go
```diff
@@ -487,14 +487,14 @@ func (suite *MTOShipmentServiceSuite) TestUpdateMTOShipmentStatus() {
 		currentTime := time.Now()
 		diff := currentTime.Sub(*actualApprovedAt)
 		diffInSeconds := diff.Seconds()
-		oneSecond := 1.000000
+		twoSeconds := 2.000000
```
monfresh, line 490: This was failing intermittently because it took longer than 1 second to perform the update. I bumped it up to 2 seconds and ran the test suite a few times without issues.

pkg/models/move_document_test.go
```diff
@@ -99,7 +99,7 @@ func (suite *ModelSuite) TestFetchApprovedMovingExpenseDocuments() {
 	}
 
 	// When: the logged in user is an office user
-	officeUser := testdatagen.MakeDefaultOfficeUser(suite.DB())
+	officeUser := testdatagen.MakeStubbedOfficeUser(suite.DB())
```
duncan-truss, line 102: Should this user have the ppm role?

pkg/models/move_test.go
```diff
@@ -129,122 +129,87 @@ func (suite *ModelSuite) TestFetchMove() {
...
-		SelectedType: &selectedMoveType,
-		Show:         swag.Bool(true),
-	}
-	move, verrs, err := orders.CreateNewMove(suite.DB(), moveOptions)
-	suite.NoError(err)
-	suite.False(verrs.HasAny(), "failed to validate move")
-	move.Orders = orders
-	reason := "SM's orders revoked"
+	suite.Run("adds reason if provided", func() {
+		move := testdatagen.MakeMove(suite.DB(), testdatagen.Assertions{Stub: true})
```
duncan-truss, line 143: I think this move is in the draft status whereas in the old test it was in the submitted status.

pkg/models/move_test.go
```diff
@@ -129,122 +129,87 @@ func (suite *ModelSuite) TestFetchMove() {
...
-	suite.NoError(err)
-	suite.Equal(MoveStatusCANCELED, move.Status, "expected Canceled")
-	suite.Equal(&reason, move.CancelReason, "expected 'SM's orders revoked'")
+		suite.NoError(err)
+		suite.Equal(MoveStatusCANCELED, move.Status, "expected Canceled")
+		suite.Equal(&reason, move.CancelReason, "expected 'SM's orders revoked'")
+	})
 
-}
+	suite.Run("does not cancel a move that is already canceled", func() {
```
duncan-truss, line 153: Do we still needs these because of the PPM even though it's now tested under your move router code?

pkg/models/move_document_test.go
monfresh: I doesn't matter for this test. `FetchMoveDocuments` only checks if the user is an office user.

pkg/models/move_test.go
monfresh: The move router tests don't test cancellation yet. There's a JIRA story to move the tests from the model to the router.

pkg/models/move_test.go
monfresh: It doesn't matter for cancellation. A move can be canceled from any status.

pkg/models/user_test.go
```diff
@@ -26,18 +24,12 @@ func (suite *ModelSuite) TestUserCreation() {
 		LoginGovEmail: userEmail,
 	}
 
-	if verrs, err := suite.DB().ValidateAndCreate(&newUser); err != nil || verrs.HasAny() {
```
duncan-truss, line 29: Sometimes we do have database check constraints like enum types or where the model doesn't match the db such as a nullable type (run `model-vet` to see). I assume those errors would surface elsewhere but there are cases where Validate and ValidateAndCreate won't be 1:1

pkg/models/user_upload_test.go
```diff
@@ -24,75 +22,16 @@ func (suite *ModelSuite) Test_UserUploadCreate() {
...
-	}
-
-	if verrs.Count() != 0 {
-		t.Errorf("did not expect UserUpload validation errors: %v", verrs)
-	}
-
-	if uploadUser.ID.String() != id.String() {
-		t.Errorf("wrong uuid for UserUpload: expected %s, got %s", id.String(), uploadUser.ID.String())
-	}
+	suite.verifyValidationErrors(&uploadUser, expErrors)
```
duncan-truss, line 34: This make me think there should be validation errors but we are testing for an empty map?  Maybe we can be more explicit in the naming.

pkg/models/user_test.go
monfresh: Agreed. From what I've seen, we have separate tests for those, such as the duplicate UUID test below this one.

pkg/models/user_upload_test.go
monfresh: Good point. This is an existing function. Instead of using it when no errors are expected, we could be consistent and use these assertions:
```golang
suite.NoError(err)
suite.False(verrs.HasAny(), "Error validating model")
```

pkg/testdatagen/make_service_member.go
```diff
@@ -141,3 +142,15 @@ func MakeExtendedServiceMember(db *pop.Connection, assertions Assertions) models
...
 	return serviceMember
 }
+
+// MakeStubbedServiceMember returns a stubbed service member that is not stored in the DB
+func MakeStubbedServiceMember(db *pop.Connection) models.ServiceMember {
+	return MakeServiceMember(db, Assertions{
+		ServiceMember: models.ServiceMember{
+			ID:     uuid.Must(uuid.NewV4()),
+			User:   MakeStubbedUser(db),
+			UserID: uuid.Must(uuid.NewV4()),
```
duncan-truss, line 152: The UserID and the User.ID are not going to match here because they are each generating their own uuid.

duncan-truss (approved):
Nice work, it saved me a whopping 60 seconds locally.

pkg/testdatagen/make_service_member.go
monfresh: Good catch! I just pushed a fix.

donaldthai (approved):
🚀

PR #6624, opened 2021-05-13
[MB-7254] Create randomized moves for services counseling in db dev seed 

duncan-truss:
## Description

As part of load testing the GHC API, we may need to rely on the queue queries to return a set of pre-populated moves that we can store and then query and manipulate in different ways.  The services counseling queue was quite bare outside of a few test moves.  While it's still a good idea to create new moves from load testing I figure that will be harder to keep up to date and porting the logic over may be a good amount of work.

I created a randomized move function that
- Randomly chooses a fake name from our predefined list
- Randomly chooses a submitted at date in the past year
- Randomly chooses an affiliation so we get moves in the USMC queue as well
- Randomly chooses origin duty stations (within the GBLOC if non-Marines) and destination duty stations
- Randomly chooses between the Needs Counseling and Services Counseling Completed statuses

If I had used the basic `math/rand` instead of `pkg/random` package I would have needed annotations everywhere although it is a test file.

We should probably stop the MakeDutyStation factory from creating ones with random names and assign an existing one or create a default one if none exist.  Otherwise we will end up with unreadable names.

Here is the corresponding [milmove_load_testing PR ](https://github.com/transcom/milmove_load_testing/pull/66)for fetching queue and move tasks

I did not convert the TOO moves queue yet since it had more moves in it and didn't know if I'd be successful with this first part.

In the future we may want a way to reconcile the orders dates with the move/shipment dates.  Same with the duty station addresses and service member addresses.

## Reviewer Notes

## Setup

```sh
make db_dev_reset
make db_dev_e2e_populate
```

1. Local login as the Services Counselor user either the regular or USMC version
2. Toggle the move statuses to see there are moves in the needs counseling and services counseling completed statuses.
3. Check the variation in the different queue row data.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7254) for this change

## Screenshots

Screen Shot 2021-05-13 at 1 36 50 PM


monfresh:
What do you think about putting this in a separate script, or put it behind an env var? This adds about 16 seconds to `make db_dev_e2e_populate`, and most people will not need those 50 moves in their day-to-day work.

duncan-truss:
> What do you think about putting this in a separate script, or put it behind an env var? This adds about 16 seconds to `make db_dev_e2e_populate`, and most people will not need those 50 moves in their day-to-day work.

@monfresh I timed it on my machine and am seeing no difference but our machines may vary:

```
make db_dev_reset
make db_dev_migrate
make db_dev_e2e_populate
```

Screen Shot 2021-05-13 at 3 55 33 PM

Screen Shot 2021-05-13 at 3 55 42 PM

If it is a substantial delay I can put it behind a flag or make a seperate load testing scenario.

monfresh:
On the master branch, `make db_dev_e2e_populate` takes around 30-32 seconds consistently. How is it only taking 6 seconds on your machine?

When I switch to your branch, it takes about 40-42 seconds.

duncan-truss:
> On the master branch, `make db_dev_e2e_populate` takes around 30-32 seconds consistently. How is it only taking 6 seconds on your machine?
> 
> When I switch to your branch, it takes about 40-42 seconds.

@monfresh The last number gives the total elapsed time so it was 21 seconds for master versus 28 seconds on the new branch.

monfresh:
Ah, I missed that. I still think it's worth it to have it not run that part by default. Any additional time adds up. We should probably also not have that loop of 12 moves be on by default, but we can fix that later.

monfresh:
After thinking about this some more, I realized that the only reason we added those initial 12 moves was to test pagination when we were first working on it. We know it works, so I don't think we need those 12 moves anymore. Pagination is not something we need to continually test locally. So, for this PR, I would propose one of these options:

- we replace those 12 moves with the random ones you created, but keep it at 12. That way, the script doesn't take longer to run.
- we delete that original loop, and we place the new random moves loop behind a feature flag, or comment it out by default and let people know to uncomment it if they need to do load testing.

duncan-truss:
> After thinking about this some more, I realized that the only reason we added those initial 12 moves was to test pagination when we were first working on it. We know it works, so I don't think we need those 12 moves anymore. Pagination is not something we need to continually test locally. So, for this PR, I would propose one of these options:
> 
>     * we replace those 12 moves with the random ones you created, but keep it at 12. That way, the script doesn't take longer to run.
> 
>     * we delete that original loop, and we place the new random moves loop behind a feature flag, or comment it out by default and let people know to uncomment it if they need to do load testing.

@monfresh Yea I remember it was to get it over the lowest possible page limit.  I would need to do more work thought to make random TOO and TIO queue data realistic.  I think it's worth filling up a single page with seed data that is the point of the scenario after all that you have a mix of data that you can use to test different features and you're more likely to use the app and find issues when it a fresh environment with no data in my opinion. If people don't want that then they can run `db_dev_reset` and `db_dev_migrate` and create their own users and data from scratch.

I think design appreciates having enough data where they don't have to do a ton of manual work and remember this seed is now being loaded when the ephemeral environments are spinning up so there is value there.  So I think a little extra processing time is a good tradeoff to save others time.  Other teams are more likely to test things in the office side if we make it more accessible and not put silly blockers in their way.

monfresh:
If I comment this out:
```
for i := 1; i < 12; i++ {
    createDefaultHHGMoveWithPaymentRequest(db, userUploader, primeUploader, logger, models.AffiliationAIRFORCE)
}
```
and run `make db_dev_e2e_populate` in your branch, I end up with 8 moves in my SC queue, 18 in my TOO queue, and 11 payment requests in my TIO queue. Is that not enough data to test with?

duncan-truss:
> If I comment this out:
> 
> ```
> for i := 1; i < 12; i++ {
>     createDefaultHHGMoveWithPaymentRequest(db, userUploader, primeUploader, logger, models.AffiliationAIRFORCE)
> }
> ```
> 
> and run `make db_dev_e2e_populate` in your branch, I end up with 8 moves in my SC queue, 18 in my TOO queue, and 11 payment requests in my TIO queue. Is that not enough data to test with?

I guess it depends what you're testing, no I don't think they are sufficiently diverse or necessarily reflect reality and should also simulate some realistic load when using the app.  I wasn't working with TIO moves in this PR but I'm sure we'll get there for GHC load testing.

monfresh:
I understand the diversity and realism problem. My question is what do we lose by removing the 12 identical moves generated by looping through `createDefaultHHGMoveWithPaymentRequest`? Conversely, how does keeping those 12 identical moves solve the diversity and realism problems?

donaldthai (approved):
This is very cool! Data looks way more realistic now 🚀

Ryan-Koch (approved):

PR #6625, opened 2021-05-13
hardcode ppm estimates to account for no tariff400ng data: MB-8060

lynzt:
## Description

We have stopped loading rate data which is used for pricing a PPM moves. The data from our last load has data thru May 15, 2021 so we have started to see errors when pricing PPM moves that use dates beyond May 15th. PPMs will be addressed in Outcome 7 and till then, we're going to hardcode the return value of a PPM move to prevent errors when we have missing PPM rate data.

## Reviewer Notes

Ensure I didn't bypass a test that should not be bypassed.

## Setup
run the server tests
`make server_test`

run the e2e tests
`yarn install`
`make server_run`
`make client_run`
`yarn test:e2e` -- i never got the e2e tests to pass locally... I might try the docker container next, but they're passing on circle.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8060)


pkg/rateengine/rateengine.go
```diff
@@ -96,70 +97,35 @@ func (re *RateEngine) computePPM(
 	lhDiscount unit.DiscountRate,
 	sitDiscount unit.DiscountRate) (cost CostComputation, err error) {
 
+	fmt.Println("*** computePPM")
```
nelztruss, line 100: debug print line accidentally left in?

pkg/rateengine/rateengine.go
```diff
@@ -176,6 +142,97 @@ func (re *RateEngine) computePPM(
 	return cost, nil
 }
 
+// // computePPM Calculates the cost of a PPM move.
```
nelztruss, line 145: suggestion: just remove the commented out code, it'll still exist in Git history if it's ever needed again... Optionally if needed, leave a prose comment pointing out that this logic exists in history if someone might need it again.

nelztruss (approved):
Just a couple of non-blocking nits

pkg/rateengine/rateengine.go
```diff
@@ -176,6 +142,97 @@ func (re *RateEngine) computePPM(
...
+// 	cost.Scale(prorateFactor)
+
+// 	re.logger.Info("PPM cost computation",
+// 		zap.String("moveLocator", re.move.Locator),
+// 		zap.Object("cost", cost),
+// 	)
+
+// 	return cost, nil
+// }
+
```
felipe-lee, line 235: I think this is more me not knowing enough about PPM estimates and in particular where tarriff400ng data comes in. Do the other PPM calculation functions not need to be modified? Like one of the error messages in the next one mentions linehaul cost couldn't be calculated. The message in the jira ticket was also about linehaul, but idk if there's different kinds and so the other one isn't an issue.

pkg/rateengine/rateengine.go
lynzt: I left this function commented out because I was not going to go to step thru the code and find other functions that are called only from here and delete them as part of this effort... I'll add a comment to mention fixing the issues in outcome 7

pkg/rateengine/rateengine.go
lynzt: yup... good catch - deleting

pkg/rateengine/rateengine.go
lynzt: `computePPM` is the main function that calculates the PPM estimates. It is used by other PPM functions but this is the base one... If you take a look, the next function you mention calls `computePPM` after looking to see if there's a discount.

nelztruss (approved):
🚢

PR #6630, opened 2021-05-13
[MB-4004] Fix bugs in our custom SQL parser/splitter

mboren:
## Description

We do a bit of parsing to the SQL code in our migrations to split the files into individual statements that we can run with `RawQuery`. As you can imagine, this is a bit error prone.

This splitting code works for all the migrations that we've written, but has some edge cases that we hit when we try to run database dumps through it. This is preventing us from squashing migrations.

Some sources of problems:
- Escaping quotes. To escape quotes within a string in SQL, you double them up. So `'I can''t believe they said ''hello'''` yields `I can't believe they said 'hello'`. Strings can include semicolons and newlines so we have to know when we're inside of quotes in order to avoid splitting statements prematurely.
- Loading data via `COPY ... FROM stdin;` is much faster than loading with `INSERT` statements, and we're loading a LOT of records in the squashed migrations, so it's important to use this syntax. However, the way records are formatted when loaded with `COPY` is a bit weird. The column values are separated by tabs, strings are not quoted, and lines do not end with semicolons.

So we have to keep track of whether we're in a quoted string accurately, which is challenging, and we have to have different rules for statements loaded with COPY.

- escaped quotes at the end of a string weren't counted correctly
- semicolons in strings in data that was within a `COPY` would split the statement
- empty columns in data loaded with `COPY` resulted in trailing whitespace, which was trimmed off.

### quote counting
My addition to the quote counting is a bit hacky and fragile. I took a stab at something that was integrated better but it's hard to get out of that code unscathed, and i backed off.
I feel like all of the quote counting logic could be replaced by toggling quoted status whenever we hit an odd number of consecutive quotes. Even numbered groups of consecutive quotes are either empty strings (`''`), escaped quotes within strings `'...''...'`, or strings containing only escaped quotes (`''''`). I tried this and i ran into problems and idk that it's worth it.

### trimming whitespace
I changed from trimming left and right whitespace to just trimming left. Records loaded with `COPY` could have significant whitespace on either side, if they start or end with columns containing empty strings, but in practice I've only seen it at the end of records, as the first column is generally the primary key.
Maybe it makes more sense to not trim any whitespace by default, and have the splitter do it just for non-copy lines.

## Setup
[largely taken from this doc](https://docs.google.com/document/d/1SGvItAZ27xtvzIgGHl582ySp_XwuVjHLkUlz6JvYupU/edit#)
We're going to get a dump from the Test DB and then use it to build up the Dev DB so that we can compare the two.

```sh
rm bin/milmove
make bin/milmove
make db_test_reset
make db_test_migrate
pg_dumpall -h localhost -U postgres -p 5433 --roles-only --no-owner --file squashed_roles.sql 
pg_dump -h localhost -U postgres -p 5433 test_db --schema-only --no-owner --file squashed_schema.sql
pg_dump -h localhost -U postgres -p 5433 test_db -a -f squashed_data.sql
```
**Make the following changes to the new files:**
- squashed_roles.sql
   - Comment out the “CREATE ROLE postgres” and the “ALTER ROLE postgres …” lines.  The Postgres docker container we’re using sets up that role by default.
- squashed_schema.sql
  - Comment out the creation of the table and any indexes on the schema_migration table.  The schema_migration table is created by the migration process itself.  Leave the schema_migration comments and grants in there.

**Make migrations**
- Go to the migrations manifest and comment out every existing migration line (with a “#” at the beginning of the line).
- Make the following empty migrations (note that this will also change the migrations manifest -- the files should be added to the manifest in this order):
  - A “secure” migration for the roles: `generate-secure-migration squashed_roles`
  - A “schema” migration for the schema: `milmove gen migration -n squashed_schema`
  - A “secure” migration for the data: `generate-secure-migration squashed_data`
- Copy the corresponding files from above into each of the generated migration files.
- Run the new migrations: `make db_dev_reset db_dev_migrate`

### Testing with deployed migrations
You can get deployed migrations with `make run_stg_migrations` or `make run_prd_migrations`.
Then you can dump those using the instructions above, except you'll need to change the port to `5434` in the `pg_dump` commands.

### Comparing databases
To make sure the data is right, we need to compare the records in the original database that we dumped from to the database that we applied the new migrations to.
I used [PostgresCompare](https://www.postgrescompare.com/) for this.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4004) for this change
* [this ](https://docs.google.com/document/d/1SGvItAZ27xtvzIgGHl582ySp_XwuVjHLkUlz6JvYupU/edit#) explains more about the approach used.


jacquelineIO (commented):
I was able to run the steps successfully. One thing to note is that you will have to rebuild the `milmove` binary.

```sh
rm bin/milmove; make bin/milmove
```

pkg/migrate/ReadInSQL.go
```diff
@@ -28,5 +29,9 @@ func ReadInSQLLine(line string, dropComments bool, dropSearchPath bool) string {
 		return ""
 	}
 
-	return strings.TrimSpace(line)
+	// When loading data with `COPY ... FROM stdin;` we can have trailing tabs that are significant
+	// if a record has an empty string value for the last column.
+	// We could also have preceding whitespace that is significant if the first column value is empty,
+	// but that has not happened in practice yet.
+	return strings.TrimLeftFunc(line, unicode.IsSpace)
```
reggieriser, line 36: Should we have a test that incorporates whether the trim happened as desired?  FWIW, it doesn't look like we had something that exercised the trimming before, either.

reggieriser: I just saw below that there's a `TestSplitStatementsCopyFromStdinTrailingEmptyColumns` that's testing out the trimming via `SplitStatements_test.go`.  So disregard my earlier comment; we're likely good here unless you want something that does a test in `ReadInSQL_test.go`.

reggieriser (approved):
LGTM.  Tried both local migrations and stg migrations and didn't see any unexpected diffs when comparing the before/after via the method in the instructions.

PR #6635, opened 2021-05-14
MB-8073 update payment request to interchange control number constraint

gidjin:
## Description

As part of the debugging process for sending EDI858s and receiving EDI997s it was discovered that a 997 file can have the same ICN assigned by the sender as the original EDI858 used in a previous send. This lead to violation of the constraint on the table that tracks ICNs to EDI Type to Payment Requests. This PR updates the constraint so this is not an issue. It also filters the lookup of a payment request to those with a matching ICN and are of type EDI858, since that is the only type we send at this time.

## Reviewer Notes

The test for failure on invalid EDI997 seemed to be testing more than it needed too, and the cases I removed were difficult to replicate after adding the support to use the EDI type from the AK2 segment. Feedback on this is welcome.

## Setup

Since you can only generate and send an 858 in staging to expect a 997 all you can do here is make sure the tests pass.
```sh
make db_test_reset db_test_migrate
go test -v --failfast ./pkg/services/invoice
```

### Acceptance

Get a payment request in the reviewed state

Review the logs

* [X] The requirements listed in [Querying the Database Safely](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#querying-the-database-safely) have been satisfied.
  * [X] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://github.com/transcom/mymove/wiki/migrate-the-database#zero-downtime-migrations))
  * [X] Have been communicated to #g-database

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8073) for this change


jacquelineIO (approved):
lgtm

migrations/app/schema/20210514045435_update_payment_request_to_interchange_control_numbers_constraint.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE payment_request_to_interchange_control_numbers ADD CONSTRAINT interchange_control_number_payment_request_id_edi_type_uniq_key UNIQUE (interchange_control_number, payment_request_id, edi_type);
```
reggieriser, line 1: I think we usually try to prefix the name of the constraint with the table name (to create a namespace to avoid possible name conflicts across tables).  But I have a feeling the identifier will be too long if we do this, so I think this is OK.

reggieriser (approved):
Migration LGTM. 🚢

migrations/app/schema/20210514045435_update_payment_request_to_interchange_control_numbers_constraint.up.sql
gidjin: yeah I had to shorten `unquie` to `uniq` because it was 2 chars over the max otherwise. Open to better name if we think of it.

PR #6608, opened 2021-05-11
MB-3171 Add endpoint for services counselor to edit an HHG shipment

mlm55:
## Description

Add ability to edit hhg shipment

## Reviewer Notes

1. Should we be checking for user type (service counselor/TOO/TIO). If so, where?

## Setup

`make server_run`
Set up [postman](https://github.com/transcom/mymove/wiki/Test-Admin-and-Office-APIs-with-Postman) and hit `http://officelocal:3000/ghc/v1/move_task_orders/:moveTaskOrderID/mto_shipments/:shipmentID`

## Code Review Verification Steps

* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#logging)
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3171) for this change

PR #6647, opened 2021-05-17
MB-8135 updateMTOShipment won't blank out addresses 

sandy-wright:
## Description

Recently, I had updated the Prime API YAML file to have a new structure for address fields in shipments so that they could each have an individual description. However, this introduced a bug to the `updateMTOShipment` endpoint wherein an address would have its values blanked out if it was not included in the request body. Fixing this involved updating the `AddressModel` function in `payload_to_model` to accurately identify blank addresses. I also added a test and caught another little bug in the meantime.

## Reviewer Notes

One other effect of the update to the Prime API's address structure is the return values. Instead of omitting the field if a null value was sent back from the server, now the response will look like:

```json
"secondaryDeliveryAddress": {
  "city": null,
  "postalCode": null,
  "state": null,
  "streetAddress1": null
},
```

There is no way to get around this if we want to have visible documentation for each of these fields. The question - is this acceptable? Is the documentation not worth the potential confusion with the response? I'd appreciate feedback on that!

## Setup

Populate your db and run your server:

```sh
```

Using the Prime API testing tool of your choice ([Prime API Client](https://github.com/transcom/prime_api_deliverable/wiki/Prime-API-Client) or [Postman](https://github.com/transcom/prime_api_deliverable/wiki/Using-Postman)):

* Run `fetchMTOUpdates` to grab a shipment's ID and eTag 
* Save a record of one of its addresses (`pickupAddress`, `destinationAddress`, `secondaryPickupAddress`, or `secondaryDeliveryAddress`) so that you can refer to it later
* Go to `updateMTOShipment`. Fill in the ID and eTag in the path params and `IfMatch` header, respectively.
* Set up a _small_ request body. Something like:
  ```json
  {
    "scheduledPickupDate": "2021-06-23"
  }
  ```
* Send the request.
* Check out your chosen address field in the response. Are all the values the same as they were before you sent the update? Then yay, that's a success!

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8135) for this change.

### DIFF
--- a/pkg/services/mto_shipment/mto_shipment_updater.go
+++ b/pkg/services/mto_shipment/mto_shipment_updater.go
@@ -280,7 +280,7 @@ func (f *mtoShipmentUpdater) updateShipmentRecord(dbShipment *models.MTOShipment
 
 		if newShipment.SecondaryPickupAddress != nil {
 			if dbShipment.SecondaryPickupAddressID != nil {
-				newShipment.PickupAddress.ID = *dbShipment.SecondaryPickupAddressID
+				newShipment.SecondaryPickupAddress.ID = *dbShipment.SecondaryPickupAddressID
 			}
 
 			err := tx.Save(newShipment.SecondaryPickupAddress)
--- a/pkg/handlers/primeapi/mto_shipment_test.go
+++ b/pkg/handlers/primeapi/mto_shipment_test.go
@@ -448,6 +448,49 @@ func (suite *HandlerSuite) TestUpdateMTOShipmentHandler() {
 
 	})
 
+	suite.T().Run("Successful PATCH - Update that does not include addresses will not blank out data", func(t *testing.T) {
+		// Get the default shipment that has PickupAddress populated:
+		addressShipment := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+			Move: move, // Prime-available move
+		})
+
+		eTagAddressShipment := string(etag.GenerateEtag(addressShipment.UpdatedAt))
+		paramsAddressShipment := mtoshipmentops.UpdateMTOShipmentParams{
+			HTTPRequest:   req,
+			MtoShipmentID: *handlers.FmtUUID(addressShipment.ID),
+			Body:          ClearNonUpdateFields(&addressShipment),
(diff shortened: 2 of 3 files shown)


pkg/services/mto_shipment/mto_shipment_updater.go
```diff
@@ -280,7 +280,7 @@ func (f *mtoShipmentUpdater) updateShipmentRecord(dbShipment *models.MTOShipment
 
 		if newShipment.SecondaryPickupAddress != nil {
 			if dbShipment.SecondaryPickupAddressID != nil {
-				newShipment.PickupAddress.ID = *dbShipment.SecondaryPickupAddressID
+				newShipment.SecondaryPickupAddress.ID = *dbShipment.SecondaryPickupAddressID
```
felipe-lee, line 283: @sandy-wright This is the test we modified in @mlm55 's pr to start seeing if the addresses are properly being updated https://github.com/transcom/mymove/blob/6d8765a8b4266dfa752b9d569ff1928333cf7d38/pkg/services/mto_shipment/mto_shipment_updater_test.go#L151-L177

felipe-lee: Specifically this:
https://github.com/transcom/mymove/pull/6608/files/813a6ebbcafebd21299d30e733db2c0e380124b2..c1f1391b2dfa798a988b405caafb869e2853895f

Just to try seeing some changes come through.

abbyoung (approved):
lgtm! thank you for catching and fixing

pkg/services/mto_shipment/mto_shipment_updater.go
pearl-truss: I pulled down this PR and added those test and I see that they're passing 🥳

sandy-wright: woo, thank you!! Should I add those tests to mine, or just let your PR take care of that?

sandy-wright: ...mm I'll actually go ahead and merge. I got them to pass too, and I would just be duplicating the work in that PR

PR #6650, opened 2021-05-18
MB-8384: txdb investigation

ahobson:
## Description

This uses go-txdb to create a transaction that is rolled back at the end of each test. This should give some speedups

## Reviewer Notes

I tried to make each commit independent and as self contained as possible, so reviewing each commit in turn might be a good strategy.

## Setup

```sh
make server_test_setup
time make server_test_standalone
```

For example:
```
# on master
$ go test -count=1 ./pkg/handlers/internalapi/
ok  	github.com/transcom/mymove/pkg/handlers/internalapi	44.086s
# on this branch
$ go test -count=1 ./pkg/handlers/internalapi/
ok  	github.com/transcom/mymove/pkg/handlers/internalapi	26.561s
```

## References

* [investigate go-txdb to speed up test suite](MB-8384) for this change

## Screenshots

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -401,8 +401,11 @@ endif
 mocks_generate: bin/mockery ## Generate mockery mocks for tests
 	go generate $$(go list ./... | grep -v \\/pkg\\/gen\\/ | grep -v \\/cmd\\/)
 
+.PHONY: server_test_setup
+server_test_setup: db_test_reset db_test_migrate redis_reset db_test_truncate
+
 .PHONY: server_test
-server_test: db_test_reset db_test_migrate redis_reset server_test_standalone ## Run server unit tests
+server_test: server_test_setup server_test_standalone ## Run server unit tests
 
 .PHONY: server_test_standalone
 server_test_standalone: ## Run server unit tests with no deps
@@ -647,16 +650,16 @@ else
 	psql postgres://postgres:$(PGPASSWORD)@localhost:$(DB_PORT_TEST)?sslmode=disable -c 'CREATE DATABASE $(DB_NAME_TEST);'
 endif
 
-.PHONY: db_test_truncate
-db_test_truncate: ## Truncate e2e db
-	@echo "Truncate the ${DB_NAME_TEST} database..."
-	psql postgres://postgres:$(PGPASSWORD)@localhost:$(DB_PORT_TEST)/$(DB_NAME_TEST)?sslmode=disable -c 'TRUNCATE users CASCADE; TRUNCATE uploads CASCADE; TRUNCATE webhook_subscriptions CASCADE; TRUNCATE traffic_distribution_lists CASCADE'
-
 .PHONY: db_test_run
 db_test_run: db_test_start db_test_create ## Run Test DB
 
 .PHONY: db_test_reset
 db_test_reset: db_test_destroy db_test_run ## Reset Test DB (destroy and run)
+
+.PHONY: db_test_truncate
+db_test_truncate:
+	@echo "Truncating ${DB_NAME_TEST} database..."
+	DB_PORT=$(DB_PORT_TEST) DB_NAME=$(DB_NAME_TEST) ./scripts/db-truncate
 
 .PHONY: db_test_migrate_standalone
 db_test_migrate_standalone: bin/milmove ## Migrate Test DB directly
--- a/pkg/handlers/ghcapi/mto_shipment_test.go
+++ b/pkg/handlers/ghcapi/mto_shipment_test.go
@@ -2,8 +2,9 @@ package ghcapi
 
 import (
 	"fmt"
+	"log"
+	"net/http"
 	"net/http/httptest"
-	"testing"
 	"time"
 
 	"github.com/transcom/mymove/pkg/models/roles"
@@ -37,7 +38,16 @@ import (
 	"github.com/transcom/mymove/pkg/testdatagen"
 )
 
-func (suite *HandlerSuite) TestListMTOShipmentsHandler() {
+type listMTOShipmentsSubtestData struct {
+	mtoAgent       models.MTOAgent
+	mtoServiceItem models.MTOServiceItem
+	shipments      models.MTOShipments
+	params         mtoshipmentops.ListMTOShipmentsParams
+}
+
+func (suite *HandlerSuite) makeListMTOShipmentsSubtestData() (subtestData *listMTOShipmentsSubtestData) {
+	subtestData = &listMTOShipmentsSubtestData{}
+
 	mto := testdatagen.MakeDefaultMove(suite.DB())
 	mtoShipment := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
 		Move: mto,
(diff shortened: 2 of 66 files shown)


monfresh:
@ahobson I'm reviewing it this week!

ahobson:
> @ahobson I'm reviewing it this week!

Please let me know how I can help make it easier to review. I could do a walkthrough via zoom if that would be helpful

monfresh:
A walkthrough would be great. I put a meeting on your calendar for tomorrow. Feel free to move to a better time if that one doesn't work.

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -683,22 +725,33 @@ func (suite *HandlerSuite) TestListMTOShipmentsHandler() {
...
+		firstCreatedShipment := subtestData.shipments[0]
+		nextCreatedShipment := subtestData.shipments[1]
+		if nextCreatedShipment.CreatedAt.Before(firstCreatedShipment.CreatedAt) {
+			firstCreatedShipment = subtestData.shipments[1]
+			nextCreatedShipment = subtestData.shipments[0]
+		}
+		_, err := time.Parse(time.RFC3339, okResponse.Payload[0].CreatedAt.String())
+		if err != nil {
+			suite.TestLogger().Fatal("unable to parse string time")
+		}
```
felipe-lee, line 744: This part can be removed. I'd deleted it in a previous pr because I changed how we tested what this bit was trying to test. It's essentially lines 746-751 now do what this was doing, without resorting to checking times.

ahobson: I believe I addressed this

pkg/testingsuite/pop_suite_txn.go
```diff
@@ -0,0 +1,168 @@
...
+	return nil, nil
+}
+
+// TransactionContextOptions returns the current pop.Transaction
+// Needed to implement the pop.store interface, but won't actually be called
+func (store *PopSuiteTxnStore) TransactionContextOptions(ctx context.Context, _ *sql.TxOptions) (*pop.Tx, error) {
+	return nil, nil
+}
+
+// WrapTxnPopConnection sets up the pop Connection for this test
```
monfresh, line 101: Did you mean `openTxnPopConnection` in this comment to match the function name?

pkg/testingsuite/pop_suite_txn.go
```diff
@@ -0,0 +1,168 @@
...
+	"fmt"
+	"log"
+	"regexp"
+
+	"github.com/DATA-DOG/go-txdb"
+	"github.com/gobuffalo/pop/v5"
+	"github.com/jmoiron/sqlx"
+)
+
+// use a driver that sqlx knows uses a doller binding like postgres
```
monfresh, line 15: doller -> dollar

pkg/parser/pricing/parse_test.go
```diff
@@ -39,20 +32,17 @@ func TestPricingParserSuite(t *testing.T) {
...
 
-	hs.tableFromSliceCreator = dbtools.NewTableFromSliceCreator(hs.DB(), logger, true, false)
-
 	hs.xlsxFile, err = xlsx.OpenFile(hs.xlsxFilename)
 	if err != nil {
 		logger.Panic("could not open XLSX file", zap.Error(err))
 	}
 
 	suite.Run(t, hs)
-	hs.PopTestSuite.TearDown()
```
monfresh, line 55: Was this removal intentional?

pkg/services/ghcimport/ghc_rateengine_importer_test.go
```diff
@@ -93,7 +93,6 @@ func TestGHCRateEngineImportSuite(t *testing.T) {
 	}
 
 	suite.Run(t, hs)
-	hs.PopTestSuite.TearDown()
```
monfresh, line 96: In this case, I think this removal makes sense because this package defines a `TearDownSuite` function that calls `suite.PopTestSuite.TearDown()`. Is that correct?

pkg/services/move_task_order/move_task_order_service_test.go
```diff
@@ -12,15 +12,9 @@ type MoveTaskOrderServiceSuite struct {
...
-	suite.FatalNoError(err)
-}
-
 func TestMoveTaskOrderServiceSuite(t *testing.T) {
 	ts := &MoveTaskOrderServiceSuite{
-		testingsuite.NewPopTestSuite(testingsuite.CurrentPackage()),
+		testingsuite.NewPopTestSuite(testingsuite.CurrentPackage(), testingsuite.WithPerTestTransaction()),
 	}
 	suite.Run(t, ts)
-	ts.PopTestSuite.TearDown()
```
monfresh, line 25: Was this removal intentional?

monfresh (approved):
This is great! Thanks for the hard work on this one, and for saving us time now and in the future.

pkg/handlers/ghcapi/mto_shipment_test.go
```diff
@@ -1744,30 +1875,42 @@ func (suite *HandlerSuite) TestCreateMTOShipmentHandler() {
 		suite.NotEmpty(typedResponse.Payload.InvalidFields)
 	})
 
-	suite.T().Run("POST failure - 422 - invalid input, missing pickup address", func(t *testing.T) {
-		fetcher := fetch.NewFetcher(builder)
-		creator := mtoshipment.NewMTOShipmentCreator(suite.DB(), builder, fetcher)
+	//
+	// TODO: this test panics with a null pointer exception. Duncan
+	// said he would look into a fix - ahobson 3 June 2021
```
monfresh, line 1880: This is failing because the validation errors are nil. The handler is receiving `services.InvalidInputError` from the service, but without validation errors, and then it's assuming there are validation errors and calls `payloadForValidationError`, which in turn calls `NewValidationErrorsResponse` which tries to iterate through `range verrs.Keys()`, but `verrs` is nil.

felipe-lee: There's a jira bug to fix this at some point: https://dp3.atlassian.net/browse/MB-8326

monfresh: The reason this is happening is because by the time the payload has arrived at the service, the `shipment.PickupAddress` is `nil`, and so the service is returning `services.NewInvalidInputError(uuid.Nil, nil, nil, "PickupAddress is required to create an HHG or NTS type MTO shipment")`, with `nil` validation errors.

It looks like this handler is not following the same pattern we use in other handlers, which is to pass in `validate.NewErrors()` as the last argument of `payloadForValidationError`. However, that means that the `InvalidFields` will be empty, but this test is expecting them to NOT be empty, which doesn't make sense.

It seems like we should not be calling `NewInvalidInputError` without any validation errors if we expect the `InvalidFields` to be populated.

cc @duncan-truss

duncan-truss: Yep you're right I think that's what I outlined here https://dp3.atlassian.net/browse/MB-8326

Makefile
```diff
@@ -654,6 +657,11 @@ db_test_run: db_test_start db_test_create ## Run Test DB
 .PHONY: db_test_reset
 db_test_reset: db_test_destroy db_test_run ## Reset Test DB (destroy and run)
 
+.PHONY: db_test_truncate
+db_test_truncate:
+	@echo "Truncating ${DB_NAME_TEST} database..."
+	DB_PORT=$(DB_PORT_TEST) DB_NAME=$(DB_NAME_TEST) ./scripts/db-truncate
+
```
felipe-lee, line 664: I just merged something that also added this `make` target, but doesn't do it as well. It also included a `db_dev_truncate`, which could be converted to use this same script with the appropriate env vars. Likely should be done in this pr though, otherwise we'll end up with duplicate targets.

pkg/handlers/ghcapi/mto_shipment_test.go
monfresh: Looking at the other APIs, they define a `ValidationError` and `ClientError` methods in `model_to_payload`, but when I tried doing the same for the GHC API, it failed because it's expecting a different type. The difference is that in the Swagger YAML definition of `ValidationError` for the GHC API, it's defining `additionalProperties` as `type: string`, whereas the others APIs expect an array to represent a list of errors for the invalid field.

felipe-lee (approved):
LGTM, thanks for all the work you did. Just a comment on the `Makefile` targets regarding some changes that I just merged into master

pkg/testingsuite/pop_suite_txn.go
ahobson: yep! fixing now

pkg/services/ghcimport/ghc_rateengine_importer_test.go
ahobson: Yep, that's exactly right!

pkg/handlers/ghcapi/mto_shipment_test.go
monfresh: OK, I get it now. The easiest fix is to update the `payloadForValidationError` function so that it only includes the `InvalidFields` if validation errors are present:
```golang
func payloadForValidationError(title string, detail string, instance uuid.UUID, validationErrors *validate.Errors) *ghcmessages.ValidationError {
    payload := &ghcmessages.ValidationError{
        ClientError:   *payloadForClientError(title, detail, instance),
    }

    if validationErrors != nil {
        payload.InvalidFields = handlers.NewValidationErrorsResponse(validationErrors).Errors
    }

    return payload
}
```
And then this test won't panic anymore, but the last assertion needs to be removed because the InvalidFields is going to be empty in this case. The fact that it was not empty before was probably because the test was not doing what we thought it was.

monfresh: I'll push this change up since I'm already in here.

Makefile
ahobson: Yep, converted!

PR #6658, opened 2021-05-19
MB-7388 Allow Counselor to soft delete shipment

monfresh:
## Description

Services Counselors should be able to remove shipments before submitting
the move to the TOO. Per this 
[ADR](https://github.com/transcom/mymove/blob/master/docs/adr/0038-soft-delete.md), 
we don't want to permanently delete the shipment from the DB. Instead,
we can set its `deleted_at` field to indicate it was removed.

The app should then ensure that deleted shipments are not shown to the
TOO or other places where we return shipments. Note that this has not
yet been implemented in this PR. Depending on how many places will need 
to be updated, it might make sense to do this work in a separate PR.

Also, note that I called things `Shipment` without the `MTO` prefix because we want to eventually rename `MTO` and `MoveTaskOrder` to `Move`, and because there is only one kind of shipment now. According to Reggie, there used to be a `shipments` table/model used for storing HHGs in the pre-GHC world, so it was probably just a way to distinguish the two when we had both for a while.

## Reviewer Notes

You can check that the event notification code is wired up correctly by running the handler test "Returns 204 when all validations pass" and looking at the log for this line:
```
INFO	event/notification.go:316	event.NotificationEventHandler: No notification needed to be created.
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7388) for this change


pkg/services/mto_shipment/shipment_deleter.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+	}
+
+	return &shipment, nil
+}
+
+func (f *shipmentDeleter) verifyShipmentCanBeDeleted(shipment *models.MTOShipment) error {
+	move := shipment.MoveTaskOrder
+	if move.Status != models.MoveStatusDRAFT && move.Status != models.MoveStatusNeedsServiceCounseling {
+		return services.NewForbiddenError("A shipment can only be deleted if the move is in Draft or NeedsServiceCounseling")
```
donaldthai, line 60: I think this should be a different error since we're checking the state of the move is correct before updating. So maybe `409 Conflict` error instead?

pkg/handlers/ghcapi/mto_shipment.go
```diff
@@ -134,3 +137,66 @@ func (h PatchShipmentHandler) Handle(params mtoshipmentops.PatchMTOShipmentStatu
...
 }
+
+// DeleteShipmentHandler soft deletes a shipment
+type DeleteShipmentHandler struct {
+	handlers.HandlerContext
+	services.ShipmentDeleter
+}
+
+// Handle soft deletes a shipment
+func (h DeleteShipmentHandler) Handle(params shipmentops.DeleteShipmentParams) middleware.Responder {
```
donaldthai, line 148: Noticed that we're not checking the eTag for this handler. Is that something we want to check for the delete handler?

pkg/services/mto_shipment/shipment_deleter.go
monfresh: My reasoning was that this would only happen if someone was trying to use the API on their own to try to circumvent what they can do on the website, hence the Forbidden error. If they're using the app normally, there should never be a way to delete a shipment that's not in Draft or in NeedsServiceCounseling.

Also, according to [this site](https://httpstatuses.com/409), 409 would not be the right code in this situation:
```
The request could not be completed due to a conflict with the current state of the target resource. 
This code is used in situations where the user might be able to resolve the conflict and resubmit the 
request.

The server SHOULD generate a payload that includes enough information for a user to recognize the 
source of the conflict.

Conflicts are most likely to occur in response to a PUT request. For example, if versioning were being 
used and the representation being PUT included changes to a resource that conflict with those made 
by an earlier (third-party) request, the origin server might use a 409 response to indicate that it can't 
complete the request. In this case, the response representation would likely contain information useful 
for merging the differences based on the revision history.
```

pkg/handlers/ghcapi/mto_shipment.go
monfresh: I don't think we need it here because we're not updating any fields that would result in any conflicts. I think eTag is meant for situations where you have concurrent requests updating the same object, and where you don't want one change to overwrite another. 

In this case, only one service counselor would be deleting a shipment. No one else would be trying to delete a shipment at the same time. And even if they did, the end result would be the same, except with a slightly different `deleted_at` timestamp. And once a move is in the counselor queue, the service member wouldn't be able to change the shipment either I think, so there shouldn't be any concurrent shipment updates at this point.

Does this sound right to you?

duncan-truss: I think it makes sense not to re-delete any shipments, if someone makes a request to delete a shipment that is already soft deleted they should receive a 404 not found instead of overwriting.  I think if you add a where clause to your findShipment to only find non-deleted shipments that would accomplish that.

monfresh: That's a great point! I pushed a fix for it.

pkg/services/mto_shipment/shipment_deleter.go
```diff
@@ -0,0 +1,64 @@
...
+			return nil, services.NewNotFoundError(shipmentID, "while looking for shipment")
+		}
+	}
+
+	return &shipment, nil
+}
+
+func (f *shipmentDeleter) verifyShipmentCanBeDeleted(shipment *models.MTOShipment) error {
+	move := shipment.MoveTaskOrder
+	if move.Status != models.MoveStatusDRAFT && move.Status != models.MoveStatusNeedsServiceCounseling {
```
duncan-truss, line 59: Good verification!

swagger/ghc.yaml
```diff
@@ -425,6 +401,31 @@ paths:
       description: Gets all shipments for a move task order
       operationId: listMTOShipments
       summary: Gets all shipments for a move task order
+  '/shipments/{shipmentID}':
+    delete:
+      summary: Soft deletes a shipment by ID
+      description: Soft deletes a shipment by ID
+      operationId: deleteShipment
+      tags:
+        - shipment
```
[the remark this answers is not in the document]
monfresh: Yes, I think the end goal is to rename `mtoShipment` to `shipment`, but I didn't want to rename everything in this PR. I wanted to introduce the idea that going forward, we want to use `shipment`. What do you think?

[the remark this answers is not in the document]
duncan-truss: I think it's probably okay to drop the mto, it's not like service items which some are on the MTO and some are on the Payment Request.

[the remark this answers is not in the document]
monfresh: Great! Is this good to be approved?

pkg/handlers/ghcapi/mto_shipment.go
```diff
@@ -134,3 +137,66 @@ func (h PatchShipmentHandler) Handle(params mtoshipmentops.PatchMTOShipmentStatu
...
+
+	if !session.IsOfficeUser() || !session.Roles.HasRole(roles.RoleTypeServicesCounselor) {
+		logger.Error("user is not authenticated with service counselor office role")
+		return shipmentops.NewDeleteShipmentForbidden()
+	}
+
+	shipmentID := uuid.FromStringOrNil(params.ShipmentID.String())
+	moveID, err := h.DeleteShipment(shipmentID)
+	if err != nil {
+		logger.Error("DeleteMTOShipment error: ", zap.Error(err))
```
duncan-truss, line 159: The error here will be in a separate part of the json if you were trying to include it in the message too.

pkg/handlers/ghcapi/mto_shipment.go
```diff
@@ -134,3 +137,66 @@ func (h PatchShipmentHandler) Handle(params mtoshipmentops.PatchMTOShipmentStatu
...
+	// will not change. However, we should make sure to add a test in
+	// mto_shipment_test.go that verifies the audit got saved.
+	h.triggerShipmentDeletionEvent(shipmentID, moveID, params)
+
+	return shipmentops.NewDeleteShipmentNoContent()
+}
+
+func (h DeleteShipmentHandler) triggerShipmentDeletionEvent(shipmentID uuid.UUID, moveID uuid.UUID, params shipmentops.DeleteShipmentParams) {
+	logger := h.LoggerFromRequest(params.HTTPRequest)
+	fmt.Println("moveID is:", moveID)
```
duncan-truss, line 185: Errant debugging statement

pkg/services/mto_shipment/shipment_deleter_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+			// Reset the deleted_at field to nil to allow the shipment to be
+			// deleted a second time when testing the other move status (a
+			// shipment can only be deleted once)
+			shipmentInDB.DeletedAt = nil
+			suite.MustSave(&shipment)
+		}
+	})
+
+	suite.T().Run("Returns not found error when the shipment is already deleted", func(t *testing.T) {
```
duncan-truss, line 84: 👏

duncan-truss (approved):
Nice work, I appreciate all the tricky tests you wrote.

pkg/handlers/ghcapi/mto_shipment.go
monfresh: Oops! Thanks for catching that.

pkg/handlers/ghcapi/mto_shipment.go
monfresh: I was following the logging format we use elsewhere in this file. I suppose we could remove the colon, but I'll leave it like the other ones for now.

pkg/services/mto_shipment/shipment_deleter.go
monfresh: Thanks!

PR #6660, opened 2021-05-19
[MB-8098] prevent GHC api handler from updating PR status to unapproved statuses

pearl-truss:
## Description

This PR fixes a bug where the GHC UpdatePaymentRequestStatusHandler had support for changing the payment request to _any_ status. This PR restricts the handler to only being able to update a payment request to REVIEWED or REVIEWED_AND_ALL_SERVICE_ITEMS_REJECTED

## Reviewer Notes

It wasn't specifically called out in the ticket, but this PR also restricts the handler from updating a PR to a pending status. I'm assuming that's desired as well

## Setup

I don't think theres a way to review this other than a review of the code and unit test

```sh
make server_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8098) for this change
* Relevant [slack thread](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1620857715412400)

### DIFF
--- a/pkg/handlers/ghcapi/payment_request.go
+++ b/pkg/handlers/ghcapi/payment_request.go
@@ -136,17 +136,14 @@ func (h UpdatePaymentRequestStatusHandler) Handle(params paymentrequestop.Update
 	now := time.Now()
 	existingPaymentRequest.Status = models.PaymentRequestStatus(params.Body.Status)
 
-	// Let's map the incoming status to our enumeration type
-	switch existingPaymentRequest.Status {
-	case models.PaymentRequestStatusReviewed, models.PaymentRequestStatusReviewedAllRejected:
-		existingPaymentRequest.ReviewedAt = &now
-	case models.PaymentRequestStatusSentToGex:
-		existingPaymentRequest.SentToGexAt = &now
-	case models.PaymentRequestStatusReceivedByGex:
-		existingPaymentRequest.ReceivedByGexAt = &now
-	case models.PaymentRequestStatusPaid:
-		existingPaymentRequest.PaidAt = &now
+	if existingPaymentRequest.Status != models.PaymentRequestStatusReviewed && existingPaymentRequest.Status != models.PaymentRequestStatusReviewedAllRejected {
+		payload := payloadForValidationError("Unable to complete request",
+			fmt.Sprintf("Incoming payment request status should be REVIEWED or REVIEWEDALLREJECTED instead it was: %s", existingPaymentRequest.Status.String()),
+			h.GetTraceID(), validate.NewErrors())
+		return paymentrequestop.NewUpdatePaymentRequestStatusUnprocessableEntity().WithPayload(payload)
 	}
+
+	existingPaymentRequest.ReviewedAt = &now
 
 	// If we got a rejection reason let's use it
 	if params.Body.RejectionReason != nil {
--- a/pkg/handlers/ghcapi/payment_request.go
+++ b/pkg/handlers/ghcapi/payment_request.go
@@ -138,7 +138,7 @@ func (h UpdatePaymentRequestStatusHandler) Handle(params paymentrequestop.Update
 
 	if existingPaymentRequest.Status != models.PaymentRequestStatusReviewed && existingPaymentRequest.Status != models.PaymentRequestStatusReviewedAllRejected {
 		payload := payloadForValidationError("Unable to complete request",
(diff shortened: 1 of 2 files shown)


pkg/handlers/ghcapi/payment_request.go
```diff
@@ -136,18 +136,15 @@ func (h UpdatePaymentRequestStatusHandler) Handle(params paymentrequestop.Update
...
-	case models.PaymentRequestStatusPaid:
-		existingPaymentRequest.PaidAt = &now
+	if existingPaymentRequest.Status != models.PaymentRequestStatusReviewed && existingPaymentRequest.Status != models.PaymentRequestStatusReviewedAllRejected {
+		payload := payloadForValidationError("Unable to complete request",
+			fmt.Sprintf("Incoming payment request status should be REVIEWED or REVIEWEDALLREJECTED instead it was: %s", existingPaymentRequest.Status.String()),
+			h.GetTraceID(), validate.NewErrors())
+		return paymentrequestop.NewUpdatePaymentRequestStatusUnprocessableEntity().WithPayload(payload)
 	}
 
+	existingPaymentRequest.ReviewedAt = &now
```
monfresh, line 146: What do you think about moving this logic (lines 139-146) into the service? We want to minimize the amount of business logic in handlers.

pkg/handlers/ghcapi/payment_request_test.go
```diff
@@ -301,6 +301,34 @@ func (suite *HandlerSuite) TestUpdatePaymentRequestStatusHandler() {
 		suite.NotNil(payload.ReviewedAt)
 	})
 
+	suite.T().Run("prevent handler from updating payment request status to unapproved statuses", func(t *testing.T) {
+		nonApprovedPRStatuses := [5]ghcmessages.PaymentRequestStatus{"SENT_TO_GEX", "RECEIVED_BY_GEX", "PAID", "EDI_ERROR", "PENDING"}
+
+		for _, nonApprovedPRStatus := range nonApprovedPRStatuses {
+			pendingPaymentRequest := testdatagen.MakeDefaultPaymentRequest(suite.DB())
```
monfresh, line 308: Since we are mocking the fetcher, we can use `MakeStubbedPaymentRequest(suite.DB())` so that we don't hit the DB.

pkg/handlers/ghcapi/payment_request.go
pearl-truss: I think this logic is specific to this ghc handler, I don't think we want to apply it more generally in the service

monfresh: Ah, then I think this is an indication that the service is too broad. Handlers and Services should correspond to a business concept, as opposed to the DB action being performed. Yes, we are updating the payment request status, but in what context? By naming a handler/service `UpdatePaymentRequestStatus`, it makes it sound like any status can be set, which is not the case. 

In the GHC API, this endpoint is only used in the context of a TIO reviewing a payment request, and the only statuses that can be set are REVIEWED or REVIEWED_AND_ALL_SERVICE_ITEMS_REJECTED. So, a better name for this service might be `PaymentRequestReviewer`, or perhaps `PaymentRequestRouter`, similar to our `MoveRouter`, which would have a function called `ReviewPaymentRequest` which would only allow those 2 statuses to be set.

I see that the support API is also using the `PaymentRequestStatusUpdater` service, and it looks like it's also not using it correctly because it's allowing various statuses to be set. But for now, we'll just deal with the GHC API, so I propose we create a new service for the GHC API. I would also rename the GHC handler to something like `ReviewPaymentRequestHandler`, so that it's clear what it's doing.

I'd be happy to pair with you on this so we can refactor it.

pearl-truss: @monfresh I'm not sure if this refactor is in scope for this ticket. I'll bring it up at A-team's standup

pearl-truss: @monfresh thanks for raising this btw. At standup we decided to create a [new story](https://dp3.atlassian.net/browse/MB-8170) to look into this some more since it's towards the end of a sprint. The ticket I created is for someone to look into changes to both the GHC and Support api

pearl-truss: Oh I missed your comment from above. The ticket is already created

monfresh: Awesome! That story looks wonderful. 👩🏼‍🍳

gidjin (approved):
LGTM 🚀

PR #6661, opened 2021-05-19
MB-7653 Services counselor sees visual cues on allowances view

donaldthai:
## Description

> As a services counseling user
> When there are special allowances fields in the allowances table that need my attention
> Then I see a visual indicator on following fields to catch my attention: Pro-gear, Spouse pro-gear, RME, OCIE

## Reviewer Notes

- I used an outline styling to produce the visual cue. I chose this route since outlines don't change the position of the elements on the page.

## Setup

```sh
make server_run
make office_client_run
```

Test
1. Login as services counselor on office app
2. Click on any move
3. Scroll down to allowances section
4. Note the visual cues

Storybook
```sh
make storybook
```

1. Navigate to `http://localhost:6006/?path=/story/office-components-allowanceslist--visual-cues`
2. Note the component with visual cues

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7653) for this change

## Screenshots

Screen Shot 2021-05-19 at 12 41 31 PM

### DIFF
--- a/src/components/Office/DefinitionLists/AllowancesList.jsx
+++ b/src/components/Office/DefinitionLists/AllowancesList.jsx
@@ -35,6 +35,7 @@ const AllowancesList = ({ info, showVisualCues }) => {
           <dd data-testid="dependents">{info.dependents ? 'Authorized' : 'Unauthorized'}</dd>
         </div>
         <div
+          data-testid="progear-row"
           className={classNames(descriptionListStyles.row, {
             [`${descriptionListStyles.rowWithVisualCue}`]: showVisualCues,
           })}
@@ -43,6 +44,7 @@ const AllowancesList = ({ info, showVisualCues }) => {
           <dd data-testid="progear">{formatWeight(info.progear)}</dd>
         </div>
         <div
+          data-testid="spouseprogear-row"
           className={classNames(descriptionListStyles.row, {
             [`${descriptionListStyles.rowWithVisualCue}`]: showVisualCues,
           })}
@@ -51,6 +53,7 @@ const AllowancesList = ({ info, showVisualCues }) => {
           <dd data-testid="spouseProgear">{formatWeight(info.spouseProgear)}</dd>
         </div>
         <div
+          data-testid="rme-row"
           className={classNames(descriptionListStyles.row, {
             [`${descriptionListStyles.rowWithVisualCue}`]: showVisualCues,
           })}
@@ -59,6 +62,7 @@ const AllowancesList = ({ info, showVisualCues }) => {
           <dd data-testid="rme">{formatWeight(info.requiredMedicalEquipmentWeight)}</dd>
         </div>
         <div
+          data-testid="ocie-row"
           className={classNames(descriptionListStyles.row, {
             [`${descriptionListStyles.rowWithVisualCue}`]: showVisualCues,
           })}
--- a/src/components/Office/DefinitionLists/AllowancesList.jsx
+++ b/src/components/Office/DefinitionLists/AllowancesList.jsx
@@ -9,6 +9,10 @@ import { formatWeight, formatDaysInTransit } from 'shared/formatters';
 import friendlyBranchRank from 'utils/branchRankFormatters';
 
 const AllowancesList = ({ info, showVisualCues }) => {
+  const visualCuesStyle = classNames(descriptionListStyles.row, {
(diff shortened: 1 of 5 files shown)


src/components/Office/DefinitionLists/AllowancesList.test.jsx
```diff
@@ -73,4 +73,12 @@ describe('AllowancesList', () => {
     render(<AllowancesList info={withUnauthorizedOcie} />);
     expect(screen.getByTestId('ocie').textContent).toEqual('Unauthorized');
   });
+
+  it('renders visual cues classname', () => {
+    render(<AllowancesList info={info} showVisualCues />);
+    expect(screen.getByTestId('progear-row').className).toContain('rowWithVisualCue');
+    expect(screen.getByTestId('spouseprogear-row').className).toContain('rowWithVisualCue');
+    expect(screen.getByTestId('rme-row').className).toContain('rowWithVisualCue');
+    expect(screen.getByTestId('ocie-row').className).toContain('rowWithVisualCue');
```
felipe-lee, line 82: I think it'd be good to try not using test ID for queries, especially based on https://testing-library.com/docs/queries/about/#priority and https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-the-wrong-query

But because of a problem with the way react testing library looks for roles (well really a bug in the package RTL uses), we can't easily grab `<dt>` or `<dd>` elements. I know in this case you're looking at the div surrounding them but I would say you could grab `dt` or `dd`, then look at it's parent to verify it has the expected class. You could still do it even with the bug, but you'd need to use `screen.getByText` and use either the `dt` or `dd` text.

All that to say, I think in general we should move away from using test ids. In this case because of the bug, I think it's more of a gray area.

src/styles/descriptionList.module.scss
```diff
@@ -14,12 +14,25 @@
...
+    content: "";
+  }
+
   dt,
   dd {
     @include u-border-top('1px');
     @include u-border-top('base-lighter');
 
-    @include u-padding-x(1);
+    @include u-padding-x(2);
```
felipe-lee, line 35: I just want to note that since this is a stylesheet used in both customer and office components, it's changing things on both apps. But if design is ok with that, then ignore me.

felipe-lee (commented):
Overall LGTM. Just two comments, one up to you, the other hinges on what design thinks.

src/components/Office/DefinitionLists/AllowancesList.jsx
```diff
@@ -33,19 +34,39 @@ const AllowancesList = ({ info }) => {
           <dt>Dependents</dt>
           <dd data-testid="dependents">{info.dependents ? 'Authorized' : 'Unauthorized'}</dd>
         </div>
-        <div className={descriptionListStyles.row}>
+        <div
+          data-testid="progear-row"
+          className={classNames(descriptionListStyles.row, {
```
duncan-truss, line 39: Because this is only calculated once for all of the rows on the showVisualCues prop should we just call classNames once?  Will this go away once the move has completed services counseling?

src/components/Office/DefinitionLists/AllowancesList.stories.jsx
```diff
@@ -6,6 +6,11 @@ import AllowancesList from './AllowancesList';
 export default {
   title: 'Office Components/AllowancesList',
   component: AllowancesList,
+  argTypes: {
```
duncan-truss, line 9: 👍

src/components/Office/DefinitionLists/AllowancesList.jsx
donaldthai: Yea, let me put that classname calls in a variable. Not sure about the visual cues disappearing after submitting a move. I think there's another story that determines the behavior of how the visual cues work though.

monfresh:
Should the visual indicators remain visible even after the counselor submits the move?

src/components/Office/DefinitionLists/AllowancesList.jsx
donaldthai: There is a note in the jira story:
```
For now we’ll always show the indicator for those fields – later we’ll take another pass at this in order to see how we want to handle when the indicator should go away.
```

brooksr4 (approved):
Designs look good

felipe-lee (approved):
LGTM!

duncan-truss (approved):

PR #6662, opened 2021-05-19
MB-8022 MTOShipments - Improve null date display

shimonacarvalho:
## Description

Sets the dates in mtoshipment to be `x-nullable: true` and `x-omitempty: false`.
Basically this means yes, they can be null when the server returns the payload, and no, don't omit them from the payload if so.
So what you want to see is, dates that are null in the db, should show up as follows:
``` 
"scheduledPickupDate": null`
```

## Setup

Delete all the date values in a shipment using a db tool like dBeaver. 
Then use `getMTO` endpoint to view the associated MTO. All dates should be listed but set to null in the shipment.
```
actualPickupDate
approvedDate
firstAvailableDeliveryDate
primeEstimatedWeightRecordedDate
requiredDeliveryDate
scheduledPickupDate
requestedPickupDate
```

Also check that you can update the following:
```
actualPickupDate
firstAvailableDeliveryDate
scheduledPickupDate
requestedPickupDate
```

And that you get an error on updating these dates (cannot be updated)
```
approvedDate
primeEstimatedWeightRecordedDate
requiredDeliveryDate
```

### DIFF
--- a/cmd/prime-api-dev/scripts/payment_requests.go
+++ b/cmd/prime-api-dev/scripts/payment_requests.go
@@ -609,7 +609,7 @@ func (pr *paymentRequestsData) displayUpdateShipmentMenu() (bool, menuType, erro
 			if err != nil {
 				log.Fatal("Cannot get date input", err)
 			}
-			shipment.RequestedPickupDate = strFmtDate
+			shipment.RequestedPickupDate = &strFmtDate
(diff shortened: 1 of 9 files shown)


eamahanna (approved):
This looks good! Should there be follow up tickets created to have all nullable fields return `null`?

PR #6665, opened 2021-05-20
MB-8074 fix duplicate 997 processing

gidjin:
## Description

If the `ProcessEDIDeleteFiles` flag is false files are not removed from the SFTP server, which means we reprocess the same file and generate a duplicate key error. This PR changes the code so that we look for a duplicate before trying to save it and it outputs the state of the flag when the job runner starts and during the reader if it is off.

## Reviewer Notes

Is this the right approach to a duplicate file? ICNs are unique to each file sent so an EDI997 with the same ICN and same payment request number is a duplicate. There is nothing to do more on our end, not sure if this should be reported as an error or ignored.

I can easily just change this so that it's logging the flag.

## Setup

I don't think there is a way to test the code change locally since we don't have an sftp server. You can check and run the tests `make server_test`.

Then check the output of the job runner for the new row.

```sh
PROCESS_EDI_DELETE_FILES=false go run ./cmd/milmove-tasks process-edis
...
{"level":"info","ts":"2021-05-25T06:43:51.529Z","caller":"milmove-tasks/process_edis.go:180","msg":"ProcessEDIDeleteFiles is false"}
...

PROCESS_EDI_DELETE_FILES=true go run ./cmd/milmove-tasks process-edis
...
{"level":"info","ts":"2021-05-25T06:43:51.529Z","caller":"milmove-tasks/process_edis.go:180","msg":"ProcessEDIDeleteFiles is true"}
...
```

Acceptance is verification in staging that the above appears working.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8074) for this change


Ryan-Koch (approved):
I don't have the most context around this part of the app, but the logic I see here makes sense as far as checking for a duplicate goes and it seems to run as expected based on the test results.

PR #6667, opened 2021-05-20
[MB-7381] Services counselor can edit move contact info

hannaliebl:
## Description

* Add `<CustomerInfo />` component that wraps the existing `<CustomerContactInfoForm />` form fields and allows a service counselor to update move contact info
* Update the GHC API to return suffix and middle name info from customer GET
* Update the GHC API to allow editing the suffix and middle name of a customer
* Make the backup fields required to match the requirements
* Display the customer name according to the new requirement that includes the suffix and middle name and add a utility function to do that

## Reviewer Notes

* I used the default grid container width, but it may need to be less wide than that - is there a class or container that matches the mocks?
* ~~I'm having issues with the test I added locally (`"messageParent" can only be used inside a worker`), so I'm curious if it will happen on CI. Is there anything wrong with the way I've set up the CustomerInfo test? I will add more assertions about the rest of the form fields being populated once I can get the first one working.~~ Edit: I think I figured this out, I will add some more tests soon.

## Setup

```sh
make server_run
make client_run
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7381) for this change

## Screenshots

https://user-images.githubusercontent.com/3903208/119006547-7bcc8f00-b95e-11eb-8e61-c5eb9a00cc33.mov


monfresh:
The order of the name parts in the Customer Info section no longer matches the order in the header. Before, they were consistent. Should we keep them consistent going forward? Is there a separate story to fix the header?

Also, the order in the Customer Info section seems unusual to me. Typically, if there's a comma, it's when the Last Name is first. If my name is `Laura W. Henderson, Jr.`, it would be strange to see my name formatted like it is now: 
```
Laura Jr., Henderson W.
```

Are we using a specific style guide for this? For example, using the [APA](https://blog.apastyle.org/apastyle/2012/03/jr-sr-and-other-suffixes-in-apa-style.html), we would format the name like this:
```
Henderson, Laura W., Jr.
```

src/pages/Office/CustomerInfo/CustomerInfo.test.jsx
```diff
@@ -0,0 +1,38 @@
...
+describe('CustomerInfo', () => {
+  it('populates initial field values', () => {
+    render(
+      <MockProviders initialEntries={['moves/CDG3TR/customer-info']}>
+        <CustomerInfo customer={mockCustomer} isLoading={false} isError={false} />{' '}
+      </MockProviders>,
+    );
+    expect(screen.getByLabelText('First name').value).toEqual('John');
+  });
+});
```
monfresh, line 38: Can we add some more tests for the various combinations (some with suffix, some without, some with middle name, some without), and confirm the expected result?

monfresh: Or add a unit test for `src/utils/formattedCustomerName.js`

src/pages/Office/ServicesCounselingMoveDetails/ServicesCounselingMoveDetails.jsx
```diff
@@ -63,7 +64,7 @@ const ServicesCounselingMoveDetails = () => {
   }
 
   const customerInfo = {
-    name: `${customer.last_name}, ${customer.first_name}`,
+    name: formattedCustomerName(customer.first_name, customer.last_name, customer.suffix, customer.middle_name),
```
monfresh, line 67: I see the bug now. The `formattedCustomerName` function takes the last name as the first argument, but you are passing in the first name. Does JavaScript have the equivalent of [required keyword arguments in Ruby](https://thoughtbot.com/blog/ruby-2-keyword-arguments#keyword-arguments-vs-positional-arguments)? That is, instead of just passing in the argument, you have to pass in its description as well, and when you use named keywords, the order doesn't matter. For example:
```
formattedCustomerName(first: customer.first_name, last: customer.last_name)
```

src/utils/formattedCustomerName.js
```diff
@@ -0,0 +1,12 @@
...
+    return `${last} ${suffix}, ${first} ${middle}`;
+  }
+  if (suffix.length > 0 && middle.length === 0) {
+    return `${last} ${suffix}, ${first}`;
+  }
+  if (suffix.length === 0 && middle.length > 0) {
+    return `${last}, ${first} ${middle}`;
+  }
+  return `${last}, ${first}`;
+}
```
monfresh, line 12: If we were to use the APA style guide, I think this can be simplified to something like this:
```js
const middleSuffixString = [middle, suffix].filter(Boolean).join(', ');
const lastFirstString = `${last}, ${first} `;
return lastFirstString + middleSuffixString;
```

src/pages/Office/CustomerInfo/CustomerInfo.test.jsx
hannaliebl: Yes I am working on adding those

src/pages/Office/ServicesCounselingMoveDetails/ServicesCounselingMoveDetails.jsx
hannaliebl: Yes, this is a bug, I've corrected it, thanks!

src/utils/formattedCustomerName.js
hannaliebl: Yeah that does seem better. The ticket was wrong and it should be `[last_name], [first_name] [middle_name], [suffix]`

hannaliebl: I stuck with the explicitness because the possibility of "Last, First, Suffix" and "Last, First Middle" doesn't work with the above, I don't think. The formatting is based on the google doc in the ticket (it's an open link, so won't link it here.)

pkg/services/office_user/customer/customer_updater.go
```diff
@@ -80,6 +80,14 @@ func (s *customerUpdater) UpdateCustomer(eTag string, customer models.ServiceMem
 			existingCustomer.Telephone = customer.Telephone
 		}
 
+		if customer.Suffix != nil {
+			existingCustomer.Suffix = customer.Suffix
+		}
+
+		if customer.MiddleName != nil {
+			existingCustomer.MiddleName = customer.MiddleName
+		}
```
duncan-truss, line 89: In the case of needing to remove the middle name or suffix this would still allow us to set the blank string here?  It may not really matter but if the name or suffix was not nil but had a length zero should we set it to nil?

src/pages/Office/CustomerInfo/CustomerInfo.jsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const [mutateCustomerInfo] = useMutation(updateCustomerInfo, {
+    onSuccess: (data, variables) => {
+      const updatedCustomer = data.customer[variables.customerId];
+      queryCache.setQueryData([CUSTOMER, variables.customerId], {
+        customer: {
+          [`${variables.customerId}`]: updatedCustomer,
+        },
+      });
+      queryCache.invalidateQueries(CUSTOMER);
```
duncan-truss, line 32: This would invalidate all customer queries because they all share the customer key, we can target only the query for this customer with `queryCache.invalidateQueries([CUSTOMER, variables.customerId])`.  That is the query that the header uses but the move details page actually returns the customer as part of the orders query.  If you don't have the order id handy to invalidate that one too, could you leave a TODO reminder?

src/utils/formattedCustomerName.js
monfresh: How about this? All tests are passing:
```js
const lastFirst = [`${last}, ${first}`];
const lastFirstMiddle = [lastFirst, middle].filter(Boolean).join(' ');

return suffix.length > 0 ? `${lastFirstMiddle}, ${suffix}` : lastFirstMiddle;
```

brooksr4 (approved):
Looks good

src/pages/Office/ServicesCounselingMoveInfo/ServicesCounselingMoveInfo.jsx
```diff
@@ -36,6 +37,10 @@ const ServicesCounselingMoveInfo = () => {
             <ServicesCounselingMoveDocumentWrapper />
           </Route>
 
+          <Route path="/counseling/moves/:moveCode/customer-info" exact>
```
donaldthai, line 40: According to the [URL sheet](https://docs.google.com/spreadsheets/d/1VWtagZWTcmgqfWaLdQlUHKBJJVtqBXlEJ_KuMMOSZMs/edit#gid=1627844212&range=E7), the path should be something like `/counseling/moves/:moveCode/customer`

src/pages/Office/ServicesCounselingMoveDetails/ServicesCounselingMoveDetails.jsx
```diff
@@ -198,7 +199,11 @@ const ServicesCounselingMoveDetails = () => {
               title="Customer info"
               editButton={
                 counselorCanEdit && (
-                  <Link className="usa-button usa-button--secondary" to="#">
+                  <Link
+                    className="usa-button usa-button--secondary"
+                    data-testid="edit=customer-info"
```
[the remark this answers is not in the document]
hannaliebl: Ah, nope, thanks!

pkg/services/office_user/customer/customer_updater.go
hannaliebl: Yeah, that seems wise, thanks!

hannaliebl: I did this [here](https://github.com/transcom/mymove/pull/6667/commits/b372c8a52954e983b1ec01919cd419a2433d515d), does this seem ok?

src/pages/Office/CustomerInfo/CustomerInfo.jsx
hannaliebl: I invalidated the customer id specifically in https://github.com/transcom/mymove/pull/6667/commits/2d6ac52db1af645857ba5345d110f1f0d3e8b9b4

For the order id, I could pass that in via props from `order.id` here, right? https://github.com/transcom/mymove/blob/master/src/pages/Office/ServicesCounselingMoveInfo/ServicesCounselingMoveInfo.jsx#L19 Is that the correct value to invalidate?

duncan-truss: Yea that would be the one to add, but we also don't have caching turned on yet that's why it gets updated after submitting the form.

pkg/services/office_user/customer/customer_updater.go
duncan-truss: Yea that's what I had in mind, we could call whatever Go's version of trim is too if they entered only white space.  What you have seems good.

src/pages/Office/CustomerInfo/CustomerInfo.test.jsx
```diff
@@ -0,0 +1,56 @@
...
+    expect(screen.getByDisplayValue(mockCustomer.phone).value).toEqual(mockCustomer.phone);
+    expect(screen.getByDisplayValue(mockCustomer.backup_contact.phone).value).toEqual(
+      mockCustomer.backup_contact.phone,
+    );
+    // to get around the two inputs labeled "Email" on the screen
+    expect(screen.getByDisplayValue(mockCustomer.email).value).toEqual(mockCustomer.email);
+    expect(screen.getByDisplayValue(mockCustomer.backup_contact.email).value).toEqual(
+      mockCustomer.backup_contact.email,
+    );
+    expect(screen.getByLabelText('Address 1').value).toEqual(mockCustomer.current_address.street_address_1);
```
duncan-truss, line 50: To make this match what's on the page I think you need `street_address_2` as well

duncan-truss: I guess we aren't displaying the street address 2 value at all now that I look.  It seems like we should incorporate it in the current address that gets displayed but yea it's not in the mock.

duncan-truss (approved):
Overall looks really solid well done.  I think the page needs a gray background color and I made a comment on our display of the street address 2 value but those are minor.

PR #6672, opened 2021-05-20
[MB-7596] Return error on nil OriginDutyStation

mboren:
## Description
When a move with a nil OriginDutyStation in its orders went through routing, we were assuming that the move shouldn't go through services counseling. 
I think we should return an error instead, because it doesn't really make sense to route a move without an OriginDutyStation; its got no GBLOC, so it's not going to show up in the TOO queue.

You can't create a move without an OriginDutyStation from the UI, so this case should only occur if there's a bug somewhere else.

If a user does hit this case, it won't let them submit their move and they won't see a useful error message, which isn't great,
but I think it's better than being able to submit their move into the void?

## Setup
Go through the signup flow for a new service member, and stop when you get to the page where you enter your signature and submit the move.

Get the order id for the move (this query gets the most recently updated)
```sql
select id from orders order by updated_at desc limit 1;
```
Set the OriginDutyStation to NULL (replace the id with the result from the last query):
```sql
UPDATE orders
SET origin_duty_station_id = NULL
WHERE id = 'a833e56d-1a75-497c-95fb-aeae8f5c763c';
```

Now attempt to submit the move. It should fail to submit, with an error message that looks like this:
image

## Acceptance
You can't trigger this error with the UI or API, so I think all you can do is make sure I didn't break routing by creating a move that requires services counseling, and making sure it shows up in the services counselor queue.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7596) for this change

### DIFF
--- a/pkg/services/move/move_router.go
+++ b/pkg/services/move/move_router.go
@@ -86,15 +86,16 @@ func (router moveRouter) needsServiceCounseling(move *models.Move) (bool, error)
 
 	var originDutyStation models.DutyStation
 
-	if orders.OriginDutyStationID != nil && *orders.OriginDutyStationID != uuid.Nil {
-		originDutyStation, err = models.FetchDutyStation(router.db, *orders.OriginDutyStationID)
-		if err != nil {
-			router.logger.Error("failure finding the origin duty station", zap.Error(err))
-			return false, services.NewInvalidInputError(*orders.OriginDutyStationID, err, nil, "unable to find origin duty station")
-		}
-		return originDutyStation.ProvidesServicesCounseling, nil
+	if orders.OriginDutyStationID == nil || *orders.OriginDutyStationID != uuid.Nil {
+		return false, services.NewInvalidInputError(orders.ID, err, nil, "orders missing OriginDutyStation")
 	}
-	return false, nil
+
+	originDutyStation, err = models.FetchDutyStation(router.db, *orders.OriginDutyStationID)
+	if err != nil {
+		router.logger.Error("failure finding the origin duty station", zap.Error(err))
+		return false, services.NewInvalidInputError(*orders.OriginDutyStationID, err, nil, "unable to find origin duty station")
+	}
+	return originDutyStation.ProvidesServicesCounseling, nil
 }
 
 // sendToServiceCounselor makes the move available for a Service Counselor to review
--- a/pkg/services/move/move_router.go
+++ b/pkg/services/move/move_router.go
@@ -86,7 +86,7 @@ func (router moveRouter) needsServiceCounseling(move *models.Move) (bool, error)
 
 	var originDutyStation models.DutyStation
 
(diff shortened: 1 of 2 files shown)


pkg/services/move/move_router_test.go
```diff
@@ -62,11 +62,13 @@ func (suite *MoveServiceSuite) TestSubmitted() {
 		suite.Contains(err.Error(), "not found looking for move.OrdersID")
 	})
 
-	suite.Run("returns error when needsServicesCounseling cannot find move", func() {
+	suite.Run("returns error when OriginDutyStation is missing", func() {
 		move := testdatagen.MakeDefaultMove(suite.DB())
-		move.Orders.OriginDutyStation = nil
-		move.Orders.OriginDutyStationID = nil
-		suite.NoError(suite.DB().Update(&move))
+		order := move.Orders
```
pearl-truss, line 67: What was the issue with the first way of using move.Orders? Asking just for my own curiosity!

pkg/services/move/move_router.go
```diff
@@ -86,15 +86,16 @@ func (router moveRouter) needsServiceCounseling(move *models.Move) (bool, error)
...
 
-	if orders.OriginDutyStationID != nil && *orders.OriginDutyStationID != uuid.Nil {
-		originDutyStation, err = models.FetchDutyStation(router.db, *orders.OriginDutyStationID)
-		if err != nil {
-			router.logger.Error("failure finding the origin duty station", zap.Error(err))
-			return false, services.NewInvalidInputError(*orders.OriginDutyStationID, err, nil, "unable to find origin duty station")
-		}
-		return originDutyStation.ProvidesServicesCounseling, nil
+	if orders.OriginDutyStationID == nil || *orders.OriginDutyStationID == uuid.Nil {
+		return false, services.NewInvalidInputError(orders.ID, err, nil, "orders missing OriginDutyStation")
```
pearl-truss, line 90: Probably outside of this PR but I'm curious why is the client showing a 500 if this case triggers a 422 which is more specific?

pearl-truss (approved):
I got back the same error that's expected from the PR description

pkg/services/move/move_router_test.go
mboren: My first attempt here didn't save the change to the orders to the DB. The test was only passing because my code was also broken, so it would return an error for the wrong reason every time.
I don't feel super confident with how DB models are saved during tests so I tried a bunch of different permutations and this is what worked.

pkg/services/move/move_router.go
mboren: That's a great question. I just checked and it looks like the error gets passed to `ResponseForError` in `errors.go`, which looks like it handles errors a bit differently than I expected. It looks like there might be a few other places in `move_router.go` where a 500 is probably returned when we don't expect it to be.

PR #6675, opened 2021-05-21
MB-8054 initial back-end tracing implementation

nelztruss:
## Description

This PR represents approximately the most minimal bits of code needed to emit the most basic tracing information, at most one trace for each request that is received by the server.

To note, there are really only ~4 lines of code that are doing instrumentation. Adding more instrumentation over time will make the tracing even more valuable and useful, but there's already a lot of bang-for-buck with just these few lines. (A near-term enhancement I am planning is to add some instrumentation around every database interaction.)

The `configureTracing(...)` function is the most sophisticated thing in this PR:
* I have specifically targeted ONLY local development environments for now, but if we want to enable tracing in deployed environments in the future, the design flexibility of OpenTelemetry means that this configuration is the only thing that would need to be enhanced a bit to support those environments.
* It defaults to a `noop` implementation of tracing, so that tracing is only enabled for those that opt in to it. 
* The `STDOUT` option will do the tracing, but only emit it to your console logs. 
* The `LOCAL_COLLECTOR` option will end up showing errors in the log (but won't actually break any functionality) unless there is some other environmental setup done (e.g. configuring and running another docker container). In the pursuit of a small iterative PR, I will be adding the tools to do that setup in a later PR.

## Setup

If you'd like to see the trace event bodies in the console, you just add an environment variable when running the server:

```sh
$ OTEL_CONFIG=STDOUT make server_run
```

## References

* [OTel - Getting Started](https://opentelemetry.io/docs/go/getting-started/) explains parts of the configuration

## Screenshots

After doing the configuration, this is a partial view of what a trace looks like in Honeycomb:

image


cmd/milmove/tracing.go
```diff
@@ -0,0 +1,62 @@
...
+				logger.Error("shutdown problems with tracing exporter", zap.Error(err))
+			}
+		}
+	default:
+		logger.Info("opentelemetry not enabled")
+	}
+
+	// set the global configuration
+	otel.SetTracerProvider(tp)
+	otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}))
```
nelztruss, line 59: TBH, I am not a huge fan of relying on global state, but this is the current best practice with OTel.

ahobson:
Closing in favor of https://github.com/transcom/mymove/pull/7304

PR #6686, opened 2021-05-25
MB-7956 Update create item and crate with DCRT service item

eamahanna:
## Description

This PR creates a new tesdatagen func `MakeMTOServiceItemDomesticCrating`.  This function creates a DCRT service item and its associated item and crate. I updated devseed, e2e populate and another file to use this function when generating DCRT service items. 

## Reviewer Notes

I saw a DCRT service item was created in `bandwidth.go`. I'm not sure what the file is used for, but it also creates a DCRT service item, so I updated it to use the new function as well. Is that okay?

## Setup

To verify complete the following steps:

- Execute `make db_dev_reset`
- Execute `make db_dev_e2e_populate`
- Execute `make server_run`
- Check with `fetchMTOUpdates` using postman and verify that all crating service items (DCRT) have an ITEM and CRATE. You can search for `DCRT` in the response to quickly find the crating service items.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7956) for this change

### DIFF
--- a/pkg/testdatagen/make_mto_service_item.go
+++ b/pkg/testdatagen/make_mto_service_item.go
@@ -288,7 +288,7 @@ func MakeMTOServiceItemDomesticCrating(db *pop.Connection, assertions Assertions
 
 	mtoServiceItem := MakeMTOServiceItem(db, assertions)
 
-	// Create crating item
+	// Create item
 	MakeMTOServiceItemDimension(db, Assertions{
 		MTOServiceItem: mtoServiceItem,
 		MTOServiceItemDimension: models.MTOServiceItemDimension{
--- a/pkg/testdatagen/make_mto_service_item.go
+++ b/pkg/testdatagen/make_mto_service_item.go
@@ -274,3 +274,40 @@ func MakeRealMTOServiceItemWithAllDeps(db *pop.Connection, serviceCode models.Re
 	log.Panicf("couldn't create service item service code %s not defined", serviceCode)
 	return models.MTOServiceItem{}
 }
+
+// MakeMTOServiceItemDomesticCrating makes a domestic crating service item and its associated item and crate
+func MakeMTOServiceItemDomesticCrating(db *pop.Connection, assertions Assertions) models.MTOServiceItem {
+	// Set service item type
+	assertions.ReService = models.ReService{
+		ID: uuid.FromStringOrNil("68417bd7-4a9d-4472-941e-2ba6aeaf15f4"), // DCRT - Domestic Crating
+	}
+
+	// Add description
+	description := "Decorated horse head to be crated."
+	assertions.MTOServiceItem.Description = &description
+
+	mtoServiceItem := MakeMTOServiceItem(db, assertions)
+
+	// Create crating item
+	MakeMTOServiceItemDimension(db, Assertions{
+		MTOServiceItem: mtoServiceItem,
+		MTOServiceItemDimension: models.MTOServiceItemDimension{
+			Length: 10000,
+			Height: 5000,
+			Width:  2500,
+		},
+	})
+
+	// Creat crate
+	MakeMTOServiceItemDimension(db, Assertions{
+		MTOServiceItem: mtoServiceItem,
+		MTOServiceItemDimension: models.MTOServiceItemDimension{
+			Type:   models.DimensionTypeCrate,
+			Length: 11000,
+			Height: 6000,
+			Width:  3500,
+		},
+	})
+
+	return mtoServiceItem
+}
--- a/pkg/testdatagen/make_mto_service_item.go
+++ b/pkg/testdatagen/make_mto_service_item.go
@@ -282,30 +282,18 @@ func MakeMTOServiceItemDomesticCrating(db *pop.Connection, assertions Assertions
 		ID: uuid.FromStringOrNil("68417bd7-4a9d-4472-941e-2ba6aeaf15f4"), // DCRT - Domestic Crating
 	}
 
-	// Add description
-	description := "Decorated horse head to be crated."
-	assertions.MTOServiceItem.Description = &description
-
 	mtoServiceItem := MakeMTOServiceItem(db, assertions)
 
 	// Create item
 	MakeMTOServiceItemDimension(db, Assertions{
 		MTOServiceItem: mtoServiceItem,
-		MTOServiceItemDimension: models.MTOServiceItemDimension{
-			Length: 10000,
-			Height: 5000,
-			Width:  2500,
-		},
 	})
 
 	// Creat crate
 	MakeMTOServiceItemDimension(db, Assertions{
 		MTOServiceItem: mtoServiceItem,
 		MTOServiceItemDimension: models.MTOServiceItemDimension{
-			Type:   models.DimensionTypeCrate,
-			Length: 11000,
-			Height: 6000,
-			Width:  3500,
+			Type: models.DimensionTypeCrate,
 		},
 	})
 
--- a/pkg/testdatagen/make_mto_service_item.go
+++ b/pkg/testdatagen/make_mto_service_item.go
@@ -287,17 +287,19 @@ func MakeMTOServiceItemDomesticCrating(db *pop.Connection, assertions Assertions
 	})
 
 	// Create item
-	MakeMTOServiceItemDimension(db, Assertions{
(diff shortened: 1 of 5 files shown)


pkg/testdatagen/make_mto_service_item.go
```diff
@@ -274,3 +274,40 @@ func MakeRealMTOServiceItemWithAllDeps(db *pop.Connection, serviceCode models.Re
 	log.Panicf("couldn't create service item service code %s not defined", serviceCode)
 	return models.MTOServiceItem{}
 }
+
+// MakeMTOServiceItemDomesticCrating makes a domestic crating service item and its associated item and crate
+func MakeMTOServiceItemDomesticCrating(db *pop.Connection, assertions Assertions) models.MTOServiceItem {
+	// Set service item type
+	assertions.ReService = models.ReService{
+		ID: uuid.FromStringOrNil("68417bd7-4a9d-4472-941e-2ba6aeaf15f4"), // DCRT - Domestic Crating
```
sandy-wright, line 282: Is it safe to do this? Do we know this service will exist in the database before this function is used? It might in `devseed.go`, but I suspect this could cause an issue if this was used in tests

pkg/testdatagen/make_mto_service_item.go
```diff
@@ -274,3 +274,40 @@ func MakeRealMTOServiceItemWithAllDeps(db *pop.Connection, serviceCode models.Re
...
+	}
+
+	// Add description
+	description := "Decorated horse head to be crated."
+	assertions.MTOServiceItem.Description = &description
```
sandy-wright, line 287: What if the caller tries to pass in a different description as an assertion?

pkg/testdatagen/make_mto_service_item.go
```diff
@@ -274,3 +274,40 @@ func MakeRealMTOServiceItemWithAllDeps(db *pop.Connection, serviceCode models.Re
...
+	// Creat crate
+	MakeMTOServiceItemDimension(db, Assertions{
+		MTOServiceItem: mtoServiceItem,
+		MTOServiceItemDimension: models.MTOServiceItemDimension{
+			Type:   models.DimensionTypeCrate,
+			Length: 11000,
+			Height: 6000,
+			Width:  3500,
+		},
+	})
```
sandy-wright, line 310: I could see someone wanting to use this to test different dimensions, so we should be able to pass in this info as an assertion, I think.

pkg/testdatagen/make_mto_service_item.go
sandy-wright: I think the solution here would be to check if an assertion for the ID was passed in, and if not, create this service code. So this line would go back up into devseed, probably.

sandy-wright (changes requested):
Thanks for adding this `testdatagen` function - it's gonna be super helpful as folks work on the crating feature! I have some notes of things that we should probably change so that it's a little more useful and follows our standards for these functions.

pkg/testdatagen/make_mto_service_item.go
eamahanna: So my only hesitation for adding this back into devseed is it was in every single creation for domestic crating service items in devseed, e2e and the other data gen file. There is also a migration that creates the service table with values `20191101201107_create-re-services-table-with-values.up.sql
  14,3: ('68417bd7-4a9d-4472-941e-2ba6aeaf15f4', 'DCRT', 'Dom. Crating', now(), now()),`

pkg/testdatagen/make_mto_service_item.go
eamahanna: That's fine, I can take this out.

pkg/testdatagen/make_mto_service_item.go
eamahanna: Sounds good

pkg/testdatagen/make_mto_service_item.go
sandy-wright: oh okay, I think that should be fine then!

pkg/testdatagen/make_mto_service_item.go
```diff
@@ -274,3 +274,28 @@ func MakeRealMTOServiceItemWithAllDeps(db *pop.Connection, serviceCode models.Re
...
+		MTOServiceItem: mtoServiceItem,
+	})
+
+	// Creat crate
+	MakeMTOServiceItemDimension(db, Assertions{
+		MTOServiceItem: mtoServiceItem,
+		MTOServiceItemDimension: models.MTOServiceItemDimension{
+			Type: models.DimensionTypeCrate,
+		},
+	})
```
sandy-wright, line 298: Unfortunately, this still won't pass on any assertions the user might have passed in for the dimensions. I could see someone trying to use this like:
```go
cratingItem := testdatagen.MakeMTOServiceItemDomesticCrating(suite.DB(), testdatagen.Assertions{
    MTOServiceItem: models.MTOServiceItem{
        Dimensions: models.MTOServiceItemDimensions{
            models.MTOServiceItemDimension{
               Type: models.DimensionTypeItem,
               Length: 10000,
			   Height: 5000,
			   Width:  2500,
            },
            models.MTOServiceItemDimension{
               Type: models.DimensionTypeCrate,
               Length: 1000,  // maybe we're testing that the crate can't be smaller?
			   Height: 500,
			   Width:  200,
            },
        },
    },
})
```

Being able to handle that full example will be kind of complicated, but I do think we should be passing on assertions in some way, like:
```go
dimAssertions := assertions.MTOServiceItemDimension
MakeMTOServiceItemDimension(db, Assertions{
		MTOServiceItem: mtoServiceItem,
		MTOServiceItemDimension: models.MTOServiceItemDimension{
			Type: models.DimensionTypeCrate,
			Length: dimAssertions.Length,
			...etc
		},
	})
```

pkg/testdatagen/make_mto_service_item.go
```diff
@@ -274,3 +274,28 @@ func MakeRealMTOServiceItemWithAllDeps(db *pop.Connection, serviceCode models.Re
...
+
+	return mtoServiceItem
```
sandy-wright, line 300: It might also be nice to add those two crating records to the output of the service item here. Something like:
```go
shipment := MakeMTOShipment(db, assertions)
	if move != nil {
		// This will allow someone to easily create multiple test shipments for one move
		move.MTOShipments = append(move.MTOShipments, shipment)
	}
	return shipment
``` 
(from `testdatagen/make_mto_shipment.go`)

They are being linked on the database, so any retrieval will get them back. However, if someone uses this function and then expects to see the crating records on the output service item, I don't think they'll see them unless we explicitly add them to that model.

This is optional though! I don't think this one is a big deal either way.

pkg/testdatagen/make_mto_service_item.go
eamahanna: Makes sense. I didn't really think it through all the way yesterday 😓  I just pushed an update that should work now.

pkg/testdatagen/make_mto_service_item.go
eamahanna: Done!

pkg/testdatagen/make_mto_service_item.go
```diff
@@ -274,3 +274,32 @@ func MakeRealMTOServiceItemWithAllDeps(db *pop.Connection, serviceCode models.Re
...
+	})
+
+	// Create item
+	dimensionItem := MakeMTOServiceItemDimension(db, Assertions{
+		MTOServiceItemDimension: assertions.MTOServiceItemDimension,
+		MTOServiceItem:          mtoServiceItem,
+	})
+
+	// Creat crate
+	assertions.MTOServiceItemCrate.Type = models.DimensionTypeCrate
```
sandy-wright, line 296: Oh, nice! Would you be willing to rename this to `MTOServiceItemCrateDimension` or `MTOServiceItemDimensionCrate`? Just because I confused myself when I was first reading through this. Also there is a wee typo in the comment above this!

sandy-wright (approved):
Thanks for making those changes! Looks good - I had one last comment about renaming the new assertion, but it's just a suggestion. And approved so you can merge right away if you do change it 😄

pkg/testdatagen/make_mto_service_item.go
eamahanna: Good call! Sure, thanks for all of your review on this PR. It wasn't my finest work lol.

PR #6697, opened 2021-05-26
[MB-8079] Migrate Modal component from React-USWDS package

kylehilltruss:
## Description

The `Modal` component in the React-USWDS library is deprecated, and will be removed with that library's 2.x release sometime in June.

This pull request:
1. wholesale ports the entire soon-to-be-deprecated component (including logic, stories, tests, and SCSS) into `/src/components/MigratedModal`, 
2. refactors it to match MM coding conventions, 
3. makes the stories look slightly prettier given the global styling in MM, 
4. and then updates the import targets of any components which had previously imported any (`Modal` or `Modal`-related subcomponent) from React-USWDS.

(The component is named as such because we already have a `Modal` component in `/src/components`, which primarily encapsulates the deprecated USWDS version. However, as other components in this project are also importing directly from the deprecated USWDS components, without the pre-existing wrapper, it didn't seem sensible to me to migrate this as a subcomponent of the existing MM `Modal`.)

## Reviewer Notes

* Ensure that everything looks good in storybook and Happo -- as we're just porting over the logic from one location to another, there should be no visual changes
* Ensure that all tests pass
* Ensure that the revised modals work within the relevant contexts of the application
* Ensure that all references to the deprecated `Modal` component and sub-components have been removed:
  * Clone `@trussworks/react-uswds` locally
  * In that repo, delete the entire directory of `/src/components/Modal`, and all content after `/** Truss-designed components */` in `/src/index.ts`
  * Run`yarn build` in that repo
  * In the MyMove repo, `yarn add ~/path/to/react-uswds`
  * Run `yarn storybook` and check that modal components load and operate correctly

## Setup

```sh
make client_run
```

## References

* Closes [MB-8079](https://dp3.atlassian.net/browse/MB-8079).

### DIFF
--- a/src/components/Modal/Modal.jsx
+++ b/src/components/Modal/Modal.jsx
@@ -8,7 +8,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 import styles from './Modal.module.scss';
 
-import { Modal as USWDSModal, connectModal as connectUSWDSModal } from 'components/MigratedModal/MigratedModal';
+import { Modal as MigratedModal, connectModal as connectMigratedModal } from 'components/MigratedModal/MigratedModal';
 
 const Modal = ({ className, ...props }) => {
   const classes = classnames(styles.Modal, className);
@@ -32,7 +32,7 @@ const Modal = ({ className, ...props }) => {
     return () => {};
   });
 
-  return <USWDSModal className={classes} {...props} />;
+  return <MigratedModal className={classes} {...props} />;
 };
 
 Modal.displayName = 'MilMoveModal';
@@ -88,17 +88,17 @@ export const connectModal = (Component) => {
   }
 
   const ConnectedModal = (props) => {
-    // connectUSWDSModal handles isOpen prop & renders with container & overlay
-    const ConnectedUSWDSModal = connectUSWDSModal(Component);
+    // connectMigratedModal handles isOpen prop & renders with container & overlay
+    const ConnectedMigratedModal = connectMigratedModal(Component);
 
     // Render into portal element if it exists
     const MODAL_ROOT_ID = 'modal-root';
     const modalContainer = document.getElementById(MODAL_ROOT_ID);
     if (modalContainer) {
-      return ReactDOM.createPortal(<ConnectedUSWDSModal {...props} />, modalContainer);
+      return ReactDOM.createPortal(<ConnectedMigratedModal {...props} />, modalContainer);
     }
 
-    return <ConnectedUSWDSModal {...props} />;
+    return <ConnectedMigratedModal {...props} />;
   };
 
   ConnectedModal.displayName = `Connected${getDisplayName(Component)}`;
--- a/src/components/Modal/Modal.jsx
+++ b/src/components/Modal/Modal.jsx
@@ -3,10 +3,12 @@ import React, { useEffect } from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-import { Modal as USWDSModal, connectModal as connectUSWDSModal, Button } from '@trussworks/react-uswds';
+import { Button } from '@trussworks/react-uswds';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 import styles from './Modal.module.scss';
+
+import { Modal as USWDSModal, connectModal as connectUSWDSModal } from 'components/MigratedModal/MigratedModal';
 
 const Modal = ({ className, ...props }) => {
   const classes = classnames(styles.Modal, className);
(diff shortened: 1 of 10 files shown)


Shkeating (approved):
looks good to me!
good

kylehilltruss:
Heads up -- there actually _is_ a legitimate CSS problem that I've discovered around the modal migration, which would cause modals on mobile widths to overflow the horizontal space on the screen. (Details below.) Until I fix this and get the Happo tests re-run to a satisfactory level, please don't merge this.

[nerd stuff]
What's happening is that the deprecated, base component that we're porting from USWDS has `min-width: 400px;` set on it. Since our rendering target includes Safari at 375px wide, we don't want that min-width; the existing `Modal` component in MM that wraps the USWDS version has `min-width: 0;` set as a result. 

Both of these rules are set at the same CSS specificity (single-class level) and so the specificity tie is essentially broken by "which rule gets loaded last." For various reasons, that order has changed since the component was migrated.

The simple solution would be to just delete the `min-width` property from the `MigratedModal`, but as _other_ components in MM now directly import the `MigratedModal` and don't necessarily have a `min-width` set for themselves, they're implicitly relying on the USWDS value. Therefore, I need to also set a `min-width` in those other components, if warranted.

CSS: it's okay

kylehilltruss:
> Heads up -- there actually _is_ a legitimate CSS problem that I've discovered around the modal migration, which would cause modals on mobile widths to overflow the horizontal space on the screen. 

This should now be fixed.

Shkeating:
> Heads up -- there actually _is_ a legitimate CSS problem that I've discovered around the modal migration, which would cause modals on mobile widths to overflow the horizontal space on the screen. (Details below.) Until I fix this and get the Happo tests re-run to a satisfactory level, please don't merge this.
> 
> [nerd stuff]
> What's happening is that the deprecated, base component that we're porting from USWDS has `min-width: 400px;` set on it. Since our rendering target includes Safari at 375px wide, we don't want that min-width; the existing `Modal` component in MM that wraps the USWDS version has `min-width: 0;` set as a result.
> 
> Both of these rules are set at the same CSS specificity (single-class level) and so the specificity tie is essentially broken by "which rule gets loaded last." For various reasons, that order has changed since the component was migrated.
> 
> The simple solution would be to just delete the `min-width` property from the `MigratedModal`, but as _other_ components in MM now directly import the `MigratedModal` and don't necessarily have a `min-width` set for themselves, they're implicitly relying on the USWDS value. Therefore, I need to also set a `min-width` in those other components, if warranted.
> 
> CSS: it's okay

I'm glad you are as excited about CSS as I am. Good catch!!

Shkeating (approved):

src/components/Modal/Modal.jsx
```diff
@@ -3,11 +3,13 @@ import React, { useEffect } from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-import { Modal as USWDSModal, connectModal as connectUSWDSModal, Button } from '@trussworks/react-uswds';
+import { Button } from '@trussworks/react-uswds';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 import styles from './Modal.module.scss';
 
+import { Modal as USWDSModal, connectModal as connectUSWDSModal } from 'components/MigratedModal/MigratedModal';
```
YanZ777, line 11: Not sure if the named import of USDSModal should be renamed since it technically isn't USWDS anymore?

src/components/Modal/Modal.module.scss
```diff
@@ -1,7 +1,7 @@
 @import 'shared/styles/colors';
 @import 'shared/styles/_basics';
 
-.Modal {
+div.Modal {
```
YanZ777, line 4: Was this change to address the styling changes that Happo was complaining about?

src/components/Modal/Modal.jsx
kylehilltruss: I suppose I could change it. I went with a policy of minimal changes to the underlying logic.

src/components/Modal/Modal.module.scss
kylehilltruss: Yep; it turned out that the easiest solution was to actually just increase the specificity of this selector rule, so that it breaks the tie with the `MigratedModal` component (class + element > class).

src/components/Modal/Modal.jsx
kylehilltruss: Changed named imports to `MigratedModal` and `connectMigrateModal`.

Ronolibert (approved):
Looks good! Everything behaving as expected and thanks for fixing that CSS issue!

jenniferlynparsons (commented):
Looks good (I know, merged already!). 

A follow up ticket would be good to refactor the two modal components into one, if possible. Since we now have control over all of it, having the wrapper doesn't make sense.

PR #6704, opened 2021-05-27
[MB-8170] DRAFT implementation of Proposal 1 for refactoring the paymentRequestStatusUpdator service

pearl-truss:
## Description

This PR  is a draft PR to illustrate the first proposed way of refactoring the paymentRequestStatusUpdater service. More context on both proposals can be found in the [discovery doc](https://docs.google.com/document/d/1FocswpJeJC-hvGntm7tbE77OnCGo3QgAM4j-ifPpPBQ/edit?pli=1).

This PR shows proposed changes to refactor the paymementRequestStatusUpdator service by breaking it out into two new services:

- UpdateReviewedPaymentRequestStatus which allows the caller to update a payment request to either a  REVIEWED or REVIEWED_AND_ALL_SERVICE_ITEMS_REJECTED status. This is used by both the support api and the ghc api
- UpdateProcessedPaymentRequestStatus which allows the caller to update a payment request to either a SENT_TO_GEX, RECEIVED_BY_GEX, EDI_ERROR or PAID status. This is used by our job runner which is a part of the support api

The desire(need?) to refactor paymementRequestStatusUpdator service  was identified in a PR to address a [bug](https://github.com/transcom/mymove/pull/6660#discussion_r636189811) where the GHC API was allowing payment request status updates to SENT_TO_GEX, EDI_ERROR, RECEIVED_BY_GEX and PAID. We only ever want payment request to be updated to those statuses via our [job runner](https://github.com/transcom/mymove/blob/master/pkg/services/payment_request/payment_request_reviewed_processor.go). This also adheres to a guideline of removing logic from handlers

## Reviewer Notes

- While working on this, I realized another option could be to keep one service and refactor it so that it takes a type argument. For instance the ghc api could call services.PaymentRequestStatusUpdator(&paymentRequest, eTag, "review") and the service can check the type and apply the logic checks for "review", "processed". This would make it easier if in the future we added new payment request statuses that we wanted to apply specific logic to (without making the logic have to live in the handler). I created a gist for this approach here 

## Setup

```sh
make server_run
make server_test
```

You can also trigger the job runner to check that the payment request status changes to SENT_TO_GEX

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8170) for this change
* [discovery doc](https://docs.google.com/document/d/1FocswpJeJC-hvGntm7tbE77OnCGo3QgAM4j-ifPpPBQ/edit?pli=1)


pkg/handlers/ghcapi/payment_request.go
```diff
@@ -127,24 +126,13 @@ func (h UpdatePaymentRequestStatusHandler) Handle(params paymentrequestop.Update
 
 	// Let's fetch the existing payment request using the PaymentRequestFetcher service object
 	existingPaymentRequest, err := h.PaymentRequestFetcher.FetchPaymentRequest(paymentRequestID)
+	existingPaymentRequest.Status = models.PaymentRequestStatus(params.Body.Status)
```
jacquelineIO, line 129: Maybe I'm missing something, but there isn't anything really happening to check the previous status or current state before setting to new status? E.g., payment request should be in "submitted" state and should have meet some prereqs before going to REVIEWED.

jacquelineIO: Let comment in design doc to hopefully clarify my intent.

PR #6708, opened 2021-05-27
[MB-8027] UpdateMTOShipment should not be able to update addresses

shimonacarvalho:
## Description

This started out as just making sure that you can't update addresses that are already populated. 
However in the course of adding tests, I realized that the tests were impossibly convoluted. 
So I cleaned up the tests a lot. Sorry it's a lot of changes - so many that the PR collapses the file :) 
Please **uncollapse mto_shipment_test.c during review.**

Instead of a 1000 line long function, there are now 3 functions for handler, addresslogic and datelogic.
Those should get tested in `server_test`.

Please try **creating and updating**  addresses using this endpoint. You should be able to create, you should not be able to update. The four addresses are :
```
PickupAddress
DestinationAddress
SecondaryPickupAddress
SecondaryDeliveryAddress
```

Do **at least one regular and one secondary address** as the logic for those is slightly different. 

### References
[JIRA Story](https://dp3.atlassian.net/browse/mb-8027)


pkg/handlers/primeapi/api_test.go
```diff
@@ -53,3 +55,36 @@ func TestHandlerSuite(t *testing.T) {
 	suite.Run(t, hs)
 	hs.PopTestSuite.TearDown()
 }
+
```
shimonacarvalho, line 58: Added some helper functions for address comparisons

pkg/handlers/primeapi/mto_shipment_test.go
```diff
@@ -288,84 +291,41 @@ func ClearNonUpdateFields(mtoShipment *models.MTOShipment) *primemessages.MTOShi
 }
 
 func (suite *HandlerSuite) TestUpdateMTOShipmentHandler() {
+
 	primeEstimatedWeight := unit.Pound(500)
-	primeEstimatedWeightDate := testdatagen.DateInsidePeakRateCycle
 	primeActualWeight := unit.Pound(600)
+
+	// Create an available shipment in DB
```
shimonacarvalho, line 298: Simplified the shared db objects to be just 2 - shipment and minimalShipment.

pkg/handlers/primeapi/mto_shipment_test.go
```diff
@@ -288,84 +291,41 @@ func ClearNonUpdateFields(mtoShipment *models.MTOShipment) *primemessages.MTOShi
...
-
-	req := httptest.NewRequest("PATCH", fmt.Sprintf("/mto-shipments/%s", mtoShipment.ID.String()), nil)
-	eTag := etag.GenerateEtag(mtoShipment.UpdatedAt)
-	params := mtoshipmentops.UpdateMTOShipmentParams{
-		HTTPRequest:   req,
-		MtoShipmentID: *handlers.FmtUUID(mtoShipment.ID),
-		Body:          ClearNonUpdateFields(&mtoShipment),
-		IfMatch:       eTag,
-	}
 	planner := &routemocks.Planner{}
```
shimonacarvalho, line 323: Removed shared `params` object. Since this is the update payload that is being tested, it is easier to understand if each test includes it's own payload, rather than patching one payload repeatedly.

pkg/handlers/primeapi/mto_shipment_test.go
```diff
@@ -1286,3 +1517,41 @@ func (suite *HandlerSuite) TestUpdateMTOShipmentStatusHandler() {
...
+// This is to be strictly used for dates as it drops any time parameters in the comparison
+func (suite *HandlerSuite) equalDatePtr(expected *time.Time, actual *strfmt.Date) {
+	if expected == nil || actual == nil {
+		suite.Nil(expected)
+		suite.Nil(actual)
+	} else {
+		isoDate := "2006-01-02" // Create a date format
+		suite.Equal(expected.Format(isoDate), time.Time(*actual).Format(isoDate))
+	}
+}
```
sandy-wright, line 1532: Should this be a generic test suite function? I could see it being useful in other tests!

sandy-wright (approved):
This is great!! The tests are a huge improvement, and this works as expected. There are still a couple of bugs with this endpoint though:

* If you pass in an ID in one of the address payloads when you're trying to create a new address on the shipment, it tries to save that ID in the database. This causes a server error. We need to fix this so that an input UUID for create is caught before it gets to that point (and we may already have a bug ticket for it - I'll add one if not).
* Secondary objects for the shipment's service items aren't populated with data. For example, a crating service item was attached to the shipment I updated, but it returned null info for the CRATE and the ITEM dimensions (even though that info _was_ in the database, so it wasn't like before where the service item had bad data). 

Those are both bug tickets we can add for later! Thanks for slogging through those tests, wow

pkg/handlers/primeapi/mto_shipment_test.go
shimonacarvalho: I was unsure if this could be useful elsewhere but if you think it could be used elsewhere I could make it shared.

sandy-wright: I'm fine with leaving it here until we actually have a moment where we're like "Ah! I could use this function!" And then we can move it haha

pkg/services/mto_shipment/mto_shipment_updater.go
```diff
@@ -759,8 +760,9 @@ func CalculateRequiredDeliveryDate(planner route.Planner, db *pop.Connection, pi
 		distance, distance, weight, weight).First(&ghcDomesticTransitTime)
 
 	if err != nil {
-		return nil, err
+		return nil, errors.Errorf("failed to find transit time for shipment of %d lbs weight and %d mile distance", weight, distance)
```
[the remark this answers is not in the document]
shimonacarvalho: It was returning `"requiredDeliveryDate": "sql: no rows returned"` to the user. ugh.

pkg/handlers/primeapi/mto_shipment_test.go
shimonacarvalho: I moved it already hehe. Just waiting for the tests to complete to merge.

PR #6720, opened 2021-06-02
MB-8223: Design and copy changes to Move 101 page

kylehilltruss:
## Description

Design requested copy and layout [changes to the Move 101 page](https://zpl.io/aNed4xv), which displays as the first screen in the third part of the Customer flow; this PR implements those changes.

In addition, this PR also

* creates a couple of baseline storybook entries for this component
* updates tests for the modified component (since it relied on specific copy and `data-testid`s that changed)
* deletes the entire `/src/images` folder, since the contents were only being used on the old version of this page pre-redesign
* deletes the entire `MilmoveCard` component (plus tests/SCSS/stories), since it too was only being used on the old version of this page pre-redesign

## Reviewer Notes

There's not really much in the way of functionality changes here, just content presentation. It's worth noting that a few things do change between mobile and tablet/desktop widths (notably: section heading font size and presentation, and icon sizes), so you might want to check out that everything looks sensible at a variety of window dimensions.

You can view this in storybook now, under Customer / pages / Move 101.

## Setup

```sh
make server_run
make client_run
```
then check that the page that shows after clicking on "Plan your shipments" from the Customer home page looks good on both mobile and desktop widths.

## References

* Closes [MB-8223](https://dp3.atlassian.net/browse/MB-8223).


src/pages/MyMove/MovingInfo.test.jsx
```diff
@@ -17,19 +17,17 @@ describe('MovingInfo component', () => {
   };
 
   it('renders the expected content', () => {
-    const { queryAllByTestId, queryByText } = render(<MovingInfo {...testProps} />);
+    const { queryByText, queryAllByRole } = render(<MovingInfo {...testProps} />);
```
felipe-lee, line 20: This can also be changed to not capture anything being returned and instead using `screen.queryByText` below. `screen` can be imported like `render` is.

https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

src/pages/MyMove/MovingInfo.test.jsx
```diff
@@ -17,19 +17,17 @@ describe('MovingInfo component', () => {
...
 
-    expect(queryByText('Tips for planning your shipments')).toBeInTheDocument();
-    expect(queryAllByTestId('shipmentsAlert').length).toBe(1);
+    expect(queryByText('Things to know about selecting shipments')).toBeInTheDocument();
```
felipe-lee, line 22: Since this is for a header, can you change to using 
```js
getByRole('heading', {level: 1, name: 'Things to know about selecting shipments'})
```

https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-the-wrong-query

src/pages/MyMove/MovingInfo.test.jsx
```diff
@@ -17,19 +17,17 @@ describe('MovingInfo component', () => {
...
   });
 
   it('renders with no errors when entitlement weight is 0', () => {
-    const { queryAllByTestId, queryByText } = render(<MovingInfo {...testProps} entitlementWeight={0} />);
+    const { queryByText, queryAllByRole } = render(<MovingInfo {...testProps} entitlementWeight={0} />);
 
-    expect(queryByText('Tips for planning your shipments')).toBeInTheDocument();
-    expect(queryAllByTestId('shipmentsAlert').length).toBe(0);
-    expect(queryAllByTestId('shipmentsSubHeader').length).toBe(4);
+    expect(queryByText('Things to know about selecting shipments')).toBeInTheDocument();
```
felipe-lee, line 30: Same for this one as above re: `getByRole` for the h1

src/pages/MyMove/MovingInfo.test.jsx
```diff
@@ -17,19 +17,17 @@ describe('MovingInfo component', () => {
...
   it('renders the expected content', () => {
-    const { queryAllByTestId, queryByText } = render(<MovingInfo {...testProps} />);
+    const { queryByText, queryAllByRole } = render(<MovingInfo {...testProps} />);
 
-    expect(queryByText('Tips for planning your shipments')).toBeInTheDocument();
-    expect(queryAllByTestId('shipmentsAlert').length).toBe(1);
+    expect(queryByText('Things to know about selecting shipments')).toBeInTheDocument();
     expect(queryByText(/7,000 lbs/)).toBeInTheDocument();
-    expect(queryAllByTestId('shipmentsSubHeader').length).toBe(4);
+    expect(queryAllByRole('heading').length).toBe(6);
```
felipe-lee, line 24: I think generally we should only use `query...` when you're checking that something _isn't_ on a page since it's the only one that won't throw an error, but we should use `get...` or `find...` if we expect something to exist: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence

src/pages/MyMove/MovingInfo.test.jsx
```diff
@@ -17,19 +17,17 @@ describe('MovingInfo component', () => {
...
 
   it('renders with no errors when entitlement weight is 0', () => {
-    const { queryAllByTestId, queryByText } = render(<MovingInfo {...testProps} entitlementWeight={0} />);
+    const { queryByText, queryAllByRole } = render(<MovingInfo {...testProps} entitlementWeight={0} />);
 
-    expect(queryByText('Tips for planning your shipments')).toBeInTheDocument();
-    expect(queryAllByTestId('shipmentsAlert').length).toBe(0);
-    expect(queryAllByTestId('shipmentsSubHeader').length).toBe(4);
+    expect(queryByText('Things to know about selecting shipments')).toBeInTheDocument();
+    expect(queryAllByRole('heading').length).toBe(5);
```
felipe-lee, line 31: Same comment about `query...` vs `get/find...`

src/pages/MyMove/MovingInfo.jsx
```diff
@@ -1,23 +1,39 @@
...
+    </Grid>
+  </Grid>
+);
+
+IconSection.propTypes = {
+  icon: string.isRequired,
+  headline: string.isRequired,
+  children: node.isRequired,
+};
+
```
felipe-lee, line 36: I think it's more common in our code for this to be its own component, but I could definitely be wrong so I think it'd be good if someone commented on this. It could even possibly be a shared (customer + office) component in case the office side ever needs it.

felipe-lee (commented):
The js/jsx look mostly good to me. Left a few comments though. Also, going to the page, it looks like the design spec to me, but I'll leave that for design to approve or not.

src/pages/MyMove/MovingInfo.jsx
YanZ777: I tend to making it into its own component if we plan on reusing this.
I would be interested in hearing what design has to say in particular with regards to reusing this format. If the answer is yes, they are going to make more designs with this layout, then I would vote for making this be its own component.
I am open to hearing why not its own component though.

src/pages/MyMove/MovingInfo.jsx
```diff
@@ -36,100 +52,39 @@ export class MovingInfo extends Component {
...
-                The government will pay to move that much weight. Your whole move, no matter how many shipments it
-                takes.
+            <h1 className={styles.ShipmentsHeader}>Things to know about selecting shipments</h1>
+            <SectionWrapper className={styles.Wrapper}>
+              {entitlementWeight !== 0 && (
+                <IconSection
+                  icon="weight-hanging"
+                  headline={`You can move ${formatWeight(entitlementWeight)} in this move.`}
+                >
+                  You’ll have to pay for any excess weight the government moves.
```
YanZ777, line 62: I think we tend to use &apos;

src/pages/MyMove/MovingInfo.test.jsx
kylehilltruss: Good point -- updated this (and all the other test-based comments).

kylehilltruss:
• https://my-milmove-pr-6720.mymove.sandbox.truss.coffee
• https://admin-milmove-pr-6720.mymove.sandbox.truss.coffee
• https://office-milmove-pr-6720.mymove.sandbox.truss.coffee
• https://prime-milmove-pr-6720.mymove.sandbox.truss.coffee

src/pages/MyMove/MovingInfo.jsx
```diff
@@ -36,100 +52,39 @@ export class MovingInfo extends Component {
...
-                    </p>
-                    <p>When things get complicated or you have questions during your move, they are there to help.</p>
-                    <p>
-                      It’s OK if things change after you submit your move info. Your movers or your counselor will make
-                      things work.
-                    </p>
-                  </CardBody>
-                </MilmoveCard>
-              </CardGroup>
+                If you use movers, they will be your point of contact throughout your move and can also help you make
```
cosburn, line 70: I think it's a better experience on screen readers if text is inside semantic HTML elements like <p> (this is currently outputting inside a <div>). That might also provide some paragraph spacing so the <br>s wouldn't be needed.

cosburn:
Looking really really good! Just a couple notes (and I added a comment with a semantics note), otherwise looks awesome!

1 - can we add some spacing below the h2s on desktop view - 8px/1unit
2 - can we remove some of the extra spacing caused by the `ul` so that the spacing between sections is equal

Screen Shot 2021-06-03 at 10 45 09 AM
Screen Shot 2021-06-03 at 10 45 25 AM

src/pages/MyMove/MovingInfo.jsx
kylehilltruss: @cosburn Do you think you might reuse this design element (the individual sections with the greyed out icons on the top left) anywhere else in the future?

src/pages/MyMove/MovingInfo.jsx
kylehilltruss: Fixed.

kylehilltruss:
w/r/t point 1 here https://github.com/transcom/mymove/pull/6720#issuecomment-853981737:

There's actually a loose, component-specific but globally-applied CSS selector [here](https://github.com/transcom/mymove/blob/master/src/scenes/PpmLanding/PpmSummary.css#L1-L3) that causes all `h2`s (and only h2s) to have a `margin-bottom: 0`, which means you see something different in Storybook (when that component's CSS isn't loaded) and in the application proper. We should probably look into fixing that in the future.

Updated the styling to match both requests.

YanZ777 (approved):
Code changes look good to me if you don't need to extract out the IconSection into its own component.

cosburn (approved):
Looks great!

PR #6722, opened 2021-06-03
[MB-3669]  use zip3 when determining line haul vs short haul

pearl-truss:
## Description

This PR updates our distanceZip3 and DistanceZip5 lookups to no longer use `50 miles` as the discriminator between linehaul and shorthaul. Instead the check should is if the zip3 is the same.

If pickup zip3 == destination zip3 --> shorthaul
If pickup zip3 != destination zip3 --> linehaul

## Setup

```sh
make server_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3669) for this change

### DIFF
--- a/pkg/payment_request/service_param_value_lookups/distance_zip3_lookup.go
+++ b/pkg/payment_request/service_param_value_lookups/distance_zip3_lookup.go
@@ -55,7 +55,10 @@ func (r DistanceZip3Lookup) lookup(keyData *ServiceItemParamKeyData) (string, er
 		return "", err
 	}
 
-	if distanceMiles >= 50 {
+	pickupZip3 := pickupZip[:3]
+	destinationZip3 := destinationZip[:3]
+
+	if pickupZip3 != destinationZip3 {
 		miles := unit.Miles(distanceMiles)
 		mtoShipment.Distance = &miles
 		err = db.Save(&mtoShipment)
--- a/pkg/payment_request/service_param_value_lookups/distance_zip3_lookup.go
+++ b/pkg/payment_request/service_param_value_lookups/distance_zip3_lookup.go
@@ -2,6 +2,7 @@ package serviceparamvaluelookups
 
 import (
 	"database/sql"
+	"fmt"
 	"strconv"
 
 	"github.com/gofrs/uuid"
@@ -55,9 +56,12 @@ func (r DistanceZip3Lookup) lookup(keyData *ServiceItemParamKeyData) (string, er
 		return "", err
 	}
 
+	if len(pickupZip) != 5 || len(destinationZip) != 5 {
+		return "", services.NewInvalidInputError(*mtoServiceItem.MTOShipmentID, fmt.Errorf("Shipment must have valid pickup and destination zip codes"), nil, "")
+	}
+
 	pickupZip3 := pickupZip[:3]
 	destinationZip3 := destinationZip[:3]
-
 	if pickupZip3 != destinationZip3 {
 		miles := unit.Miles(distanceMiles)
 		mtoShipment.Distance = &miles
--- a/pkg/payment_request/service_param_value_lookups/distance_zip3_lookup.go
+++ b/pkg/payment_request/service_param_value_lookups/distance_zip3_lookup.go
@@ -56,11 +56,13 @@ func (r DistanceZip3Lookup) lookup(keyData *ServiceItemParamKeyData) (string, er
 		return "", err
 	}
 
+	errorMsgForPickupZip := fmt.Sprintf("Shipment must have valid pickup zipcode. Received: %s", pickupZip)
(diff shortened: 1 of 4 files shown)


pkg/payment_request/service_param_value_lookups/distance_zip3_lookup.go
```diff
@@ -55,7 +55,10 @@ func (r DistanceZip3Lookup) lookup(keyData *ServiceItemParamKeyData) (string, er
 		return "", err
 	}
 
-	if distanceMiles >= 50 {
+	pickupZip3 := pickupZip[:3]
+	destinationZip3 := destinationZip[:3]
```
reggieriser, line 59: Can we trust that the `pickupZip` and `destinationZip` will always have at least length 3?  Those are coming from the database in this case, so hopefully that's true because of validations we do before it gets into the DB, but just wanted to mention it since the code will panic otherwise.

pkg/payment_request/service_param_value_lookups/distance_zip5_lookup_test.go
```diff
@@ -38,8 +51,52 @@ func (suite *ServiceParamValueLookupsSuite) TestDistanceZip5Lookup() {
...
+		//RA: Linter flags errcheck error: Ignoring a method's return value can cause the program to overlook unexpected states and conditions.
+		//RA: Functions with unchecked return values in the file are used fetch data and assign data to a variable that is checked later on
+		//RA: Given the return value is being checked in a different line and the functions that are flagged by the linter are being used to assign variables
+		//RA: in a unit test, then there is no risk
+		//RA Developer Status: Mitigated
+		//RA Validator Status: Mitigated
+		//RA Modified Severity: N/A
+		// nolint:errcheck
+		suite.DB().
+			Find(&mtoShipment, mtoServiceItem.MTOShipmentID)
```
reggieriser, line 99: Is there any reason we can't just do a `suite.NoError(err)` here (and in the other places in tests you touched) so that we don't have to include these annotation comments?  If we add a new place we annotate, my understanding is that we have to also get ISSO approval (see [this Wiki page](https://github.com/transcom/mymove/wiki/Guide-to-Static-Analysis-Annotations-for-Disabled-Linters)).  Hopefully we can avoid that.

reggieriser (commented):
Generally looks good and tests passed.  I did have a question about a possible panic as well as some linter annotations that I wanted to run past you before approving.

pkg/payment_request/service_param_value_lookups/distance_zip5_lookup_test.go
pearl-truss: That's a good catch! I believe it's fixed now

pkg/payment_request/service_param_value_lookups/distance_zip3_lookup_test.go
```diff
@@ -1,12 +1,3 @@
-//RA Summary: gosec - errcheck - Unchecked return value
```
mboren, line 1: love the removed annotations! 🎉

pkg/payment_request/service_param_value_lookups/distance_zip3_lookup.go
```diff
@@ -56,8 +56,11 @@ func (r DistanceZip3Lookup) lookup(keyData *ServiceItemParamKeyData) (string, er
 		return "", err
 	}
 
-	if len(pickupZip) != 5 || len(destinationZip) != 5 {
-		return "", services.NewInvalidInputError(*mtoServiceItem.MTOShipmentID, fmt.Errorf("Shipment must have valid pickup and destination zip codes"), nil, "")
+	if len(pickupZip) < 5 {
+		return "", services.NewInvalidInputError(*mtoServiceItem.MTOShipmentID, fmt.Errorf("Shipment must have valid pickup zipcode. Received: %s", pickupZip), nil, fmt.Sprintf("Shipment must have valid pickup zipcode. Received: %s", pickupZip))
+	}
+	if len(destinationZip) < 5 {
+		return "", services.NewInvalidInputError(*mtoServiceItem.MTOShipmentID, fmt.Errorf("Shipment must have valid destination zipcode. Received: %s", destinationZip), nil, fmt.Sprintf("Shipment must have valid destination zipcode. Received: %s", destinationZip))
```
reggieriser, line 63: Nit: You could put the `Sprintf`'d literal text that's repeated here into a `string` variable and just reference that `string` in the two places to make it easier to keep them in sync.  I'm not sure I completely follow why we need it in both places, but I assume that's an idiosyncrasy of the `InvalidInputError` and how this propagates up the chain.  Not a blocker, though.

reggieriser (approved):
Added one nit, but not a blocker.  LGTM!  :shipit:

PR #6724, opened 2021-06-03
MB-7694-make-pickup-address-viewable

NamibiaTorres:
## Description

Provide the user (Services Counselor) a view containing the additional pick-up address with the rest of the shipment information when it is present.

## Reviewer Notes

This is my first time writing a jest test for a conditional. Does the logic work out ok? 

Also, please provide extra scrutiny on the conditional: If a secondary address is present, then it is displayed. Did I write the conditional following convention?

## Setup

To run the unit test:
```sh
yarn test src/components/Office/ShipmentDisplay/ShipmentDisplay.test.jsx
```

To see whether secondary address is viewable by the Service Counselor (SC):
1. `make server_run` `make client_run`
2. Local  sign-in into **officelocal**
3. Select **Service Counselor** sign-in option
4. Select an account with a secondary address and see if that secondary address appears in the view card with the address data.
5. Repeat step for for an account with _no secondary address_ and make sure that the the secondary address view is not present.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7694) for this change

## Screenshots

Screen Shot 2021-06-03 at 9 15 47 AM
Screen Shot 2021-06-03 at 9 19 06 AM


src/components/Office/ShipmentDisplay/ShipmentDisplay.jsx
```diff
@@ -47,6 +47,12 @@ const ShipmentDisplay = ({ shipmentType, displayInfo, onChange, shipmentId, isSu
             <dt>Current address</dt>
             <dd>{displayInfo.currentAddress && formatAddress(displayInfo.currentAddress)}</dd>
           </div>
+          {displayInfo.secondaryPickupAddress && [
+            <div className={styles.row}>
+              <dt>Second pickup address</dt>
+              <dd>{displayInfo.secondaryPickupAddress && formatAddress(displayInfo.secondaryPickupAddress)}</dd>
```
eamahanna, line 53: We already checked for the existence of secondaryPickupAddress, so we don't need to check again. You can have just the formatAddress .

src/components/Office/ShipmentDisplay/ShipmentDisplay.test.jsx
```diff
@@ -43,6 +67,12 @@ describe('Shipment Container', () => {
     );
     expect(wrapper.find('div[data-testid="shipment-display"]').exists()).toBe(true);
   });
+  it('renders secondary address info when present', () => {
+    const wrapper = shallow(
+      <ShipmentDisplay shipmentId="1" displayInfo={secondaryAddressInfo} onChange={jest.fn()} isSubmitted={false} />,
+    );
+    expect(wrapper.find('div[data-testid="shipment-display"]').exists()).toBe(true);
```
eamahanna, line 74: Check that it renders when present and that it doesn't exist when not present.

src/components/Office/ShipmentDisplay/ShipmentDisplay.test.jsx
```diff
@@ -22,6 +22,30 @@ const info = {
...
+    state: 'TX',
+    postal_code: '78234',
+  },
+  secondaryAddress: {
+    street_address_1: '800 S 2nd St',
+    city: 'San Antonio',
+    state: 'TX',
+    postal_code: '78234',
+  },
+  destinationAddress: {
```
eamahanna, line 40: I think you can use the info object in the definition (I could be wrong though)
```
const secondaryAddressInfo = {
   info,
  secondaryAddress: ...
}
```

pkg/handlers/internalapi/internal/payloads/model_to_payload.go
```diff
@@ -67,17 +67,18 @@ func MTOAgents(mtoAgents *models.MTOAgents) *internalmessages.MTOAgents {
...
+		ShipmentType:           internalmessages.MTOShipmentType(mtoShipment.ShipmentType),
+		CustomerRemarks:        mtoShipment.CustomerRemarks,
+		PickupAddress:          Address(mtoShipment.PickupAddress),
+		SecondaryPickupAddress: Address(mtoShipment.SecondaryPickupAddress),
+		DestinationAddress:     Address(mtoShipment.DestinationAddress),
+		CreatedAt:              strfmt.DateTime(mtoShipment.CreatedAt),
+		UpdatedAt:              strfmt.DateTime(mtoShipment.UpdatedAt),
+		Status:                 internalmessages.MTOShipmentStatus(mtoShipment.Status),
+		ETag:                   etag.GenerateEtag(mtoShipment.UpdatedAt),
 	}
```
sandy-wright, line 82: The Service Counselors view uses the GHC API, not the Internal API, so I don't think this change does anything for us here. I would remove it.

sandy-wright (changes requested):
This doesn't work currently - there are a couple of changes you'll need to do to make the second pickup address show up:

* You need to make sure the `secondaryPickupAddress` field is getting passed into the `<ShipmentDisplay>` component. You'll do that in this chunk of code: https://github.com/transcom/mymove/blob/master/src/pages/Office/ServicesCounselingMoveDetails/ServicesCounselingMoveDetails.jsx#L56-L64
* You need to make sure that info is coming back from the database. I did this in my PR as well: https://github.com/transcom/mymove/blob/9a776b0389e41cbbba17bcc7b35a5dc2de96538e/pkg/handlers/ghcapi/mto_shipment.go#L73

Then it should work! And you'll have to update the tests for the Services Counselor component ☺️

pkg/handlers/internalapi/internal/payloads/model_to_payload.go
NamibiaTorres: Ok, thanks! We weren't sure where exactly it would go.

NamibiaTorres: Looks like the ghc api payload already has secondaryPickupAddress, so I guess no changes need to made there?

sandy-wright: Yup, looks like that part was already good to go!

src/components/Office/ShipmentDisplay/ShipmentDisplay.jsx
```diff
@@ -47,6 +47,12 @@ const ShipmentDisplay = ({ shipmentType, displayInfo, onChange, shipmentId, isSu
             <dt>Current address</dt>
             <dd>{displayInfo.currentAddress && formatAddress(displayInfo.currentAddress)}</dd>
           </div>
+          {displayInfo.secondaryPickupAddress && (
+            <div className={styles.row}>
+              <dt>Second pickup address</dt>
+              <dd>{formatAddress(displayInfo.secondaryPickupAddress)}</dd>
+            </div>
```
sandy-wright, line 54: This is a picky request, but would you be willing to call this prop `secondPickupAddress` instead? It seems like the prop names are matching the labels - not the shipment field names - in this component

NamibiaTorres: Done! Made the update. Great suggestion, definitely don't want to confuse things.

src/components/Office/ShipmentDisplay/ShipmentDisplay.test.jsx
```diff
@@ -28,6 +28,16 @@ const info = {
   counselorRemarks: 'counselor approved',
 };
 
+const secondaryAddressInfo = {
+  info,
+  secondaryAddress: {
```
sandy-wright, line 33: This needs to have the name you'd use in the component - so `secondPickupAddress`. It would be good to name the `const` to something like `secondPickupAddressInfo` too, so that it's more explicit.

src/components/Office/ShipmentDisplay/ShipmentDisplay.test.jsx
```diff
@@ -68,6 +78,16 @@ describe('Shipment Container', () => {
...
   });
+  it('renders secondary address info when present', () => {
+    render(
+      <ShipmentDisplay shipmentId="1" displayInfo={secondaryAddressInfo} onChange={jest.fn()} isSubmitted={false} />,
+    );
+    expect(screen.getByText('Second pickup address')).toBeInTheDocument();
+  });
+  it('renders no secondary address info when it is not present', () => {
+    render(<ShipmentDisplay shipmentId="1" displayInfo={info} onChange={jest.fn()} isSubmitted={false} />);
+    expect(screen.getByText('Second pickup address')).not.toBeInTheDocument();
```
sandy-wright, line 89: When I run the tests locally, there seems to be an error with this line. It might be related to the incorrect naming above though, so once you fix that it might fix all the other errors in the test.

sandy-wright (commented):
LGTM! Only thing left to handle is an error in the `ShipmentDisplay.test.jsx` file. The rest looks good! Once that's passing, I'm good to approve 😄

sandy-wright (approved):
🚀 ✨

PR #6726, opened 2021-06-03
MB-6464 generalize filters

nelztruss:
## Description

Reduce code duplication by making a utility for converting URL query parameters into concrete `QueryFilter` types.

### DIFF
--- a/pkg/handlers/adminapi/access_code_test.go
+++ b/pkg/handlers/adminapi/access_code_test.go
@@ -91,27 +91,19 @@ func (suite *HandlerSuite) TestIndexAccessCodesHandler() {
 }
 
 func (suite *HandlerSuite) TestIndexAccessCodesHandlerHelpers() {
-	queryBuilder := query.NewQueryBuilder(suite.DB())
-	handler := IndexAccessCodesHandler{
-		HandlerContext:        handlers.NewHandlerContext(suite.DB(), suite.TestLogger()),
-		NewQueryFilter:        query.NewQueryFilter,
-		AccessCodeListFetcher: accesscode.NewAccessCodeListFetcher(queryBuilder),
-		NewPagination:         pagination.NewPagination,
-	}
-
 	suite.T().Run("test both filters present", func(t *testing.T) {
 
 		s := `{"move_type":"PPM", "code":"ABC123"}`
-		qfs := handler.generateQueryFilters(&s, suite.TestLogger())
+		qfs := generateQueryFilters(suite.TestLogger(), &s, accessCodeFilterConverters)
 		expectedFilters := []services.QueryFilter{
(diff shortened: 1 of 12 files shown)


eamahanna (commented):
This looks like a good approach to me!  Let me know when you finalize the PR and I will do another review.

pkg/handlers/adminapi/access_code_test.go
```diff
@@ -91,35 +91,27 @@ func (suite *HandlerSuite) TestIndexAccessCodesHandler() {
...
 
 		s := `{"move_type":"PPM", "code":"ABC123"}`
-		qfs := handler.generateQueryFilters(&s, suite.TestLogger())
+		qfs := generateQueryFilters(suite.TestLogger(), &s, accessCodeFilterConverters)
 		expectedFilters := []services.QueryFilter{
 			query.NewQueryFilter("move_type", "=", "PPM"),
 			query.NewQueryFilter("code", "=", "ABC123"),
 		}
-		suite.Equal(expectedFilters, qfs)
+		suite.ElementsMatch(expectedFilters, qfs) // order not important
```
sandy-wright, line 102: Oh, this is a handy test function

pkg/handlers/adminapi/query_filters.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/transcom/mymove/pkg/handlers"
+	"github.com/transcom/mymove/pkg/services"
+)
+
+// generateQueryFilters is helper to convert filter params from a json string into a slice of services.QueryFilter
+//
+// The convertFns is a mapping from the filter parameter name(s) to the function that will operate on the
+// associated non-empty filter parameter value
+func generateQueryFilters(log handlers.Logger, params *string, convertFns map[string]func(string) []services.QueryFilter) (results []services.QueryFilter) {
```
sandy-wright, line 16: This looks great! But I'm not convinced that this should be in the `adminapi` package - it's a generic utility (now) that has no Admin API dependencies. I would consider moving this either to the base `handlers` level, or all the way to the `query` package in `./pkg/services/`

sandy-wright (approved):
LGTM! I have a suggestion to move the new `genericQueryFilters` function to a more generic/useful package, but it's not a hard requirement

pkg/handlers/adminapi/query_filters.go
nelztruss: Per your comment I went looking, and no where else under `handlers` does anyone actually pull their filters off of the query parameters... All usages just create `[]services.QueryFilter` by hand.

So while someone else _could_ find it useful, there isn't currently anywhere else that _does_ need to use this. So, for now I'm leaning towards keeping it under `adminapi`, until there's a demonstrated need to give it broader application.

PR #6731, opened 2021-06-03
MB-8275 Allow shipment to be diverted

monfresh:
## Description

This PR allows a TOO to request a shipment to be diverted, which in the backend translates to sending the status `Diversion Requested` to the `PatchMTOShipmentStatus` endpoint.

While working on this, I saw an opportunity to refactor the state machine logic into a separate `ShipmentRouter` service to match what we do in `MoveRouter`. This also surfaced a bunch of buggy tests which I fixed.

I also took the opportunity to clean up the `ghc.yaml` to specify only the statuses that can be set via the `PatchMTOShipmentStatus`, so I removed `SUBMITTED` and added all the other ones (except DRAFT).

You'll notice changes to `api.go` to allow passing in the logger to the MoveRouter. These changes might seem unrelated to this PR, and on their own, they are. I was planning on using the logger in the Shipment Router, but the router is currently only called in the `MTOShipmentUpdater` service. I thought about injecting the router in the handler, and then passing it in as an argument to the `MTOShipmentUpdater` service. But I think eventually, we want to create separate endpoints for each shipment status, to make it easier to deal with authorization rules, because only certain roles or entities will be able to set certain statuses. When that happens, we can revisit the architecture.

## Reviewer Notes

## Setup

```sh
echo "Code goes here"
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8275) for this change


pkg/services/mto_shipment/shipment_router.go
```diff
@@ -0,0 +1,153 @@
...
+// }
+
+// func (router shipmentRouter) logMove(shipment *models.MTOShipment) {
+// 	router.logger.Info("Move log",
+// 		zap.String("Move.ID", move.ID.String()),
+// 		zap.String("Move.Locator", move.Locator),
+// 		zap.String("Move.Status", string(move.Status)),
+// 		zap.String("Move.OrdersID", move.OrdersID.String()),
+// 	)
+// }
```
monfresh, line 153: I will remove this.

monfresh: Or rather, remove the findShipment, and turn on logging to match what we're doing in the MoveRouter.

pkg/services/mto_shipment/shipment_router.go
```diff
@@ -0,0 +1,153 @@
...
+
+	shipment.Status = models.MTOShipmentStatusRejected
+	shipment.RejectionReason = reason
+
+	return nil
+}
+
+// RequestDiversion is called when the TOO has requested that the Prime divert the shipment.
+func (router shipmentRouter) RequestDiversion(shipment *models.MTOShipment) error {
+	if shipment.Status != models.MTOShipmentStatusApproved {
```
monfresh, line 118: Previously, requesting a diversion was only possible if the current state of the shipment was `CancellationRequested`, but in Slack, it sounded like the shipment needs to be in the `Approved` status. It would be great to get confirmation on this.

pkg/handlers/internalapi/api.go
```diff
@@ -34,119 +36,112 @@ func NewInternalAPI(context handlers.HandlerContext) *internalops.MymoveAPI {
...
-		// Unable to get logger to pass in for the instantiation of
-		// move.InitNewMoveRouter(h.DB(), logger),
-		// This limitation has come up a few times
-		// - https://dp3.atlassian.net/browse/MB-2352 (story to address issue)
-		// - https://ustcdp3.slack.com/archives/CP6F568DC/p1592508325118600
-		// - https://github.com/transcom/mymove/blob/c42adf61735be8ee8e5e83f41a656206f1e59b9d/pkg/handlers/primeapi/api.go
-		// As a temporary workaround move.InitNewMoveRouter
-		// is called directly in the handler
+		ctx,
+		move.NewMoveRouter(ctx.DB(), ctx.LoggerFromContext(context.Background())),
```
Ryan-Koch, line 72: Nice, I didn't know about `context.Background()` this revealed I have some reading about the Go context interface to do.

pkg/services/mto_shipment/shipment_router.go
```diff
@@ -0,0 +1,153 @@
+package mtoshipment
```
Ryan-Koch, line 1: Thanks for setting up a router pattern for shipments! This will give us some flexibility as state machine transitions get more complex potentially down the road.

pkg/services/mto_shipment/shipment_router.go
```diff
@@ -0,0 +1,153 @@
...
+			transitionToStatus:        models.MTOShipmentStatusDiversionRequested,
+			transitionAllowedStatuses: &[]models.MTOShipmentStatus{models.MTOShipmentStatusApproved},
+		}
+	}
+	shipment.Status = models.MTOShipmentStatusDiversionRequested
+
+	return nil
+}
+
+// func (router *shipmentRouter) findShipment(shipment *models.MTOShipment) (*error) {
```
Ryan-Koch, line 131: Is this commented out section removable or do we want to hold onto it?

pkg/services/mto_shipment/shipment_router.go
monfresh: You're welcome! I think eventually, we might want to have separate endpoints for each status to keep the logic for each separate, and to make it easier to add authorization if only certain roles can perform certain status changes.

pkg/services/mto_shipment/shipment_router.go
monfresh: I will remove it.

pkg/services/mto_shipment/shipment_router.go
Ryan-Koch: What you've done here is correct, the shipment will be in an `Approved` status as opposed to the `CancellationRequested` status. There's a bit of a visual flow that shows why over here: https://miro.com/app/board/o9J_kqcaaWY=/?moveToWidget=3074457357116150204&cot=14 . But the `CancellationRequested` flow is a different thing, I'm not sure why that was originally being used here, it does apply to `Cancel()` though which it looks like carried over correctly.

Ryan-Koch (approved):

pkg/services/mto_shipment/shipment_router.go
```diff
@@ -0,0 +1,153 @@
...
+func (router shipmentRouter) RequestDiversion(shipment *models.MTOShipment) error {
+	if shipment.Status != models.MTOShipmentStatusApproved {
+		return ConflictStatusError{
+			id:                        shipment.ID,
+			transitionFromStatus:      shipment.Status,
+			transitionToStatus:        models.MTOShipmentStatusDiversionRequested,
+			transitionAllowedStatuses: &[]models.MTOShipmentStatus{models.MTOShipmentStatusApproved},
+		}
+	}
+	shipment.Status = models.MTOShipmentStatusDiversionRequested
```
donaldthai, line 126: So along with the status, the `diversion` property needs to be updated too. Set to `true`.

monfresh: I was just thinking about that! I was wondering if it had to be set at this point, or wait until the diversion gets approved. Is it possible for a diversion request to be denied? If so, then maybe we should wait until it's approved before setting the Diversion field? @Ryan-Koch thoughts?

Ryan-Koch: The prime should be setting this to `true` once they've decided to accept the diversion request and mark the shipment. I believe the work to enable them to set that during their shipment update is over here: https://github.com/transcom/mymove/pull/6576/files#diff-61462ff1f1c3c378fa84dac247390a2b0b2cd35d5de3c2ea7a54d3a858f50572R56 . So no need to set the flag at this stage just yet.

pkg/handlers/ghcapi/mto_shipment_test.go
```diff
@@ -335,48 +341,16 @@ func (suite *HandlerSuite) TestPatchMTOShipmentHandler() {
...
-		okResponse := response.(*mtoshipmentops.PatchMTOShipmentStatusOK)
-		suite.Equal(mtoShipment.ID.String(), okResponse.Payload.ID.String())
-		suite.NotNil(okResponse.Payload.ETag)
-
-		// Check that webhook notification was stored
-		suite.HasWebhookNotification(mtoShipment.ID, handlerContext.GetTraceID())
-
-	})
-
-	suite.T().Run("Successful patch with no webhook notification - On an unsubmitted move", func(t *testing.T) {
```
monfresh, line 368: This test is redundant. The fact that a webhook notification is not generated when the move is not yet available to the prime has nothing to do with this handler or the service that this handler uses. The logic that determines when a webhook notification is generated is defined in the webhook notification code, which is already tested there.

pkg/services/mto_shipment/mto_shipment_updater_test.go
```diff
@@ -749,6 +764,23 @@ func (suite *MTOShipmentServiceSuite) TestUpdateMTOShipmentStatus() {
...
+		})
+		eTag = etag.GenerateEtag(shipmentToDivert.UpdatedAt)
+
+		_, err := updater.UpdateMTOShipmentStatus(
+			shipmentToDivert.ID, models.MTOShipmentStatusDiversionRequested, nil, eTag)
+		suite.DB().Find(&shipmentToDivert, shipmentToDivert.ID)
+
+		suite.NoError(err)
+		suite.Equal(models.MTOShipmentStatusDiversionRequested, shipmentToDivert.Status)
+	})
```
monfresh, line 783: Now that I added tests for all the state machine logic in `shipment_router_test.go`, we should be able to remove all of these redundant state machine tests from this file.

pkg/services/mto_shipment/shipment_router.go
```diff
@@ -0,0 +1,146 @@
...
+	db *pop.Connection
+}
+
+// NewShipmentRouter creates a new shipmentRouter service
+func NewShipmentRouter(db *pop.Connection) services.ShipmentRouter {
+	return &shipmentRouter{db}
+}
+
+// Submit is used to submit a shipment at the time the customer submits
+// their move.
```
monfresh, line 23: I need to update this comment to specify that this should also be called when the Prime creates a move.

PR #6735, opened 2021-06-03
[MB-8239] Admin can pass in nil for user active status

shimonacarvalho:
## Description

Admin app was always passing in the user active status no matter whether it was changed or not. 
This turned out to be due to the fact that it was a required field in the adminapi. 
Changed the field to be not required, then had to fix some logic to make sure the payload_to_model wasn't replacing nil with false, then fixed and added tests. 

Please test in browser 
- adminlocal:3000 for review, admin.stg.move.mil for acceptance 
- Go to the userEdit page for each of the following: Admin Users, Office Users, and Users
- Edit fields are not related to `active`
- Check in dev tools that the payload sent does not include active. 
- Make sure the `active` field does NOT change after you save the edit.
- Then edit the `active` field, and make sure it does save. 

## References

* [Jira story](https://dp3.atlassian.net/browse/mb-8239) for this change
* [User Management](https://github.com/transcom/mymove/wiki/User-Management)


pkg/handlers/adminapi/users_test.go
```diff
@@ -212,79 +213,72 @@ func (suite *HandlerSuite) TestIndexUsersHandler() {
...
-
-	revokeMilSession := true
-	revokeAdminSession := false
-	revokeOfficeSession := true
-	active := true
-	deactivate := false
-	revoke := true
-
-	params := userop.UpdateUserParams{
-		HTTPRequest: req,
```
shimonacarvalho, line 249: params should always go in the sub function since they are specific to each test

pkg/handlers/adminapi/users_test.go
```diff
@@ -613,8 +585,8 @@ func (suite *HandlerSuite) TestUpdateUserHandler() {
 		).Return(nil, err, nil).Once()
 
 		userUpdater.On("UpdateUser",
-			mock.Anything,
-			mock.Anything,
+			userID,
+			mock.AnythingOfType("*models.User"),
 		).Return(nil, nil, err).Once()
```
shimonacarvalho, line 590: Better to not use mock.Anything

pkg/handlers/adminapi/payloads/payload_to_model.go
```diff
@@ -21,7 +21,9 @@ func UserModel(user *adminmessages.UserUpdatePayload, id uuid.UUID) (*models.Use
 		ID: uuid.FromStringOrNil(id.String()),
 	}
 
-	if user.Active != nil {
+	if user.Active == nil { // active status was nil in payload
```
shimonacarvalho, line 24: So there was something finicky to deal with here. 
The payload active status could be nil, false and true. 
The model active status could false or true. It cannot be nil. 
So if the payload was nil, there was no way to set the model status because it *should* be set to nil but can't. 
So if the payload is nil, I need the original unchanged status from the db, and I set the model to that.

pkg/handlers/adminapi/users.go
```diff
@@ -133,30 +133,39 @@ func (h UpdateUserHandler) Handle(params userop.UpdateUserParams) middleware.Res
...
+	if err == nil {
+		err = h.DB().Find(&dbUser, userID)
+	}
 	if err != nil {
-		logger.Error(fmt.Sprintf("UUID Parsing for %s", params.UserID.String()), zap.Error(err))
+		err = fmt.Errorf(fmt.Sprintf("No user found for ID: %s", params.UserID.String()))
+		logger.Error("updateUserHandler Error", zap.String("error", err.Error()))
+		return userop.NewUpdateUserNotFound()
 	}
+
```
nelztruss, line 147: I believe I see that you are trying to do ONE error handling for two different classes of errors (e.g. incorrect format of ID; ID not found in the DB), but I think that makes this section of code more onerous than just handling each problem separately.

By rewriting the error, you are losing the contextual information from the cause of the error, potentially blinding someone from debugging an underlying issue here.

You don't need to do a `zap.String(string, string)`, there's a native `zap.Error(error)` that does the work for you.

pkg/handlers/adminapi/users.go
```diff
@@ -133,30 +133,39 @@ func (h UpdateUserHandler) Handle(params userop.UpdateUserParams) middleware.Res
...
 	// Currently, only updating the Active property is supported.
 	// If you want to add support for additional properties, edit UpdateUser.
-	user, verrs := payloads.UserModel(payload, userID)
-
-	if err != nil {
-		logger.Error(fmt.Sprintf("User Model Parsing for %s", params.UserID.String()), zap.Error(verrs))
+	// Also we need to retrieve the user's original status from the db to correctly create the update model
+	user, verrs := payloads.UserModel(payload, userID, dbUser.Active)
+	if verrs != nil && verrs.HasAny() {
+		err = fmt.Errorf(fmt.Sprintf("Could not parse payload: %s", verrs.String()))
```
nelztruss, line 154: Hrm... An `Errorf` wrapped around a `Sprintf` seems unnecessary here, they achieve basically the same thing... Also, if you use the `%w` (wrap) format directive, you don't need to turn `verrs` into a string, you can just wrap the error itself.

nelztruss (commented):
A couple of non-blocking nits around error handling.

pkg/handlers/adminapi/users.go
shimonacarvalho: Oh did not know that %w would pull out what's in verrs? Will try it.

pkg/handlers/adminapi/users.go
shimonacarvalho: True... I'll split it out. Usually we do more fancy error handling with a full payload of details, title, instance, which is why I use this pattern. But then I realized admin code doesn't use this type of error handling.

Good point on the zap fields. Thanks!

pkg/handlers/adminapi/users.go
shimonacarvalho: Oh wait, did you think verrs is of type error? It's actually an array of fieldname/description strings for validation errors so I'm not sure %w works on it. We use that when we want to return multiple errors. However since the underlying function only returns one error - I suppose I could turn it into an error versus a verrs.

Here's what I get if I just use %w
```
"error": "Could not parse payload: %!!(MISSING)w(*validate.Errors=&{map[User:[payload is nil]] 0xc0026b8198})"}
```

nelztruss: Oh whoa. Yupperz, I sure did make that assumption!

nelztruss (approved):
LGTM!

PR #5939, opened 2021-02-19
[MB-6753] Removed referenceID from createMoveTaskOrder

shimonacarvalho:
## Description

During load testing we found we can set the referenceID to anything. This shouldn't be the case, referenceID is supposed to be generated by the server and fits a pattern `dddd-dddd`. So this feature of the createMoveTaskOrder endpoint breaks other parts of the flow. This is a fix. 

Changed support api endpoint `createMoveTaskOrder` to make this field `readOnly`. And then made sure we discard the value that gets sent in in favor of a generated ID.

## Setup

Create a move in Postman using createMoveTaskOrder, send in a referenceID, result should not use what you sent but have a valid referenceID.

Also tests were updated, you can run with:

```sh
go test ./pkg/handlers/supportapi/ --testify.m TestCreateMoveTaskOrderRequestHandler -v
```

## References

* [Jira story](https://dp3.atlassian.net/browse/mb-6753) for this change
* [Postman Setup](https://github.com/transcom/mymove/wiki/setup-postman-to-make-mutual-tls-api-calls) for mtls calls
* [Support Redoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/transcom/mymove/master/swagger/support.yaml#operation/updateMoveTaskOrderStatus) in master
* [Updated Support Redoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/transcom/mymove/718124137dd139b7dd4cd6601084039c88f13fb6/swagger/support.yaml#operation/createMoveTaskOrder) for this branch

PR #6765, opened 2021-06-08
[MB-3557] SupportAPI: Unneeded ReferenceID on creation of move task order

Ronolibert:
## Description

Removes test checking for ReferenceID that is not used since it is automatically generated

## Setup
`make server_test`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3557) for this change

### DIFF
--- a/pkg/gen/supportapi/embedded_spec.go
+++ b/pkg/gen/supportapi/embedded_spec.go
@@ -1804,7 +1804,7 @@ func init() {
           ]
         },
         "referenceId": {
-          "description": "Unique ID associated with this Order.\n\nNo two MoveTaskOrders may have the same ID.\nAttempting to create a MoveTaskOrder may fail if this referenceId has been used already.\n",
+          "description": "Unique ID associated with this Order. Generated upon creation of a MoveTaskOrder\n\nNo two MoveTaskOrders may have the same ID.\n",
           "type": "string",
           "readOnly": true,
           "example": "1001-3456"
@@ -4505,7 +4505,7 @@ func init() {
           ]
         },
         "referenceId": {
-          "description": "Unique ID associated with this Order.\n\nNo two MoveTaskOrders may have the same ID.\nAttempting to create a MoveTaskOrder may fail if this referenceId has been used already.\n",
+          "description": "Unique ID associated with this Order. Generated upon creation of a MoveTaskOrder\n\nNo two MoveTaskOrders may have the same ID.\n",
           "type": "string",
           "readOnly": true,
           "example": "1001-3456"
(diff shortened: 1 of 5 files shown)


YanZ777 (commented):
The tests ran and passed for me, but someone else should also review this.

monfresh:
Should we add `readOnly: true` to the other places `referenceID` is defined in `prime.yaml` and `ghc.yaml`?

monfresh:
After looking at this more closely, I can understand why that test was added. It's because the `post` endpoint is referencing the same `MoveTaskOrder` definition that is reused in the `GET` request as well. And even though we want to return the `referenceID` in GET requests, we don't want to allow passing it in when creating a Move, so we set it to `readOnly: true`. However, it looks like Swagger doesn't enforce that. On line 51 of `/pkg/services/support/move_task_order/move_task_order_creator.go`, I added this line:
```go
fmt.Println("payload.ReferenceID:", payload.ReferenceID)
```
and then ran the test that you deleted in this PR, and I can see "some terrible reference id" that was passed in the Payload.

So, what we're really testing here is that the `InternalCreateMoveTaskOrder` ignores the reference ID in the payload

I am surprised that `readOnly: true` doesn't do anything. Did we think that it did?

In general, I prefer to be explicit in the swagger definitions, and instead of reusing Model definitions, I would rather duplicate some items but make it clear which fields can be passed in. But that's beyond the scope of this PR.

So, for now, maybe we should keep the test?

Pinging @shimonacarvalho who last touched this test.

monfresh:
Yeah, it looks like this test was added because we were actually using the reference ID from the payload instead of generating it ourselves. Here's the PR Shimona opened about 3 months ago: https://github.com/transcom/mymove/pull/5939/files

I think Shimona updated this test after I wrote the original JIRA ticket. The way the test was originally written, it seemed like we didn't need it, but now that it's testing something that was broken, it seems valid.

Having said that, we don't have similar tests for other read-only fields, like the Move Locator. I think it's easier to test fields that are not allowed by specifying a specific set of allowed fields only, and then if we try to pass in something that's not allowed in the Body, then we can use the Body validation and expect that it returns an error:

```
```

Ronolibert:
@monfresh I think intuitively the expectation was that `readOnly` would give some sort of validation error. You're right though in that in other tests we aren't really testing fields that are provided but never used. I can see it going either way

monfresh:
It turns out it was a bug in `go-swagger`, which has since been fixed I think 6 months ago, but we're still on an older version. Hopefully, we'll upgrade soon and we should be able to see an error now.

PR #6768, opened 2021-06-08
MB-8247 add orders link for orders amendment flow (front end)

pearl-truss:
## Description

This PR adds a new screen that includes an "upload documents" link (that currently doesn't go anywhere). This is shown under the Orders step after a customer submits their move

## Reviewer Notes
Previously the edit button was set to a width of 42px which worked when the button's label was "edit" however this component allows for the button label to be anything and that strict 42px width forced wrapping for longer labels like "upload documents". So I switch the width to auto for the amended orders step. 

## Setup

There are 2 ways to view these changes:

### Storybook
```sh
make storybook
```
And go to Customer Components -> Step -> Orders Amended

### Customer App
```sh
make server_run && make client_run
```
Create a customer move and after submitting the move verify the UI changes

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8247) for this change
*  [Mobile designs](https://app.zeplin.io/project/5f7cbab735d969035eb65339/screen/609d62985cd95a5ac6506697) for this change
*  [desktop designs](https://app.zeplin.io/project/5f7cbab735d969035eb65339/screen/5fbc14a28b1d51555d1114d1) for this change

## Screenshots

Screen Shot 2021-06-08 at 5 09 33 PM
Screen Shot 2021-06-08 at 5 09 41 PM


amanda5iu (approved):
Looks good to me!

reggieriser (approved):
LGTM 🚀 -- both storybook and running through a move went as expected.  Code looks good too, but I may not be up on all the latest front-end patterns.  Probably want to get a look from someone more recently in the front-end.

Also looks like Danger is complaining like yesterday -- hopefully a merge with master gets past that for now.

src/components/Customer/Home/Step/Step.module.scss
```diff
@@ -18,6 +29,7 @@
   }
 }
 
+
 .accept {
```
pearl-truss, line 36: @jenniferlynparsons out of curiosity with styes like .accept and .edit-btn that aren't prefixed with step or within .step-container, are these being applied everywhere they appear (beyond this component) or is there something happening to scope them to the Step component?

src/pages/MyMove/Home/index.jsx
```diff
@@ -279,24 +279,45 @@ export class Home extends Component {
...
-                    ) : (
-                      <Description>Upload photos of each page, or upload a PDF.</Description>
-                    )}
+                    <Description>Make sure to keep your personal information up to date during your move.</Description>
                   </Step>
+                  {!this.hasSubmittedMove && (
+                    <Step
+                      complete={this.hasOrders}
+                      completedHeaderText="Orders uploaded"
+                      editBtnLabel={this.hasOrders && !this.hasSubmittedMove ? 'Edit' : ''}
```
mboren, line 288: I think including `!this.hasSubmittedMove` in this condition might be redundant, and could be removed. On the other hand, it's not wrong and might make the code clearer, so I don't feel strongly about it.

src/components/Customer/Home/Step/index.jsx
```diff
@@ -39,8 +39,10 @@ const Step = ({
     secondaryBtnClassName,
   );
 
+  const containerCN = classnames(styles['step-container'], styles[containerClassName]);
```
jenniferlynparsons, line 42: No need to change this. We don't follow this pattern anywhere else in the code except when we have to do some kind of logic on the classnames.

src/components/Customer/Home/Step/Step.module.scss
jenniferlynparsons: Yup! It's still getting scoped. If you look in the inspector you'll see things like `Step_accept__bjZEP` as a class name.

jenniferlynparsons (changes requested):
It's looking good! I just made one small request to keep things consistent.

src/components/Customer/Home/Step/index.jsx
pearl-truss: I tried adding a className to this div and wasn't getting picked up until I made this change. I can try again

src/pages/MyMove/Home/index.jsx
pearl-truss: Good catch, I think I should just remove it

YanZ777 (approved):
Looks good to me.

src/components/Customer/Home/Step/index.jsx
pearl-truss: @jenniferlynparsons whenever you have time can you check the current diff? I reverted the classname changes here but now the .step-amended-orders styles aren't getting picked up and I'm not sure if there's something happening in the background I'm not familiar with

jenniferlynparsons: Okay! I think I know where the issue is, but fiddled with it and couldn't solve it. Might be worth reaching out to Shauna and even Suz (because it's a CSS Modules thing). There's something I'm not grokking about how this works.

If you look at the rendered class names, you'll see `Step_step-container__IlFLr step-amended-orders` which I am pretty sure means that `step-amended-orders` is outside of the CSS module control. I don't have a full grasp on this and couldn't find the right similar scenario anywhere else. 

I've got some other things on my plate, but I'll try to come back to this asap.

pearl-truss: Thanks for checking! @Shkeating @suzubara if you have time to look into this and/or know what's happening with the scss modules please let us know! Essentially I'm passing in 'step-amended-orders' into the `containerName` prop on the Step component and it's not getting picked up. I was able to get it to work by including it in the classnames function: ` classnames(styles['step-container'], styles[containerClassName])` however this doesn't match other patterns in the frontend like JP pointed out earlier

jenniferlynparsons (approved):

PR #6776, opened 2021-06-09
[MB-7683] TOO can send request for shipment diversion to GHC Prime

hannaliebl:
## Description

* Add Request Diversion button to shipments on MTO page
* Utilize existing `mutateMTOShipmentStatus` and `setMessage` setup to change the status of a shipment to `DIVERSION_REQUESTED` and set the proper flash message as well as add diversion requested tag to shipment.

## Reviewer Notes
I need to add some tests, but I have questions, so I thought I would put this up a bit early.

~There is weirdness when there is more than one shipment: the list re-orders, I think because another GET request gets made after update (possibly from clearing the query cache somehow? I'm not really sure.) I don't exactly know how to handle this - I'm curious whether there is already a set sort order on the backend that can be updated, maybe? Is there another example of us handling this?~

~The flash message also disappears immediately when there is more than one shipment. I'm not sure why, but maybe someone more familiar can help me out. It uses the old Redux flash technique that removes the message on `focusout`, but I find it odd that it happens when there are multiple shipments but not when there is only one.~

EDIT: I think I fixed the above in https://github.com/transcom/mymove/pull/6776/commits/3e5c1abf64510f7fb1b48ce9a28f1af2eeb1ead4 because now the GET request orders the shipments (same way it is done on the internal API.)

## Setup

Setup the app as usual, login as TOO, go to any move with shipments, approve them if need be, then on MTO page request diversion for a shipment. You should see the successful flash message and the diversion requested tag should show up on the shipments.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7683) for this change


src/components/Office/ShipmentAddresses/ShipmentAddresses.module.scss
```diff
@@ -1,3 +1,11 @@
 .mtoShipmentAddresses td {
   vertical-align: top;
 }
+
+.rightAlignButton {
```
jenniferlynparsons, line 5: Would it make sense to name this `rightAlignButtonWrapper` since it's not the button style itself, but the flex wrapper around it?

We've done this kind of thing enough that we might want to write a ticket for a mixin and refactor.

jenniferlynparsons (approved):
Looks good, just one suggested change, but it's optional.

It also would be good to check with @transcom/truss-design to see if the "request diversion" button should be hidden if the shipment is in a "diversion requested" state. Especially since clicking it a second time does nothing.

hannaliebl:
@jenniferlynparsons ah good idea, the ticket says the link will always render, but it is easy to hide so I could definitely do it here...I will wait to see what design says.

amanda5iu (commented):
Can we change the label to sentence case? `Request diversion`. Other than that it's looking good!

hannaliebl:
@amanda5iu I've updated that!

amanda5iu (approved):

PR #6780, opened 2021-06-10
MB-8148 Show roles not deleted

donaldthai:
## Description

This PR fixes a bug that the soft deleted roles were being populated in the session. The soft deleted roles don't show anymore after logging out and back in.

## Reviewer Notes

- Caveat is that the deleted roles will still be in the session while the user is still logged in. After logging out and back in, the user roles will be properly populated.
- The model `Role` being used in the `User` model is actually suppose to use `UsersRoles`. `Role` is meant to reference the actual roles we have that exists while the `UsersRoles` references what roles the user is tied to. I'll make a new story to capture this.
- There's a custom assertion condition logic to check if roles match the fetched list in the test. The assertion doesn't check for duplication but that shouldn't be a problem since the list of roles are unique. Something to keep note of while reviewing.

## Setup

```sh
make db_dev_e2e_populate
make server_run
make office_client_run
```

### Test with local sign-in (works with Login.gov too!)
1. Open up both the office app and admin app in separate tabs

**Office app (http://officelocal:3000)**

2. In the office app, log in as a multi-role user `too_tio_services_counselor_role@office.mil`
3. Click `Change user role` link at the top left corner
4. Note the current roles: `TOO`, `TIO`, and `Services Counselor`

**Admin app (http://adminlocal:3000)**

5. Switch to the admin app and log in as the admin
6. Navigate to office users queue by clicking on the left tab `Office Users`
7. Search office user `too_tio_services_counselor_role@office.mil`
8. Click on user
9. Click `Edit`
10. De-select any user role that's checked

**Office app (http://officelocal:3000)**

11. Back to Office app tab
12. Log out
13. Log back in with same user
14. Click on `Change user role` and note the difference 

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8148) for this change

### DIFF
--- a/pkg/models/user.go
+++ b/pkg/models/user.go
@@ -177,7 +177,14 @@ func FetchUserIdentity(db *pop.Connection, loginGovID string) (*UserIdentity, er
 		return nil, ErrFetchNotFound
 	}
 	identity := &identities[0]
-	roleError := db.Load(identity, "Roles")
+	roleError := db.RawQuery(`SELECT roles.created_at,
+									   roles.id,
+									   roles.role_name,
+									   roles.role_type,
+									   roles.updated_at
+									FROM roles AS roles
+									WHERE id in (select role_id from users_roles
+										where deleted_at is null and user_id = ?)`, identity.ID).All(&identity.Roles)
 	if roleError != nil {
 		return nil, roleError
 	}
--- a/pkg/models/user.go
+++ b/pkg/models/user.go
@@ -177,8 +177,7 @@ func FetchUserIdentity(db *pop.Connection, loginGovID string) (*UserIdentity, er
 		return nil, ErrFetchNotFound
 	}
 	identity := &identities[0]
-	roleError := db.RawQuery(`SELECT *
-									FROM roles AS roles
+	roleError := db.RawQuery(`SELECT * FROM roles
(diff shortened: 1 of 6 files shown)


Ryan-Koch (approved):

pkg/models/user.go
```diff
@@ -177,7 +177,14 @@ func FetchUserIdentity(db *pop.Connection, loginGovID string) (*UserIdentity, er
...
 	identity := &identities[0]
-	roleError := db.Load(identity, "Roles")
+	roleError := db.RawQuery(`SELECT roles.created_at,
+									   roles.id,
+									   roles.role_name,
+									   roles.role_type,
+									   roles.updated_at
+									FROM roles AS roles
+									WHERE id in (select role_id from users_roles
+										where deleted_at is null and user_id = ?)`, identity.ID).All(&identity.Roles)
```
monfresh, line 187: Minor nit: since we're selecting all the fields from `roles`, can we simplify the first part to this?
```
SELECT * FROM roles WHERE id ...
```

And if we don't actually need all the fields, can we only select the ones we want?

monfresh (approved):
Thanks for fixing this and for adding the test. This looks great!

PR #6788, opened 2021-06-11
[MB-8331] Create page for uploading amended orders

mboren:
## Description

I added a page to upload amended orders and linked it to the "upload documents" button on the home screen.
I temporarily hooked the uploaded files for this page to the existing orders, so you will see a table of orders that were already uploaded and will be able to upload/delete files. The final version of this page *will not* have access to previously uploaded orders files, I just did this because the backend functionality we need to do amended orders is coming next week, and I wanted to get the props and stuff for this page mostly figured out.

I made the `Save` and `Cancel` buttons take you back to the home screen, but otherwise they have no effect.

I've commented the places that we'll need to update when backend changes are ready.
 
## Reviewer Notes
The uploaded file table is broken on mobile, and we have the work to fix that captured in another ticket.

If it's better to have the file uploader do nothing at all instead of showing the wrong orders as a placeholder, like it does now, I'm happy to take that logic out.

This is the first frontend-only story I've done, ever, and this is the only React code I've written outside of a tutorial I worked through years ago, so any tips are appreciated.

## Setup

Create a move through the customer interface and submit it, then click the "Upload documents" button on the homepage.
It should take you to a page that matches the screenshots below.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8331) for this change

## Screenshots

Here's a screenshot of the new page:
image

The table that overflows in this view is going to be fixed in [MB-8101](https://dp3.atlassian.net/browse/MB-8101)
image


pearl-truss:
I haven't looked at your code yet but from reading your note about the links/buttons I think you can reuse the WizardNavigation component. It looks like theres an "edit mode" that'll give you a save button and a cancel link

Screen Shot 2021-06-11 at 9 32 48 AM

mboren:
> I think you can reuse the WizardNavigation component. 

nice, I'll give that a shot

src/pages/MyMove/AmendOrders/AmendOrders.test.jsx
```diff
@@ -0,0 +1,79 @@
+import { React } from 'react';
```
YanZ777, line 1: I the AmendOrders.module.scss file is empty, it should be deleted.

YanZ777:
You might want to note in the description that mobile view is going to be off. From what I understood, there's another ticket to work on the styling for the uploads table, correct?

src/pages/MyMove/AmendOrders/AmendOrders.jsx
```diff
@@ -0,0 +1,138 @@
...
+  };
+
+  const handleSave = () => {
+    // push(generalRoutes.HOME_PATH);
+  };
+  const handleCancel = () => {
+    // push(generalRoutes.HOME_PATH);
+  };
+
+  useEffect(() => {
```
YanZ777, line 58: I believe this is causing an infinite loop. I think you only want to do this once on mount, so you'll pass an empty array as the second argument to useEffect.

YanZ777 (changes requested):
Please fix the infinite requests. 
Also, does this need to be branched off of `pm-mb-8247-add-orders-link-added-front-end`?

src/pages/MyMove/AmendOrders/AmendOrders.jsx
mboren: eslint won't let me pass in an empty array because I'm using `serviceMemberId` and `updateOrders` in `useEffect`. I could just pass those as dependencies, but from what I'm reading, it seems like using a function in the dependencies array will not work the way I want it to. Still looking into this.

mboren: I tried to just adding `serviceMemberId` and `updateOrders` to the deps array and it seems to have fixed the infinite loop. From reading up about this, I was under the impression that using `updateOrders` as a dep would not work due to function comparison stuff, but it seems fine? I don't really understand this.

YanZ777: At quick glance, I would have thought that adding `updateOrders` would have caused an infinite rendering loop as well, but when I look at the code again, it wouldn't because `updateOrders` is passed in a prop and not defined within this component itself. 
If it were defined in this component, then adding it as a dependency would cause the component to re-render, which would cause the component to re-define the function, which would trigger this useEffect, and cause another re-render, etc.

Could you be more specific about which aspect of function comparison would not work?

src/pages/MyMove/AmendOrders/AmendOrders.jsx
```diff
@@ -0,0 +1,138 @@
...
+    </GridContainer>
+  );
+};
+
+AmendOrders.propTypes = {
+  serviceMemberId: PropTypes.string.isRequired,
+  updateOrders: PropTypes.func.isRequired,
+  currentOrders: OrdersShape,
+  uploads: UploadsShape,
+  // push: PropTypes.func.isRequired,
```
YanZ777, line 111: I think you might get push for free if you call `useHistory()` from within the component. 
```
import { useHistory } from 'react-router-dom';
...
const { history } = useHistory();
history.push(somePath);
...
```

src/pages/MyMove/AmendOrders/AmendOrders.jsx
mboren: I think your explanation makes sense. It makes sense to me that a function defined inside the component would cause a loop.

src/pages/MyMove/AmendOrders/AmendOrders.jsx
mboren: Oh cool. I saw push passed in as a prop to some of our other components, and assumed that was necessary for testing or something.

YanZ777 (approved):
It looks good to me, but I'd like someone else to take a look, if possible.

src/pages/MyMove/AmendOrders/AmendOrders.test.jsx
```diff
@@ -0,0 +1,79 @@
...
+  };
+
+  beforeEach(() => {
+    getOrdersForServiceMember.mockImplementation(() => Promise.resolve(testOrdersValues));
+  });
+
+  it('loads orders on mount', async () => {
+    const { queryByText, queryByRole } = render(<AmendOrders {...testProps} />);
+
+    expect(queryByText('Loading, please wait...')).toBeInTheDocument();
```
kylehilltruss, line 54: We don't have this pattern in too many places yet, but the best practice with React Testing Library is to use `getByWhatever` or `findByWhatever`, rather than `queryByWhatever`, when you expect something to be in the document. 

`queryBy` is just intended to be used for checking that something _isn't_ in the document, since the other two accessors throw errors when they don't find something while query returns `null`.

See: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence

src/pages/MyMove/AmendOrders/AmendOrders.test.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+    await waitFor(() => {
+      expect(queryByText('Loading, please wait...')).not.toBeInTheDocument();
```
kylehilltruss, line 57: Generally, we don't want to use `waitFor` anymore either (again, this isn't a common pattern in our code base yet). You can replace this as:

```js
// findByWhatever throws an error if something isn't found
// so we want to move the awaited positive assertion to the top
expect(await findByRole('heading', { name: 'Upload orders', level: 5 }))).toBeInTheDocument()

// Per above, use queryBy when something _shouldn't_ be there
expect(queryByText('Loading, please wait...')).not.toBeInTheDocument()

// Other assertions in the waitFor go here
```

See, in the same blog post as above: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find

src/pages/MyMove/AmendOrders/AmendOrders.test.jsx
```diff
@@ -0,0 +1,79 @@
...
+      expect(getOrdersForServiceMember).toHaveBeenCalled();
+      expect(testProps.updateOrders).toHaveBeenCalledWith(testOrdersValues);
+    });
+  });
+
+  it('renders the save button', async () => {
+    const { queryByText } = render(<AmendOrders {...testProps} uploads={[]} />);
+
+    await waitFor(() => {
+      expect(queryByText('Save')).toBeInTheDocument();
```
kylehilltruss, line 69: This (and the cancel button test) can just be simplified down to

```js
expect(await findByText('Save')).toBeInTheDocument()
```

kylehilltruss (commented):
I don't see any blocking problems with the code, leaving aside the uploaded file table overflow. I've left comments in the tests just because the exact same comments were useful for me this time last month :).

You have a mixture of `.then`-based promises in the main component logic, and `await`s in the test file; is there a reason for this, or just is a function of the code that you were adapting from?

This is definitely the kind of component that would be well-served by adding in a storybook entry, both for reference and for speeding up the development feedback cycle.

src/pages/MyMove/AmendOrders/AmendOrders.test.jsx
mboren: Thanks, this is helpful! I was very much cargo culting here.

mboren: I haven't been able to successfully replace queryBy with findBy or getBy. I'll see if I can get some help with this on Monday.

src/pages/MyMove/AmendOrders/AmendOrders.test.jsx
mboren: Gotcha. I've gotten rid of all the `waitFor`s.

src/pages/MyMove/AmendOrders/AmendOrders.test.jsx
mboren: Thanks!

src/pages/MyMove/AmendOrders/AmendOrders.test.jsx
mboren: I got it! I think I was just misunderstanding the difference between findBy and getBy.

kylehilltruss (approved):

PR #6805, opened 2021-06-14
Update shipment tests

monfresh:
## Description

Small updates to tests based on PR feedback in #6767.

One thing that I just discovered and that was surprising to me is that using `.On` when mocking doesn't test that the mocked object was actually called in the case where it doesn't get called at all. For example, the ShipmentApprover service calls the ShipmentRouter service. We expect `ShipmentRouter` to receive a call to `Approve`. If in the test, we do this:

```golang
shipmentRouter.On("Approve", &createdShipment).Return(nil)
```
and then we comment out the call to `router.Approve` in the ShipmentApprover, the test still passes!

However, if the `ShipmentApprover` passes in the wrong object to `router.Approve`, the test will catch that.

So, to make sure the call is indeed made, we can add another assertion after the service call is made:

```golang
shipmentRouter.AssertNumberOfCalls(t, "Approve", 1)
```

Testify also has a `AssertCalled` function, but I couldn't get it to work. It was complaining about the memory address of the shipment being wrong and I couldn't figure out why.

## Reviewer Notes

## Setup

```sh
echo "Code goes here"
```

## References

## Screenshots

### DIFF
--- a/pkg/services/mto_shipment/shipment_approver_test.go
+++ b/pkg/services/mto_shipment/shipment_approver_test.go
@@ -238,5 +238,6 @@ func (suite *MTOShipmentServiceSuite) TestApproveShipment() {
 		_, err := approver.ApproveShipment(shipment.ID, eTag)
 
 		suite.NoError(err)
+		shipmentRouter.AssertNumberOfCalls(t, "Approve", 1)
 	})
 }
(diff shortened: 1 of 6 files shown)


donaldthai (approved):
Looks great, Moncef!

For that different shipment memory address when you tried to test with `AssertCalled`, I think that's because in each of the shipment services it's fetching the shipment before calling the mocked function. So whatever shipment memory address you passed in the mock isn't the same as the one actually being used. What's interesting is that it seems the `.On` call doesn't really care about the memory address though based on your tests 😄 .

PR #6810, opened 2021-06-15
MB-3728 refactor validator pattern

nelztruss:
## Description

Use a slightly lighter-weight way to apply validation checks.

## Reviewer Notes

Much of the new "rules" are just re-statements of the previous logic in the newer format.


nelztruss:
NB: I usually try to make my PRs *much* smaller than 1600 lines of flux. 😱 Mea culpa. But a lot of the LoC are just moving/repackaging the logic in `rules.go` and its associated test.

pkg/services/mto_agent/rules.go
```diff
@@ -0,0 +1,156 @@
...
+var basicChecks = []mtoAgentValidator{
+	checkShipmentID(),
+	checkAgentID(),
+}
+
+var primeChecks = append(
+	basicChecks,
+	checkContactInfo(),
+	checkAgentType(),
+)
```
sandy-wright, line 23: Hmm, I don't like the fact that we can't include the `checkPrimeAvailability` function here because of the way we have to pass in the `mtoAvailabilityChecker`. It feels misleading to have a list of checks here that gets modified when we initialize the updater/creator. 

Maybe some kind of wrapper function would be better - 
```go
func primeChecks(checker *services.MoveTaskOrderChecker) []mtoAgentValidator {
    checks := append(
        basicChecks,
        checkContactInfo(),
        checkAgentType(),
    )
    if checker != nil {
        return append(checks, checkPrimeAvailability(checker))
    }
    return checks
}
```
? Then we would be more explicit about what's needed for Prime. 

I used a pointer so that it gives folks the option to skip the Prime-availability check if necessary (like if we want to check these rules from a shipment service object). That might also mean we want them to be exportable, but I'm fine with skipping this functionality until we actually have a concrete use-case

pkg/services/mto_agent/validation.go
```diff
@@ -0,0 +1,79 @@
...
+func (fn mtoAgentValidatorFunc) Validate(ctx context.Context, newer models.MTOAgent, older *models.MTOAgent, ship *models.MTOShipment) error {
+	return fn(ctx, newer, older, ship)
+}
+
+// validateMTOAgent checks an MTOAgent against a passed-in set of business rule checks
+// Validation errors are aggregated and returned to be reported on en masse. Other errors,
+// of types not specifically associated with validations, are treated with higher priority
+// and returned immediately, ignoring any accumulated validation errors and short circuiting
+// the execution of any further mtoAgentValidator instances.
+func validateMTOAgent(
```
sandy-wright, line 34: I kind of think this should be exportable. I would imagine places where we'd want to validate an object without doing the default update or create actions performed in the Updater/Creator. 

For example, if we're using the `CreateMTOShipment` service object to create a shipment with agents, we'd probably want to reuse this validation, but do the create action within a transaction in the shipment service object.

sandy-wright: ^ We don't have to do that now, it's mostly something to keep in mind

sandy-wright (commented):
Overall, this looks really good and is a solid refactor. I have a comment about the `primeChecks` function though - I don't like that we have to append to it when we initialize the Updater and Creator service objects, so it would be nice to think of a better solution there

pkg/services/mto_agent/rules.go
nelztruss: Yeah, I think I get what you mean about the "misleading" aspect.

Honestly, those package variables are really just helpers that don't really add a lot of value. So, I think I'll take a stab at getting rid of them, rather than building more structure around them. (This is fully exposing my internal reaction of "less code is more".)

pkg/services/mto_agent/validation.go
nelztruss: My gut reaction is to not get too "reusable" with this function. It's really not that large a cognitive load to re-implement, and it's hard to make it super generic (for other types). 

To me, I feel like exposing this would run afoul of the [Go Proverb](https://go-proverbs.github.io/): "A little copying is better than a little dependency."

pkg/services/mto_agent/rules.go
sandy-wright: Ooh, I do really like that idea

pkg/services/mto_agent/validation.go
sandy-wright: That's totally fair imo, plus we _don't_ actually have this use-case yet, so no need to overcode for a possibility

pkg/services/mto_agent/mto_agent_creator.go
```diff
@@ -1,55 +1,69 @@
...
+			checkShipmentID(),
+			checkAgentID(),
+		},
+		primeChecks: []mtoAgentValidator{
+			checkShipmentID(),
+			checkAgentID(),
+			checkContactInfo(),
+			checkAgentType(),
+			checkPrimeAvailability(mtoAvailabilityChecker),
+		},
```
sandy-wright, line 34: I do like this, I think this is a good solution to the misleading constant/not constant problem. Plus it leaves room for different rules for different actions (if needed)

pkg/services/mto_agent/mto_agent_creator.go
```diff
@@ -1,55 +1,69 @@
...
-	mtoAgent, err = mtoagentvalidate.ValidateAgent(&agentData, validatorKey)
+	err = validateMTOAgent(context.TODO(), *mtoAgent, nil, mtoShipment, checks...)
 	if err != nil {
 		return nil, err
 	}
 
+	// TODO: this basically operates as a way to make a (shallow?) copy of the
+	// mtoAgent (because pass-by-value parameter); haven't dug into
+	// why this is necessary
+	mtoAgent = mergeAgent(*mtoAgent, nil)
```
sandy-wright, line 65: Alright, this was bugging the heck out of me, so I looked into it! I _think_ I get why this causes a problem when removed now, and hopefully I can explain it well enough.

This step creates a copy of the agent that was passed in. It is now a new object with a new address in the stack. If we _don't_ do this, the original object with the _original_ address is the one that gets modified. This means that our service object will no longer be idempotent because the input was modified during the create, but otherwise, you wouldn't think this would cause a problem.

However, the root of the problem is the tests themselves and the fact that they share input objects. This means in one test, we'll pass in an agent for a successful create, that agent will be modified by the service object because it will now get an ID and createdAt/updateAt timestamps, and then the next test will try to use the same agent object that was now updated. 

Because we check the agent ID to make sure we are in fact trying to _create_ and not _update_, the next test will fail with an InvalidInputError.

So there are two solutions: 
1. Make sure none of the tests share input, instead making copies of agents when needed. Or nullifying the IDs in between each test.
2. Keep this line for the sake of ensuring idempotency. 

I think I would lean towards #2 (maybe both?) in this case, but it would definitely warrant a comment so that someone later on doesn't try to remove it.

nelztruss: Yah, that's one thing I've noticed in the existing code... There's a lot of "modify the contents of a passed in instance". Whereas I generally tend towards "pure" functions with few side-effects, e.g. can spit out a new thing but won't modify a passed in thing.

I *do* have some feels about "production code" making allowances because of "test code" escalating/inflating idempotency needs... But this currently conflicts with my other oft-touted maxim "tests are a first-class consumer of your production code", so I don't have great advice about this yet. (Hrm... Except... Well, generally I *do* believe tests should operate independently.)

I propose for now that I just keep the current comment, and maybe over time we'll get our patterns in line to tidy this up.

sandy-wright: Would a possible solution to this kind of issue be _not_ passing in a pointer as the input `mtoAgent`? That would have a number of ripple effects, so I'm not advocating for doing that now (I agree, it's fine to leave this comment for now), but just wondering what a better approach would look like

sandy-wright (approved):
Looks good!

pkg/services/mto_agent/mto_agent_creator.go
nelztruss: NB: Sandy and I took this conversation offline/f2f to brainstorm about future patterns.

PR #6811, opened 2021-06-15
[MB-7680] TOO can identify newest orders uploaded in TOO UI

jenniferlynparsons:
## Description

TOO sees the creation date for the uploaded orders on the document viewer, sorted by creation date.

## Reviewer Notes

The Go tests are broken and I'm not sure why. I've added a second document creation call to the seed data generator functions so that's where the culprit might be.

## Setup

```sh
make server_run
make office_client_run

// Some changes are visible with
make storybook
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7680) for this change

## Screenshots
Screen Shot 2021-06-16 at 2 35 43 PM
Screen Shot 2021-06-16 at 2 35 57 PM

### DIFF
--- a/pkg/testdatagen/make_order.go
+++ b/pkg/testdatagen/make_order.go
@@ -42,14 +42,6 @@ func MakeOrder(db *pop.Connection, assertions Assertions) models.Order {
 			UserUploader: assertions.UserUploader,
 		})
 		document.UserUploads = append(document.UserUploads, u)
-		u2 := MakeUserUpload(db, Assertions{
-			UserUpload: models.UserUpload{
-				DocumentID: &document.ID,
-				Document:   document,
-				UploaderID: sm.UserID,
-			},
-		})
-		document.UserUploads = append(document.UserUploads, u2)
 	}
 
 	defaultOrderNumber := "ORDER3"
--- a/pkg/testdatagen/make_order.go
+++ b/pkg/testdatagen/make_order.go
@@ -42,6 +42,14 @@ func MakeOrder(db *pop.Connection, assertions Assertions) models.Order {
 			UserUploader: assertions.UserUploader,
 		})
 		document.UserUploads = append(document.UserUploads, u)
+		u2 := MakeUserUpload(db, Assertions{
+			UserUpload: models.UserUpload{
+				DocumentID: &document.ID,
+				Document:   document,
+				UploaderID: sm.UserID,
+			},
+		})
+		document.UserUploads = append(document.UserUploads, u2)
 	}
 
 	defaultOrderNumber := "ORDER3"
--- a/pkg/testdatagen/scenario/e2ebasic.go
+++ b/pkg/testdatagen/scenario/e2ebasic.go
@@ -92,45 +92,6 @@ func createHHGNeedsServicesCounselingWithLocator(db *pop.Connection, locator str
 	})
 }
 
-func createHHGWithMultipleOrderUploads(db *pop.Connection) {
-	document := testdatagen.MakeDocument(db, testdatagen.Assertions{})
-	u := testdatagen.MakeUserUpload(db, testdatagen.Assertions{
-		UserUpload: models.UserUpload{
-			DocumentID: &document.ID,
-			Document:   document,
-		},
-	})
-	document.UserUploads = append(document.UserUploads, u)
-
-	u2 := testdatagen.MakeUserUpload(db, testdatagen.Assertions{
-		UserUpload: models.UserUpload{
-			DocumentID: &document.ID,
-			Document:   document,
-		},
-	})
-
-	document.UserUploads = append(document.UserUploads, u2)
-
-	ordersMU := testdatagen.MakeOrder(db, testdatagen.Assertions{
-		Order: models.Order{
-			UploadedOrders: models.Document{
-				UserUploads: document.UserUploads,
-			},
-		},
-	})
-
-	moveMU := testdatagen.MakeMove(db, testdatagen.Assertions{
-		Move: models.Move{
-			Locator: "MULTOR",
-		},
-		Order: ordersMU,
-	})
-
-	testdatagen.MakeMTOShipment(db, testdatagen.Assertions{
-		Move: moveMU,
-	})
-}
-
 // Run does that data load thing
 func (e e2eBasicScenario) Run(db *pop.Connection, userUploader *uploader.UserUploader, primeUploader *uploader.PrimeUploader, logger Logger) {
 	// Testdatagen factories will create new random duty stations so let's get the standard ones in the migrations
@@ -2398,5 +2359,4 @@ func (e e2eBasicScenario) Run(db *pop.Connection, userUploader *uploader.UserUpl
 	createHHGNeedsServicesCounselingWithLocator(db, "SCE2ET")
 	createHHGNeedsServicesCounselingWithLocator(db, "SCE3ET")
 	createHHGNeedsServicesCounselingWithLocator(db, "SCE4ET")
-	createHHGWithMultipleOrderUploads(db)
 }
--- a/pkg/testdatagen/scenario/e2ebasic.go
+++ b/pkg/testdatagen/scenario/e2ebasic.go
@@ -92,6 +92,45 @@ func createHHGNeedsServicesCounselingWithLocator(db *pop.Connection, locator str
 	})
 }
 
+func createHHGWithMultipleOrderUploads(db *pop.Connection) {
+	document := testdatagen.MakeDocument(db, testdatagen.Assertions{})
+	u := testdatagen.MakeUserUpload(db, testdatagen.Assertions{
+		UserUpload: models.UserUpload{
+			DocumentID: &document.ID,
+			Document:   document,
+		},
+	})
+	document.UserUploads = append(document.UserUploads, u)
+
+	u2 := testdatagen.MakeUserUpload(db, testdatagen.Assertions{
+		UserUpload: models.UserUpload{
+			DocumentID: &document.ID,
+			Document:   document,
+		},
+	})
+
+	document.UserUploads = append(document.UserUploads, u2)
+
+	ordersMU := testdatagen.MakeOrder(db, testdatagen.Assertions{
+		Order: models.Order{
+			UploadedOrders: models.Document{
+				UserUploads: document.UserUploads,
+			},
+		},
+	})
+
+	moveMU := testdatagen.MakeMove(db, testdatagen.Assertions{
+		Move: models.Move{
+			Locator: "MULTOR",
+		},
+		Order: ordersMU,
+	})
+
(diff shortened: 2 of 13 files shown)


pkg/testdatagen/make_order.go
```diff
@@ -42,6 +42,14 @@ func MakeOrder(db *pop.Connection, assertions Assertions) models.Order {
...
 		})
 		document.UserUploads = append(document.UserUploads, u)
+		u2 := MakeUserUpload(db, Assertions{
+			UserUpload: models.UserUpload{
+				DocumentID: &document.ID,
+				Document:   document,
+				UploaderID: sm.UserID,
+			},
+		})
+		document.UserUploads = append(document.UserUploads, u2)
```
monfresh, line 52: What do you think about moving this additional upload to a specific move in devseed? Otherwise, this will create an unnecessary second upload everywhere we call `MakeOrder` in our tests. We also probably want to modify the `CreatedAt` date to test the sort order.

monfresh: Even better, you don't have to add anything to `devseed` because we already have `bandwidth.go`, which creates multiple uploads. You can run `make db_bandwidth_up`, and it will create several moves with specific Move Codes, such as `S150KB`. You can see all the move codes inside the various calls to the `makeMoveForOrders` function in `bandwidth.go`.

Now that I look at it more closely, it looks like this was modified since I originally wrote it, and now each move only has one upload instead of a single move having multiple uploads. I don't remember why we changed this. In any case, it should be easy to add another file in the `filterFile` declaration, such as:

```golang
filterFile := &[]string{"150Kb.png", "2mb.png"}
```

The only thing missing is the different dates to test the sort order. I tried modifying the `CreatedAt` field for the `UserUpload` in `makeOrdersForServiceMember`, and I can see if being modified correctly, but when the script finishes running, the DB doesn't have the modified date for some reason.

jenniferlynparsons: This is an interesting option. Though I think adding it to the devseed is a little easier to then test that move specifically. I didn't write any move-specific tests so I don't know that it matters. The sorting is just a basic a basic `.sort` on the frontend so there's not much there to test. The creation date is probably more important to test.

monfresh: Heh. It looks like Pop always sets `CreatedAt` to `time.Now()`. It doesn't look like it can be modified. I tried a bunch of things and nothing worked. https://github.com/gobuffalo/pop/issues/591

src/components/DocumentViewer/Menu/Menu.jsx
```diff
@@ -10,35 +11,43 @@ import styles from './Menu.module.scss';
...
-const DocViewerMenu = ({ isOpen, files, handleClose, selectedFileIndex, handleSelectFile }) => (
-  <div data-testid="DocViewerMenu" className={classnames(styles.docViewerMenu, { [styles.collapsed]: !isOpen })}>
-    <div className={styles.menuHeader}>
-      <h3>Documents</h3>
-      <div className={styles.menuControls}>
-        <Button data-testid="closeMenu" type="button" onClick={handleClose} unstyled className={styles.menuClose}>
-          <FontAwesomeIcon icon="times" title="Close menu" aria-label="Close menu" />
-        </Button>
+const DocViewerMenu = ({ isOpen, files, handleClose, selectedFileIndex, handleSelectFile }) => {
+  const sortedFiles = files.sort((a, b) => moment(a.createdAt) - moment(b.createdAt));
```
duncan-truss, line 15: This looks like it sorts the files in the menu correctly but it will load the first file in the array that is passed into the DocumentViewer component and set that as the selectedFile regardless of the order the menu is sorted in.  You can see this by changing one of the created dates in your story.

You likely will either need to move the sort functionality into the DocumentViewer component or change all the page specific components to pass in the files in the sorted order they want displayed.

duncan-truss:
If we have a very long file name, we are never going to see the Added on date displayed.  Maybe we can change the html so that we always display the Added on date and only truncate the file name portion not the entire paragraph contents.

Screen Shot 2021-06-16 at 5 25 11 PM

pkg/testdatagen/scenario/e2ebasic.go
```diff
@@ -92,6 +92,45 @@ func createHHGNeedsServicesCounselingWithLocator(db *pop.Connection, locator str
 	})
 }
 
+func createHHGWithMultipleOrderUploads(db *pop.Connection) {
```
jenniferlynparsons, line 95: Adding this didn't break anything, but also didn't actually produce a move to test against. I'm not sure what information is missing from this move.

duncan-truss: This is the cypress seed file, I think you meant to add this to devseed scenario right?

donaldthai: Did you wanted to add this function in the `devseed.go` file instead?

jenniferlynparsons: YES! 🤦

duncan-truss: can't beat Donald 🙇

src/components/DocumentViewer/Menu/Menu.jsx
duncan-truss: On second look this doesn't seem to be showing the newest orders first.  I think your comparator function should return 1 if file b is newer than file a.  Can we just do a `moment(b.createdAt) > moment(a.createdAt)` comparison instead? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

src/components/DocumentViewer/DocumentViewer.test.jsx
```diff
@@ -14,28 +14,37 @@ import sampleGIF from './sample3.gif';
...
     contentType: 'image/png',
     url: samplePNG,
+    createdAt: '2021-06-15T15:09:26.979879Z',
   },
   {
+    id: 4,
     filename: 'Test File 4.gif',
     contentType: 'image/gif',
     url: sampleGIF,
+    createdAt: '2021-06-15T15:09:26.979879Z',
```
duncan-truss, line 47: I think we should make these createdAt dates unique so you can test the sorting here and see that the newest date is displayed first.

src/components/DocumentViewer/Menu/Menu.jsx
duncan-truss: It's not clear to me what the overloaded greater than operator would do against 2 moment objects.  Should we be more explicit and use the built in method https://momentjs.com/docs/#/query/is-before/

jenniferlynparsons: I like that much better! I'm going to update this in the couple other places this kind of sort is being used.

jenniferlynparsons: Scratch that. `.sort` requires a number returned from the callback so I'm leaving it as is.

duncan-truss: You could always do `(a, b) =>  moment(b.createdAt).isAfter(moment(a.createdAt)) ? 1 : 0`

src/components/DocumentViewer/DocumentViewer.test.jsx
```diff
@@ -125,6 +57,14 @@ describe('DocumentViewer component', () => {
     expect(menu.prop('files')).toBe(mockFiles);
   });
 
+  it('renders the file creation date with the correctly sorted props', () => {
+    expect(component.find('li button').at(0).text()).toBe('Test File 4.gif  Uploaded on 16-Jun-2021');
+  });
+
+  it('renders the title bar with the correct props', () => {
+    expect(component.find('[data-testid="documentTitle"]').text()).toBe('Test File 4.gif   - Added on 16 Jun 2021');
```
duncan-truss, line 65: What's with the extra whitespace in the title?

src/components/DocumentViewer/DocumentViewer.jsx
```diff
@@ -72,14 +75,17 @@ const DocumentViewer = ({ files }) => {
 
   const selectedFilename = filenameFromPath(selectedFile.filename);
 
+  const selectedFileDate = moment(selectedFile.createdAt).format('DD MMM YYYY');
```
duncan-truss, line 78: We likely have a formatter defined for this in src/shared/formatters.js

src/components/DocumentViewer/Menu/Menu.jsx
```diff
@@ -25,12 +26,16 @@ const DocViewerMenu = ({ isOpen, files, handleClose, selectedFileIndex, handleSe
         const itemClasses = classnames(styles.menuItemBtn, {
           [styles.active]: i === selectedFileIndex,
         });
-        const filename = filenameFromPath(file.filename);
+        const fileName = filenameFromPath(file.filename);
+        const fileDate = moment(file.createdAt).format('DD-MMM-YYYY');
```
duncan-truss, line 30: similar point here about an existing formatter

src/components/DocumentViewer/DocumentViewer.test.jsx
```diff
@@ -125,6 +57,14 @@ describe('DocumentViewer component', () => {
     expect(menu.prop('files')).toBe(mockFiles);
   });
 
+  it('renders the file creation date with the correctly sorted props', () => {
+    expect(component.find('li button').at(0).text()).toBe('Test File 4.gif  Uploaded on 16-Jun-2021');
```
duncan-truss, line 61: 👍

duncan-truss (approved):
Looks good!

src/components/DocumentViewer/Menu/Menu.jsx
jenniferlynparsons: Yeah, that would work. Does that make it easier to reason about? I think so. In fact, having it explicitly return a 1 or 0 makes a lot of sense rather than coercing the values and having a potentially negative number.

jenniferlynparsons: Wait, I was wrong in that thinking. Sort should be able to handle a negative number so I'm going to leave this as is. 

Less than 0: "a" is sorted to be a lower index than "b".
Zero: "a" and "b" are considered equal, and no sorting is performed.
Greater than 0: "b" is sorted to be a lower index than "a".

🎉

src/components/DocumentViewer/DocumentViewer.test.jsx
jenniferlynparsons: Spacing weirdness that I've now replaced with some CSS padding.

src/components/DocumentViewer/DocumentViewer.jsx
jenniferlynparsons: I guess the main reason to use the `formatDate` method in formatters is to provide default values? That makes sense.

samadden (approved):
LGTM

PR #6812, opened 2021-06-15
MB-8177 Update devseed data for multiple addresses

felipe-lee:
## Description

We want to be able to have a service member that has shipments with a variety of destination address combos.

## Reviewer Notes

I also created a wiki page with info on the test data generation. Could you please review that too? https://github.com/transcom/mymove/wiki/Test-Data-Generation

I updated `MakeMTOShipment` to default secondary addresses to be blank. This meant a few tests had to be updated to account for that change, but the reason I think it's worth it is so that we can have more control over the data, and hopefully also have more realistic data.

I also changed `MakeMTOShipment` to only accept the address assertions at the top level, like our docs say we should be doing it https://github.com/transcom/mymove/wiki/Understanding-Testdatagen-Functions#assertions-on-dependent-records. That necessitated a few changes to some tests to get them to follow the standard. There are other parts of `MakeMTOShipment` that used nested assertions, but I'm counting them as out of scope just because this is already getting bigger than I'd intended.

I tweaked the `Makefile` targets a bit to clean up some repeated bits. Lmk if any of it seems off. I also added more tables  for them to truncate because when re-running the commands to populate the databases, I was getting conflicts. Though, #6650 comes with a much  better truncation script so that'll make it nicer.

I also made a few updates to a few of the other devseed functions to clean up unused params and to stop a few from accidentally creating extra service members or shipments (from missing or misplaced assertions).

## Setup

1. Fill you DB with the data:
    1. Full reset and fill:

        ```sh
        make db_dev_fresh
        ```

    2. Or only truncate and fill:

        ```sh
        make db_dev_e2e_populate
        ```

2. Run the server and customer client:

    ```sh
    make server_run
    ```

    ```sh
    make client_run
    ```

3. Use the local log in to log in as `multple-destinations@unsubmitted.hhg`
4. Note that each of the shipments has a different combination of destination addresses.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8177) for this change


kylehilltruss (approved):
This is all looking good by me. And thank you for the _very_ clear [companion page](https://github.com/transcom/mymove/wiki/Test-Data-Generation) in the wiki!

YanZ777 (commented):
I've added @gidjin and @donaldthai to review since it looks like they know what they are talking about based in the corresponding slack thread.

PR #6822, opened 2021-06-16
[MB-8365] Adjust automatic service item creation/approval rules for diversion case

hannaliebl:
## Description

Given I'm a TOO using the milmove system
Then I need certain service items to be created without duplicates and in the approved
When I approve a diverted shipment

This also relates to https://dp3.atlassian.net/browse/MB-8310, which will actually allow the Prime contractor to get a move into this state. Reading that ticket might be helpful for understanding.

## Reviewer Notes

I could use help setting up a shipment with existing service items for the blank test.

## Setup

```sh
make db_dev_e2e_populate
make server_run
make client_run
```

Login to office app as TOO, look at the move with `EXSTS1` locater. Approve the shipment, and notice that there is just one service item on it: domestic line haul (which is correct, in this case, as it was the only existing service item and so no others were auto-created.)

Comment out the changes I made in `pkg/services/mto_shipment/mto_shipment_updater.go` and then re-populate the dev DB and go to the same move as above. Approve the shipment, but this time note that, incorrectly, there are now the default service items as well as the existing service item.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8365) for this change

### DIFF
--- a/pkg/services/mto_shipment/mto_shipment_updater_test.go
+++ b/pkg/services/mto_shipment/mto_shipment_updater_test.go
@@ -540,6 +540,10 @@ func (suite *MTOShipmentServiceSuite) TestUpdateMTOShipmentStatus() {
 		}
 	})
 
+	suite.T().Run("A shipment with existing service items does not create more", func(t *testing.T) {
+		// TODO
+	})
+
 	suite.T().Run("If we act on a shipment with a weight that has a 0 upper weight it should still work", func(t *testing.T) {
 		// This is testing that the Required Delivery Date is calculated correctly.
 		// In order for the Required Delivery Date to be calculated, the following conditions must be true:
--- a/pkg/services/mto_shipment/mto_shipment_updater_test.go
+++ b/pkg/services/mto_shipment/mto_shipment_updater_test.go
@@ -443,6 +443,25 @@ func (suite *MTOShipmentServiceSuite) TestUpdateMTOShipmentStatus() {
 			Status: models.MTOShipmentStatusRejected,
 		},
 	})
+
+	shipmentWithExistingServiceItem := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+		Move: mto,
+		MTOShipment: models.MTOShipment{
+			Status: models.MTOShipmentStatusSubmitted,
+		},
+	})
+	serviceItemID, _ := uuid.NewV4()
+	testdatagen.MakeMTOServiceItem(suite.DB(), testdatagen.Assertions{
+		MTOServiceItem: models.MTOServiceItem{
+			ID: serviceItemID,
+		},
+		Move:        mto,
+		MTOShipment: shipmentWithExistingServiceItem,
+		ReService: models.ReService{
+			Code: models.ReServiceCodeDLH,
+		},
+	})
+
 	shipment.Status = models.MTOShipmentStatusSubmitted
 	eTag := etag.GenerateEtag(shipment.UpdatedAt)
 	status := models.MTOShipmentStatusApproved
@@ -541,7 +560,19 @@ func (suite *MTOShipmentServiceSuite) TestUpdateMTOShipmentStatus() {
 	})
 
 	suite.T().Run("A shipment with existing service items does not create more", func(t *testing.T) {
-		// TODO
+		eTag = etag.GenerateEtag(shipmentWithExistingServiceItem.UpdatedAt)
+		fetchedShipment := models.MTOShipment{}
+		serviceItems := models.MTOServiceItems{}
+
+		_, err := updater.UpdateMTOShipmentStatus(shipmentWithExistingServiceItem.ID, status, nil, eTag)
+		suite.NoError(err)
+
+		err = suite.DB().Find(&fetchedShipment, shipmentWithExistingServiceItem.ID)
+		suite.NoError(err)
+
+		err = suite.DB().EagerPreload("ReService").Where("mto_shipment_id = ?", shipmentWithExistingServiceItem.ID).All(&serviceItems)
+		suite.NoError(err)
+		suite.Equal(1, len(serviceItems))
 	})
 
 	suite.T().Run("If we act on a shipment with a weight that has a 0 upper weight it should still work", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


hannaliebl:
@sandy-wright I tagged you because this relates to your ticket, I think (https://dp3.atlassian.net/browse/MB-8310) and maybe you'd like to look at the devseed scenario I wrote and see if it can be improved/changed based on your work?

pkg/services/mto_shipment/mto_shipment_updater_test.go
```diff
@@ -443,6 +443,25 @@ func (suite *MTOShipmentServiceSuite) TestUpdateMTOShipmentStatus() {
...
+		MTOServiceItem: models.MTOServiceItem{
+			ID: serviceItemID,
+		},
+		Move:        mto,
+		MTOShipment: shipmentWithExistingServiceItem,
+		ReService: models.ReService{
+			Code: models.ReServiceCodeDLH,
+		},
+	})
+
```
monfresh, line 464: This block of code should go inside the subtest on line 563. The reason this test is failing is because this shipment with the existing service item is creating the DLH ReService code, but then the test setup is trying to create the DLH again (along with other codes) starting on line 465. Since you can't have duplicate codes, the test blew up during the test setup, before it got to any of the subtests.

monfresh:
I have a few thoughts about this. In the case of a TOO approving a diversion, we shouldn't have to worry about service items being created because we should be using the new `/shipments/{shipmentID}/approve-diversion` endpoint for this, which doesn't create service items.

The one scenario where we might want to prevent creating service items is when the Prime creates a new shipment with service items and that the TOO is approving for the first time. Some questions I have around that are:
* Is this a possible scenario?
* If so, do we assume that the Prime has created the correct service items, and that no other service items should be created?
* If all of these hold true, and we know for sure we never want to create any additional service items on a shipment that already has some, then we should add the logic + test from this PR to these files as well, since these are the files we want to use going forward:
* `shipment_approver.go`
* `shipment_approver_test.go `

hannaliebl:
@monfresh as best I understand, the scenario we need to account for is the Prime creating a new shipment, but then the old, existing shipment not getting additional service items added to it. I don't know about the Prime creating one with service items 😕

donaldthai:
> I have a few thoughts about this. In the case of a TOO approving a diversion, we shouldn't have to worry about service items being created because we should be using the new `/shipments/{shipmentID}/approve-diversion` endpoint for this, which doesn't create service items.
> 
> The one scenario where we might want to prevent creating service items is when the Prime creates a new shipment with service items and that the TOO is approving for the first time. Some questions I have around that are:
> 
> * Is this a possible scenario?
> * If so, do we assume that the Prime has created the correct service items, and that no other service items should be created?
> * If all of these hold true, and we know for sure we never want to create any additional service items on a shipment that already has some, then we should add the logic + test from this PR to these files as well, since these are the files we want to use going forward:
> * `shipment_approver.go`
> * `shipment_approver_test.go `

So the move will be available to the prime so technically once they create the new shipment, they then _can_ create "accessorial" service items which are manually created. These "accessorial" service items are different since they aren't created automatically like the "standard" service items. We should determine if the shipment has these "standard" service items or not then create them if they aren't there.

Here's a spreadsheet that differentiates the service items: https://docs.google.com/spreadsheets/d/1zeS5J2LVd_cyB2TW2rQPu1hHqKSXT-UY7RpUzfvTi0k/edit#gid=0

pkg/testdatagen/scenario/devseed.go
```diff
@@ -3527,6 +3527,58 @@ func createHHGNoShipments(db *pop.Connection) {
...
+			Locator:            "EXSTS1",
+			OrdersID:           orders.ID,
+			Status:             models.MoveStatusAPPROVALSREQUESTED,
+			SelectedMoveType:   &hhgMoveType,
+			AvailableToPrimeAt: swag.Time(time.Now()),
+		},
+	})
+
+	mtoShipment := testdatagen.MakeMTOShipment(db, testdatagen.Assertions{
+		MTOShipment: models.MTOShipment{
```
duncan-truss, line 3559: Could you set diversion here to true?

pkg/testdatagen/scenario/devseed.go
```diff
@@ -3617,6 +3669,7 @@ func (e devSeedScenario) Run(db *pop.Connection, userUploader *uploader.UserUplo
 	createMoveWithHHGMissingOrdersInfo(db, userUploader)
 
 	createHHGMoveWith10ServiceItems(db, userUploader)
+	createMoveWithExistingServiceItem(db, userUploader)
```
duncan-truss, line 3672: I would add diversion to this name to clarify it's purpose, your comment explains it well in the function itself.

pkg/services/mto_shipment/mto_shipment_updater_test.go
```diff
@@ -540,6 +541,39 @@ func (suite *MTOShipmentServiceSuite) TestUpdateMTOShipmentStatus() {
 		}
 	})
 
+	suite.T().Run("A shipment with existing service items does not create more", func(t *testing.T) {
+		shipmentWithExistingServiceItem := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+			Move: mto,
+			MTOShipment: models.MTOShipment{
```
duncan-truss, line 547: Let's set the diversion flag to true here as well to reflect reality.

monfresh:
What scenario are we trying to prevent here? In what situation would we approve a shipment that already has standard service items created? 

If the only case where that can happen is when the TOO is approving a diversion on an existing shipment that was previously approved, then this PR is not needed. That is addressed by using the `/approve-diversion` endpoint, which does not do anything to service items since this is an existing shipment.

For **new** shipments, those will go through the `/approve` endpoint, which *does* create standard service items. So, the only question we should ask if I'm understanding correctly, is whether it's possible for a **new** shipment to be created and already have one or more of the standard service items that it is required to have. Only if the answer is "yes" do we need to add logic to the `/approve` endpoint to prevent duplicate standard service items.

monfresh:
Another thought that came to me: if a shipment can't have duplicate service items, why not add a DB uniqueness constraint?

hannaliebl:
Closing this for now - we may come back to this in the future if need be.

PR #6823, opened 2021-06-16
[MB-7581] Add AddAmendedOrders endpoint/handler/service

pearl-truss:
## Description

This PR sets us up so that we can tie uploaded amended orders to the original order

## Reviewer Notes

Please double check my approach. There's a summary below of the changes in this PR and the work left to do in the [follow up story](https://dp3.atlassian.net/browse/MB-8335)

- Created a new endpoint called `upload_amended_orders` it should be called with the orderID in the url and param body == [upload](https://github.com/transcom/mymove/blob/master/pkg/models/upload.go) record(s) that will be attached to the order via the new UploadAmendedOrdersservice 

- When the endpoint is hit, the `UploadAmendedOrdersHandler` calls the new `UploadAmendedOrders` function (which is a part of the `OrderUpdater` service). 
- The service is essentially empty in this PR but in [MB-8335](https://dp3.atlassian.net/browse/MB-8335) it should be updated so that it checks the order for an existing UploadedAmendedOrders document and if there aren't any, it should generate a document ID and attach the incoming documents to. And if they're already are amended orders then it should just attach the incoming documents to the existing documents (in the Order.UploadedAmendedOrders field)

## Open Questions

- Currently, this PR is assuming the `upload_amended_orders` endpoint will receive a document, but looking through the relationships between document -> user_upload -> upload, I'm now wondering if this endpoint should work with an upload instead?
- I think this was answered and the pr reflects the switch to the endpoint supplying an upload instead of a document record

## Setup

Nothing is calling this code yet

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7581) for this change
* [discovery doc](https://docs.google.com/document/d/1lnOfe4AJNP5WfK7Ke9GTjI900LcXCXZ3JXKtRmOpAoU/edit#bookmark=id.wwokxevpa7v7)


pkg/handlers/internalapi/api.go
```diff
@@ -63,6 +64,10 @@ func NewInternalAPI(ctx handlers.HandlerContext) *internalops.MymoveAPI {
 	internalAPI.OrdersCreateOrdersHandler = CreateOrdersHandler{ctx}
 	internalAPI.OrdersUpdateOrdersHandler = UpdateOrdersHandler{ctx}
 	internalAPI.OrdersShowOrdersHandler = ShowOrdersHandler{ctx}
+	internalAPI.OrdersAddAmendedOrdersHandler = AddAmendedOrdersHandler{
```
jacquelineIO, line 67: Where is the upload orders handler? (I know you are doing amended but I expected to the see the original upload here too)

pkg/handlers/internalapi/orders.go
```diff
@@ -235,3 +236,39 @@ func (h UpdateOrdersHandler) Handle(params ordersop.UpdateOrdersParams) middlewa
...
 	return ordersop.NewUpdateOrdersOK().WithPayload(orderPayload)
 }
+
+// AddAmendedOrdersHandler adds amended orders to an order via PATCH /orders/{orderId}/add_amended_orders
+type AddAmendedOrdersHandler struct {
+	handlers.HandlerContext
+	services.OrderUpdater
+}
+
+// Handle ... updates an order from a request payload
```
jacquelineIO, line 246: don't need the dots here

pkg/services/order/order_updater.go
```diff
@@ -94,6 +95,26 @@ func (f *orderUpdater) UpdateAllowanceAsCounselor(orderID uuid.UUID, payload ghc
 	return f.updateOrder(orderToUpdate)
 }
 
+// AddAmendedOrders add amended order documents to an existing order
+func (f *orderUpdater) AddAmendedOrders(existingOrder models.Order, payload *internalmessages.DocumentPayload, eTag string) (*models.Order, uuid.UUID, error) {
```
jacquelineIO, line 99: I wonder if this function would be called "UploadAmendedOrders"?

jacquelineIO (commented):
Looks good so far.

pkg/services/order/order_updater.go
pearl-truss: Yeah I think that'll make more sense

pkg/handlers/internalapi/api.go
mboren: The handler for uploading files is `internalAPI.UploadsCreateUploadHandler`, it's around line 103 in this file

pkg/services/event/internal_endpoint.go
```diff
@@ -214,6 +217,10 @@ var internalEndpoints = EndpointMapType{
 		APIName:     InternalAPIName,
 		OperationID: "showOrders",
 	},
+	InternalAddAmendedOrdersEndpointKey: {
+		APIName:     InternalAPIName,
+		OperationID: "addAmendedOrders",
```
jacquelineIO, line 222: I think this didn't get updated.

pearl-truss: good catch!

jacquelineIO (approved):
lgtm

pkg/handlers/internalapi/orders.go
```diff
@@ -235,3 +236,39 @@ func (h UpdateOrdersHandler) Handle(params ordersop.UpdateOrdersParams) middlewa
...
+	if err != nil || verrs.HasAny() {
+		return handlers.ResponseForVErrors(logger, verrs, err)
+	}
+
+	orderPayload, err := payloadForOrdersModel(h.FileStorer(), *newOrder)
+	if err != nil {
+		return handlers.ResponseForError(logger, err)
+	}
+	return ordersop.NewUpdateOrdersOK().WithPayload(orderPayload)
+}
```
monfresh, line 274: As a reminder for future PRs, the best practice we want to follow is that business logic should live in services, not in handlers, including fetching objects from the DB and saving them. Handlers should be very lean and the only thing they should be responsible for is presenting what the service returns.

pkg/handlers/internalapi/orders.go
```diff
@@ -235,3 +236,39 @@ func (h UpdateOrdersHandler) Handle(params ordersop.UpdateOrdersParams) middlewa
...
+		return handlers.ResponseForError(logger, err)
+	}
+	order, err := models.FetchOrderForUser(h.DB(), session, orderID)
+	if err != nil {
+		return handlers.ResponseForError(logger, err)
+	}
+
+	newOrder, _, err := h.OrderUpdater.UploadAmendedOrders(order, params.AmendedOrders, params.IfMatch)
+	if err != nil {
+		return handlers.ResponseForError(logger, err)
```
monfresh, line 261: I believe we don't want to use this `ResponseForError` pattern anymore because it only takes into account a specific set of errors. It's mostly used in the internalapi, but we want to stop using it and follow the new pattern we've been using in the other APIs. Here are a couple Wiki articles we have about error handling. I'll also write a new one that shows how to create a brand new handler and what we expect it to look like.

https://github.com/transcom/mymove/wiki/API-Errors
https://github.com/transcom/mymove/wiki/handle-backend-errors

PR #6824, opened 2021-06-16
MB-3728 update interface for future refactoring

nelztruss:
## Description

In the spirit of smaller PRs, this is the first in a series.
This is mostly a non-functional change, meaning I changed the shape of some interfaces, but there is almost no logic/functionality that is different.
This change buys me some flexibility so that I can eventually build in some extensible business logic validation stuff in the service layer, like I did on #6810. This will also let me start moving some business logic stuff that the handlers are currently doing back into the service layer, where they belong.


sandy-wright (approved):
LGTM! As we discussed the poor naming scheme of the APIs, I think it _might_ be more helpful to name these functions `UpdateMTOShipmentCustomer` and `UpdateMTOShipmentOffice` instead of `Internal` and `GHC` respectively, but ultimately it'll be a weak point no matter what.

PR #6827, opened 2021-06-17
[MB-2352] Service object should match the API to Service Object pattern in the API

mboren:
## Description
I'm pretty much doing exactly what Moncef said to do in this post: https://ustcdp3.slack.com/archives/CP6F568DC/p1622735371219500

There were a few places were we have had to set up objects inside of a handler's `Handle` function, instead of storing it in the handler's context struct because the objects needed loggers that we weren't sure how to get access to when we needed it.

The affected handlers were: `primeAPI.PaymentRequestCreateUploadHandler` ~and
`supportAPI.PaymentRequestProcessReviewedPaymentRequestsHandler`.~

Thanks to Moncef, now we know that we can use the `context` library to set up the loggers that we need by doing `HandlerContext.LoggerFromContext(context.Background)`.

So I took advantage of that to move ~`PaymentRequestReviewedProcessor` into a field on `ProcessReviewedPaymentRequestsHandler` and~ `PaymentRequestUploadCreator` into a field on `CreateUploadHandler`.

As part of this, I renamed the variable `context` to `ctx` in a bunch of places, so we could use the `context` library without name clashes. This makes the patch a lot harder to skim through, so I'm sorry about that.

**I took changes to the support api files out of this PR because the changes to the prime API were much simpler. I'm going to split the remaining work into another ticket.**
## Setup
We need to upload a proof of service doc with the prime api to make sure that this works.
You should be able to follow the instructions for 2mil2mob up to the point where the prime uploads proof of service docs:
https://docs.google.com/presentation/d/15p49649oQWmo4Qveg0Ujg8aoiRMGCcSWZU_qeiRhPG0/edit#slide=id.gd42f88a641_4_19
I actually ran into issues when trying to do this locally, I think I have postman setup wrong for local testing.

I'm fairly certain there's an easier way to get to the proof of service upload than this, but just don't have the time tonight to figure it out. Hopefully we can figure it out Monday morning.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2352) for this change

### DIFF
--- a/pkg/handlers/primeapi/api.go
+++ b/pkg/handlers/primeapi/api.go
@@ -22,8 +22,8 @@ import (
 )
 
 // NewPrimeAPIHandler returns a handler for the Prime API
-func NewPrimeAPIHandler(context handlers.HandlerContext) http.Handler {
-	builder := query.NewQueryBuilder(context.DB())
+func NewPrimeAPIHandler(ctx handlers.HandlerContext) http.Handler {
+	builder := query.NewQueryBuilder(ctx.DB())
 	fetcher := fetch.NewFetcher(builder)
 
 	primeSpec, err := loads.Analyzed(primeapi.SwaggerJSON, "")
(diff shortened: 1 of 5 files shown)


pkg/handlers/supportapi/api.go
```diff
@@ -23,84 +35,115 @@ import (
...
-func NewSupportAPIHandler(context handlers.HandlerContext) http.Handler {
-	queryBuilder := query.NewQueryBuilder(context.DB())
+func NewSupportAPIHandler(ctx handlers.HandlerContext) http.Handler {
+	queryBuilder := query.NewQueryBuilder(ctx.DB())
 	supportSpec, err := loads.Analyzed(supportapi.SwaggerJSON, "")
 	if err != nil {
 		log.Fatalln(err)
 	}
 
+	v := viper.New()
```
mboren, line 45: The whole chunk of code starting here feels awkward. Is it OK to read env vars here? is there a way I can make it fit better? Is it ok for errors in this setup to break the whole support API?

pkg/handlers/primeapi/api.go
```diff
@@ -1,6 +1,7 @@
 package primeapi
 
 import (
+	"context"
```
reggieriser, line 4: I wonder if it would be better here to [alias this import](https://stackoverflow.com/questions/10408646/how-to-import-and-use-different-packages-of-the-same-name) to something else (like `gocontext "context"`, but open to a better alias name) since there's just one use of this package in this file instead of changing all the existing `context` references to `ctx` which makes this one `api.go` different from the other API collections.

pkg/handlers/supportapi/api.go
reggieriser: I concur -- this feels odd.  I'm wondering if we even need to do this setup here since we have an `icnSequencer` and a `gexSender` already in the handler context that we could just read and use.  Looking at `serve.go` where those are set up, I think [icnSequencer](https://github.com/transcom/mymove/blob/929a3f3eb44bd46394fe25b38f63c7b89ecadb31/cmd/milmove/serve.go#L627-L640) and [gexSender](https://github.com/transcom/mymove/blob/929a3f3eb44bd46394fe25b38f63c7b89ecadb31/cmd/milmove/serve.go#L593-L620) are set up mostly the same (main diff seems to be that we use a different version for test vs. a real environment, which seems like something we'd want to do here too?).  I didn't dive that deep, though.  Might be worth some more investigation, or at least another story to try to clean this up.

pkg/handlers/primeapi/api.go
mboren: I'm certainly open to it, I think Moncef has already switched an API collection over to using `ctx` so I was following in his footsteps. https://ustcdp3.slack.com/archives/CP6F568DC/p1622735371219500v

In the absence of moncef's post about it, I would definitely go with the alias, but I don't have strong feelings about it as long as we standardize on *something*.

pkg/handlers/supportapi/api.go
mboren: Reggie and I talked and decided to split this work off into another story because there were some differences in how the GexSenders set up certs, and it's going to take a little more digging to make sure we're doing the right thing.

pkg/handlers/primeapi/api.go
reggieriser: OK, I didn't realize we already had a precedent.  I tend to think that using `ctx` as well as `context` to refer to two different things in the same function is confusing, but I'm not going to block over it since it's stylistic and we've already done it elsewhere.  It'd be pretty easy at this point to change the other case of this if Moncef was onboard, but I agree that standardizing on something is preferable to multiple styles.

reggieriser (approved):
I was able to get a proof of service upload working locally and the revised code LGTM.  May want to note the URL for the story for addressing the other more complicated case we discussed.

PR #6786, opened 2021-06-10
[MB-4149] Upgrade xlsx to version 2.x

reggieriser:
## Description

This PR upgrades the `xlsx` Go dependency from version 1.x to version 2.x.  The latest version is 3.x, but this is an intermediate step given that 3.x has a lot of breaking changes that affect us and will take a little time to sort out.

We currently only use `xlsx` in the `ghc-pricing-parser` and `ghc-transit-time-parser` utilities.

## Reviewer Notes

- [What's new in 2.x](https://github.com/tealeg/xlsx/releases/tag/v2.0.0)
- [Code changes between version 1.05 and 2.0.1](https://github.com/tealeg/xlsx/compare/v1.0.5...v2.0.1)
- Readme notes for [2.x](https://github.com/tealeg/xlsx#version-2xx) and [3.x](https://github.com/tealeg/xlsx#version-3xx).
- I removed the reference to `xlsx` in tools.go as I'm not sure why it's needed.  It's not a CLI like some other items in that file where we don't reference it directly in code; we do have direct references to `xlsx` so there's no issue of it getting deleted via a `go mod tidy` or other go tooling.

## Setup

- `make clean server_build bin/ghc-pricing-parser bin/ghc-transit-time-parser`
- `make server_test` to ensure server tests still pass for parsers.
- See [this PR](https://github.com/transcom/mymove/pull/4250) for instructions on running `ghc-pricing-parser`.
- See [this PR](https://github.com/transcom/mymove/pull/3571) for instructions on running `ghc-transit-time-parser`.

## Code Review Verification Steps

* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4149) for this change

PR #3571, opened 2020-02-18
MB-1384 Transit times parser

donaldthai:
## Description

Shamelessly forked code from the pricing parser.

The GHC transit time parser to parse domestic transit times. The parser helped to produced [this migration](https://github.com/transcom/mymove/pull/3570) which produces a csv file so we can import into the database.

## Reviewer Notes

- Work in progress, needs some polishing

## Setup

Add any steps or code to run in this section to help others prepare to run your code:

1. Build parser binary
```sh
make bin/ghc-transit-time-parser
```

2. Download [XSLX here](https://docs.google.com/spreadsheets/d/1p2L_rvf-hsqp8djsSamuF1Dby5QxGSKS/edit#gid=1617474798) 

3. Run parser to generate a CSV file to import into your database table
```sh
bin/ghc-transit-time-parser --display --save-csv --filename [path to your file]/Appendix_C\(i\)_-_Transit_Time_Tables.xlsx 
```

4. Use your favorite database tool to import the data (ie. TablePlus)

## Code Review Verification Steps

* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/tree/master/docs/backend.md#logging)
* [ ] The requirements listed in
 [Querying the Database Safely](https://github.com/transcom/mymove/tree/master/docs/backend.md#querying-the-database-safely)
 have been satisfied.

* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-1384) for this change

PR #6830, opened 2021-06-17
[MB-8385] Upgrade xlsx to version 3.x

reggieriser:
## Description

This PR updates the `xlsx` Go dependency from version 2.x to version 3.x (after doing an intermediate 1.x to 2.x upgrade last week in #6786).  There were breaking changes that impacted us in 3.x, so you will find a lot of code changes along with the dependency upgrade.

Specifically, we were most affected by these changes:

- a `Sheet` struct no longer has a `Rows` slice field.
- a `Row` struct no longer has a `Cells` slice field.

Our usual pattern in all of our parsers/validators used those two fields above.  To adjust, I changed the code to use a `for` loop to iterate over the desired row indices, and adjusted our `getCell` method to take a `Sheet` along with a row/column index rather than the `Cells` slice and a column index.  It then uses a `sheet.Cell` call to get the specific cell at the row/column index.

Another change in the library is that the `sheet.Cell` call can return an error.  Given the way we have set things up, that should be pretty unlikely in normal operation.  But given how often we call `getCell` from the parsers, it could get a little unwieldy to have to always check and handle an error.  Since this code is only used as part of a CLI, I made a `mustGetCell` wrapper that will just panic with some context in the event of an error from `getCell`.  I'm open to other ideas on how to handle this, including checking every error from `getCell` explicitly if that feels better.

## Reviewer Notes

- [What's new in 3.x](https://github.com/tealeg/xlsx/releases/tag/v3.0.0)
- [Code changes between version 2.0.1 and 3.2.3](https://github.com/tealeg/xlsx/compare/v2.0.1...v3.2.3)
- Readme notes for [3.x](https://github.com/tealeg/xlsx#version-3xx).
- We have some pre-existing duplication of helper functions between the `ghc-pricing-parser` code and the `ghc-transit-time-parser` code.  To avoid making this story even more complicated, I will write up another story to consider refactoring/consolidating those into public shared functions with appropriate tests.
- I just noticed as I was writing up this PR that the maintainer of `xlsx` has [just added a note](https://github.com/tealeg/xlsx#not-actively-maintained) to the README in the last couple of days that they are not actively maintaining the library anymore.  There have been quite a few changes merged since the last formal release in Nov 2020, so perhaps someone else will pick up the maintenance.  But I'll create a separate story to capture this so we can evaluate later if we need to move to `excelize` or some other XLSX library (update: [story is here](https://dp3.atlassian.net/browse/MB-8475)).  I think we should still move forward with this upgrade for now.

## Setup

- `make clean server_build bin/ghc-pricing-parser bin/ghc-transit-time-parser`
- `make server_test` to ensure server tests still pass for parsers.
- See [this PR](https://github.com/transcom/mymove/pull/4250) for instructions on running `ghc-pricing-parser` and `pricing-import`.
- See [this PR](https://github.com/transcom/mymove/pull/3571) for instructions on running `ghc-transit-time-parser`.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8385) for this change

### DIFF
--- a/pkg/parser/pricing/parse_domestic_other_prices.go
+++ b/pkg/parser/pricing/parse_domestic_other_prices.go
@@ -27,10 +27,10 @@ var parseDomesticOtherPricesPack processXlsxSheet = func(params ParamConfig, she
 	sheet := params.XlsxFile.Sheets[xlsxDataSheetNum]
 	for rowIndex := rowIndexStart; rowIndex < sheet.MaxRow; rowIndex++ {
 		packPrice := models.StageDomesticOtherPackPrice{
-			ServicesSchedule:   getCell(sheet, rowIndex, servicesScheduleColumn),
-			ServiceProvided:    getCell(sheet, rowIndex, serviceProvidedColumn),
-			NonPeakPricePerCwt: getCell(sheet, rowIndex, nonPeakPriceColumn),
-			PeakPricePerCwt:    getCell(sheet, rowIndex, peakPriceColumn),
+			ServicesSchedule:   mustGetCell(sheet, rowIndex, servicesScheduleColumn),
+			ServiceProvided:    mustGetCell(sheet, rowIndex, serviceProvidedColumn),
+			NonPeakPricePerCwt: mustGetCell(sheet, rowIndex, nonPeakPriceColumn),
+			PeakPricePerCwt:    mustGetCell(sheet, rowIndex, peakPriceColumn),
 		}
 
 		if packPrice.ServicesSchedule != "" {
@@ -65,10 +65,10 @@ var parseDomesticOtherPricesSit processXlsxSheet = func(params ParamConfig, shee
 	sheet := params.XlsxFile.Sheets[xlsxDataSheetNum]
 	for rowIndex := rowIndexStart; rowIndex < sheet.MaxRow; rowIndex++ {
 		sitPrice := models.StageDomesticOtherSitPrice{
-			SITPickupDeliverySchedule: getCell(sheet, rowIndex, servicesScheduleColumn),
-			ServiceProvided:           getCell(sheet, rowIndex, serviceProvidedColumn),
-			NonPeakPricePerCwt:        getCell(sheet, rowIndex, nonPeakPriceColumn),
-			PeakPricePerCwt:           getCell(sheet, rowIndex, peakPriceColumn),
+			SITPickupDeliverySchedule: mustGetCell(sheet, rowIndex, servicesScheduleColumn),
+			ServiceProvided:           mustGetCell(sheet, rowIndex, serviceProvidedColumn),
+			NonPeakPricePerCwt:        mustGetCell(sheet, rowIndex, nonPeakPriceColumn),
+			PeakPricePerCwt:           mustGetCell(sheet, rowIndex, peakPriceColumn),
 		}
 
 		if sitPrice.SITPickupDeliverySchedule != "" {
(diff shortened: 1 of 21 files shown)


pkg/parser/pricing/parse_domestic_other_prices.go
```diff
@@ -112,18 +112,19 @@ func verifyPackUnpackPrices(params ParamConfig, sheetIndex int) error {
 	const headerIndexEnd = headerIndexStart + 1
 
 	// Verify header strings
-	dataRows := params.XlsxFile.Sheets[sheetIndex].Rows[headerIndexStart:headerIndexEnd]
-	for _, row := range dataRows {
-		if header := removeWhiteSpace(getCell(row.Cells, servicesScheduleColumn)); header != "ServicesSchedule" {
+	sheet := params.XlsxFile.Sheets[sheetIndex]
+	for rowIndex := headerIndexStart; rowIndex < headerIndexEnd; rowIndex++ {
+		// for _, row := range dataRows {
```
mboren, line 117: nit: this leftover commented line can be deleted

reggieriser: Good catch!  Looks like I missed this one and one other in this same file that I've now fixed.  I did a search on the other files but they seemed OK already.

pkg/parser/pricing/shared.go
```diff
@@ -27,13 +27,32 @@ type headerInfo struct {
...
-	if len(cells) > i {
-		return cells[i].String()
+// A safe way to get a cell's value (as a string) from a sheet
+func getCell(sheet *xlsx.Sheet, rowIndex, colIndex int) (string, error) {
+	if rowIndex < 0 || rowIndex >= sheet.MaxRow || colIndex < 0 || colIndex >= sheet.MaxCol {
+		return "", fmt.Errorf("cell coordinates are out of bounds")
 	}
 
-	return ""
+	if sheet == nil {
```
mboren, line 36: nit: i think this nil check should be the first thing in the function because the conditional before it uses `sheet.MaxRow`

pkg/parser/transittime/shared.go
```diff
@@ -16,12 +16,31 @@ import (
...
-func getValueFromSheet(sheet *xlsx.Sheet, row int, col int) string {
-	if sheet != nil {
-		return sheet.Cell(row, col).String()
+func getValueFromSheet(sheet *xlsx.Sheet, row int, col int) (string, error) {
+	if row < 0 || row >= sheet.MaxRow || col < 0 || col >= sheet.MaxCol {
+		return "", fmt.Errorf("cell coordinates are out of bounds")
 	}
 
-	return ""
+	if sheet == nil {
```
mboren, line 24: I think this `nil` check should come before the previous check

pkg/parser/pricing/shared.go
reggieriser: Good eyes!  And you're too gracious calling that a "nit" -- that's a straight up bug. 😉  Should be fixed now.

pkg/parser/transittime/shared.go
reggieriser: Fixed along with the other similar case you noted.

mboren:
I looked over the code pretty thoroughly and I think it looks good. I'm planning to come back to this and run the tests later tonight before giving a final approval (my brain is fried from staring at this patch :lolsob:), unless somebody else beats me to it.

mboren (approved):
I ran `pricing-import` for both master and this branch, then diffed the resulting db dumps. After accounting for differing UUIDs, timestamps, and line ordering, all the lines match.

PR #6832, opened 2021-06-17
[MB-8320]: setup the tasks to deploy to the demo-env

sheenamt:
Initial PR is to test circle setup 

Copy all the exp tasks to the demo env


.circleci/config.yml
```diff
@@ -1298,7 +1411,49 @@ jobs:
...
+
+  # `deploy_demo_app` updates the server-TLS app service in the demo environment
+  deploy_demo_app:
+    executor: mymove_small
+    environment:
+      - APP_ENVIRONMENT: 'demo'
+    steps:
+      - aws_vars_demo
+      - deploy_app_steps:
+          compare_host: '' # leave blank since we want demo to be able to roll back
```
carterjones, line 1442: What do you mean by "roll back"? I'm just trying to understand the comment. I also see in `scripts/deploy-app`, `scripts/deploy-app-migrations`, and `scripts/deploy-app-tasks` that the compare host is set to my.demo.dp3.us. I'm wondering which approach we should use: no compare host, my.demo.dp3.us, or something else entirely.

carterjones: We investigated this together and it looks like setting `compare_host` to the empty string allows the demo environment to be deployed regardless of its current state, which is what we want. The `compare_host` value serves as a check to ensure that only the current or a prior version is running. Given that demo can be experimental in nature, we don't want to enforce this constraint. Therefore, setting compare_host to the empty string is fine _and_ leaving it defined in the various scripts is also fine. We can remove the value in the scripts, but since it is defined in experimental, we can follow suit here. No change appears to be needed.

carterjones (approved):

reggieriser (approved):
I'm approving from the migrations perspective.  Here's what I checked:

1. The `demo` migrations all matched the current `exp` migrations with the exception of `20210630143013_add_demo_users.up.sql` as expected (no-op in `exp`, adds a user/admin_user in `demo`).
1. I didn't see any new secure migrations in master that we're missing.
1. I ran `make run_demo_migrations` and I see the new user/office_user as expected.  I can't log in as that user since it's not me, but I know that Sheena has tried that.
1. I can login to admin.demo.dp3.us fine using my credentials.

A couple of notes:
- I only looked at the migration work; others with more infra context should bless the rest of the PR.
- [nit] I did notice that the no-op file in the other deployed environments has a header about it being a `-- Local test migration`.  That's true for a local migration in dev, but not accurate for a deployed migration.  But it's still a no-op as expected, and I see we have this wrong comment in some other no-ops, so certainly not a blocker.
- We'll need to be careful in rolling this out to make sure we notify anyone with an in-progress secure migration since that migration won't be in S3 for demo unless they merge this PR in and re-upload the secure migrations to all environments including demo.
- I paired with Sheena on some of the migration stuff, so any other new eyes on this are welcome.

lynzt (approved):
only looked at the secure migration, but that looks good to me 🚢 
nice work @sheenamt 🎉

travelar (approved):

PR #6833, opened 2021-06-17
MB-8310 Set diverted shipments to SUBMITTED status, and move to APPROVALS REQUESTED

sandy-wright:
## Description

This PR updates the shipment service objects so that they will transition all diverted shipments to the "SUBMITTED" status, and it will move their respective moves to the "APPROVALS REQUESTED" status.

## Reviewer Notes

Ultimately, I think we need to add a new endpoint/service object for the diversion action. A complete diversion involves updating an existing shipment, creating a new shipment, _and_ updating a move. It's a unique request with combo actions that right now need to be completed separately. However, when I got into doing that, it got messy... So I think the work to create a new Prime API endpoint like `divertMTOShipment` should be in its own ticket.

**Also,** I noticed that the Prime API does not return the move status in its responses, which is definitely a bug. I'll add a ticket to fix that to the backlog.

## Setup

* Spin up your local environment:
   ```sh
   ```

* Find some APPROVED moves. Since you can't see the move's status in the Prime API, you'll need to use some combo of the Prime API, the Support API, the office interface, and/or a database viewer to ensure you have a move with the right status.

* Once you have a move, use the `createMTOShipment` endpoint in the Prime API to add a new shipment. This shipment does _not_ have to be part of a diversion.
   * Verify that the shipment comes back with the SUBMITTED status.
   * The move should now have the APPROVALS REQUESTED status. Any time the Prime adds a new shipment, regardless of it being a diversion or not, the move should transition to this status.

* Either mark this move and shipment as APPROVED again using the office interface/GHC API, or find a new move and shipment. We want an existing shipment this time.

* With the approved shipment's ID and eTag (you can get the eTag of the shipment using `getMoveTaskOrder` in the Prime API), use the `updateMTOShipment` endpoint to mark the shipment as diverted. Use this body:
   ```json
   {
     "diversion": true
   }
   ```
   * Verify that this update returns the shipment with the SUBMITTED status.
   * Check that the move has also changed to the APPROVALS REQUESTED status.

* Feel free to also test updating a non-diverted shipment and checking that the status **don't** change unexpectedly.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8310) for this change.


pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -216,6 +216,18 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
...
 
+		// transition the move to "Approvals Requested" if a shipment was created with the "Submitted" status:
+		if shipment.Status == models.MTOShipmentStatusSubmitted && move.Status == models.MoveStatusAPPROVED {
+			err = move.SetApprovalsRequested()
+			if err != nil {
+				return fmt.Errorf("%e", err)
+			}
+			verrs, err = txBuilder.UpdateOne(&move, nil)
+			if verrs != nil || err != nil {
+				return fmt.Errorf("%#v %e", verrs, err)
```
nelztruss, line 227: I'm not sure what this is trying to do. The `%e` verb looks like it is only associated with floating point numbers: https://golang.org/pkg/fmt/

Hrm... Is the `%e` maybe intended to be the `%w` (~wrap) verb? 
If that's the case:
* for the first error, just return it unwrapped if there's nothing else being added in the `fmt.Errorf(...)` call.
* for the second, should this prioritize the non-validation errors firstly, and only return the `verrs` if the regular error is nil?

nelztruss (changes requested):
Just a couple of error-handling nits

pkg/services/mto_shipment/mto_shipment_creator.go
sandy-wright: ooh good looking out - I had just thoughtlessly copy/pasted this (shame). I think it is supposed to be `%w`, I've seen that precedent in other places. Changes incoming 👍

sandy-wright: I went with `fmt.Errorf("%v", verrs.Errors)` because that seemed like the most explicit way to show all the validation errors, but let me know if you think something else would be better

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -216,6 +216,21 @@ func (f mtoShipmentCreator) CreateMTOShipment(shipment *models.MTOShipment, serv
...
+			if err != nil {
+				return err
+			}
+			verrs, err = txBuilder.UpdateOne(&move, nil)
+			if err != nil {
+				return err
+			}
+			if verrs != nil {
+				return fmt.Errorf("%v", verrs.Errors)
+			}
```
nelztruss, line 231: Two more nits about this:
* I would also check if `verrs.HasAny()` before returning the error (otherwise, if `verrs` is non-nil, but has zero errors in it, you'll still be reporting an error)
* I don't believe you need to expand the `verrs` in an `fmt.Errorf(...)`, as the `*validate.Errors` type actually implements the `error` interface itself.

nelztruss (changes requested):

pkg/services/mto_shipment/mto_shipment_creator.go
sandy-wright: Added the HasAny() check!

If we don't format the `verrs`, we'll lose any information we have from the "field" value for the the errors. `verrs.Error()` will look like:
```
<verr message 1>
<verr message 2>
```
instead of:
```
map[<verr field 1>:[<verr message 1>] <verr field 2>:[<verr message 2>]]
```

Since these are going to be returned as server errors and only visible to devs in the logs, the second format seemed more helpful for the sake of debugging. I'm not sure it really makes a significant difference though, so I went ahead and switched to returning it unformatted. But might be something to keep in mind

nelztruss: Gotcha, thanks for pointing out what the functional difference in results are between those two options.

I am now cranky at the `validate` package for collecting a bunch of information, and then if you use the standard `Error()` interface, it hides all the information that you went through the trouble of collecting.

nelztruss (approved):

PR #6840, opened 2021-06-22
[MB-8101] customer can see uploaded files for amended orders

mboren:
## Description

In this PR, I changed the page for amended orders to show amended orders instead of previously uploaded orders, which were there as a placeholder.

I also added an extra check to `payloadForOrdersModel` to fix a bug. When I created an amended order and then log out and then back in (as the service member), we get into a state where this function is called with a non-`nil` `order.UploadedAmendedOrdersID` and a `nil` `order.UploadedAmendedOrders`, which causes a 500 and kicks you back to the login screen. I didn't track this down to its root cause, but I'm guessing this is an eager preload related thing.

## Reviewer Notes

The page does not match the designs yet. I'm splitting visual changes into a separate PR.

## Setup
The endpoint to create a document for amended orders is still in progress, so you have to do some manual DB editing to create another document to test with.

Start by creating a service member, going through the onboarding flow, and submitting a move.
Then click the "Upload documents" button on the home screen to go to the amended orders page:
image
You should not see any amended orders yet, so it should look something like this:
image

If you're doing acceptance testing on staging, you're done. if you're testing locally, we can edit the DB to fake uploading a file for amended orders. You could get some of this from fetch-mto-updates, but we'll need to do some queries regardless, so might as well do it all.

Get the `id` and `user_id` for your service member from the `service_members` table
```sql
-- if you've got a bunch of users, you will probably want to filter on name or DoD ID
select * from service_members;
--                   id                  |               user_id                |   edipi    | affiliation | rank | first_name | middle_name | last_name | suffix |  telephone   | secondary_telephone |             personal_email              | phone_is_preferred | email_is_preferred |        residential_address_id        |      backup_mailing_address_id       |         created_at         |         updated_at         |           duty_station_id            | requires_access_code
-- --------------------------------------+--------------------------------------+------------+-------------+------+------------+-------------+-----------+--------+--------------+---------------------+-----------------------------------------+--------------------+--------------------+--------------------------------------+--------------------------------------+----------------------------+----------------------------+--------------------------------------+----------------------
--  58985910-6c0c-4256-aea8-2505a9a08620 | 72068b0c-ed23-4a77-beff-1d2bb3e713b0 | 1111111111 | COAST_GUARD | E_3  | Jason      |             | Ash       |        | 555-555-5555 | 555-555-5555        | 20210621195043-58c786181a14@example.com |                    | t                  | 0967bbba-216e-400a-b486-d843a4e42381 | 3eb888dd-dd2c-4051-a653-24c0b3cbfe47 | 2021-06-21 19:50:43.982679 | 2021-06-21 19:51:21.139363 | d7285e29-85f3-4839-b233-2512e18317cd | f
```

create a document record (you can use the same document ID that i'm using here, but must replace the service member id)
```sql
INSERT INTO documents (id, created_at, updated_at, service_member_id, deleted_at)
VALUES ('97c2874f-6438-4270-8d48-6a80a6547a48', now(), now(), '58985910-6c0c-4256-aea8-2505a9a08620', NULL)
;
```

Find id for the current order for your service_member
```sql
select id from orders where service_member_id = '58985910-6c0c-4256-aea8-2505a9a08620';
-- My order_id was: deeb4d61-5803-492c-b87b-d42cf76ed81f
```

Now update the order to point to the new document (replace id with the `order_id` from the previous query).
```sql
update orders
set uploaded_amended_orders_id = '97c2874f-6438-4270-8d48-6a80a6547a48'
where id = 'deeb4d61-5803-492c-b87b-d42cf76ed81f';
```
Now that we've created a new document and hooked it up to our orders, we're going to steal one of the original orders files that were uploaded for this user, and make it look like it was uploaded as an amended order file.

Get the id of one of the files that you uploaded for this user (replace the ID in this query with your `user_id`)
```sql
select id from user_uploads where uploader_id = '72068b0c-ed23-4a77-beff-1d2bb3e713b0';
-- My user_upload_id is: a4b2e84b-a49d-4f4c-bab8-af87559d4e1b
```

Update the document ID for that upload record
```
update user_uploads
set document_id = '97c2874f-6438-4270-8d48-6a80a6547a48'
where id = 'a4b2e84b-a49d-4f4c-bab8-af87559d4e1b';
```

Now that you've set all that up, you should be able to log in as this user, click "Upload Documents", and see the file you set up in the Uploads Table
image

## Acceptance
We can't set up amended orders, so acceptance should be pretty simple.
Either create or find a new move that has orders documents uploaded and has been submitted.
Then navigate to the amended orders page. If you see no uploaded files listed, then the changes were successful.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8101) for this change

## Screenshots


reggieriser (commented):
The code changes make sense to me, although someone with more recent front-end experience might want to look it over.  When testing, I was able to get the new file to show up by hacking my local database as described and refreshing the page.  However, if I go back to the main "submitted move" page (or log in again as per the instructions), I don't see the "Upload Documents" link nor any section about amended orders.  That sounds like it may be expected since you say "we can't set up amended orders yet", but just wanted to double-check since the instructions say to click the "upload documents" button again after logging in.

mboren:
I was able to reproduce the issue Reggie described above. I think it comes down to a problem with my instructions rather than with the code. My instructions initially said to create a move with one orders file uploaded, and then mess with the DB to use that file for amended orders instead.
Turns out, if you submit a move, and then mess with the DB and get rid of all of the uploaded files for its orders, the home screen gets kind of broken:


If you upload multiple files while setting up the move initially, and then follow my instructions to associate just one of them with amended orders, the home screen will not be broken.

YanZ777 (approved):
FE changes look good to me. (And I would like to make use of them as well.)

reggieriser (approved):

PR #6845, opened 2021-06-22
[MB-8415] Edit orders review amended orders sets move status to approved

duncan-truss:
## Description

As part of supporting diversions due to amended orders updated by the customer, the TOO should be able to see the amended orders document displayed on the edit orders page.  The TOO will now have an option to acknowledge the updated orders by checking a form input in the sidebar.  If the move previously changed status from approved to approvals requested then acknowledging the orders should set the move status back to approved.

This PR creates a new database migration to store a timestamp for when the TOO has checked they have reviewed the amended orders.  Without this it was not possible or too complex to reason about if the TOO had viewed the most recent amended orders.  This feeds into JP's PR for when to display the tag for alerting the TOO that there are new orders awaiting the office user.

## Reviewer Notes

I don't think the orders amendment upload handler on the customer side is done so we can't do things like setting the amendment acknowledged at timestamp to nil if there are superseding orders.

## Setup

```sh
make server_run
make office_client_run
make db_dev_e2e_populate
```

1. Login at the TOO and find the move with locator code `AMDORD`.
2. Click on the Move Task Order tab and approve or reject the crating and uncrating service items.
3. The status should remain in Approvals requested because of the existence of amended orders
3. Click on Edit orders from the Move Details page
4. The 2 amended orders documents should be the first ones shown in the document viewer.
5. There should be an unchecked checkbox at the bottom of the form attesting that you've reviewed the neweest orders docs.
6. The move status, which was previously in the Approvals requested status should now be in the approved status after saving the form.

You could use this same move to verify that reviewing the amended orders does not change the move status to Approved if there are still outstanding service items requested from the prime.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8415) for this change
* [Zeplin](https://zpl.io/VK7exZJ)

## Screenshots

Screen Shot 2021-06-24 at 9 49 45 AM
Screen Shot 2021-06-24 at 9 50 01 AM
Screen Shot 2021-06-24 at 9 50 09 AM
Screen Shot 2021-06-24 at 10 23 14 AM
Screen Shot 2021-06-24 at 10 23 22 AM
Screen Shot 2021-06-24 at 9 51 39 AM


duncan-truss:
I have more tests and error cases to write but I think it's in a state where others can leave feedback.

pkg/handlers/ghcapi/orders.go
```diff
@@ -109,6 +116,23 @@ func (h UpdateOrderHandler) Handle(params orderop.UpdateOrderParams) middleware.
 
 	h.triggerUpdateOrderEvent(orderID, moveID, params)
 
+	// the move status may need set back to approved if the amended orders upload caused it to be in approvals requested
+	if params.Body.OrdersAcknowledgement != nil && *params.Body.OrdersAcknowledgement && updatedOrder.UploadedAmendedOrdersID != nil && updatedOrder.AmendedOrdersAcknowledgedAt != nil {
+		moveRouter := move.NewMoveRouter(h.DB(), logger)
+		approvedMove, approveErr := moveRouter.ApproveAmendedOrders(*updatedOrder)
```
duncan-truss, line 122: When I initially wrote this I missed that the move id is being returned from the orderUpdater above, so I will either refactor it if I have time or leave a comment to that effect.

pkg/handlers/ghcapi/orders.go
```diff
@@ -109,6 +116,23 @@ func (h UpdateOrderHandler) Handle(params orderop.UpdateOrderParams) middleware.
...
+		if approveErr != nil {
+			if errors.Is(approveErr, models.ErrInvalidTransition) {
+				return handleError(services.NewConflictError(moveID, approveErr.Error()))
+			}
+			return handleError(approveErr)
+		}
+
+		updateErr := h.moveUpdater.UpdateApprovedAmendedOrders(approvedMove)
```
duncan-truss, line 130: I don't know if this really should be in a transaction with updating the orders but was not keen on adding it to the order service.

pkg/services/move/move_router.go
```diff
@@ -268,6 +268,48 @@ func (router moveRouter) CompleteServiceCounseling(move *models.Move) error {
 	return nil
 }
 
+// ApproveAmendedOrders sets the move status to APPROVED if it's status was set to
+// APPROVALS_REQUESTED because of the customer amending their orders.  If there are service items
+// needing review from the prime the status should remain in APPROVALS_REQUESTED
+func (router moveRouter) ApproveAmendedOrders(orders models.Order) (models.Move, error) {
+	var move models.Move
+	err := router.db.EagerPreload("MTOServiceItems").
```
duncan-truss, line 276: At first I thought I could do the check for service items status here in a single query but then my move variable would be an empty object in the case of no service items in the SUBMITTED status therefore I ended up looping through the service item statuses.

lynzt (approved):
only looked at the migration, but that LGTM 🚢

migrations/app/schema/20210622184646_add_amended_orders_acknowledgement_timestamp.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE orders
+    ADD COLUMN amended_orders_acknowledged_at timestamp;
```
monfresh, line 2: Can we please specify "timestamp with time zone"? I'm not sure what the default is, but I raised a question a while back in #prac-engineering as to being consistent. In the past, we've used both `timestamp with time zone` and `timestamp without time zone`, but I think we should always be using `with time zone` because that's the best practice recommended by Postgres, and is what Rails does now. https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp_.28without_time_zone.29

[Original Slack thread](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1617664477206100)

pkg/handlers/ghcapi/orders.go
monfresh: Can we move this logic to a service object? We want to establish a best practice of keeping handlers lean. Also, could we move this logic into a well-named function? It's hard to reason about those 4 conditionals. Can we come up with a name that describes what is happening?

migrations/app/schema/20210622184646_add_amended_orders_acknowledgement_timestamp.up.sql
duncan-truss: Thanks for the reminder, I will be more vigilant about that.  I've read similar discussions about the time zone debate with Postgres.  Unfortunately JP merged my changes into their branch and merged so I can't simply modify it here.

pkg/handlers/ghcapi/orders.go
duncan-truss: I'm game to move the conditional into a better named function, i don't think it warrants a service object, it's calling the move router and move updater service objects already.

pkg/services/move/move_router.go
```diff
@@ -275,6 +275,48 @@ func (router moveRouter) CompleteServiceCounseling(move *models.Move) error {
 	return nil
 }
 
+// ApproveAmendedOrders sets the move status to APPROVED if it's status was set to
```
monfresh, line 278: it's -> its

pkg/services/move/move_router.go
```diff
@@ -275,6 +275,48 @@ func (router moveRouter) CompleteServiceCounseling(move *models.Move) error {
...
+// APPROVALS_REQUESTED because of the customer amending their orders.  If there are service items
+// needing review from the prime the status should remain in APPROVALS_REQUESTED
```
monfresh, line 280: did you mean needing review from the TOO?

duncan-truss: Ahh yea maybe I should reword to state accessorial service items created by the prime waiting for TOO review.

pkg/services/move/move_router.go
```diff
@@ -275,6 +275,48 @@ func (router moveRouter) CompleteServiceCounseling(move *models.Move) error {
...
+func (router moveRouter) ApproveAmendedOrders(orders models.Order) (models.Move, error) {
```
monfresh, line 281: I'm not sure this function belongs in the move router. It is operating on an order. What do you think about moving it to a separate service?

duncan-truss: I've updated the method definition to accept the move ID argument but wanted it to be clear why it's been called. I thought I had to work around not having the move id but saw it was returned by the order updater in the handle.

pkg/services/move/move_router.go
```diff
@@ -275,6 +275,48 @@ func (router moveRouter) CompleteServiceCounseling(move *models.Move) error {
...
+	err := router.db.EagerPreload("MTOServiceItems").
+		Where("moves.orders_id = ?", orders.ID).
+		First(&move)
+
+	if err != nil {
+		router.logger.Error("failure encountered querying for move associated with orders", zap.Error(err))
+		return models.Move{}, fmt.Errorf("failure encountered querying for move associated with orders, %s, id: %s", err.Error(), orders.ID)
+	}
+
+	if move.Status != models.MoveStatusAPPROVALSREQUESTED {
```
monfresh, line 292: I think this block is duplicating logic in the move router's `Approve` function. I think we can remove it and have everything still work, no? I think this is what we want:
1. Fetch the move and its service items
2. If there are service items needing review, don't change the move's status
3. If there aren't any service items needing review, call `MoveRouter.Approve`

duncan-truss: I'm not sure I completely understand.  I don't think we can rely on Approve() solely because that allows approval from other statuses beyond APPROVALS REQUESTED which I was trying to be conservative about here.

pkg/services/move/move_router.go
monfresh: I think the scope of routers, like the move router and shipment router, is to transition from one status to another, and where each function operates on a unique start and end. Here, we are not introducing any new statuses, we're just determining whether or not we should approve the move. I think this logic belongs elsewhere.

pkg/services/move/move_router.go
monfresh: If I remove this conditional, the test you wrote in move_router_test.go passes, which tells me that this code is not needed. Either that, or the test is not sufficient. Theoretically, this function should only be called after amended orders have been uploaded, which we know sets the status to Approvals Requested. 

Do we really need the extra check? I would say no, but if you really think we need it, I would propose a different error message because the current message implies that we are trying to approve the move, but in some cases, we won't (depending on the service items). I think the message should say something like "Cannot approve amended orders because the move status is not Approvals Requested".

pkg/services/mto_service_item/mto_service_item_updater.go
```diff
@@ -130,6 +130,12 @@ func (p *mtoServiceItemUpdater) UpdateMTOServiceItemStatus(mtoServiceItemID uuid
 				break
 			}
 		}
+
+		// All service items may be reviewed but amended orders should keep the move in
+		// APPROVALS REQUESTED status
+		if move.Orders.UploadedAmendedOrdersID != nil && move.Orders.AmendedOrdersAcknowledgedAt == nil {
+			moveShouldBeMoveApproved = false
+		}
```
monfresh, line 138: I wonder if there is a concept here waiting to be extracted. We seem to be making similar checks — about whether or not the move should stay in Approvals Requested or go to Approved — in several places.

monfresh: Just something to think about as a follow-up.

pkg/services/move/move_router.go
duncan-truss: To generate the conflict 409 error response I wrote a test in the handler orders_test.go when the move was in services counseling status https://github.com/transcom/mymove/pull/6845/files#diff-7ed9d49b328573d1e4b4c17ef8455b68ba22cedf910612f033634107ba60688aR333.  You're right it's an unlikely scenario my thinking was just we don't know how things will evolve so it might catch someone using it unknowingly.  I can adopt that error message too.

pkg/handlers/ghcapi/orders.go
monfresh: Upon further inspection, I see what's going on. This existing handler was updating orders, and in some cases, we also want to approve amended orders. The general rule of thumb is that we should avoid conditionals in handlers, apart from checking errors returned by a service. So, I would propose one of these two solutions:

1. Move this logic into the existing `UpdateOrderAsTOO` function
OR
2. Create a new endpoint for approving amended orders since it is a distinct event from other types of order updates.

If the typical scenario of approving amended orders doesn't involve updating other order attributes, i.e. if creating a new endpoint will result in only one endpoint being called, then I would vote for option 2.

I won't block the PR for this, but I think we do want one of those 2 options, and if we don't choose one in this PR, I think we should write a follow-up JIRA ticket and work on it next week. I'm open to being convinced otherwise.

monfresh (approved):
This was a lot of work! Thanks for putting this together. I tested it locally, and it works as described. One thing I noticed is that it isn't clear which orders are the amended ones versus the original ones. Will there be some kind of visual indication to differentiate them? And I wonder if the checkbox should only show up when viewing the amended orders?

As for the logic for checking if amended orders require approval, I firmly believe that does not belong in the handler, but we can refactor that next week. It might be a good candidate for the validator pattern, along with the logic for determining whether or not a move should be approved or approvals requested based on the service items status. Here's an example of extracting this logic into a validator service: https://github.com/transcom/mymove/pull/6866/files?diff=split&w=1

duncan-truss:
> This was a lot of work! Thanks for putting this together. I tested it locally, and it works as described. One thing I noticed is that it isn't clear which orders are the amended ones versus the original ones. Will there be some kind of visual indication to differentiate them? And I wonder if the checkbox should only show up when viewing the amended orders?
> 
> As for the logic for checking if amended orders require approval, I firmly believe that does not belong in the handler, but we can refactor that next week. It might be a good candidate for the validator pattern, along with the logic for determining whether or not a move should be approved or approvals requested based on the service items status. Here's an example of extracting this logic into a validator service: https://github.com/transcom/mymove/pull/6866/files?diff=split&w=1

Thanks for the review @monfresh.  There was work previously to sort all order document files from newest to oldest and display the uploaded on date so that is how they will be displayed by default.  JPs branch shows the NEW tag on the prior Move Details screen if the move was already approved to draw attention.  The checkbox should only be displayed if their are amended orders in the form for the TOO.  We could go a step further and only show it if the move is in the approvals requested status as well.  I considered making it disabled once it has a value but not sure it makes a difference if we ignore all future values to update it.

I'm up for slimming down the handler and also want to make sure we're not pushing  all the complexity to the frontend or duplicating another service object.  I'd be game for a dedicated ticket instead of one this large.

Shkeating (approved):
looks good! Approving, but did want to flag that the pdfs did not show up for me on the Review Bot deployment.

PR #6851, opened 2021-06-23
MB-8175 verify that adding addtional destination address works end to end

YanZ777:
## Description

Added to the existing Cypress Tests for ntsr and hhg to check that a secondary destination address can be added.

## Reviewer Notes

Is there a better way to cypress?

## Setup

```sh
make e2e_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8175) for this change


cypress/integration/mymove/hhg.js
```diff
@@ -98,6 +98,14 @@ function customerSetsUpAnHHGMove() {
   cy.get(`select[name="delivery.address.state"]`).select('CA');
   cy.get(`input[name="delivery.address.postal_code"]`).type('91111').blur();
 
+  // secondary delivery location
+  cy.get('input[title="Yes, I have a second destination location"]').check({ force: true });
```
lynzt, line 102: I would target this based on a `data-testid` selector.

I am curious if we're NOT using the `data-testid` for a reason? If not, I see us NOT using `data-testid` in other cypress tests... I think I would advocate for us to modify those other tests too - but not in this PR.

cypress/integration/mymove/ntsr.js
```diff
@@ -58,6 +58,8 @@ function customerEditsNTSRShipmentFromReviewPage() {
   cy.get('button[data-testid="edit-ntsr-shipment-btn"]').contains('Edit').click();
   cy.get('input[name="delivery.requestedDate"]').clear().type('01/01/2022').blur();
   cy.get('input[name="delivery.address.street_address_1"]').clear().type('123 Maple street');
+  cy.get('input[title="Yes, I have a second destination location"]').check({ force: true });
```
lynzt, line 61: same here w/ targeting using `data-testid`

lynzt (commented):
Question about how we're targeting in cypress tests...

cypress/integration/mymove/hhg.js
kylehilltruss: So this is interesting. According to the [guiding principles](https://testing-library.com/docs/queries/about#priority) of React Testing Library, `data-testid` should be the selector of last resort, as it's a selector that's deliberately imperceptible to anyone who would actually be using the software, and so it makes testing less useful.

But obviously this isn't RTL, and [Cypress' own best practices](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements) state that they favor using `data-` attributes as a selector for its tests. 

There's very little functional difference in how *we're* using Cypress and RTL, so this is one of those giant cognitive code-switching annoyances that I'd love to iron out. IMHO I'd open this up to a wider discussion and figure out some sort of project-wide standard on what *our* ideal practices actually are for both testing frameworks; that's... probably outside the scope of this particular pull request review for me though.

src/components/Customer/MtoShipmentForm/MtoShipmentForm.jsx
```diff
@@ -348,6 +348,7 @@ class MtoShipmentForm extends Component {
                                         <div className={formStyles.radioGroup}>
                                           <Field
                                             as={Radio}
+                                            data-testid="yes-secondary-delivery"
```
lynzt, line 351: name this `has-secondary-delivery` to match the id... and also match Marjo's `has-secondary-pickup` in her PR...

lynzt (commented):
Can a SM edit the secondary destination address? if so, we should probably add a check that we can edit properly in `customerReviewsMoveDetailsAndEditsHHG`  (I know we're not doing that for other fields so maybe this would fall into making our e2e tests more robust and can be in a new story). Thoughts?

YanZ777:
> Can a SM edit the secondary destination address? if so, we should probably add a check that we can edit properly in `customerReviewsMoveDetailsAndEditsHHG` (I know we're not doing that for other fields so maybe this would fall into making our e2e tests more robust and can be in a new story). Thoughts?

Yea, that would be a good idea.

lynzt (approved):
LGTM 🚢

PR #6865, opened 2021-06-23
[MB-8311] Set move status to approvals requested when customer uploads amended orders

hannaliebl:
## Description

As a TOO
When a customer uploads a new set of orders
Then I need the move to be transitioned to the Approvals Requested status
So that I can notice and process changes to the move

## Reviewer Notes

The upload amended order functionality is not yet done, as best I am aware. As best I understand, though, this will cover the functionality we need to route a move to approvals requested status when a customer submits a move with amended orders. The status of that resubmitted move should still be "approved," and so I just used the existing function to set the status to approvals requested, but renamed it to make it more generic.

## Setup

```sh
make server_test
```
Maybe there is a better way to validate this, pending other work? Maybe I'm wrong and this can be verified in a different way?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8311) for this change


pkg/services/move.go
```diff
@@ -29,7 +29,7 @@ type MoveRouter interface {
 	Approve(move *models.Move) error
 	Cancel(reason string, move *models.Move) error
 	CompleteServiceCounseling(move *models.Move) error
-	SendToOfficeUserToReviewNewServiceItems(move *models.Move) error
+	SendToOfficeUser(move *models.Move) error
```
monfresh, line 32: Since we changed this, we need to regenerate the mocks: `make mocks_generate`

monfresh (approved):
This looks great! I think what we have here makes sense. The only thing I can think of that we should double check is if the move was already submitted to the TOO, they would be looking it up a "New Move" in the filters. But then if the customer uploads the amended orders before the TOO has had a chance to review the move, now the move won't show up under "New move" anymore, but instead under "Approvals Requested". I'm not sure if that makes a difference.

Ryan-Koch (approved):

PR #6866, opened 2021-06-23
MB-3728 validator pattern to UpdateMTOShipment

nelztruss:
## Description

Introduce the validator pattern to the `UpdateMTOShipment` service layer. Converts one shared validation (`checkStatus()`), and converts one Prime-specific validation (`checkAvailToPrime(...)`).

## Reviewer Notes

There are _a lot_ of business-logic validations that are occurring in the `handlers` layer, that would be more appropriate in the `services` layer, but I had to control my urge to go an refactor everything... After this PR the validator pattern will be available for any further refactorings.

### DIFF
--- a/pkg/handlers/primeapi/mto_shipment.go
+++ b/pkg/handlers/primeapi/mto_shipment.go
@@ -195,77 +195,57 @@ type UpdateMTOShipmentHandler struct {
 func (h UpdateMTOShipmentHandler) Handle(params mtoshipmentops.UpdateMTOShipmentParams) middleware.Responder {
 	logger := h.LoggerFromRequest(params.HTTPRequest)
 
-	mtoShipmentID := uuid.FromStringOrNil(params.MtoShipmentID.String())
+	// Validate payload, remove readOnly fields, convert to model
+	mtoShipment, fieldErrs := UpdateMTOShipmentModel(params.MtoShipmentID, params.Body)
+	if fieldErrs != nil {
+		logger.Error("primeapi.UpdateMTOShipmentHandler error - extra fields in request", zap.Error(fieldErrs))
 
-	// Check that MTO is available to prime
-	mtoAvailableToPrime, err := h.mtoShipmentUpdater.MTOShipmentsMTOAvailableToPrime(mtoShipmentID)
+		errPayload := payloads.ValidationError("Fields that cannot be updated found in input",
+			h.GetTraceID(), fieldErrs)
+
+		return mtoshipmentops.NewUpdateMTOShipmentUnprocessableEntity().WithPayload(errPayload)
+	}
+
+	// Get the associated shipment from the database
+	var dbShipment models.MTOShipment
+	err := h.DB().EagerPreload("PickupAddress",
+		"DestinationAddress",
+		"SecondaryPickupAddress",
+		"SecondaryDeliveryAddress",
+		"MTOAgents").Find(&dbShipment, params.MtoShipmentID)
 	if err != nil {
-		logger.Error("primeapi.UpdateMTOShipmentHandler error - MTO is not available to prime", zap.Error(err))
+		return mtoshipmentops.NewUpdateMTOShipmentNotFound().WithPayload(payloads.ClientError(handlers.NotFoundMessage, err.Error(), h.GetTraceID()))
+	}
+
+	// Validate further prime restrictions on model
+	mtoShipment, validationErrs := h.checkPrimeValidationsOnModel(mtoShipment, &dbShipment)
+	if validationErrs != nil && validationErrs.HasAny() {
+		logger.Error("primeapi.UpdateMTOShipmentHandler error - extra fields in request", zap.Error(validationErrs))
+
+		errPayload := payloads.ValidationError("Invalid data found in input",
+			h.GetTraceID(), validationErrs)
+
+		return mtoshipmentops.NewUpdateMTOShipmentUnprocessableEntity().WithPayload(errPayload)
+	}
+
+	logger.Info("primeapi.UpdateMTOShipmentHandler info", zap.String("pointOfContact", params.Body.PointOfContact))
+	mtoShipment, err = h.mtoShipmentUpdater.UpdateMTOShipmentPrime(params.HTTPRequest.Context(), mtoShipment, params.IfMatch)
+	if err != nil {
+		logger.Error("primeapi.UpdateMTOShipmentHandler error", zap.Error(err))
 		switch e := err.(type) {
 		case services.NotFoundError:
-			return mtoshipmentops.NewUpdateMTOShipmentNotFound().WithPayload(payloads.ClientError(handlers.NotFoundMessage, e.Error(), h.GetTraceID()))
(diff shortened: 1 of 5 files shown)


pkg/services/mto_shipment/validation.go
```diff
@@ -0,0 +1,71 @@
...
+	"context"
+
+	"github.com/gobuffalo/pop/v5"
+	"github.com/gobuffalo/validate/v3"
+
+	"github.com/transcom/mymove/pkg/models"
+	"github.com/transcom/mymove/pkg/services"
+)
+
+type validator interface {
```
nelztruss, line 13: Since the proposed set of `validator`s are limited to the `Update...` operation, was considering naming this `updateValidator`, but that pre-supposes that these are NOT useful for the `Create...` operation. Alternatively, if there is a demonstrated need for a different `createValidator` type, it would not be too much overhead to rename this interface since it is entirely internal.

pkg/services/mto_shipment/validation.go
```diff
@@ -0,0 +1,71 @@
...
+			}
+		}
+	}
+	if verrs.HasAny() {
+		result = services.NewInvalidInputError(newer.ID, nil, verrs, "Invalid input found while updating the shipment.")
+	}
+	return result
+}
+
+func checkStatus() validator {
```
monfresh, line 44: AFAIK, it should never be possible to set a shipment's status to DRAFT or SUBMITTED via the API. DRAFT is the default status for all new shipments, and once a shipment is created, it gets automatically set to the SUBMITTED status in the code, not because someone passed that status in. Also, the pattern I introduced was to get rid of the Status field entirely and instead use specific endpoints for transitioning shipments. For example, in the GHC API, we have these endpoints:
* `shipments/{shipmentID}/approve`
* `shipments/{shipmentID}/approve-diversion`
* `shipments/{shipmentID}/request-diversion`
* `shipments/{shipmentID}/request-cancellation`

And we have follow-up stories to update the Prime API to follow that pattern as well. Instead of duplicating logic for validating statuses, we want to use the functions in the `ShipmentRouter` service: https://github.com/transcom/mymove/blob/master/pkg/services/mto_shipment/shipment_router.go

pkg/services/mto_shipment/validation.go
```diff
@@ -0,0 +1,71 @@
...
+	return validatorFunc(func(_ context.Context, newer *models.MTOShipment, _ *models.MTOShipment) error {
+		verrs := validate.NewErrors()
+		if newer.Status != "" && newer.Status != models.MTOShipmentStatusDraft && newer.Status != models.MTOShipmentStatusSubmitted {
+			verrs.Add("status", "can only update status to DRAFT or SUBMITTED. use UpdateMTOShipmentStatus for other status updates")
```
monfresh, line 48: Unless things have changed, I believe that the only status the Prime API supports is canceling a shipment. If you look at `UpdateMTOShipmentStatus` in `prime.yaml` you'll see that it only allows `CANCELED` as the value of the `Status` field. Since that's the only transition that can be performed via the Prime API, we will replace the `UpdateMTOShipmentStatus` with a `shipments/{shipmentID}/cancel` endpoint that doesn't take any Body and just calls a service that cancels the shipment using the `Cancel` function in the ShipmentRouter.

pkg/handlers/primeapi/mto_shipment.go
```diff
@@ -195,77 +195,57 @@ type UpdateMTOShipmentHandler struct {
...
+	// Get the associated shipment from the database
+	var dbShipment models.MTOShipment
+	err := h.DB().EagerPreload("PickupAddress",
+		"DestinationAddress",
+		"SecondaryPickupAddress",
+		"SecondaryDeliveryAddress",
+		"MTOAgents").Find(&dbShipment, params.MtoShipmentID)
+	if err != nil {
+		return mtoshipmentops.NewUpdateMTOShipmentNotFound().WithPayload(payloads.ClientError(handlers.NotFoundMessage, err.Error(), h.GetTraceID()))
+	}
```
monfresh, line 218: I thought we wanted to move this kind of business logic out of handlers and into services. Is it the handler's responsibility to fetch objects from the DB and validate them?

nelztruss: You are correct, over time we would want to move this out of the handlers into a lower layer.

However, I didn't actually add this code, just un-indented it... (This file is best viewed with the "Hide whitespace changes" option.)

pkg/services/mto_shipment/validation.go
nelztruss: This check is just a re-statement of what previously was in `mto_shipment_updater.go` at lines ~114-116. For this PR, I was aiming at a structural-only (meaning no change to existing logical functionality) change.

pkg/services/mto_shipment/validation.go
nelztruss: As I was saying in a (below?) comment, I didn't change the logical operations that were performed, just changed the structure of *where* some checks happened, e.g. moving business logic checks from handlers to service layer (or in this specific case I moved a validation out of a mutation).

pkg/services/mto_shipment/validation.go
monfresh: Ah, I see. I usually go all out and fix things as I touch them, so I was reviewing this PR with that mindset, and also because there might be some overlap with this story: https://dp3.atlassian.net/browse/MB-8343

pkg/services/mto_shipment/validation.go
```diff
@@ -0,0 +1,71 @@
...
+		}
+		return verrs
+	})
+}
+
+func checkAvailToPrime(db *pop.Connection) validator {
+	return validatorFunc(func(_ context.Context, newer *models.MTOShipment, _ *models.MTOShipment) error {
+		var mto models.Move
```
monfresh, line 56: Could we name the variable `move` instead of `mto`? Last year, Reggie and I worked on consolidating a bunch of DB tables, but there are still references to the old tables, like MoveTaskOrder, or MTO. We want to stop using those and use "Move" instead.

pkg/services/mto_shipment/validation_test.go
```diff
@@ -0,0 +1,37 @@
+package mtoshipment
+
+import (
+	"context"
+
+	"github.com/transcom/mymove/pkg/models"
+)
+
+func (suite *MTOShipmentServiceSuite) TestUpdateValidations() {
```
monfresh, line 9: should this function be named "TestCheckStatus"?

monfresh: Ah, I see now. It's testing the validations when updating a shipment, and I presume more subtests will be added? Did you mean to also test `checkAvailToPrime`? Would it be less confusing to call this `TestShipmentValidations`?

eamahanna: I agree that a test should be added for checkAvailToPrime.

pkg/services/mto_shipment/validation.go
nelztruss: Will do!

pkg/services/mto_shipment/validation_test.go
nelztruss: Will do!

monfresh (commented):
LGTM. Thanks for making the changes and adding the test.

monfresh (approved):

PR #6870, opened 2021-06-24
[MB-8335] upload amended orders part 2

pearl-truss:
## Description

This PR added logic to the new UploadedAmendedOrders service so that when called with userUpload payload, it'll check the associated order for an existing UploadedAmendedOrders document. If it doesn't exist, it will create the document and attach the userUpload payload to the document. Otherwise it'll attach the incoming payload to the existing order.UploadedAmendedOrders document

## Reviewer Notes

I'm not 100% sure it was the right approach to create and save userUpload and upload records. I did it because of the inter-related relationships between the documents and uploads

## Setup

No where in the app calls this service yet. You can try testing this in postman against the `/orders/{ordersId}/upload_amended_orders` endpoint. The expected payload looks something like this

```sh
{
  "ordersID": "<uuid>",
  "ifMatch": "<etag>",
  "body": {
        "id": "<uuid>",
        "document": {
            "serviceMemberID": "<uuid>"
        },
        "upload": {
            "id":          "<uuid>",
            "url":         "https://file.test",
            "bytes":       123,
            "contentType": "application/pdf",
            "filename":    "file.pdf",
            "checksum":    "ImGQ2Ush0bDHsaQthV5BnQ=="
        },
        "uploadID":   "<uuid>",
        "uploaderID": "<uuid>"
    }
}
```

NOTE: I haven't been able to use this method because I haven't been able to get the if-match/e-tag returned on a GET request for an order. And you need an etag in order to submit the request

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8335) for this change


pkg/services/order/order_updater.go
```diff
@@ -364,5 +588,46 @@ func (f *orderUpdater) updateOrder(order models.Order) (*models.Order, uuid.UUID
 		return nil, uuid.Nil, transactionError
 	}
 
-	return &order, order.Moves[0].ID, nil
+	var moveID uuid.UUID
```
pearl-truss, line 591: I added this logic because I discovered it panicked if the order had no moves associated with it

swagger/internal.yaml
```diff
@@ -1658,6 +1661,39 @@ definitions:
...
+        example: c56a4180-65aa-42ec-a945-5fd21dec0538
+      document_id:
+        type: string
+        format: uuid
+        example: c56a4180-65aa-42ec-a945-5fd21dec0538
+      document:
+        $ref: '#/definitions/DocumentPayload'
+      upload:
+        $ref: '#/definitions/UploadPayload'
+      uploader_id:
```
mboren, line 1679: Do we need to get `uploader_id` as a parameter here? it seems like we could get it on the backend. I think `uploader_id` controls which user can access the file, and it seems like it would be better if that was not controlled by the client. I don't see how anybody could do anything nefarious with it, but it smells like a potential security issue.

pkg/services/order.go
```diff
@@ -20,7 +20,7 @@ type OrderFetcher interface {
 //OrderUpdater is the service object interface for updating fields of an Order
 //go:generate mockery --name OrderUpdater
 type OrderUpdater interface {
-	UploadAmendedOrders(order models.Order, payload *internalmessages.UploadPayload, eTag string) (*models.Order, uuid.UUID, error)
+	UploadAmendedOrders(orderID uuid.UUID, payload *internalmessages.UserUploadPayload, eTag string) (*models.Order, error)
```
mboren, line 23: I'm curious about the decision to use `UserUpload` instead of `Upload` for the payload. I think from an API perspective `Upload` fits better, but I don't know what problems you ran into.

jacquelineIO: `UserUpload` means there is a User ID that owns the `upload` and the `upload` needs to be created. If you are only returning uploads for viewing, then you can use uploads.

jacquelineIO: In general we expect the upload to be owned by a Prime and a User.

pkg/services/order/order_updater.go
```diff
@@ -177,6 +181,80 @@ func orderFromTOOPayload(existingOrder models.Order, payload ghcmessages.UpdateO
...
+func (f *orderUpdater) amendedOrderFromUserUploadPayload(order models.Order, payload *internalmessages.UserUploadPayload) (models.Order, error) {
+	var userUpload models.UserUpload
+	upload := models.Upload{
+		ID:          uuid.FromStringOrNil(payload.Upload.ID.String()),
+		UploadType:  models.UploadTypeUSER,
+		ContentType: *payload.Upload.ContentType,
+		Checksum:    payload.Upload.Checksum,
+	}
+	if payload.Upload.Filename != nil {
+		userUpload.Upload.Filename = *payload.Upload.Filename
```
jacquelineIO, line 193: `userUpload.Upload.Filename` and `upload.Filename` are redundant. Can you just use `upload.Filename` for all of these if statements. And then set `userUpload.Upload = upload` on line 209 where the `upload.ID` is being set.

pkg/services/order/order_updater.go
```diff
@@ -177,6 +181,80 @@ func orderFromTOOPayload(existingOrder models.Order, payload ghcmessages.UpdateO
...
+		}
+		savedAmendedOrdersDoc, err := f.saveDocumentForAmendedOrder(amendedOrdersDoc)
+		if err != nil {
+			return models.Order{}, err
+		}
+
+		if savedAmendedOrdersDoc.ID != uuid.Nil {
+			userUpload.DocumentID = &savedAmendedOrdersDoc.ID
+		}
+		if savedAmendedOrdersDoc != nil {
```
jacquelineIO, line 226: Need to check for Nil before using the struction on line 223. Basically, line 223 will fail if this is really Nil.

jacquelineIO: This check could probably be combined to 

```
if savedAmendedOrdersDoc != nil && savedAmendedOrdersDoc.ID != uuid.Nil {}
```

pkg/services/order/order_updater.go
```diff
@@ -177,6 +181,80 @@ func orderFromTOOPayload(existingOrder models.Order, payload ghcmessages.UpdateO
...
+		order.UploadedAmendedOrders = updatedAmendedDoc
+
+	} else {
+		if order.UploadedAmendedOrders != nil {
+			userUpload.Document = *order.UploadedAmendedOrders
+			userUpload.DocumentID = &order.UploadedAmendedOrders.ID
+		}
+
+		if order.UploadedAmendedOrders != nil {
+			order.UploadedAmendedOrders.UserUploads = append(order.UploadedAmendedOrders.UserUploads, userUpload)
```
jacquelineIO, line 250: Why are there 2 different `UserUploads = append(order.UploadedAmendedOrders.UserUploads, userUpload)` needed in the `if` statement but not in the `else`. I'm supicous that maybe 2 aren't needed in the `if` but there could be a reason for it. e.g., a new eagar fetch should load one of these, and maybe you purposely aren't wanting to do that. Just curious what the difference is here.

pkg/services/order/order_updater.go
```diff
@@ -286,6 +364,35 @@ func allowanceFromCounselingPayload(existingOrder models.Order, payload ghcmessa
 	return order
 }
 
+func (f *orderUpdater) saveDocumentForAmendedOrder(doc models.Document) (*models.Document, error) {
```
jacquelineIO, line 367: There is nothing specific happening here for an amended order, just looks like a `Document` save, is there another function already doing this?

jacquelineIO: If this is not already a function somewhere, this might should be moved to something more specific to the Docment model. I see there is a `FetchDocument` function, but not noticing a save.

jacquelineIO (commented):
I didn't get to run this locally. But did see some things in the code that might need to changed a little.

jacquelineIO:
This branched was merged into a new branch, and a new PR was created https://github.com/transcom/mymove/pull/6908. 
The new PR has been merged and this PR can be closed now.

PR #6881, opened 2021-06-25
pmohal CAC Update [Initial] / MB-8626

pmohal:
## Description

This is part of the procedure to do end to end testing which requires uploading a file generated with the help of a CAC [see https://github.com/transcom/mymove/wiki/use-mtls-with-cac ]

Please reference Jira: https://dp3.atlassian.net/browse/MB-8626


reggieriser (changes requested):
The migrations that are there look good and it ran fine locally, but it looks like you're missing the no-op migration file in `prd`.  I believe this means when the `prd` migrations run, they would fail with a missing file.  See [step 2 in this Wiki page](https://github.com/transcom/mymove/wiki/use-mtls-with-cac#uploading-cac-secure-migration).

Glad to pair if you need help on sorting this out.

reggieriser:
One other nit ... can you change the title of the issue to be more descriptive?  The titles end up in our deploy notes.  If there's a JIRA ticket for this, include that in the title as well.

pmohal:
@reggieriser Done and Done, let me know if you need me to do anything else

PR #6897, opened 2021-06-29
MB-8476 Wire up the Customer save button for uploading amended orders

felipe-lee:
## Description

GIVEN a customer has received updated orders
WHEN uploading updated documentation for their new orders
THEN the customer has a save button to click after uploading 
AND THEN receives a success message or an error message.

A note that the functionality to upload amended orders is still being worked on in #6908. Thus, in order to test this, you'll have to use the devseed data we set up (see setup).

For the backend, we added a new endpoint similar to the other ones that change the move's status. We also updated the move router service to account for the new statuses that the move can be in when it's being moved to a different status. We have DB saving at the handler level because it seems this service object doesn't have it and other handlers in this file also seem to handle their own saving. 

For the frontend, the save button now hooks up to the new endpoint. Tests are added to ensure that the save and cancel work as expected.

## Reviewer Notes

Please confirm that the `move_router_tests` have been written accurately. 

The frontend does not currently handle the uploading of documents

## Setup

1. Run server, customer client, and office client (you'll have to say yes to this one using a different port than the default).

    ```sh
    make server_run
    ```

    ```sh
    make client_run
    ```

    ```sh
    make office_client_run
    ```

2. Log in as a service member (use`hhg@multiple.pickup.amendedOrders.submitted (milmove)`)
3. On the home page, click on "Upload documents"
4. Click Save. There should be two documents uploaded already, and you won't be able to upload more as that's a WIP in another PR (#6908). 
5. Log in as a TOO. You should see the move as one that needs approval again.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8476) for this change


src/pages/MyMove/AmendOrders/AmendOrders.jsx
```diff
@@ -36,8 +36,8 @@ export const AmendOrders = ({ uploads, updateOrders, serviceMemberId, currentOrd
   };
   const handleUpload = (file) => {
     // TODO Temporarily using the original uploaded orders, will change to use amended orders once that is available
-    const documentId = currentOrders?.uploaded_orders?.id;
-    return createUploadForDocument(file, documentId);
+    const documentId = currentOrders?.uploaded_amended_orders?.id;
+    return createAmendedOrdersUploadForDocument(file, documentId);
```
mboren, line 40: you might already be aware of this, but just in case: we expect orders to have `null` for `uploaded_amended_orders` if a service member has not uploaded any orders before, so we're working on a separate endpoint for uploading amended order files right now. That endpoint will take `order_id` instead of `document_id` and will create the document if it doesn't exist.
Nothing you need to change for this, just wanted to make sure we're in sync

Ronolibert (changes requested):
When I to upload an amended order, the list does not update and when I click `Save` I am getting `orders.uploadAmendedOrders failed: Error: Required parameter ordersId is not provided`

felipe-lee:
> When I tried to upload an amended order with the `hhg@multiple.pickup.amendedOrders.submitted` account, the list does not update and when I click `Save` I am getting `orders.uploadAmendedOrders failed: Error: Required parameter ordersId is not provided`

Thanks for testing it out. I actually just updated the description for the PR because uploading amended orders isn't functional yet. Our pr is more for once you do have them, you can click submit and it gets routed correctly. 

As for the error on save, it's working for me, I wonder if it gets into a weird state when you try uploading orders and then saving.

mboren:
I suspect that there might be an issue with the way the devseed data is set up because the table should be showing just the file name and not the full path. I don't know if that has to do with this PR or is a pre-existing issue with test fixtures. I'm pretty sure this only affects the test data and not the actual functionality though, so not super worried about it.

image

looks like the `uploads` table is getting filled inconsistently:
image


also the table breaking for long filenames is 100% my fault. I'll file a bug for it.

src/pages/MyMove/AmendOrders/AmendOrders.test.jsx
```diff
@@ -65,9 +83,70 @@ describe('Amended Orders Upload page', () => {
...
+      userEvent.click(saveButton);
+
+      await waitFor(() => {
+        expect(mockPush).toHaveBeenCalledWith(generalRoutes.HOME_PATH);
+      });
+    });
+
+    it('shows an error if the API returns an error', async () => {
+      submitAmendedOrders.mockImplementation(() =>
+        // Disable this rule because makeSwaggerRequest does not throw an error if the API call fails
```
Ronolibert, line 123: 👍

Ronolibert (approved):
Ahh ok, I misunderstood what the scope of this card is. I tested it and now behaves as expected

src/pages/MyMove/AmendOrders/AmendOrders.jsx
mboren: to clarify: you might want coordinate with @jacquelineIO when you merge your PR because her changes in https://github.com/transcom/mymove/pull/6908 will conflict with this and should take precedence.

mboren (approved):
I think this looks good. The Happo diffs for the file uploader are weird, I was seeing exactly the same thing last night (https://ustcdp3.slack.com/archives/CP6F568DC/p1625185771091500), I think they can be ignored.

src/pages/MyMove/AmendOrders/AmendOrders.jsx
felipe-lee: Since our pr isn't actually meant to deal with handling uploads, maybe we should just undo this bit and leave it for the other pr to set up

PR #6899, opened 2021-06-29
MB-7702 Remove state machine functions from Move model

monfresh:
## Description

This PR completes the relocation of Move-related state machine functions from the Move model to the MoveRouter service. There shouldn't be any functional changes. We're just replacing calls to the Move model with calls to the MoveRouter.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7702) for this change


pkg/services/mto_shipment/mto_shipment_updater.go
```diff
@@ -399,17 +401,14 @@ func (f *mtoShipmentUpdater) updateShipmentRecord(dbShipment *models.MTOShipment
 				return err
 			}
 
-			// If approved, transition the status to APPROVALS REQUESTED
-			if move.Status == models.MoveStatusAPPROVED {
```
monfresh, line 403: @sandy-wright I removed this conditional to allow `f.moveRouter.SendToOfficeUser(&move)` to return an error in case the status is not "Approved".

nelztruss (commented):
One technique I'd suggest: when you're doing something like a find/replace (e.g. "context" -> "ctx"), isolate that into its own commit so that reviewers can walk through the individual commits and reduce the cognitive breadth of what they have to parse all in one go.

monfresh:
That's a great point. In hindsight, I should have also created separate commits for each function that was removed from the Move model.

Ryan-Koch (approved):

PR #6904, opened 2021-06-30
Nix node14

ahobson:
## Description

Configure nix to use node 14


monfresh (approved):
Thanks!

lynzt (approved):
Thanks for catching this!!

PR #4686, opened 2020-08-26
Kim mb 3716 update update shipment service

kimallen:
## Description

The service for updating the `mto_shipment` was prime-specific and included validations that only applied to the prime.  This PR moves all of the prime-specific validations to the primeapi `UpdateMTOShipmentHandler` and makes the Update service generic.  It also adds in the fields that are missing in the updater service.  
The service can now additionally update these fields:
- requestedPickupDate
- requestedDeliveryDate
- PrimeActualWeight
- FIrstAvailableDeliveryDate
- mtoAgents (adding new ones and updating existing ones)
-customerRemarks

This should allow the customer update mto_shipment api to use the service to update the shipment
## Reviewer Notes

[Discussion on whether services should be generic](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1597880829068300)

The MTOAgents are now set up so that they can be updated if they exist, and created if they are not.  In the future they should be separated out into their own service and have an eTag applied.

The Address field updates are not currently working, and Team Roci currently has a [story](https://dp3.atlassian.net/browse/MB-3691) in their sprint to fix it.
There will also need to be added logic to create a new address field if it does not exist.

Ideally the function in the service, `CalculateRequiredDeliveryDate`, which is used by both the service and the prime handler, should have its own tests to test its functionality.  Right now the functionality is tested by the tests moved into the handler.  Another story could be written for moving the tests of that function.

## Setup

run tests for `services/mto_shipment`
run tests for `handlers/primeapi/mto_shipment`

## Code Review Verification Steps

* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-3716) for this change
[address bug story](https://dp3.atlassian.net/browse/MB-3691)
[Discussion on whether services should be generic](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1597880829068300)

PR #6919, opened 2021-07-01
Fix bug that nullifies shipment ApprovedDate

monfresh:
In PR #4686, we introduced a bug that nullified a shipment's
`ApprovedDate` if a `ScheduledPickupDate` was included in the payload.
This affects approving diverted shipments because we determine which
endpoint to hit based on the presence of the `ApprovedDate`.

When a shipment is diverted, the original shipment that was created by
the customer gets updated by the Prime to set its `Diversion` field to
`true`. In addition, the Prime creates a new shipment. Both of these
shipments have the `SUBMITTED` status, but only the new shipment should
have service items automatically created for it.

To distinguish between these two scenarios, we created separate
endpoints in the GHC API: `shipments/{shipmentID}/approve` for new
shipments, and `shipments/{shipmentID}/approve-diversion` for existing
shipments that had previously been approved by the TOO. The way the
front end determines which endpoint to call is by looking at the
shipment's `ApprovedDate` because only the previously-existing shipment
would have that set. The new shipment hasn't been approved yet by the
TOO.

But because of this bug, both shipments end up with a nil
`ApprovedDate`, and the frontend sends both to the `approve` endpoint,
which creates service items, and so the existing shipment ends up with
duplicate service items.

## Description

## Reviewer Notes

## Setup

```sh
echo "Code goes here"
```

## References

## Screenshots

### DIFF
--- a/pkg/services/mto_shipment/mto_shipment_updater_test.go
+++ b/pkg/services/mto_shipment/mto_shipment_updater_test.go
@@ -233,7 +233,7 @@ func (suite *MTOShipmentServiceSuite) TestMTOShipmentUpdater() {
 	})
 
 	suite.T().Run("Successful update to a minimal MTO shipment", func(t *testing.T) {
-		// Minimal MTO Shipment has only pickup address created by default
+		// Minimal MTO Shipment has no associated addresses created by default.
 		// Part of this test ensures that if an address doesn't exist on a shipment,
 		// the updater can successfully create it.
 		oldShipment := testdatagen.MakeDefaultMTOShipmentMinimal(suite.DB())
@@ -258,7 +258,6 @@ func (suite *MTOShipmentServiceSuite) TestMTOShipmentUpdater() {
 			ScheduledPickupDate:              &scheduledPickupDate,
 			RequestedDeliveryDate:            &requestedDeliveryDate,
 			ActualPickupDate:                 &actualPickupDate,
-			ApprovedDate:                     &firstAvailableDeliveryDate,
 			PrimeActualWeight:                &primeActualWeight,
 			PrimeEstimatedWeight:             &primeEstimatedWeight,
 			FirstAvailableDeliveryDate:       &firstAvailableDeliveryDate,
@@ -275,7 +274,6 @@ func (suite *MTOShipmentServiceSuite) TestMTOShipmentUpdater() {
 		suite.True(scheduledPickupDate.Equal(*newShipment.ScheduledPickupDate))
 		suite.True(requestedDeliveryDate.Equal(*newShipment.RequestedDeliveryDate))
 		suite.True(actualPickupDate.Equal(*newShipment.ActualPickupDate))
-		suite.True(firstAvailableDeliveryDate.Equal(*newShipment.ApprovedDate))
 		suite.True(firstAvailableDeliveryDate.Equal(*newShipment.FirstAvailableDeliveryDate))
 		suite.True(primeEstimatedWeightRecordedDate.Equal(*newShipment.PrimeEstimatedWeightRecordedDate))
 		suite.Equal(primeEstimatedWeight, *newShipment.PrimeEstimatedWeight)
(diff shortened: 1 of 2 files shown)


pkg/services/mto_shipment/mto_shipment_updater_test.go
```diff
@@ -290,7 +290,7 @@ func (suite *MTOShipmentServiceSuite) TestMTOShipmentUpdater() {
 		suite.Equal(newPickupAddress.ID, *newShipment.PickupAddressID)
 		suite.Equal(secondaryPickupAddress.ID, *newShipment.SecondaryPickupAddressID)
 		suite.Equal(secondaryDeliveryAddress.ID, *newShipment.SecondaryDeliveryAddressID)
-
+		suite.NotNil(newShipment.ApprovedDate)
```
sandy-wright, line 293: You might want to use "NotEmpty" which checks if it's nil _or_ a zero value (0001-01-01). Either way though.

pkg/services/mto_shipment/mto_shipment_updater_test.go
```diff
@@ -235,10 +235,12 @@ func (suite *MTOShipmentServiceSuite) TestMTOShipmentUpdater() {
 		// the updater can successfully create it.
 		oldShipment := testdatagen.MakeMTOShipmentMinimal(suite.DB(), testdatagen.Assertions{
 			MTOShipment: models.MTOShipment{
-				Status: models.MTOShipmentStatusDraft,
+				Status: models.MTOShipmentStatusApproved,
```
sandy-wright, line 238: I'm not sure you should change this test - making sure these updates work with a Draft is important too. A new test might be the best way to go about it.

monfresh: That's a good point, although I don't think this test is accurately reflecting updates that would happen on a draft shipment. For example, it is updating fields like Diversion, PrimeEstimatedWeightRecordedDate, and Counselor Remarks. Would those fields ever be passed in to update a draft shipment?

sandy-wright: That's a good point, this test definitely isn't realistic. I don't think you need to refactor it as part of this bug fix, unless you want to, but I also don't think it would be a good idea to drop coverage of the DRAFT status. A minimal test that just updates the `scheduledPickupDate` of an approved shipment would be enough for this.

duncan-truss: I agree this test is calling UpdateMTOShipmentCustomer which can not update most of these fields but would likely be in draft status.  We should be testing UpdateMTOShipmentPrime I think.

monfresh: Here are some observations after looking at the code more closely:

* The code under test does not take into account the status of the shipment, so currently, a shipment can be updated regardless of its current status. The status has no bearing on the test. The fact that changing the status in this test didn't break the test confirms this. Therefore,  we are not losing any coverage.
* If the current shipment status **should** affect which fields are allowed to be updated, then that is an existing bug and we should open a ticket to fix it. AFAICT, we are not checking for the status in the handlers either.
* The `UpdateMTOShipmentOffice`, `UpdateMTOShipmentCustomer` and `UpdateMTOShipmentPrime` functions all do exactly the same thing, except that the Prime one also includes a validation to make sure the move is available to the prime.
* Tests should only include details that are relevant to the test. The fact that the test specified the draft status was misleading and confusing because it made us think the Draft status was relevant to the test. 
* The comment at the top of this test indicates this test is making sure that if a shipment does not have an address, one can be created for it. Also note the comment is wrong about `MakeMTOShipmentMinimal` having a pickup address by default. It does not have any associated addresses at all. I'll fix that.
* We don't have any unit tests for `UpdateMTOShipmentOffice` and `UpdateMTOShipmentPrime`, and `UpdateMTOShipmentPrime` is being mocked in the handler test.

Also, I just remembered that I had already removed the status from this test in another PR I worked on this week, and that has been merged already.

Since we don't have coverage for `UpdateMTOShipmentPrime`, I will leave this test as is, except for updating the comment. I will write a ticket to make sure we add the missing tests, and to make sure we only pass in fields that are relevant to each scenario.

I will write a separate test to make sure the ApprovedDate doesn't get nullified when passing in the ScheduledPickupDate.

monfresh: On second thought, what we're really trying to prevent is that the ApprovedDate should never get nullified when updating a shipment, regardless of which fields are passed in. So, for each of the functions that are used for updating a shipment, we want to include a test that passes in all the possible fields.

monfresh:
@sandy-wright @duncan-truss This is ready for another review please. I added two new tests to make sure the ApprovedDate doesn't get nullified: one for `UpdateMTOShipmentCustomer`, and one for `UpdateMTOShipmentPrime`. Can you please double check the fields for each to see if I am passing in any that should not be allowed, or if any fields are missing?

I also noticed another bug where we are allowing the `PrimeEstimatedWeightRecordedDate` to be passed in the payload in the shipment updater service. I will fix that in a separate PR.

pkg/services/mto_shipment/mto_shipment_updater_test.go
```diff
@@ -287,7 +285,45 @@ func (suite *MTOShipmentServiceSuite) TestMTOShipmentUpdater() {
...
+			DestinationAddressID:     &newDestinationAddress.ID,
+			PickupAddress:            &newPickupAddress,
+			PickupAddressID:          &newPickupAddress.ID,
+			SecondaryPickupAddress:   &secondaryPickupAddress,
+			SecondaryDeliveryAddress: &secondaryDeliveryAddress,
+			RequestedPickupDate:      &requestedPickupDate,
+			RequestedDeliveryDate:    &requestedDeliveryDate,
+			CustomerRemarks:          &customerRemarks,
+			CounselorRemarks:         &counselorRemarks,
+		}
```
duncan-truss, line 321: This looks good the only fields we could possibly add are Agents, ShipmentType, and Status that I see in the customer params request but I'm okay with what's here now.

sandy-wright (approved):
LGTM!

pkg/services/mto_shipment/mto_shipment_updater_test.go
```diff
@@ -422,6 +458,63 @@ func (suite *MTOShipmentServiceSuite) TestMTOShipmentUpdater() {
...
+		scheduledPickupDate := time.Date(2019, time.March, 17, 0, 0, 0, 0, time.UTC)
+		requestedDeliveryDate := time.Date(2019, time.March, 30, 0, 0, 0, 0, time.UTC)
+		updatedShipment := models.MTOShipment{
+			ID:                         oldShipment.ID,
+			DestinationAddress:         &newDestinationAddress,
+			DestinationAddressID:       &newDestinationAddress.ID,
+			PickupAddress:              &newPickupAddress,
+			PickupAddressID:            &newPickupAddress.ID,
+			SecondaryPickupAddress:     &secondaryPickupAddress,
+			SecondaryDeliveryAddress:   &secondaryDeliveryAddress,
```
duncan-truss, line 492: The prime endpoint says addresses can only be updated so this may not be realistic if they don't exist on the initial MTOShipment but Sandy would know best.

duncan-truss (approved):
Awesome thanks for tracking this down!

pkg/services/mto_shipment/mto_shipment_updater_test.go
sandy-wright: Oh shoot we need to fix that description - the Prime can indeed use this endpoint to add new addresses.

PR #6906, opened 2021-06-30
MB-8107: Adds helper text to Customer home screen when user has unapproved amended orders

kylehilltruss:
## Description

Design requested that a helper be added to the Customer home page, which displays once a user has uploaded amended orders documents until the TOO has approved the amendments. This PR adds that helper and the conditional logic to display it.

There is a storybook entry added for the helper, and also the state within in the Customer home page that would display this helper.

## Reviewer Notes

Ensure that the added test cases are useful -- the Home page component's tests are written with enzyme, which I haven't used in quite a while. (It didn't seem worth the effort to convert it to RTL at this time.)

## Setup

Right now, this change can only be seen in the story added to storybook (`Customer Components` / `Pages` / `Home`). 

Once the backend-focused branch *mb-8476-customer-has-a-save-button-for-uploading-amended-orders* is merged into master, you'll be able to upload and save amended orders as a Customer with a submitted move and see the helper message display on the home page. 

```sh
make db_dev_e2e_populate
make client_run
make server_run
```

(For the sake of clarity, this PR can be merged safely as-is; it will just have no practical effect until the complementary branch is merged.)

## Code Review Verification Steps

* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* Closes [MB-8107](https://dp3.atlassian.net/browse/MB-8107).
* Requires [MB-8476](https://dp3.atlassian.net/browse/MB-8476) to be completed before this will practically work.

## Screenshots

There's not really a "before"; [here's the after](https://user-images.githubusercontent.com/83614364/124020251-08378c00-d9b8-11eb-9dfc-1a6ba28435e7.png).

PR #6922, opened 2021-07-01
[MB-8344] Show alert on home screen after saving amending orders

mboren:
## Description

After a service member uploads files for amended orders, and until the TOO approves the new orders, the service member should see an alert on the home screen to reassure them that they've successfully submitted the new documents.

## Reviewer Notes

I'm piggybacking on @kylehilltruss's [branch](https://github.com/transcom/mymove/pull/6906) , in which he creates some help text that displays in the same scenario that this alert is displayed in.

The formatting doesn't quite match the designs; there's a bigger gap between the text and the top of the alert box than there should be. I'm guessing this is something to do with my use of `<p>` tags. I'm still fiddling with this, but any suggestions are welcome.

Should I move the alert into its own little component in a separate file, like the Helpers are?

Are the tests adequate?

I used a `data-testid` to select the alert because I don't think I have anything else to query on that would distinguish it from other `<Alert>`s we may put on the page in the future. If this is not the best way to do this, let me know.

## Setup

We can't see this in the app yet (devseed data and the ability to upload amended orders are both in progress right now). However, you can see it in storybook:
http://localhost:6006/?path=/story/customer-components-pages-home--amended-orders

and compare against the designs:
- [mobile](https://zpl.io/V4NRGqy)
- [desktop](https://zpl.io/VKd7B5A)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8344) for this change

### mobile
image

### tablet
image

### DIFF
--- a/src/pages/MyMove/Home/index.jsx
+++ b/src/pages/MyMove/Home/index.jsx
@@ -139,7 +139,7 @@ export class Home extends Component {
             The transportation office will review your new documents and update your move info. Contact your movers to
             coordinate any changes to your move.
           </p>
-          You don&apos;t need to do anything else in MilMove.
+          <p>You don&apos;t need to do anything else in MilMove.</p>
         </Alert>
       );
     }
--- a/src/pages/MyMove/Home/index.jsx
+++ b/src/pages/MyMove/Home/index.jsx
@@ -46,6 +46,7 @@ import ConnectedFlashMessage from 'containers/FlashMessage/FlashMessage';
 import { HistoryShape, MoveShape, MtoShipmentShape, OrdersShape, UploadShape } from 'types/customerShapes';
 import requireCustomerState from 'containers/requireCustomerState/requireCustomerState';
 import { profileStates } from 'constants/customerStates';
+import Alert from 'shared/Alert';
 
 const Description = ({ className, children, dataTestId }) => (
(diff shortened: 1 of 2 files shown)


kylehilltruss (approved):
Looks good by me 👍 

(Edit: Want to run this by someone from @transcom/truss-design?)

src/pages/MyMove/Home/index.jsx
```diff
@@ -130,6 +130,21 @@ export class Home extends Component {
     return 'Set up your shipments';
   }
 
+  renderAlert = () => {
+    if (this.hasUnapprovedAmendedOrders) {
+      return (
+        <Alert type="success" slim data-testid="unapproved-amended-orders-alert">
+          <p>
```
Shkeating, line 137: take the p element out of here, its adding extra space at the top and bottom
image

Shkeating (changes requested):

src/pages/MyMove/Home/index.jsx
Shkeating: for the second l'd leave it so you get the correct amount of space between the paragraphs

mboren: amazing. i tried every other conceivable permutation of `<p>` and `<br/>` except that one. it works!
image

Shkeating (approved):
beautiful!

chef's kiss

PR #6935, opened 2021-07-06
Fix truncation in db_e2e_up

monfresh:
In PR #6812, we cleaned up the Makefile by extracting some common setup,
like checking the app and truncating the DB, into separate, reusable
commands. Part of that was extracting the test DB truncation into a
`db_test_truncate` command that only truncated some tables. Earlier,
PR #6650 had also introduced a `db_test_truncate` command that calls
`scripts/db-truncate`, which truncates ALL the tables. For e2e tests,
we only want to truncate a certain set of tables, but for server tests,
we want to truncate everything.

While merging #6650, which happened after #6812 landed, we missed this
distinction between server and e2e tests, and so running
`make e2e_test_fresh` or `make db_e2e_init` or `make db_e2e_up` started
failing because they generate data that depends on certain tables
existing.

To fix this, we are now adding a separate `db_e2e_test_truncate` that
is only used for e2e tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -661,6 +661,11 @@ db_test_truncate:
 	@echo "Truncating ${DB_NAME_TEST} database..."
 	DB_PORT=$(DB_PORT_TEST) DB_NAME=$(DB_NAME_TEST) ./scripts/db-truncate
 
+.PHONY: db_e2e_test_truncate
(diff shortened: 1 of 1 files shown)


ahobson (approved):
🚀

PR #6943, opened 2021-07-07
Updated alpine from 3.12.3 to 3.13.5

reggieriser:
## Description

Draft PR to see if an upgrade to alpine 3.13.5 will pass checks.  It has previously failed an ECR scan, but I saw some indications online that it may pass now.  This may be a good intermediate step until we can get to 3.14.0 (ECR fails for it too).

### DIFF
--- a/Dockerfile.migrations
+++ b/Dockerfile.migrations
@@ -1,4 +1,4 @@
-FROM alpine:3.12.3
+FROM alpine:3.13.5
 
 # hadolint ignore=DL3017
 RUN apk upgrade --no-cache busybox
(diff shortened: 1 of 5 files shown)


reggieriser:
Looks like 3.13.5 still fails an ECR scan with a false positive for now.  I was hoping 3.13.x would work even if 3.14.x was failing in our last dependabot update.  Closing this while we wait for a fix.

Related:
- https://twitter.com/ariadneconill/status/1399554672571072513
- https://github.com/aws/containers-roadmap/issues/798

PR #6946, opened 2021-07-08
MB-8645 Remove image from crating sevice items

monfresh:
## Description

We originally thought we might need to support uploading images for
crating service items, but that's not the case anymore, so we're
removing references to crating images from the front-end. The backend
portion never got implemented, so there's nothing to remove there.

Since we were not using the `ServiceItemTable`, I deleted it, and renamed `ServiceItemTableHasImg` to `ServiceItemTable`. I did this in a separate commit in case that was not the right thing to do.

## Reviewer Notes

As a TOO, view the Move with Locator 4RGFTQ and click on the MTO tab.
- [ ] Verify that the Domestic crating row appears correctly

Run `make storybook`
- [ ] Verify that Tables -> Service Item Tables shows the correct design

## Setup

```sh
make server_run
make db_dev_e2e_populate
make office_client_run
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8645) for this change


src/components/Office/ServiceItemDetails/ServiceItemDetails.module.scss
```diff
@@ -15,18 +15,13 @@
   @include u-margin-top(2);
 }
 
-.detailImage {
+.detailCrating {
```
monfresh, line 18: Now that we don't have images, is this style still needed for crating items?

duncan-truss: I checked and the other service items all have a `<div>` so I think we should keep that at least.  You should be able to remove the `@include u-display(flex)` because we don't need to worry about the horizontal layout with the image removed.  The `.detailCrating` is still providing a needed scope otherwise we'd need to give the paragraph a class so it doesn't apply to all global paragraphs.

src/components/ServiceItemTable/index.jsx
```diff
@@ -1,45 +1,148 @@
...
-);
+import { SERVICE_ITEM_STATUS } from '../../shared/constants';
+import { MTOServiceItemCustomerContactShape, MTOServiceItemDimensionShape } from '../../types/order';
+
+import styles from './index.module.scss';
+
+import { formatDateFromIso } from 'shared/formatters';
+import ServiceItemDetails from 'components/Office/ServiceItemDetails/ServiceItemDetails';
+
+const ServiceItemTable = ({
```
duncan-truss, line 15: Ahh I see this is the former contents of the ServiceItemTableHasImg component.  Should we move this into the `src/components/Office/` folder? I don't think it's being used anywhere else, it just got created before that organization.

duncan-truss (approved):
🚀

src/components/Office/ServiceItemDetails/ServiceItemDetails.test.jsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
```
jenniferlynparsons, line 2: Wooo! RTL!

src/components/ServiceItemTable/index.jsx
jenniferlynparsons: If we do this, can we switch from the `index.js` pattern to the "named after the component" pattern as well? It might mean updating a few imports.

jenniferlynparsons (commented):
Working as expected so I can approve, based on changes Duncan suggested.

brooksr4 (approved):
Design looks right

PR #6957, opened 2021-07-08
[MB-8505] Add Cypress test to cover the amended orders flow

pearl-truss:
## Description

This PR adds a cypress test to Orders.js to test the amended orders flow. It starts with a customer that has a previously submitted move and ensures the the upload documents link is visible to them and that they upload amended orders and return to the starting page with a success message

## Reviewer Notes

Is there anything missing from the amended orders flow that I should be testing for?

Also, looking into if I need to use cy.intercept for the api call to /amendOrders

## Setup

```sh
make e2e_test
```
Once cypress spins up you should navigate to the test under mymove/orders

If you have issues running the cypress test you can also try running 

```sh
make e2e_test_fresh
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8505) for this change
* [docs for running cypress test](https://github.com/transcom/mymove/wiki/run-e2e-tests) 
* [Cypress Testing Strategy](https://docs.cypress.io/guides/guides/network-requests#Testing-Strategies) docs

### DIFF
--- a/cypress/integration/mymove/orders.js
+++ b/cypress/integration/mymove/orders.js
@@ -3,6 +3,10 @@ describe('orders entry', function () {
     cy.prepareCustomerApp();
   });
 
+  beforeEach(() => {
+    cy.logout();
+  });
+
   it('will accept orders information', function () {
     // needs@orde.rs
     cy.apiSignInAsPpmUser('feac0e92-66ec-4cab-ad29-538129bf918e');
@@ -48,6 +52,28 @@ describe('orders entry', function () {
     cy.contains('Continue Move Setup').click();
     cy.location().should((loc) => {
       expect(loc.pathname).to.eq('/orders/upload');
+    });
+  });
+
+  it('will allow amended orders upload', function () {
+    const userId = '6016e423-f8d5-44ca-98a8-af03c8445c94';
+    cy.apiSignInAsUser(userId);
+    cy.contains('Upload documents').click();
+
+    cy.location().should((loc) => {
+      expect(loc.pathname).to.eq('/orders/amend');
+    });
+
+    // cy.get('.filepond--label-action').click();
+    cy.upload_file('.filepond--root', 'top-secret.png');
+    // cy.get('button').contains('Save').click();
+    cy.nextPage();
+
+    cy.location().should((loc) => {
+      expect(loc.pathname).to.eq('/');
+    });
+    cy.get('.usa-alert--success').within(() => {
+      cy.contains('The transportation office will review your new documents');
     });
   });
 });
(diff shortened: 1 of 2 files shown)


jacquelineIO (commented):
This looks good to me. I can't answer this question for 
> Also, looking into if I need to use cy.intercept for the api call to /amendOrders
so it might be good to ask in frontend channel about this.

jacquelineIO (approved):

pearl-truss:
> This looks good to me. I can't answer this question for
> 
> > Also, looking into if I need to use cy.intercept for the api call to /amendOrders
> > so it might be good to ask in frontend channel about this.

Thanks for reviewing! Should I change the name of the uploaded file as well?

And since posting that review note, I read in the docs that `cy.intercept` is for stubbing out API calls which they suggest to use sparingly and only for critical paths because it can slow down the test. And I've added @kylehilltruss as a reviewer so I'll wait to see if I should change this or not

cypress/integration/mymove/orders.js
```diff
@@ -50,4 +56,26 @@ describe('orders entry', function () {
...
+    const userId = '6016e423-f8d5-44ca-98a8-af03c8445c94';
+    cy.apiSignInAsUser(userId);
+    cy.contains('Upload documents').click();
+
+    cy.location().should((loc) => {
+      expect(loc.pathname).to.eq('/orders/amend');
+    });
+
+    cy.upload_file('.filepond--root', 'top-secret.png');
+    cy.get('[data-filepond-item-state="processing-complete"]', { timeout: fileUploadTimeout }).should('have.length', 1);
```
duncan-truss, line 70: I'm not seeing this selector anywhere but I see it's the same one used by the other Cypress upload tests.  I might have a preference to test for asserting that the file information exists like `1 FILES UPLOADED` or the name of the file with the size.  It might make things easier to debug.

I think using cy.intercept and cy.wait is preferable over setting custom timeout overrides like this is doing.  It may pass locally but can become flakier in CI it seems. You're right you can use that intercept() to return mocked response payloads but we are using the real responses in all our Cypress tests I think.  Looking in the network tab it looks like there is a patch and a get request that are triggered on amended orders upload.

duncan-truss (approved):
This looks good, the small file being uploaded and the small amount of api data being sent/requested may mean the wait statements aren't necessary yet.  We can keep an eye on it to see how it performs under repeated CI.

cypress/integration/mymove/orders.js
pearl-truss: Thanks! I switched to using cy.intercept and cy.wait and got rid of the timeout. However, I haven't figured out how to assert that file information exists.

PR #6974, opened 2021-07-09
MB-8733 Fix Console Warning on Home Page

YanZ777:
Give a fallback value for uploaded amended order documents

## Description

There won't always be a value for the uploaded amended orders. It's marked as required for the props, but give it a default value to fall back to if there aren't uploaded amended orders.

## Reviewer Notes

Took off the required for `orders` and `uploadedAmendedOrderDocuments` since they technically aren't required for the home component (it can be in different states that don't have `orders` or `uploadedAmendedOrderDocuments`). 
Used default prop value for `uploadedAmendedOrderDocuments` rather than having it get defaulted in the `mapStateToProps`.
Left the default for orders inside of `mapStateToProps` since the `selectCurrentOrders` returns null instead of an empty object, which causes the site to crash.

## Setup

```sh
make client_run
make server_run
```

Go through the workflow up to the home page screen and verify that there are no console warnings. 

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8733) for this change

## Screenshots

### DIFF
--- a/src/pages/MyMove/Home/index.jsx
+++ b/src/pages/MyMove/Home/index.jsx
@@ -136,9 +136,7 @@ export class Home extends Component {
         <Alert type="success" slim data-testid="unapproved-amended-orders-alert">
           The transportation office will review your new documents and update your move info. Contact your movers to
           coordinate any changes to your move.
-          <br />
-          <br />
-          You don&apos;t need to do anything else in MilMove.
+          <p>You don&apos;t need to do anything else in MilMove.</p>
         </Alert>
       );
     }
--- a/src/pages/MyMove/Home/index.jsx
+++ b/src/pages/MyMove/Home/index.jsx
@@ -454,7 +454,7 @@ const mapStateToProps = (state) => {
     isProfileComplete: selectIsProfileComplete(state),
     orders: selectCurrentOrders(state) || {},
     uploadedOrderDocuments: selectUploadsForCurrentOrders(state),
-    uploadedAmendedOrderDocuments: selectUploadsForCurrentAmendedOrders(state),
+    uploadedAmendedOrderDocuments: selectUploadsForCurrentAmendedOrders(state) || [],
     serviceMember,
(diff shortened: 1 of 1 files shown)


felipe-lee (commented):
I feel like if we're going to provide a default value, we might as well turn off the required bit and give it a default in `defaultProps` instead of setting a default value in the `mapStateToProps` function.

Same with `orders`. That one we both set a default in the `mapStateToProps` and the `defaultPropTypes`, though I know that's not what this pr is for, just happened to notice it.

YanZ777:
> I feel like if we're going to provide a default value, we might as well turn off the required bit and give it a default in `defaultProps` instead of setting a default value in the `mapStateToProps` function.
> 
> Same with `orders`. That one we both set a default in the `mapStateToProps` and the `defaultPropTypes`, though I know that's not what this pr is for, just happened to notice it.

Yea, I think that makes more sense. I switched it to use the default props rather than defaulting in `mapStateToProps` because those props aren't technically required by the component itself (it's possible to have a home page state with no `orders` or `uploadedAmendedOrderDocuments`).

src/pages/MyMove/Home/index.jsx
```diff
@@ -452,9 +453,9 @@ const mapStateToProps = (state) => {
   return {
     currentPpm: selectCurrentPPM(state) || {},
     isProfileComplete: selectIsProfileComplete(state),
-    orders: selectCurrentOrders(state) || {},
+    orders: selectCurrentOrders(state),
```
felipe-lee, line 456: Since the default it was using before was an empty object, I think line 443 might need to be changed from `null` to `{}`

felipe-lee (approved):

src/pages/MyMove/Home/index.jsx
YanZ777: I actually put this back in because `selectCurrentOrders` returns null if there are no orders, which causes the page to crash.
I left the default prop as {} though, which should not cause problems.

PR #6985, opened 2021-07-13
MB-8507 move edit orders out of scenes

YanZ777:
## Description

Split out the `EditOrders` file so that there's an `EditOrdersForm` and an `EditOrders` page. Deleted the `EditOrders` file. 
There should be no functional changes and everything should operate as they did before. 

## Reviewer Notes

These changes don't cover:
- adding tests
- adding storybook
- refactoring this page to use formik
- fixing the various bugs with editing orders

Additionally, I took out the formValues prop from being sent over to the `EditOrdersForm` because it wasn't being used in `EditOrdersForm`. The currentOrders were being used as the intialValues instead. I decided to leave it as is since when this page gets migrated to formik, the currentOrders will be used. 

I left some of the props as `PropTypes.shape({})` because those particular props will be removed once migrated over to formik.

I've added a new shape type for the uploaded orders for reusability. 

I've changed the home page to lazy load the edit orders page.

## Setup

```sh
make client_run
make server_run
```
Make sure that the edit orders page is still accessible (even if editing doesn't work)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8507) for this change

## Screenshots


src/pages/MyMove/EditOrders.jsx
```diff
@@ -29,91 +18,10 @@ import {
...
-
-EditOrdersForm = withContext(
-  reduxForm({
-    form: editOrdersFormName,
-  })(EditOrdersForm),
-);
+import EditOrdersForm from 'components/Customer/EditOrdersForm/EditOrdersForm';
+import { OrdersShape, HistoryShape } from 'types/customerShapes';
+import { EntitlementShape, ExistingUploadsShape } from 'types';
+import 'scenes/Review/Review.css';
```
kylehilltruss, line 24: Q: do we want to migrate the CSS as well, or is that outside the scope of this ticket?

src/pages/MyMove/EditOrders.jsx
```diff
@@ -147,22 +62,23 @@ class EditOrders extends Component {
   };
 
   submitOrders = (fieldValues) => {
-    const { setFlashMessage, entitlement } = this.props;
+    const { setFlashMessage, entitlement, updateOrders, currentOrders, spouseHasProGear, history } = this.props;
 
     let entitlementCouldChange = false;
+    const modifiedFieldValues = fieldValues;
```
kylehilltruss, line 68: If `fieldValues` is an object, I think `modifiedFieldValues` is just being set by reference here, and I think modifying the new object might also modify the underlying one.

If `fieldValues` is just a flat object, you can make a quickie copy with `const modifiedFieldValues = { ...fieldValues }`.

kylehilltruss (commented):
Looks good, aside from a question (to which the answer can definitely be "out of scope") and one small thing.

src/pages/MyMove/EditOrders.jsx
YanZ777: I left the CSS as is since it looked like there were quite a few other files dependent on it. I also didn't feel super comfortable removing `Review.css` here either. 
I think once all the other. files are migrated, we should consider migrating.

src/pages/MyMove/EditOrders.jsx
YanZ777: Unfortunately, it's not a flat object.

kylehilltruss (approved):
LGTM!

PR #6987, opened 2021-07-13
[MB-8757] Update golangci-lint to latest version

reggieriser:
## Description

This PR updates our `golangci-lint` pre-commit hook to the latest version (1.41.1).  The updated linter end up flagging some new potential issues.  I have attempted to address those in this PR.  There is one linter suppression that needs review and approval.

## Reviewer Notes

I have put inline comments on each of the issues describing the linter feedback and how I addressed it.

## Setup

Execute `pre-commit run -a golangci-lint` to run all files through the updated linter.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8757) for this change

### DIFF
--- a/pkg/iampostgres/iampostgres_test.go
+++ b/pkg/iampostgres/iampostgres_test.go
@@ -1,6 +1,7 @@
 package iampostgres
 
 import (
+	"errors"
 	"testing"
 	"time"
 
@@ -14,9 +15,14 @@ type RDSUTest struct {
 	passes []string
 }
 
-func (r RDSUTest) GetToken(endpoint string, region string, user string, iamcreds *credentials.Credentials) (string, error) {
+func (r *RDSUTest) GetToken(endpoint string, region string, user string, iamcreds *credentials.Credentials) (string, error) {
+	if len(r.passes) == 0 {
+		return "", errors.New("no passwords to rotate")
+	}
+
+	// Rotate the slice: first item goes to back of slice
 	pass := r.passes[0]
-	r.passes = append(r.passes[:0], r.passes[1:]...)
+	r.passes = append(r.passes[1:], pass)
 
 	return pass, nil
 }
@@ -33,7 +39,7 @@ func TestEnableIamNilCreds(t *testing.T) {
 
 	EnableIAM("server", "8080", "us-east-1", "dbuser", "***",
 		nil,
-		rdsu,
+		&rdsu,
 		tmr,
 		logger,
 		shouldQuitChan)
@@ -61,7 +67,7 @@ func TestGetCurrentPassword(t *testing.T) {
 
 	EnableIAM("server", "8080", "us-east-1", "dbuser", "***",
 		credentials.NewStaticCredentials("id", "pass", "token"),
-		rdsu,
+		&rdsu,
 		tmr,
 		logger,
 		shouldQuitChan)
@@ -91,7 +97,7 @@ func TestGetCurrentPasswordFail(t *testing.T) {
 
 	EnableIAM("server", "8080", "us-east-1", "dbuser", "***",
 		credentials.NewStaticCredentials("id", "pass", "token"),
-		rdsu,
+		&rdsu,
 		tmr,
 		logger,
 		shouldQuitChan)
@@ -147,7 +153,7 @@ func TestEnableIAMNormal(t *testing.T) {
 	// Start cycling through the list of passwords.
 	EnableIAM("server", "8080", "us-east-1", "dbuser", "***",
 		credentials.NewStaticCredentials("id", "pass", "token"),
-		rdsu,
+		&rdsu,
 		tmr,
 		logger,
 		shouldQuitChan)
--- a/pkg/middleware/recovery.go
+++ b/pkg/middleware/recovery.go
@@ -32,7 +32,7 @@ func Recovery(logger Logger) func(inner http.Handler) http.Handler {
 						fields = append(fields, zap.Error(err))
 					} else {
 						fields = append(fields, zap.Any("object", obj))
-						fields = append(fields, zap.String("stacktrace", fmt.Sprintf("%s", debug.Stack())))
+						fields = append(fields, zap.String("stacktrace", string(debug.Stack())))
 					}
 					logger.Error("http request panic", fields...)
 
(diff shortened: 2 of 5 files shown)


pkg/iampostgres/iampostgres_test.go
```diff
@@ -14,9 +15,14 @@ type RDSUTest struct {
...
-func (r RDSUTest) GetToken(endpoint string, region string, user string, iamcreds *credentials.Credentials) (string, error) {
+func (r *RDSUTest) GetToken(endpoint string, region string, user string, iamcreds *credentials.Credentials) (string, error) {
+	if len(r.passes) == 0 {
+		return "", errors.New("no passwords to rotate")
+	}
+
+	// Rotate the slice: first item goes to back of slice
 	pass := r.passes[0]
-	r.passes = append(r.passes[:0], r.passes[1:]...)
+	r.passes = append(r.passes[1:], pass)
```
reggieriser, line 25: Linter issue: `SA4005: ineffective assignment to field RDSUTest.passes (staticcheck)`

The problem was that this method was being used with a value receiver and thus the `r.passes` assignment would not make a permanent change to the struct when the method finished (since it was operating on a copy) -- i.e., it was ineffective.  So, I changed it to a pointer receiver which makes it effective again and makes the linter happy.  However, some tests started failing after that with out-of-bounds errors because it was operating in a different (wrong?) way by removing the first element from the slice each time it was called.  After pairing with @YanZ777 , we felt the intention was to rotate the passwords in the slice as opposed to removing the first one.  The code has been changed to do so and the tests pass fine now.

pkg/middleware/recovery.go
```diff
@@ -32,7 +32,7 @@ func Recovery(logger Logger) func(inner http.Handler) http.Handler {
 						fields = append(fields, zap.Error(err))
 					} else {
 						fields = append(fields, zap.Any("object", obj))
-						fields = append(fields, zap.String("stacktrace", fmt.Sprintf("%s", debug.Stack())))
+						fields = append(fields, zap.String("stacktrace", string(debug.Stack())))
```
reggieriser, line 35: Linter issue: `S1025: the argument's underlying type is a slice of bytes, should use a simple conversion instead of fmt.Sprintf (gosimple)`

This was a pretty straightforward simplification.

pkg/paperwork/generator.go
```diff
@@ -50,13 +50,7 @@ type pdfCPUWrapper struct {
...
 func (pcw pdfCPUWrapper) Merge(files []io.ReadSeeker, w io.Writer) error {
 	var rscs []io.ReadSeeker
-	for _, f := range files {
-		frsc, ok := f.(io.ReadSeeker)
-		if !ok {
-			return errors.Errorf("file %T does not implement io.ReadSeeker", f)
-		}
-		rscs = append(rscs, frsc)
-	}
+	rscs = append(rscs, files...)
```
reggieriser, line 53: Linter issue: `S1040: type assertion to the same type: f already has type io.ReadSeeker (gosimple)`

Since the function parameter requires a slice of `io.ReadSeeker` already, there was no reason to do a type assertion here.  Once I removed that, the loop was no longer needed and a simple `append` could accomplish the same thing.

pkg/server/server_test.go
```diff
@@ -247,6 +247,16 @@ func (suite *serverSuite) testTLSConfigWithRequest(tlsVersion uint16) {
...
+
+	//RA Summary: gosec - G402 - TLS MinVersion too low
+	//RA: The linter flagged this line of code because we are passing in a tlsVersion which could be deemed too low.
+	//RA: The code is part of a test function that tests TLS configuration with multiple TLS levels (currently 1.2 and 1.3).
+	//RA: It is executed as part of our test suite and is not included in the production system.
+	//RA Developer Status: Mitigated
+	//RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+	//RA Validator:
+	//RA Modified Severity:
+	// #nosec G402
```
reggieriser, line 259: Linter issue: `G402: TLS MinVersion too low. (gosec)`

This appears to be a false positive.  We have parameterized the TLS version in this test function so we could test multiple versions (currently only 1.2 and 1.3, both of which should be OK).  That appears to confuse the linter.  This is also only used in tests and not production code.

pkg/server/server_test.go
```diff
@@ -340,7 +350,9 @@ func (suite *serverSuite) TestTLSConfigWithRequestNoClientAuth() {
 
 	// Check the TLS connection directly
 	// This should fail and conn should be nil
-	config := tls.Config{}
+	config := tls.Config{
+		MinVersion: tls.VersionTLS13,
+	}
```
reggieriser, line 355: Linter issue: `G402: TLS MinVersion too low. (gosec)`

I added a `MinVersion` here rather than taking the default (which per the docs, gets version 1.0 which is too low).  This test isn't concerned with the TLS version, so adding a valid one should be fine.

pkg/server/server_test.go
```diff
@@ -401,6 +413,7 @@ func (suite *serverSuite) TestTLSConfigWithInvalidAuth() {
 	config := tls.Config{
 		RootCAs:      invalidCaCertPool,
 		Certificates: invalidCertificates,
+		MinVersion:   tls.VersionTLS13,
```
reggieriser, line 416: Linter issue: `G402: TLS MinVersion too low. (gosec)`

I added a `MinVersion` here rather than taking the default (which per the docs, gets version 1.0 which is too low).  This test isn't concerned with the TLS version, so adding a valid one should be fine.

travelar (approved):
Appreciate the comments.

pkg/iampostgres/iampostgres_test.go
reggieriser: @YanZ777 @carterjones ... I particularly wanted to call out this change to you two as I believe you have the most context on this code.

pkg/server/server_test.go
reggieriser: @Leofscott Here's the annotation/suppresion that needs your review/approval.

pkg/iampostgres/iampostgres_test.go
carterjones: That seems like a reasonable assumption. Given that it's a test case and that 1) the assignment wasn't functioning previously and 2) the tests continue to pass, this approach makes sense to me.

Leofscott (approved):
LGTM

gidjin (approved):
Changes look good to me, feel free to wait for a non A-team eng review too if you want.

YanZ777 (approved):
Looks good to me!

PR #6993, opened 2021-07-14
MB-8671: Intercept XHR responses and dispatch actions

kylehilltruss:
## Description

One of our July slice goals is to display a universal error message to Customer users whenever an action that they take causes an XHR to be responded to with a 500 status code. This pull request implements a `requestInterceptor` function which:

* attaches to the `internal` SwaggerClient, 
* listens to every XHR response that comes back, 
* and dispatches an action to the redux store declaring whether or not that response contained a 500 error code.

The first two parts of that checklist are relatively straightforward, but the third presented a unique challenge, as redux is not designed for arbitrary event dispatch and "side effects." There are several popular options for managing redux side effects ([thunks](https://github.com/reduxjs/redux-thunk) and [sagas](https://redux-saga.js.org/)), which we use elsewhere in our application, but neither is really appropriate for our use case here.

The solution I came up with was to create a middleware function which, upon application startup, receives a reference to the active redux store and maintains it in a closure; another function is exported which allows our `responseInterceptor` to make dispatches using that closed over reference.

## Reviewer Notes

There should be no perceptible change to the application's display of information or performance. (It actually might be a little faster, because we're instantiating one fewer copy of SwaggerClient now.)

Also this is... uh, a bit of a hack; I am very interested if reviewers know of better strategies here!

## Setup

Run the application normally. There should be no perceptible change to the content or display, as this ticket is just setting up scaffolding for future work in this slice.

```sh
make server_run
make client_run
```

## References

* Closes [MB-8671](https://dp3.atlassian.net/browse/MB-8671).

### DIFF
--- a/src/services/internalApi.js
+++ b/src/services/internalApi.js
@@ -1,6 +1,6 @@
 import Swagger from 'swagger-client';
 
-import { makeSwaggerRequest, requestInterceptor } from './swaggerRequest';
+import { makeSwaggerRequest, requestInterceptor, responseInterceptor } from './swaggerRequest';
 
 let internalClient = null;
 
@@ -10,6 +10,7 @@ export async function getInternalClient() {
     internalClient = await Swagger({
       url: '/internal/swagger.yaml',
       requestInterceptor,
+      responseInterceptor,
     });
   }
 
--- a/src/services/swaggerRequest.js
+++ b/src/services/swaggerRequest.js
@@ -4,6 +4,8 @@ import { normalize } from 'normalizr';
 import * as Cookies from 'js-cookie';
 
 import * as schema from 'shared/Entities/schema';
+import { interceptInjection } from 'store/interceptor/injectionMiddleware';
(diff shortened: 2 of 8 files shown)


YanZ777:
Asking for clarification: What should I expect to happen if I get the server to throw a 500?

kylehilltruss:
> Asking for clarification: What should I expect to happen if I get the server to throw a 500?

Speaking practically, the user should expect to see an error message on whatever page they're on (copy TBD, but something like "there was a server error, try again or call this phone number"), but the work to display that is in a separate ticket.

As a developer for this pull request, you should expect there to be an action fired through the active redux store equal to `{ type: "INTERCEPT_RESPONSE", hasError: true }`.

YanZ777 (approved):
Approving because testing it out, looks like it is intercepting all the service calls that I would expect it to intercept.

jenniferlynparsons (approved):
Pair reviewed these changes with @kylehilltruss and this all looks good!

PR #6995, opened 2021-07-14
MB-8665 Move order validations to order_updater

monfresh:
When we first added validations to prevent a TOO from submitting a move
while certain order fields were missing, we made the mistake of putting
the validations in the Order model as opposed to the service. Back then,
we only had one generic service for updating orders, so it might have
made sense at the time.

Since then, we've added more complexity, such as requiring separate
validations for counselors versus TOOs, and allowing customers to
upload amended orders even if the move has made it past service
counseling but hasn't been updated by the TOO yet.

Instead of trying to fit all of these different scenarios into a
generic validation in the Order model, it now makes more sense to
only add the validation where it applies, which is only when a TOO is
updating an order and when they are approving a move.

## Description

## Reviewer Notes

## Setup

```sh
echo "Code goes here"
```

## References

## Screenshots

### DIFF
--- a/pkg/models/move.go
+++ b/pkg/models/move.go
@@ -564,13 +564,7 @@ func SaveMoveDependencies(db *pop.Connection, move *Move) (*validate.Errors, err
 			}
 		}
 
-		order := &move.Orders
-		err := db.Load(order, "Moves")
-		if err != nil {
-			responseError = errors.Wrap(err, "Error Loading Order")
-			return transactionError
-		}
-		if verrs, err := db.ValidateAndSave(order); verrs.HasAny() || err != nil {
+		if verrs, err := db.ValidateAndSave(&move.Orders); verrs.HasAny() || err != nil {
 			responseVErrors.Append(verrs)
 			responseError = errors.Wrap(err, "Error Saving Orders")
 			return transactionError
--- a/pkg/services/move_task_order/move_task_order_updater.go
+++ b/pkg/services/move_task_order/move_task_order_updater.go
@@ -14,6 +14,7 @@ import (
 	movetaskorderops "github.com/transcom/mymove/pkg/gen/primeapi/primeoperations/move_task_order"
 	"github.com/transcom/mymove/pkg/models"
 	"github.com/transcom/mymove/pkg/services"
+	"github.com/transcom/mymove/pkg/services/order"
 	"github.com/transcom/mymove/pkg/services/query"
 	"github.com/transcom/mymove/pkg/unit"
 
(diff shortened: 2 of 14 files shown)


pkg/models/move.go
```diff
@@ -564,13 +564,7 @@ func SaveMoveDependencies(db *pop.Connection, move *Move) (*validate.Errors, err
 			}
 		}
 
-		order := &move.Orders
-		err := db.Load(order, "Moves")
```
monfresh, line 568: Previously, we needed to load the Moves because the Order validation was checking the move status, but now that we moved the validation to the service level, we don't even need to check the move status anymore because the Move already has to be in the right status by the time the validation is called.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -117,75 +110,83 @@ func (o moveTaskOrderUpdater) MakeAvailableToPrime(moveTaskOrderID uuid.UUID, eT
...
+		if transactionError != nil {
+			return &models.Move{}, transactionError
 		}
 	}
 
 	return move, nil
 }
 
+func (o *moveTaskOrderUpdater) updateMove(tx *pop.Connection, move models.Move, checks ...order.Validator) error {
+	if verr := order.ValidateOrder(&move.Orders, checks...); verr != nil {
```
monfresh, line 146: This is where the order validation happens.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -117,75 +110,83 @@ func (o moveTaskOrderUpdater) MakeAvailableToPrime(moveTaskOrderID uuid.UUID, eT
...
+			if includeServiceCodeCS {
+				err = o.createServiceItem(tx, *move, models.ReServiceCodeCS)
 			}
-			return &models.Move{}, err
-		}
-		if verrs != nil {
-			return &models.Move{}, verrs
-		}
 
-		// CreateMTOServiceItem may have updated the move status so refetch as to not return incorrect status
```
monfresh, line 178: Since we are calling `CreateMTOServiceItem` with a `nil` `MTOShipmentID`, it will return way before the part where it modifies the Move status, so this comment is incorrect. In this scenario, `CreateMTOServiceItem` will never change the move status, and it shouldn't. We expect the move to always be `Approved` when we first make it available to the Prime.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -117,75 +110,83 @@ func (o moveTaskOrderUpdater) MakeAvailableToPrime(moveTaskOrderID uuid.UUID, eT
...
+	verrs, err := tx.ValidateAndUpdate(&move)
+
+	if verrs != nil && verrs.HasAny() {
+		return services.NewInvalidInputError(move.ID, nil, verrs, "")
+	}
+
+	return err
+}
+
+func (o *moveTaskOrderUpdater) createServiceItem(tx *pop.Connection, move models.Move, code models.ReServiceCode) error {
```
Ryan-Koch, line 159: Since this private function is only used here to create move level service items (those with a nil `MTOShipmentID`) would it make sense to rename this function to something that makes that distinction clear? When I was reading through this I initially went 'oh wait do we use this function to create other types service items?' after I saw the hard-coded `MTOShipmentID: nil` key value pair below. It might help others avoid a similar brain stumble later.

monfresh: Good idea.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -84,7 +84,10 @@ func (o moveTaskOrderUpdater) UpdateStatusServiceCounselingCompleted(moveTaskOrd
 	return move, nil
 }
 
-//MakeAvailableToPrime updates the status of a MoveTaskOrder for a given UUID to make it available to prime
+// MakeAvailableToPrime approves a Move, makes it available to prime, and
```
Ryan-Koch, line 87: I really like this comment addition ✨

Ryan-Koch (approved):

PR #6994, opened 2021-07-14
MB-8254 Pricing details domestic origin shuttle service

donaldthai:
## Description

This PR creates a UI for the pricing details of the `Domestic Origin Shuttle Service (DOSHUT)` service item. The pricing details are displayed during the payment service item review and the payment requests page after it's been fully reviewed.

## Reviewer Notes

Is there anything you would like reviewers to give additional scrutiny?
- The worked is based off the new design for shuttle service: https://app.zeplin.io/project/60b8e558e0e96f9861d9d89d/screen/60d3ab32b21fe7baf90b4ff4
- Some param data are not available yet such as:
  - PriceRateOrFactor
  - EscalationCompounded
- Pricing data is also being worked on by A-Team: https://dp3.atlassian.net/browse/MB-2610

## Setup

Add any steps or code to run in this section to help others prepare to run your code:

```sh
make db_dev_e2e_populate
make server_run
make office_client_run
```

### Test
1. Login as TIO role office app
2. Search for move with move code `PARAMS` and click to go to move details
3. Click on `Payment requests` header link
4. Click on `Review service items` button
5. Click `Next` button until the `Domestic Origin Shuttle Service` service item shows up
6. Click on `Show calculations` link

### Storybook
```
make storybook
```
1. Visit http://localhost:6006/?path=/story/office-components-serviceitemcalculations--doshut

Notes: You may have to approve all service items if you want to see the calculations on the payment requests page

## Code Review Verification Steps

* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8254) for this change
* [Shuttle service mockup](https://app.zeplin.io/project/60b8e558e0e96f9861d9d89d/screen/60d3ab32b21fe7baf90b4ff4) 

## Screenshots

Payment requests page
image

Payment service item review page
image

PR #6996, opened 2021-07-14
[MB-2610] Create destination and origin shuttling pricer

pearl-truss:
## Description

This PR adds pricers for domestic shuttling. It also updates the lookups for billed, actual and estimated weights to use the weights tied to the mtoService item for shuttling service items instead of the weights tied to the mtoShipment

## Reviewer Notes

There's a few things to look closely at:
- I updated the[ Acceptance Testing docs](https://github.com/transcom/mymove/wiki/Acceptance-Testing-Payment-Requests#domestic-destination-shuttling-ddshut) with instructions for pricing DDSHUT and DOSHUT. I  want to make sure the formula and query looks correct. Especially since accessorial prices are PerUnitCent instead of PriceCents 
- Any thoughts on if the approach to update the lookups for billed, actual and estimated weights to use the weights tied to the mtoService item for shuttling service items is sufficient 

## Setup

```sh
db_dev_e2e_populate
make server_run
```

create a payload.json file (you can use these exact IDs if you ran e2e populate
```json
{
    "body": {
      "isFinal": false,
      "moveTaskOrderID": "d4d95b22-2d9d-428b-9a11-284455aa87ba",
      "serviceItems": [
        {
          "id": "801c8cdb-1573-40cc-be5f-d0a24034894b"
        },
        {
            "id": "2b0ce635-d71b-4000-a22a-7c098a3b6ae9"
        }
      ]
    }
  }
```

`prime-api-client --insecure create-payment-request --filename payload.json | jq`

You can also see the TIO calculations for origin shuttling ([since TTV has completed this work](https://github.com/transcom/mymove/pull/6994))

1. Login as TIO role office app
2. Search for move with move code PARAMS and click to go to move details
3. Click on Payment requests header link
4. Click on Review service items button
5. Click Next button until the Domestic Origin Shuttle Service service item shows up
6. Click on Show calculations link

You can use this query to verify the price and escalation values
```sql
select per_unit_cents, escalation_compounded from re_domestic_accessorial_prices dap
   inner join re_services on dap.service_id = re_services.id
   inner join re_contracts on re_contracts.id = dap.contract_id
   inner join re_contract_years on re_contracts.id = re_contract_years.contract_id
   where re_services.code = 'DDSHUT'
    and re_contracts.code = 'TRUSS_TEST'
    and dap.services_schedule = 3
    and '2020-03-15' between re_contract_years.start_date and re_contract_years.end_date;
```

## Acceptance
Create a DDSHUT and/or DOSHUT service item 
```json
{
  "body": {
    "moveTaskOrderID": "<insert MTO ID>",
    "mtoShipmentID": "<insert shipment ID>",
    "modelType": "MTOServiceItemShuttle",
    "reServiceCode": "DOSHUT",
    "estimatedWeight": 1400,
    "actualWeight": 1400,
    "reason": "Explanation of why a shuttle service is required",
    "description": "Description of shuttling"
  }
}
```

```
prime-api-client --cac --hostname api.stg.move.mil --port 443 create-mto-service-item --filename create_doshut.json | jq
```

You can then use the service item ID to create a payment request 
```json
{
    "body": {
      "isFinal": false,
      "moveTaskOrderID": "<MTO_ID>",
      "serviceItems": [
        {
          "id": "<SerivceItemID>"
        }
      ]
    }
  }
```

```
prime-api-client --cac --hostname api.stg.move.mil --port 443 create-payment-request --filename create_payment_request.json | jq
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2610) for this change

## Screenshots
Screen Shot 2021-07-16 at 12 35 58 PM


pkg/services/ghcrateengine/pricer_helpers.go
```diff
@@ -204,6 +205,56 @@ func priceDomesticPickupDeliverySIT(db *pop.Connection, pickupDeliverySITCode mo
...
+	if err != nil {
+		return 0, nil, fmt.Errorf("Could not lookup Domestic Accessorial Area Price: %w", err)
+	}
+
+	contractYear, err := fetchContractYear(db, domAccessorialPrice.ContractID, requestedPickupDate)
+	if err != nil {
+		return 0, nil, fmt.Errorf("Could not lookup contract year: %w", err)
+	}
+
+	basePrice := domAccessorialPrice.PerUnitCents.Float64() * weight.ToCWTFloat64()
```
gidjin, line 237: Looking at the pricing spreadsheet the price needs to be multiplied by the average cwt weight of all the shipments times the number of shipments. I'm not sure how we should be handling it with our current setup.

> For each service schedule - multiply the price proposed by the number of shipments and the average hundred weight of each shipment and sum across schedules

gidjin (commented):
This looks good so far, I think how you have improved the lookups for weights is good. I left a small nit that applies to all 3. As for the actual calculation we may need to do some asking since the pricing spreadsheet is Price * num shipments * avg cwt weight of those shipments. I'm not sure if that's the average overall of all shuttles across a time period or all shipments across one move. I kind of suspect the former which will be more difficult to price.

pkg/payment_request/service_param_value_lookups/weight_billed_actual_lookup.go
```diff
@@ -13,18 +14,40 @@ type WeightBilledActualLookup struct {
...
 
-	// Make sure there's an actual weight since that's nullable
-	actualWeight := r.MTOShipment.PrimeActualWeight
-	if actualWeight == nil {
-		// TODO: Do we need a different error -- is this a "normal" scenario?
-		return "", fmt.Errorf("could not find actual weight for MTOShipmentID [%s]", r.MTOShipment.ID)
+		if estimatedWeight == nil {
+			// TODO: Do we need a different error -- is this a "normal" scenario?
+			return "", fmt.Errorf("could not find estimated weight for MTOServiceItemID [%s]", keyData.MTOServiceItem.ID)
+		}
```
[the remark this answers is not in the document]
pearl-truss: I put them in the switch statements so that error message could return either the MTOShipmentID or the MTOServiceItemID depending on the case. I agree that is repeating code, is there a straightforward way of keeping that specificity in the error message if it's out of the switch? I know in JS you can do ternary expressions for inline conditionals but I don't think Go has anything like that

pkg/services/ghcrateengine/pricer_query_helpers.go
```diff
@@ -64,6 +64,23 @@ func fetchDomServiceAreaPrice(db *pop.Connection, contractCode string, serviceCo
 	return domServiceAreaPrice, nil
 }
 
+func fetchAccessorialPrice(db *pop.Connection, contractCode string, serviceCode models.ReServiceCode, schedule int) (models.ReDomesticAccessorialPrice, error) {
```
reggieriser, line 67: I'm not seeing a test for this function, but it looks like we have tests for the other functions in this file (granted, it's a private function, and I know sometimes we don't have separate tests for private functions).

pkg/services/ghcrateengine/service_item_pricer_test.go
```diff
@@ -63,7 +63,9 @@ func (suite *GHCRateEngineServiceSuite) TestGetPricer() {
 		{models.ReServiceCodeDLH, &domesticLinehaulPricer{}},
 		{models.ReServiceCodeDSH, &domesticShorthaulPricer{}},
 		{models.ReServiceCodeDOP, &domesticOriginPricer{}},
-		{models.ReServiceCodeDDP, &domesticDestinationPricer{}},
```
pearl-truss, line 66: Whoops, I didn't mean to remove DDP here, will address

pkg/services/ghcrateengine/domestic_destination_shuttling_pricer_test.go
```diff
@@ -0,0 +1,113 @@
...
+const (
+	testServiceSchedule      = 2
+	testBasePriceCents       = unit.Cents(353)
+	testContractYearName     = "Test Year 1"
+	testEscalationCompounded = 1.125
+	testWeight               = unit.Pound(4000)
+	testPriceCents           = unit.Cents(15885) // testBasePriceCents * (testWeight / 100) * testEscalationCompounded
+)
+
+var testRequestedPickupDate = time.Date(testdatagen.TestYear, time.June, 5, 7, 33, 11, 456, time.UTC)
```
reggieriser, line 23: I'd like to see these identifiers namespaced with the service item code being tested, like `ddshutTestServiceSchedule`.  That's what we do for the other similar tests.  The reason I believe we have that convention is to avoid name conflicts on some of these common names since all these tests are within the same package.  We could have a single `testServiceSchedule` shared across all of them, but that also makes one small change potentially ripple to all the pricing in all the tests.

gidjin:
> This looks good so far, I think how you have improved the lookups for weights is good. I left a small nit that applies to all 3. As for the actual calculation we may need to do some asking since the pricing spreadsheet is Price * num shipments * avg cwt weight of those shipments. I'm not sure if that's the average overall of all shuttles across a time period or all shipments across one move. I kind of suspect the former which will be more difficult to price.

Looks like the pricing is correct per USTC. Just wanted to reflect that here as well.

reggieriser (approved):
LGTM!  Thanks for making the variable name changes and adding the additional test.  🚀

PR #6997, opened 2021-07-14
[MB-8756] Update remaining pre-commit hooks

reggieriser:
## Description

In #6987, I updated the pre-commit hook for `golangci-lint` to the latest version.  Since it required some fixes and annotations, I kept it separate from other pre-commit updates to make that PR more focused.

In this PR, I am updating the remaining versioned hooks.  This includes:

- `pre-commit-hooks` updated from v3.4.0 to v4.0.1
- `markdownlint-ci` updated from v0.26.0 to v0.27.1.

In addition, while running all files through all pre-commit hooks, I found that a few JavaScript files were being flagged by prettier.  Those have been fixed in this PR.  There was also an eslint issue around a dependency in our custom `eslint-plugin-ato` linter.  After talking with @Ronolibert , we decided it was safe to turn that check off because the dependency is in the parent node modules, but it's not being recognized properly by the linter.

Similarly, there were also some issues flagged by `shell-lint` that required some slight adjustments.  Those are also fixed here. 

## Reviewer Notes

I've added additional context inline where needed.

## Setup

1. `make clean server_build client_build build_tools`
1. `pre-commit run -a`

Verify that all pre-commit checks run successfully.

Also, you can run `pre-commit autoupdate` and verify that there are no additional hooks that need updating.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8756) for this change

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,12 +12,11 @@ repos:
         types: [go]
 
   - repo: git://github.com/pre-commit/pre-commit-hooks
-    rev: v3.4.0
+    rev: v4.0.1
     hooks:
       - id: check-json
       - id: check-merge-conflict
       - id: check-yaml
-        exclude: config/database.yml # database.yml is not a valid yaml file, it contains go templating
       - id: detect-private-key
         #RA Summary: detect-private-key - Private key found
         #RA: detect-private-key detected a private key in source control
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -61,7 +61,7 @@ repos:
         entry: bash -c 'exec golangci-lint run -v ${GOLANGCI_LINT_VERBOSE} -j=${GOLANGCI_LINT_CONCURRENCY:-1}' # custom bash so we can override concurrency for faster dev runs
 
   - repo: https://github.com/igorshubovych/markdownlint-cli
-    rev: v0.26.0
+    rev: v0.27.1
     hooks:
       - id: markdownlint
         entry: markdownlint --ignore .github/*.md
--- a/eslint-plugin-ato/.eslintrc
+++ b/eslint-plugin-ato/.eslintrc
@@ -1,0 +1,5 @@
+{
+  "rules": {
+    "import/no-extraneous-dependencies": "off"
+  }
+}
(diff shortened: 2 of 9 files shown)


eslint-plugin-ato/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "import/no-extraneous-dependencies": "off"
+  }
+}
```
reggieriser, line 5: Linter issue: `'eslint' should be listed in the project's dependencies. Run 'npm i -S eslint' to add it` ([for code like this](https://github.com/transcom/mymove/blob/ac6adc0d6854c5d5af7a7b2b4e51c13d2247f43c/eslint-plugin-ato/tests/no-unapproved-annotation.test.js#L1))

As mentioned in the PR description, we have `eslint` in the root of the project, but it's just not getting identified by eslint.  It appears we turn this check off in other parts of the code as well (like with Storybook).

scripts/generate-secure-migration
```diff
@@ -6,7 +6,8 @@
 
 set -eu -o pipefail
 
-readonly dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+readonly dir
```
reggieriser, line 10: Linter issue: `SC2155: Declare and assign separately to avoid masking return values`

Here is [more detail about SC2155](https://www.shellcheck.net/wiki/SC2155).  In a nutshell, you just need to split up the declaration from the assignment in some cases as noted in the link.  There were 6 total cases of this that I fixed.

.pre-commit-config.yaml
```diff
@@ -12,12 +12,11 @@ repos:
         types: [go]
 
   - repo: git://github.com/pre-commit/pre-commit-hooks
-    rev: v3.4.0
+    rev: v4.0.1
     hooks:
       - id: check-json
       - id: check-merge-conflict
       - id: check-yaml
-        exclude: config/database.yml # database.yml is not a valid yaml file, it contains go templating
```
reggieriser, line 20: This file is no longer in the repo, so I removed the exclusion.

travelar (approved):

jacquelineIO (approved):
lgtm ✅

PR #6998, opened 2021-07-14
[MB-1573] Implement lookup for CubicFeetCrating

mboren:
## Description

The lookup for `CubicFeetCrating` finds the length, width, and height of the crate for a given service item, and multiplies them to get the volume.

I also added a migration to change the types of both the `CubicFeetCrating` and `CubicFeetBilled` service item param keys from `INTEGER` to `DECIMAL`. The designs show 2 decimal places for these, and since we're storing as a string anyway there's not much reason to use an integer to store fractions of a cubic foot.

## Reviewer Notes

- I had to preload `Dimensions` when setting up lookups to make sure that I could access the field in the new lookup. If there's a better way to do this, let me know.
- I've been going back and forth about moving the thousandth of an inch to feet conversion to `unit.ThousandthInch`. I haven't so far because there are a few potential numerical pitfalls here and I wanted everything in one place so those would be easier to spot, but I'm open to refactoring that out.

## Setup
I don't think there's any code that uses this lookup yet, so just make sure migrations run:
```sh
make db_dev_reset
make db_dev_migrate
```
```sql
select key, type from service_item_param_keys where key in ('CubicFeetCrating', 'CubicFeetBilled');
```
you should see:

key                        | type
----------------- | -------------
CubicFeetBilled | DECIMAL
CubicFeetCrating | DECIMAL

and make sure tests pass:
```sh
make server_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-1573) for this change

### DIFF
--- a/pkg/payment_request/service_param_value_lookups/cubic_feet_crating_lookup.go
+++ b/pkg/payment_request/service_param_value_lookups/cubic_feet_crating_lookup.go
@@ -16,11 +16,11 @@ type CubicFeetCratingLookup struct {
 }
 
 func (c CubicFeetCratingLookup) lookup(keyData *ServiceItemParamKeyData) (string, error) {
-	// Each service item has an array of dimensions. We expect there to be at most one
-	// dimension for the crate size.
+	// Each service item has an array of dimensions. There is a DB constraint preventing
+	// more than one dimension of each type for a given service item, so we just have to
+	// look for the first crating dimension.
 	for _, dimension := range keyData.MTOServiceItem.Dimensions {
 		if dimension.Type == models.DimensionTypeCrate {
-			// is there a classier way to work with unit.ThousandthInches?
 			lengthFeet := dimension.Length / thousandthInchesPerFoot
 			heightFeet := dimension.Height / thousandthInchesPerFoot
 			widthFeet := dimension.Width / thousandthInchesPerFoot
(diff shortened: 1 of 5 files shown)


pkg/payment_request/service_param_value_lookups/cubic_feet_crating_lookup.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+// CubicFeetCratingLookup does lookup for CubicFeetCrating
+type CubicFeetCratingLookup struct {
+}
+
+func (c CubicFeetCratingLookup) lookup(keyData *ServiceItemParamKeyData) (string, error) {
+	// Each service item has an array of dimensions. There is a DB constraint preventing
+	// more than one dimension of each type for a given service item, so we just have to
+	// look for the first crating dimension.
```
jacquelineIO, line 21: Thank you for the comments!

pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go
```diff
@@ -44,7 +44,7 @@ func ServiceParamLookupInitialize(
 
 	// Get the MTOServiceItem
 	var mtoServiceItem models.MTOServiceItem
-	err := db.Eager("ReService").Find(&mtoServiceItem, mtoServiceItemID)
+	err := db.Eager("ReService", "Dimensions").Find(&mtoServiceItem, mtoServiceItemID)
```
jacquelineIO, line 47: This will run for every service item and we probably don't want that. You can move it down to the switch statement similar to `db.Eager("PickupAddress", "DestinationAddress")` based on the service code.

jacquelineIO (approved):
lgtm. one change to consider left a comment inline. but I'm good with approving in the meantime

pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go
mboren: Thanks, I suspected I was doing the wrong thing there. I moved it into the switch.

Ryan-Koch (approved):

PR #7001, opened 2021-07-15
[MB-8606] scratch code to test creating crating and shuttling service items to …

jacquelineIO:
…the EDI

## Description

Drafting how to [map out the shuttle service item.](https://docs.google.com/spreadsheets/d/1h_mqb-iStgfTcu1llWHoA6CO4J9UuBCZ/edit#gid=1072177065)

Drafting how to [map out the crating service item](https://docs.google.com/spreadsheets/d/1h_mqb-iStgfTcu1llWHoA6CO4J9UuBCZ/edit#gid=1640574675). 

Sent out the samples to USBANK to check the syntax/formatting. Waiting to hear some feedback from them. 

## Reviewer Notes

## Setup

The UT will fail for these changes. But you are able to see the output of the EDI, which was the intent. Run the below command and scroll up to the top of the output and find the EDI string.

```sh
go test ./pkg/services/invoice   -testify.m TestAllGenerateEdi -v
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8606) for this change

## Screenshots

n/a

### DIFF
--- a/pkg/edi/segment/l0.go
+++ b/pkg/edi/segment/l0.go
@@ -10,9 +10,12 @@ type L0 struct {
 	LadingLineItemNumber   int     `validate:"min=1,max=999"`
 	BilledRatedAsQuantity  float64 `validate:"required_with=BilledRatedAsQualifier"`
 	BilledRatedAsQualifier string  `validate:"required_with=BilledRatedAsQuantity,omitempty,len=2"`
-	Weight                 float64 `validate:"required_with=WeightQualifier WeightUnitCode"`
-	WeightQualifier        string  `validate:"required_with=Weight WeightUnitCode,omitempty,eq=B"`
-	WeightUnitCode         string  `validate:"required_with=Weight WeightQualifier,omitempty,eq=L"`
+	Weight                 float64 `validate:"required_with=WeightQualifier"`
+	WeightQualifier        string  `validate:"required_with=Weight,omitempty,eq=B"`
+	Volume                 float64 `validate:"required_with=VolumeUnitQualifier,omitempty"`
+	VolumeUnitQualifier    string  `validate:"required_with=Volume,omitempty,eq=E,len=1"`
+	LadingQuantity         int     `validate:"omitempty"`
+	WeightUnitCode         string  //`validate:"oneof=L CRT"`
 }
 
 // StringArray converts L0 to an array of strings
@@ -32,6 +35,13 @@ func (s *L0) StringArray() []string {
 		billedRatedAsQuantity = strconv.FormatFloat(s.BilledRatedAsQuantity, 'f', 3, 64)
 	}
 
+	var volume string
+	if s.Volume == 0 {
+		volume = ""
+	} else {
+		volume = strconv.FormatFloat(s.Volume, 'f', 3, 64)
+	}
+
 	return []string{
 		"L0",
 		strconv.Itoa(s.LadingLineItemNumber),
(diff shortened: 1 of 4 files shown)


pkg/edi/segment/l0.go
```diff
@@ -10,9 +10,12 @@ type L0 struct {
 	LadingLineItemNumber   int     `validate:"min=1,max=999"`
 	BilledRatedAsQuantity  float64 `validate:"required_with=BilledRatedAsQualifier"`
 	BilledRatedAsQualifier string  `validate:"required_with=BilledRatedAsQuantity,omitempty,len=2"`
-	Weight                 float64 `validate:"required_with=WeightQualifier WeightUnitCode"`
-	WeightQualifier        string  `validate:"required_with=Weight WeightUnitCode,omitempty,eq=B"`
-	WeightUnitCode         string  `validate:"required_with=Weight WeightQualifier,omitempty,eq=L"`
+	Weight                 float64 `validate:"required_with=WeightQualifier"`
```
gidjin, line 13: Does this also need an `omitempty`? Seems like it might based on the specs

pkg/edi/segment/l0.go
```diff
@@ -10,9 +10,12 @@ type L0 struct {
...
+	WeightQualifier        string  `validate:"required_with=Weight,omitempty,eq=B"`
+	Volume                 float64 `validate:"required_with=VolumeUnitQualifier,omitempty"`
+	VolumeUnitQualifier    string  `validate:"required_with=Volume,omitempty,eq=E,len=1"`
+	LadingQuantity         int     `validate:"omitempty"`
+	WeightUnitCode         string  //`validate:"oneof=L CRT"`
```
gidjin, line 18: I'm wondering why the validation for `WeightUnitCode` is commented out.

gidjin (approved):
As far as I can tell this looks like the right approach with what we know now.

pkg/edi/segment/l0.go
jacquelineIO: @gidjin it was failing, but these are the ones that I was thinking that we needed and I didn't bother to figure out why it was failing

pkg/edi/segment/l0.go
jacquelineIO: @gidjin that makes sense

jacquelineIO:
This is was proof of concept PR. Not looking to merge this into the code base. But can be looked at for an example of what is expected.

pkg/edi/segment/l0.go
```diff
@@ -40,12 +50,10 @@ func (s *L0) StringArray() []string {
...
 		// TODO: will need to fill in the blank fields for crating
+		volume,
+		s.VolumeUnitQualifier,
 		"",
-		"",
-		"",
-		"",
-		"",
-		s.WeightUnitCode,
+		s.WeightUnitCode, // Packaging Form Code
```
reggieriser, line 56: I'm wondering if this newer layout is a little off.  I found [this thread in Slack](https://ustcdp3.slack.com/archives/CP6F568DC/p1595967981167800) which references [this EDI documentation](https://www.ustranscom.mil/cmd/associated/dteb/files/transportationics/dt858c41.pdf).  Page 67 has the specs for the `L0` segment.  That seems to imply that `Packaging Form Code` (slot 9) and `Weight Unit Code` (slot 11) are two different things.  In this new layout, you've got `WeightUnitCode` at slot 9.  It used to be at slot 11 (which is what the doc implies).  Is that 858 doc the one we should be using as a reference?

PR #7002, opened 2021-07-15
Mb 8465 403 forbidden error fix

eamahanna:
## Description

DO NOT MERGEEEE

## Reviewer Notes

## Setup

```sh
echo "Code goes here"
```

## References

## Screenshots


pkg/auth/authentication/auth.go
```diff
@@ -349,14 +349,15 @@ func (h LogoutHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			fmt.Fprint(w, logoutURL)
 		} else {
 			// Can't log out of login.gov without a token, redirect and let them re-auth
+			auth.DeleteCSRFCookies(w)
```
eamahanna, line 352: I think this will delete the stored cookies. I'm not sure if I need to delete the session as well.

src/layout/LogoutOnInactivity.jsx
```diff
@@ -23,6 +23,7 @@ export class LogoutOnInactivity extends React.Component {
     this.idleTimer = null;
     this.onActive = this.onActive.bind(this);
     this.onIdle = this.onIdle.bind(this);
+    this.loggedOut = false;
```
eamahanna, line 26: I am sure there is a better way to do this. This was just a quick and dirty fix.

pkg/auth/authentication/auth.go
```diff
@@ -349,14 +349,24 @@ func (h LogoutHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			err := resetUserCurrentSessionID(session, h.db, h.logger)
+			if err != nil {
+				h.logger.Error("failed to reset user's current_x_session_id")
+			}
+			err = h.sessionManager(session).Destroy(r.Context())
+			if err != nil {
+				h.logger.Error("failed to destroy session")
+			}
+			auth.DeleteCSRFCookies(w)
+			h.logger.Info("no session id found")
```
monfresh, line 361: Can we move this to line 352 so that as soon as the else statement is hit, we log this scenario before we try to do anything else? Also, can we change the message to "session exists but has an empty IDToken"?

pkg/auth/authentication/auth.go
```diff
@@ -323,7 +323,6 @@ func NewLogoutHandler(ac Context, db *pop.Connection) LogoutHandler {
 
 func (h LogoutHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	session := auth.SessionFromRequestContext(r)
-
 	if session != nil {
 		redirectURL := h.landingURL(session)
 		if session.IDToken != "" {
```
monfresh, line 328: What I'm seeing locally, is that if you sign out manually, via the Sign Out link in the top right, the session is properly populated, and the ID Token exists. However, if you let the app sign you out automatically, the session is not being passed into the handler correctly, and the ID Token is always empty.

Can we add a logger statement here to see if the same thing happens in experimental?

```golang
fmt.Println("SESSION IS PRESENT AFTER CALLING LOGOUT and ITS IDTOKEN IS NOT EMPTY")
```

monfresh: I meant `h.logger.Info`, not `fmt.Println`

YanZ777:
I'm seeing some unusual behavior.
So the other day, on staging, I let my account 403. 
Login to exp. I'm 403ed.
Clear my cookies with the lock icon to the left of the url. 
Login, but I see the inactive banner with the countdown. 
Click on the page before the countdown stops. The banner just restarts the countdown. 
Let the countdown finish without clicking on the page - I see the logout calls fail with 500s
```
{title: "Internal Server Error", instance: "feaf4160-4fba-472c-8a10-7c2fc36ca105",…}
detail: "An unexpected server error has occurred."
instance: "feaf4160-4fba-472c-8a10-7c2fc36ca105"
title: "Internal Server Error"
```
Click on the page again - restarts the countdown.
Refresh the page - logged out.

pkg/auth/authentication/auth.go
monfresh: If this is also the case in experimental, that means that all of those continuous logout requests are not going to login.gov. They're just redirecting to the landing page. My theory is that they start to fail because we are making too many of them, and we have some kind of rate limit on the AWS side that is causing the 403.

eamahanna:
OKay, so I ended up fixing the repeated 403 calls, the inactivity banner, the logout from the inactivity banner and if there is no sessionTokenID case. Everything but the no session id token can be tested locally. Thank you @YanZ777 and @monfresh

pkg/auth/authentication/auth.go
eamahanna: see comment below

pkg/auth/authentication/auth.go
```diff
@@ -349,14 +350,29 @@ func (h LogoutHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			fmt.Fprint(w, logoutURL)
 		} else {
 			// Can't log out of login.gov without a token, redirect and let them re-auth
+			h.logger.Info("session exists but has an empty IDToken")
+
+			if session.UserID != uuid.Nil {
+				err := resetUserCurrentSessionID(session, h.db, h.logger)
```
eamahanna, line 356: This was causing the 500 errors we were seeing, so I put a check to only complete this step if there is a user id.

src/layout/LogoutOnInactivity.jsx
```diff
@@ -44,11 +46,13 @@ export class LogoutOnInactivity extends React.Component {
...
-    const { timeLeftInSeconds } = this.state;
-    const { history } = this.props;
+    const { timeLeftInSeconds, loggedIn } = this.state;
+    const { history, logOut } = this.props;
 
-    if (timeLeftInSeconds === 0) {
+    if (timeLeftInSeconds === 0 && loggedIn) {
       LogoutUser().then(() => {
+        this.setState({ loggedIn: false });
+        logOut();
```
eamahanna, line 55: This logout function being omitted was why the session wasn't getting properly passed to the logout handler. Now this log out works just as if you were hitting the Sign Out button.

src/layout/LogoutOnInactivity.jsx
```diff
@@ -44,11 +46,13 @@ export class LogoutOnInactivity extends React.Component {
...
   countdown = () => {
-    const { timeLeftInSeconds } = this.state;
-    const { history } = this.props;
+    const { timeLeftInSeconds, loggedIn } = this.state;
+    const { history, logOut } = this.props;
 
-    if (timeLeftInSeconds === 0) {
+    if (timeLeftInSeconds === 0 && loggedIn) {
       LogoutUser().then(() => {
+        this.setState({ loggedIn: false });
```
eamahanna, line 54: This removes prevents the 403 calls being repeatedly made after being logged out.

YanZ777 (approved):
Emily and I talked through this, everything makes sense and looks good!

PR #7003, opened 2021-07-15
MB-8163: add a note about configuring dangerjs

carterjones:
## Description

Add a note about how to configure the DangerJS API token.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8163) for this change


sheenamt (approved):
Thanks for adding this info!

PR #7007, opened 2021-07-16
MB-8755 front end error tracking adr

mlm55:
## Description

We should have a way to know if front end errors are happening. Many SAAS products
need to be FedRAMPed in order to conform with ATO standards. Here we've
investigated a few different SAAS products.

## Reviewer Notes

Does anyone have experience/thoughts on the different services?

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -685,6 +685,7 @@ sortBy
 SauceLabs
 Screener.io
 prepending
+ - docs/adr/0063-error-tracking.md
 FedRAMPed
 Rollbar
 Datadog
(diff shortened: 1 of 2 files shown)


lynzt (commented):
something to also consider as a +/- is any time/resources needed from infra to setup/maintain...

rogeruiz:
This ADR has been migrated over to https://github.com/transcom/mymove-docs/compare/rogeruiz/adrs/migrate-to-docusaurus...mb-8755-front-end-error-tracking-adr?expand=1

Please use the link above to create a PR with the changes from this branch. Do not modify the base branch as GitHub will track the work as it gets merged in an switch it to the default branch automatically.

PR #7020, opened 2021-07-20
MB-8725 Update regular setup and nix setup documentation

felipe-lee:
## Description

Trying to update the setup instructions so hopefully anyone could follow one set or the other and be set up quickly.

Some of the README updates were just me moving things around. E.g. I moved the project layout and makefile sections out of the setup section because they weren't actual setup instructions, just more info on the project. Some sections that I just moved I didn't really look at much so if something seems wrong or out of date, let me know.

There are some related changes also happening in the engineering playbook if you're up for reviewing that too, though those changes are strictly focused on `nix`: https://github.com/trussworks/Engineering-Playbook/pull/187

## Reviewer Notes

I added a few `Makefile` targets to closely mimic some of the setup being done before. I tested them locally but they could probably use another test by someone.

## Setup

Follow the instructions for either of the setups? Only if you want to actually do all the steps though. My main hope is that they are more or less clear. Or as clear as they can be in a huge readme file...

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8725) for this change


Makefile
```diff
@@ -66,8 +66,11 @@ help:  ## Print the help documentation
 # This target ensures that the pre-commit hook is installed and kept up to date
 # if pre-commit updates.
 .PHONY: ensure_pre_commit
-ensure_pre_commit: .git/hooks/pre-commit ## Ensure pre-commit is installed
+ensure_pre_commit: .git/hooks/pre-commit install_pre_commit ## Ensure pre-commit is installed
 .git/hooks/pre-commit: /usr/local/bin/pre-commit
```
ahobson, line 70: Note `/usr/local/bin/pre-commit` won't exist if nix is used. We have so many hard coded paths that assume homebrew

felipe-lee: Yeah, that's why I split off the part of this target that installed the hooks to the `install_pre_commit` target so that the nix instructions would only call that target, while the regular setup still calls this full ones. 

That's part of what I'm going to write up in my findings for this ticket. My story was about seeing what it would take to switch over, and based on what I've seen, some work is needed to address some of the hard-coded paths (I wonder if we could just do `type -p <command>` instead or something similar). It would also be nice to maybe clean up some of the makefile targets in general that are used for setup. Or maybe instead of a bunch of makefile targets, just have a script people can run that will set things up.

rogeruiz: We could leverage `which` to verify install path. _I don't mind migrating things over to lookup command paths_ as a part of my onboarding work if you wanna focus on this story. wdyt @felipe-lee?

felipe-lee: I'd be good with that. Though instead of `which`, using `type` or `command`. This post has a lot more details: https://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then

README.md
```diff
@@ -129,152 +166,315 @@ Use your work email when making commits to our repositories. The simplest path t
...
 
-### Setup: Project Layout
+NOTE: If you need any of the packages/tools for other things that you won't use `nix` for, you can possibly set things
+up so that they both work, but you'll just have to set up your `PATH` properly. And even then, there may be other steps
+necessary which aren't documented here.
 
-All of our code is intermingled in the top level directory of mymove. Here is an explanation of what some of these directories contain:
+1. If you had previously been using `direnv`, `awscli`, `aws-vault`, `chamber`, and `pre-commit` installed via
+   `homebrew`, uninstall them.
+    1. `brew uninstall direnv awscli aws-vault chamber pre-commit`
```
rogeruiz, line 233: wdyt of making these code blocks rather than inline code? Specifically for commands that users need to run. The idea behind this change is that a code block has a copy button that can be used to copy the command into the terminal rather than having the user need to copy it manually. This would require making this an unordered list or not having this be a list at all. Codeblocks can't be added as a single line and if it were multi-line, then Markdown would start the numbers over on line 235.

felipe-lee: Oh that's a great idea. I can go through and update the ones in the setup sections to be easy to copy.

rogeruiz: awesome. Also, is it actually necessary to run uninstall here? I understand that there will be redundant installation of things but, I thought Nix would be configured to only use it's dependencies due to the `direnv`? I'm still fairly new to Nix, but if we don't need to uninstall things, I'm not sure it's a good idea to recommend it. Since the Nix dependencies won't be available once you leave the directory and change Nix profiles. cc: @ahobson

felipe-lee: The main reason I put this in is so people don't potentially run into any `PATH` issues. But I put a note above about being able to keep them installed, so long as they're willing to make sure their path is ok.

> NOTE: If you need any of the packages/tools for other things that you won't use `nix` for, you can possibly set things
up so that they both work, but you'll just have to set up your `PATH` properly. And even then, there may be other steps
necessary which aren't documented here.

If y'all think it's best to leave this out though, we can.

rogeruiz: I like the note! I think it's enough to warn folks by putting that there, we could even include a ⚠️ emoji as well. I don't think we should explicitly tell folks to uninstall things that they may be managing through other means e.g. Homebrew.

If we're trying to have Nix adoption across the whole team, we should leverage the isolation that Nix creates with mounted volumes / `NIX_PROFILES` and make sure users understand that their `PATH` is going to be something that needs to be checked first during any troubleshooting process for `transcom/mymove`.

For example, I currently have Bash installed on my laptop through both macOS-default and Nix. Outside of `transcom/mymove` the newer Nix-installed Bash is used, but outside of this directory I expect to use Apple's Bash even if it is older than what we expect in this project. 

Lastly, if we tell folks to uninstall things, we could later run into issues with documentation updating as we add/remove more tools. I like the idea of just having users run `nix/update.sh` and not have to worry about what is being installed or uninstalled via a `nix-env -i` command. If folks want to dig into what is actually being installed, then they can read the `nix/default.nix` file (e.g. `grep -onE 'name = ".+"' -- nix/default.nix`). Another example is that this list of things to uninstall is incomplete. For instance, I would need to uninstall a number of things such as `jq` and `shellcheck` since I have those installed according to my own personal workflows.

felipe-lee: Yeah that's true. I can take out this installation step. Though I'll have to mess with the next section on `asdf` and `nodenv` to see how do side by side a bit before taking that step out about disabling them. Unless you've messed with it and it works out well. I feel like they should be ok for the same `PATH` reasons. Like as long as `nix` is the first in the path, it should be ok, but I'd like to verify for those two because the instructions to disable those pre-dated this pr.

felipe-lee: I tested it out and it looks like as long as your `PATH` is set up properly, even the version managers are ok to have. I updated the section to mark it as optional with the caveat of the note. I took out the step about removing individual packages since like you said the list is really longer than it was. I left the version manager instructions because they take more to disable or uninstall. Do you think I should just remove that step too and only leave the note about the `PATH` and existing installs?

README.md
```diff
@@ -98,280 +110,731 @@ To create an account at login.gov, use your regular `truss.works` email and foll
 To create an account in the sandbox, follow the same instructions, but [in the sandbox server](https://idp.int.identitysandbox.gov/sign_up/enter_email).  Do _not_ use your regular email address in the sandbox.
 **Tip**: You can use the plus sign `+` to create a new truss email address.  `name+some_string@truss.works` will be treated as a new address, but will be routed to `name@truss.works`.
 
+## Project Layout
+
+All of our code is intermingled in the top level directory of `mymove`. Here is an explanation of what some of these
+directories contain:
+
+* `.circleci`: Directory for CircleCI CI/CD configuration
```
YanZ777, line 118: 🎉

README.md
```diff
@@ -98,280 +110,731 @@ To create an account at login.gov, use your regular `truss.works` email and foll
...
+```shell
+git clone git@github.com:transcom/mymove.git
+```
+
+```shell
+cd mymove
+```
+
+You will then find the code at `~/Projects/mymove`. You can check the code out anywhere EXCEPT inside your `$GOPATH`.
+So this is customization that is up to you.
```
YanZ777, line 212: Minor, but this sentence is phrased a bit oddly. I would either take it out or rephrase it.

README.md
```diff
@@ -98,280 +110,731 @@ To create an account at login.gov, use your regular `truss.works` email and foll
...
+[EditorConfig](http://editorconfig.org/) allows us to manage editor configuration (like indent sizes) with a
+[file](https://github.com/transcom/ppp/blob/master/.editorconfig) in the repo. Install the appropriate plugin in your
+editor to take advantage of that if you wish.
+
+### Setup: Nix
+
+If you need help with this setup, you can ask for help in the
+[Truss slack #code-nix channel](https://trussworks.slack.com/archives/C01KTH6HP7D).
+
+1. [Initial Setup](#nix-initial-setup)
```
YanZ777, line 225: Maybe I'm missing something here, but are these all supposed to be 1.s?

YanZ777 (approved):
The changes look reasonable to me, but someone who actually uses nix might want to review it as well.

README.md
felipe-lee: Oh I didn't add this, I just moved it. But I am glad it exists. Might need some updates in a future iteration though

README.md
felipe-lee: Yeah true. I feel like it doesn't add anything so I'll take it out.

README.md
felipe-lee: Yeah, it's a way of not having to change numbers if you move things around. When github renders it, it gives each a proper number.

README.md
YanZ777: Still. 👍🎉

PR #7030, opened 2021-07-20
MB-8673: Adds logic to interceptor reducer to handle events

kylehilltruss:
## Description

This pull request combines the work done in previous tickets ([MB-8670](https://dp3.atlassian.net/browse/MB-8670) and [MB-8671](https://dp3.atlassian.net/browse/MB-8671)), wiring up the interceptor reducer to listen to `INTERCEPT_RESPONSE` events and update it's internal state accordingly.

Any user action that causes an XHR request to be sent can actually cause _several_ requests to be sent; we want the state to be updated to reflect than an error response has been returned if _any_ of those XHR responses have done so, regardless of which request it was. Since our interception method doesn't have any way logically connect responses to the user actions that caused them, the way we're going to operationalize what a "user action" is will be waiting for a gap of at least three seconds between response interceptions.

How it works:

* If an intercepted XHR response returns a 500 status code, set the value of `hasRecentError` to `true` in the returned state.
* If an intercepted XHR response does _not_ return a 500 status code, *and* the value of `timestamp` in the current state is more than 3000ms from right now, set the value of `hasRecentError` to false in the returned state.
* Always set the value of `timestamp` in the returned state to be the current JS timestamp (from `Date.now()`).

## Reviewer Notes

The actual work is pretty simple, but let's make sure that this code is readable, and that what we're doing around the 3000ms delay makes sense from a product perspective.

You can test that this is working in the running application by inspecting the active redux store state. `timestamp` (and `hasRecentError`) should update whenever an XHR response has been received.

## Setup

There's no visual changes to the application for this work, but other tickets will use the `hasRecentError` value from the store to determine whether or not to display an error message in the application.

```sh
make server_run
make client_run
```

## References

* Closes [MB-8673](https://dp3.atlassian.net/browse/MB-8673).


src/store/interceptor/reducer.test.js
```diff
@@ -1,7 +1,46 @@
...
+        hasRecentError: true,
+        timestamp: Date.now() - SHORT_RESPONSE_GAP_MS,
+      };
+
+      const newState = interceptorReducer(oldState, interceptResponse(false));
+      expect(newState.hasRecentError).toEqual(true);
+      expect(newState.timestamp - oldState.timestamp).toBeGreaterThanOrEqual(SHORT_RESPONSE_GAP_MS);
+    });
+
+    it('updates hasRecentError value if timestamp is stale', () => {
```
felipe-lee, line 35: I'm not sure if there'd be value in adding the case where it was already false and it's stale. So like this could be more of a "resets hasRecentError value if timestamp is stale" and checks both cases of `true` and `false` end up being `false`.

felipe-lee (commented):
I think I might need some help testing this. I see the fields in the store, but when I force an error from the backend server (400 or 500), the `hasRecentError` doesn't update to be `true`. I'm probably testing it incorrectly though

src/store/interceptor/reducer.js
```diff
@@ -1,10 +1,39 @@
...
+
+      if (action.hasError) {
+        return {
+          ...state,
+          hasRecentError: true,
+          timestamp,
+        };
+      }
+
+      if (timestamp > state.timestamp + ACTION_RESPONSE_INTERVAL_MS) {
```
Ronolibert, line 23: 👍

Ronolibert (approved):

src/store/interceptor/reducer.test.js
kylehilltruss: Added a test case for this.

felipe-lee (approved):
LGTM. I was testing the 500 error by killing the BE server but locally that seems to mess up the FE server too. After we talked I triggered it like you said by returning an internal server error explicitly and that worked as expected.

PR #7031, opened 2021-07-20
[MB-8383] Update go-swagger from 0.24.0 to 0.27.0

mboren:
## Description
so uh this is a big one. 

The goal was to get a fix for a Swagger bug that allowed passing read only parameters in POST/PATCH/PUT requests.
Now, we should return an error when this happens, instead of silently ignoring it.

Most of the code changes that I had to make to get this mess to compile and work right are completely unrelated to that though. Most of my edits come from a change to how swagger handles fields that are marked "required".
They are stored as pointers instead of values now.

This means we have to do a lot of pointer shenanigans and check for nils in a few places where we weren't before.

All these values are checked if they are nil by swagger before they hit our application code, so all of the new nil checks should be unnecessary. I felt like I should still check them though. As a compromise, I didn't write any tests to make sure that these checks are working though. Willing to add these tests if anybody thinks it's necessary/asks nicely.

We also have some changes due to swagger's variable name generation changing.
"air-force" used to be transformed to "AirForce", but now it is "AirDashForce".

## Reviewer Notes
scrolling pro-tip: the files are sorted by path in the "files changed" tab, and all the generated files are in the middle, so look at the first couple of files, and then once you hit something in `pkg/gen`, scroll *all the way* to the bottom (you will probably want to use a keyboard shortcut for this if you don't want to spend 10 years rolling your mouse wheel)  to skip over all the generated files, and then work your way up.

The AC in the original ticket suggests adding a test to the CreateMoveTaskOrder handler to make sure that you cannot pass in a reference ID. However, this doesn't test anything that happens in this update because swagger will catch readOnly values before they reach the handler. I'm not really sure where to add an automated test for this, suggestions are welcome.

## Setup
First, rebuild stuff and regenerate swagger files
```
make clean
make server_build
make server_generate
```
Make sure that your working directory is clean (`git status -uno` doesn't show any modified files). This tells us that the correct version of swagger is running and that I didn't miss any files.

Make a POST request that sets a `readOnly` parameter, and make sure that it fails.
```sh
make server_run
```

```json
{
  "body": {
    "contractorID": "5db13bb4-6d29-4bdb-bc81-262f4513ecf6",
    "order": {
      "customer": {
        "agency": "AIR_FORCE",
        "dodID": "5698008567",
        "email": "leo_spaceman_sm@example.com",
        "firstName": "Lisa",
        "lastName": "Lee",
        "phone": "555-555-5555",
        "rank": "E_1"
      },
      "departmentIndicator": "AIR_FORCE",
      "destinationDutyStationID": "2d5ada83-e09a-47f8-8de6-83ec51694a86",
      "entitlement": {
        "authorizedWeight": 8000,
        "dependentsAuthorized": true,
        "totalDependents": 5,
        "totalWeight": 5000
      },
      "issueDate": "2021-07-21",
      "orderNumber": "Order1234",
      "ordersType": "PERMANENT_CHANGE_OF_STATION",
      "ordersTypeDetail": "HHG_PERMITTED",
      "originDutyStationID": "d7285e29-85f3-4839-b233-2512e18317cd",
      "rank": "E_6",
      "reportByDate": "2021-07-20",
      "status": "DRAFT",
      "tac": "1234",
      "uploadedOrdersID": "00000000-0000-0000-0000-000000000000"
    },
    "ppmType": "FULL",
    "referenceId": "asdfasdf",
    "selectedMoveType": "PPM",
    "status": "DRAFT"
  }
}
```

```
prime-api-client --insecure support-create-move-task-order --filename createmtoreadonly.json
```
In this example, we're sending `referenceId` which is read only, so we should get a 422 response that looks something like this:
```
HTTP/1.1 422 Unprocessable Entity
{
    "title": "Validation Error",
    "instance": "<your instance id here>",
    "detail": "referenceId in body is readOnly"
}
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8383) for this change
* [the discussion thread on this issue](https://github.com/go-swagger/go-swagger/issues/2400)  has an explanation for the change to required fields


pkg/handlers/internalapi/orders.go
```diff
@@ -241,10 +246,14 @@ func (h UpdateOrdersHandler) Handle(params ordersop.UpdateOrdersParams) middlewa
 		return handlers.ResponseForError(logger, err)
 	}
 
+	if payload.OrdersType == nil {
+		return handlers.ResponseForError(logger, errors.New("Missing OrdersType")) // TODO error handling
```
reggieriser, line 250: Is that TODO still applicable?  It looks like you're handling the error already, or is that referring to something else?

reggieriser (approved):
Wow, this is a big PR!  I glanced through some representative generated files just to get a feel for what all had been changed.  I did look at all the milmove changes.  The changes make sense to me based on what I saw in the generated files.  Tests pass and I get the appropriate error back when I try to send something in on a read-only field like you show.  I had a few minor comments.  Great job!  🚀 

I'd love to have @Ronolibert or somebody else who was interested in this upgrade also look at the PR if possible before merging.  We'll also need to be careful that we merge master into this branch right before the final merge to avoid missing anything with generated files.  I think we'll also need to alert the team when this PR is merged to master so they can merge master into their branches to minimize any conflicts or generated code diffs that are caught only at build time.

reggieriser:
Oh, one other comment -- just a reminder to update the docker image hash in this PR after you've merged in the swagger upgrade PR in the `circleci-docker` repo.

pkg/handlers/internalapi/orders.go
mboren: not applicable anymore, thanks for catching this.

pkg/handlers/internalapi/backup_contacts.go
```diff
@@ -11,6 +13,7 @@ import (
 )
 
 func payloadForBackupContactModel(contact models.BackupContact) internalmessages.ServiceMemberBackupContactPayload {
+	permission := internalmessages.BackupContactPermission(contact.Permission)
```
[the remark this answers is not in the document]
mboren: Thanks for the feedback. I think I agree. It's a small thing, but I was feeling paranoid about doing it all one way and then finding out I needed to change it all.

PR #7038, opened 2021-07-21
MB-8838 Updating setup documentation from an onboarding perspective

rogeruiz:
## Description

As a new member for the Roci team, I'm going through the initial setup documentation and updating things as I see necessary. In order to not duplicate effort, I'm collaborating with @felipe-lee by merging this into #7020 instead of `master`. My patch series _will focus on the initial developer experience_ rather than any of the effort of refactoring or migrating folks over to using Nix.

The commit messages in this patch series will be granular and descriptive about what the changes are and why. This should help keep things tidy as I'm not entirely sure of the specific scope of all the changes.

## Reviewer Notes

If you have the time, go through setting up this repository in a new directory or new machine. But the majority of the scrutiny should go to reading through the documentation and making sure that it makes sense.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8838) for this change

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,14 @@ To create an account in the sandbox, follow the same instructions, but [in the
 sandbox server](https://idp.int.identitysandbox.gov/sign_up/enter_email).  Do
 _not_ use your regular email address in the sandbox.
 
+### Creating alternative users with the same email address
+
+You can use the plus sign `+` to create a new Truss email address.
+`name+some_string@truss.works` will be treated as a new address, but will be
+routed to your `name@truss.works` email automatically. Don't use this for the
+office-side of account creation. It's helpful to use these types of accounts for
+the customer-side accounts.
+
 ## Project Layout
 
 All of our code is intermingled in the top level directory of `mymove`. Here is an explanation of what some of these
--- a/README.md
+++ b/README.md
@@ -104,10 +104,14 @@ As of 3/6/2018, DDS has confirmed that support for IE is limited to IE 11 and Ed
 
 ## Login.gov
 
-You'll need accounts for login.gov and the login.gov sandbox.  These will require two-factor authentication, so have your second factor (one of: phone, authentication app, security key, CAC) on hand.
-To create an account at login.gov, use your regular `truss.works` email and follow [the official instructions](https://login.gov/help/creating-an-account/how-to-create-an-account/).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -427,6 +431,21 @@ To fix the missing variables issue, you can do one of the following things:
 * `export DB_DEBUG=0`
   * We default this to be true in `.envrc`, but if you want to silence the sql logs locally, you can set this to 0.
 
+##### Troubleshooting direnv & chamber
+
+Make sure you have the latest version of Chamber that supports the `env` command
+option. If may run into the following error if the version of Chamber you have
```
felipe-lee, line 437: I think there's a typo here at the beginning of the second sentence. 
Like I think it should be "You may run..." instead of "If may run..."

Otherwise this is great!

README.md
```diff
@@ -103,10 +103,14 @@ As of 3/6/2018, DDS has confirmed that support for IE is limited to IE 11 and Ed
...
-To create an account in the sandbox, follow the same instructions, but [in the sandbox server](https://idp.int.identitysandbox.gov/sign_up/enter_email).  Do _not_ use your regular email address in the sandbox.
-**Tip**: You can use the plus sign `+` to create a new truss email address.  `name+some_string@truss.works` will be treated as a new address, but will be routed to `name@truss.works`.
+You'll need accounts for login.gov and the login.gov sandbox.  These will
+require two-factor authentication, so have your second factor (one of: phone,
+authentication app, security key, CAC) on hand.  To create an account at
+login.gov, use your regular `truss.works` email and follow [the official
+instructions](https://login.gov/help/creating-an-account/how-to-create-an-account/).
+To create an account in the sandbox, follow the same instructions, but [in the
+sandbox server](https://idp.int.identitysandbox.gov/sign_up/enter_email).  Do
+_not_ use your regular email address in the sandbox.
```
felipe-lee, line 113: I think we should leave this part in
```markdown
**Tip**: You can use the plus sign `+` to create a new truss email address. `name+some_string@truss.works` will be treated as a new address, but will be routed to `name@truss.works`.
```

Because we need to do this a lot, especially on the customer side (I probably created like 10 accounts this way just for demo prep one time). Maybe it should be updated to explicitly say that you should use these aliased emails on the customer side.

There should also probably a comment indicating that for the office side you likely _do_ need to use your regular email address as someone likely already set you up with the proper auths using your real email address.

rogeruiz:
Thanks for the comments @felipe-lee. I'm currently leaving this as a draft until I completely go through the docs for now.

felipe-lee (approved):
LGTM

PR #7039, opened 2021-07-21
Enables Milmove system to handle repeated orders amendment cycles

Ryan-Koch:
## Description

These changes should make it so that when the customer uploads an additional orders amendment after the first one, that the TOO app will be able to notice and flag the new set of orders in the same way that occurs when this happens for the first time. The way this is handled is by ensuring that when we go through the `MoveRouter.Submit()` path that we clear out the orders acknowledgement data field for their `Orders` if it already exists. This then allows the flags to reset on the client side on the office app and effectively puts things in the expected situation when an orders amendment process starts.

## Setup

```sh
# get some data
make db_dev_e2e_populate

# start your server and client up
make server_run
make client_run
```

Once you're up and running you can navigate over to the `AMENDO` move in the office app and make sure that things are all approved. After that, log in as the customer for this move: `hhg@multiple.pickup.amendedOrders.submitted` and go ahead and upload a new orders document. When you hit save after uploading you should then be able to go look at the same move as a TOO and see the flag along with the status change to `Approvals Requested`.

By flag, I mean this visual:
image

### DIFF
--- a/pkg/services/move/move_router.go
+++ b/pkg/services/move/move_router.go
@@ -71,7 +71,7 @@ func (router moveRouter) Submit(move *models.Move) error {
 				ordersForMove.AmendedOrdersAcknowledgedAt = nil
 				_, err = tx.ValidateAndSave(&ordersForMove)
 				if err != nil {
-					router.logger.Error("failure resetting orders AmendedOrdersAcknowledgeAt field when routing move with amended orders to office user / TOO queue", zap.Error(err))
+					router.logger.Error("failure resetting orders AmendedOrdersAcknowledgeAt field when routing move submission with amended orders ", zap.Error(err))
 					return err
 				}
 				router.logger.Info("Successfully reset orders acknowledgement")
--- a/pkg/services/move/move_router.go
+++ b/pkg/services/move/move_router.go
@@ -48,10 +48,38 @@ func (router moveRouter) Submit(move *models.Move) error {
 		}
 		router.logger.Info("SUCCESS: Move sent to services counseling")
 	} else if move.Orders.UploadedAmendedOrders != nil {
-		err = router.SendToOfficeUser(move)
-		if err != nil {
-			router.logger.Error("failure routing move with amended orders to office user / TOO queue", zap.Error(err))
-			return err
+		router.logger.Info("Move has amended orders")
+		transactionError := router.db.Transaction(func(tx *pop.Connection) error {
+			err = router.SendToOfficeUser(move)
+			if err != nil {
+				router.logger.Error("failure routing move with amended orders to office user / TOO queue", zap.Error(err))
+				return err
+			}
+			// Let's get the orders for this move so we can wipe out the acknowledgement if it exists already (from a prior orders amendment process)
+			var ordersForMove models.Order
+			err = router.db.Find(&ordersForMove, move.OrdersID)
+			if err != nil {
+				return err
+			}
+			// Here we'll nil out the value (if it's set already) so that on the client-side we'll see view this change
+			// in status as 'new orders' that need acknowledging by the TOO.
+			// We shouldn't need more complicated logic here since we only hit this point from calling Submit().
+			// Other circumstances like new MTOServiceItems will be calling SendToOfficeUser() directly.
+			router.logger.Info("Determining whether there is a preexisting orders acknowledgement")
+			if ordersForMove.AmendedOrdersAcknowledgedAt != nil {
+				router.logger.Info("Move has a preexisting acknowledgement")
+				ordersForMove.AmendedOrdersAcknowledgedAt = nil
+				_, err = router.db.ValidateAndSave(&ordersForMove)
+				if err != nil {
+					router.logger.Error("failure resetting orders AmendedOrdersAcknowledgeAt field when routing move with amended orders to office user / TOO queue", zap.Error(err))
+					return err
+				}
+				router.logger.Info("Successfully reset orders acknowledgement")
+			}
+			return nil
+		})
+		if transactionError != nil {
+			return transactionError
 		}
 		router.logger.Info("SUCCESS: Move with amended orders sent to office user / TOO queue")
 	} else {
--- a/pkg/services/move/move_router.go
+++ b/pkg/services/move/move_router.go
@@ -52,7 +52,7 @@ func (router moveRouter) Submit(move *models.Move) error {
 		transactionError := router.db.Transaction(func(tx *pop.Connection) error {
 			err = router.SendToOfficeUser(move)
 			if err != nil {
-				router.logger.Error("failure routing move with amended orders to office user / TOO queue", zap.Error(err))
+				router.logger.Error("failure routing move submission with amended orders", zap.Error(err))
 				return err
 			}
 			// Let's get the orders for this move so we can wipe out the acknowledgement if it exists already (from a prior orders amendment process)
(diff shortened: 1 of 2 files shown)


pkg/services/move/move_router.go
```diff
@@ -48,10 +48,38 @@ func (router moveRouter) Submit(move *models.Move) error {
 		}
 		router.logger.Info("SUCCESS: Move sent to services counseling")
 	} else if move.Orders.UploadedAmendedOrders != nil {
-		err = router.SendToOfficeUser(move)
-		if err != nil {
-			router.logger.Error("failure routing move with amended orders to office user / TOO queue", zap.Error(err))
-			return err
+		router.logger.Info("Move has amended orders")
+		transactionError := router.db.Transaction(func(tx *pop.Connection) error {
```
Ryan-Koch, line 52: Wrapped this in a transaction because we'll potentially wind up changing both the `Move` and the `Orders` values. I suspect we don't want to do one without the other if one fails.

pkg/services/move/move_router.go
```diff
@@ -48,10 +48,38 @@ func (router moveRouter) Submit(move *models.Move) error {
...
+			err = router.SendToOfficeUser(move)
+			if err != nil {
+				router.logger.Error("failure routing move with amended orders to office user / TOO queue", zap.Error(err))
+				return err
+			}
+			// Let's get the orders for this move so we can wipe out the acknowledgement if it exists already (from a prior orders amendment process)
+			var ordersForMove models.Order
+			err = router.db.Find(&ordersForMove, move.OrdersID)
```
monfresh, line 60: Did you mean to use the `tx` variable instead of `router.db`? Otherwise, it won't be using the transaction connection.

pkg/services/move/move_router.go
```diff
@@ -48,10 +48,38 @@ func (router moveRouter) Submit(move *models.Move) error {
...
+			}
+			// Here we'll nil out the value (if it's set already) so that on the client-side we'll see view this change
+			// in status as 'new orders' that need acknowledging by the TOO.
+			// We shouldn't need more complicated logic here since we only hit this point from calling Submit().
+			// Other circumstances like new MTOServiceItems will be calling SendToOfficeUser() directly.
+			router.logger.Info("Determining whether there is a preexisting orders acknowledgement")
+			if ordersForMove.AmendedOrdersAcknowledgedAt != nil {
+				router.logger.Info("Move has a preexisting acknowledgement")
+				ordersForMove.AmendedOrdersAcknowledgedAt = nil
+				_, err = router.db.ValidateAndSave(&ordersForMove)
```
monfresh, line 72: Same comment here about using `tx`

pkg/services/move/move_router.go
```diff
@@ -48,10 +48,38 @@ func (router moveRouter) Submit(move *models.Move) error {
...
+			err = router.db.Find(&ordersForMove, move.OrdersID)
+			if err != nil {
+				return err
+			}
+			// Here we'll nil out the value (if it's set already) so that on the client-side we'll see view this change
+			// in status as 'new orders' that need acknowledging by the TOO.
+			// We shouldn't need more complicated logic here since we only hit this point from calling Submit().
+			// Other circumstances like new MTOServiceItems will be calling SendToOfficeUser() directly.
+			router.logger.Info("Determining whether there is a preexisting orders acknowledgement")
+			if ordersForMove.AmendedOrdersAcknowledgedAt != nil {
```
monfresh, line 69: I wonder if this logic belongs in the service that creates the amended orders as opposed to here. The Move Router shouldn't be modifying fields. Its responsibility is only to send moves to the appropriate place based on certain conditions. If we want to support multiple amended orders, I think it would make more sense to reset the `AmendedOrdersAcknowledgedAt` field in the same place that handles uploading/saving the latest amended orders.

monfresh: Keeping order and move updates separate also eliminates the added complexity of transactions.

pkg/services/move/move_router.go
Ryan-Koch: Nice catch, I definitely wanted to do that instead. I'll fix that.

pkg/services/move/move_router.go
Ryan-Koch: That's a fair point, I agree it that it would be a nicer way to organize this. I think the tricky bit to it is that there's two steps the customer is going through for this, there's the upload step and a step where they're doing hitting a button to save their changes. That second button is what's doing the move submission bit. Because the interaction and subsequently the API requests are separated out like that, I figured it made more sense to carry this change in with the submission action, since the customer could in theory back out of the process before doing that part.

pkg/services/move/move_router.go
```diff
@@ -48,10 +48,38 @@ func (router moveRouter) Submit(move *models.Move) error {
...
 	} else if move.Orders.UploadedAmendedOrders != nil {
-		err = router.SendToOfficeUser(move)
-		if err != nil {
-			router.logger.Error("failure routing move with amended orders to office user / TOO queue", zap.Error(err))
-			return err
+		router.logger.Info("Move has amended orders")
+		transactionError := router.db.Transaction(func(tx *pop.Connection) error {
+			err = router.SendToOfficeUser(move)
+			if err != nil {
+				router.logger.Error("failure routing move with amended orders to office user / TOO queue", zap.Error(err))
```
duncan-truss, line 55: I don't think this is unclear necessarily but it could end up in services counseling but this is just an error for failure so it's not going anywhere I guess.

Ryan-Koch: That's reasonable. I'll clear up the language so that it mentions failing to submit a move with amended orders. That language is more in line with what the customer action is on that end.

pkg/services/move/move_router_test.go
```diff
@@ -116,6 +117,32 @@ func (suite *MoveServiceSuite) TestMoveSubmission() {
...
+			Move: models.Move{
+				Status: models.MoveStatusAPPROVED,
+			},
+			Order: order,
+		})
+		suite.NotNil(move.Orders.AmendedOrdersAcknowledgedAt)
+		err := moveRouter.Submit(&move)
+		suite.NoError(err)
+		var updatedOrders models.Order
+		err = suite.DB().Find(&updatedOrders, move.OrdersID)
```
duncan-truss, line 140: You may be able to call Load() here on the move instead of finding the orders but I think they are basically equivalent.

duncan-truss (approved):

pkg/services/move/move_router.go
monfresh: One potential issue with this logic is if a customer has already uploaded amended orders, and the TOO has already reviewed them, then if the customer adds a new PPM shipment, this will trigger this logic, which will reset the `AmendedOrdersAcknowledgedAt` even though it shouldn't be. What's strange is that I see the log statements from these lines,  but the `AmendedOrdersAcknowledgedAt` field does not end up being nil, so it all works out in the end, but I'm wondering if there's a bug hiding somewhere.

monfresh: OK, I think I see the issue. It's a transaction issue, much like the one I ran into last week, and so once we fix the way we use the DB, this bug will surface and do the wrong thing. When amended orders are uploaded, the final DB change we make to the order is setting the `AmendedOrdersAcknowledgedAt` field to `nil`, and so that change gets successfully made in the DB. However, when a customer adds a new shipment, we are setting the `AmendedOrdersAcknowledgedAt` field to `nil` inside a transaction, but then the handler calls `saveMoveDependencies`, but it's using a separate DB connection, and so it doesn't see that the Order was updated, and when it saves it, it doesn't modify the `AmendedOrdersAcknowledgedAt` field.

So, this shows that my initial thought about separating the scenarios makes sense because we don't want to be triggering this logic when a customer adds a new shipment.

PR #7040, opened 2021-07-21
Remove Erin's cac access

LeDeep:
## Description

Removing Erin's cac access.

* [Jira story](https://dp3.atlassian.net/browse/MMDM-3015) for this change


reggieriser (approved):
LGTM.  I ran the migrations and saw that the record was no longer in the `client_certs` table.

PR #7041, opened 2021-07-21
MB-8801 add system error component

mlm55:
## Description

We want a general error that we can display on the Office and Customer apps when there are errors outside of the user’s control. This PR creates a SystemError component and adds it to storybook.

## Setup

```sh
make storybook
```
http://localhost:6006/?path=/story/components-system-error--system-error-component

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8801) for this change

### DIFF
--- a/src/components/SystemError/index.jsx
+++ b/src/components/SystemError/index.jsx
@@ -1,0 +1,10 @@
+import React from 'react';
+
+const SystemError = () => (
+  <div className="usa-alert--system-error">
+    Something isn&apos;t working, but we&apos;re not sure what. Wait a minute and try again. If that doesn&apos;t fix
+    it, contact the Technical Help Desk and give them this code: <strong>[trace ID]</strong>.
+  </div>
+);
+
+export default SystemError;
--- a/src/components/SystemError/SystemError.stories.jsx
+++ b/src/components/SystemError/SystemError.stories.jsx
@@ -1,0 +1,14 @@
+import React from 'react';
+
+import SystemError from './index';
+
+export default {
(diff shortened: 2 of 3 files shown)


src/components/SystemError/index.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+const SystemError = () => (
+  <div className="usa-alert--system-error">
```
Ronolibert, line 4: Do you know where the docs are for this `usa-alert--system-error` class? I've looked on uswds but can't find its existence

Ronolibert (approved):
Just had a question for curiosity but otherwise looks good

src/components/SystemError/index.jsx
mlm55: src/shared/styles/_overrides.scss

src/components/SystemError/index.jsx
```diff
@@ -0,0 +1,10 @@
...
+    Something isn&apos;t working, but we&apos;re not sure what. Wait a minute and try again. If that doesn&apos;t fix
+    it, contact the Technical Help Desk and give them this code: <strong>[trace ID]</strong>.
```
YanZ777, line 6: I think we want to be able to pass the trace id as props, but I am not sure if that should be handled in this story or in the one that actually hooks up the trace id.

YanZ777 (approved):
Approving as my question is not blocking, and I think putting the work in a following story will not significantly increase the scope,

cosburn (approved):
Looks good!

PR #7044, opened 2021-07-22
[MB-7727] Remove direct connection to Syncada via SFTP

pearl-truss:
## Description

This PR removes the direct connection to syncada via sftp along with cli commands that are now redundant since we have the job runner task

## Reviewer Notes

* Are there any other pieces of code I should also delete?
* Is there I deleted that should be brought back?
* I kept the read from syncada files because they were being used with gex sftp client
* I haven't been able to generate an EDI locally or in staging so I'm not sure what's wrong but that's what I've been looking into

## Setup

```sh
echo "Code goes here"
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7250) for this change


pearl-truss:
Whoops, I had multiple branches for this story and I accidentally opened a PR off of the wrong branch

PR #7045, opened 2021-07-22
[MB-7727] remove sftp connect

pearl-truss:
## Description

This PR removes the direct connection to syncada via sftp along with cli commands that are now redundant since we have the job runner task

## Reviewer Notes

* Are there any other pieces of code I should also delete?
* Is there I deleted that should be brought back?
* I kept the read from syncada files because they were being used with gex sftp client

## Setup

```sh
{
  "body": {
    "sendToSyncada": true,
    "readFromSyncada": false,
    "deleteFromSyncada": false
  }
}
```

And then try manually triggering the job runner
```sh
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7250) for this change


gidjin (commented):
looks good so far. I had a question about `pkg/services/invoice/syncada_sftp_sender.go` should it be removed or changed? It appears to depend on the syncada specific environment vars as apposed to the gex ones.

Also looks like it's used in `pkg/services/payment_request/payment_request_reviewed_processor.go` which might need to be updated as well.

gidjin (approved):
Changes look good. We may want to rename some of the methods but I don't think we need to at this time.

PR #7069, opened 2021-07-26
MB-7432 Update fetchMTOUpdates query

sandy-wright:
## Description

This PR adds a new Prime endpoint - `fetchMTOUpdatesFast`. This will allow us to test the new functionality without forcing folks to roll off the old `fetchMTOUpdates` immediately.

## Setup

Start your server:

```sh
make db_dev_e2e_populate && server_run
```

Then using [Postman](https://github.com/transcom/prime_api_deliverable/wiki/Using-Postman), use the Prime API to hit the `fetchMTOUpdatesFast` endpoint. Compare this with the `fetchMTOUpdates` endpoint. Mess around with `since` times as well to try and filter the results down. 

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7432) for this change.
* [Performance Testing `fetchMTOUpdates`](https://github.com/transcom/mymove/wiki/Performance-Testing-fetchMTOUpdates) explains more about the approach used.


eamahanna (commented):
Do you know how many people are using fetchMTOUpdates and how large of an effort it will be to change to the fast version?

sandy-wright:
> Do you know how many people are using fetchMTOUpdates and how large of an effort it will be to change to the fast version?

The biggest one is in `cmd/prime-api-dev/payment_requests.go`, which I _think_ @transcom/truss-a-team uses pretty heavily in their testing and development. It fully relies on `fetchMTOUpdates` to get all Prime payment requests. There are a couple ways we could address this:

- Add an endpoint to the Prime API that lists payment requests. Frankly, I feel like we should probably have one of these for service items and shipments too (with paging and the ability to filter so they don't turn into messes like `fetchMTOUpdates`).
- Update the `listMoves` endpoint in the Support API to be able to filter for only Prime-available moves if desired. Right now it lists everything, all the time

Neither would be super time-intensive, with the Support API one being the easiest to pick up. @transcom/truss-a-team would probably have an opinion on which method they'd prefer

sandy-wright:
I felt like having both endpoints live might be preferable to some folks who wanted to spend more time looking at the query, but maybe not. I can rip it out and do the Support API update if y'all think that is the best way to go

duncan-truss:
@sandy-wright It looks like adding more data got rid of some of the sequential scans that I think Lindsay alluded to.  It's definitely faster it may be hard to compare with the much smaller response size.

I don't see indexes on the `updated_at` columns of the various table so those could be worth adding to the list.

When I didn't send a value for the since parameter I was still seeing the zero timestamp value in the SQL statement there may be something getting initialized you didn't expect somewhere.

```
Hash Join  (cost=258.76..356.45 rows=497 width=155)
  Hash Cond: (orders.id = moves.orders_id)
  Join Filter: ((moves.updated_at >= '2021-07-16 00:00:00'::timestamp without time zone) OR (orders.updated_at >= '2021-07-16 00:00:00'::timestamp without time zone) OR (hashed SubPlan 1))
  ->  Seq Scan on orders  (cost=0.00..81.08 rows=2108 width=35)
  ->  Hash  (cost=46.36..46.36 rows=497 width=144)
        ->  Seq Scan on moves  (cost=0.00..46.36 rows=497 width=144)
              Filter: ((available_to_prime_at IS NOT NULL) AND show)
  SubPlan 1
    ->  HashAggregate  (cost=166.43..198.24 rows=3181 width=16)
          Group Key: mto_shipments.move_id
          ->  Append  (cost=0.00..158.48 rows=3181 width=16)
                ->  Seq Scan on mto_shipments  (cost=0.00..78.29 rows=2103 width=16)
                      Filter: (updated_at >= '2021-07-16 00:00:00+00'::timestamp with time zone)
                ->  Seq Scan on mto_service_items  (cost=0.00..18.46 rows=597 width=16)
                      Filter: (updated_at >= '2021-07-16 00:00:00+00'::timestamp with time zone)
                ->  Seq Scan on payment_requests  (cost=0.00..14.01 rows=481 width=16)
                      Filter: (updated_at >= '2021-07-16 00:00:00'::timestamp without time zone)
```

sandy-wright:
> When I didn't send a value for the since parameter I was still seeing the zero timestamp value in the SQL statement there may be something getting initialized you didn't expect somewhere.
available_to_prime_at IS NOT NULL) AND show)

Ooh nice catch, that was a bug in my handler! Whoops!

eamahanna (approved):
LGTM

sandy-wright:
I'm going to go ahead and merge this - but I welcome more feedback on the query for future PRs! Some notes on work we'll have in future tickets:

- Adding indexes on move IDs and orders ID
- Add an index on `availableToPrimeAt` and `show`
- Maybe adding an index on `updatedAt` (although I'm torn about this and want to do some more research to understand why I feel torn about it)
- Eventually deprecating the old `fetchMTOUpdates` endpoint

Some things I'm thinking about but we might not change:
- Experimenting with UNION ALL instead of UNION to see if there's a notable performance difference
- Should we... rename this endpoint? Maybe `listMoves`? `fetchMTOUpdates` is such a mouthful - I'm gonna keep simmering that thought.

PR #7070, opened 2021-07-26
[MB-8801] System Error component style changes

Ronolibert:
## Description

Changed system error font size to match design and remove duplicate storybook story to avoid confusion

## Reviewers notes

this is a continuation of https://github.com/transcom/mymove/pull/7041 but there was a duplicate system error storybook story that we wanted to remove in lieu of a dedicated system error component

## Setup
`make storybook`
SystemError component should be under `Alerts` in storybook

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8801) for this change

## Screenshots

Before:
image

After: 
image


lynzt (changes requested):
let's keep the break in the `system alert` text to ensure a break looks ok

src/components/SystemError/SystemError.stories.jsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import SystemError from './index';
 
 export default {
-  title: 'Components/System Error',
+  title: 'Components/Alerts/System Error',
```
felipe-lee, line 6: I'm not sure if putting this in the same section as the other alerts is confusing or not since usually our folders only contain variations on the same type of component. Though I guess `Fieldsets` section has multiple components in one folder so maybe it's not confusing.

Ronolibert: Yeah we're not sure. This was originally in the Alerts section but not its own component. I think it makes sense for it to live there for now since it is used in the same situations as other alerts even though it doesn't have the same design language

lynzt: Design wanted to keep this under the `Alert` folder... @cosburn 
This is probably a good topic for the FE group...

lynzt (approved):
LGTM 🚢

cosburn (approved):
Nicely done!

YanZ777 (approved):

PR #7083, opened 2021-07-28
MB-8774 refactor edit orders paage to use formik

YanZ777:
## Description

Update the EditOrders and EditOrdersForm to use Formik.
Add tests for both. 
Add storybook component for the EditOrdersForm. 
There should be no visual difference from switching to Formik, but adding a picture anyways.

## Reviewer Notes

These changes do not fix the existing bugs with the edit orders page, that will be handled in [this jira ticket](https://dp3.atlassian.net/browse/MB-8511).
Are the tests written up ok?

## Setup

```sh
make client_run
make server_run
make storybook
make client_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8774) for this change

## Screenshots

milmovelocal_3000_moves_3a8c9f4f-7344-4f18-9ab5-0de3ef57b901_review_edit-orders


src/components/Customer/EditOrdersForm/EditOrdersForm.test.jsx
```diff
@@ -0,0 +1,422 @@
...
+    });
+    const submitButton = screen.getByRole('button', { name: 'Save' });
+
+    await waitFor(() => {
+      expect(submitButton).toBeDisabled();
+      expect(
+        screen.queryByText(
+          'You entered the same duty station for your origin and destination. Please change one of them.',
+        ),
+      ).toBeInTheDocument();
```
felipe-lee, line 234: I think it's best keep only the main thing we need to wait for wrapped in the `waitFor` block. That way future us doesn't think we have to wait for all of it. Like in this case, the text should be there once the button is disabled.

src/components/Customer/EditOrdersForm/EditOrdersForm.test.jsx
```diff
@@ -0,0 +1,422 @@
...
+  it('validates the new duty station against the current duty station', async () => {
+    render(<EditOrdersForm {...testProps} currentStation={{ name: 'Luke AFB' }} />);
+
+    userEvent.selectOptions(screen.getByLabelText('Orders type'), 'PERMANENT_CHANGE_OF_STATION');
+    userEvent.type(screen.getByLabelText('Orders date'), '08 Nov 2020');
+    userEvent.type(screen.getByLabelText('Report-by date'), '26 Nov 2020');
+    userEvent.click(screen.getByLabelText('No'));
+
+    // Test Duty Station Search Box interaction
+    fireEvent.change(screen.getByLabelText('New duty station'), { target: { value: 'AFB' } });
```
felipe-lee, line 219: Can this be switched to use `userEvent.type` or whatever the equivalent is in this case (i think we are just typing though). Or does that not work here for some reason?

src/components/Customer/EditOrdersForm/EditOrdersForm.test.jsx
```diff
@@ -0,0 +1,422 @@
...
+  it('submits the form when its valid', async () => {
+    render(<EditOrdersForm {...testProps} existingUploads={existingUploads} />);
+
+    userEvent.selectOptions(screen.getByLabelText('Orders type'), 'PERMANENT_CHANGE_OF_STATION');
+    userEvent.type(screen.getByLabelText('Orders date'), '08 Nov 2020');
+    userEvent.type(screen.getByLabelText('Report-by date'), '26 Nov 2020');
+    userEvent.click(screen.getByLabelText('No'));
+
+    // Test Duty Station Search Box interaction
+    fireEvent.change(screen.getByLabelText('New duty station'), { target: { value: 'AFB' } });
```
felipe-lee, line 267: Same comment here as above about `userEvent.type` vs `fireEvent.change`

src/components/Customer/EditOrdersForm/EditOrdersForm.test.jsx
```diff
@@ -0,0 +1,422 @@
...
+    };
+
+    it('pre-fills the inputs', async () => {
+      render(<EditOrdersForm {...testProps} initialValues={testInitialValues} />);
+
+      expect(await screen.findByRole('form')).toHaveFormValues({
+        new_duty_station: 'Yuma AFB',
+      });
+
+      expect(await screen.findByLabelText('Orders type')).toHaveValue(testInitialValues.orders_type);
```
felipe-lee, line 343: Not something to change, more just for info. After the first `await screen.findBy*`, the rest can be `screen.getBy*` without `await`. The only time you need to then do `await screen.findBy*` again should be if you make changes to the page and need to wait for the updates to take effect.

felipe-lee (commented):
This looks great! Sorry about the delay in reviewing. One thing I noticed that seems to be different for some reason between the current version of the edit forms page and this pr's is that I can still upload documents currently, but in the PR version, I can't. Not sure if maybe there's some other difference in the service members I'm using that's causing the difference in behavior.

Staging:
image

PR:
Screen Shot 2021-07-29 at 17 05 35

src/components/Customer/EditOrdersForm/EditOrdersForm.jsx
```diff
@@ -1,45 +1,62 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { reduxForm, Field } from 'redux-form';
+import { Formik, Field } from 'formik';
```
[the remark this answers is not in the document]
YanZ777: I think the only difference is that the file upload is also included here.

src/components/Customer/EditOrdersForm/EditOrdersForm.jsx
```diff
@@ -1,45 +1,62 @@
...
+    new_duty_station: Yup.object()
+      .shape({
+        name: Yup.string().notOneOf(
+          [currentStation?.name],
+          'You entered the same duty station for your origin and destination. Please change one of them.',
+        ),
+      })
+      .nullable()
+      .required('Required'),
+  });
```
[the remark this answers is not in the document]
YanZ777: I left them separate mostly because the schema needs access to the orders type option and the current station. 
Should I make a utility function that returns a schema though instead?

src/components/Customer/EditOrdersForm/EditOrdersForm.jsx
```diff
@@ -51,18 +68,47 @@ const EditOrdersForm = ({
...
+              <DatePickerInput
+                name="issue_date"
+                label="Orders date"
                 required
+                renderInput={(input) => (
+                  <>
+                    {input}
+                    <Hint>
+                      <p>Date your orders were issued.</p>
+                    </Hint>
```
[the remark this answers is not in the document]
YanZ777: Yes, the hint text should be moved! Thanks for the reminder!

src/components/Customer/EditOrdersForm/EditOrdersForm.jsx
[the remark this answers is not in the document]
felipe-lee: Ah ok. Well then idk if it would be best to combine them or not. We have other forms that we use for both add and edit, but that can also make them really clunky. I'm fine with keeping them separate and combining if we find it's too much overhead to maintain them separately.

src/components/Customer/EditOrdersForm/EditOrdersForm.jsx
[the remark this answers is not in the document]
felipe-lee: Ah I didn't think about that. I think a function could be nice to have, but maybe overkill unless we find we need this in more places.

src/components/Customer/EditOrdersForm/EditOrdersForm.test.jsx
YanZ777: That's a good point! I haven't tried.

src/pages/MyMove/EditOrders.jsx
```diff
@@ -25,25 +24,44 @@ import {
...
+import { OrdersShape } from 'types/customerShapes';
+import { formatYesNoInputValue, formatYesNoAPIValue } from 'utils/formatters';
+import { ORDERS_TYPE_OPTIONS } from 'constants/orders';
+import { dropdownInputOptions } from 'shared/formatters';
 import { EntitlementShape, ExistingUploadsShape } from 'types';
-import 'scenes/Review/Review.css';
+import { DutyStationShape } from 'types/dutyStation';
 
-const EditOrders = ({
+export const EditOrders = ({
```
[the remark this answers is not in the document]
YanZ777: Exported for the tests!

src/components/Customer/EditOrdersForm/EditOrdersForm.jsx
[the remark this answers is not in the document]
YanZ777: Actually the way the DatePickerInput is set up, I can't move the hint text between the label and the field.
The MtoShipmentForm uses FieldSet to get around this issue. So it's still a question for @transcom/truss-design!

src/components/Customer/EditOrdersForm/EditOrdersForm.jsx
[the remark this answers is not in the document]
YanZ777: Onboarding has filling out the orders information and uploading orders as separate pages. I think if we decide to combine them, that would be a question for design with regards to changing the onboarding flow.

YanZ777:
> This looks great! Sorry about the delay in reviewing. One thing I noticed that seems to be different for some reason between the current version of the edit forms page and this pr's is that I can still upload documents currently, but in the PR version, I can't. Not sure if maybe there's some other difference in the service members I'm using that's causing the difference in behavior.

Hmmm the upload feature should definitely still be there. Let me look into it more.

YanZ777:
> This looks great! Sorry about the delay in reviewing. One thing I noticed that seems to be different for some reason between the current version of the edit forms page and this pr's is that I can still upload documents currently, but in the PR version, I can't. Not sure if maybe there's some other difference in the service members I'm using that's causing the difference in behavior.

Ok this should be fixed!

src/components/Customer/EditOrdersForm/EditOrdersForm.test.jsx
felipe-lee: Looks like it worked! Hopefully that makes it less flaky.

src/components/Customer/EditOrdersForm/EditOrdersForm.jsx
```diff
@@ -51,67 +77,100 @@ const EditOrdersForm = ({
...
-          </form>
-        </div>
-      </div>
-    </div>
+
+            <div className={formStyles.formActions}>
+              <WizardNavigation
+                editMode
+                onCancelClick={onCancel}
+                disableNext={!isValid || isSubmitting}
```
felipe-lee, line 138: I think it might be good to keep the check for having at least one document since users can in theory delete files (though not in practice rn because it's broken). Though I guess the save button doesn't actually do anything with uploads 🤔 so I guess it's ok to take it out like you did in the last commit.

felipe-lee (approved):
Woo, looks as expected now!

src/components/Customer/EditOrdersForm/EditOrdersForm.jsx
YanZ777: I added the uploaded_orders to the schema, so the schema has a min(1) on the array.

felipe-lee: Oh right! Awesome, thanks for pointing that out. I saw it, but my head did not make the connection for some reason...We don't have all our custom emoji here 😞

amanda5iu (approved):
Looking good!

cosburn (approved):
UI looks good!

PR #7089, opened 2021-07-28
[MB-1572] Explicit look up function for billed cubic feet

SirenaBorracha:
## Description

New lookup has been added for `CubicFeetBilled`.
We bill based on the size of the crate, and always bill for at least 4 cubic feet, so this lookup takes the crate size, and applies the 4 cubic foot minimum.

This lookup uses the lookup for `CubicFeetCrating`, parses it from a string, and checks it against the minimum. An alternative approach would have been to recalculate the volume in this lookup. This would have allowed us to skip a string conversion, which would be great, but it would have also led to some duplication of logic. 

We also did some refactoring in this PR:
- Added a new unit: `CubicFeet`
- Added a `Volume()` helper function to `MTOServiceItemDimension` so we could move the calculation out of the `CubicFeetCrating` lookup

## Setup
We don't have any code that directly uses this lookup yet, so there's not much testing to do other than making sure tests pass.
```sh
make server_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-1572) for this change


pkg/models/mto_service_item_dimensions.go
```diff
@@ -56,3 +56,9 @@ func (m *MTOServiceItemDimension) Validate(tx *pop.Connection) (*validate.Errors
 func (m MTOServiceItemDimension) TableName() string {
 	return "mto_service_item_dimensions"
 }
+
+// Volume calculates Length x Height x Width
+func (m *MTOServiceItemDimension) Volume() unit.CubicFeet {
+	volume := m.Length.ToFeet() * m.Width.ToFeet() * m.Height.ToFeet()
```
gidjin, line 62: The conversion to feet could loose accuracy as it will be float math vs int math. Maybe it's not a big deal here, but I think we should return the volume in ThousandthInch and only call `ToFeet()` when needed at the very end.

pkg/payment_request/service_param_value_lookups/cubic_feet_billed_lookup.go
```diff
@@ -0,0 +1,39 @@
...
+
+// CubicFeetBilledLookup does lookup for CubicFeetBilled
+type CubicFeetBilledLookup struct {
+	Dimensions models.MTOServiceItemDimensions
+}
+
+func (c CubicFeetBilledLookup) lookup(keyData *ServiceItemParamKeyData) (string, error) {
+	// call CubicFeetCratingLookup
+	// convert string to number
+	// do the math
```
gidjin, line 22: Probably can remove these comments or maybe restate them

pkg/payment_request/service_param_value_lookups/cubic_feet_billed_lookup.go
```diff
@@ -0,0 +1,39 @@
...
+
+	cubicFeet, err := CubicFeetCratingLookup(c).lookup(keyData)
```
gidjin, line 24: if Volume is now a function we can call, couldn't we just recalculate the volume here instead of having to parse from a string? I believe we do something similar in other look ups that depend on data calculated in external look ups.

pkg/payment_request/service_param_value_lookups/cubic_feet_crating_lookup.go
```diff
@@ -18,12 +16,7 @@ func (c CubicFeetCratingLookup) lookup(keyData *ServiceItemParamKeyData) (string
 	// look for the first crating dimension.
 	for _, dimension := range c.Dimensions {
 		if dimension.Type == models.DimensionTypeCrate {
-			lengthFeet := dimension.Length.ToFeet()
-			heightFeet := dimension.Height.ToFeet()
-			widthFeet := dimension.Width.ToFeet()
-
-			volume := lengthFeet * heightFeet * widthFeet
-			return fmt.Sprintf("%.2f", volume), nil
+			return dimension.Volume().String(), nil
```
gidjin, line 19: This is where I'd expect the `ToFeet` call to be.

pkg/models/mto_service_item_dimensions.go
mboren: Yeah that makes sense. I was originally doing this conversion first to avoid overflow, but I got my math wrong and we should be fine multiplying them all first. If we have a crate bigger than like 180ft x 180 ft x 180ft then we'll start to overflow ints for the volume. I'm not sure what the max crate size is, but i'm sure it's a lot less than that.

pkg/payment_request/service_param_value_lookups/cubic_feet_billed_lookup.go
mboren: Yeah, I can do that. I was waffling between that and the current implementation.

pkg/unit/volume.go
```diff
@@ -0,0 +1,22 @@
...
+
+// CubicFeet represents cubic feet
+type CubicFeet float64
+
+// CubicThousandthInch represents values in cubic thousandths of an inch
+type CubicThousandthInch int
+
+// String converts a CubicFeet value into a string
+func (c CubicFeet) String() string {
+	return fmt.Sprintf("%.2f", c)
```
gidjin, line 15: Wondering if this should be `%.3f` since we are storing thousandth of an inch. Maybe the design only called for 2 decimal places?

gidjin (approved):
Thanks for making the updates. I left one question but feel free to adjust it without a re-review pending an answer from documentation or design on whether we should return feet to 2 or 3 decimal places.

pkg/unit/volume.go
mboren: Yeah I only see 2 in the design, https://app.zeplin.io/project/60b8e558e0e96f9861d9d89d/screen/60d3ab32b21fe7baf90b4ff4

PR #7091, opened 2021-07-28
[MB-2504] SystemError on error office queue page

Ronolibert:
## Description

Show `SystemError` component when there is an error on any office page

## Setup
Locally, change `GetMovesQueueHandler` function located [here](https://github.com/transcom/mymove/blob/e1a30ff02bbdb771977b15c9ecad25ae9acbee9d/pkg/handlers/ghcapi/queues.go#L31) to this so it will always error:
```
func (h GetMovesQueueHandler) Handle(params queues.GetMovesQueueParams) middleware.Responder {

	if !session.IsOfficeUser() || !session.Roles.HasRole(roles.RoleTypeTOO) {
		logger.Error("user is not authenticated with TOO office role")
		return queues.NewGetMovesQueueForbidden()
	}

	return queues.NewGetMovesQueueInternalServerError()
}

```
Log in to a TOO
You should see the error
`make server_run`
`make office_client_run`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-2504) for this change

## Screenshots

Queue pages (TOO/TIO/TXO/Services Counselor)
image

TIO/TOO/TXO Move details
image

Services Counselor Move Details
image

### DIFF
--- a/src/components/SystemError/SystemError.module.scss
+++ b/src/components/SystemError/SystemError.module.scss
@@ -7,4 +7,10 @@
   padding-top: 0.8rem;
   padding-left: 1.6rem;
   padding-right: 1.6rem;
+  line-height: 20px;
+  & > a,
+  a:visited {
+    color: black;
+    text-decoration: underline;
+  }
 }
--- a/src/components/SystemError/SystemError.module.scss
+++ b/src/components/SystemError/SystemError.module.scss
@@ -1,3 +1,10 @@
+@import 'shared/styles/colors';
+
 .systemError {
   font-size: 1rem;
+  background-color: $warning;
+  padding-bottom: 0.8rem;
+  padding-top: 0.8rem;
+  padding-left: 1.6rem;
+  padding-right: 1.6rem;
 }
--- a/src/pages/Office/ServicesCounselingMoveInfo/ServicesCounselingMoveInfo.jsx
+++ b/src/pages/Office/ServicesCounselingMoveInfo/ServicesCounselingMoveInfo.jsx
@@ -1,8 +1,10 @@
 import React, { Suspense, lazy, useState, useEffect } from 'react';
 import { Switch, useParams, Redirect, Route, useHistory } from 'react-router-dom';
+import { useSelector } from 'react-redux';
 
 import 'styles/office.scss';
 import CustomerHeader from 'components/CustomerHeader';
+import SystemError from 'components/SystemError';
 import { servicesCounselingRoutes } from 'constants/routes';
 import { useTXOMoveInfoQueries } from 'hooks/queries';
 import LoadingPlaceholder from 'shared/LoadingPlaceholder';
@@ -18,7 +20,7 @@ const CustomerInfo = lazy(() => import('pages/Office/CustomerInfo/CustomerInfo')
 
 const ServicesCounselingMoveInfo = () => {
   const [customerEditAlert, setCustomerEditAlert] = useState(null);
-
+  const { hasRecentError, traceId } = useSelector((state) => state.interceptor);
   const onCustomerInfoUpdate = (alertType) => {
     if (alertType === 'error') {
       setCustomerEditAlert({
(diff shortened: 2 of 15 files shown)


src/components/SystemError/SystemError.module.scss
```diff
@@ -1,3 +1,10 @@
+@import 'shared/styles/colors';
+
 .systemError {
   font-size: 1rem;
+  background-color: $warning;
```
Ronolibert, line 5: I ported over the styles from `usa-alert--system-error` into this class because the stylesheet that `usa-alert--system-error` is a part of was not imported on the office app. Importing it would lead to many side unwanted side effects

src/pages/Office/index.test.jsx
```diff
@@ -77,10 +78,26 @@ describe('Office App', () => {
...
+            initialEntries={['/moves/queue']}
+          >
+            <ConnectedOffice />
+          </MockProviders>,
+        );
+        expect(screen.getByText('Technical Help Desk').closest('a')).toHaveAttribute(
+          'href',
+          'https://move.mil/customer-service#technical-help-desk',
+        );
+        expect(screen.getByTestId('system-error').textContent).toEqual(
```
Ronolibert, line 97: I used `data-testid` instead of getByText because the API behaves unexpectedly when there is a combination of plain text nodes mixed with other dom elements. More context: https://github.com/testing-library/dom-testing-library/issues/410

YanZ777 (approved):
The changes look good to me. 

Would this trace ID be useful during MilMob? 
Also, with regards to MilMob, we should check to see where people should be writing up their tickets. If it's not through service now, then I think someone needs to tell the milmob to not use the link in the error message.

Ronolibert:
@YanZ777 Yup! The traceId should help for logging purposes but currently this will only catch 500s

felipe-lee (commented):
Tested it out and works mostly as expected, but there is one issue. If you change focus to any other window and then come back, the page tries to reload and thus shows a different trace-id. It happens so fast it can even be hard to catch. I only noticed because I wanted to compare the visible trace id with the request and noticed the difference (and the fact that there were new request/responses in the network tab.

Ronolibert:
@felipe-lee Ahh yeah, that is sort of a limitation right now. We're not storing a stack of all the trace IDs that have errored. I think the same thing would happen if multiple endpoints failed consecutively. It'll visually cycle through all the trace IDs rapidly and mainly show the last one. Not sure we're going to tackle that issue for now but it is something of note for future iterations of it

cosburn (approved):
Looks good!

src/pages/Office/index.jsx
```diff
@@ -258,13 +272,23 @@ OfficeApp.propTypes = {
   userIsLoggedIn: PropTypes.bool,
   userRoles: UserRolesShape,
   activeRole: PropTypes.string,
+  hasRecentError: PropTypes.bool.isRequired,
+  traceId: PropTypes.string.isRequired,
+  history: PropTypes.shape({
+    location: PropTypes.shape({
+      pathname: PropTypes.string,
+    }),
+  }),
```
felipe-lee, line 281: Most of the office side uses `useHistory` I think instead of passing history around so it might be good to use that instead to be consistent.

felipe-lee:
> @felipe-lee Ahh yeah, that is sort of a limitation right now. We're not storing a stack of all the trace IDs that have errored. I think the same thing would happen if multiple endpoints failed consecutively. It'll visually cycle through all the trace IDs rapidly and mainly show the last one. Not sure we're going to tackle that issue for now but it is something of note for future iterations of it

@Ronolibert Ah ok, as long as we have it on our radar to address (or at least think through and decide if we want to), I'm good with how it works for now.

src/pages/Office/ServicesCounselingMoveInfo/ServicesCounselingMoveInfo.jsx
```diff
@@ -55,7 +57,15 @@ const ServicesCounselingMoveInfo = () => {
   return (
     <>
       <CustomerHeader order={order} customer={customerData} moveCode={moveCode} />
-
+      {hasRecentError && (
```
YanZ777, line 60: Maybe out of scope / not worthwhile, but should we consider taking this out into its own component?

PR #7118, opened 2021-08-03
[MB-8941] Reweighs and excess weight database models

duncan-truss:
## Description

Outcome 5 for reweighs/excess weight defined new requirements that we needed to extend our data model to meet.  The main change includes a new table for reweigh requests to be recorded per shipment.  This is an optional association for a shipment and we expect sometimes the prime will receive a request but not be able to perform the reweigh and update the weight.  In that scenario they will need to provide a justification as to why they could not complete the reweigh.

The new columns added to the mto_shipments table allow the TIO to cap the billable weight and provide a justification for what they've set the value to.

The new columns added to the moves table support linking an excess weight file upload from the Prime that the customer was counseled around that concern.

We did not need a column around max billable weight at the move level because authorized_weight on the entitlements table is going to fulfill this need to be updatable by both the TOO and the TIO.

## Reviewer Notes

Jacquie and others have expressed a need to store the derived final billable weight that a shipment will be priced at.  I know there was some concern over making sure this is always recalculated and does not get in a stale state but it sounds like it would simplify invoicing work from having to look in multiple places to make a determination.

I am under the impression that we still want to store all weights in pounds but in the UI I'm seeing cwt (1 cwt = 100 pounds?) used.  I know we're already doing this for some pricing parameters but does anyone foresee any issues there?

It might be a good exercise to think of what an ideal data model could look like now that we have many more weights and know who is primarily going to be updating them.  I tried to not make radical change here because that would make for a lot of reworking of api/payloads and service objects to consolidate weights.  We could start the conversation of where we would like to move to over time.  It does look like at some point we do want to know the reweigh location and time because the customer can go attend.

## Setup

To view the database diagram and comments in schemaspy:
```sh
make schemaspy
open tmp/schemaspy/index.html
```

To test the validators run:
```sh
go test ./pkg/models -testify.m TestReweighValidation
go test ./pkg/models -testify.m TestMTOShipmentValidation
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8941) for this change

## Screenshots

Screen Shot 2021-08-04 at 5 17 11 PM
Screen Shot 2021-08-06 at 11 42 01 AM
Screen Shot 2021-08-04 at 5 19 01 PM


migrations/app/schema/20210802205220_create_reweighs_table_excess_weights_columns.up.sql
```diff
@@ -0,0 +1,53 @@
+RESET ROLE;
```
duncan-truss, line 1: A previous migration set the Role to 'master', which is not able to alter existing tables and add columns.  This persisted between migrations so we need to reset it back to owner, which is 'postgres' I believe.

pkg/models/mto_shipments.go
```diff
@@ -97,6 +97,7 @@ type MTOShipment struct {
 	Diversion                        bool              `db:"diversion"`
 	RejectionReason                  *string           `db:"rejection_reason"`
 	Distance                         *unit.Miles       `db:"distance"`
+	Reweigh                          *Reweigh          `has_one:"reweighs"`
```
jacquelineIO, line 100: What's the relationship between `Reweigh` and `PrimeActualWeight`? I'm going to continue this question in Slack.

migrations/app/schema/20210802205220_create_reweighs_table_excess_weights_columns.up.sql
duncan-truss: Carter addressed this in a subsequent migration so I was able to remove this.

migrations/app/schema/20210802205220_create_reweighs_table_excess_weights_columns.up.sql
```diff
@@ -0,0 +1,51 @@
...
+    shipment_id uuid NOT NULL
+        CONSTRAINT reweighs_shipment_id_fkey
+            REFERENCES mto_shipments,
+    requested_at timestamp with time zone NOT NULL,
+    requested_by reweigh_requester NOT NULL,
+    weight integer,
+    verification_reason text,
+    verification_provided_at timestamp with time zone,
+    created_at timestamp NOT NULL,
+    updated_at timestamp NOT NULL
```
monfresh, line 20: Should these two also have `with time zone`?

migrations/app/schema/20210802205220_create_reweighs_table_excess_weights_columns.up.sql
```diff
@@ -0,0 +1,51 @@
...
+);
+
+CREATE INDEX reweighs_shipment_id_idx ON reweighs (shipment_id);
+
+COMMENT ON TABLE reweighs IS 'A reweigh represents a request from different users or the system for a shipment to be reweighed by the movers';
+COMMENT ON COLUMN reweighs.shipment_id IS 'A foreign key that points to the mto_shipments table for which shipment is being reweighed. There should only be one reweigh request per shipment';
+COMMENT ON COLUMN reweighs.requested_at IS 'The date and time when the reweigh request was initiated';
+COMMENT ON COLUMN reweighs.requested_by IS 'The type of user who requested the reweigh, including automated requests determined by the milmove system';
+COMMENT ON COLUMN reweighs.weight IS 'The reweighed weight in pounds (lbs) of the shipment submitted by the movers';
+COMMENT ON COLUMN reweighs.verification_reason IS 'If a reweigh was requested but was not able to be performed the movers can provide an explanation';
```
monfresh, line 30: Based on the description of this column, I'm wondering if there is a better name than `verification_reason`. The description sounds more like the request was denied or that the reweigh will not be performed for some reason. Would something like `prime_rejection_reason` be better?

monfresh: Or maybe `prime_justification` like you mentioned in the PR description?

migrations/app/schema/20210802205220_create_reweighs_table_excess_weights_columns.up.sql
duncan-truss: I did investigate this and I don't think it's encouraged to use `with time zone` because pop assumes they are of type `timestamp` (without time zone) in their own fizz language https://github.com/gobuffalo/pop/issues/23 https://github.com/gobuffalo/fizz/blob/3416f0e6618fcf1c0f1c7b45892c6bbb5d8eb65d/columns.go#L26-L27.  We definitely do it other places though (mto_shipments, mto_service_items, ...), I know our database env has been generally set to TZ=UTC so it hasn't hurt us.

migrations/app/schema/20210802205220_create_reweighs_table_excess_weights_columns.up.sql
duncan-truss: Yea I can see how that is less than clear, let me go back and find where I pulled it from and double check what language we're using.

pkg/models/mto_shipments.go
```diff
@@ -92,11 +92,14 @@ type MTOShipment struct {
 	PrimeEstimatedWeight             *unit.Pound       `db:"prime_estimated_weight"`
 	PrimeEstimatedWeightRecordedDate *time.Time        `db:"prime_estimated_weight_recorded_date"`
 	PrimeActualWeight                *unit.Pound       `db:"prime_actual_weight"`
+	BillableWeight                   *unit.Pound       `db:"billable_weight"`
+	BillableWeightJustification      *string           `db:"billable_weight_justification"`
```
gidjin, line 96: Currently when pricing we calculate the billableWeight. I think the assumption is that the look up in `pkg/payment_request/service_param_value_lookups/weight_billed_actual_lookup.go` will need to change to look at this field instead. Which I believe will be part of https://dp3.atlassian.net/browse/MB-3034

@jacquelineIO is this your understanding?

@duncan-truss I assume that BillableWeight is a field that the Prime will not be able to adjust correct? Just MilMove or the TIO?

donaldthai: IMO, I think since `BillableWeight` is meant to be a cap for the shipment, it should really be something like `BillableWeightCap` or `ShipmentBillableWeightCap so it'll come off clearer.

migrations/app/schema/20210802205220_create_reweighs_table_excess_weights_columns.up.sql
duncan-truss: I see it called `proof` other places like in the Prime API planned story but don't remember where I initially saw it referred to as verification. @NamibiaTorres @rogeruiz Do you have any further opinions/insight on this name?

pkg/models/mto_shipments.go
duncan-truss: @gidjin Yes BillableWeight here is going to be updatable by the TIO but likely read only or just not returned to the Prime in the payload.

@donaldthai I can change it to BillableWeightCap if that's clearer to what's reflected in the comments.  I think it's already namespaced to shipment so not as worried about needing mtoShipment.shipmentBillableWeightCap

Ryan-Koch (approved):

PR #7119, opened 2021-08-03
MB-8929 Add Semantic API Endpoints ADR

rogeruiz:
## Description

This patch series adds an ADR for updating the design for the Prime API endpoints to a more semantic design based on actions being performed on what is considered a logical-singular resource. Please view the Jira ticket referenced below for any additional information.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8929) for this change


docs/adr/0064-designing-semantic-api-endpoints.md
```diff
@@ -0,0 +1,135 @@
...
+    that are being modified.
+- ➖ Logical semantic actions create scenarios that give the Prime method
+    handlers a lot of different edge cases that would otherwise be better
+    handled by specific steps when modifying shipments.
+
+## Pros and cons of the alternatives
+
+### _Do nothing and document_
+
+- ➕ Less engineering work is needed from Roci engineers.
```
eamahanna, line 130: I think another couple of pros are: 
* The prime API remains flexible and does assume what the prime will want
* The REST APIs are easier to maintain

eamahanna: * Leaving the API as is, allows the API to follow standard milmove patterns, which reduces the amount of on boarding and confusion for new team members.

docs/adr/0064-designing-semantic-api-endpoints.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Considered Alternatives
+
+- *Do nothing to the Prime and document these edge cases of logical semantic
+  actions on a resource*.
+- ✅ *Create new semantic methods for single actions on resources*
+
+## Decision Outcome
+
+### Chosen Alternative: *Create new semantic methods for single actions on resources*
```
eamahanna, line 88: To be honest, I disagree with adding semantic endpoints until we have discussions with the prime to determine what they need. I think until that point, we should leave our APIs as flexible as possible because it is always easier to add functionality if needed than to remove it.

I am interested in other people's thoughts as well as product's and design's thoughts. @crwallace @pmohal

docs/adr/0064-designing-semantic-api-endpoints.md
```diff
@@ -0,0 +1,135 @@
...
+    ensuring a new shipment ID is not created if the method fails to update a
+    shipment ID with a `diversion` property)
+- ➖ Logical semantic actions create methods which may have side effects that
+    are not apparent to the user due to obfuscation of the underlying resources
+    that are being modified.
+- ➖ Logical semantic actions create scenarios that give the Prime method
+    handlers a lot of different edge cases that would otherwise be better
+    handled by specific steps when modifying shipments.
+
+## Pros and cons of the alternatives
```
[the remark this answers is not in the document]
rogeruiz: I only considered doing nothing and adding new endpoints that perform these new actions. But we could add others, I couldn't think of others based on @sandy-wright Slack explanation around the ADR. https://ustcdp3.slack.com/archives/CP497TGAU/p1628027092104800?thread_ts=1628022467.095500&channel=CP497TGAU&message_ts=1628027092.104800

docs/adr/0064-designing-semantic-api-endpoints.md
rogeruiz: I agree. I'll add these in. Thanks @eamahanna

docs/adr/0064-designing-semantic-api-endpoints.md
rogeruiz: I may have not been clear, but I don't think we'll be removing API endpoints but rather adding some new ones that do a lot more under-the-hood than the other APIs we currently have in place. I can refactor the background and proposal sections here to be more succinct. 

I am in agreement with your sentiment around adding functionality is better than removing functionality. I don't want this ADR to remove any functionality. I definitely misspoke to this earlier in Slack when I didn't understand what the ADR was proposing.

docs/adr/0064-designing-semantic-api-endpoints.md
```diff
@@ -0,0 +1,135 @@
...
+- ➕ Ability for the Prime engineers to perform extra actions without having to
+    modify actions if the underlying details of an action change over time.
+- ➖ Method handlers for the Prime would need to handle specific edge cases for
+    the user in case of a failure during a logical semantic action (e.g.
+    ensuring a new shipment ID is not created if the method fails to update a
+    shipment ID with a `diversion` property)
+- ➖ Logical semantic actions create methods which may have side effects that
+    are not apparent to the user due to obfuscation of the underlying resources
+    that are being modified.
+- ➖ Logical semantic actions create scenarios that give the Prime method
```
rogeruiz, line 122: To me this con is a big deal because it puts a lot of the responsibility on the engineering team to get this correct and could result in a lot of issues because "semantic actions" have a lot of side effects across different resources versus our current RESTful actions which require a lot more steps from the user of the API.

rogeruiz (commented):
Leaving a comment on the con on how this could affect supporting "semantic actions".

docs/adr/0064-designing-semantic-api-endpoints.md
ahobson: I thought I remember @monfresh proposing that we should move towards semantic action endpoints for other parts of our app. Is that right?

docs/adr/0064-designing-semantic-api-endpoints.md
```diff
@@ -0,0 +1,140 @@
...
+- ➕ Better logical relationship with a given action and resource from the
+    perspective of the Prime user (e.g. a given action performed on a given
+    resource without exposing details to the Prime user).
+- ➕ Ability for the Prime engineers to perform extra actions without having to
+    modify actions if the underlying details of an action change over time.
+- ➖ Method handlers for the Prime would need to handle specific edge cases for
+    the user in case of a failure during a logical semantic action (e.g.
+    ensuring a new shipment ID is not created if the method fails to update a
+    shipment ID with a `diversion` property)
+- ➖ Logical semantic actions create methods which may have side effects that
```
monfresh, line 119: I would consider this a plus if those side effects are required. Sometimes what happens behind the scenes is not relevant to the user. And if it is, documenting the side effects lets the user know that everything they expect to happen does with this single endpoint.

docs/adr/0064-designing-semantic-api-endpoints.md
monfresh: Yes, that's right. This is beneficial for all the APIs that use the pattern where we reuse the same handler and/or service object for various scenarios. Just because it's been the "standard" so far doesn't mean we should keep using it.

For example, previously in the GHC API, all shipment actions that affected a shipment's status used the same `PatchMTOShipmentStatus` handler, and required the Status to be passed in to the endpoint. Since the same handler was reused for all the scenarios, the `UpdateMTOShipmentStatus` service function started growing in complexity and we had to add hard-to-read conditionals to deal with differences between the scenarios.

This is not the way to design an app or an API. Instead of naming an endpoint based on what it's doing at the database level, it should represent the business concept that the consumer of the API is familiar with. If I want to approve a shipment, being able to make a request to an endpoint like `/shipments/{shipmentID}/approve` is a lot clearer than `/shipments/{shipmentID}/status` and having to remember or check the docs for which Status value I should be passing.

As a user, I don't care how the app represents an approved shipment. I just care that when I call the `approve` endpoint, the shipment gets approved and that it can move on to its next stage in the move process.

Also, by having endpoints where users can pass in values is riskier than having the logic controlled by the app only. For example, with a generic endpoint for updating a shipment's status, it's easier for bugs to appear where the user can pass in a status they shouldn't be allowed to, and it requires more guards in the app to prevent that.

If there are required side effects that always happen when performing an action, I would want the app to handle that for me instead of having to remember to do that myself via multiple requests to different endpoints. For example, when a shipment is approved, it sometimes requires setting the required delivery date and creating shipment-level service items. That should not be the responsibility of the API user to know when to perform those side effects.

Using endpoints with verb names that represent the specific action is good API practice and should not affect the Prime's ability to perform their duties. Using the diversion example at the top of this doc, those steps are all required in order to create a diversion. I could be wrong, but I don't think any one of those steps can be omitted. What is the advantage of making the Prime make several requests when they could use one endpoint that performs all the required steps for them?

monfresh: Having one endpoint for the diversion example also reduces risk of incorrect data in the DB. If the Prime has to manually create the new shipment, there's a risk that they set the wrong pickup address. If all the steps can be automated inside the app, it removes much if not all of the risk of incorrect data coming from the Prime.

rogeruiz: Responding to the shorter of the two messages @monfresh. I agree with that. Automated steps inside the app are better from a UX perspective. I'm a bit worried about having a `cancelMTODivergentShipment` or simply a `cancelMTOShipment` endpoint that has to do a lot more work to decouple the two shipments.

If we don't use that type of action-based endpoint to do all those automated steps (decoupling divergent shipments back to a single shipment) within the app, there's a lot less chance of error because the steps would be:

1. Delete the divergent shipment
2. Update the original (remaining) shipment with `{ "diversion": false }` and update (`PATCH`) the address (destination/pickup).

docs/adr/0064-designing-semantic-api-endpoints.md
rogeruiz: hmm, [see my comment above](https://github.com/transcom/mymove/pull/7119#discussion_r686334264). 👀 but I don't see this as a plus from the Prime API engineers perspectives. We'll have to create a number of tests for different scenarios when there are method handlers that do a lot of different things. I'm not saying this is 100% a bad thing, just want to be clear that it's a con in relation to developer (Roci) happiness for the Prime API.

rogeruiz: [This is the probably the more relevant comment](https://github.com/transcom/mymove/pull/7119#discussion_r685345955) 👀 when it comes to what I'm trying to express. It might make sense to change this into a ➕ given what you've said @monfresh

docs/adr/0064-designing-semantic-api-endpoints.md
rogeruiz: > As a user, I don't care how the app represents an approved shipment. I just care that when I call the approve endpoint, the shipment gets approved and that it can move on to its next stage in the move process.

I 100% agree here. I like this.

> What is the advantage of making the Prime make several requests when they could use one endpoint that performs all the required steps for them?

I agree with you about the advantages for the Prime lie with having action-based endpoints. The only current advantage for having the Prime do multiple API calls is that we would not have to update the API and rather just add some documentation around this to the Prime.

ahobson: IMHO, the fact that a generic REST API isn't self documenting is a con. For example, with a single generic endpoint, certain operations require certain fields to be set, but we can't express that with swagger. We can get much closer to that with semantic endpoints.

rogeruiz:
Oh boy! So I got a bit more context here and I doubt that semantic actions makes sense for the Prime API. I will update things a bit more to explain myself. thanks @shimonacarvalho!

sandy-wright:
This is a great discussion, and a super complicated issue. I proposed this ADR, but even I am torn about it.

I would suggest the following decision instead:
**Maintain the Prime API as a REST API, but add the RPC-ish endpoints to the Support API (or the Prime API Client CLI) to help our developers understand and test the Prime API**

I do think there's a lot of benefit in having these endpoints. Part of the point of MilMove is asserting a more consistent, regulated move process for our service members. Having well-defined and restrictive actions as our endpoints helps with this end goal.

However, it's true that we don't know how the Prime will be using the API. It's entirely possible that they'll have their own system handling things their way. I don't think this should stop us from making the development process easier on our engineers by adding these endpoints, but perhaps we shouldn't bake it into the Prime API just yet. Hence, I propose that we edit the decision to move the RPC actions somewhere else.

rogeruiz:
> From Roci's sprint planning meeting.

We're going to be shelving this ADR for now and close this. An outcome of this is that we'll be documenting these sort of multi-pronged actions better so that folks can do these actions with the multiple steps that are required.

PR #7135, opened 2021-08-05
[MB-4699] Show weight summaries on MTO page

hannaliebl:
## Description

Show the summary of weights for a move on the MTO page.

## Reviewer Notes

Just double check that the values I'm showing in the boxes matches the actual data I'm passing in...I think I got it, but another set of eyes is always good. I also think the way I set up the reduce means that null values are handled ok.

## Setup

make server_run, make client_run, run through a few cases, ideally.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4699) for this change

## Screenshots

Screen Shot 2021-08-05 at 1 38 31 PM


src/components/Office/WeightDisplay/WeightDisplay.module.scss
```diff
@@ -6,6 +6,11 @@
   @include u-bg('white');
   @include u-radius(3px);
   border: 1px solid $border-color;
+  flex: 1 1 0px;
```
duncan-truss, line 9: I think this style should live in the MoveTaskOrder.module.scss that way if this component gets used elsewhere across the app there isn't an assumption of how they will be positioned in the flexbox.  You won't be able to use the .WeightDisplay class because it will be getting a classnames identifier put on it but you could add a static one that will let you select it.

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -285,6 +286,14 @@ export const MoveTaskOrder = ({ match, ...props }) => {
...
   }
 
+  const moveWeightTotal = mtoShipments?.reduce((prev, current) => {
+    return prev + current.primeActualWeight;
+  }, 0);
+
+  const estimatedWeightTotal = mtoShipments?.reduce((prev, current) => {
+    return prev + current.primeEstimatedWeight;
+  }, 0);
+
```
duncan-truss, line 296: Do you get a NaN sum here if the move has multiple shipments with one missing estimated or actual weight?

src/pages/Office/TXOMoveInfo/TXOTab.module.scss
```diff
@@ -69,6 +69,12 @@
   }
 }
 
+.weightHeader {
```
duncan-truss, line 72: I would go ahead and start a new styles file for MoveTaskOrder if it doesn't have one.  I'm not sure this style will be shared amongst the other TXO pages will it?

src/components/Office/WeightDisplay/WeightDisplay.module.scss
jenniferlynparsons: I think this change would be fine. I'm not sure if we're actually going to use the component anywhere else styled with the border and the rest of it, but agree that adding the flex here makes it less... flexible. 😄

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
jenniferlynparsons: It would probably be useful to write a test case around this to confirm.

src/pages/Office/TXOMoveInfo/TXOTab.module.scss
jenniferlynparsons: +1 It's specific enough to warrant moving into a MoveTaskOrder file, especially if the flex styling is moved as well.

jenniferlynparsons (commented):
Looking good. There's a few suggested changes and another test or two around the reducers would be helpful.

src/components/Office/WeightDisplay/WeightDisplay.module.scss
hannaliebl: I think the best way to do this is to pass in/add a className prop to the weight display component, and I can use a USWDS utility for the class itself, I think. Or do you mean I should just add a static class to the weight component but then style it in the TXO styles?

hannaliebl: Hmm actually I might just do this with a child combinator selector on the parent...I will push it up and see what you all think :D

duncan-truss: Yep I think that's an equally good alternative!

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
hannaliebl: I added a bunch of tests. One thing is that the way I'm doing it now is that if all the weights are not defined or null, it still returns 0. I'm not sure this is ideal, or if this will ever happen, though?

hannaliebl:
This is ready for review - I just pushed a commit to show the emdash rather than 0 when the weights are not set or not sent. The tests are updated with this as well.

jenniferlynparsons (approved):
LGTM!

brooksr4 (approved):

PR #3499, opened 2020-02-08
MB-1139 TOO approve reject shipments

donaldthai:
## Description

Given that I am TOO viewing a shipment requested by the Prime

When I click the Approve button (see other story) 

The shipment is approved (temporary indicatory works) 

When I click the Reject button 

Then I should see a modal 

And it should have a text box with the following instructions: 

“Enter a reason for rejection.”

And the TOO is not able to reject until they enter a reason for rejection. 

And the TOO has the option to cancel the rejection. 

And the TOO can click a reject button once they’ve entered a reason for rejecting. 

Notes:

There are not designs for this. We are still in skeleton mode. 

The rejection modal above should be built as a component. It will be reused for service items and payment request rejections. 

## Setup

Add any steps or code to run in this section to help others prepare to run your code:

```sh
make db_dev_e2e_populate
make server_run
make office_client_run
```

Once you set up everything you should be able to just follow this link after logging in: http://officelocal:3000/too/customer-moves/6fca843a-a87e-4752-b454-0fac67aa4988/customer/6ac40a00-e762-4f5f-b08d-3ea72a8e4b63

Then you can approve/reject a shipment.

## Code Review Verification Steps

* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/tree/master/docs/backend.md#logging)
* [ ] The requirements listed in
 [Querying the Database Safely](https://github.com/transcom/mymove/tree/master/docs/backend.md#querying-the-database-safely)
 have been satisfied.
* Any new migrations/schema changes:
  * [ ] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://github.com/transcom/mymove/tree/master/docs/database.md#zero-downtime-migrations))
  * [ ] Have been communicated to #dp3-engineering
  * [ ] Secure migrations have been tested using `scripts/run-prod-migrations`
* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-1139) for this change

## Screenshots

approve reject shipment

PR #7138, opened 2021-08-05
MB-4704 Add endpoint to request reweigh

monfresh:
## Description

This PR adds a GHC API endpoint to allow a TOO to request a reweigh on a shipment.
Some assumptions I made for the validations:
- We should only allow a reweigh request if the shipment is either Approved or Diversion Requested.
- A reweigh request is denied if a reweigh already exists for this shipment
- Since this endpoint creates a reweigh, and we are not modifying any shipment fields, I assumed we don't need to pass in and check the eTag.
- The handler should return a Reweigh in the payload. I also included the Shipment in the Reweigh's payload. I'm not sure if we need that.

## Reviewer Notes

The DB and model changes are pulled in from #7118 since this PR depends on those tables existing. 

## Setup

```sh
echo "Code goes here"
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4704) for this change


pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -683,3 +683,21 @@ func QueuePaymentRequests(paymentRequests *models.PaymentRequests) *ghcmessages.
 
 	return &queuePaymentRequests
 }
+
+// Reweigh payload
+func Reweigh(reweigh *models.Reweigh) *ghcmessages.Reweigh {
+	strfmt.MarshalFormat = strfmt.RFC3339Micro
```
duncan-truss, line 689: This doesn't feel safe to me to be mutating the strfmt MarshalFormat, why do we need this exactly?

swagger/ghc.yaml
```diff
@@ -3233,6 +3269,44 @@ definitions:
...
+        format: date-time
+        type: string
+      requestedBy:
+        $ref: '#/definitions/ReweighRequester'
+      verificationProvidedAt:
+        x-nullable: true
+        format: date-time
+        type: string
+      verificationReason:
+        example: "to double check if the customer does indeed have excess weight"
```
duncan-truss, line 3288: This may go back to your comment about name of this field, but I think it's providing the reason why the reweigh was not conducted.

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
monfresh: Oops. That was copy pasta from another payload.

monfresh: I'm not sure why we're doing that in other places.

swagger/ghc.yaml
monfresh: Yeah, I hadn't seen your DB comments, and so I assumed that's what that field was for. I'll update it.

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
monfresh: It looks like we're doing it in the MTOShipment definition, and I would guess perhaps for the CreatedAt and UpdatedAt fields.

pkg/handlers/ghcapi/mto_shipment.go
```diff
@@ -637,3 +637,66 @@ func (h RequestShipmentCancellationHandler) triggerRequestShipmentCancellationEv
...
+	if !session.IsOfficeUser() || !session.Roles.HasRole(roles.RoleTypeTOO) {
+		logger.Error("Only TOO role can Request a shipment reweigh")
+		return shipmentops.NewRequestShipmentReweighForbidden()
+	}
+
+	shipmentID := uuid.FromStringOrNil(params.ShipmentID.String())
+	reweigh, err := h.RequestShipmentReweigh(params.HTTPRequest.Context(), shipmentID)
+
+	if err != nil {
+		logger.Error("RequestShipment error: ", zap.Error(err))
```
duncan-truss, line 660: The error is going to be added to the json under it's own key `msg: "RequestShipment error: ", error: "Shipment Not Found"` not sure if you wanted to string format it into the message or leave that in a separate field.

pkg/handlers/ghcapi/mto_shipment.go
```diff
@@ -637,3 +637,66 @@ func (h RequestShipmentCancellationHandler) triggerRequestShipmentCancellationEv
...
+
+		switch e := err.(type) {
+		case services.NotFoundError:
+			return shipmentops.NewRequestShipmentReweighNotFound()
+		case services.InvalidInputError:
+			payload := payloadForValidationError("Validation errors", "RequestShipmentReweigh", h.GetTraceID(), e.ValidationErrors)
+			return shipmentops.NewRequestShipmentReweighUnprocessableEntity().WithPayload(payload)
+		case services.PreconditionFailedError:
```
duncan-truss, line 668: I think you chose not to pass in an eTag for the shipment so we won't be getting this error is my understanding.

pkg/handlers/ghcapi/mto_shipment.go
```diff
@@ -637,3 +637,66 @@ func (h RequestShipmentCancellationHandler) triggerRequestShipmentCancellationEv
...
+			return shipmentops.NewRequestShipmentReweighPreconditionFailed().WithPayload(&ghcmessages.Error{Message: handlers.FmtString(err.Error())})
+		case mtoshipment.ConflictStatusError:
+			return shipmentops.NewRequestShipmentReweighConflict().WithPayload(&ghcmessages.Error{Message: handlers.FmtString(err.Error())})
+		default:
+			return shipmentops.NewRequestShipmentReweighInternalServerError()
+		}
+	}
+
+	moveID := reweigh.Shipment.MoveTaskOrderID
+	h.triggerRequestShipmentReweighEvent(shipmentID, moveID, params)
```
duncan-truss, line 678: Good job remembering this.

pkg/services/mto_shipment/shipment_reweigh_requester.go
```diff
@@ -0,0 +1,82 @@
...
+	if verrs != nil && verrs.HasAny() {
+		invalidInputError := services.NewInvalidInputError(shipment.ID, nil, verrs, "Could not save the reweigh while requesting the reweigh as a TOO.")
+
+		return nil, invalidInputError
+	}
+	if dbErr != nil {
+		return nil, dbErr
+	}
+
+	err := f.db.Load(&reweigh, "Shipment")
```
duncan-truss, line 76: I don't think anything will have changed on the shipment, could we just set it directly on the reweigh model from the local param?

pkg/services/mto_shipment/shipment_reweigh_requester.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (f *shipmentReweighRequester) createReweigh(ctx context.Context, shipment *models.MTOShipment, checks ...validator) (*models.Reweigh, error) {
+	if verr := validateShipment(ctx, shipment, shipment, checks...); verr != nil {
+		return nil, verr
+	}
+
+	now := time.Now()
+	reweigh := models.Reweigh{
+		RequestedBy: models.ReweighRequesterTOO,
```
duncan-truss, line 61: We'll likely need to let the handler specify this value eventually but I'm okay if it stays for now until the Prime work gets done.

pkg/services/mto_shipment/shipment_reweigh_requester.go
monfresh: The handler needs the Move ID for the event notification, and in order for the handler to be able to read the shipment's MoveTaskOrderID, the Shipment has to be loaded. I'm open to other solutions.

duncan-truss: Yea that's what led me here, I assumed in your findShipment method you loaded the shipment model that you pass to this function so I assumed you had access to the shipment.move_id.

pkg/handlers/ghcapi/mto_shipment_test.go
```diff
@@ -1325,6 +1325,165 @@ func (suite *HandlerSuite) TestRequestShipmentCancellationHandler() {
...
+	})
+
+	suite.Run("Returns a 403 when the office user is not a TOO", func() {
+		officeUser := testdatagen.MakeServicesCounselorOfficeUser(suite.DB(), testdatagen.Assertions{Stub: true})
+		uuid := uuid.Must(uuid.NewV4())
+		reweighRequester := &mocks.ShipmentReweighRequester{}
+
+		reweighRequester.AssertNumberOfCalls(suite.T(), "RequestShipmentReweigh", 0)
+
+		req := httptest.NewRequest("POST", fmt.Sprintf("/shipments/%s/request-cancellation", uuid.String()), nil)
```
duncan-truss, line 1375: Some of these are named `request-cancellation` instead of `request-reweigh` but the right handler is being called.

pkg/services/mto_shipment/shipment_reweigh_requester.go
monfresh: Good point. For now, I was not trying to generalize too early. I was thinking of this specifically for the TOO.

duncan-truss (approved):
Approved! My db migration PR is approved but didn't get much feedback.  I think that needs to go in first, I did rename billable_weight to billable_weight_cap that you can update.  Not sure if I should hold off longer or not.

pkg/handlers/ghcapi/mto_shipment.go
monfresh: Good catch!

pkg/handlers/ghcapi/mto_shipment.go
monfresh: Argh! More copy paste errors, and I noticed some more in the other handlers.

monfresh: It looks like we're following this same pattern in this entire file, but we've been specifying the API and handler in the `msg`, like `ghcapi.UpdateShipmentHandler`. Sounds like we should standardize the way we log errors in handlers. For now, I'll follow this pattern.

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
monfresh: This is the only place in the entire app where we are doing this. Here is the commit that introduced this, which didn't use `strfmt`, so maybe it was added by mistake. I'm gonna remove it.

monfresh: https://github.com/transcom/mymove/pull/3499/commits/e8062a158d9b7e76d26f5a7e0a40dce6d606000c

pkg/services/mto_shipment/shipment_reweigh_requester.go
monfresh: Oh, I see what you mean! Sorry, it's been a long week! `reweigh.Shipment = *shipment` works, and saves us a DB query. Thanks for noticing this!

monfresh:
Thanks for the thorough review! I think I addressed everything.

Ryan-Koch (approved):

PR #7139, opened 2021-08-05
[MB-9035] Remove readOnly from PaymentRequestID

mboren:
## Description

The latest version of go-swagger prevents us from sending `readOnly` parameters in the body of a request.
Since we need to send `paymentRequestID` in the body of requests to `/payment-requests/process-reviewed`, it can no longer be `readOnly`.
## Setup
Make a request to this endpoint:
`PATCH https://primelocal:9443/support/v1/payment-requests/process-reviewed`
with this body:
```json
{
    "sendToSyncada": false,
    "readFromSyncada": false,
    "deleteFromSyncada": false,
    "paymentRequestID": "46ffbf9c-153b-48b5-b2b1-b7c99853f4bd",
    "status": "REVIEWED"
}
```
The actual `paymentRequestID value doesn't matter. If you like, you can send the ID of a real payment request, and then you should get a success and not an error, but that's not technically necessary because the swagger validation happens earlier.

You should see:
```json
{
    "detail": "Error finding Payment Request with ID: 46ffbf9c-153b-48b5-b2b1-b7c99853f4bd",
    "instance": "11f1f76b-f6c9-49fc-acc8-5d61ee9fabdb",
    "title": "Not Found Error"
}
```
If you see instead:
```json
{
    "title": "Validation Error",
    "instance": "4e901c86-b5cd-4d18-bd88-0f7050906668",
    "detail": "paymentRequestID in body is readOnly"
}
```
Then this fix did not work.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9035) for this change

### DIFF
--- a/pkg/gen/supportapi/embedded_spec.go
+++ b/pkg/gen/supportapi/embedded_spec.go
@@ -2096,7 +2096,6 @@ func init() {
         "paymentRequestID": {
           "type": "string",
           "format": "uuid",
-          "readOnly": true,
           "example": "1f2270c7-7166-40ae-981e-b200ebdf3054"
         },
         "readFromSyncada": {
@@ -4811,7 +4810,6 @@ func init() {
         "paymentRequestID": {
           "type": "string",
           "format": "uuid",
-          "readOnly": true,
           "example": "1f2270c7-7166-40ae-981e-b200ebdf3054"
         },
         "readFromSyncada": {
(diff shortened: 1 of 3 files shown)


reggieriser (approved):
LGTM!  Was able to get it to work locally fine (no longer blocking due to a readonly error).

PR #7143, opened 2021-08-06
MB-9050: Deploy the TLS app to demo

ahobson:
* This fixes MB-9050

## Description

We need to deploy multiple applications, including the TLS app used by the API endpoint.

I'm not entirely sure this is right. In particular, I'm not sure what to use for the `health_check_hosts`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1446,6 +1446,18 @@ jobs:
           health_check_hosts: my.demo.dp3.us,office.demo.dp3.us,admin.demo.dp3.us
           ecr_env: demo
 
+  # `deploy_demo_app_client_tls` updates the mutual-TLS service in the demo environment
+  deploy_demo_app_client_tls:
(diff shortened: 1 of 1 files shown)


sheenamt (approved):
I'm also not positive about which site to use for the health_check_hosts, but api.demo.dp3.us does exist so its worth trying

.circleci/config.yml
```diff
@@ -1446,6 +1446,18 @@ jobs:
           health_check_hosts: my.demo.dp3.us,office.demo.dp3.us,admin.demo.dp3.us
           ecr_env: demo
 
+  # `deploy_demo_app_client_tls` updates the mutual-TLS service in the exp environment
```
reggieriser, line 1449: Nit: Comment still mentions `exp`.

ahobson: Thank you, I fixed this

PR #7145, opened 2021-08-06
`openapi` bundling proof-of-concept

sandy-wright:
## Description

This is a proof-of-concept PR for the [ADR to switch to using `openapi`](https://github.com/transcom/mymove/pull/7144) to compile our API specifications. To make the switch, I had to:

- Create a new directory `swagger-def/` for our split API files. The compiled files go in `swagger/` as usual so that we don't have to change our integrations or `go-swagger` settings.
- Change some linters. For example, as we move over, we can't have the alphabetical definitions rule be an error anymore. All compiler-added definitions go at the end, so gradually moving things over could cause these errors.
- Update the Makefile commands for generating code. I had to add a `swagger_generate` step. Note that this will **not** warn you about file sync issues the first time it runs so that tests don't fail. It will from then on.
- Write some README files to explain how to add new code to the `swagger-def/` directory.

I also updated the styling for the generated Redocs for the Prime and Support APIs. This was totally superfluous and I would've excluded it had I thought more about the PR process for this, but alas.

Note that the generated files are only different due to spacing/indentation from the compiler. Otherwise, it looks the same as what we're used to. Also, the compiled YAMLs have big diffs because of all the spacing changes - I couldn't figure out how to make these settings line up with what we're used to so that the changes weren't so large. But rest assured, they are otherwise the same as they were.

## Setup

To compile your API files for the first time, run:

```sh
make swagger_generate
```

This will create a stamp file in your repo (it's an ignored file) indicating the last time you ran this command. Once this file exists, it will check the `swagger/` folder for changes that have happened since this time and warn you that you might be overriding your work. 

Try typing something in `swagger/support.yaml` and run the command again. You should see:

```sh
Unexpected changes found in swagger build files. Code may be overwritten.
Continue with rebuild? [y/N] : 
```

Enter `n` and it should stop the build.

Next, try making some changes in `swagger-def/definitions/Upload`. This time, run:

```sh
make server_generate
```

This command now has `swagger_generate` as a prerequisite, so that build should also happen. Go ahead and enter `y` at the override prompt so the build executes. You should see that both `swagger/prime.yaml` AND `swagger/support.yaml` have changed. You should also see that some generated files have changes for both APIs.

For the next test, run your server!

```sh
make server_run
```

Instead of looking at the `swagger/` YAML files for changes, it should now be looking at `swagger-def/` to know when to trigger a rebuild. 

Once your server is up, try making a change to one of the YAML files in `swagger-def/`. You should see a full rebuild and then notice that the API you changed (and perhaps both) has been updated in `swagger/` as well. 

Now, once the server has rebuilt, try making a change to `swagger/support.yaml`. You shouldn't see anything happen - the server should not be triggered for a rebuild.

After that, make another change to one of the YAML files in `swagger-def/` again. If it rebuilt, it could potentially override and lose that change you made to `swagger/support.yaml`! Instead, you should see the same prompt that you saw before, and it should halt the entire rebuild. This time, it won't be able to take any input, so it just stalls the whole `server_run` process.

To make the override, you'll need to cancel and run either `swagger_generate` or `server_generate` so that you can input `y` to the prompt.

And that's the full dev cycle! Hopefully it feels fairly similar to what you had to do before.

## References

* [ADR](https://github.com/transcom/mymove/pull/7144) for this change.


swagger-def/.spectral.yml
```diff
@@ -131,9 +131,9 @@ rules:
 
 # DEFINITIONS
   definitions-sort:
-    description: Definitions MUST be sorted alphabetically
-    #message: "{{error}}"
-    severity: error
+    description: Definitions should be sorted alphabetically
+    #message: "{{warn}}"
+    severity: warn
```
gidjin, line 136: Curious as to why we relaxed this to warning?

.pre-commit-config.yaml
```diff
@@ -64,7 +64,7 @@ repos:
     rev: v0.27.1
     hooks:
       - id: markdownlint
-        entry: markdownlint --ignore .github/*.md
+        entry: markdownlint --ignore .github/*.md --ignore swagger-def
```
gidjin, line 67: I'm curious as to why we don't want to run markdownlint on markdown files in `swagger-def`

swagger-def/.spectral.yml
sandy-wright: Good question! When we compile the API spec, all definitions that were external (so in this example, `Upload`, `ValidationError`, etc) are added into the compiled spec at the end of the list. As we gradually move definitions out of the main spec and into their own files, we'll have less and less control about where they appear in the final file. 

So, with a compiler handling our organization for us, we don't have to be as strict about this requirement anymore. Plus external files will be automatically sorted alphabetically within the `definitions/` folder.

Makefile
```diff
@@ -304,8 +304,21 @@ pkg/assets/assets.go:
...
+check_swagger_generate: .swagger_build.stamp ## Check that the build files haven't been manually edited to prevent overwrites
+.swagger_build.stamp: $(shell find swagger -type f -name *.yaml)
+ifneq ("$(wildcard .swagger_build.stamp)","")
+	@echo "Unexpected changes found in swagger build files. Code may be overwritten."
+	@read -p "Continue with rebuild? [y/N] : " ANS && test "$${ANS}" == "y" || (echo "Exiting rebuild."; false)
+endif
+
+.PHONY: swagger_generate
+swagger_generate: check_swagger_generate .client_deps.stamp ## Bundles the API definition files into a complete specification
+	yarn build-redoc
```
gidjin, line 317: Does this trigger the compilation of the yaml files?

sandy-wright: yes, it does 👍

Shkeating (approved):
not 100% sure why design got tagged on this one, but I tried to follow the instructions and got through a lot of the tasks 😎

.pre-commit-config.yaml
sandy-wright: ahhh, I remembered the reason I added this: the markdown files in `swagger-def` will often be snippets. The description of an endpoint is kind of like a blurb within a larger document. This means it doesn't make sense to start each one with an <h1> element, and that's a blocking rule in `markdownlint`

sandy-wright: (but I can disable that rule inline)

ahobson: Could we add a comment to the config that explains why the rule is there?

sandy-wright: I removed this file from the rule and just ignored the one offending rule in the `markdownlintrc`. I felt like linting for an `h1` element as the very first line isn't a critical rule for us (but please let me know if you disagree!)

eamahanna (commented):
Is there a plan or guide on how to migrate existing definitions to follow this format?

rogeruiz:
> Is there a plan or guide on how to migrate existing definitions to follow this format?

@eamahanna I believe that this proof-of-concept PR will attempt to migrate all the current definitions, but I'm not entirely sure about things that are currently in-flight. I believe we discussed that we would need to more generally communicate or coordinate merge conflicts after this gets merged in. But I don't believe a guide was discussed. It's certainly a good idea to update any documentation related to modifying Swagger files to leverage this new organizational style.

sandy-wright:
> Is there a plan or guide on how to migrate existing definitions to follow this format?


> @eamahanna I believe that this proof-of-concept PR will attempt to migrate all the current definitions, but I'm not entirely sure about things that are currently in-flight. I believe we discussed that we would need to more generally communicate or coordinate merge conflicts after this gets merged in. But I don't believe a guide was discussed. It's certainly a good idea to update any documentation related to modifying Swagger files to leverage this new organizational style.


Good question - I didn't have a specific plan in place for this. This PR will move all APIs into the new folder so that everyone _can_ write definitions in the new style if they want. In my head, I was leaving it up to devs to migrate definitions on a case-by-case basis whenever they felt it was appropriate. I think better guidelines would be great though.

My thought process is that new definitions should be placed in the definitions folder and referenced in the API. And then old definitions can be migrated as it comes up? Does that sound sufficient?

eamahanna:
I think that makes sense @sandy-wright I just want to make sure devs know this change is coming. It would probably make sense to over share this in prac engineering and in an all together/demo.

ahobson:
It seems like we've struggled with the "migrate on demand" pattern because we don't actually get around to the migration. I wonder if we could find a way to add converting all the existing definition to the TRA backlog so we don't leave ourselves in a halfway state. Thoughts?

.markdownlintrc
```diff
@@ -6,5 +6,6 @@
     "siblings_only": true
   },
   "MD014": false,
+  "MD041": false,
```
sandy-wright, line 9: Rule **MD041** forces every single markdown file to start with an `#h1` header. This doesn't make sense for the snippets we'll be including in our API spec because they're often subsections of a part of a page, instead of their own page.

I could have disabled it inline, but I'd have to do that for every file in that directory essentially. I thought it would be fine to disable it overall because it's a fairly minor rule (imo)

.pre-commit-config.yaml
```diff
@@ -104,6 +104,7 @@ repos:
         language: script
         files: swagger/*
         types: [yaml]
+        exclude: swagger-def/*
```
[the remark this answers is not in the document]
sandy-wright: oh, yup! This script checks that the YAML files in the given directory are in a valid swagger format. The `swagger-def/*` files are partial swagger (since they have not yet been compiled) and they will fail this check.

Although it looks like only the files in `swagger/*` would have been checked anyway, it actually still includes the `swagger-def/*` files and will fail without the exclusion.

It is fine to exclude these files because they compile into the files being checked anyway - so in effect, they are still being validated. Furthermore, it is the files in `swagger/*` that are used to generate our Go code, so they're the important ones in that respect.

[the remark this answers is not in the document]
ahobson: Since this is a yaml file, couldn't we add a comment to the file about this so the explanation stays around?

jacquelineIO (approved):
**IS3 Review**: lgtm thank you for the explanations

sandy-wright:
> It seems like we've struggled with the "migrate on demand" pattern because we don't actually get around to the migration. I wonder if we could find a way to add converting all the existing definition to the TRA backlog so we don't leave ourselves in a halfway state. Thoughts?

I think one problem with doing that for the Open API defs in particular is that we may not want to move everything over to a separate definition file. There are going to be cases where something is specific to one API, and sometimes it might make sense to keep those definitions in that API's base file instead of breaking it out. 

We can probably make tickets for tasks along the lines of breaking out all status definitions (those should really always be the same), probably the Move definitions, the Shipment, Service Items, and Payment Requests. Update/Create-specific definitions should probably stay in their respective API files (at least, to start with). 

And things like the `If-Match` header can _definitely_ be shared, so that can be pulled out into a ticket. But I suppose I'm just trying to communicate that there is some nuance here, and there may not be a clear boundary around "migrate to the new pattern"

swagger-def/support.yaml
```diff
@@ -0,0 +1,1879 @@
...
+        format: uuid
+        in: path
+        name: paymentRequestID
+        required: true
+        type: string
+    get:
+      summary: getPaymentRequestEDI
+      description: |
+        Returns the EDI (Electronic Data Interchange) message for the payment request identified
+        by the given payment request ID. Note that the EDI returned in the JSON payload will have \n where there
```
felipe-lee, line 482: Small thing, but is there a reason for the `\n` in here?

swagger-def/tags/moveTaskOrder.yaml
```diff
@@ -0,0 +1,4 @@
+name: moveTaskOrder
+description: >
```
felipe-lee, line 2: Why do some descriptions start with `>` while others (next file, `swagger-def/tags/mtoShipment.yaml`) start with `|`?

swagger-def/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+* **Prime API**
+* **Support API** (non-production)
+
+Their respective entrypoints are:
+
+* `prime.yaml`
+* `support.yaml`
+
+Both of these YAML files contain references to the entire API definition.
```
felipe-lee, line 14: Now that you've ported over the other APIs, this should be updated to reflect that.

felipe-lee (approved):
LGTM, just a few questions and comments. Thanks for putting in all the effort you have on this!

swagger-def/support.yaml
sandy-wright: Oof, no idea! It actually shows up in the redoc, so it's truly superfluous. I think we might have a some issues like this in other descriptions too (and some of them use `<br />` tags, which I don't think we should be doing), so we'll want to revisit this later on. I will go ahead and delete this now though since... why not? Haha

swagger-def/tags/moveTaskOrder.yaml
sandy-wright: This is an OpenAPI thing! They are two different ways of communicating how the markdown should behave. `>` tells OpenAPI/ReDoc to ignore line breaks and some additional white space when rendering the markdown. `|` means that it should take all white space literally and include blank lines where indicated. 

So with `>` you'd have to do:
```yaml
description: >
  Paragraph one.


  Paragraph two.
```
to make a blank line show up. The difference is pretty subtle, but I think they're both viable options? At least I don't have a strong enough opinion to pick one over the other as best practices, but I'm definitely open to other opinions.

swagger-def/README.md
sandy-wright: yesss, nice catch

swagger-def/tags/moveTaskOrder.yaml
felipe-lee: Ah ok. I'm fine with both existing in our stuff, just hadn't known what they did. When I was looking for some reason I was only getting to different flavors of markdown and these both mean different things in those so I wasn't sure.

eamahanna (approved):

PR #7164, opened 2021-08-10
[MB-9081] Add GEX environment variables to TLS environments

reggieriser:
## Description

We were having a problem using our support endpoint to trigger the sending of invoices to GEX in our staging environment.  The job runner that picks up reviewed invoices every 5 minutes seemed to be working fine.  Digging into the logs, it appeared that we were getting an "unsupported protocol scheme" error from our support endpoint.  The URL it was trying to use only had the query string portion and no protocol/host/directory.  That led us to realize that the TLS environment (where support endpoints are reached) for staging did not have the `GEX_URL` set (the app environment and the job runner task did, though).  This PR attempts to fix that discrepancy.

Related Slack thread: https://ustcdp3.slack.com/archives/CP6F568DC/p1628526734214000

## Reviewer Notes

- I added both `GEX_SEND_PROD_INVOICE` and `GEX_URL` to the TLS environments for stg/exp/demo.  Technically, exp/demo are blocked from sending invoices right now, but I went ahead and added them to those environments for consistency since those variables were already in other `*.env` files for those environments.
- I also added the environment variables to the `*-niprnet.env` files too since those appear to be exact replicas of the current TLS configs.
- There were some task configs in exp/stg that did not have `GEX_SEND_PROD_INVOICE` set explicitly to `false`.  That's the default value, but we did explicitly include that environment variable in other configs (and seems like a good idea to reinforce that we're not sending prod invoices).
- I didn't touch anything related to prod.  The support API is not deployed there, so it's not possible to hit the job runner trigger endpoint there anyway.

## Setup

I'm not sure of a way to test this for real until it hits staging.  Staging is cleared to send test invoices to GEX, but others can't at the moment.  Let me know if there's another way to do it.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9081) for this change
* [Related Slack thread](https://ustcdp3.slack.com/archives/CP6F568DC/p1628526734214000)


gidjin (commented):
Changes look good. ~I noticed you didn't update the prd files, seems like a good time to do that too, to keep them in sync.~

gidjin (approved):
I read the PR reviewer notes after reading the changes. I agree with your analysis, so I'm marking this as approved.

travelar (approved):
Seems in line with what I would expect and I don't see any problems with this.

PR #7166, opened 2021-08-10
[MB-7692] CAT II: consistent-return finding

mboren:
## Description
In this PR, I turned on the [eslint `consistent-return` rule](https://eslint.org/docs/rules/consistent-return) for the directories `src/shared` and `src/scene`, and fixed all the errors that popped up.

There are a lot of cases where javascript will implicitly return `undefined` from a function, and this rule forces us to handle those cases more explicitly, to prevent bugs from `undefined` values popping up in places where we don't expect them.

In most cases that eslint flagged, our code was clearly written with the knowledge that `undefined` would be returned, so we can fix the errors reported by the linter without changing behavior by explicitly returning `undefined`.

The only cases where I did more than that were `if` statements that I rearranged to get rid of an implicit return.

## Reviewer Notes

- There are a few cases where we hit a problem or end up in an unexpected state, and use `console.error` and return `undefined`. I think for the purposes of this PR, it's OK to leave these as they are, but we're going to have to come back to them soon when we remove `console` statements for MB-7690.
I don't know how we should handle these without a system like Sentry or OpenTelemetry or something to send errors to. I know that stuff is in the works, but I'm not sure what we should do before we have it.

- I ran into some dependency struggles that I don't really understand. Yarn told me to install some things so I did. Please let me know if I've done this wrong.

## Setup

Make sure that the linter runs without displaying any errors:
```sh
yarn run lint
```

## Acceptance
No acceptance for this.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7692) for this change
* [eslint docs for this error](https://eslint.org/docs/rules/consistent-return)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
     "@fortawesome/free-regular-svg-icons": "^5.15.4",
     "@fortawesome/free-solid-svg-icons": "^5.15.4",
     "@fortawesome/react-fontawesome": "^0.1.15",
+    "@material-ui/core": "^4.12.3",
     "@trussworks/react-file-viewer": "https://github.com/trussworks/react-file-viewer",
     "@trussworks/react-uswds": "^2.0.0",
     "bytes": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
     "@fortawesome/free-solid-svg-icons": "^5.15.4",
     "@fortawesome/react-fontawesome": "^0.1.15",
     "@material-ui/core": "^4.12.3",
+    "@material-ui/icons": "^4.11.2",
     "@trussworks/react-file-viewer": "https://github.com/trussworks/react-file-viewer",
     "@trussworks/react-uswds": "^2.0.0",
     "bytes": "^3.0.0",
--- a/src/scenes/Moves/Ppm/Weight.jsx
+++ b/src/scenes/Moves/Ppm/Weight.jsx
@@ -141,9 +141,7 @@ export class PpmWeight extends Component {
     if (currentEstimate >= 500 && currentEstimate < 1500) {
       return <img className="icon" src={trailerGray} alt="trailer-gray" data-testid="vehicleIcon" />;
     }
-    if (currentEstimate >= 1500) {
-      return <img className="icon" src={truckGray} alt="truck-gray" data-testid="vehicleIcon" />;
-    }
+    return <img className="icon" src={truckGray} alt="truck-gray" data-testid="vehicleIcon" />;
   }
 
   chooseEstimateText(currentEstimate) {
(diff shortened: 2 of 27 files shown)


package.json
```diff
@@ -4,6 +4,8 @@
     "@fortawesome/free-regular-svg-icons": "^5.15.4",
     "@fortawesome/free-solid-svg-icons": "^5.15.4",
     "@fortawesome/react-fontawesome": "^0.1.15",
+    "@material-ui/core": "^4.12.3",
+    "@material-ui/icons": "^4.11.2",
```
gidjin, line 8: @Ronolibert and I found this issue too when enabling the linter in these directories. See https://ustcdp3.slack.com/archives/CP6F568DC/p1628280575183800 but since we use these deps I'm okay with it.

src/scenes/Moves/Ppm/Weight.jsx
```diff
@@ -255,6 +252,9 @@ export class PpmWeight extends Component {
         </Fragment>
       );
     }
+
+    // if there is no error to show
+    return undefined;
```
gidjin, line 257: This feels like an anti react pattern, but I don't have another solution. Maybe @Ronolibert or @lynzt have thoughts on this? Though maybe we just leave it like this for it to be fixed once PPM work begins and this code is migrated / updated to live in the new locations.

src/shared/formatters.js
```diff
@@ -105,6 +105,7 @@ export function formatDimensionsToThousandthInches(dimensions) {
   dimensions.length = formatToThousandthInches(dimensions.length);
   dimensions.width = formatToThousandthInches(dimensions.width);
   dimensions.height = formatToThousandthInches(dimensions.height);
+  return;
```
gidjin, line 108: Why are most places `return undefined` but not here?

gidjin (approved):
Looks good to me, not sure that we can even really run the PPM code at the moment. I'm approving but feel free to give @Ronolibert a chance to review before merging.

src/shared/formatters.js
mboren: `eslint` wants us to not mix uses of `return;`, `return <expr>`, and `<no return statement at end>` within a function. So this return is to match the earlier one on line 102.

reggieriser (commented):
There's also an .eslintrc in the cypress directory that looks almost identical to the two modified in this PR.  Should we enable the `consistent-return` rule there as well?

src/scenes/Moves/Ppm/Weight.jsx
mboren: I could pull the `(rateEngineError || hasEstimateError)` check out of this function and change the function to something like `if shorthaul {...} else {...}` to eliminate the `undefined` return. Is that more like what you're thinking?
```jsx
  chooseEstimateErrorText(rateEngineError) {
    if (hasShortHaulError(rateEngineError)) {
      return (
        <render  short haul error />
      );
    } else {
      return (
        <render other error />
      );
    }   
```
```jsx
{(hasEstimateError || rateEngineError) && this.chooseEstimateErrorText(rateEngineError)}
```

gidjin: Yeah that feels more like what I expect for react

mboren: I started to implement this, but it was a bit hard to test this code and have confidence that i'm not subtly breaking anything, so I think I'm gonna leave it as it is for now.

PR #7172, opened 2021-08-10
MB-9083 fix parsing of edi files with carriage returns

gidjin:
## Description

It was discovered recently that we are failing to correctly parse the 824s coming back from Syncada. This is due to those files having only carriage returns as the line delimeter. We are using the [bufio](https://pkg.go.dev/bufio) package to read in the files and by default it breaks lines up using it's [ScanLines](https://cs.opensource.google/go/go/+/refs/tags/go1.16.7:src/bufio/scan.go;l=350) method which handles lines delimited by newlines `\n` or carriage return + newlines `\r\n` but not only carriage returns `\r`. This PR adds a custom line splitter for the use with the scanner from bufio that handles all 3 cases.

## Reviewer Notes

Do we need the logger.go files I deleted? I don't think so but wanted to confirm.

## Setup

There isn't really a way to test the code other than running server tests or receiving an 824 which will happen in staging.

```sh
make db_test_reset db_test_migrate
make server_test
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9083) for this change


mboren (approved):
lgtm and tests pass.

PR #7180, opened 2021-08-11
MB-8951-index-moves-migration

NamibiaTorres:
## Description

Add a new migration to allow the prime to search the moves table when `available_to_prime_at` is set to true and search by the indexed show field.

## Reviewer Notes
Was the combo search query written correctly?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8951) for this change
* [this article](https://github.com/transcom/mymove/wiki/Database-Migrations#Creating-Migrations) explains more about the approach used.


migrations/app/schema/20210809190551_add_index_to_move_fields.up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX moves_available_to_prime_and_show_idx ON moves(show, available_to_prime_at);
```
reggieriser, line 1: It looks like we already have an index on `show` by itself.  I believe the composite index will be used if we're just searching for the first field of the composite index ([reference 1](https://dba.stackexchange.com/questions/27481/is-a-composite-index-also-good-for-queries-on-the-first-field), [reference 2](https://dba.stackexchange.com/questions/27481/is-a-composite-index-also-good-for-queries-on-the-first-field)).  So, we may want to get rid of the `show`-only index as part of this.

Speaking of `show`, there's also diminishing returns on an index for a boolean field -- or any field with low cardinality ([reference](https://stackoverflow.com/questions/42972726/postgresql-create-index-for-boolean-column)).  A partial index may be better for `show`.

That said, it can be tricky to know what the "right" indexes will be because it's influenced by the query structure and the data in the table.  You'd probably have to do some `EXPLAIN ANALYZE` calls on realistic data to get a better sense of how a specific index is helping or not.

reggieriser (commented):
The migration ran fine when I tried it, but I wonder if we also need to get rid of the existing `show` index.  I also added a note about indexing low cardinality columns like `show`.

migrations/app/schema/20210809190551_add_index_to_move_fields.up.sql
reggieriser: I took a quick look at how we use this in `ListPrimeMoveTaskOrders` (I believe that's the new, faster query) and this is the condition:
```sql
WHERE moves.available_to_prime_at IS NOT NULL AND moves.show = TRUE
```

If that's what we're trying to optimize, the better index might be a partial index on that exact condition, something like this:

```sql
CREATE INDEX moves_available_to_prime_and_show_partial_idx2 ON moves (show, available_to_prime_at)
    WHERE moves.available_to_prime_at IS NOT NULL AND moves.show = TRUE
```

I'm not sure if I have the exact syntax right, but here's the [Postgres docs on partial indexes](https://www.postgresql.org/docs/12/indexes-partial.html).

Note the index above is specific to that specific `WHERE` clause.  YMMV on other queries involving `show` or `is_available_to_prime_at`.

NamibiaTorres: Wondering if we should keep the `show` index because another team uses it regularly. There is a slack conversation about it: https://ustcdp3.slack.com/archives/CP6PTUPQF/p1628537056088700

The combo index I wrote only benefits Roci at the moment. Also, I'm new to indexing the database, and might have some followup questions after reading through the helpful resources you linked. Thanks for the feedback!

NamibiaTorres: > Wondering if we should keep the `show` index because another team uses it regularly. There is a slack conversation about it: https://ustcdp3.slack.com/archives/CP6PTUPQF/p1628537056088700
> 
> The combo index I wrote only benefits Roci at the moment. Also, I'm new to indexing the database, and might have some followup questions after reading through the helpful resources you linked. Thanks for the feedback!

NamibiaTorres: > It looks like we already have an index on `show` by itself. I believe the composite index will be used if we're just searching for the first field of the composite index ([reference 1](https://dba.stackexchange.com/questions/27481/is-a-composite-index-also-good-for-queries-on-the-first-field), [reference 2](https://dba.stackexchange.com/questions/27481/is-a-composite-index-also-good-for-queries-on-the-first-field)). So, we may want to get rid of the `show`-only index as part of this.
> 
> Speaking of `show`, there's also diminishing returns on an index for a boolean field -- or any field with low cardinality ([reference](https://stackoverflow.com/questions/42972726/postgresql-create-index-for-boolean-column)). A partial index may be better for `show`.
> 
> That said, it can be tricky to know what the "right" indexes will be because it's influenced by the query structure and the data in the table. You'd probably have to do some `EXPLAIN ANALYZE` calls on realistic data to get a better sense of how a specific index is helping or not.

Wondering if we should keep the show index because another team uses it regularly. There is a slack conversation about it: https://ustcdp3.slack.com/archives/CP6PTUPQF/p1628537056088700

The combo index I wrote only benefits Roci at the moment. Also, I'm new to indexing the database, and might have some followup questions after reading through the helpful resources you linked. Thanks for the feedback!

NamibiaTorres: > reference

@reggieriser  I'm wondering if we should keep the show index because another team uses it regularly. There is a slack conversation about it: https://ustcdp3.slack.com/archives/CP6PTUPQF/p1628537056088700

The combo index I wrote only benefits Roci at the moment. Also, I'm new to indexing the database, and might have some followup questions after reading through the helpful resources you linked. Thanks for the feedback!

reggieriser: > @reggieriser I'm wondering if we should keep the show index because another team uses it regularly.

Understood.  The [Postgres docs](https://www.postgresql.org/docs/12/indexes-multicolumn.html) say: `A multicolumn B-tree index can be used with query conditions that involve any subset of the index's columns, but the index is most efficient when there are constraints on the leading (leftmost) columns.`  So I don't think we'd be losing anything by getting rid of the solo `show` index because it's already being indexed as the first column of the multicolumn index and queries on `show` alone should use the multicolumn index.  But keeping both means we may be essentially indexing `show` twice, which could be more potential work on inserts/updates to maintain both indexes on `show`.  Hard to say if that's a significant impact or not, but the `moves` table is likely to have a lot of rows so how we index may matter more than some other tables.

duncan-truss: Great points on use of the partial indexes.  It's good to know that a multicolumn index on `(availalble_to_prime_at, show)` will still be able to work with queries that are only using the show condition, I don't think that is the case for all databases or maybe only recently in Postgres. I don't know if they're talking about space efficiency, performance, or both.  Are we making prime queries and queue queries in the same amount do we think is something else to consider what to prioritize.  While moves is a core table I don't think we are wildly inserting more than writing so probably can sustain a redundant index that speeds up our reads if that read ratio holds.

duncan-truss: I see now that you defined the order as `(show, available_to_prime_at)` whereas in the original slack thread I assumed they would be reversed as the reason why to keep the singular`show` index.

reggieriser (approved):
Talked it over with @NamibiaTorres and we decided to get rid of the existing `show`-only column.  Approving now that has been done.

We also talked about creating another story to research partial indexes and places in our schema where they may be a better fit.  I didn't want to hold up this PR any further, though.

PR #7187, opened 2021-08-12
MB-8924 TOO can request a reweigh for a shipment frontend

jenniferlynparsons:
## Description

Added the Button and modal for reweighs, reconfigured the weight display to be two boxes instead of one.

## Reviewer Notes

There are still a few outstanding questions about the behavior of the "Request reweigh" button after the reweigh is requested.

## Setup

Make sure data is up to date:

```
make db_dev_e2e_populate
```

Build and run the office app:

```
make server_run
make office_client_run
```

To check storybook changes locally:

```
make storybook
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8924) for this change

## Screenshots

Screen Shot 2021-08-12 at 6 51 25 PM
Screen Shot 2021-08-12 at 6 51 40 PM
Screen Shot 2021-08-12 at 6 58 00 PM


Shkeating (approved):
lookin good 😎

src/components/Office/ShipmentWeightDetails/ShipmentWeightDetails.test.jsx
```diff
@@ -1,25 +1,71 @@
...
+      <ShipmentWeightDetails
+        estimatedWeight={4500}
+        actualWeight={5000}
+        shipmentInfo={shipmentInfo}
+        handleRequestReweighModal={handleRequestReweighModal}
+      />,
+    );
+
+    const estWeight = await screen.findByText('Estimated weight');
+    expect(estWeight).toBeTruthy();
```
duncan-truss, line 24: I was reading about this on PR since I haven't used RTL all that much.  find* will return `{}` if there are no results and is also described as like the get* queries but with a wait for when we expect new elements to appear.  Should we just do `expect(getByText('Estimated weight'))` here if these weights are always going to be rendered visible and we don't have to test if it's true or in the document?

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -166,6 +203,15 @@ export const MoveTaskOrder = ({ match, ...props }) => {
     });
   };
 
+  const handleReweighShipment = (mtoShipmentID, eTag) => {
+    mutateMTOShipmentRequestReweigh({
+      shipmentID: mtoShipmentID,
+      operationPath: 'shipment.requestShipmentReweigh',
```
duncan-truss, line 209: I think it's okay to move this hardcoded operationPath to the ghcApi.js function, that would be more consistent with the way it's done elsewhere.

src/components/Office/ShipmentWeightDetails/ShipmentWeightDetails.test.jsx
jenniferlynparsons: That's a really good question. I think it goes outside of RTL because it's more of a Jest question: does an `expect` have to have a matcher? I've not seen an example in docs or in the wild set up with the pattern you're proposing.

... Checked the Jest docs:

> You will rarely call expect by itself. Instead, you will use expect along with a "matcher" function to assert something about a value.

So, while using the `expect` by itself might technically, work, it's not what's... expected. 😝

src/services/ghcApi.js
```diff
@@ -173,6 +173,23 @@ export function updateMTOShipmentStatus({
   );
 }
 
+export function updateMTOShipmentRequestReweigh({
+  shipmentID,
+  operationPath,
+  ifMatchETag,
+  normalize = true,
```
duncan-truss, line 180: I don't think this is used anywhere else, if we set normalize to false here then it will be less work on the response interceptor and in your callback function you can get rid of the `object.keys()` work

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -157,6 +159,41 @@ export const MoveTaskOrder = ({ match, ...props }) => {
...
+      const dataMtoShipments = Object.keys(data.mtoShipments);
+
+      const updatedMTOShipment = dataMtoShipments.filter((obj) => {
+        return obj.shipmentID === variables.shipmentID;
+      });
+      // Update mtoShipments with our updated status and set query data to match
+      mtoShipments[mtoShipments.findIndex((shipment) => shipment.id === updatedMTOShipment.id)] = updatedMTOShipment;
+      queryCache.setQueryData([MTO_SHIPMENTS, updatedMTOShipment.moveTaskOrderID, false], mtoShipments);
+      // InvalidateQuery tells other components using this data that they need to re-fetch
+      // This allows the requestCancellation button to update immediately
```
duncan-truss, line 174: let's update this to mention reweigh requested as opposed request cancellation

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -157,6 +159,41 @@ export const MoveTaskOrder = ({ match, ...props }) => {
...
+      queryCache.invalidateQueries([MTO_SHIPMENTS, updatedMTOShipment.moveTaskOrderID]);
+
+      setIsReweighModalVisible(false);
+      // Must set FlashMesage after hiding the modal, since FlashMessage will disappear when focus changes
+      setMessage(`MSG_CANCEL_SUCCESS_${variables.shipmentID}`, 'success', variables.onSuccessFlashMsg, '', true);
```
duncan-truss, line 179: We should have a different key prefix here otherwise it will clash with the cancellation flash message

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -684,6 +685,9 @@ func QueuePaymentRequests(paymentRequests *models.PaymentRequests) *ghcmessages.
 
 // Reweigh payload
 func Reweigh(reweigh *models.Reweigh) *ghcmessages.Reweigh {
+	if reweigh == nil {
```
duncan-truss, line 688: I looked at this further and it is a zero struct coming back from the handler/pop query.  I think making this condition something like `if reweigh == nil || reweigh.ID == uuid.Nil {` is a good way to do less frontend work.

jenniferlynparsons: Done. That's so much nicer.

src/components/Office/RequestShipmentCancellationModal/RequestShipmentCancellationModal.jsx
```diff
@@ -3,9 +3,9 @@ import PropTypes from 'prop-types';
 import { Button } from '@trussworks/react-uswds';
 
 import { ModalContainer, Overlay } from 'components/MigratedModal/MigratedModal';
-import Modal, { ModalTitle, ModalClose, ModalActions, connectModal } from 'components/Modal/Modal';
+import Modal, { ModalTitle, ModalClose, ModalActions } from 'components/Modal/Modal';
```
duncan-truss, line 6: Good eye, delete that sucker

src/components/Office/ShipmentDetails/ShipmentDetailsMain.jsx
```diff
@@ -36,7 +42,16 @@ const ShipmentDetailsMain = ({ className, shipment, dutyStationAddresses, handle
...
         handleDivertShipment={handleDivertShipment}
       />
-      <ShipmentWeightDetails estimatedWeight={primeEstimatedWeight} actualWeight={primeActualWeight} />
+      <ShipmentWeightDetails
+        estimatedWeight={primeEstimatedWeight}
+        actualWeight={primeActualWeight}
+        shipmentInfo={{
+          shipmentID: shipment.id,
+          ifMatchEtag: shipment.eTag,
+          reweighID: shipment.reweigh.id,
```
duncan-truss, line 51: This is giving me an undefined error so let's do the optional chaining with `shipment.reweigh?.id`

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -157,6 +159,41 @@ export const MoveTaskOrder = ({ match, ...props }) => {
...
+
+  const [mutateMTOShipmentRequestReweigh] = useMutation(updateMTOShipmentRequestReweigh, {
+    onSuccess: (data, variables) => {
+      const dataMtoShipments = Object.keys(data.mtoShipments);
+
+      const updatedMTOShipment = dataMtoShipments.filter((obj) => {
+        return obj.shipmentID === variables.shipmentID;
+      });
+      // Update mtoShipments with our updated status and set query data to match
+      mtoShipments[mtoShipments.findIndex((shipment) => shipment.id === updatedMTOShipment.id)] = updatedMTOShipment;
```
duncan-truss, line 171: if you turn normalizing back to false then this will become something like:

```
// Update mtoShipments with our updated status and set query data to match
mtoShipments[mtoShipments.findIndex((shipment) => shipment.id === data.shipmentID)].reweigh = data;
```

duncan-truss (approved):
Excellent work, approving assuming you saw my last 2 comments on the undefined reweigh id and the onSuccess callback of the react-query cache that gets rid of the reweigh requested link when the modal closes.

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -160,6 +167,36 @@ export const MoveTaskOrder = ({ match, ...props }) => {
     },
   });
 
+  const [mutateMTOShipmentRequestReweigh] = useMutation(updateMTOShipmentRequestReweigh, {
+    onSuccess: (data, variables) => {
+      // Update mtoShipments with our updated status and set query data to match
+      mtoShipments[mtoShipments.findIndex((shipment) => shipment.id === data.id)] = data;
```
duncan-truss, line 173: I think the response coming back is your reweigh object so you won't find a shipment match unless you use the shipmentID
```
{
   id:
   shipmentID,
   shipment: {
   }
   ... 
}
```

PR #7188, opened 2021-08-12
[MB-7699] CAT II Fix react/no-access-state-in-setstate finding

mboren:
## Description

This was pretty mechanical. I just sent a callback to `setState` instead of using the current `state` directly, as is
recommended in [the docs](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-access-state-in-setstate.md).
This is to prevent `state` from being evaluated before calls to `setState` which prevents basing an update off of stale state.

## Reviewer Notes

The only tricky changes I made were in `src/shared/Uploader/index.jsx`, so that could use a little extra scrutiny. I think it's probably fine but my JS/React is not super strong.

## Setup

```sh
yarn run lint
```

You can see the `Uploader` component in action at `http://milmovelocal:3000/moves/<move-id>/ppm-weight-ticket`.
Just select a type, look for one of the uploader components, upload a file, and delete the file.
(This is completely outside of the normal flow and maybe shouldn't be allowed to happen, but it works for testing this).
We're mostly just looking for this to not create an infinite loop and crash the page.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7699) for this change


src/shared/Uploader/index.jsx
```diff
@@ -135,13 +135,15 @@ export class Uploader extends Component {
...
-        this.setState({
-          files: newFiles,
+        this.setState((prevState) => {
+          const newFiles = reject(prevState.files, (upload) => upload.id === uploadId);
+          if (onChange) {
+            onChange(newFiles, this.isIdle());
+          }
+          return {
+            files: newFiles,
+          };
```
gidjin, line 145: I'm not sure of the ramifications of calling `onChange` from inside the `setState` callback. Feels like it might cause an issue but I'm not sure.

Another way that might avoid that concern might like below. Though if we can get someone from #prac-frontend to weigh in that would be good.
```javascript
this.setState((prevState) => {
  const newFiles = reject(prevState.files, (upload) => upload.id === uploadId);
  return {
    files: newFiles,
  };
})
if (onChange) {
  onChange(this.state.files, this.isIdle());
}
```

gidjin (commented):
LGTM, left a comment about the changes in `src/shared/Uploader/index.jsx` and I'm leaving this unapproved hoping someone from #prac-frontend can review.

src/shared/Uploader/index.jsx
mboren: Yeah, I am not sure if it is allowed to do the `onChange` call inside `setState`. I considered your version as well, and I think we don't have any guarantees about when the `setState` thing actually happens, so I was trying to avoid `onChange` getting called with the old `files` list before the file was removed from the list. I'm kind of shooting from the hip here though.

kylehilltruss: Pretty sure that @gidjin's solution will call `onChange` with the prior state's file array; `setState` only executes and updates state after the containing function is completed. 

It's _probably_ okay to put a callback prop execution in the setState function, but I think it's a lot cleaner to be more explicit, and do something like this:

```js
const newFiles = reject(this.state.files, (upload) => upload.id === uploadId);

this.setState({ files: newFiles })
if (onChange) {
  onChange(newFiles, this.isIdle())
}
```

mboren: I think that will work 99% of the time, but if i'm reading the [react/no-access-state-in-setstate docs](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-access-state-in-setstate.md) correctly, it doesn't fix the problem I'm trying to solve in this PR.

Since multiple `setState` calls can be batched and run together, if you pass as an argument a new state that's based on an old state, you can get conflicting results. 

say the original state has 3 files: `[foo, bar, baz]`
- deleting `foo` puts `setState({files: [bar, baz]})` in a queue (but `state` is still `{files: [foo, bar, baz]}`)
- deleting `bar` puts `setState({files: [foo, baz]})` in the queue (but `state` is still `{files: [foo, bar, baz]}`)
- react runtime decides, on a whim, that now is the best time to start updating state, and both of those queued state updates get run in succession, so our final state is: `{files: [foo, baz]}`

So i think i have to pass in a function to setState that can take the current state at the time the state update is actually run, and then i also think i need to have the onChange function run at around the same time with the same updated state. This feels very awkward though.
I don't really like putting a user supplied function inside a state update like this, just seems like it's asking for problems, so i'm hoping there is a better way that I'm missing.

mboren: I've been reading more and it seems like `setState` can take a second callback that will run after the state update, maybe that's the way to go.
```js
this.setState((prevState) => {
   const newFiles = reject(prevState.files, (upload) => upload.id === uploadId);
   return {
     files: newFiles,
   };
}, () => {
  if (onChange) {
    onChange(this.state.files, this.isIdle());
  }
});
```
still feels kinda ugly but i think it works better.

mboren: i just pushed a refactored version of the above code. I still don't feel like we've nailed it, but I think we're closer.

src/shared/Uploader/index.jsx
```diff
@@ -135,12 +135,19 @@ export class Uploader extends Component {
...
+          const newFiles = reject(prevState.files, (upload) => upload.id === uploadId);
+          return {
+            files: newFiles,
+          };
+        };
+
         if (onChange) {
-          onChange(newFiles, this.isIdle());
+          this.setState(removeFile, () => {
+            onChange(this.state.files, this.isIdle());
```
mboren, line 147: hrmph, it looks like using `this.state` in this callback is probably also wrong. I'll look for another solution.

mboren: i decided to try `componentDidUpdate` to dispatch the `onChange` calls instead.

Ronolibert (approved):
Looks good! The CDU solution behaves as expected

PR #7200, opened 2021-08-17
MB-8771 Display Reweigh tags in MTO tab

monfresh:
## Description

As a TOO, when I am viewing shipments in the Move Task Order tab, I want
to easily identify shipments that have been reweighed, and those that I
have requested to be reweighed.

### Testing Reweighed shipment
1. Sign in as a TOO and search for the move with Locator REWAYD.
2. Click on the move
3. Go to the Move task order tab
4. Look for the "Shipment weight" box 
5. Verify that it has a `REWEIGHED` tag

### Testing Reweigh Requested shipment
1. Sign in as a TOO and search for the move with Locator AMDORD (or any move with approved shipments).
2. Click on the move
3. Go to the Move task order tab
4. Look for the "Shipment weight" box
5. Click on the "Request reweigh" link
6. Click on "Submit request" when the modal appears
7. When the page refreshes, verify that it has a `REWEIGH REQUESTED` tag

## Setup

```sh
make db_dev_e2e_populate
make server_run
make office_client_run
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8771) for this change

## Screenshots

Screen Shot 2021-08-17 at 11 45 14
Screen Shot 2021-08-17 at 11 45 57

### DIFF
--- a/src/components/Office/ShipmentWeightDetails/ShipmentWeightDetails.jsx
+++ b/src/components/Office/ShipmentWeightDetails/ShipmentWeightDetails.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
-import { Button } from '@trussworks/react-uswds';
+import { Button, Tag } from '@trussworks/react-uswds';
 
 import DataPointGroup from '../../DataPointGroup/index';
 import DataPoint from '../../DataPoint/index';
@@ -21,6 +21,8 @@ const ShipmentWeightDetails = ({ estimatedWeight, actualWeight, shipmentInfo, ha
           </Button>
         </div>
       )}
+      {shipmentInfo.reweighID && !shipmentInfo.reweighWeight && <Tag>reweigh requested</Tag>}
+      {shipmentInfo.reweighWeight && <Tag>reweighed</Tag>}
     </div>
   );
   return (
@@ -45,6 +47,7 @@ ShipmentWeightDetails.propTypes = {
     shipmentID: PropTypes.string,
     ifMatchEtag: PropTypes.string,
     reweighID: PropTypes.string,
+    reweighWeight: PropTypes.number,
   }).isRequired,
   handleRequestReweighModal: PropTypes.func.isRequired,
 };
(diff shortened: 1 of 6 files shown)


src/components/Office/ShipmentWeightDetails/ShipmentWeightDetails.jsx
```diff
@@ -21,6 +21,8 @@ const ShipmentWeightDetails = ({ estimatedWeight, actualWeight, shipmentInfo, ha
           </Button>
         </div>
       )}
+      {shipmentInfo.reweighID && !shipmentInfo.reweighWeight && <Tag>reweigh requested</Tag>}
```
hannaliebl, line 24: Would it be possible to add a test for these two tags showing up?

monfresh: Yeah, for sure. That was next on my to-do list.

monfresh: Done here: https://github.com/transcom/mymove/pull/7200/commits/9d12463cda909b7ad687ca5debadef5112f44677

src/components/Office/ShipmentWeightDetails/ShipmentWeightDetails.module.scss
```diff
@@ -18,4 +18,11 @@
   .shipmentWeight {
     display: flex;
   }
+
+  :global(.usa-tag) {
+    @include u-font-size('body', 3xs);
+    @include u-margin-left(1);
+    @include u-margin-right(0);
```
monfresh, line 25: Per the design image in JIRA, I was trying to push this to the right to align the edge of the label with the end of the horizontal rule. It looks like setting the margin-right to -1 will do the trick. Is that an acceptable solution? Also, @brooksr4 can you please confirm if the label's edge is indeed supposed to line up with the horizontal rule? If I look at the text labels on the left hand side, such as "Shipment weight", it doesn't lign up, so I thought to keep things balanced, we would not line the label on the right hand side? I'm not a designer though, which is why I'm asking.

hannaliebl: Oh, interesting. If that's intended, I think a negative margin is the way to go, because that alignment comes from the parent container. Maybe there's a better way, but it seems to make sense to me. Also curious if that is intended. All tags like this will have that problem, then.

hannaliebl (approved):
LGTM!

brooksr4 (approved):

PR #7201, opened 2021-08-17
MB-9074 TOO sees tags if max billable weight needs updating on MTO tab

jenniferlynparsons:
## Description

This PR adds a dismissable alert that appears when there is an excessive weight risk.

I've also done two pieces of refactoring:
- Moved the `handleScroll` function to a utility. It appeared on all three tabs with the same logic.
- Moved the tab navigation to its own component, which reduces the complexity of TXOMoveInfo and also makes the navigation testable without resorting to mocking out React.

## Setup

Make sure data is up to date:

```
make db_dev_e2e_populate
```

Build and run the office app:

```
make server_run
make office_client_run
```
Log in as a TOO and use the `RISKEX` move to test the alert.

To check storybook changes locally:

```
make storybook
```

To test the updated stories:
[TXO Tab Navigation](http://localhost:6006/?path=/story/components-txo-tab-navigation--default)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9074) for this change

## Screenshots

Screen Shot 2021-08-19 at 3 26 48 PM

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,11 +1,15 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
 import 'happo-plugin-storybook/register';
+import { addDecorator } from '@storybook/react';
+import { MemoryRouter } from 'react-router';
 
 import './storybook.scss';
 import '../src/index.scss';
 import '../src/ghc_index.scss';
 
 import '../src/icons';
+
+addDecorator((story) => <MemoryRouter initialEntries={['/']}>{story()}</MemoryRouter>);
 
 export const parameters = {
   options: {
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,6 +1,5 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
 import 'happo-plugin-storybook/register';
-import { addDecorator } from '@storybook/react';
 import { MemoryRouter } from 'react-router';
 
 import './storybook.scss';
@@ -8,8 +7,6 @@ import '../src/index.scss';
 import '../src/ghc_index.scss';
 
 import '../src/icons';
-
-addDecorator((story) => <MemoryRouter initialEntries={['/']}>{story()}</MemoryRouter>);
 
 export const parameters = {
   options: {
--- a/src/components/Office/TXOTabNav/TXOTabNav.test.jsx
+++ b/src/components/Office/TXOTabNav/TXOTabNav.test.jsx
@@ -1,0 +1,98 @@
+import React from 'react';
+import { render, screen, within } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+
+import TXOTabNav from './TXOTabNav';
+
+const basicNavProps = {
+  order: {},
+  moveCode: 'TESTCO',
+};
+
+describe('Move details tag rendering', () => {
+  it('should render the move details tab container without a tag', () => {
+    render(<TXOTabNav {...basicNavProps} />, { wrapper: MemoryRouter });
+
+    const moveDetailsTab = screen.getByTestId('MoveDetails-Tab');
+    expect(within(moveDetailsTab).queryByTestId('tag')).toBeFalsy();
+  });
+
+  it('should render the move details tab container with a tag that shows the count of items that need attention when the orders have been amended', () => {
+    const moveDetailsAmendedOrders = {
+      ...basicNavProps,
+      order: {
+        uploadedAmendedOrderID: '1234',
+      },
+    };
+    render(<TXOTabNav {...moveDetailsAmendedOrders} />, { wrapper: MemoryRouter });
+
+    const moveDetailsTab = screen.getByTestId('MoveDetails-Tab');
+    expect(within(moveDetailsTab).getByTestId('tag')).toHaveTextContent('1');
+  });
+
+  it('should render the move details tab container with a tag that shows the count of items that need attention when there are unapproved shipments', () => {
+    const moveDetailsOneShipment = {
+      ...basicNavProps,
+      unapprovedShipmentCount: 1,
+    };
+    render(<TXOTabNav {...moveDetailsOneShipment} />, { wrapper: MemoryRouter });
+
+    const moveDetailsTab = screen.getByTestId('MoveDetails-Tab');
+    expect(within(moveDetailsTab).getByTestId('tag')).toHaveTextContent('1');
+  });
+
+  it('should render the move details tab container with a tag that shows the count of items that need attention when the orders have been amended and there are unapproved shipments', () => {
+    const moveDetailsShipmentAndAmendedOrders = {
+      ...basicNavProps,
+      order: {
+        uploadedAmendedOrderID: '1234',
+      },
+      unapprovedShipmentCount: 1,
+    };
+    render(<TXOTabNav {...moveDetailsShipmentAndAmendedOrders} />, { wrapper: MemoryRouter });
+
+    const moveDetailsTab = screen.getByTestId('MoveDetails-Tab');
+    expect(within(moveDetailsTab).getByTestId('tag')).toHaveTextContent('2');
+  });
+});
+
+describe('Move task order tag rendering', () => {
+  it('should render the move task order tab container without a tag', () => {
+    render(<TXOTabNav {...basicNavProps} />, { wrapper: MemoryRouter });
+
+    const moveTaskOrderTab = screen.getByTestId('MoveTaskOrder-Tab');
+    expect(within(moveTaskOrderTab).queryByTestId('tag')).toBeFalsy();
+  });
+
+  it('should render the move task order tab container with a tag that shows the count of shipments that have excess weight risk', () => {
+    const moveTaskOrderWithExcessRisk = {
+      ...basicNavProps,
+      excessWeightRiskCount: 1,
+    };
+    render(<TXOTabNav {...moveTaskOrderWithExcessRisk} />, { wrapper: MemoryRouter });
+
+    const moveTaskOrderTab = screen.getByTestId('MoveTaskOrder-Tab');
+    expect(within(moveTaskOrderTab).getByTestId('tag')).toHaveTextContent('1');
+  });
+  it('should render the move task order tab container with a tag that shows the count of items that need attention when there are unapproved shipments', () => {
+    const moveTaskOrderWithUnapprovedServiceItem = {
+      ...basicNavProps,
+      unapprovedServiceItemCount: 1,
+    };
+    render(<TXOTabNav {...moveTaskOrderWithUnapprovedServiceItem} />, { wrapper: MemoryRouter });
+
+    const moveTaskOrderTab = screen.getByTestId('MoveTaskOrder-Tab');
+    expect(within(moveTaskOrderTab).getByTestId('tag')).toHaveTextContent('1');
+  });
+  it('should render the move task order tab container with a tag that shows the count of items that need attention when there are unapproved ServiceItems and an excessive weight risk', () => {
+    const moveTaskOrderWithUnapprovedServiceItemAndExcessWeight = {
+      ...basicNavProps,
+      excessWeightRiskCount: 1,
+      unapprovedServiceItemCount: 1,
+    };
+    render(<TXOTabNav {...moveTaskOrderWithUnapprovedServiceItemAndExcessWeight} />, { wrapper: MemoryRouter });
+
+    const moveTaskOrderTab = screen.getByTestId('MoveTaskOrder-Tab');
+    expect(within(moveTaskOrderTab).getByTestId('tag')).toHaveTextContent('2');
+  });
+});
--- a/src/pages/Office/MoveDetails/MoveDetails.jsx
+++ b/src/pages/Office/MoveDetails/MoveDetails.jsx
@@ -9,6 +9,7 @@ import classnames from 'classnames';
 import styles from '../TXOMoveInfo/TXOTab.module.scss';
 
 import 'styles/office.scss';
+import hasRiskOfExcess from 'utils/hasRiskOfExcess';
 import { MOVES, MTO_SHIPMENTS, MTO_SERVICE_ITEMS } from 'constants/queryKeys';
 import SERVICE_ITEM_STATUSES from 'constants/serviceItems';
 import { shipmentStatuses } from 'constants/shipments';
@@ -31,11 +32,12 @@ const sectionLabels = {
   'customer-info': 'Customer info',
 };
 
-const MoveDetails = ({ setUnapprovedShipmentCount, setUnapprovedServiceItemCount }) => {
+const MoveDetails = ({ setUnapprovedShipmentCount, setUnapprovedServiceItemCount, setExcessWeightRiskCount }) => {
   const { moveCode } = useParams();
   const history = useHistory();
 
   const [activeSection, setActiveSection] = useState('');
+  const [estimatedWeightTotal, setEstimatedWeightTotal] = useState(null);
 
   const { move, order, mtoShipments, mtoServiceItems, isLoading, isError } = useMoveDetailsQueries(moveCode);
 
@@ -88,6 +90,7 @@ const MoveDetails = ({ setUnapprovedShipmentCount, setUnapprovedServiceItemCount
   const submittedShipments = mtoShipments?.filter(
     (shipment) => shipment.status === shipmentStatuses.SUBMITTED && !shipment.deletedAt,
   );
+
   const approvedOrCanceledShipments = mtoShipments?.filter(
     (shipment) =>
       shipment.status === shipmentStatuses.APPROVED ||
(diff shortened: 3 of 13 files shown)


.storybook/preview.js
```diff
@@ -1,12 +1,16 @@
...
+import { addDecorator } from '@storybook/react';
+import { MemoryRouter } from 'react-router';
 
 import './storybook.scss';
 import '../src/index.scss';
 import '../src/ghc_index.scss';
 
 import '../src/icons';
 
+addDecorator((story) => <MemoryRouter initialEntries={['/']}>{story()}</MemoryRouter>);
```
jenniferlynparsons, line 12: This made it possible to create a story for the new TXO Tab Nav, otherwise React Router was cranky.

src/pages/Office/MoveDetails/MoveDetails.jsx
```diff
@@ -9,6 +9,8 @@ import classnames from 'classnames';
 import styles from '../TXOMoveInfo/TXOTab.module.scss';
 
 import 'styles/office.scss';
+import hasRiskOfExcess from 'utils/hasRiskOfExcess';
```
jenniferlynparsons, line 12: This is here to make sure the correct tag appears the same on each page.

src/pages/Office/MoveDetails/MoveDetails.jsx
```diff
@@ -114,6 +104,27 @@ const MoveDetails = ({ setUnapprovedShipmentCount, setUnapprovedServiceItemCount
     setUnapprovedServiceItemCount(serviceItemCount);
   }, [approvedOrCanceledShipments, mtoServiceItems, setUnapprovedServiceItemCount]);
 
+  useEffect(() => {
```
jenniferlynparsons, line 107: Same deal with making sure the tags on the tabs appear the same across pages. This is something that can be addressed in a separate ticket.

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -409,7 +432,7 @@ export const MoveTaskOrder = ({ match, ...props }) => {
           {sections.map((s) => {
             const classes = classnames({ active: s.id === activeSection });
             return (
-              <a key={`sidenav_${s.id}`} href={`#shipment-${s.id}`} className={classes}>
+              <a key={`sidenav_${s.id}`} href={`#s-${s.id}`} className={classes}>
```
jenniferlynparsons, line 435: This is changed so it matches the other pages' sections. As far as I can tell, it's not used for anything else.

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -419,6 +442,11 @@ export const MoveTaskOrder = ({ match, ...props }) => {
           })}
         </LeftNav>
         <FlashGridContainer className={styles.gridContainer} data-testid="too-shipment-container">
+          {isWeightAlertVisible && (
+            <Alert slim type="warning" cta={excessWeightAlertControl} className={styles.alertWithButton}>
```
jenniferlynparsons, line 446: Used the basic Alert rather than the FlashMessage because we don't need to. This also made it easier to directly use the `cta` param as the FlashMessage doesn't and shouldn't have that prop.

src/pages/Office/TXOMoveInfo/TXOMoveInfo.jsx
```diff
@@ -48,14 +51,6 @@ const TXOMoveInfo = () => {
   if (isLoading) return <LoadingPlaceholder />;
   if (isError) return <SomethingWentWrong />;
 
-  let moveDetailsTagCount = 0;
```
jenniferlynparsons, line 51: This logic was moved to the new TXOTabNav component.

src/pages/Office/TXOMoveInfo/TXOTab.module.scss
```diff
@@ -32,6 +32,21 @@
       @include u-padding-left(4);
     }
   }
+
+  .alertWithButton {
```
jenniferlynparsons, line 36: I put this styling here for now. It likely won't need to be used in exactly this way elsewhere and can be moved when/if we have another similarly styled alert.

src/utils/handleScroll.test.js
```diff
@@ -0,0 +1,15 @@
+import handleScroll from './handleScroll';
+
+const sections = [{ id: '1' }, { id: '2' }];
+
+const setActiveSection = jest.fn();
+
+describe('handleScroll', () => {
+  it('calls setActiveSection with expected new section', () => {
```
jenniferlynparsons, line 8: Additional tests could be written here, but mocking out `window` and `document` is painful. Might be worth circling around at some point.

.storybook/preview.js
duncan-truss: Can we move this to the TXOTabNav story definition, you should be able to add a decorators block inside the export default?

src/components/Office/TXOTabNav/TXOTabNav.test.jsx
```diff
@@ -0,0 +1,98 @@
...
+import TXOTabNav from './TXOTabNav';
+
+const basicNavProps = {
+  order: {},
+  moveCode: 'TESTCO',
+};
+
+describe('Move details tag rendering', () => {
+  it('should render the move details tab container without a tag', () => {
+    render(<TXOTabNav {...basicNavProps} />, { wrapper: MemoryRouter });
```
duncan-truss, line 14: Oh cool I didn't know about the wrapper option in the render method 👍

src/pages/Office/MoveDetails/MoveDetails.jsx
```diff
@@ -31,41 +33,28 @@ const sectionLabels = {
...
-    }
-  };
+  let sections = useMemo(() => {
+    return ['orders', 'allowances', 'customer-info'];
+  }, []);
 
   useEffect(() => {
     // attach scroll listener
-    window.addEventListener('scroll', handleScroll);
+    window.addEventListener('scroll', handleScroll(sections, activeSection, setActiveSection));
```
duncan-truss, line 51: Yay thanks for moving that common logic out 🎉

src/pages/Office/MoveDetails/MoveDetails.jsx
```diff
@@ -114,6 +104,27 @@ const MoveDetails = ({ setUnapprovedShipmentCount, setUnapprovedServiceItemCount
...
+      estimatedWeightCalc = mtoShipments
+        ?.filter((s) => s.primeEstimatedWeight)
+        .reduce((prev, current) => {
+          return prev + current.primeEstimatedWeight;
+        }, 0);
+    }
+
+    setEstimatedWeightTotal(estimatedWeightCalc);
+
+    if (hasRiskOfExcess(estimatedWeightTotal, order?.entitlement.authorizedWeight)) {
```
duncan-truss, line 121: The risk of excess should compare the estimated weight to the weight allowance, which should be `order.entitlements.totalWeight`.  My Jira story mistakenly said authorizedWeight too, which if unedited are generally the same value but the TOO could edit authorized weight.

https://miro.com/app/board/o9J_l7GWMwI=/?moveToWidget=3074457361167092876&cot=14

src/pages/Office/MoveDetails/MoveDetails.jsx
```diff
@@ -114,6 +104,27 @@ const MoveDetails = ({ setUnapprovedShipmentCount, setUnapprovedServiceItemCount
     setUnapprovedServiceItemCount(serviceItemCount);
   }, [approvedOrCanceledShipments, mtoServiceItems, setUnapprovedServiceItemCount]);
 
+  useEffect(() => {
+    let estimatedWeightCalc = null;
+    let excessWeightCount = null;
+
+    if (mtoShipments?.some((s) => s.primeEstimatedWeight)) {
+      estimatedWeightCalc = mtoShipments
+        ?.filter((s) => s.primeEstimatedWeight)
```
duncan-truss, line 113: I think we might need to look at the status of the mtoShipment to only count those approved but I guess if a shipment wasn't approved it wouldn't have an estimated weight.  Although things get trickier when it comes to requested cancellation, diversion requested, and ultimately diversions that are back in submitted status.

src/pages/Office/MoveDetails/MoveDetails.jsx
jenniferlynparsons: Hmm. Okay, I originally was using `totalWeight` but then I'm never able to dismiss the alert because even when the billable weight is updated, the logic for the alert is still comparing it to the total weight so it never goes away.

src/pages/Office/MoveDetails/MoveDetails.jsx
```diff
@@ -31,41 +33,28 @@ const sectionLabels = {
   'customer-info': 'Customer info',
 };
 
-const MoveDetails = ({ setUnapprovedShipmentCount, setUnapprovedServiceItemCount }) => {
+const MoveDetails = ({ setUnapprovedShipmentCount, setUnapprovedServiceItemCount, setExcessWeightRiskCount }) => {
   const { moveCode } = useParams();
   const history = useHistory();
 
   const [activeSection, setActiveSection] = useState('');
+  const [estimatedWeightTotal, setEstimatedWeightTotal] = useState(null);
```
duncan-truss, line 41: If this is only ever updated in your useEffect I don't think we need to save it to the state.  It's only going to change if the shipments change or orders entitlement change.

src/pages/Office/MovePaymentRequests/MovePaymentRequests.jsx
```diff
@@ -29,7 +30,9 @@ const MovePaymentRequests = ({
 
   const { paymentRequests, mtoShipments, isLoading, isError } = useMovePaymentRequestsQueries(moveCode);
   const [activeSection, setActiveSection] = useState('');
-  let sections = ['payment-requests'];
+  let sections = useMemo(() => {
```
duncan-truss, line 33: It probably doesn't matter but not sure why we initialize this section only to remove it if paymentRequest.length === 0 below, it seems a little backwards.

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -393,14 +409,21 @@ export const MoveTaskOrder = ({ match, ...props }) => {
...
+          Review billable weight
+        </Button>
+      </span>
+    </span>
+  );
+  const excessWeightAlertControl = (
+    <Button type="button" onClick={handleHideWeightAlert} unstyled>
+      <FontAwesomeIcon icon="times" />
+    </Button>
+  );
```
duncan-truss, line 426: I agree defining excessWeightAlertControl is more readable here because it ends up in the cta prop.  I might embed the excessWeightAlertMessage directly in the Alert children though since it's not being used anywhere else.

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -67,14 +69,17 @@ export const MoveTaskOrder = ({ match, ...props }) => {
...
   const [isReweighModalVisible, setIsReweighModalVisible] = useState(false);
   const [isWeightModalVisible, setIsWeightModalVisible] = useState(false);
+  const [isWeightAlertVisible, setIsWeightAlertVisible] = useState(false);
+
   const [selectedShipment, setSelectedShipment] = useState(undefined);
   const [selectedServiceItem, setSelectedServiceItem] = useState(undefined);
   const [sections, setSections] = useState([]);
   const [activeSection, setActiveSection] = useState('');
   const [unapprovedServiceItemsForShipment, setUnapprovedServiceItemsForShipment] = useState({});
+  const [estimatedWeightTotal, setEstimatedWeightTotal] = useState(null);
```
duncan-truss, line 79: Same comment as on MoveDetails I think this can be a computed value w/o needing state.

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -320,31 +326,37 @@ export const MoveTaskOrder = ({ match, ...props }) => {
...
 
-    if (activeSection !== newActiveSection) {
-      setActiveSection(newActiveSection);
+    if (hasRiskOfExcess(estimatedWeightTotal, order?.entitlement.authorizedWeight)) {
+      excessBillableWeightCount = 1;
+      setExcessWeightRiskCount(excessBillableWeightCount);
     }
-  };
+
+    setIsWeightAlertVisible(excessBillableWeightCount);
```
duncan-truss, line 348: I think we're mixing booleans with integers here since it was initialized to false.  You must be dreaming of tag counts 😛

src/pages/Office/MoveDetails/MoveDetails.jsx
duncan-truss: Yea that was the point I was trying to make here https://ustcdp3.slack.com/archives/CP4979J0G/p1629139907118400 that we'll need to make an api request to acknowledge either the TOO dismissing the alert or updating the authorized weight.  Either of those needs to make an api request to change the move status as well.  I think you are referring to the MTO page this is the MoveDetails one but same issue.

hannaliebl:
Just a behavioral thing that maybe is expected, but if you click on review billable weight but don't update it (or you do, but re-type the same value), you see both the success message along with the alert message. I guess it's about the intention of the review - will someone ever come in and keep the weigh the same?

Screen Shot 2021-08-20 at 11 13 47 AM

src/pages/Office/MoveDetails/MoveDetails.jsx
jenniferlynparsons: Ah, this all makes a bit more sense now. Okay, I'll change it back to `totalWeight` and await the followup ticket.

src/pages/Office/MovePaymentRequests/MovePaymentRequests.jsx
jenniferlynparsons: Good catch. I'm removing that other if statement. There's no need for it when the payment requests don't render.

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
jenniferlynparsons: It can be computed, but unlike with MoveDetails where we're only concerned with making sure the tag count state is updated, this value is also used by the WeightDetails component. Having this value in state is probably the best solution. Otherwise we'd have to implement a `useRef` in order to make sure the value stored outside the `useEffect` gets updated.

jenniferlynparsons:
> Just a behavioral thing that maybe is expected, but if you click on review billable weight but don't update it (or you do, but re-type the same value), you see both the success message along with the alert message. I guess it's about the intention of the review - will someone ever come in and keep the weigh the same?

Good question. I would think that if they update it, but possibly not enough to bump it over that 90% threshold, the alert should still be showing because it will still be at risk. So, the way I'm seeing it, I think it's valid that both the alert and flash message could be visible at the same time.

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
duncan-truss: Good point, let's keep it then.  I think we can eventually get rid of it when we have the `excess_weight_qualifed_at` field coming back in the payload along with how we store the acknowledgement.  Then it'll only need to be calculated for the weight display.

src/pages/Office/MoveDetails/MoveDetails.jsx
```diff
@@ -106,24 +105,19 @@ const MoveDetails = ({ setUnapprovedShipmentCount, setUnapprovedServiceItemCount
 
   useEffect(() => {
     let estimatedWeightCalc = null;
-    let excessWeightCount = null;
 
     if (mtoShipments?.some((s) => s.primeEstimatedWeight)) {
       estimatedWeightCalc = mtoShipments
-        ?.filter((s) => s.primeEstimatedWeight)
+        ?.filter((s) => s.primeEstimatedWeight && s.status === SERVICE_ITEM_STATUSES)
```
duncan-truss, line 111: This is showing as the object of service item statuses, did you mean to look at shipment approved status?

src/pages/Office/MoveTaskOrder/MoveTaskOrder.jsx
```diff
@@ -327,25 +327,25 @@ export const MoveTaskOrder = ({ match, ...props }) => {
...
   useEffect(() => {
-    let estimatedWeightCalc = null;
-    let excessBillableWeightCount = null;
+    let estimatedWeightCalc = 0;
+    let excessBillableWeightCount = 0;
 
     if (mtoShipments?.some((s) => s.primeEstimatedWeight)) {
       estimatedWeightCalc = mtoShipments
-        ?.filter((s) => s.primeEstimatedWeight)
+        ?.filter((s) => s.primeEstimatedWeight && s.status === SERVICE_ITEM_STATUSES.APPROVED)
```
duncan-truss, line 335: If there isn't an object of shipment statuses or you could use a literal.

duncan-truss (approved):

amanda5iu (approved):

PR #7202, opened 2021-08-17
MB-8945: Clean up MtoShipmentForm log pollution

kylehilltruss:
## Description

As part of the [TRA epic](https://dp3.atlassian.net/browse/MB-8944) to clear out console warnings/errors during front-end test runs, this pull request fulfills a ticket to resolve emitted messages from the test fixtures for `MtoShipmentForm`. 

In this case, the emitting warnings indicated an inconsistency with how the `formatSwaggerDate` function was using the `moment` library to parse date strings in multiple possible formats (from both user input and database stored values). The easiest way to solve this issue was to adapt it to utilize a `parseDate` function that's in use elsewhere in our code base, which has `moment` check the input date string for multiple valid date formats.

## Reviewer Notes

Ensure that _all_ client and integration tests pass, and that the `MtoShipmentForm.test.jsx` file does not emit any console messages during testing. There should be no perceptible UI change to the user.

## Setup

```sh
yarn test MtoShipmentForm
```

## References

* Part of epic [MB-8944](https://dp3.atlassian.net/browse/MB-8944).
* Closes [MB-8945](https://dp3.atlassian.net/browse/MB-8945).

### DIFF
--- a/src/utils/formatMtoShipment.js
+++ b/src/utils/formatMtoShipment.js
@@ -1,7 +1,17 @@
 import { isEmpty } from 'lodash';
+import moment from 'moment';
 
 import { MTOAgentType } from 'shared/constants';
-import { formatSwaggerDate, parseSwaggerDate } from 'shared/formatters';
+import { parseDate } from 'shared/dates';
+import { parseSwaggerDate } from 'shared/formatters';
+
+const formatDateForSwagger = (date) => {
+  const parsedDate = parseDate(date);
+  if (parsedDate) {
+    return moment(parsedDate).format('YYYY-MM-DD');
+  }
+  return '';
+};
 
 function formatAgentForDisplay(agent) {
   const agentCopy = { ...agent };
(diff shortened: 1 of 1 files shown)


src/utils/formatMtoShipment.js
```diff
@@ -1,7 +1,17 @@
...
+import { parseDate } from 'shared/dates';
+import { parseSwaggerDate } from 'shared/formatters';
+
+const formatDateForSwagger = (date) => {
+  const parsedDate = parseDate(date);
+  if (parsedDate) {
+    return moment(parsedDate).format('YYYY-MM-DD');
+  }
+  return '';
+};
```
felipe-lee, line 14: My main concern with this is that we already have two other functions for formatting dates for swagger. Maybe we should have a ticket to go in and consolidate them, or to at the very least co-locate them all and document when each should be used. 

https://github.com/transcom/mymove/blob/43f05998de4dc4109c0c5b8a91bfe2fb74f86d8b/src/shared/dates.js#L33-L38

https://github.com/transcom/mymove/blob/43f05998de4dc4109c0c5b8a91bfe2fb74f86d8b/src/shared/formatters.js#L134-L141

felipe-lee (approved):
Ran tests and can confirm I'm not seeing any warnings or errors for the tests in question. Just have a comment on the new function, but as I mention in the comment, might be more of a thing to focus a new ticket on because we use these functions in a lot of places.

src/utils/formatMtoShipment.js
kylehilltruss: So, I know we want to eventually move all of our code out of `/src/shared` because it's all deprecated and not getting linted; I'm reluctant to make these sort of incremental refactors here without really knowing what I'm getting into. When we do get around to moving out `/src/shared/formatters.js`, I'll likely check that the refactor performed in the copy I made for this PR is appropriate for all uses of that file, and condense it into a single function.

PR #7204, opened 2021-08-17
context config stateless services

ahobson:
## Description

We want our services to be composable, so that one service can call
another. We also want to be able to have a per request trace id
associated with a logger so that we can correlate log messages in a
single request.

Right now, services are initialized with a database connection pool
and a logger.

When using the `database/sql` package, each database
request uses a different connection to the database. Some of our
services start transactions and then want to call other services. The
"sub-service" uses its own connection pool and thus does not have
visibility to the changes made within the transaction.

The same problem exists for logging because services are logging using
the "global" logger and not one initialized per request.

This PR changes ... everything to use a new `appcontext.AppContext`
interface that contains a database connection and a logger. This makes
every service stateless, so that all of the necessary config is passed
as the first argument to every function.

This is a massively invasive change, but one that is necessary to
address the issues above.

## Reviewer Notes

## Setup

```sh
echo "Code goes here"
```

## References

## Screenshots


monfresh (approved):
Great work. Thanks for doing this. As we've discussed, this PR was not reviewed line by line. The passing tests is giving us a decent level of confidence.

pkg/db/context.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/transcom/mymove/pkg/logging"
+)
+
+type contextKey string
+
+var dbContextKey = contextKey("db")
+
+// NewContext returns a new context with a logger.
+func NewContext(ctx context.Context, db *pop.Connection) context.Context {
```
reggieriser, line 16: Do we expect to use this anywhere?  I think we determined that we didn't want to put the connection in the Go context, so maybe this is a leftover from an earlier iteration.

pkg/services/payment_request/payment_request_creator.go
```diff
@@ -18,22 +19,20 @@ import (
...
-		db:      db,
 		planner: planner,
 		pricer:  pricer,
 	}
 }
 
-func (p *paymentRequestCreator) CreatePaymentRequest(paymentRequestArg *models.PaymentRequest) (*models.PaymentRequest, error) {
-	transactionError := p.db.Transaction(func(tx *pop.Connection) error {
+func (p *paymentRequestCreator) CreatePaymentRequest(appCtx appcontext.AppContext, paymentRequestArg *models.PaymentRequest) (*models.PaymentRequest, error) {
+	transactionError := appCtx.NewTransaction(func(txnAppCfg appcontext.AppContext) error {
```
reggieriser, line 35: We should probably rename this from `txnAppCfg` to `txnAppCtx` to match our new `appCtx` naming.  May be other places like this too.

pkg/services/payment_request/payment_request_creator.go
```diff
@@ -90,7 +85,7 @@ func (p *paymentRequestCreator) CreatePaymentRequest(paymentRequestArg *models.P
 			serviceItemMessageString := " RE Service Item Code: <" + string(reServiceItem.Code) + "> Name: <" + reServiceItem.Name + ">"
 			errMessageString := mtoMessageString + prMessageString + mtoServiceItemString + serviceItemMessageString
 			// Create the payment service item
-			paymentServiceItem, mtoServiceItem, err = p.createPaymentServiceItem(tx, paymentServiceItem, paymentRequestArg, now)
+			paymentServiceItem, mtoServiceItem, err = p.createPaymentServiceItem(appCtx.DB(), paymentServiceItem, paymentRequestArg, now)
```
reggieriser, line 88: I'm pretty sure this and any other places referencing `appCtx` in the scope of this transaction block need to be using `txAppCfg` (or `txAppCtx` if we change the name).  The error I'm seeing with a `make db_dev_e2e_populate` indicates that this code can't find the payment request record created above.  That makes sense if it's done in a separate transaction and hasn't been committed yet.  It doesn't exist as far as this code can see.

pkg/services/invoice/process_edi824.go
```diff
@@ -137,26 +131,26 @@ func (e *edi824Processor) ProcessFile(path string, stringEDI824 string) error {
...
+		err = txnAppCfg.DB().Update(&paymentRequest)
 		if err != nil {
-			e.logger.Error("failure updating payment request status:", zap.Error(err))
+			txnAppCfg.Logger().Error("failure updating payment request status:", zap.Error(err))
 			return fmt.Errorf("failure updating payment request status: %w", err)
 		}
-		e.logger.Info("SUCCESS: 824 Processor updated Payment Request to new status")
-		e.logEDIWithPaymentRequest(edi824, paymentRequest)
+		txnAppCfg.Logger().Info("SUCCESS: 824 Processor updated Payment Request to new status")
+		e.logEDIWithPaymentRequest(appCtx, edi824, paymentRequest)
```
reggieriser, line 148: Looks like another place where we should use `txnAppCtx`?

pkg/services/payment_request/payment_request_reviewed_processor.go
```diff
@@ -123,14 +115,14 @@ func (p *paymentRequestReviewedProcessor) ProcessAndLockReviewedPR(pr models.Pay
 		)
 		// Send EDI string to Syncada
 		// If sent successfully to GEX, update payment request status to SENT_TO_GEX.
-		err = paymentrequesthelper.SendToSyncada(edi858cString, icn, p.gexSender, p.sftpSender, p.runSendToSyncada, p.logger)
+		err = paymentrequesthelper.SendToSyncada(appCtx, edi858cString, icn, p.gexSender, p.sftpSender, p.runSendToSyncada)
```
reggieriser, line 118: Should use `txnAppCtx` (may be some other places in this transaction block too).

pkg/appcontext/context.go
```diff
@@ -0,0 +1,49 @@
...
+func (ac *appContext) DB() *pop.Connection {
+	return ac.db
+}
+
+func (ac *appContext) Logger() *zap.Logger {
+	return ac.logger
+}
+
+func (ac *appContext) NewTransaction(fn func(appCtx AppContext) error) error {
+	// OMFG I fucking hate go and its lack of nested transactions
```
LeDeep, line 41: don't disagree, lol. wasn't sure if you felt strongly enough to leave this in the commit.

LeDeep (approved):
this is great. talking about this for a long time, it's nice to finally see it implemented. 🚀

pkg/appcontext/context.go
ahobson: Thank you for the prompt. Comment updated

pkg/db/context.go
ahobson: Thank you. This is gone now

pkg/services/payment_request/payment_request_creator.go
ahobson: Agreed! and done

pkg/services/payment_request/payment_request_creator.go
ahobson: Thank you! Fixed

pkg/services/payment_request/payment_request_reviewed_processor.go
ahobson: thank you!

pkg/services/invoice/process_edi824.go
ahobson: thank you

reggieriser (approved):
LGTM!  Great work.  Thanks for making the additional changes.  🚀

PR #7206, opened 2021-08-18
[MB-9034] create shipment cards in storybook and use in the billableweightcard component

pearl-truss:
## Description

This PR expanded the existing `ShippingList` component to optionally display shipment weight and the missing weight and over weight flags when needed. 

This PR also updated the existing `BillableWeightCard` component to use the `ShippingList` component. Which covers the work outlined in this ticket https://dp3.atlassian.net/browse/MB-9060

## Reviewer Notes

- Does the logic for when the missing weight and over weight flags make sense? Does it make sense to pass in an entitlements array?
- I moved `ShippingList` to a /components instead of in /customer/components. Does this make sense?

## Setup

```sh
make storybook
```
And then review the ShippingList story and the BillableWeightCard story

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9034) for this change

## Screenshots

Screen Shot 2021-08-18 at 12 40 59 PM
Screen Shot 2021-08-20 at 10 17 52 AM


src/components/Office/BillableWeightCard/BillableWeightCard.test.jsx
```diff
@@ -5,13 +5,27 @@ import BillableWeightCard from './BillableWeightCard';
...
       maxBillableWeight: '13,750',
       totalBillableWeight: '12,460',
       weightRequested: '12,260',
       weightAllowance: '8,000',
+      entitlements,
+      shipments,
     };
-    const { getByText } = render(<BillableWeightCard {...weights} />);
+
+    const { getByText } = render(<BillableWeightCard {...defaultProps} />);
```
Ronolibert, line 28: I think it is recommended that we use `screen.getByText` here instead of `getByText` on the base element. In this example it is functionally the same but there are some advantages to using screen vs not. More info: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

Ronolibert (approved):
Looks good! Just had a nit on using screen but otherwise looks good

src/components/Office/BillableWeightCard/BillableWeightCard.jsx
```diff
@@ -1,8 +1,10 @@
 import React from 'react';
-import { string } from 'prop-types';
+import { string, arrayOf, shape } from 'prop-types';
 import classnames from 'classnames';
 import { Button } from '@trussworks/react-uswds';
 
+import ShipmentList from '../../ShipmentList';
```
YanZ777, line 6: You can use aliases here and for the other imports. It's not that big of a difference, but I find it easier to reason about.
`'components/...'`

src/components/ShipmentList/index.jsx
```diff
@@ -72,13 +105,29 @@ const ShipmentList = ({ shipments, onShipmentClick, moveSubmitted }) => {
...
         const shipmentNumber = shipmentNumbersByType[shipmentType];
-        const canEdit = moveSubmitted ? shipmentType === 'PPM' : true;
+        let canEdit = moveSubmitted ? shipmentType === 'PPM' : true;
+        let entitlement;
+        let isOverweight;
+        let isMissingWeight;
+        if (showShipmentWeight) {
+          canEdit = false;
+          entitlement = entitlements.find((entitlment) => entitlment.shipmentId === shipment.id);
+          if (entitlement.authorizedWeight < shipment.billableWeightCap) {
```
YanZ777, line 115: I think the logic might be flipped here?

pearl-truss: Screen Shot 2021-08-19 at 9 48 15 AM

I think the logic here is correct because in the design it shows the flag as appearing when the shipment's weight is greater than the authorized weight butI'll tag @transcom/truss-design to be sure

src/components/ShipmentList/index.jsx
```diff
@@ -32,25 +43,47 @@ const ShipmentListItem = ({ shipment, onShipmentClick, shipmentNumber, canEdit,
         {showNumber && ` ${shipmentNumber}`}
       </strong>{' '}
       {/* use substring of the UUID until actual shipment code is available */}
-      <span className={styles['shipment-code']}>#{shipment.id.substring(0, 8).toUpperCase()}</span>{' '}
+      {!showShipmentWeight && (
+        <span className={styles['shipment-code']}>#{shipment.id.substring(0, 8).toUpperCase()}</span>
+      )}{' '}
```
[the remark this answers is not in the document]
pearl-truss: I dont think formatWeight applies here because this line is for displaying the shipment id. But I did update the code to use formatWeight for displaying the billable weight

src/components/ShipmentList/index.jsx
```diff
@@ -32,25 +43,47 @@ const ShipmentListItem = ({ shipment, onShipmentClick, shipmentNumber, canEdit,
...
+          <FontAwesomeIcon icon="exclamation-triangle" className={styles.warning} />
+          <span className={styles.warningText}> Over weight</span>
+        </div>
+      )}
+      {isMissingWeight && (
+        <div>
+          <FontAwesomeIcon icon="exclamation-triangle" className={styles.warning} />
+          <span className={styles.warningText}> Missing weight</span>
+        </div>
+      )}
```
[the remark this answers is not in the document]
pearl-truss: great catch!

src/components/ShipmentList/index.jsx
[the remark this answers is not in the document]
YanZ777: That makes sense! I misread.

YanZ777 (approved):
Looks good to me!

amanda5iu (approved):
The component looks good! 

Doesn't block the story, but could we stay away from numbering the shipments like `HHG 1`? I know it's just in Storybook but don't want to give anyone the wrong idea, since we don't do this in practice.

src/components/ShipmentList/index.jsx
cosburn: Are we talking about the the "over weight" flag?

It should appear when the shipment's billable weight is greater than 110% of the shipment's estimated weight.

pearl-truss: Thanks @cosburn , I updated the logic

pearl-truss:
> The component looks good!
> 
> Doesn't block the story, but could we stay away from numbering the shipments like `HHG 1`? I know it's just in Storybook but don't want to give anyone the wrong idea, since we don't do this in practice.

Good point, I update this

PR #7210, opened 2021-08-18
[MB-4698] TOO sees reweigh weight in Shipment weight section if it is lower than prime actual weight value

hannaliebl:
## Description

Add a helper util to find the lower of two values - this is used here, and will likely be used next week when we need to tally the reweighs for all shipments. Display the lower of either the prime actual weight or the reweigh weight  in the 'Shipment weight' section.

## Setup

```sh
make db_dev_e2e_populate
make server_run
make client_run
```

Use the `REWAYD` move code, and note that the shipment weight displayed is 800 lb, which is the reweighed weight, as it is less than the initial weight (980 lb.) You can confirm this by looking at the mtoShipments query, which has the `primeActualWeight` as 980 lb. You can also note that the weight display in the upper right corner is 980 lb (fixing this to be the lower of the two values is happening next week.)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-4698) for this change


pkg/testdatagen/scenario/shared.go
```diff
@@ -3555,7 +3555,7 @@ func createHHGMoveWithReweigh(db *pop.Connection, userUploader *uploader.UserUpl
 	move := testdatagen.MakeAvailableMove(db)
 	move.Locator = "REWAYD"
 	mustSave(db, &move)
-	reweighedWeight := unit.Pound(1000)
+	reweighedWeight := unit.Pound(800)
```
hannaliebl, line 3558: I updated this so we could see the lower value being used for verification.

Ryan-Koch (approved):

src/utils/returnLowestValue.js
```diff
@@ -0,0 +1,12 @@
+export default function returnLowestValue(val1, val2) {
+  if (val1 && val2) {
```
duncan-truss, line 2: I know you asked about this in WIP demo as something that bothered you stylistically.

I think if your goal is to make this as concise as possible (not necessarily better) one other way you could structure it is:
```javascript
// changing this condition allows for returning zero here instead of previously those values being falsey. The != null condition applies to undefined too.
if (val1 != null && val2 != null) {
    return Math.min(val1, val2)
}

return val1 || val2
```
I assume you don't care if null or undefined is ultimately returned but if you did you could set default param argument values for both.

hannaliebl: Nice, I like this, thank you! I always figure there is a more concise way to do things 😄

PR #7212, opened 2021-08-19
[MB-9073] Shipment estimated weight update qualifies move for excess weight and approvals requested status

duncan-truss:
## Description

When the Prime updates the estimated weight of a shipment we want to check if the sum total of the approved shipments meets or exceeds 90% of the move's weight allowance.  If the sum total of the approved shipments meets this threshold than we'll mark the move as qualified for excess weight by recording the timestamp to the `excess_weight_qualified_at` column and change the status of the move from `APPROVED` to `APPROVALS REQUESTED`.

## Reviewer Notes

I have not yet added the case for how to handle shipments that are marked as `DIVERSION REQUESTED` or `DIVERTED` because that deals with the prime creating a shipment as well not just updating an already approved shipment and disambiguating which shipment marked diversion to count.  I'm not sure yet if I'll have time to account fro that case here.

I am also working on setting the status back to `APPROVED` in the (rare?) case that estimated weight or a shipment changing status would mean the move is no longer at risk.

I did revise the way the entitlements mapping was being initialized on every lookup of the weight allowance.  While it was not a large object, we would be doing that work on every single payload request where we're calculating totalWeight or authorizedWeight at least on the GHC side.  It should cut down on unnecessary allocations even if they were only on the function stack.

## Setup

```sh
make db_dev_e2e_populate
make server_run
make office_client_run
```

1. Login as the TOO and find the move with locator `INXSWT`
2. Approve the move along with at least 1 of the shipments
3. Using postman make a PATCH request to the updateMTOShipment endpoint. Set the body json values of the estimatedWeight to be `7200` pounds and set the scheduledPickupDate to be at least 10 days from now. 
4. The move will have the status Approvals Requested and the `excess_weight_qualified_at` timestamp will be set in the database.

Screen Shot 2021-08-19 at 7 06 07 PM

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9073 for this change

## Screenshots

Screen Shot 2021-08-19 at 7 05 01 PM
Screen Shot 2021-08-19 at 7 05 18 PM
Screen Shot 2021-08-19 at 7 15 14 PM


pkg/services/move.go
```diff
@@ -36,3 +38,9 @@ type MoveRouter interface {
 	Submit(move *models.Move) error
 	SetLogger(logger Logger)
 }
+
+// MoveWeights is the exported interface for flagging a move with an excess weight risk
+//go:generate mockery --name MoveWeights --disable-version-string
+type MoveWeights interface {
```
duncan-truss, line 44: Someone please suggest a better name 😆

rogeruiz:
```suggestion
type MoveVerify interface {
```

rogeruiz: My suggestion to `Verify` is that we may use this interface for other types of verifications in the future. I'm using `Check` as the synonym here. Some more examples include: **make sure**, confirm, corroborate, validate, substantiate.

duncan-truss: For the excess weight field I used the term `qualify`.  For more context I could see this containing the code to auto trigger a reweigh request when this excess weight threshold is met.  To me those are both things that deal with the shipment weights of an entire move.

jenniferlynparsons (approved):
Looks good to me, though someone more familiar with the Prime would be helpful to put eyes on this as well.

rogeruiz:
Verified this manually.
Captura de Pantalla 2021-08-20 a la(s) 12 28 16
Captura de Pantalla 2021-08-20 a la(s) 12 31 02

pkg/services/move.go
rogeruiz: @duncan-truss interesting... So I went with `MoveVerify` because of the name of the service but maybe we should have something like:

```suggestion
type MoveQualifyWeights interface {
```

duncan-truss: I'm going to leave it alone for now and let the method names do the hardwork.  It should be easy to change as we figure out where things will live and do.  Thanks for the review @rogeruiz !

PR #7213, opened 2021-08-19
MB-8923 remove DCRTSA service code from prime and support api

eamahanna:
## Description

This PR removes the `DCRSTA` service code from the prime and support apis. This service code is not valid for MVP, so we are removing it.

## Setup

To verify all tests still run, execute `make server_run`

To verify the prime cannot create a `DCRSTA` service item, use the createMTOServiceItems endpoint and set the reServiceCode to `DCRSTA`. An example payload is below:

```
{
  "moveTaskOrderID": "5e7fce45-ada0-4e7d-b362-9c4337af35aa",
  "modelType": "MTOServiceItemDomesticCrating",
  "reServiceCode": "DCRTSA",
  "item": {
    "type": "ITEM",
    "length": 1000,
    "width": 1000,
    "height": 1000
  },
  "crate": {
    "type": "ITEM",
    "length": 1000,
    "width": 1000,
    "height": 1000
  },
  "description": "Decorated horse head to be crated.",
  "reason": "Storage items need to be picked up"
}
```

You will get the error: 
```
{
    "title": "Validation Error",
    "instance": "5bb1e77d-9419-492d-898e-571802527a06",
    "detail": "reServiceCode in body should be one of [DCRT DUCRT]"
}
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8923) for this change

### DIFF
--- a/pkg/gen/primeapi/embedded_spec.go
+++ b/pkg/gen/primeapi/embedded_spec.go
@@ -1716,7 +1716,6 @@ func init() {
               "type": "string",
               "enum": [
                 "DCRT",
-                "DCRTSA",
                 "DUCRT"
               ]
             },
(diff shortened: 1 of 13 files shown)


rogeruiz:
@eamahanna Similar question as #7215 about removing this in other areas of the app. Since `DCRTSA` is not MVP, does that mean it needs to be removed from the `pkg/models/re_service.go` and `pkg/services/ghcimport/import_re_domestic_accessorial_prices[_test].go` files as well?? I also found it here in `pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go` with the other two service codes that the endpoint now matches.

rogeruiz:
✅ Tests passed after running make server_test locally on this branch.

eamahanna:
@rogeruiz I don't know if it needs to be removed in other parts of the app. I don't want to make assumptions beyond roci. @jamesatheyDDS do you think we need to remove it from the whole app?

sandy-wright (approved):
> @rogeruiz I don't know if it needs to be removed in other parts of the app. I don't want to make assumptions beyond roci. @jamesatheyDDS do you think we need to remove it from the whole app?

I think it does. The reason we're removing it is because it's not actually a service at all, so the concept and code is meaningless to USTC. That said, this ticket is very clear that it's about the Prime and Support YAML files, so I'd say let's make another ticket to completely remove it from the app.

Anyway, approved! Meets the AC for this ticket.

PR #7285, opened 2021-09-01
[MB-9278] create service object  to calculate billable weight for a shipment

pearl-truss:
## Description

This is a draft PR for creating a service object to calculate the billable weight for a shipment. It uses the logic:
- if a shipment has a reweigh weight and an original weight, it returns the lowest weight
- if there's a billableWeightCap set that takes precedence 

## Reviewer Notes

- I found that in order to create a new field on the mto_shipment payload, I had to also create that field on the mto_shipment model as well as add the column to the database. I'm not sure if that's what we wanted
- It was brought up that we wanted to have this service return a struct that included the calculated weight, the inputs and whether the shipment had a manual override or not. I created a type to represent this struct, should it actually be a model? I temporarily have the type defined in services/mto_shipments and think I need to move it somewhere else. If there's a better place for it to live please let me know
- I think this draft PR covers all of the AC for this ticket as I understand it. Please let me know if there's anything I missed or if invoicing or the prime will need this service object to change in some way

## Setup

```sh
echo "Code goes here"
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9278) for this change


pkg/models/mto_shipments.go
```diff
@@ -95,6 +95,7 @@ type MTOShipment struct {
 	PrimeEstimatedWeightRecordedDate *time.Time        `db:"prime_estimated_weight_recorded_date"`
 	PrimeActualWeight                *unit.Pound       `db:"prime_actual_weight"`
 	BillableWeightCap                *unit.Pound       `db:"billable_weight_cap"`
+	CalculatedBillableWeight         *unit.Pound       `db:"calculated_billable_weight"`
```
donaldthai, line 98: I see why you had to add this field along with the db migration. I do want to avoid creating a new column for that calculated billable weight since we don't plan on storing it at the moment. This brings up a good question on how exactly we want to use the service object. I'll comment more in the interface.

pkg/services/mto_shipment.go
```diff
@@ -19,6 +20,20 @@ type MTOShipmentUpdater interface {
...
+	OriginalWeight           *unit.Pound
+	ReweighWeight            *unit.Pound
+	HadManualOverride        *bool
+}
+
+//ShipmentBillableWeightCalculator is the service object interface for approving a shipment diversion
+//go:generate mockery --name ShipmentBillableWeightCalculator --disable-version-string
+type ShipmentBillableWeightCalculator interface {
+	CalculateShipmentBillableWeight(appCtx appcontext.AppContext, shipmentID uuid.UUID) (BillableWeightInputs, error)
+}
```
donaldthai, line 35: This brings up a good question on how exactly we want to use the service object. Do we want to always fetch the DB in order to calculate the billable weight? Do we want to just pass in a shipment struct with the appropriate information and calculate the billable weight that way? We may need both ways but I do feel like in most cases we'll already have fetched the shipment struct and then calculate the billable weight by passing in the shipment.

Would definitely like some thoughts on this from folks!

pkg/services/mto_shipment/shipment_billable_weight.go
```diff
@@ -0,0 +1,58 @@
...
+
+	err := appCtx.DB().Q().
+		Eager("Reweigh").
+		Find(&shipment, shipmentID)
+
+	if err != nil && errors.Cause(err).Error() == models.RecordNotFoundErrorString {
+		return services.BillableWeightInputs{}, services.NewNotFoundError(shipmentID, "while looking for shipment")
+	} else if err != nil {
+		return services.BillableWeightInputs{}, err
+	}
```
donaldthai, line 35: Seems like we don't necessary have to check specifically for `RecordNotFoundErrorString` since any errors would still mean that some error happened "while looking for shipment".

swagger/ghc.yaml
```diff
@@ -2818,6 +2818,10 @@ definitions:
         x-nullable: true
         example: 2000
         type: integer
+      calculatedBillableWeight:
```
reggieriser, line 2821: We may want to make this a read-only field as I don't think we'd ever want it to be passed in.

migrations/app/schema/20210901212329_add_calculated_billable_weight_column_to_mto_shipments_table.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE mto_shipments ADD COLUMN calculated_billable_weight int;
```
reggieriser, line 1: I saw your note about having to do this, but I believe there's a workaround -- I'll add more info in the model.

pkg/models/mto_shipments.go
reggieriser, line 98: This may help -- you can mark a field as `db:"-"` and Pop shouldn't try to store or retrieve the value from the DB ([Pop docs reference](https://gobuffalo.io/en/docs/db/models/#skipping-model-fields)).

pkg/services/mto_shipment/shipment_billable_weight.go
```diff
@@ -0,0 +1,58 @@
...
+
+	hasOverride := shipment.BillableWeightCap != nil
+	return services.BillableWeightInputs{
+		CalculatedBillableWeight: calculatedWeight,
+		OriginalWeight:           shipment.PrimeActualWeight,
+		ReweighWeight:            shipment.Reweigh.Weight,
+		HadManualOverride:        &hasOverride,
+	}, nil
+}
```
donaldthai, line 58: I think we may be missing a case here where if there is no reweigh and billable weight cap, then the calculated weight should be the original weight.

reggieriser (commented):
I haven't had a chance yet to look at the whole PR (and realize you're in draft mode), but wanted to leave a couple of initial comments.

pkg/services/mto_shipment.go
pearl-truss: yeah with that said I think it makes sense to pass in the entire shipment struct.

pearl-truss:
Thanks @donaldthai and @reggieriser for the early reviews, I believe all points have been addressed in the latest push

pkg/services/mto_shipment/shipment_billable_weight.go
```diff
@@ -0,0 +1,47 @@
...
+type shipmentBillableWeightCalculator struct {
+}
+
+// NewShipmentBillableWeightCalculator updates the address for an MTO Shipment
+func NewShipmentBillableWeightCalculator() services.ShipmentBillableWeightCalculator {
+	return &shipmentBillableWeightCalculator{}
+}
+
+// CalculateShipmentBillableWeight calculates a shipment's billable weight
+func (f *shipmentBillableWeightCalculator) CalculateShipmentBillableWeight(appCtx appcontext.AppContext, shipment *models.MTOShipment) (services.BillableWeightInputs, error) {
```
donaldthai, line 20: I believe we can remove `appCtx` and `error` from the signature since they're no longer being used.

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -292,6 +292,7 @@ func MTOShipment(mtoShipment *models.MTOShipment) *ghcmessages.MTOShipment {
 		DestinationAddress:       Address(mtoShipment.DestinationAddress),
 		PrimeEstimatedWeight:     handlers.FmtPoundPtr(mtoShipment.PrimeEstimatedWeight),
 		PrimeActualWeight:        handlers.FmtPoundPtr(mtoShipment.PrimeActualWeight),
+		CalculatedBillableWeight: handlers.FmtPoundPtr(mtoShipment.CalculatedBillableWeight),
```
donaldthai, line 295: So it looks like the `mtoShipment.CalculatedBillableWeight` field will still need to be set before mapping it to the payload. I wonder if we can just remove `CalculatedBillableWeight` from the MTOShipment model and import `CalculateShipmentBillableWeight()` directly `model_to_payload.go`? That way, we can just directly calculate the field then map it to the payload.

donaldthai (commented):
This is looking really good to me. I added some additional comments.

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
pearl-truss: @donaldthai I'm not following this all the way. I've been trying something like this (not yet working but is that what you mean?)

	payload.CalculatedBillableWeight = handlers.FmtPoundPtr(services.ShipmentBillableWeightCalculatorCalculateShipmentBillableWeight(mtoShipment).CalculatedBillableWeight)

donaldthai: Yea, something like that. Maybe something similar to:
```golang
weightsCalculator := mtoshipment.NewShipmentBillableWeightCalculator()
calculatedWeights := weightsCalculator.CalculateShipmentBillableWeight(mtoShipment)

payload.CalculatedBillableWeight = handlers.FmtPoundPtr(calculatedWeights.CalculatedBillableWeight)
```

pkg/services/mto_shipment/shipment_billable_weight.go
pearl-truss: this is fixed!

pkg/models/mto_shipments.go
```diff
@@ -95,6 +95,7 @@ type MTOShipment struct {
 	PrimeEstimatedWeightRecordedDate *time.Time        `db:"prime_estimated_weight_recorded_date"`
 	PrimeActualWeight                *unit.Pound       `db:"prime_actual_weight"`
 	BillableWeightCap                *unit.Pound       `db:"billable_weight_cap"`
+	CalculatedBillableWeight         *unit.Pound       `db:"-"`
```
donaldthai, line 98: We no longer need this right?

pearl-truss: yeah I think you're right

pkg/services/mto_shipment/shipment_billable_weight.go
```diff
@@ -0,0 +1,46 @@
...
+//shipmentBillableWeightCalculator handles the db connection
+type shipmentBillableWeightCalculator struct {
+}
+
+// NewShipmentBillableWeightCalculator updates the address for an MTO Shipment
+func NewShipmentBillableWeightCalculator() services.ShipmentBillableWeightCalculator {
+	return &shipmentBillableWeightCalculator{}
+}
+
+// CalculateShipmentBillableWeight calculates a shipment's billable weight
```
YanZ777, line 18: Could we change this comment to be a bit more descriptive? Calculate feels a bit fuzzy to me. 
I realize that the method name is the same thing, but I think having a more descriptive comment would be useful for future developers.
Using what you put in the PR description works too.

YanZ777 (commented):
The changes look good to me, but someone more familiar with backend stuff should review.

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -319,6 +320,13 @@ func MTOShipment(mtoShipment *models.MTOShipment) *ghcmessages.MTOShipment {
 		payload.ScheduledPickupDate = handlers.FmtDatePtr(mtoShipment.ScheduledPickupDate)
 	}
 
+	weightsCalculator := mtoshipment.NewShipmentBillableWeightCalculator()
```
duncan-truss, line 323: I think we may want to transition this long term to live off of the handler struct that way we aren't creating a new instance for every payload.  This payload is going to be used in a lot of queries that doesn't need this info on the Services Counselor and TOO side.  I think that'll make it easier to test as well if it's being passed as a dependency to this function or called beforehand.  Not a blocker for this initial implementation.

pkg/services/mto_shipment/shipment_billable_weight.go
```diff
@@ -0,0 +1,48 @@
...
+	if shipment.Reweigh != nil {
+		if shipment.Reweigh.Weight != nil && shipment.PrimeActualWeight != nil {
+			reweighWeight = shipment.Reweigh.Weight
+			if int(*shipment.PrimeActualWeight) < int(*reweighWeight) {
+				calculatedWeight = shipment.PrimeActualWeight
+			} else {
+				calculatedWeight = reweighWeight
+			}
+		}
+	} else if shipment.Reweigh == nil && shipment.BillableWeightCap == nil {
```
duncan-truss, line 33: Minor nitpick since this is an else I think we already know shipment.Reweigh is nil so you could likely remove the first condition.

pkg/services/mto_shipment/shipment_billable_weight.go
```diff
@@ -0,0 +1,48 @@
...
+	return &shipmentBillableWeightCalculator{}
+}
+
+// CalculateShipmentBillableWeight calculates a shipment's billable weight
+// if a shipment has a reweigh weight and an original weight, it returns the lowest weight
+// if there's a billableWeightCap set that takes precedence
+func (f *shipmentBillableWeightCalculator) CalculateShipmentBillableWeight(shipment *models.MTOShipment) services.BillableWeightInputs {
+	var calculatedWeight *unit.Pound
+	var reweighWeight *unit.Pound
+	if shipment.Reweigh != nil {
```
duncan-truss, line 24: The one thing you might need to watch out for here is a zeroed Reweigh struct where the Reweigh.ID has the value of uuid.Nil especially in a payload function as Pop may eager load that.  I think it's fine since you are checking that weight != nil below but I think you want something like `shipment.Reweigh != nil && shipment.Reweigh.ID != uuid.Nil` so the primeActualWeight gets set in that case.

donaldthai (approved):
This works for Voltron's use case! Thanks for working on this! 🚀

pkg/services/mto_shipment/shipment_billable_weight.go
```diff
@@ -0,0 +1,50 @@
...
+	return &shipmentBillableWeightCalculator{}
+}
+
+// CalculateShipmentBillableWeight calculates a shipment's billable weight
+// if a shipment has a reweigh weight and an original weight, it returns the lowest weight
+// if there's a billableWeightCap set that takes precedence
+func (f *shipmentBillableWeightCalculator) CalculateShipmentBillableWeight(shipment *models.MTOShipment) services.BillableWeightInputs {
+	var calculatedWeight *unit.Pound
+	var reweighWeight *unit.Pound
+	if shipment.Reweigh != nil && shipment.Reweigh.ID != uuid.Nil {
```
reggieriser, line 26: What would happen if someone passed in a `shipment` that had not already eagerly loaded the `Reweigh` field?  Wouldn't we possibly return the wrong calculated weight (for instance, if the reweigh was less than the original weight) since we wouldn't be factoring in a legit reweigh from the DB?  Is there a way we can detect and potentially error if the `Reweigh` isn't eagerly loaded?  If not, then I think we need to highlight in the comments on this method that we require the `Reweigh` to be in place before this is called.

pearl-truss: @reggieriser that's a good point, I updated the code to throw an error if the Reweigh field hadn't been loaded

PR #7292, opened 2021-09-02
[MB-9112] TOO can approve and deny SIT Extensions

hannaliebl:
## Description

Opening a draft PR of this. So far I have done the SIT Extension approval endpoint. If this is an ok pattern/way of doing things, I hope the deny endpoint won't be too hard to work into this.

## Reviewer Notes

Outstanding questions I have:
* I have a TODO in the tests for the SIT Extension Approver service - I'm not sure this is indicative that my test is setup wrong, or that the service itself isn't properly updating the SIT extension.
* The service object I created is just for approvals - should it be for both approve and deny, though?
* Am I missing nil checks? Anything else? I've not done a net-new thing like this that updates multiple models before.

Work left:
- Handler tests
- Deny SIT extension endpoint/tests

Open to anything else

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9112) for this change


pkg/services/mto_shipment/sit_extension_approver.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	return &sitExtension, nil
+}
+
+func (f *sitExtensionApprover) approveSITExtension(appCtx appcontext.AppContext, shipment models.MTOShipment, sitExtension models.SITExtension, approvedDays *int, officeRemarks *string) (*models.MTOShipment, error) {
+	var returnedShipment models.MTOShipment
+
+	transactionError := appCtx.NewTransaction(func(txnAppCtx appcontext.AppContext) error {
+		if err := f.updateApprovedDays(txnAppCtx, sitExtension, approvedDays); err != nil {
```
hannaliebl, line 75: Is this pattern of handling every update to every field in a separate function that validates it correct?

swagger/ghc.yaml
```diff
@@ -3022,6 +3127,19 @@ definitions:
         example: MTO Shipment not good enough
     required:
       - rejectionReason
+  ApproveSitExtension:
+    properties:
+      approvedDays:
+        description: Number of days approved for SIT extension
+        type: integer
+        example: 21
```
duncan-truss, line 3135: You should be able to set a minimum value here, I'm assuming it has to be 1 or greater.

pkg/handlers/ghcapi/mto_shipment.go
```diff
@@ -719,3 +720,48 @@ func (h RequestShipmentReweighHandler) triggerRequestShipmentReweighEvent(shipme
...
+	}
+
+	if !session.IsOfficeUser() || !session.Roles.HasRole(roles.RoleTypeTOO) {
+		return handleError(services.NewForbiddenError("is not a TOO"))
+	}
+
+	shipmentID := uuid.FromStringOrNil(string(params.ShipmentID))
+	sitExtensionID := uuid.FromStringOrNil(string(params.SitExtensionID))
+	approvedDays := swag.Int(int(*params.Body.ApprovedDays))
+	officeRemarks := swag.String(*params.Body.OfficeRemarks)
```
duncan-truss, line 758: If `params.Body.OfficeRemarks` is optional it could be nil here so we shouldn't dereference it and it's being passed as a pointer below to the ApproveSITExtension service.  You should be able to pass it directly through.

pkg/handlers/ghcapi/mto_shipment.go
```diff
@@ -719,3 +720,48 @@ func (h RequestShipmentReweighHandler) triggerRequestShipmentReweighEvent(shipme
...
+		}
+	}
+
+	if !session.IsOfficeUser() || !session.Roles.HasRole(roles.RoleTypeTOO) {
+		return handleError(services.NewForbiddenError("is not a TOO"))
+	}
+
+	shipmentID := uuid.FromStringOrNil(string(params.ShipmentID))
+	sitExtensionID := uuid.FromStringOrNil(string(params.SitExtensionID))
+	approvedDays := swag.Int(int(*params.Body.ApprovedDays))
```
duncan-truss, line 757: `params.Body.ApprovedDays` should be required so I would simplify this as `approvedDays := int(*params.Body.ApprovedDays)` then update your service to take a string value instead of the string pointer.

Most of these `swag.*` helper functions are for use in the payload functions to convert values from the data models to pointers that the swagger types use.

swagger/ghc.yaml
```diff
@@ -909,6 +909,58 @@ paths:
       description: Requests a shipment reweigh
       operationId: requestShipmentReweigh
       summary: Requests a shipment reweigh
+  '/shipments/{shipmentID}/sit-extensions/{sitExtensionID}/approve':
+    parameters:
```
duncan-truss, line 913: We should have a param here that passes the sitExtension eTag.  That means in JP's SITExtension payload the eTag should be coming back set in that response.

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -360,6 +367,34 @@ func MTOAgents(mtoAgents *models.MTOAgents) *ghcmessages.MTOAgents {
...
+		DecisionDate:      strfmt.DateTime(*sitExtension.DecisionDate),
+		OfficeRemarks:     string(*sitExtension.OfficeRemarks),
+		CreatedAt:         strfmt.DateTime(sitExtension.CreatedAt),
+		UpdatedAt:         (*strfmt.DateTime)(&sitExtension.UpdatedAt),
+	}
+	return payload
+}
+
+// SITExtensions payload
+func SITExtensions(sitExtensions *models.SITExtensions) *ghcmessages.SitExtensions {
```
duncan-truss, line 388: I believe JP was adding nil checks here right?

pkg/services/mto_shipment/sit_extension_approver.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	sitExtension, err := f.findSITExtension(appCtx, sitExtensionID)
+	if err != nil {
+		return nil, err
+	}
+
+	existingETag := etag.GenerateEtag(shipment.UpdatedAt)
+	if existingETag != eTag {
+		return &models.MTOShipment{}, services.NewPreconditionFailedError(shipmentID, query.StaleIdentifierError{StaleIdentifier: eTag})
```
duncan-truss, line 39: Not a huge concern but you can return nil here instead of an empty `&models.MTOShipment{}` unless that was causing you problems.  It's only when the return value is not a pointer type that you must return an empty struct to satisfy the type.

pkg/services/mto_shipment/sit_extension_approver.go
```diff
@@ -0,0 +1,175 @@
...
+func (f *sitExtensionApprover) ApproveSITExtension(appCtx appcontext.AppContext, shipmentID uuid.UUID, sitExtensionID uuid.UUID, approvedDays *int, officeRemarks *string, eTag string) (*models.MTOShipment, error) {
+	shipment, err := f.findShipment(appCtx, shipmentID)
+	if err != nil {
+		return nil, err
+	}
+
+	sitExtension, err := f.findSITExtension(appCtx, sitExtensionID)
+	if err != nil {
+		return nil, err
+	}
```
duncan-truss, line 35: If we're going to use 2 separate queries I think we'll need to make sure that the shipmentID provided is the same as the one linked to from the SITExtension.  In theory you could update a different shipment's allotment by approving another shipment's SITExtension 😛

pkg/services/mto_shipment/sit_extension_approver.go
duncan-truss: I think instead you want to set all of the values (approvedDays, officeRemarks, status, decisionDate) on your sitExtension model that you fetched from the database.  Validate that as a whole, then we'll only have to do one update/write to the database.

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
hannaliebl: Ah, yes, I will check in with them.

hannaliebl:
@duncan-truss thanks for the review! I've addressed this initial round of feedback, I believe. And now the tests I wrote pass 😄  I'll work on the handler tests and adding the deny endpoint and then remove this from draft status. If you see anything else, let me know!

pkg/services/mto_shipment/sit_extension_approver.go
hannaliebl: Was this why my tests weren't passing? I still don't actually know why the stuff I had commented out initially worked after 4a4d582

pkg/services/mto_shipment/sit_extension_approver.go
```diff
@@ -34,9 +34,13 @@ func (f *sitExtensionApprover) ApproveSITExtension(appCtx appcontext.AppContext,
 		return nil, err
 	}
 
+	if sitExtension.MTOShipmentID != shipment.ID {
+		return nil, errors.New("SITExtension's shipment ID does not match shipment ID provided")
```
duncan-truss, line 38: This may end being treated as a 500 error because it doesn't match one of the error types declared in your handler.  I would say treat it like either a NotFoundError or InvalidInputError.

pkg/services/mto_shipment/sit_extension_approver.go
```diff
@@ -148,12 +114,12 @@ func (f *sitExtensionApprover) updateSITExtensionDecisionDate(appCtx appcontext.
 	return nil
 }
 
-func (f *sitExtensionApprover) updateSitDaysAllowance(appCtx appcontext.AppContext, shipment models.MTOShipment, approvedDays *int) (*models.MTOShipment, error) {
+func (f *sitExtensionApprover) updateSitDaysAllowance(appCtx appcontext.AppContext, shipment models.MTOShipment, approvedDays int) (*models.MTOShipment, error) {
 	if shipment.SITDaysAllowance != nil {
-		sda := int(*approvedDays) + int(*shipment.SITDaysAllowance)
+		sda := approvedDays + int(*shipment.SITDaysAllowance)
```
duncan-truss, line 119: We don't have to look at the allowance here of the orders do we?  Is that work happening elsewhere to set the shipment's SITDaysAllowance when the shipment is created?

pkg/services/mto_shipment/sit_extension_approver.go
```diff
@@ -0,0 +1,141 @@
...
+	sitExtension, err := f.findSITExtension(appCtx, sitExtensionID)
+	if err != nil {
+		return nil, err
+	}
+
+	if sitExtension.MTOShipmentID != shipment.ID {
+		return nil, errors.New("SITExtension's shipment ID does not match shipment ID provided")
+	}
+
+	existingETag := etag.GenerateEtag(shipment.UpdatedAt)
```
duncan-truss, line 41: We should be explicit maybe in the swagger docs or elsewhere that we're expecting the shipment eTag instead of the SITExtension eTag.  It's a little confusing but this happens to us elsewhere like updating entitlements of orders where we send in the orders eTag I think.  I guess it will be consistent when we are creating a new SITExtension as the TOO and that can use the shipment's eTag too.  There is the chance the SITExtension could be updated in the meantime and we'd overwrite it if it didn't update the shipment record too.

pkg/services/mto_shipment/sit_extension_approver.go
hannaliebl: Good question, I think that is being done here: https://dp3.atlassian.net/browse/MB-9243

pkg/services/mto_shipment/sit_extension_approver.go
hannaliebl: 😬  I did some searching, but how do you access/update the swagger docs? Is it done via the ghc yaml file? I wasn't sure this was the correct way to handle this, but because updating a SIT Extension, as best I understand, is always done in context of the shipment it belongs to, I thought the shipment etag made the most sense.

duncan-truss: Yea I was thinking you can add a description property to the If-Match header of this endpoint in the ghc.yaml.  Locally you can read the swagger docs at http://milmovelocal:3000/swagger-ui/ghc.html

pkg/services/mto_shipment/sit_extension_approver.go
```diff
@@ -0,0 +1,141 @@
...
+	if transactionError != nil {
+		return nil, transactionError
+	}
+
+	return &returnedShipment, nil
+}
+
+func (f *sitExtensionApprover) updateSITExtension(appCtx appcontext.AppContext, sitExtension models.SITExtension, approvedDays int, officeRemarks *string) error {
+	sitExtension.ApprovedDays = &approvedDays
+	if officeRemarks != nil {
```
duncan-truss, line 102: We don't need to check for nil here since we are just assigning the pointer value directly, unless you were trying to prevent someone from say erasing the existing remarks.

pkg/services/mto_shipment/sit_extension_denier.go
```diff
@@ -0,0 +1,125 @@
...
+
+func (f *sitExtensionDenier) denySITExtension(appCtx appcontext.AppContext, shipment models.MTOShipment, sitExtension models.SITExtension, officeRemarks *string) (*models.MTOShipment, error) {
+	var returnedShipment models.MTOShipment
+
+	transactionError := appCtx.NewTransaction(func(txnAppCtx appcontext.AppContext) error {
+		if err := f.updateSITExtension(txnAppCtx, sitExtension, officeRemarks); err != nil {
+			return err
+		}
+
+		verrs, err := appCtx.DB().ValidateAndUpdate(&shipment)
```
duncan-truss, line 83: Do we need this shipment update if nothing is changing due to the SIT Extension denial?  I could see that be necessary if we were denying a previously approved request but that would get into changing the days count.

pkg/handlers/ghcapi/mto_shipment.go
```diff
@@ -719,3 +720,92 @@ func (h RequestShipmentReweighHandler) triggerRequestShipmentReweighEvent(shipme
...
+
+	shipmentID := uuid.FromStringOrNil(string(params.ShipmentID))
+	sitExtensionID := uuid.FromStringOrNil(string(params.SitExtensionID))
+	approvedDays := int(*params.Body.ApprovedDays)
+	officeRemarks := params.Body.OfficeRemarks
+	updatedShipment, err := h.SITExtensionApprover.ApproveSITExtension(appCtx, shipmentID, sitExtensionID, approvedDays, officeRemarks, params.IfMatch)
+	if err != nil {
+		return handleError(err)
+	}
+
```
duncan-truss, line 763: I think we should trigger an event notification to the prime making them aware that the SIT Extension has been reviewed by the TOO. Here's an example of how we do it for Reweigh https://github.com/transcom/mymove/blob/624e1ab3e1745b179251f0fd22a4eb03a4cd0322/pkg/handlers/ghcapi/mto_shipment.go#L690

pkg/services/mto_shipment/sit_extension_denier.go
hannaliebl: Yeah, nothing is updated on the shipment directly, but I still want to return the shipment because ultimately the endpoint needs the shipment with the sit extension that is now denied. Can I do that here, like just do something like `appCtx.DB().Q().Find` to find the shipment (that now has a denied SIT extension associated with it.)

duncan-truss: Yea I would say you could just return the shipment that you are passing in as a param to this function but your updateSITExtension call does not return the updated SITExtension model or modify the sitExtension var directly.  So yes you could do another find and make sure you eager join the SITExtension association to get the newly denied data because ValidateAndUpdate won't fetch associations any associations to my knowledge.  It shouldn't need to be in the transaction either if it's just a select statement.

pkg/handlers/ghcapi/mto_shipment.go
duncan-truss: Sounds good thanks for following that thread

pkg/handlers/ghcapi/mto_shipment_test.go
```diff
@@ -1481,6 +1480,91 @@ func (suite *HandlerSuite) TestRequestShipmentReweighHandler() {
...
+			},
+			ShipmentID:     *handlers.FmtUUID(mtoShipment.ID),
+			SitExtensionID: *handlers.FmtUUID(sitExtension.ID),
+		}
+		response := handler.Handle(approveParams)
+		okResponse := response.(*shipmentops.ApproveSitExtensionOK)
+		payload := okResponse.Payload
+		suite.IsType(&shipmentops.ApproveSitExtensionOK{}, response)
+		suite.Equal(int64(30), *payload.SitDaysAllowance)
+		// this should work, I think, but doesn't
```
hannaliebl, line 1522: Calling this out. It seems like there is no SIT Extension at all on this, as I get a `panic: runtime error: index out of range [0] with length 0`

pkg/handlers/ghcapi/mto_shipment_test.go
```diff
@@ -1481,6 +1480,91 @@ func (suite *HandlerSuite) TestRequestShipmentReweighHandler() {
...
+				OfficeRemarks: &officeRemarks,
+			},
+			ShipmentID:     *handlers.FmtUUID(mtoShipment.ID),
+			SitExtensionID: *handlers.FmtUUID(sitExtension.ID),
+		}
+		response := handler.Handle(denyParams)
+		// okResponse := response.(*shipmentops.DenySitExtensionOK)
+		// payload := okResponse.Payload
+		suite.IsType(&shipmentops.DenySitExtensionOK{}, response)
+		// this should work, I think, but doesn't
```
hannaliebl, line 1563: Same as above - I would like to see that the SIT extension is in denied status, but I get `panic: runtime error: index out of range [0] with length 0` which seems like there isn't one on the shipment at all?

pkg/handlers/ghcapi/mto_shipment_test.go
duncan-truss: This is related to you needing to eager join the SITExtensions association in your findShipment or make sure it is set on the shipment you return from your approve/deny services.  It may be there in the database but Pop won't fetch the different associations on the shipment model unless you tell it to.

hannaliebl: I added this to the services and now these tests pass, thanks!

pkg/services/mto_shipment/sit_extension_denier.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (f *sitExtensionDenier) denySITExtension(appCtx appcontext.AppContext, shipment models.MTOShipment, sitExtension models.SITExtension, officeRemarks *string) (*models.MTOShipment, error) {
+	var returnedShipment models.MTOShipment
+
+	transactionError := appCtx.NewTransaction(func(txnAppCtx appcontext.AppContext) error {
+		if err := f.updateSITExtension(txnAppCtx, sitExtension, officeRemarks); err != nil {
+			return err
+		}
+
+		if e := appCtx.DB().Q().EagerPreload("SITExtensions").Find(&returnedShipment, shipment.ID); e != nil {
```
duncan-truss, line 87: When testing this I saw it wasn't returning the updated SITExtension I think because this is inside of a transaction you need to use the txnAppCtx to get the database and not the appCtx.  Alternatively you could use the appCtx if you do the Find after the transaction block.

duncan-truss (approved):
I tested this in Postman and got the right results in the database 🎉 The only outstanding issue I saw was the return value of the deny request did not have the updated SIT extension data but it was correct in the database.

pkg/services/mto_shipment/sit_extension_denier.go
hannaliebl: Ah, ok, that makes sense. I'll just use the `txnAppCtx` unless there is a preference to do it the other way?

pkg/handlers/ghcapi/mto_shipment.go
rogeruiz: hey @duncan-truss and @hannaliebl I understand that this is now on Roci to do this work, but I wanted to add that previously when @monfresh worked on a similar feature that triggered a notification it was done as a part of the original PR.

I don't think it needs to be done by y'all after the fact. I want to make a note of this inconsistency though when it comes to approaches for particular type of work. Thanks for having this conversation. It was very helpful in Roci Sprint planning today to get us on the same page about what kind of effort was required from Roci for this upcoming Sprint starting tomorrow.

The PR that I am referring to @monfresh is this one, https://github.com/transcom/mymove/pull/7138/files#diff-c6d42978ef86661959c232f7df52de5863e180caecd611d271153534b48cd678R682-R700. I've gone and highlighted the code around this feature.

Lastly, I will say again that I appreciate the conversation around this and referencing the MB-9088 ticket as it makes sense for Roci to take on this sort of work to at the very least get more understanding behind the backend work that went into this PR. What I'm noting here more than anything is the inconsistent approaches behind this sort of work and how it can be confusing when some team members do the extra work while others delegate it.

hannaliebl: Ah, interesting. Yeah, so maybe the best way of handling this in the future is either one ticket that covers the notification work, or the notifications ticket should have been TTV's?

PR #4313, opened 2020-06-23
MB-2737 Ignore service item params in CreatePaymentRequestPayload

noelledusahel:
## Description

The Prime may attempt to send Service Item Params through the Create Payment Request payload, but we want the source of truth to be the MTO. This PR achieves ignoring the input of service item params by converting this part of the payload to readOnly, and removing the handler function that assigns serviceItemParams to serviceItems. At this time there can be a CreatePaymentRequest success when serviceItemParams are passed, without throwing an error. 

## Reviewer Notes

Is there anything you would like reviewers to give additional scrutiny?

## Setup

- `make db_dev_e2e_populate`
- `make server_run`
- Make a `pr_payload.json` file containing the following (which references E2E data) **running with this payload should produce an error, to make it successful, then remove the params part of the payload**:
```
  {
    "body": {
      "isFinal": false,
      "moveTaskOrderID": "9c7b255c-2981-4bf8-839f-61c7458e2b4d",
      "serviceItems": [
        {
          "id": "76055c99-0990-410c-a7c9-69373b0b53eb",
          "params": [
            {
             "key": "weightActual",
              "value": "1111"
            }
          ]
        }
      ]
    }
  }
```
- `go run ./cmd/prime-api-client --insecure create-payment-request --filename pr_payload.json | jq`
- Verify parameters and that the payment request is created without the service item params passed in through the payment request payload (in this case weightActual).  

## Code Review Verification Steps

* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/wiki/backend#logging)
* [ ] The requirements listed in
 [Querying the Database Safely](https://github.com/transcom/mymove/wiki/backend#querying-the-database-safely)
 have been satisfied.
* Any new migrations/schema changes:
  * [ ] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://github.com/transcom/mymove/wiki/migrate-the-database#zero-downtime-migrations))
  * [ ] Have been communicated to #g-database
  * [ ] Secure migrations have been tested following the instructions in our [docs](https://github.com/transcom/mymove/wiki/migrate-the-database#secure-migrations)
* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
[...]

PR #7295, opened 2021-09-02
[MB-9089] Prime can issue partial invoices for SIT

mboren:
## Description

We previously only allowed the Prime to bill for SIT 29 days at a time (or, if a departure date was present, whatever was left). We want to allow more flexibility than that.
This PR changes allows the Prime to specify the start and end dates of the period they are billing for when requesting a payment request for additional days of SIT (DOASIT, DDASIT).

These dates are not used for pricing yet, that will be handled in [MB-9253](https://dp3.atlassian.net/browse/MB-9253).

I had to add two new payment service item param keys (start date and end date), and I had to update the `createPaymentRequest` endpoint to allow specifying additional params for each service item.
This was something that we used to allow, so I was able to resurrect some code that we deleted in [this PR](https://github.com/transcom/mymove/pull/4313).
Here's an example of what that looks like:
```
...
  "serviceItems": [
    {
      "id": "<DDASIT Service Item ID>",
      "params": [
        {
          "key": "SITPaymentRequestStart",
          "value": "2020-09-01"
        },
        {
          "key": "SITPaymentRequestEnd",
          "value": "2020-09-08"
        }
      ]
    }
  ]
 ...
```

I also some validation to make sure that the params passed when creating a payment request are expected. In the future, if we add other params that are passed in the same way, we'll need to add them to the allowlist.
It doesn't seem like we're going to do this very often, so I wanted to make sure there were good error messages for doing it wrong.

I spent a lot of time thinking about validation and error handling for this change, but ended up leaving out most of it from this PR because I think it warrants more testing, and this PR has been dragging on for a while.
I think the best way to handle errors is to have these params validated in the lookups that use them (in this case that would just be the number of days in SIT lookup) and bubble up errors from lookups to the prime. 

## Setup
First, create a move and approve as TOO.

Use the pricing acceptance script to create a new shipment and update dates and weights on it
but kill it with ctrl-c when it prompts to create a service item (we don't want it to create a payment request).
```sh
pricing-acceptance <move code>
```

Take the MTO ID and shipment ID from the `pricing-acceptance` script output and use them to fill in the following request body for creating a DDFSIT service item.

```json
{
  "body": {
    "moveTaskOrderID": "<your MTO ID>",
    "mtoShipmentID": "<your MTO shipment ID>",
    "modelType": "MTOServiceItemDestSIT",
    "reServiceCode": "DDFSIT",
    "timeMilitary1": "1705Z",
    "firstAvailableDeliveryDate1": "2021-09-27",
    "timeMilitary2": "0719Z",
    "firstAvailableDeliveryDate2": "2021-09-29",
    "sitEntryDate": "2021-08-01"
  }
}
```
```sh
prime-api-client --insecure create-mto-service-item --filename create_ddfsit.json
```
Creating DDFSIT will automatically create DDASIT and DDDSIT service items.
You can get the ID of the DDASIT service item from the response.
We want to create a payment request for DDASIT and pass it the new parameters.

```json
{
  "body": {
    "isFinal": false,
    "moveTaskOrderID": "<your MTO ID>",
    "serviceItems": [
      {
        "id": "<DDASIT Service Item ID>",
        "params": [
          {
            "key": "SITPaymentRequestStart",
            "value": "2020-09-01"
          },
          {
            "key": "SITPaymentRequestEnd",
            "value": "2020-09-08"
          }
        ]
      }
    ]
  }
}
```
```sh
prime-api-client --insecure create-payment-request --filename create_payment_request.json
```

Pricing doesn't actually use the new parameters yet, so we're just looking for the payment request creation to be successful.

## References
* [Jira story](https://dp3.atlassian.net/browse/MB-9089) for this change


Ryan-Koch (approved):
The migration LGTM, I'd suggest getting another review before merging though as I didn't dig into the rest of it.

gidjin (commented):
leaving comments, haven't had a chance to run the code locally yet, but should after my morning meeting ends.

pkg/handlers/primeapi/payment_request_test.go
```diff
@@ -214,10 +228,10 @@ func (suite *HandlerSuite) TestCreatePaymentRequestHandler() {
 				MoveTaskOrderID: handlers.FmtUUID(subtestData.moveTaskOrderID),
 				ServiceItems: []*primemessages.ServiceItem{
 					{
-						ID: *handlers.FmtUUID(subtestData.serviceItemID1),
+						ID: *handlers.FmtUUID(subtestData.serviceItemID3),
 						Params: []*primemessages.ServiceItemParamsItems0{
 							{
-								Key:   "weight",
+								Key:   string(models.ServiceItemParamNameSITPaymentRequestStart),
 								Value: "5678",
```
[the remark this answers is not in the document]
mboren: Done

pkg/handlers/primeapi/payment_request_test.go
```diff
@@ -227,9 +241,59 @@ func (suite *HandlerSuite) TestCreatePaymentRequestHandler() {
 			},
 		}
 		response := handler.Handle(params)
-		suite.IsType(&paymentrequestop.CreatePaymentRequestUnprocessableEntity{}, response)
+		suite.IsType(&paymentrequestop.CreatePaymentRequestCreated{}, response)
 	})
 
+	suite.Run("fail to create payment request with invalid SIT dates", func() {
```
mboren, line 247: Oops, this test was not meant to make it in here. The validation it's testing isn't implemented yet.

pkg/handlers/primeapi/payment_request.go
```diff
@@ -180,20 +180,21 @@ func (h CreatePaymentRequestHandler) buildPaymentServiceItems(payload *primemess
 	return paymentServiceItems, verrs, nil
 }
 
-func (h CreatePaymentRequestHandler) buildPaymentServiceItemParams(payloadMTOServiceItem *primemessages.ServiceItem, reService models.ReService) (models.PaymentServiceItemParams, error) {
-	/************
-	  ServiceItem.params is set to readOnly = true currently in prime.yaml. Therefore we are only checking if
-	  there were params sent. If there were params in via the create payment request then we will error out.
+func (h CreatePaymentRequestHandler) buildPaymentServiceItemParams(reServiceCode models.ReServiceCode, payloadMTOServiceItem *primemessages.ServiceItem) (models.PaymentServiceItemParams, error) {
```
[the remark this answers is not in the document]
mboren: Additional historical context: `ReService` was just being used for the error message. if you go back to the time when we allowed the prime to set any param, the signature didn't have anything from `ReService`.
```go
func (h CreatePaymentRequestHandler) buildPaymentServiceItemParams(payloadMTOServiceItem *primemessages.ServiceItem) models.PaymentServiceItemParams
```

pkg/handlers/primeapi/payment_request_test.go
```diff
@@ -214,10 +228,10 @@ func (suite *HandlerSuite) TestCreatePaymentRequestHandler() {
 				MoveTaskOrderID: handlers.FmtUUID(subtestData.moveTaskOrderID),
 				ServiceItems: []*primemessages.ServiceItem{
 					{
-						ID: *handlers.FmtUUID(subtestData.serviceItemID1),
+						ID: *handlers.FmtUUID(subtestData.serviceItemID3),
 						Params: []*primemessages.ServiceItemParamsItems0{
 							{
-								Key:   "weight",
+								Key:   string(models.ServiceItemParamNameSITPaymentRequestStart),
```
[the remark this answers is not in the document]
mboren: That makes sense. I can certainly add the missing param to this test, but I'd just as soon leave it until we're actually requiring it.

gidjin (approved):
Changes look good and I tried out creating a payment request for DDASIT and DUPK with the new params and it worked as expected. I'm find leaving the check for both dates until the pricer story.

pkg/handlers/primeapi/allowed_payment_service_item_params.go
```diff
@@ -0,0 +1,35 @@
...
+		models.ReServiceCodeDOASIT: {
+			models.ServiceItemParamNameSITPaymentRequestStart,
+			models.ServiceItemParamNameSITPaymentRequestEnd,
+		},
+		models.ReServiceCodeDDASIT: {
+			models.ServiceItemParamNameSITPaymentRequestStart,
+			models.ServiceItemParamNameSITPaymentRequestEnd,
+		},
+	}
+)
```
reggieriser, line 21: This feels like the sort of thing that could be part of the metadata we store in the `service_params` table.  I did something similar recently with adding the `is_optional` boolean flag there.  I don't think it's something to worry about for this PR, but may be worth adding a story to consider it for the future, particularly if we have more of these to add.  I think we originally had keys mapped as `PRIME` in the `service_item_param_keys` table that would be passed by the prime as part of the payment request, but that sort of changed and probably isn't used in the same way anymore.

pkg/models/service_item_param_key.go
```diff
@@ -146,6 +146,10 @@ const (
 	ServiceItemParamNameZipSITOriginHHGActualAddress ServiceItemParamName = "ZipSITOriginHHGActualAddress"
 	// ServiceItemParamNameZipSITOriginHHGOriginalAddress is the param key name ZipSITOriginHHGOriginalAddress
 	ServiceItemParamNameZipSITOriginHHGOriginalAddress ServiceItemParamName = "ZipSITOriginHHGOriginalAddress"
+	// ServiceItemParamNameSITPaymentRequestStart is the param key name SITPaymentRequestStart
+	ServiceItemParamNameSITPaymentRequestStart ServiceItemParamName = "SITPaymentRequestStart"
+	// ServiceItemParamNameSITPaymentRequestEnd is the param key name SITPaymentRequestEnd
+	ServiceItemParamNameSITPaymentRequestEnd ServiceItemParamName = "SITPaymentRequestEnd"
```
reggieriser, line 152: Nit: I think we've tried to keep this list (and the other two below) alphabetical.

reggieriser: One more note ... there's also a spot to add these to the list in the YAML files as well.  [Here are the spots](https://github.com/transcom/mymove/pull/7248/commits/9a1be43e233a4267f5f0ea6551dfd0427e88cc5f) that I changed in another PR that added params.

migrations/app/schema/20210831214156_add_sit_pr_start_and_end_dates.up.sql
```diff
@@ -0,0 +1,15 @@
+INSERT INTO service_item_param_keys
+(id,key,description,type,origin,created_at,updated_at)
+VALUES
+('76452034-86f5-4932-a0e0-a977454e1e80','SITPaymentRequestStart', 'Start of time period being billed for SIT additional days', 'DATE', 'PRIME', now(), now()),
+('d7db37e7-a92b-4612-ab7c-3647e37dc8bc','SITPaymentRequestEnd', 'End of time period being billed for SIT additional days', 'DATE', 'PRIME', now(), now());
+
+
+INSERT INTO service_params
+(id,service_id,service_item_param_key_id,created_at,updated_at)
```
reggieriser, line 9: Just noting what I mentioned in Slack ... there's a new `is_optional` param that defaults to `false` on the `service_params` table.  As long as both these params are required for `DOASIT` and `DDASIT`, then this migration is good as written.

pkg/models/service_item_param_key.go
mboren: Ok, I'll rearrange the lists and add to the YAML files.

pkg/handlers/primeapi/allowed_payment_service_item_params.go
mboren: I agree that having it in the DB would be a good alternative. I don't have any idea if we're going to pass more parameters in this fashion in the future, so it's hard know what to design for.

migrations/app/schema/20210831214156_add_sit_pr_start_and_end_dates.up.sql
mboren: Thanks for bringing this up again, I meant to ask about it last week and forgot.
These params are required when creating a payment request for `DOASIT` and `DDASIT`, but not when creating the service items.
I'm not sure what the correct value of `is_optional` is in this case, but it doesn't seem to affect anything I've done when testing. `is_optional` being false doesn't seem to prevent me from creating the service items without the params (maybe because they get created automatically when you create first day SIT?).

reggieriser: If everything is showing up in the DB as expected, then I think you're OK for now.  Right now, I believe the `is_optional` value just controls whether a lookup writes a param to the DB if it's empty (an optional one won't, but a required one will).  It also is used in validation of params before we attempt to price; if the param is required for the service item but is empty, you'll get a validation error.

reggieriser (approved):
Thanks for making the additional changes!

PR #7306, opened 2021-09-03
Add in git-attributes for generated files;

rogeruiz:
## Description

It's best practice to treat generated files as if they were binary files
so that they don't show up when you perform a `git diff` on changed
files.

## Reviewer Notes

n/a

## Setup

Make a change to any of the files in `pkg/gen` or any of the Yaml files in `swagger/` and then run `git diff` and you will see that they do not produce a visual diff and Git reports back that they are binary files.

## References

* [See other PR](https://github.com/transcom/mymove/pull/7145) for this change

## Screenshot

5locd9

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+# README: https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes
+
+# Generated Golang files by `make server_generate`
+pkg/gen/**/* binary
+
+# Generated Swagger files by `yarn build-redoc`
+swagger/*.yaml binary


.gitattributes
```diff
@@ -0,0 +1,7 @@
+# README: https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes
+
+# Generated Golang files by `make server_generate`
+pkg/gen/**/* binary
+
+# Generated Swagger files by `yarn build-redoc`
+swagger/*.yaml binary
```
sandy-wright, line 7: _nice_

sandy-wright (approved):
This looks good to me, yay! Thanks for adding the soon-to-be-generated swagger files too 😄

PR #7317, opened 2021-09-07
[MB-8898] Recalculate payment request service

reggieriser:
## Description

This PR creates a service to recalculate payment requests.  See this [Slack thread](https://ustcdp3.slack.com/archives/C021XEC7TV4/p1630100880000300?thread_ts=1628100839.073300&cid=C021XEC7TV4) for more details.

## Reviewer Notes

- I've added a new `DEPRECATED` status for payment requests, and we set the old payment request's status to `DEPRECATED` after it has been recalculated in a new payment request.  See this [Slack thread](https://ustcdp3.slack.com/archives/CP6F568DC/p1630687118113800) for more details.
- I had to revisit some previously committed test data setup that was not properly associating a service type to an MTO service item because I wanted to use that functionality in this PR.

## Setup

Currently, there is not a way to execute this from the app, but running `make server_test` will exercise the recalculating service.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-8898) for this change


pkg/services/payment_request/payment_request_repricer.go
```diff
@@ -0,0 +1,185 @@
...
+	sort.SliceStable(newPaymentServiceItems, func(i, j int) bool {
+		return newPaymentServiceItems[i].MTOServiceItem.ReService.Priority < newPaymentServiceItems[j].MTOServiceItem.ReService.Priority
+	})
+
+	newPaymentRequest.PaymentServiceItems = newPaymentServiceItems
+
+	return newPaymentRequest
+}
+
+func updateOldPaymentRequestStatus(appCtx appcontext.AppContext, oldPaymentRequest *models.PaymentRequest) error {
```
jacquelineIO, line 121: This is good, but also wondering if this should be extracted to a generic payment request that will set the status with guards. Quasi state machine. Also might be overkill, but just a thought.

pkg/services/payment_request/payment_request_repricer.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+		if verrs.HasAny() {
+			return services.NewInvalidInputError(newPaymentRequest.ID, err, verrs, "failed to validate proof of service doc")
+		}
+
+		for _, primeUpload := range proofOfServiceDoc.PrimeUploads {
+			newPrimeUpload := models.PrimeUpload{
+				ProofOfServiceDocID: newProofOfServiceDoc.ID,
+				ContractorID:        primeUpload.ContractorID,
+				UploadID:            primeUpload.UploadID,
```
jacquelineIO, line 153: This UploadID points to the actual data of the uploaded doc. So as this is written, we'd have multiple PrimeUploads pointing to the same Upload. If a PrimeUpload were to be deleted would we run into an issue? It could be that we never delete so we don't care. So I'm curious if this was intentional or if we should be creating a new Upload?  For a picture reference see https://miro.com/app/board/o9J_kuh516s=/

pkg/services/payment_request/payment_request_repricer.go
```diff
@@ -0,0 +1,185 @@
...
+
+		newPaymentRequest.ProofOfServiceDocs = append(newPaymentRequest.ProofOfServiceDocs, newProofOfServiceDoc)
+	}
+
+	return nil
+}
+
+// linkNewToOldPaymentRequest links a new payment request to the old payment request that was repriced.
+func linkNewToOldPaymentRequest(appCtx appcontext.AppContext, newPaymentRequest *models.PaymentRequest, oldPaymentRequest *models.PaymentRequest) error {
+	newPaymentRequest.RepricedPaymentRequestID = &oldPaymentRequest.ID
```
jacquelineIO, line 175: The name repriced throws me off here. I understand what it is from looking at the code, but `RepricedPaymentRequest` makes me feel like it should be holding the ID of the new payment request, but it is holding the payment of the old one. 

The oldPaymentRequest.RepricedPaymentRequestID variable could instead point to the new one. Or maybe there is a better name. I don't have a preference here. Just something to consider.

jacquelineIO (changes requested):
This is good for the payment request re-price. What will be calling this service? I was kind of thinking the MTO Shipment ID would be passed down which would be 1 layer above this. The reason being is that the weight is tied to the shipment and then the payment requests would be pulled and then this service is called. This might already be in the backlog as the next piece. Just want to make sure it is somewhere. 

The only thing that needs a real decision to keep or change (from my review) is around the copying of the `UploadID` which I left a comment for.

reggieriser:
> This is good for the payment request re-price. What will be calling this service? I was kind of thinking the MTO Shipment ID would be passed down which would be 1 layer above this. The reason being is that the weight is tied to the shipment and then the payment requests would be pulled and then this service is called. This might already be in the backlog as the next piece. Just want to make sure it is somewhere.

I think what you're describing is captured in [MB-9339](https://dp3.atlassian.net/browse/MB-9339).  That might end up being in its own service, but either way I expect that the service in this PR will be used as a building block for it.  I was hoping to keep this PR as a generic way to rebuild a payment request with current inputs, regardless of what triggers it now or in the future.

pkg/services/payment_request/payment_request_repricer.go
reggieriser: I just noticed that there is an existing `UpdatePaymentRequestStatus` service that I may be able to leverage.  I do have some guard rails in this PR that checks the status of the existing payment request to make sure it's `PENDING` before allowing it to be recalculated.  I should probably add something similar to the other service if I use it.

pkg/services/payment_request/payment_request_repricer.go
reggieriser: I'm not sure if we can delete the old upload once the old payment request has been deprecated.  If I made a new `Upload` record for each doc as well, would I have to also duplicate the uploaded file as well on S3 (presuming it's not deleted itself)?  I don't recall all the details of how uploads work off the top of my head.

pkg/services/payment_request/payment_request_repricer.go
reggieriser: I'll work on a new name.  I think I prefer having the new payment request point to the old one so we can work down a chain from current payment request to older deprecated payment requests as needed.

pkg/services/payment_request/payment_request_repricer.go
jacquelineIO: Yes @reggieriser  you'd have to save a new file to S3. I agree, I think we won't be deleting. At some point we will be archiving things. We don't know what that looks like yet. Also once we are plugged into IGC we could then delete and that might be the archive process (but that's a guess).

jacquelineIO:
> I think what you're describing is captured in [MB-9339](https://dp3.atlassian.net/browse/MB-9339). That might end up being in its own service, but either way I expect that the service in this PR will be used as a building block for it. I was hoping to keep this PR as a generic way to rebuild a payment request with current inputs, regardless of what triggers it now or in the future.

I agree with the scope of this PR.

pkg/services/payment_request/payment_request_repricer.go
reggieriser: I'm now using the service mentioned above.

pkg/services/payment_request/payment_request_repricer.go
reggieriser: Sounds like we'll have to research this more.  I'll try to look into it a little and see what we may be up against, but we may want to make this part of a separate story.

pkg/services/payment_request/payment_request_repricer.go
reggieriser: I changed the field name to `recalculation_of_payment_request_id` to make it more obvious what it contains ([see Slack thread](https://ustcdp3.slack.com/archives/CP6F568DC/p1631041087129800)).

reggieriser:
@jacquelineIO I just pushed up updates to address some of your comments.  I also introduced a `DEPRECATED` status for payment requests and renamed files, methods, functions, etc. to use "recalculate" instead of "reprice".

pkg/services/payment_request/payment_request_repricer.go
gidjin: I'd lean towards the lower effort of not duplicating files on S3 or in the uploads table, at least until we need deleting / archiving. I don't think it's worth duplicating the upload record if we are not duplicating the S3 file.

jacquelineIO (approved):
lgtm. thanks for updates

pkg/services/payment_request/payment_request_repricer.go
jacquelineIO: I agree @gidjin but want to make sure we have the right usage of Upload structurally. Re-use probably shouldn't be at the `Upload` level, it should be at the `PrimeUpload` or `ProofOfService`.

migrations/app/schema/20210903211741_recalculation_payment_request.up.sql
```diff
@@ -0,0 +1,14 @@
+-- Add column that links a new payment request to the older payment request it recalculated, if applicable.
+ALTER TABLE payment_requests
+	ADD COLUMN recalculation_of_payment_request_id uuid
+		CONSTRAINT payment_requests_recalculation_of_payment_request_id_fkey REFERENCES payment_requests;
+
+COMMENT ON COLUMN payment_requests.recalculation_of_payment_request_id IS 'Link to the older payment request that was recalculated to form this payment request (if applicable).';
```
sandy-wright, line 6: Thanks for adding this!

migrations/app/schema/20210903211741_recalculation_payment_request.up.sql
```diff
@@ -0,0 +1,14 @@
...
+
+CREATE INDEX payment_requests_recalculation_of_payment_request_id_idx ON payment_requests (recalculation_of_payment_request_id);
+
+-- Add DEPRECATED status for a payment request.
+ALTER TYPE payment_request_status
+	ADD VALUE 'DEPRECATED';
+
+COMMENT ON COLUMN payment_requests.status IS 'Track the status of the payment request through the system. PENDING by default at creation. Options: PENDING, REVIEWED, REVIEWED_AND_ALL_SERVICE_ITEMS_REJECTED, SENT_TO_GEX, RECEIVED_BY_GEX, PAID, EDI_ERROR, DEPRECATED';
```
sandy-wright, line 14: I'm starting to feel like we shouldn't list all the options in these comments, just because it's fairly burdensome to keep it up-to-date. But thanks for updating it anyway!

sandy-wright (approved):
Reviewed the migration and model changes - LGTM!

pkg/services/payment_request/payment_request_repricer.go
reggieriser: I've added [a story](https://dp3.atlassian.net/browse/MB-9417) to further investigate and address this concern.

PR #7342, opened 2021-09-10
MB-9361 Refactor service item updater

monfresh:
## Description

The original motivation for this refactor was to reuse the logic for
determining whether or not a move's status should go to Approved or
Approvals Requested, based on whether or not it has unreviewed service
items and/or unacknowledged amended orders.

While I was in the `mto_service_item_updater file`, I took the
opportunity to refactor the `UpdateMTOServiceItemStatus` function to
follow the patterns I've used in other services:

- Make the main function as short as possible
- Break the various steps into well-named, separate functions
- Perform multiple DB updates inside a transaction
- Use the validator pattern for any checks that need to happen

## Reviewer Notes

More things we can refactor:

- Move the logic in `updateServiceItem` for preventing a nil rejection reason into a validation
- Rename `UpdateMTOServiceItemStatus` to `ApproveOrRejectServiceItem`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9361) for this change

### DIFF
--- a/pkg/services/mto_service_item/mto_service_item_creator.go
+++ b/pkg/services/mto_service_item/mto_service_item_creator.go
@@ -275,7 +275,7 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(appCtx appcontext.AppContex
 			return err
 		}
 
-		if err = o.approveMoveOrRequestApproval(txnAppCtx, move.Orders, move); err != nil {
+		if err = ApproveMoveOrRequestApproval(txnAppCtx, o.moveRouter, move.Orders, move); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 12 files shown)


pkg/services/mto_service_item/mto_service_item_creator.go
```diff
@@ -327,6 +293,43 @@ func (o *mtoServiceItemCreator) CreateMTOServiceItem(appCtx appcontext.AppContex
...
+	if err != nil {
+		return err
+	}
+
+	verrs, err := appCtx.DB().ValidateAndUpdate(&move)
+
+	return HandleError(move.ID, verrs, err)
+}
+
+// MoveShouldBeApproved checks if the TOO no longer has any tasks requiring action.
```
Ryan-Koch, line 316: I like the choice to go with a function like this, nice and easy to read/understand 🙌

Ryan-Koch:
Overall looking good! The one thing I was thinking about is whether it might be worth consolidating the logic around checking whether the state should change for a parent move to something publicly consumable in the Move service. What do you think? We can also talk about it when we pick up the work next week for the SIT extension impact on Move state if that's more beneficial.

monfresh:
> The one thing I was thinking about is whether it might be worth consolidating the logic around checking whether the state should change for a parent move to something publicly consumable in the Move service. 

Having the logic in the Move service probably makes sense. Could you be more specific as to which functions you think should be moved?

Ryan-Koch:
> > The one thing I was thinking about is whether it might be worth consolidating the logic around checking whether the state should change for a parent move to something publicly consumable in the Move service.
> 
> Having the logic in the Move service probably makes sense. Could you be more specific as to which functions you think should be moved?

Sure thing. As an example there's the logic we're using [here](https://github.com/transcom/mymove/blob/f29be5b9532cd67a4aec2f342de09ccfe0e684d2/pkg/services/order/excess_weight_risk_manager.go#L173) in the order service which makes use of a series of public functions representing each of the 'can I approve or not' rules. We wind up needing to do the same thing [here](https://github.com/transcom/mymove/blob/f29be5b9532cd67a4aec2f342de09ccfe0e684d2/pkg/services/mto_service_item/mto_service_item_creator.go#L317) in the MTO Service Item service (though I think you have it so that one is using the same public functions). I was thinking it might be helpful to have the determination of whether there's something to prevent approval together in one location, the theory being it might make it easier to new checks.

monfresh:
Gotcha! Yeah, that makes sense. I'll need some more time to think about the best way to do that, so maybe next week would be better.

Ryan-Koch (approved):

PR #7352, opened 2021-09-13
[MB-9111] too can submit and auto-approve SIT extension for customer

hannaliebl:
## Description

Add an endpoint for the TOO to create a SIT extension and automatically approve it. Additionally, once created, the SIT allowance on the shipment should be update to reflect this new approved SIT extension.

## Reviewer Notes

There is room to integrate this work with what @NamibiaTorres is doing this week for the Prime API side - she is also adding the ability to create a SIT extension from the Prime side, but I think that will need to wait until another sprint at this point.

## Setup

```sh
echo "Code goes here"
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9111) for this change


pkg/services/mto_shipment/sit_extension_creator_test.go
```diff
@@ -0,0 +1,89 @@
...
+		eTag := ""
+
+		_, err := sitExtensionCreator.CreateApprovedSITExtension(suite.TestAppContext(), &sitExtensionToSave, mtoShipment.ID, eTag)
+
+		suite.Error(err)
+		suite.IsType(services.PreconditionFailedError{}, err)
+		suite.Contains(err.Error(), mtoShipment.ID.String())
+	})
+
+	suite.T().Run("Creates an approved SIT extension when all fields are valid and updates the shipment's SIT days allowance", func(t *testing.T) {
```
hannaliebl, line 56: This is the test that is not working right now.

pkg/services/mto_shipment/sit_extension_creator_test.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		updatedShipment, err := sitExtensionCreator.CreateApprovedSITExtension(suite.TestAppContext(), &sitExtensionToSave, mtoShipment.ID, eTag)
+		suite.NoError(err)
+
+		var shipmentInDB models.MTOShipment
+		err = suite.DB().Find(&shipmentInDB, mtoShipment.ID)
+		suite.NoError(err)
+		var sitExtensionInDB models.SITExtension
+		err = suite.DB().Find(&sitExtensionInDB, updatedShipment.ID)
```
monfresh, line 79: This is looking up a SIT Extension using a shipment ID. It's returning no rows found because there isn't a SIT Extension with an `id` equal to the shipment's ID. The service is returning a shipment, not a SIT Extension. 

One way to test this is to verify on line 59 that when the shipment is first created, it does not have any SIT Extensions, and then here on line 79, you would check that it does have one now. And then you could verify the attributes like you have below by looking up the first SIT Extension. Something like this:

```golang
err = suite.DB().First(&sitExtensionInDB)
```
We could also make sure only one was created by checking the length of all the SIT Extensions:
```golang
var allSITExtensions []models.SITExtension
err = suite.DB().All(&allSITExtensions)
suite.NoError(err)
suite.Equal(2, len(allSITExtensions))
```

pkg/services/mto_shipment/sit_extension_creator_test.go
```diff
@@ -0,0 +1,89 @@
...
+		suite.NoError(err)
+
+		suite.Equal(mtoShipment.ID.String(), updatedShipment.ID.String())
+		suite.Equal(requestedDays, *updatedShipment.SITDaysAllowance)
+		suite.Equal(requestedDays, *sitExtensionInDB.ApprovedDays)
+		suite.Equal(requestedDays, &sitExtensionInDB.RequestedDays)
```
monfresh, line 85: The ampersand is not needed here. It causes the test to fail. Removing it fixes the test.

pkg/handlers/ghcapi/mto_shipment.go
```diff
@@ -809,3 +809,52 @@ func (h DenySITExtensionHandler) Handle(params shipmentops.DenySitExtensionParam
...
+	services.SITExtensionCreatorAsTOO
+}
+
+// Handle creates the approved SIT extension
+func (h CreateSITExtensionAsTOOHandler) Handle(params shipmentops.CreateSitExtensionAsTOOParams) middleware.Responder {
+	logger := h.LoggerFromRequest(params.HTTPRequest)
+	appCtx := appcontext.NewAppContext(h.DB(), logger)
+	payload := params.Body
+
+	if payload == nil {
```
duncan-truss, line 825: I'm not sure we need to validate this, the swagger client would flag required parameters and respond with an error prior to hitting this handler is my understanding.  You can set `required: true` on the Body parameter for this request.  It's not hurting anything but unlikely to be reachable.

pkg/handlers/ghcapi/internal/payloads/payload_to_model.go
```diff
@@ -115,6 +115,28 @@ func AddressModel(address *ghcmessages.Address) *models.Address {
...
+	}
+
+	var approvedDays *int
+	ad := int(*sitExtension.ApprovedDays)
+	approvedDays = &ad
+
+	model := &models.SITExtension{
+		MTOShipmentID: uuid.FromStringOrNil(sitExtension.ShipmentID.String()),
+		RequestReason: models.SITExtensionRequestReason(*sitExtension.RequestReason),
+		RequestedDays: int(*sitExtension.ApprovedDays),
```
duncan-truss, line 131: I thought this was only set by the Prime, it could be confusing if it has a value here.  I see it is required in the model though.

pkg/handlers/ghcapi/internal/payloads/payload_to_model.go
```diff
@@ -115,6 +115,28 @@ func AddressModel(address *ghcmessages.Address) *models.Address {
...
 }
 
+// ApprovedSITExtensionFromCreate model
+func ApprovedSITExtensionFromCreate(sitExtension *ghcmessages.CreateSITExtensionAsTOO) *models.SITExtension {
+	if sitExtension == nil {
+		return nil
+	}
+
+	var approvedDays *int
+	ad := int(*sitExtension.ApprovedDays)
```
duncan-truss, line 125: You really only need this line and then on line 133 you can write `ApprovedDays: &ad` if you want to cut down on extra steps.

pkg/handlers/ghcapi/mto_shipment.go
```diff
@@ -809,3 +809,52 @@ func (h DenySITExtensionHandler) Handle(params shipmentops.DenySitExtensionParam
...
+		logger.Error("Invalid mto shipment: params Body is nil")
+		return shipmentops.NewCreateSitExtensionAsTOOBadRequest()
+	}
+
+	sitExtension := payloads.ApprovedSITExtensionFromCreate(payload)
+	shipmentID := uuid.FromStringOrNil(string(params.ShipmentID))
```
duncan-truss, line 831: I think you parsed this once in your payload function above if you wanted to rely on that you could.

pkg/handlers/ghcapi/mto_shipment_test.go
```diff
@@ -1572,6 +1572,47 @@ func (suite *HandlerSuite) TestDenySITExtensionHandler() {
 	})
 }
 
+func (suite *HandlerSuite) CreateSITExtensionAsTOO() {
+	suite.Run("Returns 200, creates new SIT extension, and updates SIT days allowance when validations pass", func() {
+		mtoShipment := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+			MTOShipment: models.MTOShipment{},
```
duncan-truss, line 1578: It is probably good to test when a shipment has no SITDaysAllowance yet because that's true of existing moves unless we are going to set defaults or backfill all old data in the upcoming sprints.  Can we also test when SITDaysAllowance has a value?  You should be able to set that here in the assertion.

pkg/handlers/ghcapi/mto_shipment.go
```diff
@@ -809,3 +809,52 @@ func (h DenySITExtensionHandler) Handle(params shipmentops.DenySitExtensionParam
...
+	handlers.HandlerContext
+	services.SITExtensionCreatorAsTOO
+}
+
+// Handle creates the approved SIT extension
+func (h CreateSITExtensionAsTOOHandler) Handle(params shipmentops.CreateSitExtensionAsTOOParams) middleware.Responder {
+	logger := h.LoggerFromRequest(params.HTTPRequest)
+	appCtx := appcontext.NewAppContext(h.DB(), logger)
+	payload := params.Body
+
```
duncan-truss, line 824: We should likely add authorization here for a user with the TOO role to create a SITExtension

pkg/services/mto_shipment/sit_extension_creator_as_too.go
```diff
@@ -0,0 +1,103 @@
...
+type sitExtensionCreatorAsTOO struct {
+}
+
+// NewCreateSITExtensionAsTOO creates a new struct with the service dependencies
+func NewCreateSITExtensionAsTOO() services.SITExtensionCreatorAsTOO {
+	return &sitExtensionCreatorAsTOO{}
+}
+
+// CreateSITExtensionAsTOO creates a SIT Extension with a status of APPROVED and updates the MTO Shipment's SIT days allowance
+func (f *sitExtensionCreatorAsTOO) CreateSITExtensionAsTOO(appCtx appcontext.AppContext, sitExtension *models.SITExtension, shipmentID uuid.UUID, eTag string) (*models.MTOShipment, error) {
```
duncan-truss, line 25: We could get into a situation here where the shipmentID param in the path doesn't match the one in the body.

pkg/services/mto_shipment/sit_extension_creator_as_too.go
```diff
@@ -0,0 +1,103 @@
...
+
+	var returnedShipment models.MTOShipment
+
+	transactionError := appCtx.NewTransaction(func(txnAppCtx appcontext.AppContext) error {
+		verrs, err := txnAppCtx.DB().ValidateAndCreate(sitExtension)
+		if e := f.handleError(sitExtension.ID, verrs, err); e != nil {
+			return e
+		}
+
+		updatedShipment, err := f.updateSitDaysAllowance(txnAppCtx, *shipment, *sitExtension.ApprovedDays)
```
duncan-truss, line 44: You could make this `returnedShipment, err = f.updateSitDaysAllowance(txnAppCtx, *shipment, *sitExtension.ApprovedDays)` without the `:=` if you want to avoid creating a new updatedShipment variable.  You'd just need to make returnedShipment a pointer type `*models.MTOShipment` when you declare it.

pkg/services/mto_shipment/sit_extension_creator_as_too.go
```diff
@@ -0,0 +1,103 @@
...
+	})
+
+	if transactionError != nil {
+		return nil, transactionError
+	}
+
+	return &returnedShipment, nil
+}
+
+func (f *sitExtensionCreatorAsTOO) updateSitDaysAllowance(appCtx appcontext.AppContext, shipment models.MTOShipment, approvedDays int) (*models.MTOShipment, error) {
```
duncan-truss, line 61: Does this exist elsewhere in the previous service you created? We should find a way to consolidate eventually if they are the same.

swagger/ghc.yaml
```diff
@@ -909,6 +909,47 @@ paths:
...
+          name: shipmentID
+          required: true
+          format: uuid
+          type: string
+        - in: body
+          name: body
+          schema:
+            $ref: '#/definitions/CreateSITExtensionAsTOO'
+        - in: header
+          description: We want the shipment's eTag rather than the SIT extension eTag as the SIT extension is always associated with a shipment
```
duncan-truss, line 936: Yay thanks for this clear comment 🎉

swagger/ghc.yaml
```diff
@@ -3233,6 +3274,29 @@ definitions:
         type: string
         example: Denied this extension as it does not match the criteria
         x-nullable: true
+  CreateSITExtensionAsTOO:
+    properties:
+      shipmentID:
+        example: 1f2270c7-7166-40ae-981e-b200ebdf3054
+        format: uuid
+        type: string
+      requestReason:
```
duncan-truss, line 3283: I think we should make this an enum of possible string values if it doesn't exist already.  I think we can share it eventually between the SITExtension definition and this payload but I know that makes you change the types everywhere.

pkg/handlers/ghcapi/internal/payloads/payload_to_model.go
hannaliebl: Hmm, it's unclear to me what this should be, since it's not really going through the prime at all. I can ask in the feature channel.

pkg/handlers/ghcapi/mto_shipment_test.go
```diff
@@ -1572,6 +1572,47 @@ func (suite *HandlerSuite) TestDenySITExtensionHandler() {
...
+			HTTPRequest: req,
+			IfMatch:     eTag,
+			Body: &ghcmessages.CreateSITExtensionAsTOO{
+				ApprovedDays:  &approvedDays,
+				OfficeRemarks: &officeRemarks,
+				RequestReason: &requestReason,
+				ShipmentID:    *handlers.FmtUUID(mtoShipment.ID),
+			},
+			ShipmentID: *handlers.FmtUUID(mtoShipment.ID),
+		}
```
monfresh, line 1605: Can we also validate the Body please, per our [handler testing guide](https://github.com/transcom/mymove/wiki/Testing-Handlers)?

```golang
suite.NoError(createParams.Body.Validate(strfmt.Default))
```

pkg/services/mto_shipment/sit_extension_creator_as_too.go
monfresh: Agreed. Having the shipment ID in the path should be enough. We'll have access to it from the Params, like you did in the handler, and then you can pass it around wherever it's needed.

pkg/services/mto_shipment/sit_extension_creator_as_too.go
hannaliebl: Yeah, I wasn't sure how to share functions like this when they are attached to their own service objects.

pkg/handlers/ghcapi/internal/payloads/payload_to_model.go
hannaliebl: Prakit said we should default to requested being the same as approved.

pkg/handlers/ghcapi/mto_shipment.go
```diff
@@ -855,6 +846,17 @@ func (h CreateSITExtensionAsTOOHandler) Handle(params shipmentops.CreateSitExten
...
 
+	sitExtension := payloads.ApprovedSITExtensionFromCreate(payload)
+	shipmentID := sitExtension.MTOShipmentID
+	shipment, err := h.SITExtensionCreatorAsTOO.CreateSITExtensionAsTOO(appCtx, sitExtension, shipmentID, params.IfMatch)
+	if err != nil {
+		return handleError(err)
+	}
+
+	if !session.IsOfficeUser() || !session.Roles.HasRole(roles.RoleTypeTOO) {
+		return handleError(services.NewForbiddenError("is not a TOO"))
```
duncan-truss, line 857: This looks good, make sure to include the `services.ForbiddenError` type in your above handleError func so it doesn't get treated like a 500 Internal Server error.  And add the 403 response the your ghc swagger endpoint.

duncan-truss (approved):
LGTM, I just had the one last authorization comment.

pkg/services/mto_shipment/sit_extension_creator_as_too.go
hannaliebl: Ok, that makes sense, I did it here: 643b759

PR #7354, opened 2021-09-13
[MB-9418] Add support API endpoint for recalculating payment request

reggieriser:
## Description

This PR adds a support API endpoint for recalculating a payment request (the recalculation service was added in #7317).  

## Reviewer Notes

This is landing right after @sandy-wright 's PR #7145 that refactored the swagger directory and code generation process, so please review the swagger-related files to ensure I am changing them appropriately.

## Setup

First, build the server and set up your local DB with E2E data:
- `make clean server_build db_dev_reset db_dev_e2e_populate`

You can then use the `prime-api-client` to call the support endpoint locally.  Choose a payment request ID from the E2E data.  Note that the payment request must be in `PENDING` state (if not, an error will be reported).  Some of the E2E payment requests may not be fully set up with all prerequisite data, so it's also possible to get an error on some of the `PENDING` payment requests.

Here's one that should work out of the box (just replace the ID below if you want to try a different payment request):

- `prime-api-client --insecure support-recalculate-payment-request --id ea945ab7-099a-4819-82de-6968efe131dc | jq`

The top-level payment request information should be returned from the call above.  There should be a reference to the old payment request ID in the JSON.  Also, look at your DB and verify that the old payment request is now in `DEPRECATED` status.  You can also spin up the office UI if you'd like and find the payment requests there.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9418) for this change


pkg/services/payment_request/payment_request_recalculator.go
```diff
@@ -67,6 +67,14 @@ func (p *paymentRequestRecalculator) doRecalculate(appCtx appcontext.AppContext,
...
 	}
 
+	// Set the (now) old payment request's status.  Doing this before we recalculate in case the create
+	// payment request service needs to know that this will be deprecated if reaculation is successful.
+	oldPaymentRequest.Status = models.PaymentRequestStatusDeprecated
+	_, err = p.paymentRequestStatusUpdater.UpdatePaymentRequestStatus(appCtx, &oldPaymentRequest, oldPaymentRequestEtag)
+	if err != nil {
+		return nil, err
+	}
+
```
reggieriser, line 77: Just moving this up in the transaction because I anticipate we may need to know during a recalculation that a request is being `DEPRECATED`.  I previously was doing this after the new payment request was created.  It's the same result in the end.

cmd/prime-api-client/support/recalculate_payment_request.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	// Check the config before talking to the CAC
+	err := checkRecalculatePaymentRequestConfig(v, args, logger)
+	if err != nil {
+		logger.Fatal(err)
+	}
+
+	// Get the UUID that was passed in
+	paymentRequestID := v.GetString(utils.IDFlag)
```
jacquelineIO, line 63: Can we have a choice or either the UUID or the payment request number? I thought that's what we did with some other payment request support APIs. (Not necessary)

pkg/handlers/supportapi/payment_request_test.go
```diff
@@ -777,3 +777,110 @@ func (suite *HandlerSuite) TestProcessReviewedPaymentRequestsHandler() {
...
+	suite.T().Run("golden path", func(t *testing.T) {
+		samplePaymentRequest := models.PaymentRequest{
+			ID:                   uuid.Must(uuid.NewV4()),
+			MoveTaskOrderID:      uuid.Must(uuid.NewV4()),
+			Status:               models.PaymentRequestStatusPending,
+			PaymentRequestNumber: "1111-2222-1",
+			SequenceNumber:       1,
+		}
+
+		mockRecalculator := &mocks.PaymentRequestRecalculator{}
```
jacquelineIO, line 797: do we normally mock out the success/golden path? or is it that the service it tested already we don't care here?

pkg/services/payment_request/payment_request_recalculator.go
```diff
@@ -75,13 +83,6 @@ func (p *paymentRequestRecalculator) doRecalculate(appCtx appcontext.AppContext,
 		return nil, err // Just pass the error type from the PaymentRequestCreator.
```
jacquelineIO, line 83: should we try to set the deprecated pr back to the previous status, if we bail out. I also understand if at a certain point we are failing and something bigger is wrong and there is only so much error  handling that you can do.

jacquelineIO (approved):
lgtm. ran locally. Had a couple of questions, but nothing blocking ths PR.

pkg/services/payment_request/payment_request_recalculator.go
reggieriser: Since all of this is in a transaction block, everything should be rolled back automatically if an error is returned.  It shouldn't really matter from that perspective if we set the previous transaction to deprecated before or after we create the new payment request.  But it could matter that the new payment request creation process (and the lookups and pricers) knows that this other payment request should be considered deprecated -- which should be visible if it's working within the same transaction even thought it hasn't been committed yet.

cmd/prime-api-client/support/recalculate_payment_request.go
reggieriser: I looked and couldn't find an example of a `prime-api-client` command that accepted identifiers other than UUIDs (either for payment requests or moves).  The backend recalculate service would need to change to accept either too unless I wanted to start doing database calls in `prime-api-client` which we don't do yet (there'd be some setup involved to initiate a DB connection).  So, it looks like it could get involved.  If we decide we want this, let's make it another story and do it for all the payment request `prime-api-client` commands.

I did see that we accept a payment request number on the `generate-payment-request-edi` standalone utility.  But we don't on the `create-upload` utility.  🤷‍♂️

pkg/handlers/supportapi/payment_request_test.go
reggieriser: I felt this was a reasonable case to mock the service because it in theory is already tested, plus the setup to test a real payment request is pretty complicated and does a lot of DB work (making every test run slow down).  My goal here is to test all the paths of the handler code -- like making sure the success path translates between a model and the payload types.  That said, I dug into things a little more and made a few changes to tighten up the mock and related assertions.

pkg/handlers/supportapi/payment_request_test.go
```diff
@@ -811,6 +816,8 @@ func (suite *HandlerSuite) TestRecalculatePaymentRequestHandler() {
 		}
 		response := handler.Handle(params)
 
+		mockRecalculator.AssertExpectations(t)
```
jacquelineIO, line 819: What is this?

pkg/handlers/supportapi/payment_request_test.go
```diff
@@ -819,6 +826,9 @@ func (suite *HandlerSuite) TestRecalculatePaymentRequestHandler() {
 			suite.Equal(samplePaymentRequest.Status.String(), string(payload.Status))
 			suite.Equal(samplePaymentRequest.PaymentRequestNumber, payload.PaymentRequestNumber)
 			// SequenceNumber is not on payload at all as it's an internal representation.
+			if suite.NotNil(payload.RecalculationOfPaymentRequestID.String()) {
```
jacquelineIO, line 829: are you expecting this to be Nil? Is this in a loop?

pkg/handlers/supportapi/payment_request_test.go
reggieriser: It's just making sure that the things I expected to be called (based upon the "On" with those specific params/return) were actually called.  More info here:
- https://github.com/stretchr/testify#mock-package
- https://pkg.go.dev/github.com/stretchr/testify/mock#Mock.AssertExpectations

pkg/handlers/supportapi/payment_request_test.go
reggieriser: I had a copy/paste error and should have passed `payload.RecalculationOfPaymentRequestID` rather than `payload.RecalculationOfPaymentRequestID.String()` to `NotNil`.  The `payload.RecalculationOfPaymentRequestID` could potentially be `nil` if things weren't working right.  By doing the check first, I'm avoiding a panic by trying to call `String()` on a `nil` pointer just below (which could make it harder to understand the root cause of a failed test).

pkg/handlers/supportapi/payment_request_test.go
```diff
@@ -826,7 +826,7 @@ func (suite *HandlerSuite) TestRecalculatePaymentRequestHandler() {
 			suite.Equal(samplePaymentRequest.Status.String(), string(payload.Status))
 			suite.Equal(samplePaymentRequest.PaymentRequestNumber, payload.PaymentRequestNumber)
 			// SequenceNumber is not on payload at all as it's an internal representation.
-			if suite.NotNil(payload.RecalculationOfPaymentRequestID.String()) {
+			if suite.NotNil(payload.RecalculationOfPaymentRequestID) {
```
jacquelineIO, line 829: I guess my question is, should there be something in the `else`?

reggieriser: I don't think so.  If `payload.RecalculationOfPaymentRequestID` is `nil`, then it should display an error message about that and not try to compare the string version of the ID (since there isn't one, and I believe a panic would occur if we went there).  If it's non-nil, then it compares the strings.

PR #7368, opened 2021-09-15
MB-9243 Create default sit days allowance for new shipments

felipe-lee:
## Description

We want to have a default SIT days allowance set on new HHG shipments. This should vary by customer type, but since we only have one customer type currently, I just left comments about it.

## Reviewer Notes

Based on this [slack thread](https://ustcdp3.slack.com/archives/C024LFTCQKE/p1631728090093100), the default is being added to all mto_shipments. In the future we'll need to tweak this logic to do it based on customer type.

A question came up ([slack thread](https://ustcdp3.slack.com/archives/CTNTFJSBA/p1631643730153300)) about adding a migration for existing data so I'm going to add one, but not sure if this is the best way to do it either. Also started [another related slack thread](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1631727764276800).

As I was working on this, I noticed the main test file I was updating had shared setup between the subtests, but only did it once so I moved it to a function that is now called at the beginning of each subtest. 

Related to the last point, I updated the `mto_shipment` services package to use per test transactions to bring it in line with newer standards. That's what the rest of the changes are for. Unfortunately, because we still have other packages that don't run in transactions and they truncate the full DB, there's no guarantee that some of the data we would expect from a fresh DB will be there (e.g. `ReServices`), so I've had to keep the creation of them in there. I did made those tests check first if they even needed to create them because if you have a fresh DB and only run tests within that package, you'll get a non-unique error if you just try to always create them.

## Setup

```sh
make server_test
```

You can also verify by creating a shipment through the internal, ghc, or prime APIs since they should all go through the same service object. 

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9243) for this change


pkg/services/mto_shipment/mto_shipment_creator_test.go
```diff
@@ -266,6 +355,50 @@ func (suite *MTOShipmentServiceSuite) TestCreateMTOShipment() {
...
+				suite.NoError(err)
+
+				if tt.expectNilSITDaysAllowance {
+					suite.Nil(createdShipment.SITDaysAllowance)
+				} else {
+					suite.Equal(tt.expectedSITDaysAllowance, *createdShipment.SITDaysAllowance)
+				}
+			})
+		}
+	})
```
felipe-lee, line 401: This is the main bit relevant to the ticket. The rest of the file is to update the standards of this file.

pkg/services/mto_shipment/validation_test.go
```diff
@@ -40,15 +40,17 @@ func (suite *MTOShipmentServiceSuite) TestUpdateValidations() {
 	})
 
 	suite.Run("checkAvailToPrime", func() {
+		appCtx := suite.TestAppContext()
```
felipe-lee, line 43: The changes in this file had to do with per test transactions as well because I needed to make sure the same connection was used for generating the data as for the test itself since it queries from the DB.

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -154,6 +154,12 @@ func (f mtoShipmentCreator) CreateMTOShipment(appCtx appcontext.AppContext, ship
 			shipment.Status = models.MTOShipmentStatusDraft
 		}
 
+		// Assign default SITDaysAllowance based on customer type and shipment type...but we only have service members
+		// right now. Once we introduce more, this logic will have to change.
+		if shipment.ShipmentType == models.MTOShipmentTypeHHG {
+			shipment.SITDaysAllowance = &models.DefaultServiceMemberSITDaysAllowance
+		}
+
```
felipe-lee, line 162: Main work for this ticket.

reggieriser (approved):
Reviewing migration for DB reviewers ... I think it's fine (but please get an overall PR review from another reviewer).  I was able to run migrations successfully.  It should only affect moves that are in the system at the moment this migration is run; future moves won't be affected because the migration should never run again.  But looks like you set that in code now, so that's how they'll get the default.  I'll follow up in one of the Slack threads you linked to.

One question ... I see a bunch of other shipment types with `HHG` in their title (e.g., `HHG_LONGHAUL_DOMESTIC`).  Does this apply to any of those?

pkg/models/mto_shipments.go
```diff
@@ -45,6 +45,13 @@ const (
 	MTOShipmentTypeBoatTowAway MTOShipmentType = "BOAT_TOW_AWAY"
 )
 
+// Other values will likely be added to this once we deal with different types of customers.
+// Can't be constants because we need pointers to these values and you can't get a pointer to a const.
+var (
+	// DefaultServiceMemberSITDaysAllowance is the default number of SIT days a service member is allowed
+	DefaultServiceMemberSITDaysAllowance = 90
```
jacquelineIO, line 52: Why do we have this and the mto_shipments.sit_days_allowance in the database?

felipe-lee: From my understanding, in the future we'll have other customer types that will have different default numbers.

felipe-lee:
> One question ... I see a bunch of other shipment types with `HHG` in their title (e.g., `HHG_LONGHAUL_DOMESTIC`). Does this apply to any of those?

@reggieriser Sorry, forgot to reply to your question. I asked in this [slack thread](https://ustcdp3.slack.com/archives/C024LFTCQKE/p1631901612108000?thread_ts=1631728090.093100&cid=C024LFTCQKE) and a decision was made to actually just do it for all shipment types so I've updated the code and the migration to reflect that.

pkg/models/mto_shipments.go
eamahanna: This is correct!

jacquelineIO: I still don't follow. Will there be more entries in the database? When do you use the `const` and when do you use the one in the database? Are these 2 the same (`DefaultServiceMemberSITDaysAllowance` and `mto_shipments.sit_days_allowance`)?

felipe-lee: You use the one in the DB (`mto_shipments.sit_days_allowance`) for any lookups and such. Tbh I have no idea where it's displayed or used, but anytime you need to do something with a shipment and need its SIT days allowance, you'll want to use the DB. The `const` here (`DefaultServiceMemberSITDaysAllowance`) is for the code to set the initial value that will go into the DB. I just didn't want to have this only in the service object file, so I put it here like the possible statuses and shipment types. Should I add a comment above the `DefaultServiceMemberSITDaysAllowance` declaration indicating that's what this is for?

pkg/services/mto_shipment/mto_shipment_updater_test.go
```diff
@@ -715,12 +713,30 @@ func (suite *MTOShipmentServiceSuite) TestUpdateMTOShipmentStatus() {
...
+					break
+				}
+			}
+
+			if !codeFound {
+				suite.Fail("Unexpected service code", "unexpected ReService code: %s", string(serviceItem.ReService.Code))
+			}
+
 			// Test that service item was approved within a few seconds of the current time
 			suite.Assertions.WithinDuration(time.Now(), *actualApprovedAt, 2*time.Second)
```
felipe-lee, line 736: @monfresh, do you know if it's actually important that all the approved at times are the same (line 712-713) and that they be within a short time of the current time (i.e. time of action)? Or would it be sufficient to check that before the action none are approved and the approved at time isn't set, then that after the action, they are approved (718) and that the time is set?

felipe-lee: I ask because we're a test that fails for @lynzt because it's off by 4 sec instead of 2, but passes for me and CI.

monfresh: If these assertions are flakey, then we can check for not nil instead. I wonder if there's something else going on though. 4 seconds seems like a long time.

pkg/models/mto_shipments.go
jacquelineIO: The default value is usually just to use what is already in the database. I think I understand what you are saying and I don't see why there is a `const` and a value in the database. As you have explained it, this also doesn't have anything to do with having other customer types in the future. When other types come into play, will the database have more types as well? When creating and pricing payment requests, invoicing will exclusively pull the max number of days from the database. The problem is, someone could change the `const` and/or change what is the database and they will become out of sync. Does there need to be a wrapper function to call to get the correct value? 

It sounds like the intent is that if the `const` is changed then you won't have to update the database via a migration? So you want to keep it the code and not in the database?

felipe-lee: Per the [technical implementation doc](https://docs.google.com/document/d/1Fl5gXA9PQ5Hz_cYZFzCUbVORGJZ9DfX5qqQ6bUMWtGw/edit?disco=AAAAN5NqJRo), it looks like the DB value can be updated if there are SIT extensions so the `const` is more of the initial value, and the DB value is the current value for the shipment, whether that means the default value, or the modified value after extensions are approved.

felipe-lee: As for the other types of customers, they'll have different values for the `const` which can then be used by the `mto_shipment_creator` service object when it decides which value to use based on customer type for the initial DB value. 

So I think based on that, there shouldn't be a need for a wrapper function. 

As for what if the `const` needs to be changed in the future, I agree it would be nicer if we didn't need to update the DB, but that's likely something that would have to be discussed to figure out what transcom would want to do with existing shipments.

pkg/services/mto_shipment/mto_shipment_updater_test.go
felipe-lee: Possibly, though I don't know why it would only be affecting her computer. I guess I'm curious if you get the error too if you run all the server tests on this branch. The test hers is failing on is https://github.com/transcom/mymove/blob/67772e0c0886c42314c3b10f0a68af75368b3609/pkg/services/mto_shipment/shipment_approver_test.go#L78-L148

It's a different file than this comment thread, but the way the times are being tested is the same and I couldn't target the other test properly since I didn't edit close enough to the lines in question.

felipe-lee: It could also possibly be because in that test, there's a few queries that are made between when the shipment is approved (and thus its service items too) and when the `time.Now()` is called in the assertion. So one possible mitigation we were discussing is grabbing `time.Now()` before the approval func is called, and then using that rather than using a fresh one on each loop. Well that and that the way the approval time is being grabbed, it actually doesn't need to be in a loop since we're checking the same time each loop iteration.

lynzt (approved):
Thanks for walking thru the PR and updating some fo those tests.

lynzt (approved):
oops - commented vs approvingLGTM 🚢

PR #7372, opened 2021-09-15
MB-9340 Add service to upload an excess weight record

sandy-wright:
## Description

This PR adds a service object to upload an excess weight record for a move. This involved creating a service in the `move` package to handle the upload and the update of the move, and creating a generic `CreateUpload` service that could handle uploading new files.

## Setup

You can test the full test suite with:

```sh
make server_test
```

Otherwise, you can run the new tests only:

```sh
go test ./pkg/services/move -testify.m TestCreateExcessWeightUpload -v
go test ./pkg/services/upload -testify.m TestCreateUpload -v
```

Currently, there is no handler for this service and thus no way to complete an e2e test.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9340) for this change.


pkg/services/move/excess_weight_uploader.go
```diff
@@ -0,0 +1,77 @@
...
+		if err != nil {
+			return err
+		}
+
+		now := time.Now()
+		move.ExcessWeightQualifiedAt = &now
+		move.ExcessWeightUploadID = &excessWeightUpload.ID
+		move.ExcessWeightUpload = excessWeightUpload
+
+		verrs, err := txnAppCtx.DB().ValidateAndUpdate(move)
```
eamahanna, line 62: Why are we updating the move if we are just creating the weight upload? Does the move have an FK to the weight upload?

sandy-wright: Yeah exactly, the move has the `excess_weight_upload_id` field that points to whatever document the Prime uploaded. It's the only way we can link the move to the document at the moment

pkg/services/upload/upload_creator.go
```diff
@@ -0,0 +1,69 @@
...
+	txErr := appCtx.NewTransaction(func(txnAppCtx appcontext.AppContext) error {
+		newUploader, err := uploader.NewUploader(u.fileStorer, uploader.MaxFileSizeLimit, uploadType)
+		if err != nil {
+			if err == uploader.ErrFileSizeLimitExceedsMax {
+				return services.NewBadDataError(err.Error()) // preserves the error message from the uploader err
+			}
+			return err
+		}
+
+		fileName := time.Now().Format(filenameTimeFormat) + "-" + uploadFilename
```
eamahanna, line 46: Is this how we many all files that are uploaded?

eamahanna: I think for the creating uploads there is usually a file path name as well. Do you think this is needed?

This is from `payment_request/upload_creator`
```
	newfilename := time.Now().Format(VersionTimeFormat) + "-" + filename
	uploadFilePath := fmt.Sprintf("/payment-request-uploads/mto-%s/payment-request-%s", paymentRequest.MoveTaskOrderID, paymentRequest.ID)
	uploadFileName := path.Join(uploadFilePath, newfilename)
```

pkg/services/upload/upload_creator.go
```diff
@@ -0,0 +1,69 @@
...
+		aFile, err := newUploader.PrepareFileForUpload(txnAppCtx, file, fileName)
+		if err != nil {
+			return err
+		}
+
+		newUploader.SetUploadStorageKey(fileName)
+
+		newUpload, verrs, err := newUploader.CreateUpload(txnAppCtx, uploader.File{File: aFile}, uploader.AllowedTypesAny)
```
eamahanna, line 54: Do we need to check with design and maybe infrasec about allowing any file type to be uploaded?

sandy-wright: oof, nice catch. I think it should only be PDFs/images, so let me look up what that allowedTypes option should be. I'll compare with the proof of service uploader

eamahanna (approved):
The only major question or concern I have is about what file types are allowed to be uploaded. If uploading any file type is fine, then feel free to merge. It all looks great!

pkg/services/upload/upload_creator.go
sandy-wright: my thought was that this is now a generic uploader, so whatever service uses it can make their filename whatever path they want when they call this function (although I of course completely forgot to do that myself 😓  - I'll update the move service object with a path filename like that).

As for the timestamp... I saw the proof of service uploader do that, and just slapped it in. If I'm expecting custom filenames though, I should probably rethink that. My one hang up was that I think the `storage_key` needs to be unique, and the proof of service uploader used the filename as the key. I'll poke at that some more

PR #7380, opened 2021-09-16
MB-9433 make it clear that calculate billable weight needs shipment reweigh eager loaded

YanZ777:
## Description

Cherry picked Reggie's commit over into this branch. 
Added cases to use the primeActualWeight (original weight): 
- if there is no reweigh id
- if there is no reweigh weight
- if the reweigh weight is 0.

## Reviewer Notes

Is the logic that I'm putting in make sense? Is there a less convoluted way I could have written it?
Are there any test cases that I missed?
Any other comments that I should put in there?

## Setup

```sh
make office_client_run
make server_run
go test ./pkg/services/mto_shipment
```

Login to the office as a TIO.
Select the shipment with locator MISHRW / service member name of `MissingShipmentReweigh Reweighs`.
The shipment card should and total billable weight should no longer be 0.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9433) for this change

## Screenshots


pearl-truss (approved):
🚀 changes look good to me

duncan-truss:
I think I understand what you're solving for here.  I see the Prime has no validation on reweigh weight being greater than 0 so you could get into that situation.  Likewise the model validation is only checking that it is non-negative if it has a value.  If it is eager loaded though we'll never have a reweigh id that is not uuid.nil right?  It might be worth adding a minimum to prime.yaml swagger file to reject a weight that is not greater than 0 to be safe.

pkg/services/mto_shipment/shipment_billable_weight_test.go
```diff
@@ -51,9 +53,43 @@ func (suite *MTOShipmentServiceSuite) TestShipmentBillableWeightCalculator() {
...
+		billableWeightCalculations, err := billableWeightCalculator.CalculateShipmentBillableWeight(&shipment)
+		suite.NoError(err)
+		suite.Equal(shipment.PrimeActualWeight, billableWeightCalculations.CalculatedBillableWeight)
+	})
+
 	suite.T().Run("If the shipment has an original weight and no reweigh weight and no set billable weight cap, it should return the original weight", func(t *testing.T) {
 		originalWeight := unit.Pound(900)
-		reweigh := models.Reweigh{}
+		reweigh := models.Reweigh{
+			ID: uuid.FromStringOrNil("d66d2f35-218c-4b85-b9d1-631949b9d985"),
```
reggieriser, line 74: Just a note ... if the actual UUID value doesn't matter for the test (I don't believe it does for any except the `uuid.Nil` case), another way to handle this without having to generate and add a literal UUID is to do `uuid.Must(uuid.NewV4())`.

pkg/services/mto_shipment/shipment_billable_weight_test.go
```diff
@@ -93,4 +129,58 @@ func (suite *MTOShipmentServiceSuite) TestShipmentBillableWeightCalculator() {
...
+		var dbShipment models.MTOShipment
+		err := suite.DB().Eager("Reweigh").Find(&dbShipment, shipment.ID)
+		suite.FatalNoError(err)
+
+		billableWeightCalculations, err := billableWeightCalculator.CalculateShipmentBillableWeight(&dbShipment)
+		suite.NoError(err)
+		suite.Equal(reweigh.Weight, billableWeightCalculations.CalculatedBillableWeight)
+	})
+
+	suite.T().Run("Eagerly loaded reweigh where a reweight does not exist", func(t *testing.T) {
```
reggieriser, line 151: Nit, but pointing out a typo that was my fault since the test came from my branch ... `reweight` should be `reweigh`.  My finger muscle memory just can't seem to type `weigh` instead of `weight`.

reggieriser (commented):
The logic/tests generally LGTM, but agree with @duncan-truss that we should try to stop negative weight data if we can before it gets in the DB.  A check constraint in the DB could be another safety for that.  I added a few nits that aren't mandatory things to address.

YanZ777:
> I think I understand what you're solving for here. I see the Prime has no validation on reweigh weight being greater than 0 so you could get into that situation. Likewise the model validation is only checking that it is non-negative if it has a value but that could be worth changing to require a positive value. If it is eager loaded though we'll never have a reweigh id that is not uuid.nil right? It might be worth adding a minimum to prime.yaml swagger file to reject a weight that is not greater than 0 to be safe.

Should I take out the test and the check for if the reweigh id is nil then? 

Are there existing tests that I could look at that are similar to rejecting a weight that is not greater than 0 once I add it to the prime.yaml?

duncan-truss:
> > I think I understand what you're solving for here. I see the Prime has no validation on reweigh weight being greater than 0 so you could get into that situation. Likewise the model validation is only checking that it is non-negative if it has a value but that could be worth changing to require a positive value. If it is eager loaded though we'll never have a reweigh id that is not uuid.nil right? It might be worth adding a minimum to prime.yaml swagger file to reject a weight that is not greater than 0 to be safe.
> 
> Should I take out the test and the check for if the reweigh id is nil then?
> 
> Are there existing tests that I could look at that are similar to rejecting a weight that is not greater than 0 once I add it to the prime.yaml?

Here is an example of a payload that sets a minimum in the ghc.yaml https://github.com/transcom/mymove/blob/73b0c062db0c1db95e9fb7512761291512b7a842/swagger-def/ghc.yaml#L2574

Then in a test I think if you go to validate the params you should get an error https://github.com/transcom/mymove/blob/d0fdf8b6a2a99f2de44c962a1d745b43abaaa1d2/pkg/handlers/supportapi/mto_shipment_test.go#L161

Here is an existing optional positive int validator for the model as well https://github.com/transcom/mymove/blob/c2e190f7cd51e6c74d8f9f34e718b68967a7202e/pkg/models/validators.go#L69

pkg/handlers/primeapi/reweigh_test.go
```diff
@@ -107,9 +107,56 @@ func (suite *HandlerSuite) TestUpdateReweighHandler() {
 		suite.Equal(&reason, reweighOk.Payload.VerificationReason)
 	})
 
+	suite.T().Run("Failure 422 - Failed to update reweigh weight due to bad request - zero reweigh value", func(t *testing.T) {
+		// Testcase:   reweigh us updated with the new weight of the shipment
+		// Expected:   Failure 422
+
+		// Update with weights
+
+		req := httptest.NewRequest("PATCH", fmt.Sprintf("/mto-shipments/%s/rewighs/%s", reweigh.ShipmentID.String(), reweigh.ID.String()), nil)
```
duncan-truss, line 116: Glad I'm not alone in having misspelled reweigh what feels like 100 times in the last month 😆

pkg/handlers/primeapi/reweigh_test.go
```diff
@@ -107,9 +107,56 @@ func (suite *HandlerSuite) TestUpdateReweighHandler() {
...
+			ReweighID:     *handlers.FmtUUID(reweigh.ID),
+			MtoShipmentID: *handlers.FmtUUID(reweigh.ShipmentID),
+			IfMatch:       updatedETag,
+			Body: &primemessages.UpdateReweigh{
+				Weight: &weight,
+			},
+		}
+
+		// Run swagger validations
+		suite.Error(params.Body.Validate(strfmt.Default))
```
duncan-truss, line 130: This test looks great, are we getting a weight specific error payload message that's worth checking here or just a 422?

pkg/models/validators.go
```diff
@@ -95,6 +95,21 @@ func (v *OptionalPoundIsNonNegative) IsValid(errors *validate.Errors) {
...
+type OptionalPoundIsPositive struct {
+	Name  string
+	Field *unit.Pound
+}
+
+// IsValid adds an error if the Field is less than zero
+func (v *OptionalPoundIsPositive) IsValid(errors *validate.Errors) {
+	if v.Field != nil {
+		if *v.Field <= 0 {
+			errors.Add(validators.GenerateKey(v.Name), fmt.Sprintf("%d is less than zero.", *v.Field))
```
duncan-truss, line 108: let's throw in the "or equal to zero" to the error message.

duncan-truss (approved):
Thanks for addressing those zero validation checks!

reggieriser (approved):
LGTM!  Looks like you still need to change a test to account for your new error message text, but I'm going to go ahead and approve since that should be a trivial change.

reggieriser (commented):
Actually, I just noticed that there are no tests in `validators_test.go` for your new validator.  Might be nice to add those before merging if you can.

cmd/send-to-gex/main.go
```diff
@@ -140,9 +140,13 @@ func main() {
...
+	statusCode := 0
+
 	if resp == nil {
 		log.Fatal("Gex Sender had no response")
+	} else {
+		statusCode = resp.StatusCode
 	}
 
-	fmt.Printf("status code: %v, error: %v \n", resp.StatusCode, err)
+	fmt.Printf("status code: %v, error: %v \n", statusCode, err)
```
YanZ777, line 151: I was getting some linter warnings here. Some folks get them, others don't, so I figured I would just fix them. 
I know this case won't be hit because there's a bailout for nil checks, but the linter can be rather aggressive.
[https://ustcdp3.slack.com/archives/CP6PTUPQF/p1631913587351200](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1631913587351200)
@felipe-lee

reggieriser (approved):
LGTM!  Thanks for adding the additional tests. 🚢

PR #7386, opened 2021-09-16
Adding missing generated mock for SITExtensionCreatorAsTOO

reggieriser:
## Description

After doing a `make mocks_generate` on master, I noticed I had an untracked file for `pkg/services/mocks/SITExtensionCreatorAsTOO.go`.  Looks like we just missed committing the generated mock as part of #7352.  This PR adds the missing mock.


duncan-truss (approved):
Thanks for catching this!

PR #7416, opened 2021-09-21
MB-8063 Refactor handler for updating payment service item status to use service object

Ryan-Koch:
## Description

Makes it so the `UpdatePaymentServiceItemStatusHandler` now uses the `PaymentServiceItemStatusUpdater` instead of using `QueryBuilder` from within the handler.

## Reviewer Notes
Functionality for these shouldn't change, so to test that get things running:
```
make db_dev_e2e_populate
make server_start
make office_client_start
```

and try reviewing some payment requests. Basic functionality should be unchanged from before.


pkg/handlers/ghcapi/payment_service_items_test.go
```diff
@@ -112,34 +85,13 @@ func (suite *HandlerSuite) TestUpdatePaymentServiceItemHandler() {
 
 	})
 
-	suite.Run("422 - Integration Test", func() {
-		subtestData := suite.makeUpdatePaymentSubtestData()
-		newParam := subtestData.input
-		newParam.Body.Status = ""
```
Ryan-Koch, line 118: Removed this test because it seemed to be hinging on this empty string status field. This should be covered by the swagger enum, so I don't think the test is necessary.

pkg/handlers/ghcapi/payment_service_items_test.go
```diff
@@ -53,38 +52,16 @@ func (suite *HandlerSuite) makeUpdatePaymentSubtestData() (subtestData *updatePa
 		},
 	}
 
-	// for 422 and 500 errors
-	psi := testdatagen.MakeDefaultPaymentServiceItem(suite.DB())
```
Ryan-Koch, line 57: This was only being used for the test I got rid of further down.

pkg/handlers/ghcapi/payment_service_items_test.go
```diff
@@ -53,38 +52,16 @@ func (suite *HandlerSuite) makeUpdatePaymentSubtestData() (subtestData *updatePa
...
 
 		handler := UpdatePaymentServiceItemStatusHandler{
-			HandlerContext: handlers.NewHandlerContext(suite.DB(), suite.TestLogger()),
-			Fetcher:        fetcher,
-			Builder:        *queryBuilder,
+			HandlerContext:                  handlers.NewHandlerContext(suite.DB(), suite.TestLogger()),
+			PaymentServiceItemStatusUpdater: paymentServiceItemService.NewPaymentServiceItemStatusUpdater(),
 		}
 
 		response := handler.Handle(subtestData.params)
```
monfresh, line 67: Can we also update these tests to validate the Body?
```
```
This assertion should go before the response definition.

pkg/services/payment_service_item/payment_service_item_service_test.go
```diff
@@ -0,0 +1,35 @@
...
+	logger *zap.Logger
+}
+
+// TestAppContext returns the AppContext for the test suite
+func (suite *PaymentServiceItemSuite) TestAppContext() appcontext.AppContext {
+	return appcontext.NewAppContext(suite.DB(), suite.logger)
+}
+
+func (suite *PaymentServiceItemSuite) SetupTest() {
+	err := suite.TruncateAll()
```
monfresh, line 24: For new services, we want to run tests inside a transaction instead of truncating. Here is the how-to:
https://transcom.github.io/mymove-docs/docs/dev/testing/writing-tests/Running-server-tests-inside-a-transaction

pkg/services/payment_service_item/payment_service_item_status_updater.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	// Save the record
+	verrs, err := appCtx.DB().ValidateAndSave(&paymentServiceItem)
+	if err != nil || verrs != nil && verrs.HasAny() {
+		return models.PaymentServiceItem{}, verrs, err
+	}
+	// Return the updated object
+	return paymentServiceItem, nil, nil
+}
```
monfresh, line 72: Could we break down this function into 2 or 3 smaller ones that each perform the individual steps: finding the payment service item, checking the eTag, and updating the service item? We could also check the eTag using our new preferred validation pattern. 

Here's an example:
https://github.com/transcom/mymove/blob/master/pkg/services/mto_service_item/mto_service_item_updater.go#L43

I'm happy to pair if any of these patterns are new to you.

pkg/services/payment_service_item/payment_service_item_status_updater_test.go
```diff
@@ -0,0 +1,78 @@
...
+		// Arbitrary date time that isn't the record updatedAt used here
+		badETag := etag.GenerateEtag(testdatagen.DateInsidePerformancePeriod)
+		updater := NewPaymentServiceItemStatusUpdater()
+
+		_, _, err := updater.UpdatePaymentServiceItemStatus(suite.TestAppContext(),
+			paymentServiceItem.ID, models.PaymentServiceItemStatusApproved, nil, badETag)
+
+		suite.Error(err)
+		suite.IsType(services.PreconditionFailedError{}, err)
+	})
```
monfresh, line 76: Can we also add a test if a validation fails, such as trying to deny the service item without providing a reason? It doesn't look like we were checking for this scenario, so I'm not sure if that's a requirement, but I would guess it would be, similarly to how move service items require a rejection reason when rejecting them.

Ryan-Koch: Ah yes I forgot about that requirement! I'll test for that and adjust accordingly.

pkg/handlers/ghcapi/payment_service_items_test.go
```diff
@@ -80,11 +99,26 @@ func (suite *HandlerSuite) TestUpdatePaymentServiceItemHandler() {
...
 
+	suite.Run("422 - Fails to reject without rejectionReason - Integration Test", func() {
+		subtestData := suite.makeUpdatePaymentSubtestData()
+
+		handler := UpdatePaymentServiceItemStatusHandler{
+			HandlerContext:                  handlers.NewHandlerContext(suite.DB(), suite.TestLogger()),
+			PaymentServiceItemStatusUpdater: paymentServiceItemService.NewPaymentServiceItemStatusUpdater(),
+		}
+
+		suite.NoError(subtestData.params.Body.Validate(strfmt.Default))
```
monfresh, line 116: Thanks for adding this test. Since the params are being modified, I think we need to define `subtestData.rejectWithNoReason` before validating it. Otherwise, this is validating `subtestData.params` which has a different Body. Also, since `subtestData.params` and `subtestData.rejectWithNoReason` are basically the same except for the Status, would it make more sense to update the Status inside the test without having to define a new data structure?

Something like this, which you've done in some of the other tests:
```golang
subtestData.params.Body.Status = ghcmessages.PaymentServiceItemStatusDENIED
suite.NoError(subtestData.params.Body.Validate(strfmt.Default))
response := handler.Handle(subtestData.params)
```

pkg/services/payment_service_item/payment_service_item_service_test.go
```diff
@@ -20,14 +20,9 @@ func (suite *PaymentServiceItemSuite) TestAppContext() appcontext.AppContext {
...
-func (suite *PaymentServiceItemSuite) SetupTest() {
-	err := suite.TruncateAll()
-	suite.FatalNoError(err)
-}
-
-func TestOrderServiceSuite(t *testing.T) {
+func TestPaymentServiceItemServiceSuite(t *testing.T) {
 	ts := &PaymentServiceItemSuite{
-		PopTestSuite: testingsuite.NewPopTestSuite(testingsuite.CurrentPackage()),
+		PopTestSuite: testingsuite.NewPopTestSuite(testingsuite.CurrentPackage(), testingsuite.WithPerTestTransaction()),
```
monfresh, line 25: Yay transaction tests!

pkg/handlers/ghcapi/payment_service_items_test.go
Ryan-Koch: Ah yeah that's a fair point, it is a bit heavy to create an entire new object just for one field. I'll make the changes.

pkg/services/payment_service_item/payment_service_item_status_updater.go
```diff
@@ -24,8 +24,34 @@ func NewPaymentServiceItemStatusUpdater() services.PaymentServiceItemStatusUpdat
...
+	if err != nil || verrs != nil && verrs.HasAny() {
+		return models.PaymentServiceItem{}, verrs, err
+	}
+
+	// Check the etag to make sure we're good to update (not stale)
+	currentEtag := etag.GenerateEtag(paymentServiceItem.UpdatedAt)
+	if currentEtag != eTag {
+		return models.PaymentServiceItem{}, nil, services.NewPreconditionFailedError(paymentServiceItem.ID,
+			query.StaleIdentifierError{StaleIdentifier: eTag})
+	}
```
monfresh, line 39: Nice! Thanks for breaking this up. This looks great. Since the eTag is being checked as part of the validation checks, we should be able to get rid of this section here.

Ryan-Koch: Oh good eye, I forgot to remove that block

monfresh (approved):
Looks great! Thanks for making the changes.

PR #7404, opened 2021-09-20
[MB-9288] Steel Cable Prime Simulator in Office App

ahobson:
## Description

The office app already has multiple views depending on the roles available, so add a prime simulator role and a simple view of available moves, showing user based auth to the Prime API

## Reviewer Notes

This is totally a proof of concept and ugly with no tests. I've only tested this using devlocal auth.

See the demo link here: https://office-milmove-pr-7404.mymove.sandbox.truss.coffee

## Setup

1. Start the app locally with some test data
2. Visit the office app
3. Create a new Prime Simulator user
4. See the list of available moves

## Code Review Verification Steps

* [x] If the change is risky, it has been tested in experimental before merging.
* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#logging)
* [x] The requirements listed in [Querying the Database Safely](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#querying-the-database-safely) have been satisfied.
* Any new migrations/schema changes:
  * [x] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://github.com/transcom/mymove/wiki/migrate-the-database#zero-downtime-migrations))
  * [x] Have been communicated to #g-database
  * [x] Secure migrations have been tested following the instructions in our [docs](https://github.com/transcom/mymove/wiki/migrate-the-database#secure-migrations)
* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [x] Have the Jira acceptance criteria been met for this change?

PR #7405, opened 2021-09-20
[MB-9288] Exp env testing

ahobson:
DO NOT MERGE

PR #7424, opened 2021-09-22
[MB-9288] Switch to gorilla mux

rogeruiz:
## Description

> Gorilla mux provides the feature of being able to determine which
> route a request is associated with which will be necessary for
> recording metrics associated with the route instead of the file.

> It also provides a way to only enable a route for a given host

> In addition gorilla mux is actively maintained and goji is not.

## Reviewer Notes

This was originally introduced in the PR #7404 and should be its own PR ~~as [it's not related to the Prime Simulator effort as described by @ahobson here](https://github.com/transcom/mymove/pull/7404#discussion_r712472073).~~ because it is related to both the Prime Simulator work and some metrics and observability work in the future.

## Setup

I am not entirely sure what the setup for this is. But I will make space for @ahobson to chime in if he has something to add here. I personally tested this by ensuring that the Applications all come up and that logging is still being done properly. I also ran the tests locally to verify that everything was passing. 

Since this is a low-level change to the Mux Router, I don't know of a better way of testing this without ensuring that tests are passing appropriately and that the Application's routes are working as expected.

There's no expectations for merge conflicts but this is a global change to the whole of `mymove` backend.

## References

* [Jira](https://dp3.atlassian.net/browse/MB-9288) ticket associated with this work.


rogeruiz:
Once this is merged into `master`, we can rebase #7404 to get rid of these changes from that PR so that it's less visually noisy since the majority of that effort is to get an MVP of the Prime Simulator up.

rogeruiz:
Here is some [additional context in Slack](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1632324163392800) around why this should be it's own patch series. It has been [specifically discussed here](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1632325382396100?thread_ts=1632324163.392800&cid=CP6PTUPQF) by a member of @transcom/truss-sr-eng. 

I've updated this patch series with some additional commits to ensure that things are passing in CI due to how the work was related to the changes in #7404 but how logically it should be in its own PR.

And I will be adding the necessary code to #7404 which needs this the code in this PR to exist before it can actually be useful. Because of the way Git handles changes, some of these commits may or may not cause merge conflicts, but I can take care of that when this is merged into `master`.

ahobson (approved):
I like this change! 😁

rogeruiz:
@ahobson has pushed this to EXP in #7405 as that has this SHA in its history. I've logged into the app on the following URLs and tested the API with the EXP certificates from my local machine.

- https://my.exp.move.mil
- https://office.exp.move.mil
- https://admin.exp.move.mil
- https://api.exp.move.mil

This change LGTM 🌈

PR #7432, opened 2021-09-23
MB-8841: Fixes Orders and OrdersInfoForm tests

kylehilltruss:
## Description

This pull request refactors tests in three suites to be more in line with current front-end testing best practices (using `screen` instead of methods returned from `render`, using RTL instead of enzyme, appropriately using `await findByWhatever` instead of `waitFor`); this should greatly reduce the incidence of these tests sporadically failing during our CircleCI `client_test` build jobs, of which these were probably the worst offenders in our front-end code base.

One specific change in this version of the pull request (as opposed to [a previous one](https://github.com/transcom/mymove/pull/7239)) is that a `delay` value of 100ms is added to `userEvent.type` commands in these three test suites. This has the effect of slowing down simulated input into the third party `react-select` combo box, which allows the (also simulated) async functions to catch up with autocomplete options.

I've run these client tests about 20 times in a CircleCI environment, and the changed tests have passed each time. Since each of these flaky tests seem to fail around 5-10% of the time, it gives me a reasonable level of confidence that we've actually resolved the issue this time.

## Reviewer Notes

There should be no visible changes to the UI (quite literally, all we're changing/adding are test files). Ensure that the modified tests aren't emitting any messages to the console during test runs.

## Setup

```sh
yarn test
```

## References

* Closes [MB-8841](https://dp3.atlassian.net/browse/MB-8841) and [MB-8842](https://dp3.atlassian.net/browse/MB-8842).
* [Common mistakes with React Testing Library](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library/)

### DIFF
--- a/src/components/Customer/EditOrdersForm/EditOrdersForm.test.jsx
+++ b/src/components/Customer/EditOrdersForm/EditOrdersForm.test.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import selectEvent from 'react-select-event';
 
 import EditOrdersForm from './EditOrdersForm';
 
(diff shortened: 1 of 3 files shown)


felipe-lee (approved):
LGTM, tests all ran well, I'll trust your 20 Circle CI runs vs running my own...

PR #7408, opened 2021-09-20
MB-9372 add handler for prime sit extension

eamahanna:
## Description

This PR covers quite a few things:
- Adds handler for createSitExtension for the prime api
- Updates the ghc swagger definition for `sitExtensions` to use the shared definition file
- Updates `getMove` endpoint to include sitExtensions
- Updates the response for the `updateShipment` endpoint to return sitExtensions
- Adds tests

## Reviewer Notes

I am in a bit of a time crunch this morning. I can add more detailed instructions on how to test if need be. Just ping me and let me know. This is a big PR so I wanted to allow people time to review it. I should be able to add instructions after the weekly demo.

## Setup

Add any steps or code to run in this section to help others prepare to run your code:

```sh
echo "Code goes here"
```

## Code Review Verification Steps

* [ ] If the change is risky, it has been tested in experimental before merging.
* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#logging)
* [ ] The requirements listed in [Querying the Database Safely](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#querying-the-database-safely) have been satisfied.
* Any new migrations/schema changes:
  * [ ] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://github.com/transcom/mymove/wiki/migrate-the-database#zero-downtime-migrations))
  * [ ] Have been communicated to #g-database
  * [ ] Secure migrations have been tested following the instructions in our [docs](https://github.com/transcom/mymove/wiki/migrate-the-database#secure-migrations)
* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9372) for this change

PR #7433, opened 2021-09-23
[MB-8758] Use correct operation ID for SIT extension approve/deny

hannaliebl:
## Description

https://github.com/transcom/mymove/pull/7408 changed the name from Sit to SIT in the API and that in turn meant that the operation ID that the front end uses to call the API needed to be updated.

## Setup

make server_run, make client_run

Use the `PENDNG` move code and make sure that approval/denial works.


duncan-truss (approved):
Approving and denying worked for me on the test move.

PR #7445, opened 2021-09-25
MB-9308 Update requestedPickupDate and requestedDeliveryDate on update shipment

gidjin:
## Description

This is needed as these two fields are optional on the shipment as they
are not set right away. However, during an update if they are not set to
optional in the api the payload will contain a zero value which can
trigger an update and over write a non zero value that has already been
saved. This change makes them nullable so we can check to tell when
converting the payload to a model.

## Reviewer Notes

I haven't checked everything that uses this payload, but seems like an okay change since it is on update. 
I'm not sure if this should also be applied to the `BillableWeightCap` too since it's a pointer, but I expect it would also be an issue.

## Setup

* Create or find a shipment that has a requested pickup date.
* Login as the TIO and adjust the weight of the specific shipment
* Check the database to see that RequestedPickupDate is not erased

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9308) for how we found this issue
* [this slack thread](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1632521752451000) explains more about the approach used.


jacquelineIO (approved):
lgtm. successfully edited billable shipment max from TIO UI and submitted payment request for the shipment.

PR #7450, opened 2021-09-27
[MB-7416] TIO Review Payment Request Service Items Shows Days In SIT balance

duncan-truss:
## Description

When a TIO is reviewing a payment request with a Domestic Origin or Destination Additional Day SIT on it, they should be able to see the current balance of SIT days used, pending invoice, and total remaining.

This connects to the new backend ghc endpoint that returns the SIT balance for the shipment that the service item is on.  It is only shown when the payment request needs review, not when viewing an already approved or rejected payment request's service items.

## Reviewer Notes

## Setup

```sh
make db_dev_e2e_populate
make server_run
make office_client_run
```

1. Login as the TIO office user
2. Find one of the payment requests belonging to the move code `PARSIT`.
3. See that there is one payment request that has already been reviewed with a Domestic Origin SIT that used 30 days of the shipment's allotment of 120 days.
4. Select the payment request needing review that has a Domestic Destination SIT that is pending invoice for 60 days.  The Days in SIT balance will show that there are 30 days remaining in the shipment's SIT allotment if this payment request is approved.

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-7416) for this change

## Screenshots
Screen Shot 2021-09-28 at 4 54 51 PM


duncan-truss:
I see I have too big of a margin on the service item cards in the Happo diffs, going to fix that now.

src/components/Office/ReviewServiceItems/ServiceItemCard.jsx
```diff
@@ -214,7 +220,15 @@ const ServiceItemCard = ({
                 <dl>
                   <dt>Service item</dt>
                   <dd data-testid="serviceItemName">{mtoServiceItemName}</dd>
-
+                  {isAdditionalDaySIT(mtoServiceItemCode) && (
+                    <>
+                      <dt className={styles.daysInSIT}>Days in SIT</dt>
+                      <dd>
+                        {/* eslint-disable-next-line react/jsx-props-no-spreading */}
+                        <DaysInSITAllowance className={styles.daysInSITDetails} {...shipmentSITBalance} />
```
Ryan-Koch, line 228: Here we could consider explicitly passing the params (admittedly verbose) or creating a shape for these to use in the propTypes requirement for that `DaysInSitAllowance` component in order to avoid the linter rule. 

Going the object with shape route could mean taking the shape block you have further down and tossing it in its own file with an export looking like this:
```
export const ShipmentSITBalanceShape = PropTypes.shape({
    previouslyBilledDays: PropTypes.number,
    previouslyBilledEndDate: PropTypes.string,
    pendingSITDaysInvoiced: PropTypes.number,
    pendingBilledEndDate: PropTypes.string,
    totalSITDaysAuthorized: PropTypes.number,
    totalSITDaysRemaining: PropTypes.number,
    totalSITEndDate: PropTypes.string,
  })
  ```
   That would then mean being able to do this in the proptypes for the component: 
 ```
 DaysInSITAllowance.propTypes = {
 ...
   shipmentSITBalance: ShipmentSITBalanceShape.isRequired,
 ...
 }
 ```

Comment (duncan-truss) [reply src/components/Office/ReviewServiceItems/ServiceItemCard.jsx:228]: Good call, yea I don't think it will be a shared shape elsewhere but it would eliminate the prop spreading.

**Review by Ryan-Koch — APPROVED** [review]
Was able to do see the right info locally. I made a comment about an eslint thing, but otherwise this LGTM.

File: src/components/Office/DaysInSITAllowance/DaysInSITAllowance.test.jsx [hunk 2]
```diff
@@ -0,0 +1,59 @@
...
+import DaysInSITAllowance from './DaysInSITAllowance';
+
+describe('DaysInSITAllowance Component', () => {
+  const shipmentSITBalance = {
+    previouslyBilledDays: 30,
+    previouslyBilledEndDate: '2021-06-08',
+    pendingSITDaysInvoiced: 60,
+    pendingBilledEndDate: '2021-08-08',
+    totalSITDaysAuthorized: 120,
+    totalSITDaysRemaining: 30,
```
monfresh, line 13: Not a big deal, but what do you think about using values that are all different from each other, or maybe a few more scenarios? Instead of having two test values that are both `30`, the test might feel more confident that we are doing the right calculation when the values are different. For example, if we accidentally based the remaining days on `previouslyBilledDays`, the test would still pass.

duncan-truss: That's a valid point and I've seen that mistake happen before. I designed this test to match the Zeplin mocks so those are the values I defaulted to. I think the next test has unique values though and I exposed all of the storybook argTypes as controls that lets you change values at will.

monfresh: I meant to put this comment in the server test since that's where the calculation is being performed.

duncan-truss: Ahh gotcha, the first one kind of shows 3 different values because the previously billed can be nil https://github.com/transcom/mymove/pull/7450/files#diff-4dc21ac88136778ad459f97cd329a8db981ae6a881fbcaf6b2b21df5ca844973R102-R107

Shkeating:
looks good, but I think we need to do something to get Happo to re-run?

pkg/services/payment_request/payment_request_shipments_sit_balance.go
```diff
@@ -26,9 +26,15 @@ func NewPaymentRequestShipmentsSITBalance() services.ShipmentsPaymentSITBalance
 func getStartAndEndParams(params models.PaymentServiceItemParams) (start time.Time, end time.Time, err error) {
 	for _, paymentServiceItemParam := range params {
 		if paymentServiceItemParam.ServiceItemParamKey.Key == models.ServiceItemParamNameSITPaymentRequestStart {
-			start, err = time.Parse(sitParamDateFormat, paymentServiceItemParam.Value)
+			// remove once the pricer work is done so a 500 server error isn't returned for an unparseable date
```
monfresh, line 29: Should we create a JIRA ticket so we don't forget this?

pkg/services/payment_request/payment_request_shipments_sit_balance.go
```diff
@@ -52,9 +58,13 @@ func lookupDaysInSIT(params models.PaymentServiceItemParams) (int, error) {
 	return 0, nil
 }
 
+func isAdditionalDaySIT(code models.ReServiceCode) bool {
```
monfresh, line 61: Thanks for putting this in a well-named function!

monfresh (approved):
Nice work!

monfresh:
When you're ready to merge, can you please add the "ready-to-merge" label so we can test Mergify?

duncan-truss:
> looks good, but I think we need to do something to get Happo to re-run?

@Shkeating Thanks for noticing, Kyle removed a function I needed in a different PR so I had to solve the mystery of where my formatter went 🙃

pkg/services/payment_request/payment_request_shipments_sit_balance.go
duncan-truss: Yea we probably can, if payment requests with SIT service items were created in the last month in different environments those pricing params are going to exist with "NOT IMPLEMENTED" values and we'll have an issue reviewing them if we remove this.

duncan-truss: It looks like A-team will at some point reject SIT service items without those params from being created https://ustcdp3.slack.com/archives/CP4979J0G/p1632872544319200?thread_ts=1632867211.318500&cid=CP4979J0G

Shkeating:
I think we may have added extra space on the second label in *all* Review Service Item components. maybe add the space in as a margin bottom of the gray box with the SIT info?

image

Shkeating:
> I think we may have added extra space on the second label in _all_ Review Service Item components. maybe add the space in as a margin bottom of the gray box with the SIT info?
> 
> image

This larger space is supposed to be there after reviewing the Zeplin designs. Nice catch!

Shkeating (approved):

PR #7459, opened 2021-09-28
[MB-9524] Edit billable weight validation

Ronolibert:
## Description

Add validation to edit billable weight fields

## Setup
`make server_run`
`make office_client_run`
Log in as TIO
View payment request
Review weight
Click on `Edit`
Put an empty value for weight
Expected result: Should show required message and disabled button

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9524) for this change

## Screenshots
Screen Shot 2021-09-28 at 8 22 32 AM

### DIFF
--- a/src/components/Office/BillableWeight/EditBillableWeight/EditBillableWeight.jsx
+++ b/src/components/Office/BillableWeight/EditBillableWeight/EditBillableWeight.jsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
 import * as Yup from 'yup';
-import { func, number, string } from 'prop-types';
+import { func, number, string, bool } from 'prop-types';
 import { Formik } from 'formik';
 import { Button, Fieldset, Label, Textarea } from '@trussworks/react-uswds';
 
@@ -88,14 +88,15 @@ export default function EditBillableWeight({
   estimatedWeight,
   maxBillableWeight,
   originalWeight,
+  showFieldsInitial,
   title,
   totalBillableWeight,
   weightAllowance,
 }) {
-  const [showEditBtn, setShowEditBtn] = useState(true);
+  const [showFields, setShowFields] = useState(showFieldsInitial);
 
   const toggleEdit = () => {
-    setShowEditBtn(!showEditBtn);
+    setShowFields((show) => !show);
   };
 
   const initialValues = {
@@ -106,7 +107,7 @@ export default function EditBillableWeight({
   return (
     <div className={styles.wrapper}>
       <h4 className={styles.header}>{title}</h4>
-      {showEditBtn ? (
+      {!showFields ? (
         <>
           <span>{billableWeight ? formatWeight(billableWeight) : formatWeight(maxBillableWeight)}</span>
           {billableWeightJustification && (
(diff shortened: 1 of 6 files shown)


src/components/Office/BillableWeight/EditBillableWeight/EditBillableWeight.jsx
```diff
@@ -73,6 +76,11 @@ MaxBillableWeightHintText.defaultProps = {
   estimatedWeight: null,
   weightAllowance: null,
 };
+
+const validationSchema = Yup.object({
+  billableWeight: Yup.number().min(1, 'Authorized weight must be greater than or equal to 1').required('Required'),
```
donaldthai, line 81: Wondering if we still need to do that `500 lb min` [based on the thread](https://ustcdp3.slack.com/archives/C021XEC7TV4/p1632338509125900)? Any preference here @cosburn?

donaldthai (approved):
Code looks good. I had one question and noticed one other thing around the edit shipment billable weight. The edit shipment billable weight form looks a bit weird in the card: 
image

@cosburn it look ok to you?

src/components/Office/BillableWeight/EditBillableWeight/EditBillableWeight.jsx
Shkeating: > Code looks good. I had one question and noticed one other thing around the edit shipment billable weight. The edit shipment billable weight form looks a bit weird in the card: image
> 
> @cosburn it look ok to you?

I just came here to say, while I can't find the error state of this field in the zeplin designs, that I'd give it more space above the `Required` text. My eyes want it to have the same amoutn of space as the `Remarks` label has above it, which is `24px, so maybe try that? But ultimately I'd say this is Claire's call.

Shkeating:
Should we include the error state in Happo? I didn't see any new images added.

src/components/Office/BillableWeight/EditBillableWeight/EditBillableWeight.jsx
cosburn: Looking at the way we show errors on other forms, I think part of the issue is that the `Required` text is usually larger than the label. (example from customer UI)
Screen Shot 2021-09-29 at 9 44 54 AM

I would also say, because this is in a tighter container than other forms, I think it would look cleaner to use this error treatment: https://storybook.dp3.us/?path=/story/components-forms--text-fieldset-with-error (rather than the bar on the left side).

One other note -- the way other fields work, the error state doesn't display until the field is clicked and then clicked out of. I don't think we want the error to be visible when the form is opened.

cosburn: > Wondering if we still need to do that `500 lb min` [based on the thread](https://ustcdp3.slack.com/archives/C021XEC7TV4/p1632338509125900)? Any preference here @cosburn?

This is a requirements question, we should probably get clarification from Dan on the expectations here. I will ping him in the thread to ask.

Ronolibert:
@Shkeating I have added the error state to the happo/storybook!

cosburn (approved):
Looks good!

PR #7467, opened 2021-09-29
Update health check to only check https

gidjin:
## Description

Seems http health check has broken. Since we expect to use https this changes the health check to only check for https endpoints.

Also 10 tries with each having a 15 minute time out seems excessive.
This PR reduces the timeout to 5 minutes for each of the 10 tries.

[See this thread](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1632950530011200)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -374,7 +374,7 @@ commands:
           no_output_timeout: 20m
       - run:
           name: Health Check
-          command: bin/health-checker --schemes http,https --hosts << parameters.health_check_hosts >> --tries 10 --backoff 3 --log-level info --timeout 15m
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -410,7 +410,7 @@ commands:
           no_output_timeout: 20m
       - run:
           name: Health Check
-          command: bin/health-checker --schemes http,https --hosts << parameters.health_check_hosts >> --tries 10 --backoff 3 --log-level info --timeout 15m
+          command: bin/health-checker --schemes /ttp,https --hosts << parameters.health_check_hosts >> --tries 10 --backoff 3 --log-level info --timeout 5m
```
carterjones, line 413: I'm not sure what `/http` is. Did you mean to delete this and just put `--schemes https`?

carterjones (approved):

felipe-lee (approved):

PR #7479, opened 2021-10-01
MB-9535 fix billable weight hint texts

YanZ777:
## Description

Fixed some of the logic around the hint text, refactored some tests, took out the console output being generated from the tests.

## Reviewer Notes

Is the logic around the hint text correct?
Should there be other tests? 
Are the tests split out ok?

## Setup

```sh
make office_client_run
make server_run
```

The easiest way might be to change the move with the locator `MAXCED`. 
I adjusted the max billable weight on that to be 1500 and the shipment weight to be 1560, and then played with those values.

For hint text that shouldn't be there on the edit max billable weight page, you can use the move with the locator `NOESTW`

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9535) for this change

## Screenshots

image


Shkeating (approved):

src/components/Office/BillableWeight/EditBillableWeight/EditBillableWeight.jsx
```diff
@@ -17,20 +17,27 @@ function BillableWeightHintText({
   originalWeight,
   totalBillableWeight,
 }) {
-  const showToFit = billableWeight > maxBillableWeight && billableWeight < estimatedWeight * 1.1;
+  const estimatedWeightTimes110 = estimatedWeight * 1.1;
+
+  const showToFit = billableWeight > maxBillableWeight && billableWeight > estimatedWeightTimes110;
```
donaldthai, line 22: So looking at the display logic, shouldn't this show when the `max billable weight` is exceeded by the `total billable weight` AND the `toFitValue` is below the 110% estimated weight?
image

donaldthai (commented):
I'm seeing a weird issue where the default value of the input on the edit billable weight form is using the max billable weight value: 
Oct-04-2021 16-18-31

YanZ777:
> I'm seeing a weird issue where the default value of the input on the edit billable weight form is using the max billable weight 
Good catch! I've fixed this.

src/components/Office/BillableWeight/EditBillableWeight/EditBillableWeight.jsx
YanZ777: Hmmmmm that doesn't sound quite right either, but I do think my signs are flipped here. I'll wait for Claire to weigh in on this.

cosburn (approved):
Looks like the logic is working as expected!

donaldthai (approved):
Code looks good! Since Claire is happy with the functionality, I'm happy!

PR #7643, opened 2021-10-28
MB-9977: Convert to using appcontext almost everywhere

ahobson:
## Description

Fully convert to using appcontext. As part of this, convert the `PopTestSuite` to contain a logger so we can consolidate appcontext test helpers.

The references to `pop.Connection` are now few and far between:

```
$ grep -rl pop.Connection  pkg | grep -v -E 'pkg/(db|migrate|models|testdatagen|testingsuite)'
pkg/auth/authentication/client_cert.go
pkg/cli/dbconn.go
pkg/appcontext/context.go
pkg/telemetry/db.go
pkg/handlers/contexts.go
```
## Reviewer Notes

Each commit is relatively small and could be reviewed separately. The poptestsuite conversion commit is the largest.

## Setup

The usual server tests

* [ ] Code follows the guidelines for [Logging](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide#logging)
  * [ ] Secure migrations have been tested following the instructions in our [docs](https://transcom.github.io/mymove-docs/docs/dev/contributing/database/Database-Migrations#secure-migrations)

## References

The Jira bot will link for us


felipe-lee (commented):
In general, it seems like we might be able to get rid of most of these `AppContextForTest` definitions except for the auth one, but that one can be replaced with `AppContextWithSessionForTest`.
image

I think I misunderstood you previously in regards to the linter. We only wrote it to look for structs that had `*pop.Connection`. Based on this PR, looks like we should be scanning in function params too. Is that the case? If so we can edit the linter some more to get it there.

The linter currently only allows `*pop.Connection` in the `appcontext` package, but can easily be expanded to other packages. We didn't have it skip `models` because apparently we don't use it in there (as a struct anyways). So with that in mind, here is what is flagged with the linter as is (meaning only usages in structs):
image

We can expand it to start excluding the `testingsuite` package, so struct-wise, that would only leave the `pkg/handlers/contexts.go`. That one is kind of the base of everything else though right? Meaning it should also be excluded? 

I've gotten about 2/3 of the way through haha. I'll keep reviewing tomorrow.

ahobson:
> In general, it seems like we might be able to get rid of most of these `AppContextForTest` definitions except for the auth one, but that one can be replaced with `AppContextWithSessionForTest`. 

You are right! I must have missed cleaning up/removing some. Thank you!

> 
> I think I misunderstood you previously in regards to the linter. We only wrote it to look for structs that had `*pop.Connection`. Based on this PR, looks like we should be scanning in function params too. Is that the case? If so we can edit the linter some more to get it there.
> 

Removing it from structs is most important, so I think that's a great first milestone! We could go back and restrict the usage to outside the packages I mention above, but again, that can be done in the future.

> The linter currently only allows `*pop.Connection` in the `appcontext` package, but can easily be expanded to other packages. We didn't have it skip `models` because apparently we don't use it in there (as a struct anyways). So with that in mind, here is what is flagged with the linter as is (meaning only usages in structs): image
> 
> We can expand it to start excluding the `testingsuite` package, so struct-wise, that would only leave the `pkg/handlers/contexts.go`. That one is kind of the base of everything else though right? Meaning it should also be excluded?

Yeah, the testingsuite is needed for, uh, tests. :) and yeah, `pkg/handlers/context.go` is the one place to hold onto the connection because all handlers get their connections from there.

I think the alternative would be some middleware that puts the db connection in the request context, but I'm not sure that's so much better. It makes test setup slightly harder, too.

felipe-lee:
> > In general, it seems like we might be able to get rid of most of these `AppContextForTest` definitions except for the auth one, but that one can be replaced with `AppContextWithSessionForTest`.
> 
> You are right! I must have missed cleaning up/removing some. Thank you!

No problem 😄 

> > I think I misunderstood you previously in regards to the linter. We only wrote it to look for structs that had `*pop.Connection`. Based on this PR, looks like we should be scanning in function params too. Is that the case? If so we can edit the linter some more to get it there.
> 
> Removing it from structs is most important, so I think that's a great first milestone! We could go back and restrict the usage to outside the packages I mention above, but again, that can be done in the future.

My focus this sprint is these changes in general so I can see what it takes to update the linter since now.

> > The linter currently only allows `*pop.Connection` in the `appcontext` package, but can easily be expanded to other packages. We didn't have it skip `models` because apparently we don't use it in there (as a struct anyways). So with that in mind, here is what is flagged with the linter as is (meaning only usages in structs): image
> > We can expand it to start excluding the `testingsuite` package, so struct-wise, that would only leave the `pkg/handlers/contexts.go`. That one is kind of the base of everything else though right? Meaning it should also be excluded?
> 
> Yeah, the testingsuite is needed for, uh, tests. :) and yeah, `pkg/handlers/context.go` is the one place to hold onto the connection because all handlers get their connections from there.
> 
> I think the alternative would be some middleware that puts the db connection in the request context, but I'm not sure that's so much better. It makes test setup slightly harder, too.

I'm good with keeping it as is. Just wanted confirmation to know what changes to make to the linter. Thank you!

felipe-lee (approved):
Thanks for making those changes! LGTM

pkg/services/dbtools/table_from_slice_creator_test.go
```diff
@@ -100,29 +100,21 @@ func (suite *DBToolsServiceSuite) TestCreateTableFromSlicePermTable() {
...
-			err = tx.Order("name").All(&testStructs)
+			err = txnAppCtx.DB().Order("name").All(&testStructs)
 			suite.NoError(err)
 			suite.Len(testStructs, 3)
 			for i, testStruct := range testStructs {
 				suite.Equal(validSlice[i], testStruct)
 			}
 			return nil
 		})
+		suite.NoError(txnErr)
```
gidjin, line 117: Thank you for removing the lint exception and checking the error in the test! 🎉

pkg/testdatagen/make_payment_request_to_interchange_control_number.go
```diff
@@ -24,7 +25,10 @@ func MakePaymentRequestToInterchangeControlNumber(db *pop.Connection, assertions
 		log.Panic(fmt.Errorf("Errors encountered creating random sequencer: %v", err))
 	}
 
-	icn, err := icnSequencer.NextVal()
+	// for now, hack together an appcontext so we don't have to change
```
gidjin, line 28: Is this a hack? I could see this being okay for cases where you didn't use / need logger or session. Maybe just session.

gidjin (approved):
Lots of changes, but LGTM. 🚀

pkg/testdatagen/make_payment_request_to_interchange_control_number.go
felipe-lee: Yeah I feel like it's valid to create an AppContext without a logger or session. I think what he means in this case though is that ideally the function that this lives inside of would be getting an AppContext instance rather than a pop.Connection, so we wouldn't need to be creating an instance within the function.

PR #7653, opened 2021-11-02
[MB-9796] Small query change to allow renaming the duty_stations table without downtime

mboren:
## Description
This is a small tweak to the query that populates the duty location search results. 
It shouldn't have any visible effect yet, but it sets up for a future migration.

We will be adding a migration that renames the duty_stations table. It will look something like this:
```sql
ALTER TABLE duty_stations RENAME TO duty_locations;
CREATE VIEW duty_stations AS SELECT * FROM duty_locations;
```
Creating the view allows us to do the migration before any queries have been updated, in theory.
In practice, I ran into one small issue that is fixed by this PR.
Pasted Graphic 117

The problem is that Postgres is not quite as good at figuring things out for views as it is for tables, so when this query looks at the `duty_stations` view, we'll get an error because Postgres can't tell that `duty_stations.provides_services_counseling` will not change within any of the groups.

## Reviewer Notes
Will this change the results of the query in a way that I'm not anticipating?

## Setup

- Create a new customer, 
- go through the onboarding UI up until the point where you select a duty station,
-  play with the search to make sure the results make sense.
   - I recommend comparing against another environment to make sure things match.
   -  Make sure you test cases with abbreviations (for example, "naval air station" should give you results that start with "NAS").
- After trying out a few searches, finish customer onboarding just to make sure that nothing breaks.

## References

* [Slack thread with a little context on the migration this is setting up for](https://ustcdp3.slack.com/archives/CSGDM3NUW/p1635794050000500)


eamahanna (approved):
LGTM

Ryan-Koch (approved):

PR #7695, opened 2021-11-05
MB-10191 add additional auth logging

pearl-truss:
## Description

This PR adds more robust logging to our server side authentication code

## Reviewer Notes

Are there logs I'm missing? Logs I've added that won't actually be useful?

## Setup

`make server_run`
`make client_run`

It might be hard to see all of the logs that were added (especially those related to sessions) but it's possible to see some of auth logs locally if you go to the client and sign in without selecting the local sign in button.  You should see the logs in your server terminal

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-10191) for this change


pkg/auth/cookie.go
```diff
@@ -180,6 +180,7 @@ func SessionCookieMiddleware(globalLogger *zap.Logger, appnames ApplicationServe
 
 			existingSession := sessionManager.Get(r.Context(), "session")
 			if existingSession != nil {
+				logger.Info("Existing session not found")
```
duncan-truss, line 183: I think the session does exist in this case

duncan-truss:
This wasn't anything you added but I think we're logging user is active when we meant to say inactive here:
https://github.com/transcom/mymove/pull/7695/files#diff-be793c31d522efea61c667cf49ed11f865eb47751287243daf9d0a4383372eb4R619-R627

pearl-truss:
> This wasn't anything you added but I think we're logging user is active when we meant to say inactive here:
> 
> https://github.com/transcom/mymove/blob/9c754ae9ee57a918266d8ec5c7d8ed13c1925e50/pkg/auth/authentication/auth.go#L601-L608

Great catch! I just pushed a fix

pkg/auth/authentication/auth.go
```diff
@@ -138,6 +142,7 @@ func currentUser(appCtx appcontext.AppContext) (*models.User, error) {
 	userID := appCtx.Session().UserID
 	user, err := models.GetUser(appCtx.DB(), userID)
 	if err != nil {
+		appCtx.Logger().Error("Getting the user", zap.String("email", appCtx.Session().Email), zap.Error(err))
```
gidjin, line 145: I'm not sure if we can log the email address, I believe there are PII restrictions on such things. We may need to log their user id instead. Probably should confirm with @travelar

pkg/auth/authentication/auth.go
```diff
@@ -586,9 +602,11 @@ func (h CallbackHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	userIdentity, err := models.FetchUserIdentity(appCtx.DB(), openIDUser.UserID)
 	if err == nil { // Someone we know already
 		authorizeKnownUser(appCtx, userIdentity, h, w, r, landingURL.String())
+		appCtx.Logger().Info("Authorized and known user detected", zap.String("OID_User", openIDUser.UserID), zap.String("OID_Email", openIDUser.Email))
```
gidjin, line 605: I think `openIDUser.UserID` is okay, but we might want to ask about it when we ask about the email log above.

gidjin (commented):
We need to make sure that it's okay to log `email` and `openIDUser.UserID`

pkg/auth/authentication/auth.go
pearl-truss: Erik confirmed on [slack](https://ustcdp3.slack.com/archives/CP496B8DB/p1636396030193400), IDs are okay but the emails should be replaced

gidjin (approved):
Changes look good, I'm seeing some user logging locally.

PR #7697, opened 2021-11-08
[MB-9935] Add NTS SAC and TAC to GHC orders definition and model to payload

hannaliebl:
## Description

Add NTS TAC and SAC codes to the Order API definition so that a GET request for a given order returns that information.

## Reviewer Notes

Did I miss another spot where this could be updated?

## Setup

I manually added an NTS TAC and SAC code to an existing order, and then used Postman to verify that I could use the GHC API to GET that oder and that those codes were on the response.

```sh
make server_run
make client_run
```

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9935) for this change

## Screenshots

Here's a screenshot of the NTS SAC and TAC being returned in Postman:
Screen Shot 2021-11-08 at 10 13 48 AM


reggieriser (commented):
The changes LGTM and I was able to see the NTS TAC/SAC returned in the GHC payload after setting some values for them in the DB.

I did see that there is an `Orders` definition used in `internal.yaml` and an `Order` definition in `support.yaml` and `prime.yaml`.  Do either of those need to change?  It's not clear to me.

jenniferlynparsons (commented):
Looks good to me as well. I don't have the answer to Reggie's questions, unfortunately. Worth bringing up at standup.

LeDeep:
> The changes LGTM and I was able to see the NTS TAC/SAC returned in the GHC payload after setting some values for them in the DB.
> 
> I did see that there is an `Orders` definition used in `internal.yaml` and an `Order` definition in `support.yaml` and `prime.yaml`. Do either of those need to change? It's not clear to me.



> The changes LGTM and I was able to see the NTS TAC/SAC returned in the GHC payload after setting some values for them in the DB.
> 
> I did see that there is an `Orders` definition used in `internal.yaml` and an `Order` definition in `support.yaml` and `prime.yaml`. Do either of those need to change? It's not clear to me.

I think the answer is "no" for `support.yaml` and `prime.yaml`. Since prime doesn't handle NTS shipments, they should never need to know that information. I don't think we even send the TAC/SAC codes to the prime.

For `internal.yaml`, worth a longer discussion, as I'm not sure either.

reggieriser:
> Since prime doesn't handle NTS shipments, they should never need to know that information. I don't think we even send the TAC/SAC codes to the prime.

I thought the prime could handle NTS shipments?  Isn't that why we have an external vendor flag on a shipment to differentiate between ones the prime handles and ones an external vendor handles?  I may be confused.

FWIW, it looks like the prime does get the TAC as [part of this structure](https://github.com/transcom/mymove/blob/9b1ccc784197c9a74a8937c5466af5e0a43c8074/swagger-def/prime.yaml#L1961-L1962) which seems to come back as part of the response when fetching or updating an MTO.  I haven't tested this, though, but that seems to be what the YAML is implying.

reggieriser (approved):
I'm going to go ahead and approve based on our conversation in standup yesterday about filing a separate story to investigate the other YAML files.

PR #7713, opened 2021-11-10
[MB-9796] Create postal code to GBLOC mapping table

mboren:
### What I did
I'm creating a table to map ZIP codes to GBLOCs. It is based on [ZIP CODE TO GBLOC data.csv](https://drive.google.com/file/d/1igfhDGe4rs-B2jOT2JpiVOmq-FNWW_Ly/view?usp=sharing), which was sent to us by USTC.

I also threw in a script to get a few ZIP codes for a given GBLOC, to help when setting up a move.

This addresses one part of MB-9796. The rest of the model changes will occur in separate PRs.

### Feature context
To support non-installation locations for PO12, we're changing the way we assign GBLOCs to moves.
The current method is to get the GBLOC from the transportation office of the origin duty location.
New method:
- For services counseling queue: use old method to get GBLOC (transportation office of origin duty location)
- For TIO/TOO queues:
   - If Marines: GBLOC is USMC
   - Otherwise: Get the postal code from the pickup address of the first shipment and plug it into the `postal_code_to_gbloc` table (which is what i'm creating in this PR)
   
## Reviewer Notes
We're going to be using GBLOC in a few places, and it's starting to feel a little weird to use the string value instead of a foreign key to a GBLOC table. My intuition is that adding the extra table will just add query complexity without getting us anything useful (that I can think of), but I wanted to call this out in case other people had thoughts about it.

To generate the migrations, I used some python to drop all the columns except `ZIP CODE` and `GBLOC` in `ZIP CODE TO GBLOC DATA.csv`, remove duplicates, and generate the migration text. I think this is a one-off, so I didn't make a Go script to add to the repo. As I'm writing this up though, I realize I probably should have just done it in Go to make this easier to review. The code I used is in a snippet below. Let me know if my methods need clarifying.
## Setup
This will only affect the database, and nothing uses this table yet, so it can only be tested locally.

To verify that the stuff in the DB matches the original data, we can export the table from the DB as a CSV and compare it against `ZIP CODE TO GBLOC DATA.csv`.

First, run migrations
```
make db_dev_migrate
```
Then plug the following command into `psql-dev`:
```sql
\copy (select postal_code, gbloc from postal_code_to_gbloc order by postal_code, gbloc) to 'postal_code_to_gbloc_export.csv' CSV;
```
Which should look like this:
```
$ psql-dev
psql (14.0, server 12.7 (Debian 12.7-1.pgdg100+1))
Type "help" for help.

dev_db=# \copy (select postal_code, gbloc from postal_code_to_gbloc order by postal_code, gbloc) to 'postal_code_to_gbloc_export.csv' CSV;
```

Then run a lil python to get rid of extra columns and duplicate rows in the CSV:
```python
import csv

with open('ZIP CODE TO GBLOC DATA.csv', newline='') as f:
    reader = csv.DictReader(f)
    rows = set()
    for row in reader:
        rows.add(','.join((row['ZIP CODE'], row['GBLOC'])))

rows = sorted(rows)
text = '\n'.join(rows) + '\n'

with open('outfile.csv', 'w') as f:
    f.write(text)
```

Now check and see if they match:
```sh
diff outfile.csv postal_code_to_gbloc_export.csv
```
If you get any output from `diff`, then I messed something up!

  * [x] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://transcom.github.io/mymove-docs/docs/dev/contributing/database/Database-Migrations#zero-downtime-migrations))
  * [x] Have been communicated to #g-database
  
## References


migrations/app/schema/20211109231452_create_zip_to_gbloc_table.up.sql
```diff
@@ -40855,3 +40858,5 @@ VALUES
 ('83127','KKFA'),
 ('83128','KKFA'),
 ('83414','KKFA');
+
+CREATE INDEX postal_code_to_gbloc_postal_code_idx ON postal_code_to_gbloc (postal_code);
```
Ryan-Koch: Since this is being set as a primary key I think we can get away with out this line. Postgres should be creating an index for primary key values by default. More on that here in the primary key section: https://www.postgresql.org/docs/current/sql-createtable.html . The key take away is this line: `Adding a PRIMARY KEY constraint will automatically create a unique btree index on the column or group of columns used in the constraint.`

Ryan-Koch (approved):
I made a comment about the index/primary key situation, but otherwise this seems to run okay and populate the data 🙌

migrations/app/schema/20211109231452_create_zip_to_gbloc_table.up.sql
mboren: That might explain why I couldn't measure a performance difference between creating the index before or after the insert, even though the internet assured me it was a thing. I'll remove the redundant index.

pkg/models/postal_code_to_gbloc.go
```diff
@@ -0,0 +1,10 @@
+package models
+
+// PostalCodeToGBLOC is a mapping from Postal Codes to GBLOCs
+type PostalCodeToGBLOC struct {
+	PostalCode string `db:"postal_code"`
+	GBLOC      string `db:"gbloc"`
+}
```
reggieriser, line 7: Should this have a `CreatedAt` and `UpdatedAt` to match the fields of the database?  I'm not sure that it hurts anything if this is read-only and we don't need those timestamps for anything, but just wanted to make sure that was intentional since we usually include them in the model.

reggieriser (commented):
I added an inline comment but had a couple of comments about the large migration file that GitHub wouldn't seem to let me add inline:
- Did you consider doing a `COPY FROM` instead of the `INSERT` for the 40k+ rows?  We've done that in the past for large inserts and have seen a pretty big increase in migration speed, but it may not have as much impact here.  I tried running the current migration locally in isolation and it took about 1.4 seconds on my machine.
- Nit: Looks like there's a tab vs. spaces thing happening in the `CREATE TABLE` that's messing with the indentation.

Otherwise, LGTM.  I ran the migration and the bash script fine and was able to successfully compare the two files as described.

dsspence:
It seems like postal code and GBLOC have a fixed length, it could be worth limiting the data type length to those maximums.  Another option could be to keep the datatypes and add a check constraint since the character(n) or varchar(n) don't seem to have real performance implications in Postgres https://www.postgresql.org/docs/9.1/datatype-character.html

pkg/models/postal_code_to_gbloc.go
mboren: Not intentional, I'll add those. Thanks.

mboren:
Thanks for all the feedback!

@dsspence  I did consider using a fixed length and decided not to because it seemed like it wasn't being done for these types of columns in other places and there wasn't a performance impact. But, on reflection, I think I will add length constraints because this table won't ever store user input, so it's probably useful to have the additional checks when we insert.

@reggieriser I suppose I might as well use the COPY FROM format here. It is pretty fast, but we do run these migrations a lot.

mboren:
@reggieriser (just in case you're curious) using `COPY FROM stdin` speeds things up slightly (maybe 0.1 or 0.2s on my machine). Interesting note: the speedup wasn't noticeable when i just used what i got from pg_dump, but, when i removed the timestamp columns, it's a little faster. Seems like file IO might be a bottleneck more than database operations in this case. The migration with timestamps was 2.8MB, and without them it's 440KB.

reggieriser (approved):
LGTM!  Ran it again and verified that the CSV and the database dump still matched.  Thanks for making the adjustments.

PR #7741, opened 2021-11-15
MB-9936 Added SC order updating functionality to ghc.yaml

jenniferlynparsons:
## Description

The API can be used to update TAC and SAC accounting codes and changes are saved to the database.

## Reviewer Notes

I'm not as familiar with the backend, so there may be something I missed. 

## Setup

Make sure data is up to date:

```
make db_dev_e2e_populate
```

Build and run the office app:

```
make server_run
make office_client_run
```

Run Postman with the `Updates an order (performed by a services counselor)` PATCH command and confirm that the move that is returned has the expected HHG_TAC, HHG_SAC, NTS_TAC, and NTS_SAC.

* [ ] Code follows the guidelines for [Logging](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide#logging)
  * [ ] Secure migrations have been tested following the instructions in our [docs](https://transcom.github.io/mymove-docs/docs/dev/contributing/database/Database-Migrations#secure-migrations)

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9936) for this change

### DIFF
--- a/pkg/handlers/ghcapi/orders_test.go
+++ b/pkg/handlers/ghcapi/orders_test.go
@@ -736,6 +736,10 @@ func (suite *HandlerSuite) TestCounselingUpdateOrderHandler() {
 		suite.Equal(*body.IssueDate, ordersPayload.DateIssued)
 		suite.Equal(*body.ReportByDate, ordersPayload.ReportByDate)
 		suite.Equal(*body.OrdersType, ordersPayload.OrderType)
+		suite.Equal(body.Tac, ordersPayload.Tac)
+		suite.Equal(body.Sac, ordersPayload.Sac)
+		suite.Equal(body.NtsTac, ordersPayload.NtsTac)
+		suite.Equal(body.NtsSac, ordersPayload.NtsSac)
 	})
 
 	suite.Run("Returns a 403 when the user does not have Counselor role", func() {
--- a/pkg/handlers/ghcapi/orders_test.go
+++ b/pkg/handlers/ghcapi/orders_test.go
@@ -265,6 +265,8 @@ func (suite *HandlerSuite) TestUpdateOrderHandlerWithAmendedUploads() {
 			OriginDutyStationID:   handlers.FmtUUID(originDutyStation.ID),
 			Tac:                   handlers.FmtString("E19A"),
 			Sac:                   handlers.FmtString("987654321"),
+			NtsTac:                handlers.FmtString("E19A"),
+			NtsSac:                handlers.FmtString("987654321"),
 			OrdersAcknowledgement: &ordersAcknowledgement,
 		}
 
@@ -305,6 +307,8 @@ func (suite *HandlerSuite) TestUpdateOrderHandlerWithAmendedUploads() {
 		suite.Equal(body.DepartmentIndicator, ordersPayload.DepartmentIndicator)
 		suite.Equal(body.Tac, ordersPayload.Tac)
 		suite.Equal(body.Sac, ordersPayload.Sac)
+		suite.Equal(body.NtsTac, ordersPayload.NtsTac)
+		suite.Equal(body.NtsSac, ordersPayload.NtsSac)
 		suite.NotNil(ordersPayload.AmendedOrdersAcknowledgedAt)
 
 		reloadErr := suite.DB().Reload(&move)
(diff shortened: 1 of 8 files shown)


swagger-def/ghc.yaml
```diff
@@ -2689,6 +2689,30 @@ definitions:
...
+        minLength: 4
+        maxLength: 4
+        example: 'F8J1'
+        x-nullable: true
+      nts_sac:
+        type: string
+        title: NTS SAC
+        example: 'N002214CSW32Y9'
+        x-nullable: true
     required:
```
hannaliebl, line 2716: Does `tac` need to be added here? I think according to the ticket, HHG TAC is always required? And same in the `UpdateOrderPayload`, I think. I don't actually know the best way to do this, though: enforce its existence here, or in the service object?

jenniferlynparsons: I think it doesn't need to be required here because at different points in the move (SC vs TOO) the TAC may not be available and is an optional field for some folks. I think we should stick with FE validation for now unless that scenario changes.

pkg/handlers/ghcapi/orders_test.go
```diff
@@ -396,6 +402,8 @@ func (suite *HandlerSuite) makeUpdateOrderHandlerSubtestData() (subtestData *upd
 		OriginDutyStationID: handlers.FmtUUID(originDutyStation.ID),
 		Tac:                 handlers.FmtString("E19A"),
 		Sac:                 handlers.FmtString("987654321"),
+		NtsTac:              handlers.FmtString("E19A"),
+		NtsSac:              handlers.FmtString("987654321"),
```
reggieriser, line 406: I see that this `make*` method and the one for the SC below are used in some tests that return a successful HTTP 200 status.  It doesn't look like we check that the TACs/SACs were saved (or not saved) in all of those.  One of those is for a TIO user editing orders -- should a TIO be able to change TACs/SACs?  I can't recall, but either way we should probably validate the expected behavior for the TACs/SACs occurred.

swagger/ghc.yaml
```diff
@@ -2770,6 +2770,30 @@ definitions:
...
+        minLength: 4
+        maxLength: 4
+        example: F8J1
+        x-nullable: true
+      sac:
+        type: string
+        title: HHG SAC
+        example: N002214CSW32Y9
+        x-nullable: true
+      nts_tac:
```
reggieriser, line 2785: I think these should be camelCase instead?  It looks like we're inconsistent in this YAML file, but the support and prime ones that are linted do use camelCase.  The [API Style Guide](https://transcom.github.io/mymove-docs/docs/api/guides/api-style-guide) seems to indicate camelCase for most stuff.

reggieriser (commented):
I tested out both the counselor office update endpoint and the regular office update endpoint and was able to change an NTS TAC. 🎉   I did note a couple things to look into.

I was also curious how you would ever delete an NTS TAC through the API.  I tried sending a null for the NTS TAC but that didn't delete what was there -- presumably because the null is what you get in the handler if they just don't include the field (i.e., we don't want it changed).  I think this is probably a bigger question beyond this PR, though.

reggieriser (approved):
Thanks for making the changes!  LGTM.  I did have one little nit.

One thing I mentioned in a previous comment that I think we'll have to deal with at some point is how we clear out a nullable field (like the NTS TAC/SAC) via the API.  Passing a null for the field value doesn't do it.

swagger/ghc.yaml
```diff
@@ -2733,12 +2733,12 @@ definitions:
         title: SAC
```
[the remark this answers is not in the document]
jenniferlynparsons: I'd like this to be, but changing this here and now effects the frontend codebase and breaks the build. I've left things as is for now for these fields.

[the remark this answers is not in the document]
reggieriser: OK, sounds good.  Didn't think about a title change affecting the front end.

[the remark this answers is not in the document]
jenniferlynparsons: Yeah, I could have tackled it here, but it would have taken this ticket way out of scope. I'll make a note on the related form ticket to change the naming.

PR #7742, opened 2021-11-16
[MB-9796] Part 3: Insert non-installation locations into duty_stations

mboren:
## Description
In this PR, we're adding non-installation duty locations to the `duty_stations` table.
There is a non-installation location for approximately every unique `(zip, city, state)` in the United States.
They come from [ZIP CODE TO GBLOC DATA](https://docs.google.com/spreadsheets/d/1CtE8bwOhVIFr2L0IEfOKWu72bVi6ISf2/). Note that we're using `PREFERRED LAST LINE CITY STATE NAME` for the city name, and not `CITY STATE NAME`. There are significantly more records in the data file than there are in our migration, as the data file includes multiple more granular place names for each `PREFERRED LAST LINE CITY STATE NAME`.

This will allow users to select non-installation locations by ZIP code.

However, the application doesn't fully support these locations yet because they don't have transportation offices.
I added kind of a hack to prevent things from crashing during onboarding (and a story to solve the issue in a better way), but until we update routing, the system won't know which GBLOC to use for moves originating from these locations.

This onboarding issue that I fixed happens because we try to look up contact information for the transportation office, which is `NULL`. This contact information gets included in notification messages. This contact information isn't really meaningful for non-installation moves, so I just included a placeholder for now.

Until we update routing, the system will also be unable to figure out what GBLOC to use for moves originating from these locations.

## Reviewer Notes
To generate the migration, I used [some SQL](https://gist.github.com/mboren/8bd3648878fb76d5a703ae5c703724cc) to load the CSV we were given into a temporary table and setup the new records we needed to add in other temporary tables. Then I used `pg_dump` on the new tables and did a little bit of cleanup to get a usable migration file.

Fun fact: I noticed that replacing the `created_at` and `updated_at` timestamps that we get from `pg_dump` with `NOW()` reduces the file size a lot and speeds up the migration by ~8%.

## Setup

Run migrations and make sure nothing breaks
```sh
make db_dev_migrate
```
Go through customer onboarding until you get to a duty station select box, try searching for some ZIP codes, and make sure you're getting what you would expect based on [ZIP CODE TO GBLOC DATA](https://docs.google.com/spreadsheets/d/1CtE8bwOhVIFr2L0IEfOKWu72bVi6ISf2/).
Select a non-installation location (something with a ZIP code in its name) for the origin duty location.
Finish onboarding and submit the move, just to make sure that nothing breaks.

### Check that the right duty locations provide services counseling
The most important thing to verify is that duty locations that provide services counseling are correct.
```sql
create view services_counseling_postal_codes as
select a.postal_code
from duty_stations
     join addresses a ON duty_stations.address_id = a.id
where affiliation is not null and provides_services_counseling;

-- Find non-installation locations that don't provide services counseling, and share a ZIP code with an installation location that provides services counseling
-- Should have zero results
select * from duty_stations
    join addresses a on duty_stations.address_id = a.id
where duty_stations.affiliation is null
  and duty_stations.provides_services_counseling is false
  and a.postal_code in
      (select * from services_counseling_postal_codes);

-- Find non-installation locations that provide services counseling, and share a ZIP code with an installation location that provides services counseling
-- Should have 95 results (1 more than the installation locations that provide services counseling because Tinker AFB and Oklahoma City share the same ZIP code)
select * from duty_stations
    join addresses a on duty_stations.address_id = a.id
where duty_stations.affiliation is null
  and duty_stations.provides_services_counseling
  and a.postal_code in
    (select * from services_counseling_postal_codes);

-- Find non-installation locations that provide services counseling, and don't share a ZIP code with an installation location that provides services counseling
-- Should have no results
select * from duty_stations
      join addresses a on duty_stations.address_id = a.id
where duty_stations.affiliation is null
  and duty_stations.provides_services_counseling
  and a.postal_code not in
      (select * from services_counseling_postal_codes);
```

### method for generating the migration
- Download this script: https://gist.github.com/mboren/8bd3648878fb76d5a703ae5c703724cc
- Download [ZIP CODE TO GBLOC DATA.csv](https://drive.google.com/file/d/1igfhDGe4rs-B2jOT2JpiVOmq-FNWW_Ly/view)
- Run the script
```sh
make db_dev_reset
make db_dev_migrate
psql --variable "ON_ERROR_STOP=1" postgres://postgres:mysecretpassword@localhost:5432/dev_db -f generate_non_installation_location_records.sql
```
This will create 3 new tables:
`temp_zip_to_gbloc`: this holds all the data from "ZIP TO GBLOC DATA.csv", with a few column names changed for convenience.
`temp_duty_stations`: holds all the non-installation duty locations that need to be created
`temp_addresses`: holds new addresses for the the new duty locations

To generate the migration in this PR, we just have to dump `temp_duty_stations` and `temp_addresses`
```sh
pg_dump -h localhost -U postgres dev_db --table temp_duty_stations -t temp_addresses --section=data > non_installation_duty_stations.sql
```
To transform this into the final migration, I removed the postgres config changes at the start of the file, removed the `temp_` prefix from  `temp_duty_stations` and `temp_addresses`, and replaced all the timestamps with `NOW()`
If you remove the UUIDs you should be able to diff this against my migration.

## References

## Screenshots
Here's a bunch of different search results, mostly based on stuff that was in the designs.
image
image
image
image
image
image

### DIFF
--- a/cypress/integration/office/txo/tooFlows.js
+++ b/cypress/integration/office/txo/tooFlows.js
@@ -199,7 +199,7 @@ describe('TOO user', () => {
         .type('JB McGuire-Dix-Lakehurst')
         .get('[class*="-menu"]')
         .find('[class*="-option"]')
-        .eq(1)
+        .eq(3)
         .click(0, 0);
 
       cy.get('input[name="issueDate"]').click({ force: true }).clear().type('16 Mar 2018');
--- a/migrations/app/migrations_manifest.txt
+++ b/migrations/app/migrations_manifest.txt
@@ -682,3 +682,4 @@
 20211101220502_add_nts_shipment_columns.up.sql
 20211109231452_create_zip_to_gbloc_table.up.sql
 20211112193224_support_noninstallation_locations.up.sql
+20211112205920_add_noninstallation_locations.up.sql
--- a/pkg/models/duty_station.go
+++ b/pkg/models/duty_station.go
@@ -73,6 +73,10 @@ func FetchDSContactInfo(db *pop.Connection, dutyStationID *uuid.UUID) (*DutyStat
 	if err != nil {
 		switch err {
 		case sql.ErrNoRows:
+			// This is temporary. Non-installation duty locations do not have transportation offices
+			// so we can't look up their contact information. This default here allows us to avoid
+			// crashing when we're generating notifications. In the future, we need to update all the
+			// text relating to transportation office contact info because it's not meaningful for most moves.
 			return &DutyStationTransportInfo{Name: "Unknown Office", PhoneLine: "555-555-5555"}, nil
 		default:
 			return nil, err
--- a/pkg/models/duty_station.go
+++ b/pkg/models/duty_station.go
@@ -1,6 +1,7 @@
 package models
 
 import (
+	"database/sql"
 	"time"
 
 	"github.com/gobuffalo/pop/v5"
@@ -70,7 +71,12 @@ func FetchDSContactInfo(db *pop.Connection, dutyStationID *uuid.UUID) (*DutyStat
 		LIMIT 1`
 	err := db.RawQuery(query, *dutyStationID).First(&DSTransportInfo)
 	if err != nil {
-		return nil, err
+		switch err {
+		case sql.ErrNoRows:
+			return &DutyStationTransportInfo{Name: "Unknown Office", PhoneLine: "555-555-5555"}, nil
+		default:
+			return nil, err
+		}
 	} else if DSTransportInfo.Name == "" || DSTransportInfo.PhoneLine == "" {
 		return nil, ErrFetchNotFound
 	}
@@ -95,7 +101,7 @@ func FetchDutyStationByName(tx *pop.Connection, name string) (DutyStation, error
 func FindDutyStations(tx *pop.Connection, search string) (DutyStations, error) {
 	var stations DutyStations
 
-	sql := `
+	sqlQuery := `
 with names as (
 (select id as duty_station_id, name, similarity(name, $1) as sim
 from duty_stations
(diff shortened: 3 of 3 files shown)


mboren:
I did a git goof and force pushed to resolve it. My apologies if you've got this branch checked out locally.

leia-sefkin (approved):
did a final code review, another run of local tests, and went through the move flow manually, and ran the queries you listed in the PR. everything looks good to me, seems that we're safe to ship it!

reggieriser (commented):
Looking at the migration as a DB reviewer ... it worked fine for me and took about 4 seconds or so to run locally.  I'm wondering a few things:
- It looks like we insert some columns like `street_address_2`, `street_address_3`, and `affiliation` that I think (unless I missed it) are always null.  Could we just remove those columns from the insert entirely?  We should end up with the same result and it'd be less data the migration had to process.  Not sure if that speeds things up much, but it hopefully wouldn't hurt.
- We have a lot of indexes on `duty_stations`, including a `gin` one.  Don't see any indexes on addresses other than the implied primary key one.  I know I've seen some speedups in the past by temporarily dropping indexes before a big import of data and reapplying after.  Not sure if that's worth it here -- or ramifications when dropping on a table already in use -- but just thought I'd mention it.
- Does this seem to have any significant performance impact to the duty station lookup in the customer app?  We query duty stations a lot there since we match as you type, and we just went from 500 or so to 41k+ duty stations with this migration.  Hopefully our indexes were already sufficient to handle the additional records.

mboren:
Love the idea of dropping those NULL fields, I'll definitely do that.

Great question about the performance impact on our searches. I spent more time thinking about the migration performance than the runtime performance, which is uh probably the wrong way around!
The search we're doing is looking for "similar" text, so I'm not sure how much the index can even help us. I can look into that. It's been running fine locally, but I wouldn't be surprised if it blows up when hit by multiple users at once.
We're kind of abusing full text search here when we were only asked for lookup by ZIP because it simplified both design and development. I assume ZIP search would be way more indexable, so there is probably a lot of room for performance improvement, but at the cost of additional design and implementation work.

Ryan-Koch (approved):

PR #7751, opened 2021-11-16
MB 10148 upload docs flow

kennethchow-truss:
## Description

Previously, when a Prime User was on the upload docs page, successful actions and cancelling an upload lead them back to the home page. With this change, the application will move the user back to the move details page related to the upload.

Tests were also added 

## Setup

To run the save: 

1. Run both front end and back end 
2. Go to `officelocal:3000`
3. Click on a move that has payment requests
4. Scroll to the bottom
5. Click upload documents
6. Upload a file of your choice
7. Click Save
8. Observe that the page you are on now is the move details page with a success message

To test the cancel button:
1. Repeat steps 1-5 of the last section
2. Click cancel
3. Observe that the page you are on now is the move details page


donaldthai (approved):
🎉

YanZ777 (approved):
Looks good to me!

PR #7752, opened 2021-11-17
MB-9925 Move appcontext to be in request context

felipe-lee:
## Description

Proof of concept of using `AppContext` in middleware instead of as a field on the `handlerContext` struct.

Main things to focus on are the first commit: https://github.com/transcom/mymove/pull/7752/commits/3125c5899262e17a9fd157dc801ace59a00abdd6

and the changes to `contexts.go`: https://github.com/transcom/mymove/pull/7752/commits/c97387015ad3f7aceb743b184cf6b7e5b597d2b3#diff-218b8e3f456a4d95176b572dead5bf50215b43f5b1823f6013eb899a1fec4c3c 

Tests are broken because the setup for tests that need the handler would now need to set up the appCtx in the request context, but I have a comment below highlighting how they could be fixed.

### DIFF
--- a/pkg/auth/authentication/auth_test.go
+++ b/pkg/auth/authentication/auth_test.go
@@ -11,6 +11,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/transcom/mymove/pkg/appcontext"
+
 	"github.com/stretchr/testify/mock"
 
 	"github.com/transcom/mymove/pkg/handlers"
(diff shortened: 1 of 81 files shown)


pkg/auth/authentication/auth_test.go
```diff
@@ -180,15 +182,16 @@ func (suite *AuthSuite) TestAuthorizationLogoutHandler() {
 		IDToken:         fakeToken,
 		Hostname:        OfficeTestHost,
 	}
-	ctx := auth.SetSessionInRequestContext(req, &session)
+
+	initialCtx := auth.SetSessionInRequestContext(req, &session)
+	ctx := appcontext.NewContext(initialCtx, suite.AppContextForTest())
+
```
felipe-lee, line 188: This would be something that would change for any tests that need a handler, instead of passing appCtx to the handler context creation, we would need to set it into the request context. I only did it for this one test to show what it might look like.

ahobson: I guess I need to write more about appcontext, but for me the idea is that it takes things from the Context and then passes it down to our lower level APIs. To me, the handler is the ideal place to do that and not middleware because middleware doesn't give us any type level checking that the middleware is setup correctly.

felipe-lee: Yeah I think more docs would be great. Unless we already have some and I missed them.

PR #7761, opened 2021-11-18
MB-10469: Use custom net/http go opentelemetry implementation for improvements, standardize metric names

ahobson:
## Description

We need additional instrumentation to [net/http](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/net/http) that existing open PRs would provide, but they aren't merged, apparently because the opentelemetry folks don't really want to maintain the code.

## Reviewer Notes

The [external repo has the custom net/http otel code](https://github.com/trussworks/otelhttp/pull/1)

I deployed this to loadtesting environment and it seemed to work well


duncan-truss:
When we have the time should we fork the repo or is that not desirable?

ahobson:
> When we have the time should we fork the repo or is that not desirable?

I couldn't decide! The thing is that the contrib repo has *lots* of other things in it to introspect packages we don't use, so it didn't seem like the right call.

We're also in a period of (I think) rapid iteration. Perhaps once things settle down we could open source just this bit?

gidjin (commented):
Looks good, though had questions around the ramifications of importing the Apache License.

felipe-lee (commented):
Mostly LGTM! Just one thing, I think the version of `semconv` also needs to be updated in [pkg/telemetry/init.go](https://github.com/transcom/mymove/blob/9659d9e8b98ee23c5f12a929e6cf7bec57504487/pkg/telemetry/init.go#L8).

gidjin (commented):
LGTM now that the library code is in another repo. I agree with Felipe the `pkg/telemetry/init.go` file needs to have the version for `semconv` updated to match the others.

ahobson:
> Mostly LGTM! Just one thing, I think the version of `semconv` also needs to be updated in [pkg/telemetry/init.go](https://github.com/transcom/mymove/blob/9659d9e8b98ee23c5f12a929e6cf7bec57504487/pkg/telemetry/init.go#L8).

Done

pkg/telemetry/init.go
```diff
@@ -176,7 +176,7 @@ func Init(logger *zap.Logger, config *Config) (shutdown func()) {
 	}
 
 	tracer = otel.Tracer(tracerName)
-	meter = global.Meter(metricName)
+	//meter = global.Meter(metricName)
```
felipe-lee, line 180: Just to make sure I'm understanding correctly, this change is making it so that instead of setting up this global metric provider for the repo at this level, we'll be setting up a specific one for the DB stats?

ahobson: It turns out we don't need to save the global meter as we can use the API to get it back wherever we need to. Same for the tracer, actually ...

Let me add another commit to clean that up. Thank you

felipe-lee (approved):

PR #7780, opened 2021-11-23
MB-9674 add layout css rules for min-width in the office app

Shkeating:
## Description

In its current state, the office app layout degrades to an unusable state when the screen becomes narrow. This aspect of the MilMove application will not have mobile users — as a result when our office app components were built they were built and styled to be desktop-only. 

This branch addresses that by introducing a style sheet into our office app's scss file. It tells the browser to stop having our components resize after the view port has come down to 1240px (this value was determined by when the app started to have compromised usability, caused by things that were happening like link targets touching or content overlapping). 

At this point, the app will overflow out of the viewport and x-scrolling is enabled.

## Reviewer Notes

I tested this out to make sure I caught everything that was breaking and I *think* I got it all. I also tried to ensure that this does not inhibit the experience anywhere. 

Right now, I am only loading this style sheet into the office app itself, not in Storybook. however, I have thought an easy way to reflect this in storybook would be simply loading in the style sheet on office-only components. I am contemplating introducing this change to storybook too, but would do so in a separate branch.

## Setup

```sh
make server_generate
make server_run
make office_client_run
```

peruse around all roles and let me know if I bonked anything

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9674?atlOrigin=eyJpIjoiODI5MWE5NDk5YTUxNDUwYmIwODljYjFiN2M1MGZjZTkiLCJwIjoiaiJ9) for this change

### BEFORE:
image

### AFTER: 
image


Shkeating:
holy heck it affected storybook?!?! ah

YanZ777 (approved):
Changes look good to me!

amanda5iu (approved):

PR #7798, opened 2021-11-29
MB-10497: Increase db pool size

ahobson:
## Description

Increase the pool size limit from 50, which is too small

## Reviewer Notes

This was tested in the load test environment

### DIFF
--- a/pkg/cli/dbconn.go
+++ b/pkg/cli/dbconn.go
@@ -81,12 +81,32 @@ const (
 	// SSLModeVerifyFull is the verify-full SSL Mode
 	SSLModeVerifyFull string = "verify-full"
 
+	// awsRdsT3SmallMaxConnections is the max connections to an RDS T3
+	// Small instance
+	//
+	// The T3 small instance has 2 GB
+	// https://aws.amazon.com/rds/instance-types/
+	//
+	// These docs say we can calculate the max connections
+	// https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Limits.html
+	//
+	// If correct it is
+	//
+	// LEAST({DBInstanceClassMemory/9531392}, 5000)
+	//
+	// DBInstanceClassMemory = 2147483648
+	// so 2147483648 / 9531392 = 225.3 which is less than 5000
+	//
+	// we deploy two containers for the AWS service, so divide that in
+	// half
+	// 225 / 2 =~ 110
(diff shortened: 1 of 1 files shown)


ahobson:
> Have you been looking at the reported DB metrics around how long our connections are being kept open for or blocked and waiting? I see there is a max connection lifetime that is currently unlimited https://github.com/gobuffalo/pop/blob/257e6c6f7d110a8048caf1325e0b4ca933ea3a1b/connection_details.go#L45.
> 
> Did you see at all in your reading that the database tends to like only 50% of it's max connections in use? It seems a little counterintuitive that requiring more idle connections would lead to more manageable load but that's what I was reading in a few places.

I started to try to gather some [db metrics](https://github.com/transcom/mymove/blob/master/pkg/telemetry/db.go) and I have been experimenting with other (db metrics)[https://console.amazonaws-us-gov.com/cloudwatch/home?region=us-gov-west-1#metricsV2:graph=~(view~'timeSeries~stacked~false~region~'us-gov-west-1~start~'2021-11-22T18*3a00*3a00.000Z~end~'2021-11-22T20*3a00*3a00.000Z~metrics~(~(~'ECS*2fAWSOTel*2fApplication~'db.waitduration~'OTelLib~'github.com*2ftranscom*2fmymove*2fdb)));query=~'*7bECS*2fAWSOTel*2fApplication*2cOTelLib*7d] in loadtest.

Since we have two containers, each container would use 50% of the db connections (I think).

Looking at charts, I think we max out the db CPU if both containers can have ~225 connections open.

I do feel sure 50 is too low. I'm not positive what the max should be. Should we set the limit even lower in the container to ensure we don't max out the DB CPU? I think we need to do some more experimentation, but it [looks to me like ~112 connections is stable](https://console.amazonaws-us-gov.com/cloudwatch/home?region=us-gov-west-1#metricsV2:graph=~(metrics~(~(~'ECS*2fAWSOTel*2fApplication~'db.open_connections~'OTelLib~'github.com*2ftranscom*2fmymove*2fdb))~view~'timeSeries~stacked~false~region~'us-gov-west-1~start~'2021-11-22T18*3a00*3a00.000Z~end~'2021-11-22T23*3a23*3a59.000Z~stat~'Maximum~period~300);query=~'*7bECS*2fAWSOTel*2fApplication*2cOTelLib*7d*20open_connections)

PR #7799, opened 2021-11-29
MB-10522: pin otel collector version, use aws ecr repo, use metrics config

ahobson:
## Description

In loadtest, the metrics stopped working, seemingly after an update to the otel collector version. Pinning to v0.14.0 and using the metrics config seems to have fixed it.

Then, there was a problem somewhere in The Cloud where AWS couldn't pull down the image from docker hub, so use an AWS ECR public image.

## Reviewer Notes

The [docs](https://github.com/aws-observability/aws-otel-collector/blob/main/docs/developers/ecs-demo.md) suggest we should use the metrics config "... to consume application metrics, traces (using OTLP and Xray) and container resource utilization metrics". It's really not clear to me how exactly it is different from the other provided ECS config, but this seems to work in the load test environment.

### DIFF
--- a/cmd/ecs-deploy/task_def.go
+++ b/cmd/ecs-deploy/task_def.go
@@ -615,7 +615,7 @@ func taskDefFunction(cmd *cobra.Command, args []string) error {
 		containerDefinitions = append(containerDefinitions,
 			&ecs.ContainerDefinition{
 				Name:      aws.String("otel-" + containerDefName),
-				Image:     aws.String("amazon/aws-otel-collector"),
+				Image:     aws.String("amazon/aws-otel-collector:v0.14.0"),
 				Essential: aws.Bool(true),
 				Command: aws.StringSlice([]string{
 					"--config=/etc/ecs/ecs-cloudwatch-xray.yaml",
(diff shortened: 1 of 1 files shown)


cmd/ecs-deploy/task_def.go
```diff
@@ -615,10 +615,10 @@ func taskDefFunction(cmd *cobra.Command, args []string) error {
 		containerDefinitions = append(containerDefinitions,
 			&ecs.ContainerDefinition{
 				Name:      aws.String("otel-" + containerDefName),
-				Image:     aws.String("amazon/aws-otel-collector"),
+				Image:     aws.String("public.ecr.aws/aws-observability/aws-otel-collector:v0.14.0"),
```
felipe-lee, line 618: Which version was the one causing issues? I see v0.15.0 and v0.14.1. Mainly wondering if we should pin to v0.14.1 or if that was the one that was causing issues.

felipe-lee: Was looking here at the versions: https://gallery.ecr.aws/aws-observability/aws-otel-collector

ahobson: I think it was v0.14.1 that was causing problems, I'm not sure. I know v0.14.0 works so I'd like to stay on that for a bit.

felipe-lee (approved):
LGTM

PR #7801, opened 2021-11-29
Kc mb 8572 diversion shipment

kennethchow-truss:
## Description

Adds the Diversion field to the move details and update shipment form. The diversion field is a boolean checkbox that is an attribute on a shipment

## Screenshots

Below is the new diversion field on the move details page 
image

Below is the new diversion field in the update shipment page
image


YanZ777 (commented):
Looks good to me, only question is if the tests should be updated or not.

YanZ777 (approved):
Approving as we agreed that tests were not in scope.

donaldthai (approved):

PR #7811, opened 2021-11-30
MB-9669: Ensure auth API calls are never cached

ahobson:
## Description

Looking at the logs from the most recent occurrence (see the comments in jira), there doesn't seem to be a log for the `GET /auth/login-gov` before the redirect back to `/auth/login-gov/callback`. This is a bit of a shot in the dark, but let's make sure none of the API calls is ever cached.

## Reviewer Notes

Not positive this will fix the bug, but it also seems like the right thing to do regardless. We install this middleware for all our other API endpoints.

I deployed this on the load test environment

### DIFF
--- a/cmd/milmove/serve.go
+++ b/cmd/milmove/serve.go
@@ -1007,6 +1007,7 @@ func serveFunction(cmd *cobra.Command, args []string) error {
 		},
 	)
 	authMux := root.PathPrefix("/auth/").Subrouter()
+	authMux.Use(middleware.NoCache(logger))
 	authMux.Use(otelmux.Middleware("auth"))
 	authMux.Handle("/login-gov", authentication.NewRedirectHandler(authContext, handlerContext, useSecureCookie)).Methods("GET")
 	authMux.Handle("/login-gov/callback", authentication.NewCallbackHandler(authContext, handlerContext, notificationSender)).Methods("GET")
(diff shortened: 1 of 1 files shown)


duncan-truss (approved):
I am curious where we can potentially remove the no cache directive elsewhere. I think there's a security requirement around not caching PII but I could be wrong.

ahobson:
> I am curious where we can potentially remove the no cache directive elsewhere. I think there's a security requirement around not caching PII but I could be wrong.

I don't think there's any PII in these HTTP requests, but I think, in theory, a client could currently cache the `GET /auth/login-gov`. I'm not positive about that, though

duncan-truss:
> > I am curious where we can potentially remove the no cache directive elsewhere. I think there's a security requirement around not caching PII but I could be wrong.
> 
> I don't think there's any PII in these HTTP requests, but I think, in theory, a client could currently cache the `GET /auth/login-gov`. I'm not positive about that, though

I agree it makes sense for the auth to never be cached and was trying to make heads or tails of why it may exist elsewhere where it could benefit our users to cache some things.

ahobson:
> > > I am curious where we can potentially remove the no cache directive elsewhere. I think there's a security requirement around not caching PII but I could be wrong.
> > 
> > 
> > I don't think there's any PII in these HTTP requests, but I think, in theory, a client could currently cache the `GET /auth/login-gov`. I'm not positive about that, though
> 
> I agree it makes sense for the auth to never be cached and was trying to make heads or tails of why it may exist elsewhere where it could benefit our users to cache some things.

Yeah, a good cache strategy could be a way to reduce load on the server, but wow, I think it's complicated and challenging to do right. It's probably farther down our list of priorities. :grinning:

felipe-lee (approved):
LGTM. At this point we use it almost everywhere. I would say we should just apply it to all using `root` (or maybe `site`, tbh I'm not sure how those two interact). But we don't have it enabled for `static`, and that's actually probably a good place to allow cache use so the way we're doing it now might be best.

ahobson:
> LGTM. At this point we use it almost everywhere. I would say we should just apply it to all using `root` (or maybe `site`, tbh I'm not sure how those two interact). But we don't have it enabled for `static`, and that's actually probably a good place to allow cache use so the way we're doing it now might be best.

Yeah, I didn't want to get too aggressive. I'd love to redo our routes anyway to be less confusing. Maybe someday.

PR #7774, opened 2021-11-22
MB-9677 TOO view origin duty location

leia-sefkin:
## Description

This PR adds the origin duty location column to the TOO queue

## Reviewer Notes

Did I miss any tests? Anything weird or wrong with other queues that use orders? 

## Setup

To run the tests, use the following commands: make server_test_standalone and make client_test.

To view the changes in the UI, open the office app and choose the TOO user role. On the move queue, you will see the new column and can search via name. 

## Code Review Verification Steps

* [ ] If the change is risky, it has been tested in experimental before merging.
* [ ] Code follows the guidelines for [Logging](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide#logging)
* [ ] The requirements listed in [Querying the Database Safely](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide/#querying-the-database-safely) have been satisfied.
* Any new migrations/schema changes:
  * [ ] Follow our guidelines for zero-downtime deploys (see [Zero-Downtime Deploys](https://transcom.github.io/mymove-docs/docs/dev/contributing/database/Database-Migrations#zero-downtime-migrations))
  * [ ] Have been communicated to #g-database
  * [ ] Secure migrations have been tested following the instructions in our [docs](https://transcom.github.io/mymove-docs/docs/dev/contributing/database/Database-Migrations#secure-migrations)
* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).

* [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9677) for this change

## Screenshots
Screen Shot 2021-11-29 at 3 34 34 PM

PR #7821, opened 2021-12-01
MB-9675 Allow service counselor work queue to be sorted by origin location

christopherhuii:
## Summary

This PR updates the service counselor work queue table. It replaces the "Destination duty location" with "Origin duty location".

This PR branches off of #7774. #7774 should be merged first.

### Frontend

### Backend

### DIFF
--- a/pkg/handlers/ghcapi/queues.go
+++ b/pkg/handlers/ghcapi/queues.go
@@ -42,6 +42,7 @@ func (h GetMovesQueueHandler) Handle(params queues.GetMovesQueueParams) middlewa
 		DodID:                  params.DodID,
 		LastName:               params.LastName,
 		DestinationDutyStation: params.DestinationDutyStation,
+		OriginDutyLocation:     params.OriginDutyLocation,
 		Status:                 params.Status,
 		Page:                   params.Page,
 		PerPage:                params.PerPage,
@@ -150,7 +151,7 @@ type GetServicesCounselingQueueHandler struct {
 	services.OrderFetcher
 }
 
-// Handle returns the paginated list of moves for the TOO user
+// Handle returns the paginated list of moves for the services counselor
 func (h GetServicesCounselingQueueHandler) Handle(params queues.GetServicesCounselingQueueParams) middleware.Responder {
 	appCtx := h.AppContextFromRequest(params.HTTPRequest)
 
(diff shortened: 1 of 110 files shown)


pkg/handlers/ghcapi/queues.go
```diff
@@ -166,6 +166,7 @@ func (h GetServicesCounselingQueueHandler) Handle(params queues.GetServicesCouns
 		DodID:                  params.DodID,
 		LastName:               params.LastName,
 		DestinationDutyStation: params.DestinationDutyStation,
```
eamahanna, line 168: Since the Destination duty station was removed from the services counselor queue and the TOO queue, is this field used anywhere?

swagger-def/ghc.yaml
```diff
@@ -1816,6 +1817,10 @@ paths:
           name: destinationDutyStation
```
eamahanna, line 1817: Can this be removed now?

swagger-def/ghc.yaml
```diff
@@ -1776,6 +1776,7 @@ paths:
               submittedAt,
               originGBLOC,
               destinationDutyStation,
```
eamahanna, line 1778: same comment about if the destintionDutyStation field is needed.

pkg/handlers/ghcapi/queues.go
christopherhuii: I kept this here because I wasn't 100% sure if it's not being used elsewhere.

After thinking about it, I think the best move is to keep it. Even if it wasn't explicitly used, I think keeping the query parameter around makes the API more "complete". For the API portion of the story, I think of it as extending the API so that a user is able to sort/filter by origin duty location. I don't think that necessarily means removing the functionality to be able to sort/filter by destination duty location, even though the user can't do that via UI.

I'm open to discussion!

eamahanna: It's your call. I would probably remove it if it's not being used. I generally just try and take out code if its not being used, but it's your call. It's not a blocker for me.

eamahanna (approved):

pkg/handlers/ghcapi/queues.go
Ryan-Koch: The queue handlers are only used by those components. So if the queue doesn't need it then we can safely remove it.

PR #7844, opened 2021-12-03
MB-10589: Fix deploy storybook in CI, plus client build speedup

ahobson:
## Summary

We are seeing storybook builds fail in CI. It's almost certainly because we tried using a Ramdisk for speed, but storing the storybook results in the Ramdisk results in the task running out of memory.

Go ahead an upgrade to the latest storybook as @duncan-truss discovered it has some performance improvements, too.

While investigating this, I stumbled across a speedup for the client compile.

## Setup to Run Your Code

Watch this PR pass in CI


duncan-truss:
@ahobson Could you merge in master please so we can rule out some of the extra Happo diffs?

ahobson:
> @ahobson Could you merge in master please so we can rule out some of the extra Happo diffs?

Done!

duncan-truss (approved):
I am seeing some weirdness with Storybook and the Happo diffs around the Customer -> Step component.  I think it's because we are wrapping the React component in the stories instead of using decorators https://storybook.js.org/docs/react/writing-stories/decorators#story-decorators.

https://github.com/transcom/mymove/blob/master/src/components/Customer/Home/Step/Step.stories.jsx

It's worth tagging @Shkeating for a review as well.  I think we could create a story to address this issue there and elsewehere in the code.

Screen Shot 2021-12-03 at 5 33 58 PM

gidjin (commented):
LGTM! But was about to flag the same thing Duncan just did.

PR #7847, opened 2021-12-03
MB-10500 Fix find duty stations func to work for zipcodes

eamahanna:
## Summary

This PR updates the query that populates the duty station drop downs. Previously, if you entered a zipcode into the search, an exact match for the zipcode would should up low in the list rather than first. This PR updates the query, so exact postal codes now show up first in the drop down list. 

Here is a slack thread that discusses the possible queries I looked at and the testing of the queries: 
https://ustcdp3.slack.com/archives/CP4979J0G/p1638470436013800

## Setup to Run Your Code

To test this change in the UI complete the following steps. 

##### Terminal 1

Start the server.

```sh
make make server_run
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

1. Create a new customer move
2. Once you get to the `Current duty location` page, enter a postal code. You should see several zip codes come back at the top of the list that are an exact match. To test different scenarios, search for different duty station names and zip codes. The jira ticket contains a list of duty locations that previously did not show up at the top of the list. You can grab any of those zip codes and verify that the duty location now comes up near the top of the list.

Screen Shot 2021-12-03 at 9 36 55 AM

Screen Shot 2021-12-03 at 9 37 46 AM


Ryan-Koch (approved):

pkg/models/duty_station.go
```diff
@@ -123,6 +118,12 @@ from duty_station_names
 where similarity(name, $1) > 0.03
 order by sim desc
 limit 5)
+union
+(select ds.id as duty_station_id, ds.name as name, 1 as sim
+from duty_stations as ds
+inner join addresses a2 on ds.address_id = a2.id
+where a2.postal_code ILIKE $1
```
mboren, line 125: I wonder if it would make sense to filter out installations from the postal code search (`where ds.affiliation is null`). 
Right now, if you search for 88002 the first two results are:
1. White Sands Missile Range
2. White Sands Missile Range, NM 88002

I'm not sure if this is potentially confusing, or a useful feature.

mboren (approved):
I think this looks great. Left one question but I think it's a small thing.
Also, GitHub wouldn't let me leave a comment on these lines, but it would be great if you could change the two places with `LIMIT 11` in the query back to the [smaller values they had previously](https://github.com/transcom/mymove/pull/7742/files#diff-a476e89e22962de5b27cdee6259098bcbe9dd5bd0df13784d7cad5913d8d2225R119). The first limit was 5 originally, and the last one was 7.

pkg/models/duty_station.go
eamahanna: Okay! I can do that. It should only effect when zip codes are searched, so I think that makes sense. Good call out!

I can also change the limits.

PR #7562, opened 2021-10-13
MB-9033 Upgrade react-uswds and uswds together to maintain parity and prevent Happo errors

Shkeating:
## Description

Upgrades react-uswds to newest [2.2.0 version](https://github.com/trussworks/react-uswds/releases/tag/2.2.0), and upgrades uswds version to 2.11.2, which is the uswds version used in this release of react uswds. We can't go to 2.12.0 until react-uswds supports it.

This branch also accounts for a styling change that came with this dependancy update that was causing a lot of visual changes. A default margin-top was getting added abaove all `.usa-form-group`s, so to account for that this branch adds a negative margin below legends to raise those form groups back into place.

## Setup

```sh
yarn install
```

## Code Review Verification Steps
* [ ] There are no aXe warnings for UI.
* [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
* [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
* [ ] User facing changes have been reviewed by design.
* [ ] Request review from a member of a different team.
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9033?atlOrigin=eyJpIjoiNWY4OTE3OTUyMTM4NDI5Y2JhOGJhOGEwMTE4OGFhM2IiLCJwIjoiaiJ9) for this change

## Screenshots
see happo for differences

PR #7850, opened 2021-12-03
MB-9033 Upgrade react-uswds and uswds versions

Shkeating:
## Summary

Upgrades react-uswds to newest [2.4.1 version](https://github.com/trussworks/react-uswds/releases/tag/2.4.1), and upgrades uswds version to 2.11.2, which is the uswds version used in this release of react uswds. We can't go to 2.12.0 until react-uswds supports it.

This branch also accounts for a styling change that came with this dependancy update that was causing a lot of visual changes. A default margin-top was getting added above all `.usa-form-group`s, so to account for that this branch adds a negative margin below legends to raise those form groups back into place. JP has done some work on this but doesn't have it 100% solved. She standardized how the labels are used a bit better which I'm hipoing will make my work easier.

[My previous PR that I closed](https://github.com/transcom/mymove/pull/7562/files) has some additional context.

## Setup to Run Your Code

Start the Storybook locally.

```sh
yarn storybook
```

### Frontend

## Screenshots
this shouldnt cause visual changes. I will be trying to get this through with as few happo changes as possible.


Shkeating:
I think this is at a place where more has been fixed than broken. Something I did here which everyone will need to be in alignment about ahead of merge was removing the wrapping `FormGroup` component off each individual text field and drop down. This has some implications across many components, in that it shrinks the space above some text fields. This trade off was necessary because it fixed many more components that were getting extra space added.

Shkeating:
I lost my access right before I got to push this change that fixes the address fields issues. In `AddressFields.jsx`, if you wrap the contents of the `Fieldset` in a `div` with a class of `margin-top-neg-105` it will fix multiple happo failures.

image

PR #7866, opened 2021-12-07
MB-10618 Add failing test for login loop bug

gidjin:
## Summary

This bug happens when a service member leaves their onboarding incomplete. This results in a 404 on the duty station during the call to `logged_in` endpoint, which is okay but the endpoint was checking the wrong `error` variable for issues in one spot. This fix should resolve it.

[this slack thread](https://ustcdp3.slack.com/archives/CP4U2NKRT/p1638809797074200) explains more about the issue and fix.

## Setup to Run Your Code

Steps to reproduce:
1. Create/login as a customer with local sign in or The public’s one account for government. 
2. Complete the customer onboarding forms up to the orders page
3. When it’s time to choose a new duty location, select a duty location with a postal code in the name such as Columbus, OH 43209
4. Enter the rest of your orders information and click the “Next” button. You should now have made it to the orders file upload page.
5. Logout of Milmove by using the “Sign out” button
6. Attempt to login again with the same user email and you should be shown the login page instead of being logged in and resuming where you left off.

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

### Frontend

### Backend


pkg/handlers/internalapi/users_test.go
```diff
@@ -113,31 +114,67 @@ func (suite *HandlerSuite) TestServiceMemberLoggedInUserNotRequiringAccessCodeHa
...
+		//suite.MustSave(&station)
+
+		req := httptest.NewRequest("GET", "/users/logged_in", nil)
+		req = suite.AuthenticateRequest(req, sm)
+
+		params := userop.ShowLoggedInUserParams{
+			HTTPRequest: req,
+		}
+		builder := officeuser.NewOfficeUserFetcherPop()
+		handler := ShowLoggedInUserHandler{handlers.NewHandlerContext(suite.DB(), suite.Logger()), builder}
```
duncan-truss, line 169: I think the issue is the handler context has no FileStorer set so it panics when it's nil.

This looks like an example of adding the dependency to the handler for tests where it's needed https://github.com/transcom/mymove/blob/3142887e14355ed7cca1c5f3ae5d8c5c47d0b682/pkg/handlers/internalapi/orders_test.go#L57-L60

gidjin: That was it @duncan-truss I now have a properly failing test that gets the expected error `{"error": "FETCH_NOT_FOUND"}`

pkg/handlers/internalapi/users_test.go
```diff
@@ -113,31 +115,70 @@ func (suite *HandlerSuite) TestServiceMemberLoggedInUserNotRequiringAccessCodeHa
 
 func (suite *HandlerSuite) TestServiceMemberNoTransportationOfficeLoggedInUserHandler() {
 	firstName := "Joseph"
```
felipe-lee, line 117: I know this was here before, but do we really need a name for this test? I removed it locally and cut the corresponding bits out of the test setup and it seemed to run fine. Not a big deal tbh, more because the fact that we're setting it explicitly almost makes it seem like a requirement for these tests to run as expected.

felipe-lee (approved):
LGTM, I tested locally pre and post change and got the expected results. Also tested in the exp env, but only tested post-change. It seems to be working as expected!

PR #7830, opened 2021-12-02
MB-10572: Update nix hashes/versions to include support for m1 macs

ahobson:
## Summary

Note that we have to bump some versions because nix added m1 support for new packages after a certain date

## Setup to Run Your Code

```
./nix/update.sh
```

## Verification Steps for Author

These are to be checked by the author.

- [x] Request review from a member of a different team.

## Verification Steps for Reviewers

Run the update! See no errors!

PR #7875, opened 2021-12-07
Update nix psql_exe path so that it's aware of per-user profiles.

Ryan-Koch:
## Summary

I made the path in `scripts/psql-wrapper` match the change that happened [here](https://github.com/transcom/mymove/pull/7830/files#diff-d33e979799a45c7c51752e9c8d96a3e452015d1a40b1e4b6ec6a98e92c4d8430R303)

##### Terminal 1

Run database setup

```sh
make db_dev_reset
```

Does this still work for single user installs of Nix?

### DIFF
--- a/scripts/psql-wrapper
+++ b/scripts/psql-wrapper
@@ -14,7 +14,7 @@ if [ -x "/nix/var/nix/profiles/per-user/${LOGNAME}/mymove/bin/psql" ]; then
 fi
 
 if [ -n "${command[*]}" ]; then
-  exec  $psql_exe --variable "ON_ERROR_STOP=1" postgres://postgres:"${db_password}"@localhost:"${db_port}"/"${DB_NAME}" -c "${command}"
+  exec  "$psql_exe" --variable "ON_ERROR_STOP=1" postgres://postgres:"${db_password}"@localhost:"${db_port}"/"${DB_NAME}" -c "${command}"
 else
-  exec $psql_exe  --variable "ON_ERROR_STOP=1" postgres://postgres:"${db_password}"@localhost:"${db_port}"/"${DB_NAME}"
+  exec "$psql_exe"  --variable "ON_ERROR_STOP=1" postgres://postgres:"${db_password}"@localhost:"${db_port}"/"${DB_NAME}"
 fi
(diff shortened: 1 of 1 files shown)


scripts/psql-wrapper
```diff
@@ -9,12 +9,12 @@ command="${*:-}"
 
 # nix-friendly option
 psql_exe=/usr/local/bin/psql
-if [ -x /nix/var/nix/profiles/mymove/bin/psql ]; then
-  psql_exe=/nix/var/nix/profiles/mymove/bin/psql
+if [ -x "/nix/var/nix/profiles/per-user/${LOGNAME}/mymove/bin/psql" ]; then
```
felipe-lee, line 12: Should this instead check to see if `NIX_PROFILE` has been set (since the `.envrc` file sets it)? That would let us know the person is using nix, and it's only set if the location is writable. From there we could build the `pslq_exe` path using that var, so something like:

```shell
psql_exe="${NIX_PROFILE}/bin/psql"
```

felipe-lee (commented):
This works for me as a single-user install person!

One comment/question though

scripts/psql-wrapper
gidjin, line 12: Interestingly I have both paths locally. I wonder if that is due to me having nix installed very early on.

gidjin (approved):
Script appears to still work, and changes lgtm.

scripts/psql-wrapper
gidjin: This seems cleaner to me if it's possible, missed the comment as part of my review.

felipe-lee: It's because initially `.envrc` was set up for single user installs (so the `profiles/mymove` path, but in #7830 we changed to using the path for multi-user installs (which single-user installs had available, but didn't use for this repo before). So now, when `direnv` runs, it's setting up the new profile path for everyone using `nix`.

Ryan-Koch: Oh yeah that's a good point. I'll make an adjustment.

Ryan-Koch: Change made!

PR #7877, opened 2021-12-08
[MB-9678]  TOO can see non-installation moves for their GBLOC

mboren:
## Summary
In this PR, I'm updating the query used to populate the TOO queue.
We used to determine the GBLOC of moves for this queue by looking at `orders.OriginDutyStation.TransportationOffice.GBLOC`.
Now, we want to use the postal code of the move's first shipment to look up the GBLOC in the `postal_code_to_gblocs` table.
The services counseling queue will still use the transportation office GBLOC, the new logic only applies to the TOO and TIO.

There are a few different things going on in this PR
1. Fixing mistakes I made when setting up the `postal_code_to_gbloc` table
2. Updating the TOO queue query to use new logic for its GBLOC filter
3. Making sure that the Services Counseling queue isn't broken by the TOO update

### postal_code_to_gbloc goofs
These are easy to fix right now because no code prior to this PR uses this table, so
there are no backwards compatibility concerns.
1. I used a singular name: postal_code_to_gbloc. Pop prefers plural
   names, so I renamed it to `postal_code_to_gblocs`. This would have been easy to work around, but consistency is nice.
2. Pop requires tables to have an id column. So I added one. There are >40k rows to update, so the migration is quite large. If there's a better way to go about this please let me know.

### Updating TOO query
The TOO queue needs to determine the GBLOC of a move based on the postal code of the move's first shipment.
It used to look it up on the origin duty station's transportation office.

Getting the first shipment for each move, joining with `postal_code_to_gbloc`, and filtering on the GBLOC is a little complicated, and I couldn't figure out how to express it well with `pop` without resorting to a raw query.
We can't really use a raw query here because there's some application logic that builds the query up, and it's more complex than I would want to do with string concatenation.

I decided to create a view to encapsulate the hairy part of the logic: figuring out which GBLOC to assign to each `move_id`. Joining on this view is easy to do from pop. This view can also be re-used in TIO query.
As far as I know, this is the only way to do it without writing two separate queries, or doing a raw SQL query.
However, we don't use views like this anywhere else, so I'd love to get feedback about this approach and I'm super open to alternatives.
Here's [a slack thread where I asked about it](https://ustcdp3.slack.com/archives/CSGDM3NUW/p1638997875028800).

### updating tests
When testing office functionality, we now have to make sure that postal code of the first shipment on moves that we create has an entry in `postal_code_to_gblocs` that maps it to the same GBLOC as the office user.

## Setup to Run Your Code

<details>
##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Go through onboarding. Choose a non-installation location for your duty location. For this example, let's pick "Twentynine Palms, CA 92277".
2. When setting up your first shipment, choose a postal code that will give a different GBLOC than your origin duty location would, like 90210 (`KKFA` GBLOC).
2. Set up a second shipment with a zip code that belongs to the same GBLOC as your Origin Duty Location. For example 90740 for `LKNQ`.
3. Submit your move
4. Sign in as a TOO in the same GBLOC as your second shipment. Verify that you don't see the move you just set up.
5. Use the admin interface to update your office user's transportation office to one in the same GBLOC as your first shipment's ZIP. In this case, let's use `Los Angeles AFB` to get to the `KKFA` GBLOC.
6. Check the TOO queue again and verify that your move appears now.

### Frontend

### Backend

## Screenshots


mboren:
I just noticed that we have an "Origin GBLOC" column in the TOO and Services Counselor queues. 
This column currently gets the GBLOC of the origin duty station's transportation office. This works for Services Counseling, but not for the new logic in the TOO queue.
Both queues are populated by the same function, so I'm still thinking about how to fix this for the TOO queue.
image

pkg/testdatagen/scenario/e2ebasic.go
```diff
@@ -3070,6 +3070,9 @@ func (e e2eBasicScenario) Run(appCtx appcontext.AppContext, userUploader *upload
 		InnerJoin("transportation_offices", "duty_stations.transportation_office_id = transportation_offices.id").
 		All(&originDutyStationsInGBLOC)
 
+	// Map postal code 90210 (the default we use for just about everything) to the LKNQ GBLOC
+	testdatagen.MakePostalCodeToGBLOC(appCtx.DB(), "90210", "LKNQ")
```
mboren, line 3074: Running `make e2e_test_docker_office` locally shows the same failure that CircleCI is getting. Removing this line allows `make e2e_test_docker_office` to actually run, but of course the tests for TOO stuff that I changed fail.

I have no idea why this is breaking stuff though.

eamahanna: Huh, the only difference I see in how you set up function. I think all our other functions have the input parameters as (appCtx.DB(), assertions Assertions). I wonder if the shared code requires that format. Also, do you have to update anything in the `pkg/testdatagen/shared.go` file? 

I think that would be where I start to try and trouble shoot this. Update the `MakePostalCodeToGBLOC` function to match the rest of the testdatagen function call formats, and add the assertion to the `shared.go` file. 

Do you get any errors in the terminal?

mboren: Thanks, I'll take a look at that stuff. 

I get this over and over again:
```
2021/12/13 19:23:57 Problem with request: Get http://milmovelocal:4000: dial tcp 127.0.0.1:4000: connect: connection refused. Sleeping 1s
```
but other than that no errors.
So it seems to me like either the server or the client is failing to start and i wouldn't be surprised if there's a log file somewhere that i'm not aware of that has the actual error that's preventing them from starting.

mboren: I found the problem. The `postal_code_to_gbloc` table is not getting truncated when this gets run in circleci, so when I try to create another record in it, it violates the unique constraint.

pkg/models/move.go
```diff
@@ -108,6 +108,7 @@ type Move struct {
 	FinancialReviewFlag          bool                    `db:"financial_review_flag"`
 	FinancialReviewFlagSetAt     *time.Time              `db:"financial_review_flag_set_at"`
 	FinancialReviewRemarks       *string                 `db:"financial_review_remarks"`
+	ShipmentGBLOC                MoveToGBLOCs            `has_many:"move_to_gbloc" fk_id:"move_id"`
```
mboren, line 111: I'm still trying to find a way to do this without `has_many`, because it is a 1:1 relationship. There is a pop bug related to EagerPreloading `has_one` relationships, so I can't use that, but I'm hoping there's some other way for this to not be an array.

pkg/models/move_to_gblocs.go
```diff
@@ -0,0 +1,20 @@
...
+	MoveID uuid.UUID `db:"move_id"`
+	Move   Move      `belongs_to:"moves" fk_id:"move_id"`
+	GBLOC  string    `db:"gbloc"`
+}
+
+// MoveToGBLOCs is not required by pop and may be deleted
+type MoveToGBLOCs []MoveToGBLOC
+
+// TableName overrides the table name used by Pop.
+func (m MoveToGBLOC) TableName() string {
```
emahanna2021, line 18: pop requires plural table names, correctt?

migrations/app/migrations_manifest.txt
```diff
@@ -686,4 +686,5 @@
 20211112205920_add_noninstallation_locations.up.sql
 20211123155748_domestic_nts_packing_metadata.up.sql
 20211202213138_index_user_roles_table.up.sql
+20211202234555_move_to_gbloc_view.up.sql
```
eamahanna, line 689: why was this not inserted at the end of the manifest? will that cause any issues?

eamahanna (approved):
These changes look good to me and work in the UI as well. Nice job!

pkg/models/move.go
reggieriser: Have you tried a `belongs_to` by chance?  I know it may not exactly match semantically, but wondering if we might get lucky and it works without the additional burden of the array.  It's possible the view might need an `id` field instead of `move_id` -- just depends on how Pop deals with that internally which I don't know off the top of my head.

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -759,6 +759,19 @@ func QueueMoves(moves []models.Move) *ghcmessages.QueueMoves {
 			deptIndicator = ghcmessages.DeptIndicator(*move.Orders.DepartmentIndicator)
 		}
 
+		var gbloc ghcmessages.GBLOC
+		if move.Status == models.MoveStatusNeedsServiceCounseling {
+			gbloc = ghcmessages.GBLOC(move.Orders.OriginDutyStation.TransportationOffice.Gbloc)
+		} else if len(move.ShipmentGBLOC) > 0 {
+			// There is a Pop bug that prevents us from using a has_one association for
+			// Move.ShipmentGBLOC, so we have to treat move.ShipmentGBLOC as an array, even
+			// though there can never be more than one GBLOC for a move.
```
reggieriser, line 768: If you have a chance, could you file a bug with Pop about this issue?  I probably have a similar one out there for `belongs_to` or `has_many` with `EagerPreload` and pointer fields.

reggieriser (approved):
Reviewed the DB changes and they LGTM.  Nice update query with the temp table!  I looked through some of the non-migration code, but not in great detail.

One nit (would have commented inline but github wouldn't let me on the large migration file) ... you add a constraint named `unique_postal_code`, but you may want to prefix that with the table name since constraint names (I believe) are database-scoped, not table-scoped.

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
mboren: Yeah I'll put that on the todo list, not sure if i'll get to it before the holidays though

migrations/app/migrations_manifest.txt
eamahanna: Ah! Thanks :)

PR #7879, opened 2021-12-08
Psql-wrapper script use the +x pattern that we use in other places to detect NIX_Profile

Ryan-Koch:
## Summary

Makes it so non-nix users can run `make db_dev_reset`


reggieriser (approved):
Works for me (as a non-nix user)!

PR #6355, opened 2021-04-07
[MB-7520] Services counseling move queue skeleton

duncan-truss:
## Description

This PR creates a new temporary route `counseling/queue` to display the service counselor user with a queue of moves in the Needs Service Counseling and Service Counseling Completed statuses.

The work to create a distinctive Service Counselor user role will not be complete until later in the slice so for now any TOO or TIO officer user should be able to access the services counseling queue by going directly to the URL.

By default the services counseling queue will load only those moves in the Needs Service Counseling status, sorted by oldest submitted at date.  If you select Service Counseling Completed from the statuses dropdown any moves in that status will become visible as well.

Things that are not currently implemented in this story:
- **Filtering by exact Requested Move Date or Submitted At Date**
- **Sorting by Requested Move Date**

Requested Move Date displays the earliest requested pickup date of the move's shipments.  Filtering this would be more straightforward but I saw there was a story to further implement this.  However sorting would potentially require converting the order fetcher code to a custom SQL query.  Currently we're pulling out the earliest date in the payload not doing a min() calculation using Pop.

## Reviewer Notes

I decided to extend the existing order fetcher service and only add a new counseling queue handler because it is so similar to the TOO move queue.  I don't think splitting them makes sense at this moment and it helped us move faster to get this work done.  It does make the QueueMoves swagger payload serve a bit of a dual purpose though.

## Setup

Add any steps or code to run in this section to help others prepare to run your code:

```sh
make db_dev_e2e_populate
make server_run
make office_client_run
```

1. Login as a TOO or TIO office user
2. In the URL bar navigate to the page [http://officelocal:3000/counseling/queue](http://officelocal:3000/counseling/queue)
3. A single devseed move should be visible in the queue with the status Needs Service Counseling
4. Clicking on the status filter and selecting Service Counseling Complete will reveal a second move in that status

You can also use the USMC TXO user to test that the service counseling queue would work for Marine Corps moves.

## Code Review Verification Steps

* [ ] If the change is risky, it has been tested in experimental before merging.
* [ ] Code follows the guidelines for [Logging](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#logging)
* [ ] The requirements listed in [Querying the Database Safely](https://github.com/transcom/mymove/wiki/Backend-Programming-Guide#querying-the-database-safely) have been satisfied.
* Any new migrations/schema changes:
[...]

PR #7889, opened 2021-12-09
[MB-10615] Introduce NTS Shipments Scenario for PO8

rogeruiz:
## Summary

Having a database populated with moves that exercise the PO8 work is helpful to everyone working on this Slice of work. This patch series goal is to facilitate this type of work and testing.

## Setup to Run Your Code

<details>
<summary>💻 You will need to use three separate terminal to test this locally.</summary>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

##### Terminal 3

Run the sub-scenario for this

```sh
DEVSEED_SUBSCENARIO=nts_and_ntsr make db_dev_e2e_populate
```

</details>

### Additional steps

Once Terminal 3 has completed, you should have a database with only a couple of NTS related shipments on two Moves. One of the Moves can be verified by the Customer user using local sign-in. This Move has not been submitted yet. The second Move should be viewable by the TOO user and is in a Submitted state.  

Move code in DRAFT state (Customer app): `NTSSUN`

Move code in SUBMITTED state (Office app): `NTSSUB`

Both moves set `usesExternalVendor` to true.

#### Bonus
The ability to create a new user with a specific GBLOC is now available. Just visit the devlocal sign-in page.
image

- [ ] ~Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)~

### Backend


pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -745,11 +745,7 @@ func QueueMoves(moves []models.Move) *ghcmessages.QueueMoves {
 		// we can't easily modify our sql query to find the earliest shipment pickup date so we must do it here
 		for _, shipment := range move.MTOShipments {
 			if queueIncludeShipmentStatus(shipment.Status) {
-				if earliestRequestedPickup == nil {
-					earliestRequestedPickup = shipment.RequestedPickupDate
-				} else if shipment.RequestedPickupDate.Before(*earliestRequestedPickup) {
-					earliestRequestedPickup = shipment.RequestedPickupDate
-				}
+				earliestRequestedPickup = shipment.RequestedPickupDate
```
rogeruiz, line 748: I'm looking for extra scrutiny here. From what I can tell, this code was previously doing the same thing, assigning `earliestRequestedPickup` to the `RequestedPickupDate` from the shipment. 

This code is not prepared for a situation where `earliestRequestedPickup` is blank on the initial run. If it's `nil` it gets set, but I was experiencing issues where it was not `nil` and yet the second if-else conditional would fail. 

I don't want to keep this code like this, but I wanted to call it out since I had no other way of testing the DevSeed data without making this change to this handler code. 

cc: @duncan-truss

rogeruiz (commented):
Could use a couple more eyes on this @transcom/truss-pamplemoose @transcom/truss-roci

pkg/testdatagen/make_nts_shipment.go
```diff
@@ -47,7 +47,7 @@ func MakeNTSShipment(db *pop.Connection, assertions Assertions) models.MTOShipme
 		PickupAddressID:        &pickupAddress.ID,
 		SecondaryPickupAddress: &secondaryPickupAddress,
 		ShipmentType:           models.MTOShipmentTypeHHGIntoNTSDom,
-		Status:                 "DRAFT",
+		Status:                 assertions.MTOShipment.Status,
```
reggieriser, line 50: I don't think we want to set this to the assertion's status here.  It's possible the status has not explicitly been set in the assertion so the shipment could end up with a bad status (hopefully fails with typing or validation).  When it was set to `DRAFT` before, I believe a non-zero value in the assertion's status would overwrite the `DRAFT` in the `mergeModels` call on line 64.  But we'd still have a valid status even with no assertion.

rogeruiz: From the code, I'm seeing what you're referencing. Thanks for catching this. https://github.com/transcom/mymove/blob/master/pkg/testdatagen/shared.go#L170-L176

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
rogeruiz: I see the commit that introduced this change. But I was not clear on why this conditional block was causing issues for me, so I refactored it to this, but like I said, I don't want to keep the code like this.

[=> Original pull request that introduced these changes](https://github.com/transcom/mymove/pull/6355)
[=> Original commit that introduced this logic](https://github.com/transcom/mymove/commit/b05950d0b6d9a617f17688bf512e6aea6167f7a5#diff-c6ffd8ae943f145dfe0ce51c991700bdc11d2bea013fa0c913cd5a5ae4fe2d28R576-R599)

pkg/testdatagen/scenario/shared.go
```diff
@@ -829,55 +836,35 @@ func createSubmittedHHGMoveMultiplePickupAmendedOrders(appCtx appcontext.AppCont
...
-		return [7]string{
-			"583cfbe1-cb34-4381-9e1f-54f68200da1b",
-			"e6e40998-36ff-4d23-93ac-07452edbe806",
-			"f4503551-b636-41ee-b4bb-b05d55d0e856",
-			"06578216-3e9d-4c11-80bf-f7acfd4e7a4f",
-			"1bdbb940-0326-438a-89fb-aa72e46f7c72",
-			"5afaaa39-ca7d-4403-b33a-262586ad64f6",
-			"eecc3b59-7173-4ddd-b826-6f11f15338d9",
-		}
-	}
```
rogeruiz, line 843: I removed these @donaldthai and got things working with auto-generated UUIDs. Because I didn't have the context of why these were here, I thought it was best to just have the UUIDs be generated in the `createMoveWithNTSAndNTSR` function.

Was there a specific reason for these UUIDs to be hard-coded here?

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
duncan-truss: Because the queue only shows one move per row, this is finding the earliest requested pickup date across all shipments of the move so that the queue could be filtered or sorted by the soonest date.  It looks like you got ride of the date comparisons so it would just use the last value which might not be the earliest amongst the shipments.  Are you running into an issue of your seed data not having the requested pickup dates set?

pkg/testdatagen/scenario/shared.go
donaldthai: Not too sure why in this case but usually when we have hardcoded UUIDs it's because we would want to link the associated data together. Doesn't look like the IDs are being used to link the new data from the looks of it. Should be safe to remove!

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
rogeruiz: Interesting. Yeah the issue I was running into was related to this message `Go: panic: runtime error: invalid memory address or nil pointer dereference`. I did not touch much of the code internal to how the `testdatagen` package works, so I was curious what the issue is. But I will say that this is coming up when I only have a single Move in the queue. It should have two shipments in that Move though.

pkg/testdatagen/scenario/shared.go
rogeruiz: Interesting. When you mention _link the new data_, do you mean having UUIDs linked across multiple users/shipments? I could see the `subscenarios.go` code be the place to manage these UUIDs rather than in the `shared.go`. I really like the pattern of passing in a configuration to a subscenario call rather than have the subscenario take care of all the details for me.

donaldthai: Yea, for example: "Create static UUID for a move, use that move static UUID to create linked data like shipments and orders"

rogeruiz: oh no! I did not read all the functions in there. I should refactor this then to use that function rather than the generic function `uuid.Must(uuid.NewV4()).

rogeruiz: Oh wait, never mind, is there a function for creating a static UUIDs and storing those in memory through the duration of the `testdatagen` run?

donaldthai: Like a global static variable? You could just create a public variable in the `devseed.go` or `shared.go` file and reference it in the files you're working on.

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
donaldthai: I think the `RequestedPickupDate` will always be filled out before the move is transitioned over to submitted. We wouldn't run into this error normally.

pkg/testdatagen/scenario/subscenarios.go
```diff
@@ -259,3 +257,33 @@ func subScenarioMisc(appCtx appcontext.AppContext, userUploader *uploader.UserUp
...
+
+		// Create some unsubmitted Moves for Customer users
+		createMoveWithNTSAndNTSR(
+			appCtx,
+			sceneOptionsNTS{
+				shipmentMoveCode: "NTSUNS",
+				moveStatus:       models.MoveStatusDRAFT,
+				orderStatus:      models.OrderStatusDRAFT,
+			},
+		)
```
YanZ777, line 278: I'm having trouble finding these on the customer side. Would it be possible to name the customer email so that it can be found in the local login?

YanZ777 (commented):
The code looks fine to me, just would prefer the customer sample data to have a email so that it's more easily findable.

donaldthai (approved):
Looks good to me! I can find the NTS users on the customer side and can find the move with NTS shipments on the office side. 

@jenniferlynparsons mentioned that we may want to include NTS shipments marked as external but I'll leave that up to you if you want to add that or not. That's just passing in a bool value for `UsesExternalVendor` on the `MTOShipment` model.

pkg/testdatagen/scenario/subscenarios.go
rogeruiz: Ah by default the users use this as their email address. It's within the `createMoveWithNTSAndNTSR` function. I didn't want to modify that function too much. but if it makes sense to do all that outside of the function I can. I was hoping for the minimal amount of code and hopefully a TRA/ADR to refactor `testdatagen` package a bit more further down the road.

https://github.com/transcom/mymove/blob/rogeruiz/mb-10615-nts-shipment-devseed-data/pkg/testdatagen/scenario/shared.go#L851

```go
	email := fmt.Sprintf("nts.%s@nstr.%s", opts.shipmentMoveCode, opts.moveStatus)
```

If you search for `email :=` in that file, `pkg/testdatagen/scenario/shared.go`, you'll see that the emails are being hardcoded in all the other functions. With the instructions in the PR description, I only run the one scenario so there would only be two users for the Customer app. But I do agree that it would be easier to find if it was more visible somewhere.

rogeruiz:
@duncan-truss and @donaldthai I think I figured out what [the issue is with the code path I mentioned I removed](https://github.com/transcom/mymove/pull/7889/commits/8e57020d4bcc6c9a442c55ea141235731365adc5). Because I'm [not using the MoveRouter to transition a Move into a Submitted state](https://ustcdp3.slack.com/archives/C029R9LDTD4/p1639174316244400?thread_ts=1639169594.233700&cid=C029R9LDTD4)🔒, I think that's what's causing issues in the `payload_to_model.go` code. _That's my guess_. I will be transitioning the Moves with the appropriate Services call rather than doing things manually _tomorrow_. But for now I wanted to capture the error message here and my thoughts here.

If you want to check this out locally, take the latest `HEAD` of this branch and then run the `make db_dev_e2e_populate` command. Then search for the `NTSSUB` move in the TXO view and you'll see the following error in the `make server_run` terminal window.

```sh
[POP] 2021/12/15 03:40:45 sql - SELECT transportation_offices.address_id, transportation_offices.created_at, transportation_offices.gbloc, transportation_offices.hours, transportation_offices.id, transportation_offices.latitude, transportation_offices.longitude, transportation_offices.name, transportation_offices.note, transportation_offices.services, transportation_offices.shipping_office_id, transportation_offices.updated_at FROM transportation_offices AS transportation_offices WHERE id = $1 | [38376fd1-22cb-4495-a808-e11193995780]
2021-12-15T03:40:45.121Z        ERROR   middleware/recovery.go:39       http request panic      {"git_branch": "rogeruiz/mb-10615-nts-shipment-devseed-data", "git_commit": "8e57020d4bcc6c9a442c55ea141235731365adc5", "milmove_trace_id": "d7d38b33-79b9-4840-b751-b907763f00a5", "url": "/ghc/v1/queues/moves?page=1&perPage=20&sort=status&order=asc&locator=NTSSUB", "milmove_trace_id": "d7d38b33-79b9-4840-b751-b907763f00a5", "error": "runtime error: invalid memory address or nil pointer dereference"}
github.com/transcom/mymove/pkg/middleware.Recovery.func1.1.1
        /Users/rsr/Developer/trussworks/mymove/pkg/middleware/recovery.go:39
github.com/transcom/mymove/pkg/handlers/ghcapi/internal/payloads.QueueMoves
        /Users/rsr/Developer/trussworks/mymove/pkg/handlers/ghcapi/internal/payloads/model_to_payload.go:750
github.com/transcom/mymove/pkg/handlers/ghcapi.GetMovesQueueHandler.Handle
        /Users/rsr/Developer/trussworks/mymove/pkg/handlers/ghcapi/queues.go:74

```

pkg/testdatagen/scenario/subscenarios.go
YanZ777: That's fair. I think a TRA would be a good idea to refactor the devseed data.

YanZ777 (approved):
I like the changes! 
It would be nice if we could keep the options DRY, but I'm not sure if that is completely possible here. In any case, I won't block the PR over it.

PR #7893, opened 2021-12-10
[MB-10590] Add ButtonDropdown component

hannaliebl:
## Summary

Style the React USWDS so that the select looks like a button and includes the required sort icon and include it in Storybook.

## Open Questions/Comments
1. Should this live here, or in the Office subfolder?
2. I think an aria-label will be needed in spots where we include this without a form label, so I added it as a prop, but let me know if that isn't the case.
3. I eyeballed the REM measures, if there's a better way, let me know. I included the usa-button styles and overrode what I needed.
4. Like I mentioned on Slack, I can't style the actual option elements themselves, so this ended up being just styling the select element itself to make it look like a button.

## Setup to Run Your Code

Start Storybook locally.

```sh
make storybook
```

</details>

### Additional steps

1. Look at Components/ButtonDropdown

### Frontend

- [x] There are no aXe warnings for UI.
- [x] There are no new console errors in the browser devtools
- [x] There are no new console errors in the test output

## Screenshots
Screen Shot 2021-12-10 at 3 54 10 PM

Screen Shot 2021-12-10 at 3 54 18 PM


hannaliebl:
I was running into Storybook issues (custom styles not being applied), so I actually confirmed this by plopping the component into the move details page, but that might suggest I did something wrong with Storybook 😬

src/components/ButtonDropdown/ButtonDropdown.module.scss
```diff
@@ -0,0 +1,27 @@
+@import '../../shared/styles/colors';
```
YanZ777, line 1: This import can just be put in as `@import 'shared/styles/colors.scss';`

src/components/ButtonDropdown/ButtonDropdown.module.scss
```diff
@@ -0,0 +1,27 @@
...
+  text-align: left;
+  padding-right: 2.9rem;
+  &:hover {
+    background-color: $primary-dark;
+  }
+}
+
+.ButtonDropdownWrapper {
+  position: relative;
+}
```
YanZ777, line 18: I think what you actually want to do here is have this wrap the other classes in this file, and your custom styling should show up. 
Doing that however, I noticed that the icon appears off center.
e.g.
```
.ButtonDropdownWrapper {
...
    .ButtonDropdown{ ... }
    .ButtonDropdownIcon{ ... }
}
```

src/components/ButtonDropdown/ButtonDropdown.test.jsx
```diff
@@ -0,0 +1,36 @@
...
+    const { queryByTestId } = render(
+      <ButtonDropdown onChange={() => {}}>
+        <option>- Select - </option>
+        <option value="value1">Option A</option>
+        <option value="value2">Option B</option>
+        <option value="value3">Option C</option>
+      </ButtonDropdown>,
+    );
+
+    expect(queryByTestId('dropdown')).toBeInTheDocument();
```
YanZ777, line 18: Minor, but would switch these over to just do `screen.queryByTestId...` so you don't have to destructure the queries.

amanda5iu (commented):
Looking at this in Storybook, the text color is off. The button label is black and the options are white. I also tried pulling up the move details page (both SC and TOO view) but it showed the old button. Maybe I'm doing something wrong?

on load:
Screen Shot 2021-12-13 at 10 04 47 AM

hover:
Screen Shot 2021-12-13 at 10 12 59 AM
749/145847339-74e5075d-c66f-4a3b-a415-bbcc20de2ce1.png)

on click:
Screen Shot 2021-12-13 at 10 13 07 AM

src/components/ButtonDropdown/ButtonDropdown.test.jsx
hannaliebl: Ah, right, I always forget this and do the opposite, thanks, will update shortly.

hannaliebl:
@amanda5iu I fixed the firefox styles and also used the existing USWDS icon in my latest push.

amanda5iu (approved):
@hannaliebl this looks great!! Thanks for making all the styling changes 🙌

src/components/ButtonDropdown/ButtonDropdown.module.scss
```diff
@@ -0,0 +1,38 @@
+@import 'shared/styles/colors.scss';
+@import 'shared/styles/basics';
+
+.ButtonDropdown {
+  position: relative;
+  display: inline-block;
```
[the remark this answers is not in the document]
YanZ777: I think usually the container sets the width, but I don't see a problem leaving it here either.

YanZ777 (approved):
Changes look good to me!

PR #7849, opened 2021-12-03
[MB-10151][MB-10152] Domestic NTS Packing pricer

reggieriser:
## Summary

This PR creates a pricer for the Domestic NTS Packing service (`DNPK`).

[This discovery document](https://docs.google.com/document/d/1tfQAqmpdaFcV8RwQOav99ANjr3lpzKznOCUY1swr8hE/edit#heading=h.jqin6bx5hd61) explains more about the approach used.

In testing using the office and prime simulator UIs, I also found the following issues and fixed them:
- when I previously added the metadata for the pricing params in #7785, I accidentally made some of the weight fields required when they should be optional
- there's a Pop bug with `EagerPreload` that I put in a workaround for in the `ListMTOShipments` service

## Setup to Run Your Code

The tests exercise the pricer and its components, but you should be able to also test via the UI.  I had success doing the following:
- in the customer UI, create a move with an NTS shipment
- in the office UI with the service counselor role, submit the move (if needed)
- in the office UI with the TOO role, edit the orders to add the missing fields and then approve the shipment
- in the office UI with the prime simulator role, add the weights/addresses, then submit a payment request that includes domestic NTS packing
- inspect the result and you should see the JSON with the pricing information for domestic NTS packing
- in the office UI with the TIO role, you should see the domestic NTS packing in the payment request screen (note that we don't have the "show your work" calculations in place for it yet) 

## Verification Steps for Author

These are to be checked by the author.

- [ ] Request review from a member of a different team.
- [ ] Have the Jira acceptance criteria been met for this change?

## Verification Steps for Reviewers

These are to be checked by a reviewer.

### Backend

- [ ] Code follows the guidelines for [Logging](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide#logging)
- [ ] The requirements listed in [Querying the Database Safely](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide/#querying-the-database-safely) have been satisfied.

PR #7894, opened 2021-12-10
[MB-10152] Added DNPK to some of the same places where DPK is checked

reggieriser:
## Summary

I noticed in slice demo that we failed with a "destination address id" not found when issuing a payment request.  I started poking through the code looking for why this might happen.  I did notice a few places where we probably needed to have similar behavior for `DNPK` as `DPK` that had not been changed yet.

That said, I don't think this caused the demo failure as many service items would fail to be priced without a destination address (and we were pricing a whole bunch in the demo).  So it may just be the case that the dates/weights/dest address didn't save for some reason in the prime simulator (when I test it, dest address is required by the prime simulator, so we shouldn't have even been able to proceed without it).  At any rate, this helped me find a few places in the code that needed adjusting.

## Setup to Run Your Code

Use a similar setup as shown in #7849, but locally you may want to null out the destination address for your shipment in your local DB before creating the payment request to simulate it missing (since I don't see a way to keep it null through the prime simulator).

I also added some tests that should exercise my changes.

### Backend


jacquelineIO (approved):
Thanks for jumping on this Reggie. I can't do the run at the moment. But leaving approval from reviewing the code.

jacquelineIO:
@reggieriser I Ran this but did need to keep the zip as part of the destination address. I think this is a bug of pricing and not this PR. With the zip in place, I was able to successfully price.

PR #7907, opened 2021-12-13
MB-10713: Remove SetTraceID from HandlerContext and cascading changes

ahobson:
## Summary

The HandlerContext is initialized once and shared across all handlers.
It should not contain data that changes per request. (It should really be
HandlerConfig, but that change will have to wait for another day).

Remove SetTraceID and replace GetTraceID with GetTraceIDFromRequest to
be explicit about where the TraceID is coming from.

Change the tracing middleware to save/restore a UUID instead of a
string and only convert to a string when logging. This is needed for
storing the UUID in the Event.

Also change Event to take an AppContext and TraceID instead of a
Request and HandlerContext so that the service is not tied to a
handler specific item.

## Setup to Run Your Code

`make server_test`

### Backend


pkg/handlers/contexts.go
```diff
@@ -81,7 +81,6 @@ type handlerContext struct {
 	useSecureCookie       bool
 	appNames              auth.ApplicationServername
 	featureFlags          map[string]bool
-	traceID               uuid.UUID
 	sessionManagers       [3]*scs.SessionManager
 }
```
rogeruiz, line 85: Removing the `traceID` from the `handleContext` and replacing the `SetTraceID` with the `GetTraceIDFromRequest`.

pkg/services/event/notification.go
```diff
@@ -32,6 +32,7 @@ func isSourcePrime(event *Event) bool {
 // notificationSave saves the record in the webhook_notification table
 // If it fails, it will return an error.
 func notificationSave(event *Event, payload *[]byte) error {
+	appCtx := event.AppContext
```
rogeruiz, line 35: Nice!

pkg/trace/context.go
```diff
@@ -10,16 +12,16 @@ type contextKey string
 var traceIDContextKey = contextKey("milmove_trace_id")
 
 // NewContext adds the traceID string into the context and returns the new context
-func NewContext(ctx context.Context, traceID string) context.Context {
+func NewContext(ctx context.Context, traceID uuid.UUID) context.Context {
```
rogeruiz, line 15: Trace IDs are now `uuid.UUID` rather than `strings`. Straight forward change.

pkg/trace/context.go
```diff
@@ -10,16 +12,16 @@ type contextKey string
...
 	return context.WithValue(ctx, traceIDContextKey, traceID)
 }
 
 // FromContext returns a traceID that was previously added into the context, if any.
-func FromContext(ctx context.Context) string {
+func FromContext(ctx context.Context) uuid.UUID {
```
rogeruiz, line 20: Explains the function signature for the `NewContext` method above.

pkg/trace/context.go
rogeruiz: We're also setting new UUIDs whenever we set up new Contexts. I think this looks good, but I wanted to verify that that is my understanding.

rogeruiz (approved):
lgtm 🌈 thanks for fixing Event Tracing throughout the app @ahobson 

A majority of the files changed in this PR are changes outlined below: 
- `GetTraceIDFromRequest(params.HTTPRequest)` is the new method instead of `GetTraceID`. It calls `FromContext` ti get the value of the context from the `HTTPRequest`.

PR #7918, opened 2021-12-14
Create audit_log_tech_adr.md

kennethchow-truss:
## MB-10256 ADR for History and Audit Log

New ADR for history and audit log feature overall tech considerations. Data model design will be in confluence

### DIFF
--- a/docs/adr/audit_log_tech_adr.md
+++ b/docs/adr/audit_log_tech_adr.md
@@ -1,0 +1,46 @@
+# History & Audit Logging
+
+MilMove requires an audit log of actions taken on the application. In this application, several different users have different use cases for the audit logging feature.
+
+- Customer user - veteran user who is requesting the move
+- Office users - approves moves, helps veterans with their move questions
+- Prime users - does the actual moving process for the customer
+- Data warehouse users - consumes log data for analytics
+
+As a part of the user requirements gathering, the audit log data must include the following data
+
+- When the action took place
+- What action took place
+- Who took the action
+
+When considering how to implement the application, there are two main choices worth considering in the Milmove application. NoSQL vs. SQL Route
+
+## Considered Alternatives
+
+### SQL Route
+
+This approach leverages the existing technology to implement the audit log. This is the path of least resitance because engineers do not need to introduce new technology to the application. Introducing new technology will need new considerations such as CI/CD, security, knowledge, Transcom approval, etc. 
+
+#### Feasability 
+
+The more complex we make the underlying tables for the audit logging, the write speeds will be slower. Therefore, if this route is chosen, a flatter database design will help mitigate performance issues especially as the application scales.  
+
+### NoSQL Route
+
+This approach will introduce new technology. If the amount of data that is expected to be stored into the database is extremely large, a NoSQL route will help maintain write speeds. NoSQL by default has a flat data design so there is no need for extra data modeling considerations because of the flexible nature of NoSQL
+
+#### Feasability
+
+Introducing technology will have a lot of hurdles to jump through. Given the state of the Milmove application, a quicker time to launch may be preferred. If this route was chosen, developers will be introducing new AWS services to the application such as DynamoDB, Lambda, API Gateway, etc. 
+
+NoSQL may slow down consumers in retrieving data to our application. Doing any type of bulk work whether getting or updating data proves more dificult with a NoSQL table. Querying the application may be cumbersome because there is unstructured data. However, AWS includes services such as AWS Glue to help normalize unstructured data to more readabale formats. 
+
+## Comparing the Alternatives
+
+### Performance
+
+NoSQL has quicker write speeds so that the application does not slow down due to a new history log call in the API. NoSQL also has workarounds for its worse/slower readability for potential consumers
+
+### Ease of use
+
+SQL route will leverage existing technology to build out the application. Therefore, there is no additional approval needed or infrastructure resources need to build the feature. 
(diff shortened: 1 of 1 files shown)


docs/adr/audit_log_tech_adr.md
```diff
@@ -0,0 +1,46 @@
+# History & Audit Logging
+
+MilMove requires an audit log of actions taken on the application. In this application, several different users have different use cases for the audit logging feature.
+
+- Customer user - veteran user who is requesting the move
+- Office users - approves moves, helps veterans with their move questions
+- Prime users - does the actual moving process for the customer
+- Data warehouse users - consumes log data for analytics
```
rogeruiz, line 8: I'm curious if for this outcome we only need to focus on the **Office users**. I think it might be a bit out of scope to mention the other three without drawing emphasis on the **Office users**. I had this realization recently when thinking about the IGC and data modeling.

docs/adr/audit_log_tech_adr.md
```diff
@@ -0,0 +1,46 @@
...
+
+As a part of the user requirements gathering, the audit log data must include the following data
+
+- When the action took place
+- What action took place
+- Who took the action
+
+When considering how to implement the application, there are two main choices worth considering in the Milmove application. NoSQL vs. SQL Route
```
rogeruiz, line 16: On the Truss side we try to avoid creating a binary fallacy by having two choices. Usually there are three choices.

1. Do nothing
2. Do a thing
3. Do some other thing

I think this could use the section as well from the template.

docs/adr/audit_log_tech_adr.md
```diff
@@ -0,0 +1,46 @@
+# History & Audit Logging
```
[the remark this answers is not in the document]
rogeruiz: I gave the whole thing a quick draft review and think you're going in a good direction here. I suggest sticking to the sections from the `mymove/docs/adr/template.md` file for the sake of readability. It will help guide the decisions you present in a more consistent way with the other ADRs in here.

docs/adr/audit_log_tech_adr.md
```diff
@@ -0,0 +1,46 @@
...
+
+### NoSQL Route
+
+This approach will introduce new technology. If the amount of data that is expected to be stored into the database is extremely large, a NoSQL route will help maintain write speeds. NoSQL by default has a flat data design so there is no need for extra data modeling considerations because of the flexible nature of NoSQL
+
+#### Feasability
+
+Introducing technology will have a lot of hurdles to jump through. Given the state of the Milmove application, a quicker time to launch may be preferred. If this route was chosen, developers will be introducing new AWS services to the application such as DynamoDB, Lambda, API Gateway, etc. 
+
+NoSQL may slow down consumers in retrieving data to our application. Doing any type of bulk work whether getting or updating data proves more dificult with a NoSQL table. Querying the application may be cumbersome because there is unstructured data. However, AWS includes services such as AWS Glue to help normalize unstructured data to more readabale formats.
```
rogeruiz, line 36: Because of compliance, we'll be limited to services within the ATO boundary for Staging and Production environments. This means that [we'll need to leverage only the services available in GovCloud](https://aws.amazon.com/es/govcloud-us/details/#Databases).

I am not going to suggest some tech ideas I have around this just now, but I wanted to bring up the limitations we have when working against us for this decision.

rogeruiz (commented):
great start @kennethchow-truss 🙌🏽

YanZ777 (commented):
I think you might also need to update the `index.md` in this file as well.

Is the list of users meant to be exhaustive or an example?

Is there any need to discuss the read speeds for either route? It looks like write is a concern, but I'm not sure if read speed is concern as well.

docs/adr/audit_log_tech_adr.md
```diff
@@ -0,0 +1,46 @@
+# History & Audit Logging
+
+MilMove requires an audit log of actions taken on the application. In this application, several different users have different use cases for the audit logging feature.
```
YanZ777, line 3: Is the list of users meant to be exhaustive or an example?

docs/adr/audit_log_tech_adr.md
YanZ777, line 16: You might want to rephrase this as "when considering how to store the data for this feature, there are .... NoSQL, SQL, ...etc."

docs/adr/audit_log_tech_adr.md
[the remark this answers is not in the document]
kennethchow-truss: Oh thanks... I took this format from another random ADR. I will change it

docs/adr/audit_log_tech_adr.md
kennethchow-truss: In this https://docs.google.com/spreadsheets/d/1Kg4VSGFjrsq_30js0RLfslOxk8sDXOH1XvkFl0W9dmE/edit?usp=sharing, some of the tickets have roles other than office users. Shouldn't we include them in this?

docs/adr/audit_log_tech_adr.md
kennethchow-truss: Yep, I am working on a third option

docs/adr/audit_log_tech_adr.md
kennethchow-truss: All of these suggested services are on gov cloud.

docs/adr/audit_log_tech_adr.md
kennethchow-truss: It was meant to reflect the user roles in this https://docs.google.com/spreadsheets/d/1Kg4VSGFjrsq_30js0RLfslOxk8sDXOH1XvkFl0W9dmE/edit?usp=sharing

PR #7933, opened 2021-12-16
Em mb 9676 update services counselor queue

eamahanna:
## Summary

This PR updates the services counselor queue to show moves that are within the office users GLBOC based on the moves origin duty location. This also updates the queue to have no differences for an office user if they are a marine. All marine moves will be displayed based only on their origin duty station GBLOC. 

## How to Verify Changes

To verify tests run `make server_test` and `make client_test`.

To verify locally through the UI complete the following steps:
- In two separate terminals run `make server_run` and `make client_run`
- In the customer UI create a two moves: one as not a marine service member, and one as a marine service member.  Ensure each origin duty location has the same GBLOC as the office user you intend to view the queue with. I believe the GLBLOC is "KKFA"  if you use the pre-created service counselor locally.
- Once you have created the moves, log into the office app and sign in as a services counselor. 

You should see both moves in the queue!

## Screenshots
Screen Shot 2021-12-21 at 2 04 14 PM


pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -761,8 +761,9 @@ func QueueMoves(moves []models.Move) *ghcmessages.QueueMoves {
 		}
 
 		var gbloc ghcmessages.GBLOC
-		if move.Status == models.MoveStatusNeedsServiceCounseling {
-			gbloc = ghcmessages.GBLOC(move.Orders.OriginDutyStation.TransportationOffice.Gbloc)
+		if move.Status == models.MoveStatusNeedsServiceCounseling || move.Status == models.MoveStatusServiceCounselingCompleted {
+			// THIS NEEDS TO BE UPDATED!!!!
+			gbloc = ghcmessages.GBLOC(*gblocFilter)
```
eamahanna, line 766: I'm not sure if this is the best move. From the service, I am passing the GBLOC used to filter the queue up. The glboc filter value is then used to populate the move gbloc....this is a hack, but my only other idea on how to fix it is to add a view similar to Marty's for the shipment gbloc. I think we need to figure out a better way to do this....Do you all have any thoughts?

leia-sefkin: If it's not too much work using a view makes more sense to me, even if it's a bit more code. It will be more clear what's going on as well

eamahanna: That's fine with me we will just have another field on our model, which could get confusing (especially if the TIO has a different GBLOC assignment and has to add to the model), but I can go with that.

leia-sefkin: TRUE, maybe we can try to name it clearly to help somewhat?

pkg/services/order/order_fetcher.go
```diff
@@ -43,12 +43,25 @@ func (f orderFetcher) ListOrders(appCtx appcontext.AppContext, officeUserID uuid
...
+		for _, status := range params.Status {
+			if status == string(models.MoveStatusNeedsServiceCounseling) || status == string(models.MoveStatusServiceCounselingCompleted) {
+				needsCounseling = true
+			}
+		}
+	}
+
+	branchQuery := branchFilter(params.Branch, needsCounseling)
+
 	// If the user is associated with the USMC GBLOC we want to show them ALL the USMC moves, so let's override here.
```
Ryan-Koch, line 59: Would you mind updating the comment here to mention the services counseling-TOO split that's happening in the logic below?

src/constants/queues.js
```diff
@@ -43,6 +43,15 @@ export const BRANCH_OPTIONS = [
   { value: 'COAST_GUARD', label: 'Coast Guard' },
 ];
 
+export const SERVICE_COUNSELING_BRANCH_OPTIONS = [
```
Ryan-Koch, line 46: Thanks for setting up a constant for this!

src/pages/Office/ServicesCounselingQueue/ServicesCounselingQueue.jsx
```diff
@@ -106,15 +103,13 @@ const columns = (isMarineCorpsUser = false) => [
 const ServicesCounselingQueue = () => {
   const {
     // eslint-disable-next-line camelcase
-    data: { office_user },
+    // data: { office_user },
```
Ryan-Koch, line 106: Do we need the commented out bits above?

Ryan-Koch (approved):

Ryan-Koch (approved):

src/pages/Office/ServicesCounselingQueue/ServicesCounselingQueue.jsx
eamahanna: Nope!

PR #7958, opened 2021-12-23
Add migration to regrant permissions to crud user

gidjin:
## Summary

It was discovered recently that creating a new view in the demo environment is not being created with permissions for the MilMove database user `crud` to access it. This results in a `error pq: permission denied for view duty_stations` error. 

- [this slack thread](tbd)
- [this other thread]()
- see these migrations
  - migrations/app/schema/20210113173236_create_crud_role.up.sql
- Possibly related migrations
  - migrations/app/schema/20201230070814_fix_ecs_user_in_docker.up.sql
  - migrations/app/schema/20190129115001_master_role.up.sql
  - migrations/app/secure/20190129115416_ecs_user_for_rds.up.sql
  - migrations/app/secure/20190726151515_grant_all_ecs_user.up.sql
 
This needs to be tested in experimental or demo to find out if the migration as it is would be best or if there is something else going on that is preventing permissions from being on the view. Alternate solution that worked but would require a `GRANT` statement for every future view.

```sql
GRANT SELECT, INSERT, UPDATE, DELETE ON duty_stations TO crud;
GRANT SELECT, INSERT, UPDATE, DELETE ON move_to_gbloc TO crud;
```

## Setup to Run Your Code

I can't replicate this locally. I checked and my local db has a `crud` user but that user has the correct permissions on the view. If you can replicate it locally reseting the database in exp or demo won't be needed to experiment with it. I do have this deployed to Demo for now.

### Additional steps

1. Log into the office in demo and ensure no 500 error. 
2. Use the api to retrieve move with code `7799QT` from demo

- [X] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
- [X] Request review from a member of a different team.

### Frontend

### Backend

### DIFF
--- a/migrations/app/migrations_manifest.txt
+++ b/migrations/app/migrations_manifest.txt
@@ -691,3 +691,4 @@
 20211213235039_update_duty_station_location.up.sql
 20211214234555_move_to_gbloc_view.up.sql
 20211221182830_origin_duty_location_to_gbloc_view.up.sql
+20211222193432_regrant_access_to_views_for_crud.up.sql
--- a/migrations/app/schema/20211222193432_regrant_access_to_views_for_crud.up.sql
+++ b/migrations/app/schema/20211222193432_regrant_access_to_views_for_crud.up.sql
@@ -1,0 +1,16 @@
+-- For some reason the following migration steps haven't taken fully in all
+-- of our environments, specifically demo.
+--
+-- Running them again should not hurt anything.
+-- See also migrations/app/schema/20210113173236_create_crud_role.up.sql
+--
+-- Modify existing tables, views, and sequences.
+GRANT SELECT, INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA public TO crud;
+GRANT USAGE, UPDATE ON ALL SEQUENCES IN SCHEMA public TO crud;
+
+-- Modify future tables, views, and sequences.
+-- Do not include `FOR ROLE` in the following statements so that:
+-- 1. when this runs in RDS, it will apply to the role running migrations.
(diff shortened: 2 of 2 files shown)


ahobson (approved):
I reviewed the [pg docs](https://www.postgresql.org/docs/12/sql-grant.html) and I agree that this should do the trick. I wonder if when we created demo we restored from a db backup and that may not have included these role changes.

We really shouldn't be making role changes inside a migration for that reason, but that's a problem for another day.

To explain a bit more: [pg_dump](https://www.postgresql.org/docs/12/app-pgdump.html) only backs up a single database. Role level permissions, however, are at the cluster level. As that page says

> pg_dump only dumps a single database. To back up an entire cluster, or to back up global objects that are common to all databases in a cluster (such as roles and tablespaces), use pg_dumpall.

 Our migrations conflate changes to a single database with changes that affect the entire cluster. This is one reason why squashing our migrations is so difficult: we really need to separate out these role changes into a separate database configuration process.

.circleci/config.yml
```diff
@@ -26,7 +26,7 @@ references:
   #
   # In addition, it's common practice to disable acceptance tests and
   # ignore tests for demo deploys. See the branch settings below.
-  demo-branch: &demo-branch placeholder_branch_name
+  demo-branch: &demo-branch gidjin_fix_duty_station_view_permissions
```
ahobson, line 29: Do we want these branch changes to go into master? I don't think we do

gidjin: No this was to get the fix onto Demo for Dan to check out. I'll remove them before merging.

ahobson:
Thank you! Please merge when ready!

PR #7970, opened 2021-12-28
MB-10864: Add support for running tests with gotestsum

ahobson:
## Summary

The current go-junit-reporter does not include the log/failure
message in the XML, which makes it harder to see the error in
CircleCI.

gotestsum JUnit XML support does include the error message, but it
runs the tests different. This adds support for gotestsum while trying
to keep support for go-junit-reporter so we can switch back quickly if
we need to.

1. Run `make server_test` locally
2. See the tests run in CI successfully


gidjin (approved):
LGTM! looks easy to switch back to the other library if we need to.

PR #7971, opened 2021-12-28
MB-10561: create new ranking for: E-9 Special Senior Enlisted 

pearl-truss:
## Summary

This PR adds a E-9 special senior enlisted ranking that has a 14000 weight allowance without dependents and 17000 weight allowance with them

NOTE: The customer header will display the weight allowance without dependents regardless if the service member has dependents. This will be addressed in https://dp3.atlassian.net/browse/MB-9872

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the client app
2. Login as a new customer
3. Go through the customer onboarding flow. You'll notice on the DOD info page that the rank dropdown has

-  a new field: E-9 (Special Senior Enlisted)
- an updated 0-1 with 'or' replacing the '\'

### Frontend

## Screenshots
Screen Shot 2021-12-28 at 3 23 30 PM


src/services/internalApi.js
```diff
@@ -20,7 +20,7 @@ export async function getInternalClient() {
 // Attempt at catch-all error handling
 // TODO improve this function when we have better standardized errors
 export function getResponseError(response, defaultErrorMessage) {
-  return response.body?.detail || response.statusText || defaultErrorMessage;
+  return response?.body?.detail || response?.statusText || defaultErrorMessage;
```
pearl-truss, line 23: While working on this ticket, I discovered an existing bug that can be reproduced by clicking 'Edit' on the customer details page for Orders. If you try to save changes to the order, currently in the master branch nothing happens when you click save since the error message was being suppressed from being displayed. 

With this code change, you now at least see the red error badge when you click the save button. I can create a bug ticket to investigate and fix the root cause of the error

pearl-truss:
Changes are implemented. And I will add test for these changes

pkg/paperwork/forms.go
```diff
@@ -16,34 +16,35 @@ import (
...
-	internalmessages.ServiceMemberRankW1:                "W-1",
-	internalmessages.ServiceMemberRankW2:                "W-2",
-	internalmessages.ServiceMemberRankW3:                "W-3",
-	internalmessages.ServiceMemberRankW4:                "W-4",
-	internalmessages.ServiceMemberRankW5:                "W-5",
-	internalmessages.ServiceMemberRankAVIATIONCADET:     "Aviation Cadet",
-	internalmessages.ServiceMemberRankCIVILIANEMPLOYEE:  "Civilian Employee",
-	internalmessages.ServiceMemberRankACADEMYCADET:      "Service Academy Cadet",
-	internalmessages.ServiceMemberRankMIDSHIPMAN:        "Midshipman",
+	internalmessages.ServiceMemberRankE1:                      "E-1",
```
duncan-truss, line 19: I know this is existing technical debt, we likely don't want to rely on the swagger generated `internalmessages` values here. Could be nice to migrate this to use the pkg/models version if possible.

duncan-truss:
In the customer onboarding on the review page prior to the signature I also saw we are displaying the raw enum value instead of formatting into the readable label.
Screen Shot 2021-12-28 at 4 10 48 PM

pkg/paperwork/forms.go
```diff
@@ -15,36 +15,36 @@ import (
...
+	models.ServiceMemberRankE1:                      "E-1",
+	models.ServiceMemberRankE2:                      "E-2",
+	models.ServiceMemberRankE3:                      "E-3",
+	models.ServiceMemberRankE4:                      "E-4",
+	models.ServiceMemberRankE5:                      "E-5",
+	models.ServiceMemberRankE6:                      "E-6",
+	models.ServiceMemberRankE7:                      "E-7",
+	models.ServiceMemberRankE8:                      "E-8",
+	models.ServiceMemberRankE9:                      "E-9",
+	models.ServiceMemberRankE9SPECIALSENIORENLISTED: "E-9 (Special Senior Enslisted",
```
ronaktruss, line 28: I think there might be a small typo in `Enslisted`

src/services/internalApi.js
pearl-truss: I found that this bug was already logged and ticket created: https://dp3.atlassian.net/browse/MB-9875?jql=text%20~%20%22edit%20orders%22

pkg/paperwork/forms.go
```diff
@@ -15,35 +15,36 @@ import (
...
+	models.ServiceMemberRankE1:                      "E-1",
+	models.ServiceMemberRankE2:                      "E-2",
+	models.ServiceMemberRankE3:                      "E-3",
+	models.ServiceMemberRankE4:                      "E-4",
+	models.ServiceMemberRankE5:                      "E-5",
+	models.ServiceMemberRankE6:                      "E-6",
+	models.ServiceMemberRankE7:                      "E-7",
+	models.ServiceMemberRankE8:                      "E-8",
+	models.ServiceMemberRankE9:                      "E-9",
+	models.ServiceMemberRankE9SPECIALSENIORENLISTED: "E-9 (Special Senior Enlisted",
```
duncan-truss, line 28: Missing the ending parenthesis here

src/components/Customer/Review/Summary/index.jsx
```diff
@@ -235,6 +235,8 @@ export class Summary extends Component {
     const thirdSectionHasContent =
       showMoveSetup || showPPMShipmentSummary || (isReviewPage && (mtoShipments.length > 0 || currentPPM));
 
+    const formattedRank = get(schemaRank['x-display-value'], get(serviceMember, 'rank'));
```
duncan-truss, line 238: The lodash get method is used a lot in this file, we're trying to reduce the usage (https://github.com/transcom/mymove/blob/master/docs/adr/0057-lodash.md) because of newer Javascript features and it's bloated module size.  I bet we'll encounter this some in PPM work going forward.

Usually you can just use the [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) in replacing usages of get along with `||` if you need a default value.

duncan-truss: It's also weird that we are relying on the internal schema enum values in this component but that's a battle for another time I think to make those app provided constants.

src/services/internalApi.js
felipe-lee: Yeah this bug has been around for a while 😞 but I'm glad at least the error won't be suppressed anymore!

felipe-lee (commented):
Thanks for making these changes! I know it's unrelated but I'm really excited that the orders page won't silently fail anymore haha.

It seems like the `swagger-def/orders.yaml` hasn't been updated in a while (still shows O-1 without anything else). I think there's been talk of removing it, but I wonder if for the sake of consistency until then, it should also be updated:
https://github.com/transcom/mymove/blob/1ab7bfb16d3fcd095353255a8e0325d5f0346743/swagger-def/orders.yaml#L690-L692

duncan-truss:
Does this feel like a good candidate to you pull out into it's own file in `swagger-def/definitions` since it is shared amongst at least 4 APIs (if you count support and orders)?  Happy to write up a new Jira story for that work.

pearl-truss:
Do you think the happo failures are okay to ignore? I've confirmed that the rank still appears in my browser and I noticed that the "before" image in happo doesn't actually match what's seen here. It was showing that rank was already using a - instead of underscores. I looked into it more and it seems like the happo data is being hardcoded in [Review.stories.jsx](https://github.com/transcom/mymove/blob/master/src/components/Customer/Review/Review.stories.jsx#L78)

pearl-truss:
> Does this feel like a good candidate to you pull out into it's own file in `swagger-def/definitions` since it is shared amongst at least 4 APIs (if you count support and orders)? Happy to write up a new Jira story for that work.

Yeah I think it makes sense to break out

duncan-truss:
> Do you think the happo failures are okay to ignore? I've confirmed that the rank still appears in my browser and I noticed that the "before" image in happo doesn't actually match what's seen here. It was showing that rank was already using a - instead of underscores. I looked into it more and it seems like the happo data is being hardcoded in [Review.stories.jsx](https://github.com/transcom/mymove/blob/master/src/components/Customer/Review/Review.stories.jsx#L78)

@pearl-truss I think you should change the storybook data to use the enum value `E_1` instead of how they were passing in a formatted value https://github.com/transcom/mymove/blob/31bbb64e18d182aa3d2addddca1aa6c4eb83d14c/src/components/Customer/Review/Review.stories.jsx#L78.  That matches the prop data being passed in from what I can see locally.

Screen Shot 2022-01-03 at 11 52 29 AM

duncan-truss (approved):
Thanks for leaving this code better than you found it!

felipe-lee (approved):
LGTM, thanks for making these changes!

amanda5iu (approved):
LGTM!

PR #7977, opened 2021-12-30
MB-10537 (Part 1) Remove outdated duty station data

eamahanna:
## Summary

This PR is step 1 of 3 for cleaning up the duty station data. It deletes the duty stations designated to be deleted in this [spreadsheet](https://docs.google.com/spreadsheets/d/12rNzXdAMNzyB5OGJwrTMr5ts4S3j7hld/edit#gid=1611704837), which was provided by Dan. I deleted all associated data with the duty stations and updated any foreign keys that may refer to one of the duty stations to be deleted. 

## Checking Changes

To test this change locally, run `make db_dev_fresh`. This will run the migrations and remove any old duty station data. You can also run `make db_dev_migrate` to just run the migrations.


migrations/app/migrations_manifest.txt
```diff
@@ -691,3 +691,4 @@
 20211213235039_update_duty_station_location.up.sql
 20211214234555_move_to_gbloc_view.up.sql
 20211221182830_origin_duty_location_to_gbloc_view.up.sql
+20211229190325_delete_incorrect_duty_stations_12_29_2021.up.sql
```
reggieriser, line 694: Nit: Don't think we need the date at the end of this filename since it already starts with a timestamp.

migrations/app/schema/20211229190325_delete_incorrect_duty_stations_12_29_2021.up.sql
```diff
@@ -0,0 +1,337 @@
+DROP TABLE IF EXISTS list_delete;
+CREATE TEMP TABLE list_delete
+(id uuid);
```
reggieriser, line 3: Nice use of a temp table to keep things DRY.  I'm wondering if we should have an `ON COMMIT DROP` here to be more precise about the temp table getting dropped at the end of this migration (each migration is run in its own transaction IIRC).  If I'm reading [the docs](https://www.postgresql.org/docs/12/sql-createtable.html) right, the default is to drop at the end of the session, which may be fine, but I'm not sure when a "session" ends with migrations.

migrations/app/schema/20211229190325_delete_incorrect_duty_stations_12_29_2021.up.sql
```diff
@@ -0,0 +1,337 @@
...
+    ('3315baf8-6fdb-4e13-b798-76a22f9dc763'),
+    ('0c663c4c-7439-4dc1-834e-cd1b4da6963d'),
+    ('f295e563-9ee2-4bb0-bb0a-be3b02751950'),
+    ('f3746f1d-42c3-432d-8f04-5b2d4bfb09a8'),
+    ('e2249511-0387-44d6-848f-0253f4df0577'),
+    ('38567f23-f543-4dec-a05f-ca532f760a43'),
+    ('c08da0fe-74af-4631-a658-ebc243291b7f');
+
+UPDATE service_members
+  SET duty_station_id = 'a268b48f-0ad1-4a58-b9d6-6de10fd63d96'
```
reggieriser, line 331: Nit: Might be nice to put a comment here just noting what duty station this represents and/or why we're mapping all the deleted ones to this particular one.  Same for the other explicit UUID below.

migrations/app/schema/20211229190325_delete_incorrect_duty_stations_12_29_2021.up.sql
```diff
@@ -0,0 +1,337 @@
...
+  WHERE duty_station_id IN (select id from list_delete);
+UPDATE orders
+  SET new_duty_station_id = 'dca78766-e76b-4c6d-ba82-81b50ca824b9', origin_duty_station_id = 'a268b48f-0ad1-4a58-b9d6-6de10fd63d96'
+  WHERE new_duty_station_id IN (select id from list_delete) OR origin_duty_station_id IN (select id from list_delete);
```
reggieriser, line 335: If I'm following this query correctly, we would set both the new/origin duty stations to new values even if only one of the two were on the deleted list.  Is that the desired behavior, or do we want to just replace just the deleted ones?

reggieriser (commented):
This ran fine and seemed to delete the appropriate rows based on what I see in the spreadsheet.  It looks like you covered all the FK relationship remapping given the schema.  I added a few comments for consideration.

Since this will remap duty stations in our deployed environments, is there any practical concern with people "losing" moves they once saw based on their user's GBLOC?  In other words, do we need to try to remap moves to other duty stations in the same GBLOC as the deleted duty station?  That may not be a real problem (or worth the trouble), but just thought I'd mention it.

migrations/app/schema/20211229190325_delete_incorrect_duty_stations_12_29_2021.up.sql
mboren: When I run it locally, the list is not accessible after running migrations, but I agree that adding `ON COMMIT DROP` is a good idea to keep the scope as small as possible.

migrations/app/schema/20211229190325_delete_incorrect_duty_stations_12_29_2021.up.sql
mboren: That's correct. I think the choice is mostly arbitrary. 
Duty location is only used to figure out the GBLOC for services counseling, and `Hunter AAF` (gbloc: CNNQ) is the only one of the deleted duty locations the provides services counseling. So only moves originating there that are in services counseling should get lost.
I'll check if anybody is using that base on staging right now, it would be pretty easy to special case that one.

migrations/app/schema/20211229190325_delete_incorrect_duty_stations_12_29_2021.up.sql
mboren: Done.

migrations/app/schema/20211229190325_delete_incorrect_duty_stations_12_29_2021.up.sql
mboren: Added some comments.

mboren:
I'm taking this over as Emily is offboarding 😢. 
I don't think there's a big danger of moves disappearing on people given the recent GBLOC changes.
The only situation where a move would disappear, is if it originated at Hunter AAF, and you were looking for it as a services counselor. I can't find any examples of moves from Hunter AAF on staging, so I think this will be OK.

migrations/app/schema/20211229190325_delete_incorrect_duty_stations_12_29_2021.up.sql
mboren: I think I'm leaning toward sticking to changing both just because it'll make it more obvious that this migration hit the move, but I could go either way.

reggieriser (approved):
LGTM!  Thanks for making the changes and giving more context. 🚢

Ryan-Koch (approved):

PR #8000, opened 2022-01-07
MB-10966 Update to latest version of milmove-cypress image

gidjin:
## Summary

This updates to the latest cypress image, make sure the cypress tests still pass.

## Setup to Run Your Code

You can try running the tests in the docker image but if the build is passing not sure it's needed.

```sh
make e2e_docker
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,10 +13,10 @@ version: '2.1'
 
 # References for variables shared across the file
 references:
-  circleci-docker: &circleci-docker milmove/circleci-docker:milmove-app-7a49a111589830c94ab7bf750ebdb887d1f891d1
+  circleci-docker: &circleci-docker milmove/circleci-docker:milmove-app-3739e422811794dd0ebe5ab1f0244764520a18c4
 
   # the cypress image to use
-  cypress: &cypress milmove/circleci-docker:milmove-cypress-cf2249e54c69b52625f1ab04eb79f37381d0d3bf
+  cypress: &cypress milmove/circleci-docker:milmove-cypress-3739e422811794dd0ebe5ab1f0244764520a18c4
 
   # https://circleci.com/docs/2.0/databases/#optimizing-postgres-images
   postgres: &postgres circleci/postgres:12.7-ram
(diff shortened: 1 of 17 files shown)


cypress/cypress.json
```diff
@@ -4,7 +4,6 @@
   "reporterOptions": {
     "configFile": "mocha-reporter-config.json"
   },
-  "nodeVersion": "system",
```
duncan-truss, line 7: @gidjin So I think I added this change because Cypress was using their own bundled node version that could be completely different than what we have set in the repository.  It shouldn't have anything to do with the version bumps, although it will be deprecated eventually https://docs.cypress.io/guides/references/configuration#Node-version

.circleci/config.yml
```diff
@@ -13,10 +13,10 @@ version: '2.1'
 
 # References for variables shared across the file
 references:
-  circleci-docker: &circleci-docker milmove/circleci-docker:milmove-app-7a49a111589830c94ab7bf750ebdb887d1f891d1
+  circleci-docker: &circleci-docker milmove/circleci-docker:milmove-app-42dad53c62a147ccd2d497bb95115855c2bf5b92
```
reggieriser, line 16: I don't think we necessarily have to change any of the `milmove-app` references since only the `milmove-cypress` image changed (I think there are only two references to `milmove-cypress`).  Don't know that it hurts as long as we're not accidentally picking up untested modifications in `milmove-app` by using a later hash.

reggieriser (commented):
Added one note, but this seems fine to me given that integration tests passed (although may need master merged in to get past the danger warning).  Do we want to change the hashes to use a merged commit hash from `circleci-docker` instead of the branch's commit hash?  I'm not sure if that matters or not.

cypress/cypress.json
duncan-truss: Yea "system" will become the default in version 9 but in 8.5 it'll still default to "bundled" if it's not in the configuration.

duncan-truss (approved):
I think without the `nodeVersion: "system"` in the config it can be harder to debug issues when running Cypress locally with e2e_test

reggieriser (approved):

gidjin:
I put the `"nodeVersion": "system",` back in, let me know if you need a re-review @duncan-truss

PR #8003, opened 2022-01-07
MB-10885 add destination address type column to shipments table

christopherhuii:
## Summary

This PR adds a column named `destination_address_type` to the `mto_shipments` table. This column is for shipments headed for retirees and separates.

## Setup to Run Your Code
[How to migrate the database](https://transcom.github.io/mymove-docs/docs/backend/setup/database-migrations/#Creating-Migrations)

After running migrations, view the database and see the new column.


migrations/app/schema/20220107193037_add_destination_address_type_to_shipments.up.sql
```diff
@@ -0,0 +1,9 @@
+-- This address type is ONLY applicable for destination addresses for retirees and separatees
+CREATE TYPE address_type AS ENUM (
```
reggieriser, line 2: Are the enum values in this type things that only make sense for destination addresses?  If so, we may want to name it something more specific than `address_type`.

migrations/app/schema/20220107193037_add_destination_address_type_to_shipments.up.sql
```diff
@@ -0,0 +1,9 @@
...
+	'HOME_OF_RECORD',
+	'HOME_OF_SELECTION',
+	'PLACE_ENTERED_ACTIVE_DUTY',
+	'OTHER'
+);
+
+ALTER TABLE mto_shipments ADD COLUMN destination_address_type address_type;
```
reggieriser, line 9: Let's add a `COMMENT ON COLUMN` statement to document this column for schemaspy and other DB tools.  Could also do the same with the enum too (`COMMENT ON TYPE`), although we mainly have been commenting tables and columns so far.

reggieriser (commented):
Migration ran fine and I saw the nullable column and the enum added as expected.  Added a couple of notes for consideration before I approve.

migrations/app/schema/20220107193037_add_destination_address_type_to_shipments.up.sql
reggieriser: I see you added some inline comments (thanks!), but it looks like you're still missing the `COMMENT ON` Postgres statements that would cause comments to be added to the Postgres schema and could be introspected by tools.  Here are some examples from other migrations:

- [`COMMENT ON COLUMN` example](https://github.com/transcom/mymove/blob/17788b5fe89c68cbfbf68021a91210c6ba969447/migrations/app/schema/20211209182000_add_nts_recorded_weight.up.sql#L4)
-  [`COMMENT ON TYPE` example](https://github.com/transcom/mymove/blob/a9576b4b05a7578e3556e2b4bfa6e9d64b3ad0aa/migrations/app/schema/20210823225557_add_sit_extensions_table.up.sql#L11)

reggieriser (approved):
LGTM!  Thanks for the changes.  I'm approving on behalf of the DB reviewers, but probably want a team approval too before merging.

Ryan-Koch (approved):

migrations/app/schema/20220107193037_add_destination_address_type_to_shipments.up.sql
```diff
@@ -0,0 +1,16 @@
+-- Types of shipment locations for destination addresses
+CREATE TYPE destination_address_type AS ENUM (
+	'HOME_OF_RECORD',
+	'HOME_OF_SELECTION',
+	'PLACE_ENTERED_ACTIVE_DUTY',
+	'OTHER'
```
Ryan-Koch, line 6: Do you mind switching this to: `OTHER_THAN_AUTHORIZED`? This came up during our discovery meeting this morning (sorry for the short notice) as a language change that was desired.

PR #8016, opened 2022-01-11
MB-10783 allow the too to update shipments

YanZ777:
## Summary

I noticed that there were two other things on the FE that prevented TOO from updating moves, so I fixed those.

I took off the `readOnly` from the agent id for the `ghc.yaml` because it was preventing the TOO from updating moves. The customer API does not have the agent id as readOnly.

I also sanitized the `eTag` from the storage facility. 

Please let me know if there are some unknown repercussions with taking off the `readOnly` from the agent id in the `ghc.yaml` or sanitizing the `eTag` from the storage facility when sending it to the API.

There's [a discussion](https://ustcdp3.slack.com/archives/C029R9LDTD4/p1641939864040300) about removing or disabling the button from the FE side for shipments that are not editable. Removing or disabling the button is not handled in this PR as it is out of scope.

With regards to the code changes on the backend, those should be fairly straightforward as I just added more cases to check for the TOO role. 

There is an outstanding problem with how roles are handled on the backend, explained more in [this thread](https://ustcdp3.slack.com/archives/C029R9LDTD4/p1641863265026000), but that is outside the scope of this PR.

As for the tests, I've split out the tests that check if the user is allowed to update the shipment into its own function. I've moved the related services counselor tests there and added the other shipment statuses as test cases. Additionally, I've added tests for the TIO and TOO that also test all the shipment statuses.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make office_client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the office app
2. Login as a TOO (Must be TOO only! If you log in as mixed roles, strange things will happen. See [this thread](https://ustcdp3.slack.com/archives/C029R9LDTD4/p1641863265026000) for details, there is [a ticket](https://dp3.atlassian.net/browse/MB-11059) out for it.)
3. Access any moves with these locators: `HGNTSR`, `NTSSUB`, or `HNTMIS`. 
4. Attempt a shipment on the move you've selected.
5. They should be updatable.

### Frontend


pkg/services/mto_shipment/mto_shipment_updater.go
```diff
@@ -217,13 +217,24 @@ func (e StaleIdentifierError) Error() string {
...
-			if session.Roles.HasRole(roles.RoleTypeTIO) {
+			if isTIO || isTOO {
+				return true, nil
+			}
+		case models.MTOShipmentStatusCancellationRequested:
+			if isTOO {
+				return true, nil
+			}
+		case models.MTOShipmentStatusDiversionRequested:
+			if isTOO {
```
donaldthai, line 237: Seems like we should allow the TOO to be able to edit in `CANCELED` status based in the [slack thread](https://ustcdp3.slack.com/archives/C029R9LDTD4/p1642002803047400?thread_ts=1641939864.040300&cid=C029R9LDTD4).

YanZ777: Done!

donaldthai (approved):
LGTM

PR #8028, opened 2022-01-14
Add where clause to filter out test user email from devlocal service member users list

duncan-truss:
## Summary

Before A-team starts outcome work on PPM, I thought it would be good to filter out all of the test users from the devlocal sign in screen that are created when populating the database with the e2e devseed.  These are users that have the default email `first.last@login.gov.test` when the devseed calls MakeUser/MakeServiceMember without any override assertions.  When we do create PPM specific data usually the email describes the purpose such as `needs@orde.rs`.  The devlocal login sorts users by last created date and there is a user limit so it often hides the most useful premade users.

The fetchAppUserIdentities method is only used for querying the users for devlocal so does not affect any real app usage.

This doesn't impact the office app devlocal sign in list only the customer app. 

## Setup to Run Your Code

<details>

##### Terminal 1

Populate the develepment database with the e2e devseed scenario

```sh
make db_dev_e2e_populate
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. In your browser open http://milmovelocal:3000
2. Select Local Sign In from the homepage 
3. The users list should only contain the custom service member users to select from and not those created by the office.

### Frontend

### Backend

## Screenshots

Before

Screen Shot 2022-01-14 at 10 40 56 AM

After

Screen Shot 2022-01-14 at 10 36 00 AM

### DIFF
--- a/pkg/models/user_test.go
+++ b/pkg/models/user_test.go
@@ -257,29 +257,32 @@ func (suite *ModelSuite) TestFetchAppUserIdentities() {
 
 	suite.Run("service member", func() {
 
-		// Regular service member
+		// Create a user email that won't be filtered out of the devlocal user query w/ a default value of
+		// first.last@login.gov.test
+		user := testdatagen.MakeUser(suite.DB(), testdatagen.Assertions{
+			User: User{
+				LoginGovEmail: "test@example.com",
+			},
+		})
+
+		testdatagen.MakeServiceMember(suite.DB(), testdatagen.Assertions{
+			ServiceMember: ServiceMember{
+				UserID: user.ID,
+				User:   user,
+			},
+		})
+
+		// This service member will be filtered out from the result because we haven't overridden the default email
 		testdatagen.MakeDefaultServiceMember(suite.DB())
+
 		identities, err := FetchAppUserIdentities(suite.DB(), auth.MilApp, 5)
+
 		suite.NoError(err)
 		suite.NotEmpty(identities)
 		suite.Equal(1, len(identities))
-
-		if len(identities) > 1 {
-			suite.NotNil(identities[0].ServiceMemberID)
-			suite.Nil(identities[0].OfficeUserID)
-		}
-
-		// Service member is super user
-		testdatagen.MakeDefaultServiceMember(suite.DB())
-		identities, err = FetchAppUserIdentities(suite.DB(), auth.MilApp, 5)
-		suite.NoError(err)
-		suite.NotEmpty(identities)
-		suite.Equal(2, len(identities))
-
-		if len(identities) == 2 {
-			suite.NotNil(identities[1].ServiceMemberID)
-			suite.Nil(identities[1].OfficeUserID)
-		}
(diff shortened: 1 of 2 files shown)


pkg/models/user_test.go
```diff
@@ -257,29 +257,32 @@ func (suite *ModelSuite) TestFetchAppUserIdentities() {
...
-		suite.NotEmpty(identities)
-		suite.Equal(1, len(identities))
 
-		if len(identities) > 1 {
-			suite.NotNil(identities[0].ServiceMemberID)
-			suite.Nil(identities[0].OfficeUserID)
-		}
+		identities, err := FetchAppUserIdentities(suite.DB(), auth.MilApp, 5)
 
-		// Service member is super user
```
duncan-truss, line 272: There used to be a column called is_superuser on the users table but it was removed a while ago and this test wasn't doing anything so I'm deleting it.

ronaktruss:
@duncan-truss Are the user with first.last@login.gov.test not used for other parts of the application / are there other more specific logins that serve the same purpose? I know I had used some of these users when working on load testing.

felipe-lee (approved):
I like that this is a quick solution to this for the purpose of seeing useful customers to log in as, but I also think we still have an underlying issue that most of these users are also mistakenly created. People don't always realize that several testdatagen function will create a new user if you don't pass it in and will call several of these in a row which creates several users when they only needed one user.

duncan-truss:
> @duncan-truss Are the user with [first.last@login.gov.test](mailto:first.last@login.gov.test) not used for other parts of the application / are there other more specific logins that serve the same purpose? I know I had used some of these users when working on load testing.

@ronaktruss This user email is only from data created in the devseed (make db_dev_e2e_populate) and e2e_basic (make db_e2e_up) so it adds data for local/ephemeral deploys/and in Cypress tests.  It's the byproduct of creating moves used in the office for the most part where we never intend to login as the service member but there must be some record for the move to exist.

In load testing I think you are calling [createUser](https://github.com/transcom/mymove/blob/6ed72bf51a630cf834d188b707fc46f302079084/pkg/auth/authentication/devlocal.go#L410) which creates a unique email (timestamp+nonce@example.com) based on the timestamp if you aren't overriding it, same as clicking the "Create a new Milmove User" button.

You could still enter the email into the login with user email if you had one unique user with that email but it becomes useless when you have multiple with the same email.  We should probably add an input to login by user id, which is how the user list works.

duncan-truss:
> I like that this is a quick solution to this for the purpose of seeing useful customers to log in as, but I also think we still have an underlying issue that most of these users are also mistakenly created. People don't always realize that several testdatagen function will create a new user if you don't pass it in and will call several of these in a row which creates several users when they only needed one user.

@felipe-lee That's a good point I wasn't thinking so much about the test factories that say create multiple users incorrectly.  I am down to look at that further as we work on PPMs.

PR #8030, opened 2022-01-14
[MB-9276] Prime API totalWeight entitlement should include dependents if authorized

duncan-truss:
## Summary

The Prime API model to payload response code was always defaulting to using the self weight allowance value even when the entitlement/allowance had set dependents authorized to true.

The entitlements code is a little awkward because the weight allowance lookups depend on the rank/grade of the service member, which is stored as part of the orders model.  We don't actually store the weight allowance in the database, only the authorized weight/max billable weight that the TOO/TIO can use to raise the weight limit above the standard allowances.  If the authorized weight is not set it returns the value of the weight allowance (totalWeight) to the Prime API.

## Setup to Run Your Code

<details>

##### Terminal 1

Populate the database and start the server

```sh
```

##### Terminal 2

Start the UI locally.

```sh
make office_client_run
```

### Additional steps

1. Open up Postman and import the prime.yaml collection if it doesn't exist already.  Make sure you are targeting against the local Prime API Development environment.
2. Choose the move-task-order/getMoveTaskOrder request and change the value of moveID under the params tab to **HGNTSR**
3. Send the request and find the entitlement block in the response.
4. Verify that the totalWeight value is 8000 with dependents authorized instead of the 5000 lbs when dependents are unauthorized.

### Frontend

### Backend

## Screenshots

**Prime getMove request for #HGNTSR with dependents authorized:**

Screen Shot 2022-01-18 at 5 23 01 PM

**Prime getMove request for #HGNTSR after changing dependents authorized to false as the TOO office user**

Screen Shot 2022-01-18 at 5 28 16 PM

Screen Shot 2022-01-18 at 5 30 36 PM

### DIFF
--- a/pkg/handlers/primeapi/payloads/model_to_payload.go
+++ b/pkg/handlers/primeapi/payloads/model_to_payload.go
@@ -134,13 +134,12 @@ func Order(order *models.Order) *primemessages.Order {
 	if order.Grade != nil && order.Entitlement != nil {
 		order.Entitlement.SetWeightAllotment(*order.Grade)
 	}
-	entitlements := Entitlement(order.Entitlement)
 
 	payload := primemessages.Order{
 		CustomerID:             strfmt.UUID(order.ServiceMemberID.String()),
 		Customer:               Customer(&order.ServiceMember),
 		DestinationDutyStation: destinationDutyStation,
-		Entitlement:            entitlements,
+		Entitlement:            Entitlement(order.Entitlement),
 		ID:                     strfmt.UUID(order.ID.String()),
 		OriginDutyStation:      originDutyStation,
 		OrderNumber:            order.OrdersNumber,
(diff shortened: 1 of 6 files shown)


pkg/handlers/primeapi/payloads/model_to_payload.go
```diff
@@ -134,13 +134,12 @@ func Order(order *models.Order) *primemessages.Order {
 	if order.Grade != nil && order.Entitlement != nil {
 		order.Entitlement.SetWeightAllotment(*order.Grade)
```
duncan-truss, line 135: This is a necessary step because we don't store the weight allowance entitlements data in the database so this adds the data based on the grade.  If we wanted to store it in the database, it would need to be updated anytime grade is changed or dependents authorized changed.

pkg/models/ghc_entitlements.go
```diff
@@ -69,3 +69,16 @@ func (e *Entitlement) AuthorizedWeight() *int {
 		return nil
 	}
 }
+
+// WeightAllowance will return the service member's weight allotment based on their rank and if dependents are
+// authorized
+func (e *Entitlement) WeightAllowance() *int {
```
duncan-truss, line 75: I didn't think this logic belonged in the model to payload function that should just be about converting to swagger types if we can help it.

pkg/handlers/primeapi/payloads/model_to_payload.go
pearl-truss: I think it makes sense to keep this out of the database sense it's a. derived calculation

pearl-truss (approved):
LGTM

PR #8040, opened 2022-01-18
MB-9872: weight allowances

ronaktruss:
## Summary

This should resolve the mismatching weight allowance (visually displayed) on pages when a customer has dependents. Is there a good way to capture this scenario in a test? 

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

If you login as a customer to the mymove application and create a move with dependents you should see the correct weight allowance on all of the pages. See screenshot and story for references. 

### Frontend

## Screenshots
Navy with E-8 grade with dependents
Screen Shot 2022-01-18 at 2 56 12 PM
Screen Shot 2022-01-18 at 2 56 26 PM
Screen Shot 2022-01-18 at 2 56 45 PM


src/components/Customer/MtoShipmentForm/MtoShipmentForm.jsx
```diff
@@ -207,8 +207,8 @@ class MtoShipmentForm extends Component {
                     <h1>{shipmentForm.header[`${shipmentType}`]}</h1>
 
                     <Alert type="info" noIcon>
-                      Remember: You can move {serviceMember.weight_allotment.total_weight_self} lbs total. You’ll be
-                      billed for any excess weight you move.
+                      Remember: You can move {orders.authorizedWeight} lbs total. You’ll be billed for any excess weight
```
duncan-truss, line 210: Authorized weight is something that the TOO/TIO sets after the move is submitted, it does default to the weight allowance if there isn't a value yet but that also looks at the dependents authorized value that is also set by the office user.  I don't think we can rely on that before the move is submitted here in the customer onboarding.

Can we look at the value of has_dependents in the orders and then choose the total_weight_self or total_weight_dependents based on if that is true/false?

duncan-truss: Yea I think we want to be consistent everywhere in the customer UI, didn't mean to make it specific to this component/file.

duncan-truss:
@ronaktruss Do you want to [deploy this to the ephemeral environment](https://trussworks.slack.com/archives/C01T40WC4G1) so design can more easily review it?

duncan-truss:
I'm seeing inconsistent numbers after completing profile onboarding and then going back to edit the service member's grade (say changing from E-8 to E-1).  The customer header looks accurate though.

After saving the new grade the Flash message shows the wrong weight allotment and the add new shipment info looks wrong as well.

Screen Shot 2022-01-19 at 3 16 21 PM
Screen Shot 2022-01-19 at 3 16 13 PM
Screen Shot 2022-01-19 at 3 22 00 PM

I'm seeing an error when trying to change the has dependents question in the edit orders form after filling it out initially.  I want to say this is a known bug and not something you changed but could be worth fixing to test.  It seems like patchOrders api call is missing the `ordersId` value

Screen Shot 2022-01-19 at 3 27 10 PM

ronaktruss:
I think this is likely the same issue that is discussed in [MB-7998](https://dp3.atlassian.net/browse/MB-7998).

duncan-truss (approved):
Changes on the initial flow looks correct to me.  Looks like we'll take care of those edits to rank and orders in the next bug story.

dinabeigelman (approved):
Dina and James looked at it and it looks gr8

duncan-truss (approved):

PR #8047, opened 2022-01-20
[MB-10525] backend returns failure if create payment request called for external shipment

jacquelineIO:
## Summary

This is a backend story. If `CreatePaymentRequest` is called when `MTOShipment.UsesExternalVendor` is `true`, return a failure.

For acceptance, the Prime Simulator and the `prime-api-client` should not show any shipments that are using external vendors so there is no acceptance to run on staging. UT was added for the service object to verify this case.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the
2. Login as a
3.

### Frontend

### Backend


reggieriser (commented):
Had a few small comments for consideration, but this generally looks good.

I did notice there's a place in `ServiceParamLookupInitialize` where we already look up the shipment and could potentially just look at the `UsesExternalVendor` flag there and error if true, but I'm not sure that's necessarily any better.  Probably less code, but you might do a good bit of database work before you get there (although that probably only matters if we expect a payment request to be submitted on an external shipment often, which I doubt is the case).  No need to change -- just thought I'd mention it.

jacquelineIO:
> I did notice there's a place in `ServiceParamLookupInitialize` where we already look up the shipment and could potentially just look at the `UsesExternalVendor` flag there and error if true, but I'm not sure that's necessarily any better. Probably less code, but you might do a good bit of database work before you get there (although that probably only matters if we expect a payment request to be submitted on an external shipment often, which I doubt is the case). No need to change -- just thought I'd mention it.

Yes I looked at that, my thought was to keep it at the top. I'm not sure if there is a "better" way but I don't think this logic should be buried all the way down in lookups for service items. The shipment is only checked once so that shouldn't be much of a burden.

pkg/services/payment_request/payment_request_creator.go
```diff
@@ -202,6 +210,40 @@ func (p *paymentRequestCreator) CreatePaymentRequest(appCtx appcontext.AppContex
...
+	reServiceItem := mtoServiceItem.ReService
+	serviceItemMessageString := " RE Service Item Code: <" + string(reServiceItem.Code) + "> Name: <" + reServiceItem.Name + ">"
+	return mtoMessageString + prMessageString + mtoServiceItemString + serviceItemMessageString
+}
+
+func (p *paymentRequestCreator) validShipment(appCtx appcontext.AppContext, shipmentID *uuid.UUID, shipmentIDs map[string]bool) error {
+	if shipmentID != nil {
+		shipmentIDStr := shipmentID.String()
+		if _, found := shipmentIDs[shipmentIDStr]; !found {
+			shipmentIDs[shipmentIDStr] = true
```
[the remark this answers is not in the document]
jacquelineIO: The `UsesExternalVendor` check will not matter. This map is just saying that we looked at the shipment. 

However, about doing it after the `Find`, let's say the `Find` failed, are you saying that we should be checking for this shipment again, because that is what would happen here if I move `shipmentIDs[shipmentIDStr] = true` until after the err checking is done?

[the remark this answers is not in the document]
reggieriser: Now that I think about it more, it should be fine where it is.  If the `Find` fails, it's bailing out of the whole transaction anyway so it doesn't matter what happens next go around since there won't be one.  I wasn't taking all the error returns into account.

pkg/services/payment_request/payment_request_creator.go
```diff
@@ -202,6 +210,40 @@ func (p *paymentRequestCreator) CreatePaymentRequest(appCtx appcontext.AppContex
...
+) string {
+	mtoServiceItemString := " MTO Service item ID <" + mtoServiceItemID.String() + ">"
+	reServiceItem := mtoServiceItem.ReService
+	serviceItemMessageString := " RE Service Item Code: <" + string(reServiceItem.Code) + "> Name: <" + reServiceItem.Name + ">"
+	return mtoMessageString + prMessageString + mtoServiceItemString + serviceItemMessageString
+}
+
+func (p *paymentRequestCreator) validShipment(appCtx appcontext.AppContext, shipmentID *uuid.UUID, shipmentIDs map[uuid.UUID]bool) error {
+	if shipmentID != nil {
+		// shipmentIDStr := shipmentID.String()
```
jacquelineIO, line 227: need to remove this comment

donaldthai (approved):
LGTM

pkg/services/payment_request/payment_request_creator.go
```diff
@@ -202,6 +210,40 @@ func (p *paymentRequestCreator) CreatePaymentRequest(appCtx appcontext.AppContex
...
+		if _, found := shipmentIDs[*shipmentID]; !found {
+			shipmentIDs[*shipmentID] = true
+			var mtoShipment models.MTOShipment
+			err := appCtx.DB().Find(&mtoShipment, *shipmentID)
+			if err != nil {
+				appCtx.Logger().Error("paymentRequestCreator.validShipment query error", zap.Error(err))
+				return apperror.NewQueryError("MTOShipment", err, fmt.Sprintf("paymentRequestCreator.validShipment:MTOShipmentID %s", shipmentID.String()))
```
reggieriser, line 234: A `Find` or `First` can legitimately return a `sql.ErrNoRows` that isn't really a query error ([see example](https://github.com/transcom/mymove/blob/b9db7b1a6066b55f8373887c7b836a0a442eed18/pkg/services/invoice/ghc_payment_request_invoice_generator.go#L186-L193)).  Not sure if that could really happen here given the FKs in the DB.  But if it did happen, a 500 would be returned as opposed to a 404 and may raise an alert (which might be OK if we don't ever expect this).

reggieriser (approved):
LGTM!  Thanks for making the changes.  I did note something about `sql.ErrNoRows` to think about, but don't think it's a blocker.

pkg/services/payment_request/payment_request_creator.go
jacquelineIO: Thank you for the reference, I will update to do that.

PR #8075, opened 2022-01-25
MB-11181-update-format-weights-import

NamibiaTorres:
## Summary

The purpose of this PR is to update the imports for formatWeight from shared/formatters.js to utils/formatters.js. The FE is moving away from using the shared folder. To start, we moved over the imports for the  formatWeight function since it was used to reformat weight allowances in a recent PR. 

[this article](https://github.com/transcom/mymove/blob/master/docs/adr/0048-frontend-file-org.md) explains more about the FE change.

## Setup to Run Your Code
To run this code, just make sure that there are no more `import { formatWeight } from 'shared/formatters';` in any of the frontend code. You may also chose to run tests for the files that have been changed `yarn test <test_file_path>`


src/shared/formatters.test.js
```diff
@@ -4,22 +4,6 @@ import moment from 'moment';
 import PAYMENT_REQUEST_STATUS from 'constants/paymentRequestStatus';
 
 describe('formatters', () => {
-  describe('formatWeight', () => {
```
felipe-lee, line 7: Can you add this test in the utils formatter tests file: https://github.com/transcom/mymove/blob/MB-11181-update-formatWeights-import-NT/src/utils/formatters.test.js

src/shared/formatters.js
```diff
@@ -149,13 +149,13 @@ export function parseSwaggerDate(dateString) {
...
 
 // Format a weight with lbs following, e.g. 4000 becomes 4,000 lbs
-export function formatWeight(weight) {
-  if (weight) {
-    return `${weight.toLocaleString()} lbs`;
-  } else {
-    return '0 lbs';
-  }
-}
+// export function formatWeight(weight) {
```
felipe-lee, line 152: We won't be using it anymore in here so I think it's fine to completely remove this rather than just commenting it out.

felipe-lee (commented):
Mostly lgtm, just two comments

pearl-truss (commented):
Pending  Felipe's comments this looks good to go. Thanks for doing the extra work of getting this formatter in the right place :)

src/shared/formatters.test.js
NamibiaTorres: Hi Felipe, it should be there already. I added it last week. This is what I see:
`  describe('formatWeight', () => {
    describe('when formatting a integer weight', () => {
      const weight = 4000;
      const formattedWeight = formatters.formatWeight(weight);
      it('should be be formatted as expected', () => {
        expect(formattedWeight).toEqual('4,000 lbs');
      });
    });
    describe('when formatting a integer weight', () => {
      const weight = '';
      const formattedWeight = formatters.formatWeight(weight);
      it('should be be formatted as expected', () => {
        expect(formattedWeight).toEqual('0 lbs');
      });
    });
  });`

src/shared/formatters.js
NamibiaTorres: Agree! I realized I forgot to push my last commit, which was removing the commends. I'll do that now. Thanks for catching this!

NamibiaTorres: Alrighty! Just pushed that last commit. Thank you!

src/shared/formatters.test.js
felipe-lee: 😞 I missed that somehow. Thanks for correcting me!

felipe-lee (approved):
LGTM, thanks for making these changes!

PR #8076, opened 2022-01-25
MB-11140: PPM DB ADR

ronaktruss:
## Summary

We have decided that option 3 should be utilized and that we should create a new table and implement a new pattern for generic shipment information and information related to a specific shipment type.

### DIFF
--- a/docs/adr/0067-ppm-db-design.md
+++ b/docs/adr/0067-ppm-db-design.md
@@ -1,0 +1,64 @@
+# TBD
+
+**User Story:** [MB-11140](https://dp3.atlassian.net/browse/MB-11140)
+
+Currently, there is a `personally_procurred_moves` table that is used for PPMs while there is a separate `mto_shipments` table that is used for HHG and other shipments types. Since the existing table doesn't follow the current conventions and we are working on this outcome it is a good time to revisit the DB structure to see if the original design continues to be the best choice moving forward.
+
+## Decision Drivers
+
+* Time and complexity
+  * Do we have bandwidth to make these changes?
+  * Do we have enough capacity handle the unknowns or deal with future issues this decision may cause?
+  * Do we need more information?
+    * Do we have enough future capacity to delay this?
+* Flexibility
+  * How likely is this solution to support new shipment types or even existing shipments?
+* Consistency
+  * How consistent are our design patterns?
+    * Are they intuitive?
+
+## Considered Alternatives
+
+* [Use old table](#use-old-table)
+* [Modify `mto_shipments`](#modify-mto_shipments)
+* [Create a new table](#create-a-new-table)
+
+## Decision Outcome
+
+* Chosen Alternative: _[alternative 1]_
+* _[justification. e.g., only alternative, which meets KO criterion decision driver | which resolves force force | ... | comes out best (see below)]_
+* _[consequences. e.g., negative impact on quality attribute, follow-up decisions required, ...]_ <!-* optional -->
+
+## Pros and Cons of the Alternatives
+
+### Use old table
+
+Continue using the `personally_procurred_moves` table
+
+* `+` Less work, most of the plumbing is already setup and can be leveraged to account for modifications made
+* `-` The PPM code will remain siloed
+* `-` The PPM models and relationship will be unique so there will be additional overhead to learn this relationship when switching to PPMs from another area of the project
+* `-` The nomenclature is not clear. We currently have moves which contain PPMs which can cause some confusions, as we normally consider the relationship to be moves contain shipments.
+
+### Modify `mto_shipments`
+
+Incorporate the columns on the `personally_procurred_moves` table into the `mto_shipments` table
+
+* `+` This will make PPMs consistent with how we store and retrieve other shipment types
+* `+` A `CHECK` constraint can be used to ensure only PPM related fields are used with a PPM shipment type
+* `-` The `mto_shipments` table has a growing number of fields and it’s hard to know which fields are applicable for a given shipment type
+* `-` As we continue to support more types of shipments, this is beginning to feel like a workaround solution rather than the best solution
+* `-` This is a sizable amount of work, and there is some risk of the amount of work ballooning into more work that originally expecting.
+
+### Create a new table[¹](#references)
+
+Create a new table that is a child of the `mto_shipments` table that holds ppm specific info
+
+* `++` This structure is more flexible and can be implemented retroactively for old shipment types and support new shipment types
+* `-` If the existing patterns for shipments are not modified, we will continue to have multiple patterns for different types of shipments
+* `-` The most amount of work and the most unknowns (slightly more than [modifying mto_shipments](#modify-mto_shipments))
+
+## References
+
+1. [^](#create-a-new-table) [Class table inheritance](https://martinfowler.com/eaaCatalog/classTableInheritance.html)
+2. [Slack thread about this conversation](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1641937842037600)
--- a/docs/adr/0067-ppm-db-design.md
+++ b/docs/adr/0067-ppm-db-design.md
@@ -35,8 +35,8 @@ Currently, there is a `personally_procurred_moves` table that is used for PPMs w
 
 Continue using the `personally_procurred_moves` table
 
-* `+` Less work, most of the plumbing is already setup and can be leveraged to account for modifications made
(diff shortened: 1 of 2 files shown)


docs/adr/0067-ppm-db-design.md
```diff
@@ -0,0 +1,64 @@
...
+* `-` The `mto_shipments` table has a growing number of fields and it’s hard to know which fields are applicable for a given shipment type
+* `-` As we continue to support more types of shipments, this is beginning to feel like a workaround solution rather than the best solution
+* `-` This is a sizable amount of work, and there is some risk of the amount of work ballooning into more work that originally expecting.
+
+### Create a new table[¹](#references)
+
+Create a new table that is a child of the `mto_shipments` table that holds ppm specific info
+
+* `++` This structure is more flexible and can be implemented retroactively for old shipment types and support new shipment types
+* `-` If the existing patterns for shipments are not modified, we will continue to have multiple patterns for different types of shipments
```
duncan-truss, line 58: I think another positive here is that we'll be able to leverage some of the same models, payload, handlers, and service objects for shipments that are already written (more so on the office side than the customer side).  It may not be painless and will require another table join but on the office side we'll know that shipment queries will include PPM where we need them for the Services Counselor or TIO.

docs/adr/0067-ppm-db-design.md
```diff
@@ -0,0 +1,64 @@
...
+* `-` The most amount of work and the most unknowns (slightly more than [modifying mto_shipments](#modify-mto_shipments))
+
+## References
```
pearl-truss, line 61: I think the PPM confluence doc's backend section should also be linked here https://dp3.atlassian.net/wiki/spaces/MT/pages/1605238794/PPM+Bookings+Technical+Discovery#Database

docs/adr/0067-ppm-db-design.md
pearl-truss: Another plus is that this option is the most semantic with PPMs being a clear child of shipments

ronaktruss: Thanks for pointing this out. 

Could the same be said for option 2 where we just add to the existing shipments table? We would need to modify the existing code to account for the additional fields.

Right now option 1 (using the existing ppm table) is not accounted for when getting shipments on the office side. Is the work to support this significantly more work than the work needed to support this option?

docs/adr/0067-ppm-db-design.md
```diff
@@ -0,0 +1,64 @@
...
+
+Currently, there is a `personally_procurred_moves` table that is used for PPMs while there is a separate `mto_shipments` table that is used for HHG and other shipments types. Since the existing table doesn't follow the current conventions and we are working on this outcome it is a good time to revisit the DB structure to see if the original design continues to be the best choice moving forward.
+
+## Decision Drivers
+
+* Time and complexity
+  * Do we have bandwidth to make these changes?
+  * Do we have enough capacity handle the unknowns or deal with future issues this decision may cause?
+  * Do we need more information?
+    * Do we have enough future capacity to delay this?
```
pearl-truss, line 13: I'm not sure what this is referring to? Is this asking about delaying making a decision?

docs/adr/0067-ppm-db-design.md
duncan-truss: Yes, I think option 2 would get some of the same benefit of being returned in existing queries in the office and prime side.

For option 1 we'd be creating parallel API requests, handlers, and service objects on the office side most likely that don't currently exist.  I think that is absolutely more new work.  However we do have [open questions about how office users are segmented to work on PPM vs HHG](https://dp3.atlassian.net/browse/MB-10177) so I don't think we know for sure when we need to include things for viewing/editing.

docs/adr/0067-ppm-db-design.md
ronaktruss: I was trying to capture some questions that would help guide the ADR. Sometimes when there is not enough information it can be worthwhile to delay forcing a solution or choosing a "simplistic" solution that may not solve future unknowns. In this line, I was specifically trying to capture that if we did decide to go with a simplistic solution would we have future capacity to account the future work that may be necessary.

docs/adr/0067-ppm-db-design.md
```diff
@@ -0,0 +1,64 @@
...
+* _[justification. e.g., only alternative, which meets KO criterion decision driver | which resolves force force | ... | comes out best (see below)]_
+* _[consequences. e.g., negative impact on quality attribute, follow-up decisions required, ...]_ <!-* optional -->
+
+## Pros and Cons of the Alternatives
+
+### Use old table
+
+Continue using the `personally_procurred_moves` table
+
+* `+` Less work, most of the plumbing is already setup and can be leveraged to account for modifications made
```
duncan-truss, line 38: We say it's less work but it's also not up to standards (api conventions, handler errors, validations, migrating queries from models to service objects that don't exist yet) and so we have to rewrite most of it anyway even if we kept the seperate table.

docs/adr/0067-ppm-db-design.md
duncan-truss: Also if you want a TDR that talks about combining database tables here is a previous example https://github.com/transcom/mymove/blob/master/docs/adr/0055-consolidate-moves-and-mtos.md

docs/adr/0067-ppm-db-design.md
```diff
@@ -0,0 +1,67 @@
...
+* `-` The `mto_shipments` table has a growing number of fields and it’s hard to know which fields are applicable for a given shipment type
+* `-` As we continue to support more types of shipments, this is beginning to feel like a workaround solution rather than the best solution
+* `-` This is a sizable amount of work, and there is some risk of the amount of work ballooning into more work that originally expecting.
+
+### Create a new table[¹](#references)
+
+Create a new table that is a child of the `mto_shipments` table that holds ppm specific info
+
+* `++` This structure is more flexible and can be implemented retroactively for old shipment types and support new shipment types
+* `+` This is semantically the most clear and intuitive
```
Ryan-Koch, line 59: This point makes me like this option a lot. The `mto_shipments` table is already getting to a point where one has to know a lot of about how shipments work in order to understand why some fields are filled out, empty, or configured a certain way at different times. Having a separation for the PPM specific data as a child I think reduces cognitive load in a way that will be helpful to folks working on this code say 6+ months from now.

docs/adr/0067-ppm-db-design.md
```diff
@@ -0,0 +1,67 @@
...
+* Positive Outcomes: We will have a pattern that is easier to understand and will potentially be helpful in the future when implementing new shipments. The tentative plan is to test this new pattern out with PPMs, and if things go smoothly, revisit this to be the default DB pattern for all new shipment types moving forward.
+* Consequences: If this new pattern is not utilized elsewhere we may be adding more complexity by introducing another pattern that is only partially used.
+
+## Pros and Cons of the Alternatives
+
+### Use old table
+
+Continue using the `personally_procurred_moves` table
+
+* `+` Less work, most of the plumbing is already setup and can be leveraged to account for modifications made
```
felipe-lee, line 38: Small nit:

```suggestion
* `+` Less work, most of the plumbing is already set up and can be leveraged to account for modifications made.
```

docs/adr/0067-ppm-db-design.md
```diff
@@ -0,0 +1,67 @@
...
+* `-` Additional work would be needed to support he office workflows and to update the code to current standards
```
felipe-lee, line 39:
```suggestion
* `-` Additional work would be needed to support the office workflows and to update the code to current standards.
```

felipe-lee (approved):
Small suggestions, but otherwise lgtm, thanks for putting this together!

docs/adr/index.md
```diff
@@ -70,6 +70,8 @@ This log lists the architectural decisions for DP3 Infrastructure.
 - [ADR-0063](0063-use-openapi-to-compile-api-specs.md) - Using `openapi` CLI to compile API specs
 - [ADR-0064](0064-use-stateless-services-with-context.md) - Use stateless services with context
 - [ADR-0065](0065-use-office-app-for-prime-ui.md) - *Use Office application for Prime UI*
+- [ADR-0066](0066-use-custom-nullable-types-for-patch-requests.md) - Use custom nullable types for patch requests
```
gidjin, line 73: Thanks for adding the missing one!

gidjin (approved):
LGTM

PR #8082, opened 2022-01-25
[MB-11206] Remove etag from storage facility address on shipment form update

hannaliebl:
## Summary

I added a helper to remove the eTag field from a given object, and then used it in the helper that's used in the shipment form that formats the MTO Shipment for API consumption.

However, I'd like some further info about WHY this stripping of the eTag is needed - we don't do this for other address fields, for example. Is this some difference in implementation of the storage facility specifically? For example, destinationAddress and pickupAddresses on MTO shipments contain eTag values that match the parent eTag sent in the header: why is the storage facility eTag different? If this is expected and we should strip it that's fine, I just feel like this is the first time I've had to explicitly remove an eTag on a PATCH, but maybe there are other places that follow this pattern.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. On this branch, Login as a service counselor
2. Add an NTS-r shipment, or find a move where one exists (not sure there is one on devseed, though?)
3. Go back and try to edit the NTS-release shipment: you should be able to without any errors!
4. Contrast this with master: you should see a 422 error on update because the etag of the storageFacility.address is readOnly

- [X] Request review from a member of a different team.

### Frontend


donaldthai (approved):
LGTM

To answer your question around why sending back the `eTag` in a request is causing us trouble now, I think it has something to do with how the `read-only` attribute was not supported before and now it is with the recent `go-swagger` [version update](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1642111438056500?thread_ts=1642104165.046500&cid=CP6PTUPQF).

It's possible that the read-only validation is being triggered for the storage facility address **first**, even though we may edit plus send back other addresses.

Hope that answers the question!

YanZ777 (approved):
I'm not sure if it's better to create a separate function or to just have the eTag be removed in the `formatAddressForAPI`. `formatAddressForAPI` is being used by more things, so I don't know what the full repercussions are if we were to add it to that function.

PR #8088, opened 2022-01-26
MB-11055 Calculate RequiredDeliveryDate for NTS shipments

mlm55:
## Summary
Current RDD calculations require a `pickupAddress` and a `destinationAddress`, this PR updates the calculation to use the storage facility address instead of a pickup/destination address when necessary. There were 2 places that RDD might be calculated - when the prime updates a shipment and when a shipment is approved on the office side.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Run `DEVSEED_SUBSCENARIO=nts_and_ntsr make db_dev_e2e_populate `
2. Login as a TOO
3. Look at the NTSSUB move and update the missing storage facility information for the NTS and NTSR shipments
4. In the DB, add a `ScheduledPickupDate`, and make sure that `RequiredDeliveryDate` is null for the NTSR shipment
5. In the app, approve both shipments and verify that a RequiredDeliveryDate is set for the NTSR
6. Use the prime simulator to update the NTS shipment with a ScheduledPickupDate, currently the simulator requires pickup/address and destination address to be filled in, but the calculation will use the storage facility address to calculate required delivery date
7. Verify that there's a requiredDeliveryDate for the NTS shipment

### Backend


jenniferlynparsons (approved):
This all looks good and works as expected.

pkg/services/mto_shipment/mto_shipment_updater.go
```diff
@@ -714,8 +714,25 @@ func (o *mtoShipmentStatusUpdater) createShipmentServiceItems(appCtx appcontext.
...
+		(shipment.PrimeEstimatedWeight != nil || shipment.NTSRecordedWeight != nil) {
+
+		var pickupLocation *models.Address
+		var deliveryLocation *models.Address
+		weight := shipment.PrimeEstimatedWeight
+
+		switch shipment.ShipmentType {
+		case models.MTOShipmentTypeHHGIntoNTSDom:
+			pickupLocation = shipment.PickupAddress
+			deliveryLocation = &shipment.StorageFacility.Address
```
reggieriser, line 726: Same question here as above -- do we need to protect against a nil `shipment.StorageFacility`?  I'm also wondering if we're sure these associations are being brought in.  When I look at the usage of this function, I see the caller is doing a `fetchShipment` but that doesn't appear to be bringing in any associations, but maybe I'm missing something in the QueryBuilder since I haven't used that a ton.

pkg/services/mto_shipment/shipment_approver.go
```diff
@@ -85,7 +85,14 @@ func (f *shipmentApprover) ApproveShipment(appCtx appcontext.AppContext, shipmen
 
 func (f *shipmentApprover) findShipment(appCtx appcontext.AppContext, shipmentID uuid.UUID) (*models.MTOShipment, error) {
 	var shipment models.MTOShipment
-	err := appCtx.DB().Q().Eager("MoveTaskOrder", "PickupAddress", "DestinationAddress").Find(&shipment, shipmentID)
+	err := appCtx.DB().Q().Eager("MoveTaskOrder", "PickupAddress", "DestinationAddress", "StorageFacility", "PrimeEstimatedWeight", "NTSRecordedWeight").Find(&shipment, shipmentID)
```
reggieriser, line 88: No need for `PrimeEstimatedWeight` or `NTSRecordedWeight` here since those aren't associations.  Those will be loaded with the shipment regardless.

pkg/services/mto_shipment/mto_shipment_updater.go
```diff
@@ -714,8 +714,25 @@ func (o *mtoShipmentStatusUpdater) createShipmentServiceItems(appCtx appcontext.
...
+		case models.MTOShipmentTypeHHGOutOfNTSDom:
+			pickupLocation = &shipment.StorageFacility.Address
+			deliveryLocation = shipment.DestinationAddress
+			weight = shipment.NTSRecordedWeight
+		default:
+			pickupLocation = shipment.PickupAddress
+			deliveryLocation = shipment.DestinationAddress
+		}
+		requiredDeliveryDate, calcErr := CalculateRequiredDeliveryDate(appCtx, o.planner, *pickupLocation, *deliveryLocation, *shipment.ScheduledPickupDate, weight.Int())
```
reggieriser, line 735: Not seeing any test changes for this file.  Should we exercise this switching logic?

pkg/services/mto_shipment/shipment_approver.go
```diff
@@ -102,8 +109,27 @@ func (f *shipmentApprover) findShipment(appCtx appcontext.AppContext, shipmentID
...
+		(shipment.PrimeEstimatedWeight != nil || shipment.NTSRecordedWeight != nil) {
+
+		var pickupLocation *models.Address
+		var deliveryLocation *models.Address
+		var weight int
+
+		switch shipment.ShipmentType {
+		case models.MTOShipmentTypeHHGIntoNTSDom:
+			pickupLocation = shipment.PickupAddress
+			deliveryLocation = &shipment.StorageFacility.Address
```
reggieriser, line 121: Same possible nil `shipment.StorageFacility`.

pkg/services/mto_shipment/shipment_approver.go
```diff
@@ -102,8 +109,27 @@ func (f *shipmentApprover) findShipment(appCtx appcontext.AppContext, shipmentID
...
+			weight = shipment.PrimeEstimatedWeight.Int()
+		case models.MTOShipmentTypeHHGOutOfNTSDom:
+			pickupLocation = &shipment.StorageFacility.Address
+			deliveryLocation = shipment.DestinationAddress
+			weight = shipment.NTSRecordedWeight.Int()
```
reggieriser, line 126: This could panic too if we had a nil `NTSRecordedWeight`.  I know you check that one of the two weights above is non-nil, but not that the specific shipment type has the right weight field set.  Hopefully unlikely, though.

pkg/services/mto_shipment/shipment_approver_test.go
```diff
@@ -330,4 +330,128 @@ func (suite *MTOShipmentServiceSuite) TestApproveShipment() {
...
+				StorageFacilityID:    &storageFacility.ID,
+				DestinationAddress:   &destinationAddress,
+				DestinationAddressID: &destinationAddress.ID,
+			},
+		})
+
+		testCases := []struct {
+			shipment models.MTOShipment
+		}{
+			{hhgShipment},
```
reggieriser, line 434: Nit: this could be simplified to be a slice of shipments rather than a list of structs that contain a shipment.

pkg/services/mto_shipment/shipment_approver_test.go
```diff
@@ -330,4 +330,128 @@ func (suite *MTOShipmentServiceSuite) TestApproveShipment() {
...
+
+			shipmentEtag := etag.GenerateEtag(testCase.shipment.UpdatedAt)
+			_, err = approver.ApproveShipment(appCtx, testCase.shipment.ID, shipmentEtag)
+			suite.NoError(err)
+
+			fetchedShipment := models.MTOShipment{}
+			err = suite.DB().Find(&fetchedShipment, testCase.shipment.ID)
+			suite.NoError(err)
+			// We also should have a required delivery date
+			suite.NotNil(fetchedShipment.RequiredDeliveryDate)
```
reggieriser, line 454: Is the presence of a `RequiredDeliveryDate` enough to tell us if the switching logic is working correctly?  Seems like we'd need to dig a little deeper to know that we got the right date for the right address/weight combinations.

reggieriser (commented):
Looks like this is moving in the right direction, but I have a few notes for consideration regarding eager fetching, nil checks, and test coverage.

pkg/handlers/primeapi/mto_shipment.go
```diff
@@ -210,14 +212,26 @@ func (h UpdateMTOShipmentHandler) checkPrimeValidationsOnModel(appCtx appcontext
 
 	// Prime can create a new address, but cannot update it.
 	// So if address exists, return an error. But also set the local pointer to nil, so we don't recalculate requiredDeliveryDate
+	var latestPickupAddress *models.Address
+	var latestDestinationAddress *models.Address
+
+	switch dbShipment.ShipmentType {
+	case models.MTOShipmentTypeHHGIntoNTSDom:
+		latestPickupAddress = dbShipment.PickupAddress
+		latestDestinationAddress = &dbShipment.StorageFacility.Address
```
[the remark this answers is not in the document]
reggieriser: I don't think you have to worry about a panic on the pickup or destination addresses, at least not here.  They could in theory be `nil` since they're pointers, but you're not looking at some field on them like you're doing with the potentially-nil storage facility.

pkg/handlers/primeapi/mto_shipment_test.go
```diff
@@ -1023,51 +1023,99 @@ func (suite *HandlerSuite) TestUpdateMTOShipmentDateLogic() {
 	})
 
 	suite.T().Run("PATCH Success 200 RequiredDeliveryDate updated on scheduledPickupDate update", func(t *testing.T) {
-		tenDaysFromNow := now.AddDate(0, 0, 11)
-		oldShipment := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+		address := testdatagen.MakeAddress(suite.DB(), testdatagen.Assertions{})
+		storageFacility := testdatagen.MakeStorageFacility(suite.DB(), testdatagen.Assertions{})
```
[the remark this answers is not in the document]
mlm55: I think this should be ok. Even though the addresses are all the same, the function should check different addresses based on shipment type. Eg, the nts shipment in this test has a pickup address and a storage facility while the destination address is nil. If it was pulling the wrong addresses, i'd expect a panic to bubble up

pkg/services/mto_shipment/mto_shipment_updater.go
reggieriser: Thanks for adding the `nil` check.  But I'm still concerned that the associations like `PickupAddress` and `StorageFacility.Address` may not be loaded when the passed-in shipment is fetched.  Can you confirm if the associations are there?

reggieriser (commented):
I tried this out using the instructions and it worked as expected.  Thanks for making the adjustments!  I did add a note about verifying that some associations are being loaded.  If we can confirm (or fix) that, I should be ready to approve.

pkg/services/mto_shipment/mto_shipment_updater.go
mlm55: confirming that they're loaded. 
Just after the fetching the shipment Screen Shot 2022-02-07 at 11 03 02 AM

In the setRequiredDeliveryDate function
Screen Shot 2022-02-07 at 11 05 11 AM

reggieriser: Just to follow-up on this.  Marjo and I paired and we found that the `fetchShipment` (which uses the QueryBuilder) is actually doing an unbounded `Eager` which loads all first-level associations.  So it was loading `StorageFacility`.  However, we found that it was not loading `StorageFacility.Address` since an unbounded `Eager` doesn't go down the tree by default.  So, Marjo added additional code to load `StorageFacility.Address` in the latest commit.

reggieriser (approved):
LGTM! 🚢

PR #8100, opened 2022-01-28
MB-10670 ppm db migration model

NamibiaTorres:
## Summary
 DB migration for new PPM table called ppm_shipments and create new model for this new table.

## Setup to Run Your Code

You won't need to run the code for this PR, but you can check to make sure that the models set up makes sense. Note that validations will be added to the Service object for PPMs. 

Additionally, please take at the migration. One question we had was whether we need a specific character length for varchar?

### Backend


pkg/models/ppm_shipment.go
```diff
@@ -0,0 +1,61 @@
...
+	PPMShipmentStatusSUBMITTED PPMShipmentStatus = "SUBMITTED"
+	// PPMShipmentStatusAPPROVED captures enum value "APPROVED"
+	PPMShipmentStatusAPPROVED PPMShipmentStatus = "APPROVED"
+	// PPMShipmentStatusPAYMENTREQUESTED captures enum value "PAYMENT_REQUESTED"
+	PPMShipmentStatusPAYMENTREQUESTED PPMShipmentStatus = "PAYMENT_REQUESTED"
+	// PPMShipmentStatusCOMPLETED captures enum value "COMPLETED"
+	PPMShipmentStatusCOMPLETED PPMShipmentStatus = "COMPLETED"
+	// PPMShipmentStatusCANCELED captures enum value "CANCELED"
+	PPMShipmentStatusCANCELED PPMShipmentStatus = "CANCELED"
+)
```
duncan-truss, line 27: Have you all given any thought as to if we should combine this with the parent shipment status?

duncan-truss:
@NamibiaTorres To your question about if `varchar` needs a length, I think for Postgres there isn't a real benefit in terms of performance whether we introduce a char(5), varchar(5), varchar, or text since they used the same underlying data type.  If we were using another database or worried about SQL standard compatibility we may make a different choice.

Looking elsewhere we use all of the above for postal code/zip3/zip5 columns so I don't see a strong suggestion for consistency.  The question is do we want an error if someone tries to save a zip of length of say 10 (XXXXX-XXXX) instead of 5?  Or if our requirements change to accept longer ZIPs how painful will it be to adjust the datatype?  Varchar(n) only errors if the length is over so if we wanted to make sure our data is length 5 and only length 5 then adding a [constraint check](https://www.postgresql.org/docs/14/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS) may be easier than hardcoding the data type.

migrations/app/schema/20220127234422_create_ppm_shipment_table.up.sql
```diff
@@ -0,0 +1,66 @@
...
+		CONSTRAINT ppm_shipment_mto_shipment_id_fkey
+		REFERENCES mto_shipments,
+	created_at timestamp NOT NULL,
+	updated_at timestamp NOT NULL,
+	status ppm_shipment_status NOT NULL,
+	expected_departure_date timestamp,
+	actual_move_date timestamp,
+	submit_date timestamp,
+	review_date timestamp,
+	approve_date timestamp,
```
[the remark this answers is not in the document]
duncan-truss: I think we also recommend using `timestamp with time zone` except for the Pop controlled `created_at` and `updated_at` fields.  While all timestamps like this generated by the backend system are saved as UTC, it's still good to be consistent.

This recommendation talks more about it, for the case where you want to capture different external user's timezone to convert back to https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp_.28without_time_zone.29

ronaktruss:
+1 to Duncan's comments (see thumbs up). Outside of those the rest of the PR looks good.

pkg/models/ppm_shipment.go
NamibiaTorres: Looks like we're keeping it here for now, and then removing once we get consensus on whether to just use the shipment statuses.

NamibiaTorres:
> @NamibiaTorres To your question about if `varchar` needs a length, I think for Postgres there isn't a real benefit in terms of performance whether we introduce a char(5), varchar(5), varchar, or text since they use the same underlying data type. If we were using another database or worried about SQL standard compatibility we may make a different choice.
> 
> Looking elsewhere we use all of the above for postal code/zip3/zip5 columns so I don't see a strong suggestion for consistency. The question is do we want an error if someone tries to save a zip of length of say 10 (XXXXX-XXXX) instead of 5? Or if our requirements change to accept longer ZIPs how painful will it be to adjust the datatype? Varchar(n) only errors if the length is over so if we wanted to make sure our data is length 5 and only length 5 then adding a [constraint check](https://www.postgresql.org/docs/14/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS) may be easier than hardcoding the data type.

Thank you! This was an informative explanation. I have a feeling allowing for flexibility will help once we start accepting international zips as well.

migrations/app/schema/20220127234422_create_ppm_shipment_table.up.sql
```diff
@@ -0,0 +1,66 @@
...
+(
+	id uuid PRIMARY KEY NOT NULL,
+	shipment_id    uuid NOT NULL
+		CONSTRAINT ppm_shipment_mto_shipment_id_fkey
+		REFERENCES mto_shipments,
+	created_at timestamp NOT NULL,
+	updated_at timestamp NOT NULL,
+	status ppm_shipment_status NOT NULL,
+	expected_departure_date timestamp,
+	actual_move_date timestamp,
```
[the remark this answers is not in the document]
NamibiaTorres: @duncan-truss , do we also need to update the model here? Looking at the mto_shipment_model, for date fields, it looks like we use time.Time:
```
	ScheduledPickupDate              *time.Time              `db:"scheduled_pickup_date"`
	RequestedPickupDate              *time.Time              `db:"requested_pickup_date"`
	RequestedDeliveryDate            *time.Time              `db:"requested_delivery_date"`

```

I changed the db migration to `date` and would like to understand the model convention a little better too.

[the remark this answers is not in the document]
duncan-truss: That's a good question, Go doesn't have separate data types for dates and date + time values the way other languages separate them so you shouldn't have to change anything about the model.  Pop or Postgres will disregard/truncate if there is a time data included when you got to save.  In swagger and our api handlers we'll have to remember to use dates instead of datetimes as well.

migrations/app/migrations_manifest.txt
```diff
@@ -696,3 +696,4 @@
 20220107193037_add_destination_address_type_to_shipments.up.sql
 20220112013152_update_duty_locations.up.sql
 20220113172826_create_new_duty_locations.up.sql
+20220127234422_create_ppm_shipment_table.up.sql
```
duncan-truss, line 699: One last request to pluralize the name `ppm_shipments`

migrations/app/schema/20220127234422_create_ppm_shipment_table.up.sql
```diff
@@ -0,0 +1,66 @@
...
+CREATE TYPE ppm_shipment_status AS enum (
+	'DRAFT',
+	'SUBMITTED',
+	'APPROVED',
+    'PAYMENT_REQUESTED',
+    'COMPLETED',
+    'CANCELED'
+	);
+
+CREATE TABLE ppm_shipment
```
duncan-truss, line 11: Last request to pluralize the name `ppm_shipments`

migrations/app/schema/20220127234422_create_ppm_shipment_table.up.sql
```diff
@@ -0,0 +1,66 @@
...
+	advance_requested bool,
+	advance_id uuid
+		CONSTRAINT ppm_shipment_reimbursements_id_fkey
+		REFERENCES reimbursements,
+	advance_worksheet_id uuid
+		CONSTRAINT ppm_shipment_documents_id_fkey
+		REFERENCES documents
+);
+
+COMMENT on TABLE ppm_shipment IS 'Stores all PPM shipments, and their details.';
```
duncan-truss, line 45: Sorry the plural name will have to be replaced in the comments here too.

felipe-lee:
@duncan-truss  on the pluralizing, do we have that documented somewhere? I am seeing now that you mention it that all our tables have plural names, but the corresponding models are singular, which I personally find a bit weird (I'm of the camp that thinks both should be singular, but I know there's a lot of debate about this). At this point it's our pattern so I agree we should update this migration to follow it, but I'm curious if there's a reason documented, and if we should update the docs on adding a table to say that (and the reason if we have it somewhere) https://transcom.github.io/mymove-docs/docs/backend/setup/database-migrations#adding-a-new-table

felipe-lee:
@duncan-truss, @NamibiaTorres, and @ronaktruss, I realized as I was starting to add the testdatagen function that we didn't add PPM as a type of shipment so I'm going to push up a commit adding that. I have it as a separate migration, but am also fine if we want to combine the migrations since they really only make sense together anyways.

felipe-lee:
Another thing is that we didn't fully decide if the PPM shipment should keep track of its own status separately from the regular MTOShipment table or not. For now, I updated the statuses on the PPM table to match the flow on https://miro.com/app/board/o9J_lqJFWxk=/?moveToWidget=3458764515331487801&cot=14

felipe-lee:
> @duncan-truss, @NamibiaTorres, and @ronaktruss, I realized as I was starting to add the testdatagen function that we didn't add PPM as a type of shipment so I'm going to push up a commit adding that. I have it as a separate migration, but am also fine if we want to combine the migrations since they really only make sense together anyways.

Hmm, I followed the pattern that was used in https://github.com/transcom/mymove/blob/a0eb0fb0d58f06493f26bff553d78fff5fa1aa86/migrations/app/schema/20200205160112_add_enum_values_to_shipment_type_within_mto_shipments.up.sql for the migration to add PPM as a shipment type, but I was looking at the docs for postgres and see that it's possible to do something like

```sql
ALTER TYPE mto_shipment_type ADD VALUE 'PPM';
```

That seems much simpler. @duncan-truss, do you happen to know if one is preferred over the other? I found this: https://stackoverflow.com/questions/56942421/how-can-i-add-a-new-value-to-an-enum-in-postgres-without-locking-the-table

which seems to indicate that the `ALTER` method may be better since it doesn't lock things.

duncan-truss:
> > @duncan-truss, @NamibiaTorres, and @ronaktruss, I realized as I was starting to add the testdatagen function that we didn't add PPM as a type of shipment so I'm going to push up a commit adding that. I have it as a separate migration, but am also fine if we want to combine the migrations since they really only make sense together anyways.
> 
> Hmm, I followed the pattern that was used in https://github.com/transcom/mymove/blob/a0eb0fb0d58f06493f26bff553d78fff5fa1aa86/migrations/app/schema/20200205160112_add_enum_values_to_shipment_type_within_mto_shipments.up.sql for the migration to add PPM as a shipment type, but I was looking at the docs for postgres and see that it's possible to do something like
> 
> ```sql
> ALTER TYPE mto_shipment_type ADD VALUE 'PPM';
> ```
> 
> That seems much simpler. @duncan-truss, do you happen to know if one is preferred over the other? I found this: https://stackoverflow.com/questions/56942421/how-can-i-add-a-new-value-to-an-enum-in-postgres-without-locking-the-table
> 
> which seems to indicate that the `ALTER` method may be better since it doesn't lock things.

I don't know if I've ever had to look into that.  It sounds like we could do either way honestly.  The only note I see is about slightly better performance when the type is recreated than when it is appended to after the fact.  Although I think it helps that we're just appending it to the end and not inserting in at another position.

Screen Shot 2022-02-03 at 9 08 15 AM

duncan-truss:
> @duncan-truss on the pluralizing, do we have that documented somewhere? I am seeing now that you mention it that all our tables have plural names, but the corresponding models are singular, which I personally find a bit weird (I'm of the camp that thinks both should be singular, but I know there's a lot of debate about this). At this point it's our pattern so I agree we should update this migration to follow it, but I'm curious if there's a reason documented, and if we should update the docs on adding a table to say that (and the reason if we have it somewhere) https://transcom.github.io/mymove-docs/docs/backend/setup/database-migrations#adding-a-new-table

I don't know if it is documented anywhere, I take that convention for granted but keep in mind that Pop takes after Rail's Active Record database ORM.  When Milmove started we used the auto generators for migrations I think or at least used the Fizz language and I think it follows the singular model and plural database table defaults https://gobuffalo.io/en/docs/db/models/#using-the-generator

duncan-truss (approved):
Nice work this looks likes it's in a good place! The only other thing I could think of is adding a field to the MTOShipment model to hold the PPMShipment when we join/eager load.  It's probably okay if that is on the testdatagen PR.

NamibiaTorres:
> Nice work this looks likes it's in a good place! I tested on a fresh database and then on an existing one with data.
> 
> The only other thing I could think of is adding a field to the MTOShipment model to hold the PPMShipment when we join/eager load. It's probably okay if that is on the testdatagen PR.

Thanks Duncan! I'm going to merge this PR, and ask the testdatagen folks to add that field. I really appreciate all the great feedback you provided on this PR!

felipe-lee:
> > @duncan-truss on the pluralizing, do we have that documented somewhere? I am seeing now that you mention it that all our tables have plural names, but the corresponding models are singular, which I personally find a bit weird (I'm of the camp that thinks both should be singular, but I know there's a lot of debate about this). At this point it's our pattern so I agree we should update this migration to follow it, but I'm curious if there's a reason documented, and if we should update the docs on adding a table to say that (and the reason if we have it somewhere) https://transcom.github.io/mymove-docs/docs/backend/setup/database-migrations#adding-a-new-table
> 
> I don't know if it is documented anywhere, I take that convention for granted but keep in mind that Pop takes after Rails' Active Record database ORM. When Milmove started we used the auto generators for migrations I think or at least used the Fizz language and it follows the singular model and plural database table defaults https://gobuffalo.io/en/docs/db/models/#using-the-generator
> 
> https://guides.rubyonrails.org/active_record_basics.html#naming-conventions

Ah ok, cool, good to know about the auto-generated migrations following that pattern, I knew we had used fizz files before but didn't know we'd used just the auto-generated ones.

PR #8115, opened 2022-02-02
Audit Table Draft - Testing Triggers

kennethchow-truss:
In this PR I am testing triggers on DB to save diffs

Also will include querying the data


migrations/app/schema/20220119211559_create_log_table.up.sql
```diff
@@ -0,0 +1,114 @@
+CREATE TABLE log_event_types (
```
kennethchow-truss, line 1: this is the only file that you should pay attention to

PR #8121, opened 2022-02-03
[MB-11261]  Archive ppm data

pearl-truss:
## Summary

This PR archives the old PPM tables by creating new tables prefixed with `archived` and copy the data from the corresponding tables into them. 

## Setup to Run Your Code

Populate the database by running `make db_dev_e2e_populate`
Run your server `make server_run`

Open a database view and directly insert the sql query from the migration and verify that the new created tables have copied the existing data into them

Alternatively, you can run `generate-test-data` and then run the migrations to review these changes

- [x] Follows our guidelines for [Zero-Downtime Deploys](https://transcom.github.io/mymove-docs/docs/dev/contributing/database/Database-Migrations#zero-downtime-migrations)
- [x] Have been communicated to #g-database
- [x] Secure migrations have been tested following the instructions in our [docs](https://transcom.github.io/mymove-docs/docs/dev/contributing/database/Database-Migrations#secure-migrations)


migrations/app/schema/20220201155623_archive_ppm_data.up.sql
```diff
@@ -0,0 +1,58 @@
...
+    INCLUDING DEFAULTS INCLUDING CONSTRAINTS INCLUDING INDEXES
+);
+
+ALTER TABLE archived_personally_procured_moves
+	ADD CONSTRAINT archived_personally_procured_moves_move_id_fkey FOREIGN KEY (move_id) REFERENCES moves,
+	ADD CONSTRAINT archived_personally_procured_moves_advance_worksheet_id FOREIGN KEY (advance_worksheet_id) REFERENCES documents;
+
+INSERT INTO archived_personally_procured_moves SELECT * FROM personally_procured_moves;
+
+CREATE TABLE archived_move_documents(
```
duncan-truss, line 12: Related to the existing `move_documents` table, I see that we have a constraint on the document_id to cascade and delete if the move_documents row is removed.  This will be a problem if this archived_move_documents document_id is pointing to the same record in the documents table.  We should remove that ON DELETE CASCADE from `move_documents` is my read on it.

NamibiaTorres:
Additions and migration lgtm. Once Duncan's comment is addressed, I can take another look.

pkg/testdatagen/scenario/e2ebasic.go
```diff
@@ -1208,6 +1208,38 @@ func serviceMemberWithPPMReadyToRequestPayment01(appCtx appcontext.AppContext, u
 		},
 		UserUploader: userUploader,
 	})
+	moveDoc := testdatagen.MakeMoveDocument(appCtx.DB(), testdatagen.Assertions{
```
duncan-truss, line 1211: Thanks for adding this test data to verify.  This file is for the e2e_basic scenario though that Cypress uses, while the dev_seed scenario is what `make db_dev_e2e_populate` uses.  I think it's better suited in the dev seed rather than the Cypress e2e basic seed or am I missing something?

`make db_dev_e2e_populate` will also apply missing migrations before you load data so it might be a better test instruction to run just the generate-test-data command itself before running the migration that copies the data.

pkg/testdatagen/scenario/shared.go
```diff
@@ -221,18 +221,39 @@ func createPPMWithAdvance(appCtx appcontext.AppContext, userUploader *uploader.U
...
+		MovingExpenseDocument: models.MovingExpenseDocument{
+			MoveDocument:   moveDoc,
+			MoveDocumentID: moveDoc.ID,
+		},
+	})
+	testdatagen.MakeWeightTicketSetDocument(db, testdatagen.Assertions{
+		WeightTicketSetDocument: models.WeightTicketSetDocument{
+			MoveDocument: moveDoc,
+		},
+	})
```
duncan-truss, line 255: I'm not seeing either of these tables populated are you?

duncan-truss: Whoops this was my mistake, I ran the compiled generate-test-data instead of picking up the new changes with go run ..cmd/generate-test-data ...

migrations/app/migrations_manifest.txt
```diff
@@ -698,4 +698,5 @@
 20220113172826_create_new_duty_locations.up.sql
 20220126235958_remove_cac_access_carterjones_monfresh_ronolibert_and_shkeating.up.sql
 20220127234422_create_ppm_shipments_table.up.sql
+20220201155623_archive_ppm_data.up.sql
```
duncan-truss, line 701: Do you mind incrementing this time to be after the migration that was merged before and change the file name as well?

duncan-truss (approved):
LGTM! Just a comment about reordering the manifest so that this comes after the ppm_shipments table that was merged first.

migrations/app/migrations_manifest.txt
duncan-truss: Although I don't know how that might affect experimental, which has already been applied there.

reggieriser: I believe if a migration is rolled out to experimental but you later change it (either the name or the contents) before merging, then you'll need to get infra to restore the experimental database to a state before the migration was applied.  It's possible the next deploy to experimental will fail.  If the name is changed, the next experimental deploy will try to run the migration a second time since it doesn't realize that migration has already been applied.  If the contents are changed but the name stays the same, the migration won't be run again at all and the changes won't be applied to experimental.

pearl-truss: @transcom/truss-infrasec regarding this can someone restore the experimental database?

gidjin (approved):

PR #8128, opened 2022-02-04
MB-11455 : start of swagger fixes

ahobson:
## Summary

Fix some discrepancies in our swagger files and what our server generates. There's a lot more where this came from, but this is a start.

## Setup to Run Your Code

Checkout the branch, build, and try some workflows

Hopefully our e2e tests would find any problems. My manual spot checking hasn't turned anything up


YanZ777 (approved):
Looks good to me.

pkg/handlers/primeapi/move_task_order_test.go
```diff
@@ -61,6 +61,8 @@ func (suite *HandlerSuite) TestListMovesHandlerReturnsUpdated() {
 	listMovesResponse := response.(*movetaskorderops.ListMovesOK)
 	movesList := listMovesResponse.Payload
 
+	suite.NoError(movesList.Validate(strfmt.Default))
```
rogeruiz, line 64: This is an important thing to do in order to ensure that we're validating our Swagger definitions in our tests. It could be easy to do this in your tests; and if it is please do if the work takes under an hour to complete. This is an example only and it's possible there will be a larger amount of work for certain tests and handlers, so make sure you communicate that with your team in order to track that larger effort (>1 hours) work.

PR #7890, opened 2021-12-09
MB-10592 Use custom nullable types for patch requests ADR

eamahanna:
## Summary

This ADR discusses how to handle setting fields to null via update requests. Link at at the bottom of the adr for additional information.

A prototype PR is here: https://github.com/transcom/mymove/pull/7881

PR #8129, opened 2022-02-04
[MB-10661] create nullable types for PATCH requests

mboren:
## Summary
The current way we decode JSON makes it impossible to tell whether a given field was explicitly set to `null` in a request body, or if it was omitted.
We would like to have this behavior:
- if the field is present and `null`, update the field to be null
- if the field is omitted, do not update it

To get this behavior, we're implementing types to wrap values of all the primitive types
they all look kinda like this:
```go
type Bool struct {
	Present bool // Present is true if key is present in json
	Value   *bool
}
```
`String` was added in a previous PR, and I'm basing all these new types on that one.

For additional context review the following items:

- #7890 
- conversation about some alternatives: https://ustcdp3.slack.com/archives/CP6PTUPQF/p1638833895016700
- Library this is based on: https://github.com/romanyx/nullable
- Blog post about it: https://romanyx90.medium.com/handling-json-null-or-missing-values-with-go-swagger-4d7f37a2a7ca

### things to give additional scrutiny
The tests aren't set up exactly the same as most of our other tests because I was adapting them from the library this is based off of. They seem to run fine with our test runner though and don't really need access to any of our context stuff. If you think it's worthwhile, I'm definitely down to refactor them to use `testify` suites.

## Setup to Run Your Code
None of the types I'm adding are being used yet, so there's not really anything to test. 


Ryan-Koch (approved):
This is looking good to me. Since it doesn't need anything from context I don't think we need to worry about refactoring the tests here.

leia-sefkin (approved):
LGTM, nice catch

PR #8135, opened 2022-02-07
MB-11460: Support local swagger binary

ahobson:
## Summary

Support local swagger binary, which helps immensely on M1 Macs

## Setup to Run Your Code

```sh
make server_generate
```


scripts/gen-server
```diff
@@ -2,10 +2,15 @@
 
 set -eu -o pipefail
 
+if [ "${CIRCLECI+x}" ]; then
+  LOCAL_SWAGGER="/usr/local/bin/swagger"
+elif swagger version 2>/dev/null | grep -F -q 'version: 0.28.0'; then
+  LOCAL_SWAGGER=$(which swagger)
+fi
+
```
rogeruiz, line 10: nice work on cleaning up the Swagger location here.

scripts/pre-commit-swagger-validate
```diff
@@ -6,15 +6,22 @@
 
 set -eu -o pipefail
 
-filename="$1"
-readonly filename
-
 UNAME_S=$(uname -s)
 if [ "$UNAME_S" == "Linux" ]; then
```
rogeruiz, line 10: lol, this is a funny way to check if we're in CircleCI or not. I don't think we need to fix it now. but I think it should be checking for the `$CIRCLECI` variable here in the future?

rogeruiz (approved):
lgtm 🌈

thanks for doing this @ahobson. I don't have an M1 Mac to test the speeds here. I'd love to see how long it takes to do on a newer machine.

<details>
<summary>📸  gif(s)</summary>


</details>

PR #8142, opened 2022-02-08
Mb 11028 backend dates location nt

NamibiaTorres:
## Summary
Add a ppmShipmentCreator Service Object following the new validation pattern for service objects.

[this article](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/golang-guide/#service-objects) explains more about the approach used.

Create a migration to add DRAFT status

## Setup to Run Your Code

<details>
<summary>💻 You will need to make sure that all tests pass. Note that the validation tests are commented out because they are set up for the ppmShipmentUpdater, when that service object is added in the next sprint</summary>

##### Terminal 1

Run tests:

```sh
go test ./pkg/services/ppmshipment --testify.m TestValidationRules -v
```

```sh
go test ./pkg/services/ppmshipment --testify.m TestPPMShipmentCreator -v
```
</details>

### Backend


pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -132,7 +135,7 @@ func (f mtoShipmentCreator) CreateMTOShipment(appCtx appcontext.AppContext, ship
 				return fmt.Errorf("failed to create pickup address %#v %e", verrs, err)
 			}
 			shipment.PickupAddressID = &shipment.PickupAddress.ID
-		} else if shipment.ShipmentType != models.MTOShipmentTypeHHGOutOfNTSDom {
+		} else if shipment.ShipmentType != models.MTOShipmentTypeHHGOutOfNTSDom && shipment.ShipmentType != models.MTOShipmentTypePPM {
 			return apperror.NewInvalidInputError(uuid.Nil, nil, nil, "PickupAddress is required to create an HHG or NTS type MTO shipment")
```
pearl-truss, line 139: I think pickup address isn't required for PPM creation but we do need pickup postal code. If that's true, is it in scope for this ticket to add that validation?

pkg/services/ppmshipment/ppm_shipment_creator.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (f *ppmShipmentCreator) createPPMShipment(appCtx appcontext.AppContext, ppmShipment *models.PPMShipment, checks ...ppmShipmentValidator) (*models.PPMShipment, error) {
+	// Make sure we do this whole process in a transaction so partial changes do not get made committed
+	// in the event of an error.
+
+	builder := query.NewQueryBuilder()
+	fetcher := fetch.NewFetcher(builder)
+	moveRouter := moverouter.NewMoveRouter()
+	mtoShipmentCreator := mtoshipment.NewMTOShipmentCreator(builder, fetcher, moveRouter)
```
duncan-truss, line 53: We should be able to move this `mtoShipmentCreator` out to become a dependency that is passed to the NewPPMShipmentCreator(mtoShipmentCreator) constructor and set as a struct field on the `ppmShipmentCreator`.  This way we only create one instance of it in say the `pkg/handlers/internal/api.go` file when we initialize the handlers instead of creating a new one for every call to createPPMShipment() causing more memory to be used.

pkg/services/ppmshipment/ppm_shipment_creator.go
```diff
@@ -0,0 +1,86 @@
...
+		if ppmShipment.Shipment.ShipmentType == "" {
+			ppmShipment.Shipment.ShipmentType = models.MTOShipmentTypePPM
+		} else if ppmShipment.Shipment.ShipmentType != models.MTOShipmentTypePPM {
+			return apperror.NewInvalidInputError(uuid.Nil, nil, nil, "MTO shipment type must be PPM shipment")
+		}
+
+		if ppmShipment.Shipment.Status == "" {
+			ppmShipment.Shipment.Status = models.MTOShipmentStatusDraft
+		} else if ppmShipment.Shipment.Status != models.MTOShipmentStatusDraft {
+			return apperror.NewInvalidInputError(uuid.Nil, nil, nil, "Must have a DRAFT status")
```
duncan-truss, line 48: Can we add specifiers to these error messages to say if we're talking about the MTOShipment status or the PPM shipment status?

pkg/services/ppmshipment/ppm_shipment_creator.go
```diff
@@ -0,0 +1,86 @@
...
+			ppmShipment.Status = models.PPMShipmentStatusDraft
+		} else if ppmShipment.Status != models.PPMShipmentStatusDraft {
+			return apperror.NewInvalidInputError(uuid.Nil, nil, nil, "Must have a DRAFT status")
+		}
+
+		// NOTE: The ppm may require a service item for pricing.Passing an HHG service item may be sufficient for the pricer.
+		createShipment, err := f.mtoShipmentCreator.CreateMTOShipment(txnAppCtx, &ppmShipment.Shipment, nil)
+		// Check that mtoshipment is created. If not, bail out.
+		if err != nil {
+			return apperror.NewQueryError("MTOShipment", err, "")
```
duncan-truss, line 61: I don't think we know that this was a QueryError, that to me would mean the database save failed.  I think there are a lot of different error types that can result from CreateMTOShipment so we should just return the err instead of wrapping it here.

pkg/services/ppmshipment/ppm_shipment_creator.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		ppmShipment.ShipmentID = createShipment.ID
+		// Validate the ppmShipment, and return an error
+		err = validatePPMShipment(txnAppCtx, *ppmShipment, nil, &ppmShipment.Shipment)
```
duncan-truss, line 66: I think you need to set `ppmShipment.Shipment = createShipment` similar to how you updated the ShipmentID above before passing it to validatePPMShipment.  It won't get the newly saved MTOShipment unless you requeried the database for it, it only has the in-memory copy that was initially passed in unless you manually update it.

pkg/services/ppmshipment/rules.go
```diff
@@ -0,0 +1,75 @@
...
+func checkPPMShipmentID() ppmShipmentValidator {
+	return ppmShipmentValidatorFunc(func(_ appcontext.AppContext, newPPMShipment models.PPMShipment, oldPPMShipment *models.PPMShipment, _ *models.MTOShipment) error {
+		verrs := validate.NewErrors()
+		if oldPPMShipment == nil {
+			if newPPMShipment.ID != uuid.Nil {
+				verrs.Add("ID", "cannot manually set a new PPM Shipment's UUID")
+			}
+		} else {
+			if newPPMShipment.ID != oldPPMShipment.ID {
+				return apperror.NewImplementationError(
```
duncan-truss, line 42: Did you want this to be an ImplementationError? It seems like everywhere else is using verrs.

pkg/services/ppmshipment/rules.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		// Override pre-existing values with anything sent in for the update/create:
+		if !newPPMShipment.CreatedAt.IsZero() {
+			createdAt = newPPMShipment.CreatedAt
+		}
+
+		// Check that we have something in the CreatedAt field:
+		if createdAt.IsZero() {
+			verrs.Add("requestedAt", "cannot be empty")
```
duncan-truss, line 70: I think you meant `createdAt` here.  Pop adds the CreatedAt and UpdatedAt field values when we call Save().  Is there a reason we need to ever check either of these fields if Pop maintains them regardless of if we set their value to something else?

swagger/internal.yaml
```diff
@@ -2315,10 +2315,12 @@ definitions:
       - HHG_SHORTHAUL_DOMESTIC
       - HHG_INTO_NTS_DOMESTIC
       - HHG_OUTOF_NTS_DOMESTIC
+      - PPM
```
duncan-truss, line 2318: You should be changing the swagger-def/internal.yaml file, this one gets generated from that one.  We should see `pkg/gen` updates to this definition if you updated swagger-def/internal.yaml and ran `make server_generate`.

pkg/services/ppmshipment/rules_test.go
```diff
@@ -0,0 +1,152 @@
...
+						suite.Contains(verr.Keys(), "ID")
+					default:
+						suite.False(testCase.verr, "expected a *validate.Errors: %t - naid %s", err, testCase.newPPMShipment.ID)
+					}
+				})
+			}
+
+		})
+	})
+	suite.Run("CheckRequiredFields()", func() {
```
duncan-truss, line 129: I think this is the only rule without a failure test, the others look like they have both success and failure.

pkg/services/ppmshipment/ppm_shipment_creator_test.go
```diff
@@ -0,0 +1,65 @@
...
+	suite.T().Run("CreatePPMShipment - Success", func(t *testing.T) {
+		// Under test:	CreatePPMShipment
+		// Set up:		Established valid shipment and valid new PPM shipment
+		// Expected:	New PPM shipment successfully created
+		subtestData := suite.createSubtestData(testdatagen.Assertions{})
+		ppmShipmentCreator := NewPPMShipmentCreator(mtoShipmentCreator)
+		createdPPMShipment, err := ppmShipmentCreator.CreatePPMShipmentCheck(suite.AppContextForTest(), subtestData.newPPMShipment)
+
+		suite.Nil(err)
+		suite.NotNil(createdPPMShipment)
```
duncan-truss, line 52: I might want to check the default statuses are what you expect here and the shipment type gets set to PPM.

This test should fail the checkRequiredFields validation because none of those fields have values in your subtest data.  I think the issue is here where you need to pass `checks` to your validatePPMShipment function. https://github.com/transcom/mymove/pull/8142/files#diff-51bd11d6b8ccdd0a09290bf3a9944af32af2fece96be3282384c8ce94e869f8eR68

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -58,7 +59,8 @@ func (f mtoShipmentCreator) CreateMTOShipment(appCtx appcontext.AppContext, ship
 			return nil, apperror.NewInvalidInputError(uuid.Nil, nil, verrs,
 				fmt.Sprintf("RequestedPickupDate should not be set when creating a %s shipment", shipment.ShipmentType))
 		}
-	} else {
+	} else if shipment.ShipmentType != models.MTOShipmentTypePPM {
```
felipe-lee, line 62: Can you update the tests for this? This change probably leads to a change to this test: https://github.com/transcom/mymove/blob/a4630e461c3bdf43bd6b42cb32bdda7a9228bef6/pkg/services/mto_shipment/mto_shipment_creator_test.go#L71

pkg/services/mto_shipment/mto_shipment_creator.go
```diff
@@ -132,7 +134,7 @@ func (f mtoShipmentCreator) CreateMTOShipment(appCtx appcontext.AppContext, ship
 				return fmt.Errorf("failed to create pickup address %#v %e", verrs, err)
 			}
 			shipment.PickupAddressID = &shipment.PickupAddress.ID
-		} else if shipment.ShipmentType != models.MTOShipmentTypeHHGOutOfNTSDom {
+		} else if shipment.ShipmentType != models.MTOShipmentTypeHHGOutOfNTSDom && shipment.ShipmentType != models.MTOShipmentTypePPM {
```
felipe-lee, line 137: This could also use a test. I'm not sure I see one that explicitly covers this, but I do see a dedicated test for NTS shipments, so maybe one for PPM shipments would be good too: https://github.com/transcom/mymove/blob/a4630e461c3bdf43bd6b42cb32bdda7a9228bef6/pkg/services/mto_shipment/mto_shipment_creator_test.go#L209

pkg/services/ppmshipment/ppm_shipment_creator.go
```diff
@@ -0,0 +1,88 @@
...
+		// Check that mtoshipment is created. If not, bail out.
+		if err != nil {
+			return err
+		}
+
+		ppmShipment.ShipmentID = createShipment.ID
+		// Update the model with ppm shipment data:
+		ppmShipment.Shipment = *createShipment
+		// Validate the ppmShipment, and return an error
+		err = validatePPMShipment(txnAppCtx, *ppmShipment, nil, &ppmShipment.Shipment)
```
felipe-lee, line 68: Should this call to `validatePPMShipment` also be passing in the checks that the creator receives?

pkg/services/ppmshipment/ppm_shipment_creator.go
```diff
@@ -0,0 +1,88 @@
...
+		checks: []ppmShipmentValidator{
+			checkShipmentID(),
+			checkPPMShipmentID(),
+			checkRequiredFields(),
+		},
+	}
+}
+
+// CreatePPMShipmentCheck passes a validator key to CreatePPMShipment
+func (f *ppmShipmentCreator) CreatePPMShipmentCheck(appCtx appcontext.AppContext, ppmShipment *models.PPMShipment) (*models.PPMShipment, error) {
```
felipe-lee, line 31: Sorry, this might just be a me thing, but I keep getting confused with this name. Could it be something like `CreatePPMShipmentWithDefaultChecks` or something like that? I think it's just that as is without anything between `Shipment` and `Check`, it keeps throwing me off.

pkg/services/ppmshipment/ppm_shipment_creator.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+
+		if ppmShipment.Status == "" {
+			ppmShipment.Status = models.PPMShipmentStatusDraft
+		} else if ppmShipment.Status != models.PPMShipmentStatusDraft {
+			return apperror.NewInvalidInputError(uuid.Nil, nil, nil, "Must have a DRAFT status associated with PPM shipment")
+		}
+
+		// NOTE: The ppm may require a service item for pricing.Passing an HHG service item may be sufficient for the pricer.
+		createShipment, err := f.mtoShipmentCreator.CreateMTOShipment(txnAppCtx, &ppmShipment.Shipment, nil)
```
felipe-lee, line 58: Something I hadn't thought of is that I know that `pop` can have issues with transactions within transactions. The `CreateMTOShipment` function also starts a transaction so I don't know if that could cause any issues since we're calling it from within a transaction. I haven't had a chance to test if it could cause issues (and at the moment am not 100% _how_ I would test that), but we should probably check. I can look into this tomorrow unless someone else can quickly test it.

pkg/services/ppmshipment/validation_test.go
```diff
@@ -0,0 +1,29 @@
+package ppmshipment
+
```
felipe-lee, line 2: Can you add tests for the validation file?

felipe-lee (commented):
Haven't had a chance to finish reviewing it, but I'll continue my review tomorrow. For now, posting the comments I've had so far.

pkg/services/ppmshipment/validation_test.go
NamibiaTorres: Yes, this is specifically for the udpate. I commented them out so that when this PR is merged, Ronak can fill out the update tests.

pkg/services/ppmshipment/ppm_shipment_creator_test.go
duncan-truss: @NamibiaTorres To your question in standup I think passing the `checks` to validatePPMShipment is the main thing I think should be added before merging.  We can come back to some of the extra tests. @felipe-lee called out the same thing https://github.com/transcom/mymove/pull/8142#discussion_r806371136

felipe-lee (approved):
At this point, the main things that were needed have been addressed I think. Any other remaining comments can likely be addressed in the followup PRs we're going to have for this.

duncan-truss (approved):

pkg/services/ppmshipment/ppm_shipment_creator_test.go
NamibiaTorres: Great! This made sense. I ended up going through the comments with Felipe and making pushing up some changes.

pkg/services/ppmshipment/ppm_shipment_creator.go
felipe-lee: Btw, @ronaktruss and I found a bit of code with a comment that confirms transactions within transactions are not supported, but, this bit of code handles that for us by checking if we're already in one, and if so, doesn't start a new one.

https://github.com/transcom/mymove/blob/9931ccdcd4bb66484c2cd4315e149b3f6cceb780/pkg/appcontext/context.go#L59-L69

PR #8149, opened 2022-02-10
MB-11467 Make the SAC and TAC Show Up for the TOO

YanZ777:
## Summary

JP beat me to the punch. I'm just adding the ordersLOA as a prop for the approved shipments as well. And fixing the naming of stuff.
[Work done in JP's PR](https://github.com/transcom/mymove/pull/8109)

Additionally, I've put in a fix for a prop type validation on the orders type for oneOf. I've grabbed the values from the object.

There's currently a bug with the tacType and sacType not being able to be cleared. That bug is not handled in this PR. [MB-11643](https://dp3.atlassian.net/browse/MB-11643) is the new bug ticket.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make office_client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the office app
2. Login as a TOO
3. Select a move with NTS or NTS-R - `NTSSUB` has this information, as does `HGNTSR`
4. Add or edit the TAC and SAC values. 
5. Verify that those values show up in the Move Details page.

### Frontend


hannaliebl (approved):
LGTM

reggieriser (approved):
Changes LGTM.  Ran through the instructions and saw the TAC/SAC fine for the orders and was able to set them when editing NTS and NTS-R shipments.

pkg/services/mto_shipment/mto_shipment_updater.go
```diff
@@ -148,12 +148,16 @@ func setNewShipmentFields(appCtx appcontext.AppContext, dbShipment *models.MTOSh
 		dbShipment.BillableWeightJustification = requestedUpdatedShipment.BillableWeightJustification
 	}
 
-	if requestedUpdatedShipment.TACType != nil {
+	if requestedUpdatedShipment.TACType != nil && *requestedUpdatedShipment.TACType == "" {
```
hannaliebl, line 151: Is the not nil check required here and below?

reggieriser: My question on this is if we want a `requestedUpdatedShipment.TACType` of `nil` to overwrite whatever's already in the database.  The `nil` is going to be the zero value of that pointer field so it'll be `nil` even if not explicitly set (I believe this structure originates in the `MTOShipmentModelFromUpdate` function for the GHC handler).  So if the incoming JSON does not have a `tacType` set, I think we may accidentally overwrite the DB value when we really should not be changing it at all.  That said, this is an internal API, so we have some control over how things are sent.  It could end up being a candidate for one of the new custom nullable types we recently made that would allow us to distinguish whether it was present or not in the payload.

pkg/services/mto_shipment/mto_shipment_updater_test.go
```diff
@@ -660,6 +660,68 @@ func (suite *MTOShipmentServiceSuite) TestMTOShipmentUpdater() {
...
+		newShipment, err := mtoShipmentUpdater.UpdateMTOShipmentOffice(suite.AppContextForTest(), &updatedShipment, eTag)
+
+		suite.Require().NoError(err)
+		suite.Equal(newTacType, *newShipment.TACType)
+		suite.Equal(newSacType, *newShipment.SACType)
+
+		// Verify that shipment recalculate was handled correctly
+		mockShipmentRecalculator.AssertNotCalled(suite.T(), "ShipmentRecalculatePaymentRequest", mock.Anything, mock.Anything)
+	})
+
```
reggieriser, line 724: Just noting that the two tests you added are only exercising the `else` part of your new `if` statements.  I think you'd have to send an empty string in too to exercise the other side.  And if you do that, you might be able to just use a table-based test to run all three values through in a loop and eliminate some of the repetition.

pkg/services/mto_shipment/mto_shipment_updater_test.go
```diff
@@ -660,6 +660,68 @@ func (suite *MTOShipmentServiceSuite) TestMTOShipmentUpdater() {
...
+			},
+		})
+
+		eTag := etag.GenerateEtag(oldShipment.UpdatedAt)
+
+		updatedShipment := models.MTOShipment{
+			ID:      oldShipment.ID,
+			TACType: nil,
+			SACType: nil,
+		}
```
reggieriser, line 681: This sort of demonstrates what I was getting at above -- if you don't set the `TACType` and `SACType` explicitly to `nil` here, you should end up with the exact same `updatedShipment` struct, and any existing values for those fields will get wiped.

pkg/services/mto_shipment/mto_shipment_updater.go
reggieriser: Ideally, our PATCH would only modify fields that were explicitly provided in the payload.  The problem is that the usual swagger types don't allow us to differentiate between a `tacType` that has not been passed in the payload vs. one that has been passed in and set to null.  [This ADR](https://github.com/transcom/mymove/blob/master/docs/adr/0066-use-custom-nullable-types-for-patch-requests.md) and [this article](https://romanyx90.medium.com/handling-json-null-or-missing-values-with-go-swagger-4d7f37a2a7ca) talk about the issue.

That said, the `tacType` is only set by the office app currently, so we are more in control as to what is sent in by the front end and perhaps could make some assumptions.  Is the `tacType` always given as the empty string when we're trying to unset it?  We just need to be careful that we don't accidentally unset it just because the `tacType` wasn't explicitly provided on a shipment update somewhere in the office app.

YanZ777: Ahh I see. Yea, with the way it's currently set up now, tacTypes / sacTypes would get unintentionally cleared. I'll put in using the new custom nullable type.

LeDeep (approved):
looks good. verified that TAC and SAC are visible on move details page.

PR #8141, opened 2022-02-08
[MB-11220] Pricing on NTS-Release shipments should use ActualPickupDate

reggieriser:
## Summary

An NTS-Release shipment does not use a `RequestedPickupDate`, but that’s a key factor in our existing pricing as we use it to determine peak vs. non-peak and also the contract year we’re in (which determines escalation factors).  USTC has decided that `ActualPickupDate` should be used instead when pricing services on an NTS-Release shipment.  This PR adjusts the pricing process to allow this.

Since we have assumed `RequestedPickupDate` would always be our pricing reference point so far, this PR ends up changing a good bit of code.  Here are the highlights:
- A new migration adjusts the metadata around pricing parameters, introducing a new `ReferenceDate` parameter (for now, that will either contain the `RequestedPickupDate` or `ActualPickupDate` value depending on what type of shipment we're dealing with).  We also take any already priced services and add a `ReferenceDate` param with the existing `RequestedPickupDate` value.
- A new `ReferenceDate` lookup was created, and the `RequestedPickupDate` and `ActualPickupDate` lookups were modified.
- All existing pricers that currently use `RequestedPickupDate` (which is just about all of them) have been adjusted to look at `ReferenceDate` instead.  I've sorted parameters in the pricers as well.  Symbols have been adjusted.
- Existing tests have been adjusted as needed.  I added a new test of the payment request creator that exercises the pricing of a relevant service on an NTS-Release shipment.

A few notes:
- The front end UI has not been changed yet for these new params, so a service on an NTS-Release shipment will still show a `Requested Pickup Date` label with an undefined value, but the pricing should work and be correct (it previously failed the payment request).  A [task has been created](https://dp3.atlassian.net/browse/MB-11478) in Jira to fix this.
- If using the prime simulator to set an actual pickup date, you may find that you have to enter additional data (like addresses) before it will accept the change.
- Some discussion of how we approached this can be found in [this Slack thread](https://ustcdp3.slack.com/archives/C029R9LDTD4/p1643748589487019).

## Setup to Run Your Code

<details>
<summary>💻 You will need to use three separate terminals to test this locally.</summary>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

[...]

PR #8158, opened 2022-02-11
MB-11603 : Remove as many fmt.Print* statements as possible

ahobson:
## Summary

We should be using the logging package where appropriate.

It might (in the future) be worth adding a pre-commit check to prevent `fmt.Print`, so I changed some semi-valid uses of that to use logging instead.


YanZ777 (commented):
The code looks fine to me, but I'm getting an error when I try to populate the database. 
I think that's because [Reggie's PR](https://github.com/transcom/mymove/pull/8141), which does a database migration got merged, which I was testing out locally. I'm not sure if those changes need to be here or if I can somehow revert the migration locally.

YanZ777 (approved):
Rebasing seems to have worked for me. Everything else seems to be working normally.

reggieriser (approved):
The changes LGTM.  I did notice that we still have 3 or so `fmt.Println` calls in `defer` closures and wondered if we could get rid of those eventually too by either logging instead or using a named error param to return the error (could get tricky).

ahobson:
> 

Yeah, I saw those, too, but punted for now to keep the perfect from being the enemy of the good

PR #8171, opened 2022-02-15
[MB-10957] Add new columns for duty station rename

mboren:
## Summary
We want to rename:
- `service_members.duty_station_id` to `duty_location_id`
- `orders.origin_duty_station_id` to `origin_duty_location_id`
- `orders.new_duty_station_id` to `new_duty_location_id`

In this PR, I will be adding columns with the new names, so that we can gradually switch over to using them, and then drop the old versions of the columns without downtime.
I'm also adding triggers to keep the new columns in sync with the old ones.

my plan is
- Migration 1: create columns with new names and triggers to fill them in *<- You are here*
- Migration 2: copy all values from old columns to new columns
- Reference the new columns from the application code
- Migration 3: drop old columns and triggers.

I'd love to know if there's a way to cut down the number of steps here.

I don't think we use any triggers anywhere yet. Do we need to grant additional permissions?
Should I test this in exp?

## Setup to Run Your Code

Run migrations and populate devseed data
```sh
make db_dev_reset
make db_dev_e2e_populate
```
Make sure that the columns have been kept in sync:
```sql
-- Query for any records with mismatched ids (these should give no results):
select * from orders where origin_duty_station_id <> origin_duty_location_id or orders.new_duty_station_id <> orders.new_duty_location_id;
```
```sql
select * from service_members where duty_station_id <> duty_location_id;
```
If those queries returned no results (as expected) you can run the following queries just to do a quick spot check to make sure things look OK. This is to catch things like the DB being empty somehow or the test queries being wrong.
```sql
select id, origin_duty_station_id, origin_duty_location_id, new_duty_station_id, new_duty_location_id
from orders;

select id, duty_station_id, duty_location_id
from service_members;
```

## Manually test
```
make server_run
make client_run
```
Create a customer, then update their origin and new duty locations as a TOO or services counselor.
Make sure that the changes saved, and then check the DB to make sure `service_members.duty_location_id` `service_members.duty_station_id`, `orders.origin_duty_station_id`, and `orders.origin_duty_location_id` are all equal and that `orders.new_duty_station_id = orders.new_duty_location_id`

```
select m.id,  m.locator, sm.duty_station_id, sm.duty_location_id, o.origin_duty_station_id, o.origin_duty_location_id, o.new_duty_station_id, o.new_duty_location_id
from moves m
         join orders o on m.orders_id = o.id
         join service_members sm ON o.service_member_id = sm.id
where m.locator = 'YOUR MOVE LOCATOR';

```


migrations/app/schema/20220214222638_create_new_cols_for_duty_station_to_location_rename.up.sql
```diff
@@ -0,0 +1,56 @@
+-- Create copies of duty_station columns on service_members with new wording
+ALTER TABLE service_members
+	ADD COLUMN duty_location_id uuid
```
duncan-truss, line 3: Do you mind adding an index on these duty_location_id columns since there is one on the the existing duty_station_id columns as well please?

mboren: For sure. Thanks for catching that.

duncan-truss (approved):
Tested updating db manually and then in the app and behaved as expected.  Thanks for the testing instructions.

PR #8173, opened 2022-02-15
MB-11032 DRAFT estimated weights and progear form added to storybook

pearl-truss:
## Summary

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the
2. Login as a
3.

### Frontend

### Backend


src/components/Customer/PPMBooking/EstimatedWeightsProGear/EstimatedWeightsProGear.jsx
```diff
@@ -0,0 +1,196 @@
...
+    estimatedProGearWeight: Yup.number().when(['hasProGear', 'estimatedSpouseProGearWeight'], {
+      is: (hasProGear, estimatedSpouseProGearWeight) => hasProGear && !estimatedSpouseProGearWeight,
+      then: (schema) =>
+        schema
+          .required(`Enter a weight into at least one pro-gear field. If you won't have pro-gear, select No above.`)
+          .max(2000, 'Enter a weight less than 2,000 lbs'),
+      otherwise: Yup.number().max(2000, 'Enter a weight less than 2,000 lbs'),
+    }),
+    estimatedSpouseProGearWeight: Yup.number().max(500, 'Enter a weight less than 500 lbs'),
+  });
```
felipe-lee, line 35: Could you pull this out of the component (could even just be above the component definition). Mainly since there's not really a need to define this each time the component is loaded since it's effectively a constant.

src/components/Customer/PPMBooking/EstimatedWeightsProGear/EstimatedWeightsProGear.jsx
```diff
@@ -0,0 +1,196 @@
...
+                  label="Estimated weight of this PPM shipment"
+                  id="estimatedPPMWeight"
+                  mask={Number}
+                  scale={0} // digits after point, 0 for integers
+                  signed={false} // disallow negative
+                  thousandsSeparator=","
+                  lazy={false} // immediate masking evaluation
+                  suffix="lbs"
+                  warning={
+                    values.estimatedPPMWeight > entitlement?.authorizedWeight
```
felipe-lee, line 72: In the alert above where we tell customers their weight allowance we don't use the optional chaining and I feel like we should probably be consistent in that. One thing too with the optional chaining is that if, for whatever reason, `entitlement` is `null`/`undefined`, then this would always evaluate to `false`, which could be what we want, but if so, I feel like it should either be explicit in the code, or called out in a comment. 

I just tested in storybook what happens if `entitlement` is `undefined` (I didn't realize how easy storybook made it to mess with props), and got a `TypeError`, which is expected since the alert doesn't do the optional chaining, but then if I change it to allow optional chaining, it ends up in a strange state where the alert says weight entitlement is 0lbs, but there's no warning for being over the weight entitlement. Again, I don't really know what is preferred, but we should probably decide and make the necessary changes to make sure it makes sense.

src/components/Customer/PPMBooking/EstimatedWeightsProGear/EstimatedWeightsProGear.stories.jsx
```diff
@@ -0,0 +1,68 @@
...
+            <Story />
+          </Grid>
+        </Grid>
+      </GridContainer>
+    ),
+  ],
+};
+
+const Template = (args) => <EstimatedWeightsProGear {...args} />;
+
```
felipe-lee, line 25: Should we maybe have a story showing the warning by default of weight being more than entitlement?

felipe-lee (commented):
Haven't finished looking, but added some comments

src/components/Customer/PPMBooking/EstimatedWeightsProGear/EstimatedWeightsProGear.jsx
```diff
@@ -0,0 +1,196 @@
...
+          </div>
+        );
+      }}
+    </Formik>
+  );
+};
+
+EstimatedWeightsProGear.propTypes = {
+  entitlement: EntitlementShape.isRequired,
+  mtoShipment: MtoShipmentShape,
```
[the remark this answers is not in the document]
pearl-truss: Yeah I noticed this as well and went with consistency assuming we may need fields on `mtoShipment` at some point. Though I could also see a case of just pulling in `ppmShipment` directly for both forms and updating them if we ever do need fields from `mtoShipment`

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.jsx
```diff
@@ -0,0 +1,198 @@
...
+            <Form className={(formStyles.form, ppmBookingStyles.form)}>
+              <Alert type="info">{`Total weight allowance for your move: ${formatWeight(
+                entitlement?.authorizedWeight,
+              )}`}</Alert>
+              <SectionWrapper className={classnames(ppmBookingStyles.sectionWrapper, formStyles.formSection)}>
+                <h2>Full PPM</h2>
+                <p>
+                  Estimate the full weight of your PPM, including everything you plan to move. If you’re moving pro-gear
+                  in this PPM, include that weight in this estimate. Do not count weight twice, though. Do not include
+                  weight in your estimate that will be moved in other shipments.{' '}
```
felipe-lee, line 58: It looks like in the zeplin mocks, this bit of text is separate from the bit before, so I think it probably needs to go into its own `<p></p>` tags.

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.stories.jsx
```diff
@@ -0,0 +1,84 @@
...
+  onSubmit: action('submit button clicked'),
+  onBack: action('back button clicked'),
+  entitlement: {
+    authorizedWeight: 5000,
+  },
+  mtoShipment: {
+    id: '123',
+    ppmShipment: {
+      id: '123',
+      estimatedWeight: '7000',
```
felipe-lee, line 46: I'm getting a prop type error in the console because it expects this to be a number; can you update it to be a number instead?

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.stories.jsx
```diff
@@ -0,0 +1,84 @@
...
+    authorizedWeight: 5000,
+  },
+  mtoShipment: {
+    id: '123',
+    ppmShipment: {
+      id: '123',
+      hasProGear: 'true',
+      estimatedProGearWeight: '1000',
+      estimatedSpouseProGearWeight: '100',
+      estimatedWeight: '4000',
```
felipe-lee, line 65: I'm also getting a prop error for `hasProGear` because it expects a bool instead of a string. I'm not getting any for the weights, but I assume once the `hasProGear` is fixed, it will trigger on the next one in line.

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.test.jsx
```diff
@@ -0,0 +1,110 @@
...
+    it('displays secondary pro gear weight inputs when hasProGear is true', async () => {
+      render(<EstimatedWeightsProGearForm {...defaultProps} />);
+      const hasProGear = await screen.getByLabelText('Yes');
+      expect(screen.queryByLabelText('Estimated weight of your pro-gear')).toBeNull();
+      expect(screen.queryByLabelText('Estimated weight of your spouse’s pro-gear')).toBeNull();
+      userEvent.click(hasProGear);
+
+      await waitFor(() => {
+        expect(screen.queryByLabelText('Estimated weight of your pro-gear')).toBeInstanceOf(HTMLInputElement);
+        expect(screen.queryByLabelText('Estimated weight of your spouse’s pro-gear')).toBeInstanceOf(HTMLInputElement);
```
felipe-lee, line 51: Since these are things we expect to be there, the query should be `screen.getByLabelText`. The `queryByLabelText` is meant for when you expect something to not be present.

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.test.jsx
```diff
@@ -0,0 +1,110 @@
...
+        mtoShipmentProps.mtoShipment.ppmShipment.estimatedSpouseProGearWeight,
+      );
+    });
+  });
+
+  describe('validates form fields and displays error messages', () => {
+    it('marks required inputs when left empty', async () => {
+      render(<EstimatedWeightsProGearForm {...defaultProps} />);
+
+      await userEvent.click(screen.getByRole('button', { name: 'Save & Continue' }));
```
felipe-lee, line 76: This test is actually making me wonder if we should have this button disabled on page load if there is no weight, instead of disabling it once they try to submit with no data. Not sure how you feel about that, or if that's more of a design question.

felipe-lee (commented):
Thanks for making the other changes! I added a few more comments. Also, I noticed that the zeplin mocks have a `PPM` badge on the top left that aren't in the component. I'm not sure if it's meant to be at this level, or if that's going to be done at a higher level:
Screen Shot 2022-02-16 at 12 20 46

pearl-truss:
> Thanks for making the other changes! I added a few more comments. Also, I noticed that the zeplin mocks have a `PPM` badge on the top left that aren't in the component. I'm not sure if it's meant to be at this level, or if that's going to be done at a higher level: Screen Shot 2022-02-16 at 12 20 46

I left this and heading out because I think it should happen at the page level. I updated the AC in the ticket for incorporating this form into a page so that this doesn't get lost https://dp3.atlassian.net/browse/MB-10675

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.test.jsx
pearl-truss: Sounds like a question for @transcom/truss-design

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.jsx
```diff
@@ -0,0 +1,198 @@
...
+import formStyles from 'styles/form.module.scss';
+import { EntitlementShape } from 'types/order';
+import MaskedTextField from 'components/form/fields/MaskedTextField/MaskedTextField';
+import Hint from 'components/Hint/index';
+import SectionWrapper from 'components/Customer/SectionWrapper';
+import Fieldset from 'shared/Fieldset';
+import { formatWeight } from 'utils/formatters';
+
+const validationSchema = Yup.object().shape({
+  estimatedPPMWeight: Yup.number().required('Required'),
```
duncan-truss, line 23: What do you think about enforcing a min of 1 lbs for these weights so we don't accept zeros as valid?  I know the input field restricts something like negative numbers.

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.jsx
```diff
@@ -0,0 +1,207 @@
...
+  };
+
+  return (
+    <Formik initialValues={initialValues} validationSchema={validationSchema} onSubmit={onSubmit}>
+      {({ isValid, isSubmitting, handleSubmit, values }) => {
+        return (
+          <div className={classnames(styles.EstimatedWeightsProGearForm, ppmBookingStyles.formContainer)}>
+            <Form className={(formStyles.form, ppmBookingStyles.form)}>
+              <Alert type="info">{`Total weight allowance for your move: ${formatWeight(
+                entitlement?.authorizedWeight,
```
duncan-truss, line 57: authorizedWeight is really an office/prime value that we probably shouldn't be returning in the internal API honestly.  I think you're right in that it has the right default value before a move is submitted but it could change to something wildly different.

I think we should do the lookup similar to how we're doing it elsewhere on the mtoShipmentForm https://github.com/transcom/mymove/blob/96b3031f8d809ac6196a2394c272d813e508717a/src/components/Customer/MtoShipmentForm/MtoShipmentForm.jsx#L213-L215

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.jsx
```diff
@@ -0,0 +1,207 @@
...
+                  defaultValue="0"
+                  name="estimatedWeight"
+                  label="Estimated weight of this PPM shipment"
+                  id="estimatedWeight"
+                  mask={Number}
+                  scale={0} // digits after point, 0 for integers
+                  signed={false} // disallow negative
+                  thousandsSeparator=","
+                  lazy={false} // immediate masking evaluation
+                  suffix="lbs"
```
duncan-truss, line 79: Oh nice at first I thought this was a prop I never noticed before and then I saw you added it in 🙌 I think I remember there is still an issue with this component not being marked as touched if it is just clicked into and then blurred/unfocused.  Not sure if the issue is we're not passing in the `{...field}` props to the component like is happening here https://formik.org/docs/api/useField

scarequotes (changes requested):
When the field is in an error state, there's some extra space added above the field label that moves all the text down. The error indication line to the left of the field shows blank space as well. Not expected behavior. 

(Also shows up in @duncan-truss's https://github.com/transcom/mymove/pull/8177)

Screen Shot 2022-02-17 at 2 47 18 PM

duncan-truss:
> When the field is in an error state, there's some extra space added above the field label that moves all the text down. The error indication line to the left of the field shows blank space as well. Not expected behavior.
> 
> (Also shows up in @duncan-truss's #8177)
> 
> Screen Shot 2022-02-17 at 2 47 18 PM

@scarequotes Thanks for flagging this James, we didn't introduce it but it was definitely a bug 🐛 in our global styles.  I think I have a fix for it here in my PR https://github.com/transcom/mymove/pull/8177/commits/3d3e62e541ffababcd93c71bd97fd709d9e668f3

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.test.jsx
felipe-lee: Np! I got that from this article, which in general has some other great tips on writing tests with RTL: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.stories.jsx
felipe-lee, line 46: I think this one also needs to be updated to be a number, I'm getting a prop type error.

yarn.lock
```diff
@@ -2792,18 +2792,18 @@
...
-"@storybook/addons@6.4.18":
-  version "6.4.18"
-  resolved "https://registry.yarnpkg.com/@storybook/addons/-/addons-6.4.18.tgz#fc92a4a608680f2e182a5e896ed382792f6b774e"
-  integrity sha512-fd3S79P4jJCYZNA2JxA1Xnkj0UlHGQ4Vg72aroWy4OQFlgGQor1LgPfM6RaJ9rh/4k4BXYPXsS7wzI0UWKG3Lw==
-  dependencies:
-    "@storybook/api" "6.4.18"
-    "@storybook/channels" "6.4.18"
-    "@storybook/client-logger" "6.4.18"
-    "@storybook/core-events" "6.4.18"
+"@storybook/addons@6.4.19":
```
felipe-lee, line 2795: I wonder if a merge from master would fix this. I'm not really sure why the `yarn.lock` thinks it should have updates.

src/components/form/fields/MaskedTextField/MaskedTextField.jsx
```diff
@@ -51,23 +54,48 @@ const MaskedTextField = ({
...
+          lazy={lazy}
+          onAccept={(val, masked) => {
+            helpers.setValue(masked.unmaskedValue);
+            // setValue is already triggering validation for this field so we should be able to skip it in setTouched
+            helpers.setTouched(true, false);
+          }}
+          {...props}
+          {...field}
+        />
+      )}
```
felipe-lee, line 98: I wonder if it would be good to pull this out and define it before we return so that we don't have to define it twice. Mainly wondering because right now there's a potential for drift between the two, but since they're in the same file, that potential is also low so I don't mind if it stays as is.

felipe-lee: So something like:

```jsx
  /* eslint-disable react/jsx-props-no-spreading */
  const input = (
    <IMaskInput
      className={classnames('usa-input', inputClassName)}
      data-testid={inputTestId}
      type={type}
      id={id}
      name={name}
      value={value ?? defaultValue}
      mask={mask}
      blocks={blocks}
      lazy={lazy}
      onAccept={(val, masked) => {
        helpers.setValue(masked.unmaskedValue);
        // setValue is already triggering validation for this field so we should be able to skip it in setTouched
        helpers.setTouched(true, false);
      }}
      {...props}
      {...field}
    />
  );
  /* eslint-enable react/jsx-props-no-spreading */

  return (
    <FormGroup className={classnames(!!warning && !showError && `warning`, formGroupClassName)} error={showError}>
      <div className="labelWrapper">
        <Label className={labelClassName} hint={labelHint} error={showError} htmlFor={id || name}>
          {label}
        </Label>
        {optional && <OptionalTag />}
      </div>
      {showError && (
        <ErrorMessage display={showError} className={errorClassName}>
          {metaProps.error ? metaProps.error : errorMessage}
        </ErrorMessage>
      )}
      {suffix ? (
        <div className={suffix && styles.hasSuffix}>
          {input}
          {suffix && <div className="suffix">{suffix}</div>}
        </div>
      ) : (
        { input }
      )}
      {children}
    </FormGroup>
  );
```

pearl-truss: I introduced the double code blocks because it trigged client test failures for test suites that were checking for 
`    expect(field.closest('div').nextElementSibling.name).toBe(inputName);
`
And I tried this way with modifications but it was still causing 30+ client test errors. So I switched it back to how it was originally written for the time being

pearl-truss:
> > When the field is in an error state, there's some extra space added above the field label that moves all the text down. The error indication line to the left of the field shows blank space as well. Not expected behavior.
> > (Also shows up in @duncan-truss's #8177)
> > Screen Shot 2022-02-17 at 2 47 18 PM
> 
> @scarequotes Thanks for flagging this James, we didn't introduce it but it was definitely a bug 🐛 in our global styles. I think I have a fix for it here in my PR https://happo.io/a/488/p/603/compare/9931ccdcd4bb66484c2cd4315e149b3f6cceb780/fe57debf2788182ac4b16b2052076e36eb514807. We can double check it doesn't create a bunch of Happo diffs, unless it fixes the problem elsewhere too.

@scarequotes Are you okay with removing the "changes requested" flag from this PR since the UI bug has been fixed in Duncan's PR?

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.jsx
```diff
@@ -0,0 +1,209 @@
...
+                  // formatWeight will display 0lbs if the weight is undefined.
+                  // therefore if entitlement is undefined display the overweight warning
+                  warning={
+                    values.estimatedWeight > authorizedWeight || !authorizedWeight
```
duncan-truss, line 84: What is this the second part of this conditional displaying they warning for?

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.jsx
```diff
@@ -0,0 +1,209 @@
...
+                  Your final incentive amount will be determined by your finance office, based on certified weight
+                  tickets that document the actual weight you moved in your PPM.
+                </p>
+                <h3>Need help estimating your PPM&apos;s weight?</h3>
+                <p>A good guideline: Estimate 1,000 to 1,500 lbs per room.</p>
+
+                <p>
+                  If you own a lot of things for your space, estimate on the higher side. If you own less, estimate
+                  lower. The services have an official{' '}
+                  <a href="https://www.ustranscom.mil/dp3/weightestimator.cfm" target="_blank" rel="noreferrer">
```
duncan-truss, line 105: Could you also include the `noopener` value in the rel attribute to be safe?  I think defaults have changed in modern browsers but not sure we can rely on our users being up to date.

duncan-truss (approved):
Thanks for addressing my comments

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.jsx
pearl-truss: true, I removed the second part of this conditional since t's no longer needed

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.jsx
```diff
@@ -0,0 +1,212 @@
...
+                <h3>Need help estimating your PPM&apos;s weight?</h3>
+                <p>A good guideline: Estimate 1,000 to 1,500 lbs per room.</p>
+
+                <p>
+                  If you own a lot of things for your space, estimate on the higher side. If you own less, estimate
+                  lower. The services have an official{' '}
+                  <a
+                    href="https://www.ustranscom.mil/dp3/weightestimator.cfm"
+                    target="_blank"
+                    noopener
```
felipe-lee, line 106: I'm getting a console error that I think is coming from here: 

```text
Received `true` for a non-boolean attribute `noopener`.
```

I think it needs to be passed to the `rel` attribute based on what I'm seeing here: https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener

I _think_ we should be able to pass more than one value to it, using a space-separated list, so something like `rel="noopener noreferrer"`

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.stories.jsx
```diff
@@ -0,0 +1,104 @@
...
+const Template = (args) => <EstimatedWeightsProGearForm {...args} />;
+
+export const BlankEstimatedWeightsProGear = Template.bind({});
+BlankEstimatedWeightsProGear.args = {
+  onSubmit: action('submit button clicked'),
+  onBack: action('back button clicked'),
+  orders: {
+    has_dependents: true,
+  },
+  serviceMember: {
```
felipe-lee, line 33: I'm getting a prop type error for a missing `id`:

```text
Warning: Failed prop type: The prop `serviceMember.id` is marked as required in `EstimatedWeightsProGearForm`, but its value is `undefined`.
```

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.stories.jsx
```diff
@@ -0,0 +1,104 @@
...
+};
+
+export const WarningForOverweightEstimatedWeightProGear = Template.bind({});
+WarningForOverweightEstimatedWeightProGear.args = {
+  onSubmit: action('submit button clicked'),
+  onBack: action('back button clicked'),
+  orders: {
+    has_dependents: true,
+  },
+  serviceMember: {
```
felipe-lee, line 47: Same comment on the missing `id`.

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.stories.jsx
```diff
@@ -0,0 +1,104 @@
...
+};
+
+export const MTOShipmentEstimatedWeightProGear = Template.bind({});
+MTOShipmentEstimatedWeightProGear.args = {
+  onSubmit: action('submit button clicked'),
+  onBack: action('back button clicked'),
+  orders: {
+    has_dependents: true,
+  },
+  serviceMember: {
```
felipe-lee, line 68: Same comment on the missing `id`.

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.stories.jsx
```diff
@@ -0,0 +1,104 @@
...
+};
+
+export const ErrorEstimatedWeightsProGear = Template.bind({});
+ErrorEstimatedWeightsProGear.args = {
+  onSubmit: action('submit button clicked'),
+  onBack: action('back button clicked'),
+  orders: {
+    has_dependents: true,
+  },
+  serviceMember: {
```
felipe-lee, line 92: Same comment on the missing `id`.

felipe-lee (approved):
Thanks for making those changes! I'm seeing a few other console errors that I commented on, but otherwise LGTM!

scarequotes (approved):

PR #8185, opened 2022-02-17
MB-11554: Update PPMs

ronaktruss:
## Summary

Latest changes (newest -> oldest):
- Added an additional rule to ensure mtoShipment has a shipmentType of PPM
- cleaned up comments
- moved extra handler tests to out of scope (basic functionality is tested, but not specific errors (ie. 401, 403, etc)

Here are some pending tasks that need to be included in this PR:
- [x] clean up model <-> payload conversions + new model for update endpoint
- [x] Add default value of 10000 for estimated incentive
- [x] create merge function and add tests

The following are nice to haves that I would like to include, but may not be done in this PR:
- [ ] add more tests for handler
- [ ] Service object ADR / refactor
- [ ] handling ETags? (I can likely add this if there is a consensus on how we are handling this)
- [ ] Add DB migration and convert estimated weights to cents
- [ ] Add support for patching empty/nil values (I assume this will likely be incorporated as part of the edit story that we have, [MB-11350](https://dp3.atlassian.net/browse/MB-11350))
    - [ ] We will likely need to change the model a bit to handle default values for something like `sitExpected` (ref: [slack thread](https://ustcdp3.slack.com/archives/CP6F568DC/p1646085498396549))
* * * 

## Setup to Run Your Code
Please read through the code changes. To run the modified tests locally you can follow these steps:

1. setup your local tests by running:
```
make server_test_setup
```
2. run the follow 3 tests (feel free to add the -v flag after `go test` for a verbose output:
```
go test ./pkg/handlers/internalapi/ -testify.m TestUpdateMTOShipmentHandler
go test ./pkg/services/ppmshipment
go test ./pkg/services/ppmshipment/ -testify.m TestMergePPMShipment
```

### Frontend

### Backend


swagger-def/internal.yaml
```diff
@@ -2381,6 +2383,17 @@ definitions:
       - APPROVED
       - SUBMITTED
       - REJECTED
+  PPMShipmentStatus:
+    example: 'DRAFT'
+    type: string
+    readOnly: true
```
ronaktruss, line 2389: Should this be readOnly? Are we planning on handling this the same way as the status on `mto_shipment`?

pkg/models/ppm_shipment.go
```diff
@@ -49,7 +49,7 @@ type PPMShipment struct {
 	HasProGear                     *bool             `json:"has_pro_gear" db:"has_pro_gear"`
 	ProGearWeight                  *unit.Pound       `json:"pro_gear_weight" db:"pro_gear_weight"`
 	SpouseProGearWeight            *unit.Pound       `json:"spouse_pro_gear_weight" db:"spouse_pro_gear_weight"`
-	EstimatedIncentive             *int32            `json:"estimated_incentive" db:"estimated_incentive"`
+	EstimatedIncentive             *unit.Cents       `json:"estimated_incentive" db:"estimated_incentive"`
```
ronaktruss, line 52: Do we want to use Dollars here? Cents is more flexible.

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -454,13 +479,17 @@ func (suite *HandlerSuite) TestUpdateMTOShipmentHandler() {
 	recalculator := paymentrequest.NewPaymentRequestRecalculator(creator, statusUpdater)
 	paymentRequestShipmentRecalculator := paymentrequest.NewPaymentRequestShipmentRecalculator(recalculator)
 
+	// TODO: Confirm if we need to have a new handler objects, instructions seem to advice to put into a helper function and call per test
```
ronaktruss, line 482: It's not clear to me what needs to be reset during each test. As far as I can tell it looks like it is only the handler itself. Do we need to create a new updaters as well?

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -490,13 +519,42 @@ func (suite *HandlerSuite) TestUpdateMTOShipmentHandler() {
 		suite.NotEmpty(updatedShipment.Agents[0].ID)
 	})
 
+	suite.Run("Successful PATCH with PPMShipment - Integration Test", func() {
```
ronaktruss, line 522: Need to add tests for additional ppm edge cases / errors

pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -114,7 +116,22 @@ func (h UpdateMTOShipmentHandler) Handle(params mtoshipmentops.UpdateMTOShipment
 				h.GetTraceIDFromRequest(params.HTTPRequest)))
 	}
 
-	updatedMtoShipment, err := h.mtoShipmentUpdater.UpdateMTOShipmentCustomer(appCtx, mtoShipment, params.IfMatch)
+	var updatedMTOShipment *models.MTOShipment
+	var updatedPPMShipment *models.PPMShipment
+	var err error
+	err = appCtx.NewTransaction(func(txnAppCtx appcontext.AppContext) error {
```
ronaktruss, line 122: Would like to get some additional feedback regarding the usage of service objects in this manner.

pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -114,7 +116,22 @@ func (h UpdateMTOShipmentHandler) Handle(params mtoshipmentops.UpdateMTOShipment
 				h.GetTraceIDFromRequest(params.HTTPRequest)))
 	}
```
ronaktruss, line 117: My understanding is that this logic in the above block/section should likely be moved to the service object. Is that correct?

pkg/services/ppmshipment/ppm_shipment_updater_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	oldPPMShipment := testdatagen.MakeDefaultPPMShipment(suite.DB())
+
+	newPPM := models.PPMShipment{
+		ID:          oldPPMShipment.ID,
+		SitExpected: models.BoolPointer(true),
+	}
+
+	// eTag := etag.GenerateEtag(oldPPMShipment.UpdatedAt)
+	suite.T().Run("UpdatePPMShipment - Success", func(t *testing.T) {
```
ronaktruss, line 25: Based on comments in the code it looks like the expectation is to use something like suite.Run("name", func() { ... }) instead of the testify run function.

pkg/handlers/internalapi/mto_shipment.go
felipe-lee: Ah, that's a hard one because it's specific to the customer side of things, but the service objects are used across the applications. I think we might have helpers to know which app we're in though. If so, then yeah, I think this should probably be in a service object, probably in the rules file.

pkg/handlers/internalapi/mto_shipment.go
felipe-lee: I don't actually know if we have a pattern set for this, but my initial thought is that the handler shouldn't be managing transactions. I'd be curious what others think.

pkg/handlers/internalapi/mto_shipment_test.go
felipe-lee: It's a bit of a combo of what needs to be reset and what should likely not get repeated. Like any data creation should happen in the setup helper function for the tests. I think it should be the same with things that are repeated, e.g. the creation of the updater. At least if it's being created the same way in each test.

pkg/models/ppm_shipment.go
felipe-lee: I think I remember it being said in a meeting we didn't want to allow cents, only round dollars. Seems like we have a Dollars type, but it's a float. If we want to stick with using our custom types, we could add a custom one without cents.

pkg/services/ppmshipment/ppm_shipment_updater_test.go
felipe-lee: I'm not sure which comments you're specifically meaning, but we have both in the codebase in part because we never fully rolled out the transaction tests: https://transcom.github.io/mymove-docs/docs/backend/testing/running-server-tests-inside-a-transaction

swagger-def/internal.yaml
felipe-lee: Yup, it should be `readOnly` and I think it should be like the mto shipment one. This is also going to be in the other PR I opened that adds the PPMShipment to MTOShipment.

pkg/handlers/internalapi/internal/payloads/model_to_payload.go
```diff
@@ -29,6 +29,37 @@ func Address(address *models.Address) *internalmessages.Address {
 	}
 }
 
+func PPMShipment(ppmShipment *models.PPMShipment) *internalmessages.PPMShipment {
```
felipe-lee, line 32: Sorry about the double work, but this is also being done in #8206

felipe-lee (commented):
I haven't had a chance to review more thoroughly, but I tried adding responses to your questions, or as best as I could anyway.

swagger-def/internal.yaml
```diff
@@ -2413,13 +2428,113 @@ definitions:
...
+      - sitExpected
   UpdateShipment:
     type: object
     properties:
       status:
         $ref: '#/definitions/MTOShipmentStatus'
       shipmentType:
         $ref: '#/definitions/MTOShipmentType'
+      ppmShipment:
+        $ref: '#/definitions/PPMShipment'
```
duncan-truss, line 2537: I think we need a separate definition type for this instead of reusing PPMShipment. This definition makes sense as a response type or maybe a create type w/ some changes.  For the update endpoint though we can't specify those fields as required to be sent with every request.  I think from the estimated weights screen for example, we don't want the form submission to have to send the full body of the ppm shipment in the update request for PATCH requests only the fields that are changing.

pkg/models/ppm_shipment.go
ronaktruss: Yeah as far as requirements are concerned we are only showing whole dollar amounts. I suggested using cents because it looked like cents was being used more often (this would mean we would have a single set of helper functions to do conversions and modifications on the frontend and on  the backend). It would also be easier to make the switch to supporting showing fractional cents more easily should that be the case in the future without having to deal with floats.

pkg/handlers/internalapi/mto_shipment_test.go
ronaktruss: I don't think I understand why we would want something like the creation of the updater to happen to multiple times by moving it to a helper function that is run as part of each test. Wouldn't it be better not to create new instances of it? Or is there some singleton pattern behind the scenes that makes it so that we aren't actually creating additional versions while keeping the code consistent from test to test.

pkg/handlers/internalapi/mto_shipment.go
duncan-truss: I think most of the time we want to avoid transactions in the handlers.  At the same time we need to be able to compose service objects so that we're not combining distinct service objects and making them less independent and harder to test.

I could see a generic wrapper that composes service objects or embedding the update ppm service object within the mto shipment updater.  We shouldn't always be calling the update ppm shipment service object, only if the shipment is of type PPM.  That feels like a decision that the handler can delegate to the mto shipment updater?

felipe-lee: Based on the meeting we just had in the BE check-in, most were in favor of moving toward composable service objects, with having a layer on top for more complex things like shipments that would then know when to use the base mto service objects vs the ppm service objects. Next steps are to write an ADR about it. Not sure if we want to bring that into this PR though since it would expand the scope of what this was originally meant to do.

pkg/models/ppm_shipment.go
felipe-lee: We'll still need separate helpers (or modified ones) to handle returning the values without cents though since we don't want to present that data as having cents. But I'd be fine with using the cents at a db level. This would require a DB migration though because currently the field is an int.

pkg/handlers/internalapi/mto_shipment_test.go
felipe-lee: We don't use a pattern to avoid recreating things at the moment. I think we're optimizing more for ease of reuse/reading than for optimal setup test code in and of itself. We could likely implement some sort of in memory cache at a function level, or use a singleton pattern like you said. I'm not sure if we have that implemented already in other bits of this code, or this would be the first place, but it would be helpful for things like this that likely won't be different across multiple tests.

pkg/services/ppmshipment/ppm_shipment_updater.go
```diff
@@ -0,0 +1,103 @@
...
+	// Find the associated mtoShipment, return an error if not found
+	err = appCtx.DB().Find(&mtoShipment, oldPPMShipment.ShipmentID)
+	if err != nil {
+		switch err {
+		case sql.ErrNoRows:
+			return nil, apperror.NewNotFoundError(oldPPMShipment.ShipmentID, "while looking for MTOShipment")
+		default:
+			return nil, apperror.NewQueryError("MTOShipment", err, "")
+		}
+	}
```
duncan-truss, line 66: I think you could combine these into 1 query by finding on one ID and eager joining the other association.

One other thing I'm a little worried about is relying on the provided IDs because they could be replaced to update any shipment record not belonging to the service member's move.  We don't have to add that authorization in this PR.  I think we could simplify to update the PPM Shipment that is associated with the mto shipment id instead of relying on what that ppm shipment id value is set to (we may not need it in the update payload).

ronaktruss: Agreed.

Is their an authorization check on mto_shipments that prevents similar updates? If so, can you point me to where that's happening in the code? I see `CheckIfMTOShipmentCanBeUpdated` but it looks like that would allow for updates to any mto_shipment given I am a "customer".

duncan-truss: Yea I don't think that authorization is happening anywhere, nor do we necessarily want to do it ad-hoc for every handler.  We would have to tie the service member id back to the move (really the orders) and make sure the shipment belongs to the move.  Let's ignore it for now.

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -490,13 +518,48 @@ func (suite *HandlerSuite) TestUpdateMTOShipmentHandler() {
...
+		handler := UpdateMTOShipmentHandler{
+			handlers.NewHandlerContext(suite.DB(), suite.Logger()),
+			updater,
+			ppmUpdater,
+		}
+
+		existingPPMShipment := testdatagen.MakeDefaultPPMShipment(suite.DB())
```
duncan-truss, line 528: I think this might be a better test if we start with the MakeMinimalPPMShipment that only has the required fields set at creation time.  Then include more fields to update in your payload instead of just checking what values already existed from the testdatagen.

pkg/services/ppmshipment/rules.go
```diff
@@ -47,9 +47,6 @@ func checkRequiredFields() ppmShipmentValidator {
 	return ppmShipmentValidatorFunc(func(_ appcontext.AppContext, newPPMShipment models.PPMShipment, oldPPMShipment *models.PPMShipment, _ *models.MTOShipment) error {
 		verrs := validate.NewErrors()
 
-		// TODO: We'll need to add logic for when a shipment is being updated, otherwise the code below could break.
-		// Look at pkg/services/reweigh/rules.go  checkRequiredFields() for an example.
-
```
ronaktruss, line 52: @felipe-lee I removed this because I am merging the old and new ppmShipments before validating and that should do what is currently being done in the rewight/rules.go. Let me know if I missed anything.

pkg/services/ppmshipment/ppm_shipment_updater.go
```diff
@@ -0,0 +1,66 @@
...
+func (f *ppmShipmentUpdater) updatePPMShipment(appCtx appcontext.AppContext, ppmShipment *models.PPMShipment, mtoShipmentID uuid.UUID, checks ...ppmShipmentValidator) (*models.PPMShipment, error) {
+	if ppmShipment == nil {
+		return nil, nil
+	}
+
+	oldPPMShipment, err := models.FetchPPMShipmentFromMTOShipmentID(appCtx.DB(), mtoShipmentID)
+	if err != nil {
+		return nil, err
+	}
+	// if etag.GenerateEtag(oldPPMShipment.UpdatedAt) != eTag {
```
pearl-truss, line 39: Could we get rid of these lines since they're commented out?

pkg/handlers/internalapi/mto_shipment.go
pearl-truss: I agree with wanting to avoid having transactions in handlers. Since the next step is to create an ADR around composable service objects, (which may take awhile) could we capture in a code comment a tldr for moving this logic into the new layer the ADR will introduce?

pkg/models/ppm_shipment.go
```diff
@@ -64,3 +67,21 @@ type PPMShipments []PPMShipment
...
+	var ppmShipment PPMShipment
+
+	err := db.EagerPreload("Shipment").
+		Where("ppm_shipments.shipment_id = ?", mtoShipmentID).
+		First(&ppmShipment)
+
+	if err != nil {
+		switch err {
+		case sql.ErrNoRows:
+			return nil, apperror.NewNotFoundError(ppmShipment.ID, "while looking for PPMShipment")
```
duncan-truss, line 81: This ID won't exist if the query returns no results or will be uuid.Nil, you could use the mtoShipmentID instead

pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -114,7 +116,22 @@ func (h UpdateMTOShipmentHandler) Handle(params mtoshipmentops.UpdateMTOShipment
...
+	var updatedMTOShipment *models.MTOShipment
+	var updatedPPMShipment *models.PPMShipment
+	var err error
+	err = appCtx.NewTransaction(func(txnAppCtx appcontext.AppContext) error {
+		updatedMTOShipment, err = h.mtoShipmentUpdater.UpdateMTOShipmentCustomer(txnAppCtx, mtoShipment, params.IfMatch)
+		if err != nil {
+			return err
+		}
+
+		updatedPPMShipment, err = h.ppmShipmentUpdater.UpdatePPMShipmentWithDefaultCheck(txnAppCtx, mtoShipment.PPMShipment, mtoShipment.ID)
```
duncan-truss, line 128: Should we even call this update if the shipment is not of type PPM?  I know we'll get a not found error at least so maybe that's ok for the time being.

pkg/services/ppmshipment/ppm_shipment_updater.go
```diff
@@ -0,0 +1,66 @@
...
+	// }
+
+	updatedPPMShipment := mergePPMShipment(*ppmShipment, oldPPMShipment)
+
+	err = validatePPMShipment(appCtx, *updatedPPMShipment, oldPPMShipment, &oldPPMShipment.Shipment, checks...)
+	if err != nil {
+		return nil, err
+	}
+
+	transactionError := appCtx.NewTransaction(func(txnAppCtx appcontext.AppContext) error {
```
duncan-truss, line 50: I think you are already operating within a transaction here from the handler and only making a single record update, which shouldn't require a transaction to rollback I don't think.

pkg/handlers/internalapi/internal/payloads/payload_to_model.go
```diff
@@ -96,6 +97,48 @@ func MTOShipmentModelFromCreate(mtoShipment *internalmessages.CreateShipment) *m
...
+func UpdatePPMShipmentModel(ppmShipment *internalmessages.UpdatePPMShipment) *models.PPMShipment {
+	if ppmShipment == nil {
+		return nil
+	}
+
+	// Temporarily hard code this value, until we determine this
+	estimatedIncentive := int32(10000)
+
+	ppmModel := &models.PPMShipment{
+		ActualMoveDate:                 (*time.Time)(ppmShipment.ActualMoveDate),
```
duncan-truss, line 109: ActualMoveDate is a date instead of a timestamp so let's treat it like we do ExpectedDepartureDate

pkg/services/ppmshipment/ppm_shipment_updater.go
ronaktruss: I kept ETag and SitException code commented out so that when we add them its slightly easier.

pkg/handlers/internalapi/mto_shipment.go
ronaktruss: My take on this was that we really shouldn't be handling business logic in the handler and having the if check there would fall into that category. If no ppmShipment object is passed in the `updatePPMShipment` checks if its nil and should terminate early.

pkg/services/ppmshipment/ppm_shipment_updater.go
ronaktruss: Given the way that the handler is calling this function right now I think your analysis is correct. 

I kept this as is because it theoretically makes this service object reusable and NewTransactions (at least tentatively) checks if we are already inside of a transaction context and if we are returns the current context instead of creating a new one.

pkg/handlers/internalapi/internal/payloads/payload_to_model.go
ronaktruss: It looks like both the `FmtDatePtrToPopPtr` and `FmtDateTimePtrToPopPtr` are formatting it into a `*time.Time` using `time.Time(*date)`. Let me know if I am missing something.

duncan-truss: That's true they are the same type in Go, I just saw they were grouped with the timestamps.

swagger-def/internal.yaml
```diff
@@ -2392,6 +2392,8 @@ definitions:
         type: string
       shipmentType:
         $ref: '#/definitions/MTOShipmentType'
+      ppmShipment:
+        $ref: '#/definitions/PPMShipment'
```
felipe-lee, line 2396: Could you remove this? Since it was handled in #8214, which has merged now so you can pull master in.

swagger/internal.yaml
```diff
@@ -2575,13 +2577,96 @@ definitions:
...
+        type: string
+        x-nullable: true
+      reviewedAt:
+        format: date-time
+        type: string
+        x-nullable: true
+      approvedAt:
+        format: date-time
+        type: string
+        x-nullable: true
```
felipe-lee, line 2604: The endpoint caller shouldn't be able to update these three as they will be set by the backend based on status changes, so they shouldn't be part of this payload.

swagger-def/internal.yaml
```diff
@@ -2476,7 +2476,7 @@ definitions:
         x-nullable: true
         x-omitempty: false
       pickupPostalCode:
-        description: zip code, international allowed
+        description: zip code
```
felipe-lee, line 2479: Haha, thank you for updating them all, I completely missed this in my own PR 🤦🏼

pkg/services/ppmshipment/validation.go
```diff
@@ -52,3 +53,43 @@ type ppmShipmentValidatorFunc func(appcontext.AppContext, models.PPMShipment, *m
...
+		return &newPPMShipment
+	}
+
+	ppmShipment := *oldPPMShipment
+
+	ppmShipment.ActualMoveDate = services.SetOptionalDateTimeField(newPPMShipment.ActualMoveDate, ppmShipment.ActualMoveDate)
+	ppmShipment.SubmittedAt = services.SetOptionalDateTimeField(newPPMShipment.SubmittedAt, ppmShipment.SubmittedAt)
+	ppmShipment.ReviewedAt = services.SetOptionalDateTimeField(newPPMShipment.ReviewedAt, ppmShipment.ReviewedAt)
+	ppmShipment.ApprovedAt = services.SetOptionalDateTimeField(newPPMShipment.ApprovedAt, ppmShipment.ApprovedAt)
+
```
felipe-lee, line 68: These shouldn't be updateable at this point, I think. At least if I'm understanding this correctly because this is focused on merging the existing model with the requested changes right?

pkg/services/ppmshipment/rules.go
felipe-lee: Awesome, thanks!

pkg/services/ppmshipment/rules_test.go
```diff
@@ -83,7 +96,10 @@ func (suite *PPMShipmentSuite) TestValidationRules() {
 					newPPMShipment: newPPMShipment,
 					oldPPMShipment: nil,
 				},
-				// Add Update Test case here
+				"update": {
+					newPPMShipment: newPPMShipment,
+					oldPPMShipment: &models.PPMShipment{ID: newPPMShipment.ID},
```
felipe-lee, line 101: So, this test case actually highlights a gap in our validation function. Both of the PPM IDs are nil, but they shouldn't be for updates. But since they're equal to each other, the test is passing. Probably need to update the validator function, and then this test case to ensure they have an ID.

pkg/services/ppmshipment/rules_test.go
```diff
@@ -94,47 +110,46 @@ func (suite *PPMShipmentSuite) TestValidationRules() {
...
+					newPPMShipment: models.PPMShipment{ID: id1},
 					oldPPMShipment: nil,
 					verr:           true,
 				},
-				//"update": Add Update Test Case here
-			}
+				"update": {
+					newPPMShipment: models.PPMShipment{ID: id1},
+					oldPPMShipment: &models.PPMShipment{ID: id2},
+					verr:           true,
```
felipe-lee, line 129: Given the comment I put above, a third error case can be if both the new and old have a nil PPM ID.

pkg/services/ppmshipment/rules_test.go
```diff
@@ -94,47 +110,46 @@ func (suite *PPMShipmentSuite) TestValidationRules() {
...
-			destPostalcode := "99999"
-			sitExpected := false
-			shipmentID := uuid.Must(uuid.NewV4())
-
+		expectedTime := time.Now()
+		pickupPostal := "99999"
+		destPostalcode := "99999"
+		sitExpected := false
+		shipmentID := uuid.Must(uuid.NewV4())
+		suite.Run("success", func() {
```
felipe-lee, line 152: I'm curious why this change was made. Like if the intent was to reuse the values, the other subtest is still declaring its own versions of this.

pkg/services/ppmshipment/ppm_shipment_updater_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/transcom/mymove/pkg/apperror"
+	"github.com/transcom/mymove/pkg/models"
+	"github.com/transcom/mymove/pkg/testdatagen"
+	"github.com/transcom/mymove/pkg/unit"
+)
+
+func (suite *PPMShipmentSuite) TestUpdatePPMShipment() {
+	ppmShipmentUpdater := NewPPMShipmentUpdater()
+
+	oldPPMShipment := testdatagen.MakeDefaultPPMShipment(suite.DB())
```
felipe-lee, line 17: Data like this should be created on a per-test basis rather than at this level. I know right now it works, but it could break if any other tests are added that expect it (and it ties to the potential confusion I mentioned in my other comment).

pkg/services/ppmshipment/ppm_shipment_updater.go
```diff
@@ -0,0 +1,67 @@
...
+func (f *ppmShipmentUpdater) UpdatePPMShipmentWithDefaultCheck(appCtx appcontext.AppContext, ppmShipment *models.PPMShipment, mtoShipmentID uuid.UUID) (*models.PPMShipment, error) {
+	return f.updatePPMShipment(appCtx, ppmShipment, mtoShipmentID, f.checks...)
+}
+
+func (f *ppmShipmentUpdater) updatePPMShipment(appCtx appcontext.AppContext, ppmShipment *models.PPMShipment, mtoShipmentID uuid.UUID, checks ...ppmShipmentValidator) (*models.PPMShipment, error) {
+	if ppmShipment == nil {
+		return nil, nil
+	}
+
+	oldPPMShipment, err := models.FetchPPMShipmentFromMTOShipmentID(appCtx.DB(), mtoShipmentID)
```
felipe-lee, line 36: I'm wondering if instead of getting the `mtoShipmentID` passed in, we should get the full mto shipment passed in. This currently operates on the assumption that the regular mto shipment updater will be called beforehand, kind of. Like it does in the sense that it doesn't specifically handle updates to the regular mto shipment model. Given that, the other updater returns the latest version of the shipment, so no need to hit the DB another time. I think it would require updating this query to load the PPMShipment: https://github.com/transcom/mymove/blob/6998928adcebe08dd144b4d4c7e63deec46a87f0/pkg/services/mto_shipment/mto_shipment_updater.go#L261-L272

felipe-lee (commented):
I left some comments, but overall this looks good!

pkg/services/ppmshipment/rules_test.go
ronaktruss: Hmm, so update doesn't actually use the ppmID anymore. We are "inheriting" the value from the previous version found in the DB based on the mtoShipmentID and getting the ppm object associated with it. 

Since the db prevents null ids on create and we aren't using this value for update I don't think this additional test is needed. Let me know if you disagree or I am missing something.

pkg/services/ppmshipment/rules_test.go
ronaktruss: See https://github.com/transcom/mymove/pull/8185#discussion_r818748402

pkg/services/ppmshipment/validation.go
ronaktruss: Ah good catch, forgot to remove this section with the removal of the fields from the yaml file.

pkg/handlers/internalapi/internal/payloads/payload_to_model.go
```diff
@@ -155,6 +195,8 @@ func MTOShipmentModelFromUpdate(mtoShipment *internalmessages.UpdateShipment) *m
 		model.MTOAgents = *MTOAgentsModel(&mtoShipment.Agents)
 	}
 
+	model.PPMShipment = UpdatePPMShipmentModel(mtoShipment.PpmShipment)
```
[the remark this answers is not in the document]
ronaktruss: Since we are calling this in the mto_shipment handler we may be updating the mtoShipment model so the shipment we assign may not be up to date when we get to the ppm_shipment_updater where we are explicitly fetching this info from the DB.

For all practical purposes this probably isn't an issue, but this is an issue say if we update the mtoShipment to a shipmentType of anything other than PPM and then try and update the ppmShipment info.

I don't have a good sense of if we are adding other validations that utilize this mtoShipment information or if the example above is even possible through UI interactions, so it may actually be worth making the change. 

Summary: I would probably lean away from making this change because the information may not be up to date, but open to being swayed to make this change if we don't expect to use the mtoShipment info for other validations.

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -551,6 +549,30 @@ func (suite *HandlerSuite) getUpdateMTOShipmentParams(originalShipment models.MT
 	return params
 }
 
+func (suite *HandlerSuite) getUpdateMTOShipmentParamsForHHGAndPPM(originalShipment models.MTOShipment) mtoshipmentops.UpdateMTOShipmentParams {
+	serviceMember := testdatagen.MakeDefaultServiceMember(suite.DB())
```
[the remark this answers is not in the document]
ronaktruss: Yeah, right now this is mimicking the logic that is being for the `getUpdateMTOShipmentParams` function that is used in the existing tests. I did initially have plans to try and refactor this logic but decided against it because of time constrains. 

I just spent a some time looking into this and it doesn't look like there is a good way to fetch the serviceMember info but we could potentially pass the information but it would likely have to go through quite a few layers (mtoShipment -> move -> orders -> serviceMember).

pkg/handlers/internalapi/internal/payloads/payload_to_model.go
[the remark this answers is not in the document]
felipe-lee: I don't think there's a way to change the shipment type. That and this change I mentioned is more to be able to access the data in both directions, rather than only one-sided. I don't think it should really have a negative effect to add it, but am fine leaving it out and only adding it if needed. Mainly brought it up because it came up as a thing we needed for the create ppm service object, but that service object also works differently from the update one.

pkg/handlers/internalapi/mto_shipment_test.go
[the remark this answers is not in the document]
felipe-lee: Yeah, it's a few layers. That's why I was saying maybe we could flip the order of things. Like in the test, call this function to create the service member, then pass the service member to `MakeMinimalDefaultPPMShipment` and to this `getUpdateMTOShipmentParamsForHHGAndPPM` function. That way both use the same service member. It's not a huge deal right now, but it's not really mimicking the behavior of the system we want. Funnily enough though, it is highlighting the auth issue we talked about at the BE checkin.

pkg/services/ppmshipment/rules_test.go
felipe-lee: I think in that case then, the check should probably be removed from the list of checks for the updater and the logic for `update` changes should be removed from the check ppm id function. Like I'm fine with either way (adding the extra logic or removing the check), but I don't think it should stay as is because it's incorrect and only halfway there.

pkg/handlers/internalapi/mto_shipment_test.go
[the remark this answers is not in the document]
felipe-lee: We talked about this and this will be done in  a followup PR to avoid expanding this one even more.

pkg/handlers/internalapi/internal/payloads/payload_to_model.go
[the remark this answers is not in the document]
felipe-lee: We talked about this and will leave it out

felipe-lee (approved):
Awesome, thanks for making those updates. I think this is the only comment we didn't end up talking about:
https://github.com/transcom/mymove/pull/8185#discussion_r818271126

But I'm fine with that being discussed later as we're going to end up refactoring how the service objects and handlers interact anyways.

pkg/handlers/internalapi/internal/payloads/payload_to_model.go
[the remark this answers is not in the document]
ronaktruss: I added this refactor in my [last commit](https://github.com/transcom/mymove/pull/8185/commits/a0b0a634e3e92bdf3deb8521a2d38752495618af)

PR #8190, opened 2022-02-18
MB-10727 Add Destination Type to TOO Move Details

leia-sefkin:
## Summary

This work adds destination type to the TOO move details page and shipment form. Destination type is required for retirees and separatees who have HHG or NTS-R shipments only. 

I followed pretty much the same pattern as previous destination type work, although the components for TOO are setup a little differently. I also did some cleanup in the process. 

Left to do

- Look at how testdata is generated: I made some changes in the testdata files to add destination type moves for the TOO, but NTS-R shipment creation isn't working, so that at min needs fixing. 
- Check unit and e2e tests to make sure they're working and covering what's needed (I'm not confident the shipment form tests are actually looking at what we want to)
- Double check storybook (I don't think anything should be affected here but better safe)
- Make sure the move details notification count updates appropriately when destination type is missing on a shipment (and that it's updated again after the info is added) 

Screen Shot 2022-02-18 at 12 51 30 PM

<details>

1. Populate the test data. With this you should see Separatee and Retiree moves in the TOO queue locally

```sh
make dev_db_fresh
make db_dev_e2e_populate
```
```sh
make client_run
```

```sh
make server_run
```

</details>

### Frontend

## Screenshots

image


pkg/testdatagen/scenario/shared.go
```diff
@@ -1746,6 +1746,63 @@ func createHHGWithPaymentServiceItems(appCtx appcontext.AppContext, primeUploade
 	logger.Info(fmt.Sprintf("New payment request with service item params created with locator %s", move.Locator))
 }
 
+// A generic method
```
leia-sefkin, line 1749: Here I am again attempting to create more generic methods we can work with rather than repeating ourselves. I'm not 100% confident it works as expected though, as NTS-R moves don't seem to be working. May need some more tweaking

src/components/Office/RequestedShipments/RequestedShipments.jsx
```diff
@@ -29,11 +30,6 @@ const showWhenCollapsedWithGHCPrime = {
   HHG_OUTOF_NTS_DOMESTIC: ['ntsRecordedWeight', 'serviceOrderNumber', 'tacType'],
 };
 
-const errorIfMissing = {
```
leia-sefkin, line 32: this is now passed in via move details

src/pages/Office/MoveDetails/MoveDetails.jsx
```diff
@@ -36,8 +37,8 @@ const sectionLabels = {
 };
 
 const errorIfMissing = {
```
leia-sefkin, line 39: combined final object from requestedShipments and what was originally here

Ryan-Koch (approved):
I haven't looked at the test datagen part yet, but did add a client test and e2e test to give us some coverage there. I'll take a quick peek asyncronously at the NTS thing, I suspect it has something to do with it needing `usesExternalVendor` value on the shipment and the `selectedMoveType` for the move. If it's more complicated than that we may want to split this part off.

Ryan-Koch:
> I haven't looked at the test datagen part yet, but did add a client test and e2e test to give us some coverage there. I'll take a quick peek asyncronously at the NTS thing, I suspect it has something to do with it needing `usesExternalVendor` value on the shipment and the `selectedMoveType` for the move. If it's more complicated than that we may want to split this part off.

Added in that info so I think the stuff being populated in the sub scenario is good now.

PR #8203, opened 2022-02-23
[MB-10957] Rename duty station to location in APIs

mboren:
## Summary
This PR updates the duty station fields on most of the remaining models to the new duty location phrasing.

This includes backend changes, frontend changes, API changes. Basically everything except the database changes.

Sorry if it's a bit painful to review.

The biggest things changed are Orders and ServiceMembers.
Renamed fields:
- Orders.OriginDutyStation -> Order.OriginDutyLocation
- Order.DestinationDutyStation -> Order.DestinationDutyLocation (also referred to as `newDutyLocation` in some places)
- ServiceMember.DutyStation -> ServiceMember.DutyLocation (also referred to as `current_station`/`current_location` in the frontend)

I undid some changes from #8128 . That PR removed some fields from the DutyLocationPayload that weren't being used, but they were only unused because we were still using DutyStationPayload for them. Now that I've phased out the duty station version, we need all the fields.
I also had to change the way that payload was generated so that if we have a zero valued address, it turns into a nil. This helps the swagger validation, which only checks if pointers are nil and doesn't check if their contents are zero valued.
I'm not super confident that I came up with the best solution here. The address model is used in a few other places in that API, so those will be affected by this change.

The story suggests refactoring the Orders objects out for each API into a shared definition, but the different orders definitions seemed different enough that this wasn't going to be trivial. This PR is already getting huge and running late, so I've decided to punt on refactoring the orders.

## Setup to Run Your Code
This touches so many things, I think we have to mostly rely on automated tests to cover it.
```
make e2e_test
make client_test
make server_test
```

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps
I think it's worth going through the basic flow of creating a customer, updating as TOO, creating a payment request, and reviewing the payment request as the TIO.

### Frontend

### Backend


pkg/handlers/internalapi/duty_locations.go
```diff
@@ -19,13 +21,13 @@ func payloadForDutyLocationModel(location models.DutyLocation) *internalmessages
 		return nil
 	}
 	payload := internalmessages.DutyLocationPayload{
-		ID:                     handlers.FmtUUID(location.ID),
-		CreatedAt:              handlers.FmtDateTime(location.CreatedAt),
-		UpdatedAt:              handlers.FmtDateTime(location.UpdatedAt),
-		Name:                   swag.String(location.Name),
-		Affiliation:            location.Affiliation,
-		AddressID:              handlers.FmtUUID(location.AddressID),
-		TransportationOfficeID: handlers.FmtUUIDPtr(location.TransportationOfficeID),
```
mboren, line 28: i don't think i meant to remove this field, will double check

pkg/handlers/internalapi/internal/payloads/model_to_payload.go
```diff
@@ -17,6 +17,9 @@ func Address(address *models.Address) *internalmessages.Address {
 	if address == nil {
 		return nil
 	}
+	if *address == (models.Address{}) {
+		return nil
+	}
```
mboren, line 22: this is kind of a weird one and deserves a comment and probably extra scrutiny since it affects all Address payloads in the internal API.

pkg/handlers/internalapi/duty_stations.go
```diff
@@ -2,39 +2,14 @@ package internalapi
...
-		ID:          handlers.FmtUUID(station.ID),
-		CreatedAt:   handlers.FmtDateTime(station.CreatedAt),
-		UpdatedAt:   handlers.FmtDateTime(station.UpdatedAt),
-		Name:        swag.String(station.Name),
-		Affiliation: station.Affiliation,
-		AddressID:   handlers.FmtUUID(station.AddressID),
-		Address:     payloads.Address(&station.Address),
-	}
-
-	payload.TransportationOffice = payloads.TransportationOffice(station.TransportationOffice)
```
mboren, line 33: the duty location type doesnt have the full transportation office, just the ID. is this OK? did we use it anywhere? Perhaps when displaying contact info?

mboren: yep, we do need it to display contact info

pkg/handlers/internalapi/internal/payloads/model_to_payload.go
Ryan-Koch: Hm this is interesting. Was there a case where getting the zero value for `models.Address` instead of a `nil` was causing an issue?

mboren: Yeah, I meant to go back and explain this and forgot, oops. 

Sometimes we return an address as part of the `DutyLocationPayload` and sometimes we don't.

`internalmessages.DutyLocationPayload` has a `*Address` field that is optional.
Since it's optional, if swagger thinks the address is present, it will run validations on it, but if it's not present, it will just skip it.
Swagger only thinks pointer fields are missing if they are `nil`, it doesn't check to see if thing it points to is zero valued. On our end, we were sending a pointer to a zero-valued `Address`. Swagger would run validations on that, which would fail.

After writing this out, I think we should probably have different payloads or something here if the address is only included sometimes. Or maybe we should make the address field required and always return it.

src/components/Office/ShipmentContainer/ShipmentContainer.stories.jsx
```diff
@@ -86,13 +86,15 @@ export const MTOAccessorial = () => (
         state: 'CA',
         postalCode: '94535',
       })}
-      originDutyStation={object('Order.originDutyStation', {
+      originDutyLocation={object('Order.originDutyLocation', {
+        // TODO
```
Ryan-Koch, line 90: I see there's a couple of TODO comments here and in the next block. Do we need them?

src/pages/MyMove/CreateOrEditMtoShipment.jsx
```diff
@@ -47,7 +47,7 @@ export class CreateOrEditMtoShipment extends Component {
           <DateAndLocation
             mtoShipment={mtoShipment}
             serviceMember={serviceMember}
-            destinationDutyLocation={orders.new_duty_station}
+            destinationDutyLocation={orders.new_duty_location} // TODO
```
Ryan-Koch, line 50: Do we need this TODO comment? If so could we add a comment elaborating on it?

src/pages/Office/MoveAllowances/MoveAllowances.jsx
```diff
@@ -91,10 +91,10 @@ const MoveAllowances = () => {
     } = values;
     const body = {
       issueDate: order.date_issued,
-      newDutyStationId: order.destinationDutyStation.id,
+      newDutyStationId: order.destinationDutyLocation.id, // TODO
```
Ryan-Koch, line 94: Same question as the other spots as far as the TODO here goes.

Ryan-Koch (approved):
Left a couple of nit comments. LGTM!

cosburn (approved):

mboren:
This branch should work fine locally with a clean database, but it won't work for old stuff in our deployed environments until after #8227 is deployed.

src/components/Office/ShipmentContainer/ShipmentContainer.stories.jsx
mboren: Nope, I'll remove these

PR #8184, opened 2022-02-17
MB-11553 Update mto shipment creation handler to be able to create PPM shipments

NamibiaTorres:
## Summary

Is there anything you would like reviewers to give additional scrutiny?

[this article](tbd) explains more about the approach used.

## Setup to Run Your Code

<details>
<summary>💻 You will need to use three separate terminals to test this locally.</summary>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the
2. Login as a
3.

## Verification Steps for Author

These are to be checked by the author.

- [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
- [ ] Request review from a member of a different team.
- [ ] Have the Jira acceptance criteria been met for this change?

## Verification Steps for Reviewers

These are to be checked by a reviewer.

### Frontend

- [ ] User facing changes have been reviewed by design.
- [ ] There are no aXe warnings for UI.
- [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
- [ ] There are no new console errors in the browser devtools
- [ ] There are no new console errors in the test output
- [ ] If this PR adds a new component to Storybook, it ensures the component is fully responsive, OR if it is intentionally not, a wrapping div using the `officeApp` class or custom `min-width` styling is used to hide any states the would not be visible to the user.

### Backend

- [ ] Code follows the guidelines for [Logging](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide#logging)
- [ ] The requirements listed in [Querying the Database Safely](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide/#querying-the-database-safely) have been satisfied.

#### Any new migrations/schema changes:

- [ ] Follows our guidelines for [Zero-Downtime Deploys](https://transcom.github.io/mymove-docs/docs/dev/contributing/database/Database-Migrations#zero-downtime-migrations)
- [ ] Have been communicated to #g-database
- [ ] Secure migrations have been tested following the instructions in our [docs](https://transcom.github.io/mymove-docs/docs/dev/contributing/database/Database-Migrations#secure-migrations)

PR #8206, opened 2022-02-24
MB-11553 Add PPM shipment to mto shipments payload

felipe-lee:
## Summary

This PR adds PPM shipment to the MTO shipment payload.

Other changes:

* Added several helper functions to `pkg/handlers/apitests.go` to ease comparisons between models and payloads.
* Also edited at least one existing function in the `apitests.go` file to use the built-in date/datetime comparison rather than formatting things as strings and then comparing them.
* Added a testdatagen function to give back a date without time info. This was needed because of the change to date/datetime comparisons mentioned in the previous bullet.
* Updated PPM testdatagen creation functions to link the ppm shipment to the mto shipment in place. This is mainly needed if you want to be able to reverse the relationship without doing another db query.

## Setup to Run Your Code

<details>
<summary>💻 Setup</summary>

### Populate DB

You can skip this if you've already generated data with a PPM shipment.

```sh
make db_dev_fresh
```

### Server

Start the server

```sh
make server_run
```

### Client

Start the UI locally.

```sh
make client_run
```

</details>

### Additional steps

1. Go to the [local client page](http://milmovelocal:3000/)
2. Login as a client that has a PPM shipment
3. Open the browser dev tools
4. Go to the network tab and refresh the page
5. Look at the /mto-shipments network call and check the response payload. It should have the PPM info.

This is only part of the work for the ticket. The rest of the work will be in #8184


pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -798,20 +802,52 @@ func (suite *HandlerSuite) TestListMTOShipmentsHandler() {
...
 		firstShipmentReturned := okResponse.Payload[0]
 		secondShipmentReturned := okResponse.Payload[1]
+		thirdShipmentReturned := okResponse.Payload[2]
 
 		// we expect the shipment that was created first to come first in the response
 		suite.Equal(subtestData.shipments[0].ID.String(), firstShipmentReturned.ID.String())
 		suite.Equal(subtestData.shipments[1].ID.String(), secondShipmentReturned.ID.String())
+		suite.Equal(subtestData.shipments[2].ID.String(), thirdShipmentReturned.ID.String())
+
+		suite.NoError(okResponse.Payload.Validate(strfmt.Default))
```
duncan-truss, line 816: Yay thanks for remembering to validate responses.  Can we also call validate on the request `subtestData.params.Body.Validate(strfmt.Default)` as well?

swagger-def/internal.yaml
```diff
@@ -2413,6 +2415,149 @@ definitions:
...
+      expectedDepartureDate:
+        description: >
+          Date the customer expects to move.
+        format: date
+        type: string
+      actualMoveDate:
+        format: date
+        type: string
+        x-nullable: true
+        x-omitempty: false
```
ronaktruss, line 2460: I'm not sure I understand why we are using `x-omitempty: false` for some of these fields. Eg. For `actualMoveDate` what does having this do? What does sending "" over not sending anything let us do? Are we using this distinction at some other point in the code?

ronaktruss (commented):
LGTM! See comments about `x-omitempty`.

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -798,20 +802,52 @@ func (suite *HandlerSuite) TestListMTOShipmentsHandler() {
...
 
 		for i, returnedShipment := range okResponse.Payload {
 			expectedShipment := subtestData.shipments[i]
 
 			suite.Equal(expectedShipment.Status, models.MTOShipmentStatus(returnedShipment.Status))
 
+			if expectedShipment.ShipmentType == models.MTOShipmentTypePPM {
```
duncan-truss, line 823: I was wondering how the PPMShipment association got loaded without change any service query and then I see the handler is using this older code to eager load all associations on the mto shipment model.  We should try to refactor this to use the mto_shipment_fetcher.go in the future when we get a chance.

https://github.com/transcom/mymove/blob/73450b285c7c1ecf0b4754af2a8cbdaacad7b2d6/pkg/handlers/internalapi/mto_shipment.go#L193

swagger/internal.yaml
```diff
@@ -2417,6 +2419,153 @@ definitions:
...
+        format: date-time
+        type: string
+        x-nullable: true
+        x-omitempty: false
+      pickupPostalCode:
+        description: zip code, international allowed
+        format: zip
+        type: string
+        title: ZIP
+        example: '''90210'' or ''N15 3NL'''
```
duncan-truss, line 2482: We aren't accepting that second example and in fact we can add a `pattern` property for these postal codes to match the frontend validation of 5 digit only `/^(\d{5})$/`

duncan-truss (approved):
Just left some minor cleanup suggestions

pkg/handlers/internalapi/mto_shipment_test.go
felipe-lee: Yeah I'll try adding that

swagger-def/internal.yaml
felipe-lee: It's so that the payload remains consistent in what keys are present. We aren't currently using it much in the internal api, but it's something that started being done in the prime api. This is likely something that also needs an ADR

ronaktruss (approved):

swagger/internal.yaml
felipe-lee: Ah good point, I'll update this.

pkg/handlers/internalapi/mto_shipment_test.go
felipe-lee: Ah good point. I'll add a ticket for this if there isn't one already since it would likely require more work to ensure all the data we use on the FE is still being provided. I'm not confident our tests as they are actually check for all the data we expect, since even though we're adding the validation of the payload, none of the related models are marked as required in the swagger file.

swagger-def/internal.yaml
```diff
@@ -2413,6 +2415,153 @@ definitions:
...
+        format: date-time
+        type: string
+        x-nullable: true
+        x-omitempty: false
+      reviewedAt:
+        format: date-time
+        type: string
+        x-nullable: true
+        x-omitempty: false
+      pickupPostalCode:
```
ronaktruss, line 2471: @felipe-lee I think we are missing approvedAt

pkg/handlers/internalapi/mto_shipment_test.go
felipe-lee: Hmm, I was looking into this but i'm not 100% sure how to do this. The `subtestData.params.Body` is a reader and doesn't have any validation on it. It looks like `params` itself does have a way to validate things, but it requires having both the request object and the matched route. I have the request object, but the matched route object is the one I'm not sure how to get. There's a few functions that can return it, but they all work using a wrapper around the regular go request context. I think it's set up in middleware somewhere, but I'm not 100% sure where because our middleware doesn't do it afaict.

swagger-def/internal.yaml
felipe-lee: Ah, good catch, I'll add that.

felipe-lee: Fixed!

PR #8209, opened 2022-02-24
MB-11596 : Add cypress test for PPM booking -dates and location page

pearl-truss:
## Summary

This PR adds cypress test for the PPM booking for the dates and location form.

To review, check that the flows covered in the test make sense and call out anything that's missing


pearl-truss:
@jenniferlynparsons @YanZ777 I believe you both have experience with cypress testing so I've added you to this PR. Also, do you know if the happo diffs are okay to accept?

pearl-truss:
> @jenniferlynparsons @YanZ777 I believe you both have experience with cypress testing so I've added you to this PR. Also, do you know if the happo diffs are okay to accept?

I followed this thread and accepted the happo changes https://ustcdp3.slack.com/archives/CP6PTUPQF/p1646158575883389

cypress/integration/milmove/ppms/dateAndLocation.js
```diff
@@ -0,0 +1,117 @@
...
+  });
+
+  it('doesn’t allow SM to progress if date field is invalid"', () => {
+    // profile@co.mple.te
+    const userId = '3b9360a3-3304-4c60-90f4-83d687884070';
+    cy.apiSignInAsUser(userId);
+    resumePPM();
+    invalidDate();
+  });
+
```
YanZ777, line 41: I'd probably roll all these "don't allow to progress if..." into one e2e test since in my mind the user isn't going to log out and log in and try each field and get an error.

i.e.
resume PPM 
try invalid input for the postal code
put in valid input
try invalid input for the date field
put in valid input
etc. 

But I don't necessarily mind it this way either.

cypress/integration/milmove/ppms/dateAndLocation.js
```diff
@@ -0,0 +1,117 @@
...
+    cy.prepareCustomerApp();
+  });
+
+  beforeEach(() => {
+    cy.logout();
+  });
+
+  it('can submit a PPM move', () => {
+    // profile@comple.te
+    const userId = '3b9360a3-3304-4c60-90f4-83d687884070';
```
YanZ777, line 12: I'd be ok if you put just made the `userId` a top level field so that the tests have access to it.

duncan-truss:
I think it's a little confusing that we now have `milmove` and `mymove` categories, granted you're usually only using this when creating or editing a new test
Screen Shot 2022-03-01 at 12 53 14 PM

cypress/integration/milmove/ppms/dateAndLocation.js
```diff
@@ -0,0 +1,117 @@
...
+    resumePPM();
+    missingSecondaryZip();
+  });
+});
+
+function customerChoosesAPPMMove() {
+  cy.get('button[data-testid="shipment-selection-btn"]').click();
+  cy.nextPage();
+
+  cy.get('input[type="radio"]').eq(1).check({ force: true });
```
duncan-truss, line 55: This selector could be `input[value="PPM"]` if you wanted to get a single match and make the intent a little clearer.

cypress/integration/milmove/ppms/dateAndLocation.js
```diff
@@ -0,0 +1,117 @@
...
+
+  cy.location().should((loc) => {
+    expect(loc.pathname).to.match(/^\/moves\/[^/]+\/shipments\/[^/]+\/estimated-weight/);
+  });
+}
+
+function inputsInvalidPostalCodes() {
+  cy.get('[class="usa-error-message"]').should('not.exist');
+  cy.get('input[name="pickupPostalCode"]').clear().type('00000').blur();
+  cy.get('[class="usa-error-message"]').should('exist');
```
duncan-truss, line 80: I know this was even a bit tricky in react testing library when I did it, how hard is it to tell that the error is for the expected field and do you think it's worth verifying the error message?

cypress/integration/milmove/ppms/dateAndLocation.js
```diff
@@ -0,0 +1,117 @@
...
+
+  it('can submit a PPM move', () => {
+    // profile@comple.te
+    const userId = '3b9360a3-3304-4c60-90f4-83d687884070';
+    cy.apiSignInAsUser(userId);
+    customerChoosesAPPMMove();
+    submitsDateAndLocation();
+  });
+
+  it('doesn’t allow SM to progress if an invalid postal code is provided"', () => {
```
duncan-truss, line 18: Looks like there's a stray double quote at the end of some of these test names.

pearl-truss:
> I think it's a little confusing that we now have `milmove` and `mymove` categories, granted you're usually only using this UI when creating or editing a new test but the folders will still appear similar in the codebase. Screen Shot 2022-03-01 at 12 53 14 PM

I'll defer to @gidjin on this, he created the PR that I pulled in to create the new milmove category

pearl-truss:
@duncan-truss @YanZ777 thanks for the reviews! I just pushed a commit that addresses all the comments left

cypress/integration/milmove/ppms/dateAndLocation.js
```diff
@@ -0,0 +1,86 @@
...
+
+  cy.location().should((loc) => {
+    expect(loc.pathname).to.match(/^\/moves\/[^/]+\/shipments\/[^/]+\/estimated-weight/);
+  });
+}
+
+function invalidInputs() {
+  // invalid date
+  cy.get('input[name="expectedDepartureDate"]').clear().type('01 ZZZ 20222').blur();
+  cy.get('[class="usa-error-message"]').contains('Enter a complete date in DD MMM YYYY format (day, month, year).');
```
duncan-truss, line 54: I think we could define these same selectors as aliases in Cypress https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Elements

Something like `cy.get('[class="usa-error-message"]').as('errorMessage)`

cypress/integration/milmove/ppms/dateAndLocation.js
```diff
@@ -0,0 +1,86 @@
...
+  const userId = '3b9360a3-3304-4c60-90f4-83d687884070';
+  it('can submit a PPM move', () => {
+    cy.apiSignInAsUser(userId);
+    customerChoosesAPPMMove();
+    submitsDateAndLocation();
+  });
+
+  it('doesn’t allow SM to progress if form is in an invalid state', () => {
+    cy.apiSignInAsUser(userId);
+    resumePPM();
```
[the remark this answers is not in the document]
pearl-truss: great point, I just pushed pushed a commit to address

cypress/integration/milmove/ppms/dateAndLocation.js
pearl-truss: Unless I'm missing something I don't think that change would get us very much since it would just be `cy.get('@errorMessage')` instead of `cy.get('[class="usa-error-message"]')` which is only being used a handful of times

duncan-truss: I think it might be slightly more efficient because it's only getting re-queried if the underlying results change like when you got to call `cy.get('@errorMessage').next('input')` it doesn't search the whole DOM again because it already has the reference.  I thought it would just be nice to cut down on the boilerplate but not a blocking change.

duncan-truss (approved):

cypress/integration/milmove/ppms/dateAndLocation.js
pearl-truss: Thanks, that makes sense but I think we do want to requery for error messages since there are different messages depending on where in the test you are

pearl-truss: Oh what, nevermind I see what you're saying wrt to using for the `.next('input')` portion

duncan-truss: Reading the docs it sounds to me like Cypress does re-query when the DOM changes so you don't have to reassign the alias https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Stale-Elements.  Or did you try using the alias for everything and it didn't work?

pearl-truss: Oh okay, I didn't look at the docs in depth. I just pushed a commit that doesnt reassign the alias. And the test still passed locally

PR #8227, opened 2022-03-01
[MB-10957] Copy old duty station ID values to new duty location columns

mboren:
## Summary
This is part of renaming these columns. We've already set up triggers to fill in the 
duty location versions of the columns, and now it's time to copy over all the old versions of the values.

This must get deployed before #8203 .

## Setup to Run Your Code
Before checking out this branch, create devseed data
```sh
git checkout master
make db_dev_e2e_populate
```
Make sure there are some null values for the columns that we're messing with. We have to disable triggers for this because they are going to try to keep the columns synchronized.  
```sql
alter table orders disable trigger all;
UPDATE orders
    SET new_duty_location_id = NULL,
        origin_duty_location_id = NULL;
alter table orders enable trigger all;

alter table service_members disable trigger all;
update service_members
    set duty_location_id = NULL;
alter table service_members enable trigger all;
```
Check out this branch and run migrations
```sh
git checkout mb-MB-10957-copy-values-to-new-columns
make db_dev_migrate
```
Make sure that all the duty location fields are in sync with the duty station fields after the migration. If any of these 3 queries return results, there is a problem.
```sql
select * from orders where origin_duty_location_id is null and origin_duty_station_id is not null;
select * from orders where new_duty_location_id is null and new_duty_station_id is not null;
select * from service_members where duty_location_id is null and duty_station_id is not null;

```


migrations/app/schema/20220228192535_copy_values_to_new_duty_location_cols.up.sql
```diff
@@ -0,0 +1,21 @@
+UPDATE service_members
+SET duty_location_id = duty_location_id
```
leia-sefkin, line 2: looks like the second value here should be `duty_station_id` rather than **location**_id

mboren: whoops! Should be fixed now.

migrations/app/schema/20220228192535_copy_values_to_new_duty_location_cols.up.sql
```diff
@@ -0,0 +1,21 @@
+UPDATE service_members
+SET duty_station_id = duty_location_id
```
leia-sefkin, line 2: ahh so close but this should be reversed to `SET duty_location_id = duty_station_id` otherwise we'll have even more null values 😬

mboren: oof thank you.

leia-sefkin (approved):
queries look good, ran migrations and the data appears to be updated as expected. i agree we're fine to accept the risk of blocking writes to the orders table while this merge is running, but it's worth asking the question of when that risk becomes something we need to address

gidjin (approved):
LGTM, interesting note on blocking writes, but as noted I agree the risk is small given we are not in production. After we have production users we will likely need to time these things at lower traffic times.

I noticed that the columns `origin_duty_location_id`, `duty_location_id`, and `new_duty_location_id` do not have comments. Could you add comments to them or note that the comments on the old fields should be moved if there is a later story to clean up the old columns. I don't see a need to re-review after the comments are added though.

PR #8230, opened 2022-03-02
Explicitly define deps for bin/* targets in Makefile

ahobson:
* This ensures the binaries are rebuild when their dependencies change


rogeruiz (approved):
lgtm 🌈

Thanks for submitting this @ahobson and pairing on this asynchronously.

PR #8233, opened 2022-03-03
Make filter more specific to see if it helps

gidjin:
New build job for integration tests was running with a CircleCI parallelism of 4 but only had one file in the glob. So when using parallelism for the integration tests you need to ensure there are enough files to have at least one per thread.

See https://circleci.com/docs/2.0/parallelism-faster-jobs/


rogeruiz (approved):
lgtm 🌈

pairing with @gidjin and @mboren on this. The fact that there was a parallelism greater than two (>2) caused the jobs to run multiple times with only the `ppm` folder running the first time and them all jobs running on subsequent runs in CircleCI.

PR #8235, opened 2022-03-03
Error if CircleCI parallelism setting is larger than glob in e2e tests

ahobson:
## Summary

If the circleci parallelism for e2e tests is greater than the number of files, then some of the parallel runs will have an empty list of specs to run and thus will run the entire test suite. Change it so that we get an error instead

You can see an [example failure](https://app.circleci.com/pipelines/github/transcom/mymove/39586/workflows/d8e15e82-a268-4a7e-8e67-59eba81e6f68/jobs/609816)


rogeruiz:
This should be fixed by the work done here. https://github.com/transcom/mymove/pull/8233

The problem was the parallelism for the new Cypress tests.

.circleci/config.yml
```diff
@@ -632,6 +632,10 @@ commands:
           # reasonable proxy for test time
           command: |
             SPECS=$(circleci tests glob "<< parameters.spec_glob >>" | circleci tests split --split-by=filesize | paste -s -d, -)
+            if [ -z "${SPECS}" ]; then
+              echo "CircleCI Parallelism greater than number of files in glob"
```
rogeruiz, line 636:
```suggestion
              echo "CircleCI Parallelism greater than number of files in glob. Please adjust the number parallelism value to a lower number."
```

rogeruiz:
It looks like we'll need to get the latest `master` into this branch to get tests to pass. I think this is a great addition. I added a suggestion but don't let that stop us from merging this in.

rogeruiz (approved):
lgtm 🌈

rogeruiz:
@Mergify refresh

PR #8240, opened 2022-03-04
[ MB-11826 ] Fix flaky Cypress tests

rogeruiz:
## Summary

This disables the validation that Cypress uses to check if a user can
click on something. While Cypress debug messages suggest you do this,
this is an anti-pattern when it comes to testing things in Cypress since
a user cannot actually click an input[type=checkbox] because of styling
done by the USWDS. So this actually clicks on the label element instead
which is a sibling of the input element with further checks on the
`for=` attribute. This simulates the way a user would actually click on
things making the tests more aligned with how a user would actually
click on an checkbox in this application.

[this Cypress blog post about testing clicks](https://www.cypress.io/blog/2019/01/22/when-can-the-test-click/) explains more about some of the approaches used in this PR to fix the flaky tests. There are other parts of the Cypress documentation that are being used as well to make other tests less flaky as well.

I think the ~~previous commit with changes to the~~ inputs ~~around~~ need
 `focus` and `blur` events should occur when editing a text input have
done the trick to fix the flakiness. I also ran into some quirks with
out the client app is built when it comes to the Cypress tests being
run. Locally, I was seeing a bunch of things happening where various
attributes on HTML elements were not accurate with the tests.

[=> The docs on this are here](https://docs.cypress.io/api/commands/blur#Usage)

## Setup to Run Your Code

<details>
<summary>💻 You will need to use at least one terminal to test this locally.</summary>

##### Terminal 1

Start the Cypress UI locally.

```sh
make e2e_test
```

##### Headless testing

To test this without the UI, for example repeatedly to test flakiness.

```sh
SPECS=cypress/integration/office/servicescounseling/servicesCounselingFlows.js make e2e_test_docker
```

</details>

### Additional steps

1. Select the `office/txo/tooFlows.js` file to test the `is able to edit allowances` and make sure it passes and clicks the `<label>` element rather than the hidden `<input>` element.
2. Select the `cypress/integration/office/servicescounseling/servicesCounselingFlows.js` file to test the `is able to edit allowances` and make sure it passes repeatedly. 

<!--- more stuff to come later...


YanZ777 (approved):
Looks good to me! I don't know if we want to try running the CircleCI job a few times to see if it flakes or not though.

PR #8226, opened 2022-03-01
[MB-10712] Payment request cache fixes and optimizations

reggieriser:
## Summary

This ticket addresses a few things with payment requests:

- Param lookup caching can result in incorrect pricing in some scenarios.  See this [prior ticket](https://dp3.atlassian.net/browse/MB-9497) as well as [this document](https://docs.google.com/document/d/1dkSVly3TC8GxpqOEutDyUBvQKTcmQHxQ1n8c1ni-wPY/edit#heading=h.92o2nzwiokvt) for more details about this issue.  I ended up just turning the lookup cache off as discussed in [this Slack thread](https://ustcdp3.slack.com/archives/CP6F568DC/p1633465395092500) since the cache seemed to have a pretty minor effect on the number of queries executed.  I added tests (leveraged from a [previous draft PR](https://github.com/transcom/mymove/pull/7524)) to demonstrate the scenario and how turning off this portion of the cache fixes it.
- As part of my evaluation, I managed to find a few places where I could optimize existing queries (by preloading associations, reusing values already looked up, etc.).  In my local testing with a representative payment request, it looked to result in about 25% fewer queries overall.
- I kept the param map caching as is, but I did turn off an out-of-date (and seemingly not that advantageous) manual mapping previously used when the cache was off.

With the benefit of hindsight, I think we could rethink the cache and some of the payment request code to make it more efficient and easier to navigate.  But that's a bigger effort that's outside the scope of this bug ticket.

## Setup to Run Your Code

<details>
<summary>💻 You will need to use three separate terminals to test this locally.</summary>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

There should be no visible changes as a result of this PR.  Ensure tests are passing.  But you can also try out some payment requests and verify that they still succeed and the pricing seems accurate.

## Verification Steps for Author

These are to be checked by the author.

- [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
- [ ] Request review from a member of a different team.
- [ ] Have the Jira acceptance criteria been met for this change?

## Verification Steps for Reviewers

These are to be checked by a reviewer.

### Backend

- [ ] Code follows the guidelines for [Logging](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide#logging)
[...]

PR #8241, opened 2022-03-04
Fix mock issue when running db_dev_e2e_populate

reggieriser:
## Summary

In #8226, I partially turned off the payment request cache because it could lead to incorrect pricing in some scenarios (plus it didn't seem to be having the performance impact we had first thought).  Unfortunately, I didn't realize that we limit mock calls in one spot in some E2E test data setup for a payment request, so `make db_dev_e2e_populate` started to fail in `master`.  I'm not exactly sure the reasoning behind why we have the limits there, but I didn't want to remove all the limits entirely without digging in further.  So, to get us moving again, this PR just increases the limits on two distance calls (when the zip3s match) to be the minimum threshold required for a successful run.

## Setup to Run Your Code

Run `make db_dev_reset db_dev_e2e_populate` and verify that it completes successfully.


mboren (approved):
It fixes the problem for me 🎉. I agree that it seems strange to constrain the number of calls here, it seems like a pricing implementation detail that i wouldn't expect test data generation to care about.

PR #8242, opened 2022-03-04
[MB-10957] Read from new duty location columns

mboren:
## Summary
I updated raw queries and a view that were still using `duty_station_id` columns on `orders` and `service_members` to use the new versions of the columns, and updated them to use the `duty_locations` table instead of the `duty_stations` view.
I also updated models that were still using the `duty_stations` view to use `duty_locations`.
I also did some more superficial renaming of some local variables and comments to the new language.

The goal is to eliminate references to `duty_stations`, and make all reads happen on `duty_location_id` versions of columns, so we can start phasing that stuff out.

## Setup to Run Your Code

<details>
<summary>💻 You will need to use separate terminals to test this locally.</summary>

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps
There are no new features introduced in this PR, so we're just checking to make sure nothing is broken.

1. Create a move for a new customer. Pay particular attention to the order creation step and make sure that nothing weird happens.
2. Make sure that they show up in the appropriate Service Counselor or TOO queue, and that their orders look OK.

### Backend


pkg/services/duty_stations/load_duty_stations.go
```diff
@@ -24,7 +24,7 @@ const (
 	InsertTemplate string = `
 	{{range .}}
 INSERT INTO addresses (id, street_address_1, city, state, postal_code, created_at, updated_at, country) VALUES ('{{.AddressID}}', 'N/A', '{{.Address.City}}', '{{.Address.State}}', '{{.Address.PostalCode}}', now(), now(), 'United States');
-INSERT INTO duty_stations (id, name, affiliation, address_id, created_at, updated_at, transportation_office_id) VALUES ('{{.DutyLocationID}}', '{{.Stations.Name}}', 'MARINES', '{{.AddressID}}', now(), now(), '{{.To.ID}}');
+INSERT INTO duty_locations (id, name, affiliation, address_id, created_at, updated_at, transportation_office_id) VALUES ('{{.DutyLocationID}}', '{{.Stations.Name}}', 'MARINES', '{{.AddressID}}', now(), now(), '{{.To.ID}}');
 	{{end}}`
```
mboren, line 28: This entire file is not really used anymore, and i would be shocked if it worked. I updated it because i wanted to remove all references to `duty_stations`, but i'm not committing to making sure that it runs.

pkg/models/service_member.go
```diff
@@ -62,7 +62,7 @@ type ServiceMember struct {
 	Orders                 Orders                    `has_many:"orders" fk_id:"service_member_id" order_by:"created_at desc" `
 	BackupContacts         BackupContacts            `has_many:"backup_contacts" fk_id:"service_member_id"`
 	DutyLocationID         *uuid.UUID                `json:"duty_station_id" db:"duty_station_id"`
-	DutyLocation           DutyLocation              `belongs_to:"duty_stations" fk_id:"duty_station_id"`
+	DutyLocation           DutyLocation              `belongs_to:"duty_locations" fk_id:"duty_station_id"`
```
mboren, line 65: updating `fk_id` and all the other tags for these fields will come in the next PR.

migrations/app/schema/20220308211442_update_origin_duty_location_to_gbloc_view.up.sql
```diff
@@ -0,0 +1,11 @@
...
+-- and orders.origin_duty_location_id instead of orders.origin_duty_station_id
+CREATE OR REPLACE VIEW origin_duty_location_to_gbloc(id, move_id, gbloc) AS
+SELECT pctg.id,
+       m.id AS move_id,
+       pctg.gbloc
+FROM moves m
+         JOIN orders ord ON ord.id = m.orders_id
+         JOIN duty_locations dl ON dl.id = ord.origin_duty_location_id
+         JOIN addresses a ON a.id = dl.address_id
+         JOIN postal_code_to_gblocs pctg ON a.postal_code::text = pctg.postal_code::text;
```
mboren, line 11: here's the original definition for this view, to compare with:
```
marty@martys-MacBook-Pro-2 mymove % psql-dev
psql (12.7)
Type "help" for help.

dev_db=# \sv origin_duty_location_to_gbloc
CREATE OR REPLACE VIEW public.origin_duty_location_to_gbloc AS
 SELECT pctg.id,
    m.id AS move_id,
    pctg.gbloc
   FROM moves m
     JOIN orders ord ON ord.id = m.orders_id
     JOIN duty_stations ds ON ds.id = ord.origin_duty_station_id
     JOIN addresses a ON a.id = ds.address_id
     JOIN postal_code_to_gblocs pctg ON a.postal_code::text = pctg.postal_code::text
```

leia-sefkin (changes requested):
Overall the changes look good, and I was able to make a move and see it in the TOO queue, and fuss around with Duty Location in various spots with no issues. 

One problem I did see is an error in console that the Customer form is expecting a dutyStationName prop and getting "undefined", where it is required. This is after the initial contact info is submitted, before uploading orders when visiting the review page. I've attached the error I saw. The page is loading ok without this, and can understand we want to minimize the scope of these changes, but seems worth looking at before shipping the code. 

Screen Shot 2022-03-10 at 2 42 47 PM

mboren:
Thanks for catching that, I'll look into it.

mboren:
I was able to reproduce this with non-installation locations.
I don't think I introduced this issue, but I'm not 100% sure yet, still looking into it.
I'm seeing the same console error locally on master. I don't see the console error on staging, but when I look in the devtools, I see the same problematic props are undefined on staging, so maybe we're just suppressing the error somehow? IDK if that's a thing.
image

leia-sefkin (approved):

leia-sefkin:
after discussing with team we determined the console issue is due to something introduced prior to this work and fixing it will fall into another task. therefore - hitting approval as everything else looks good to 🚢

Ryan-Koch (approved):

PR #8216, opened 2022-02-26
MB-10675 Add EstimatedWeightsProGear page

felipe-lee:
## Summary

This PR adds the Estimated Weights and Pro Gear page, implementing the previously created `EstimatedWeightsProGearForm` component.

I added a testdatagen function that will create an unsubmitted move with a PPM that has filled in the weights info.

I also made a few of the testdatagen concrete type -> pointer functions be exported so they can be used in scenarios and tests.

## Setup to Run Your Code

<details>
<summary>💻 Setup </summary>

### Populate DB

You can skip this if you've already generated data with a customer that can add shipments, or if you want to do it from scratch.

```sh
make db_dev_fresh
```

### Server

Start the server

```sh
make server_run
```

### Client

Start the UI locally.

```sh
make client_run
```

</details>

### Additional steps

1. Go to the [local client page](http://milmovelocal:3000/)
    1. Alternatively, you can use the ephemeral deploy https://my-milmove-pr-8216.mymove.sandbox.truss.coffee
2. Login as a customer that hasn't submitted their move but has a ppm shipment. There are for sure two available if you use the devseed data, emails: `dates_and_locations@ppm.unsubmitted` and `estimated_weights@ppm.unsubmitted`. 
3. Select the PPM shipment
4. You'll see the dates and locations page.
5. Click "Save & Continue".
6. Now you should see the new estimated weight and pro gear page. Test it out as you see fit.
7. Clicking "Save & Continue" won't fully work at the moment because the next page is coming in a separate PR and the BE work to get the updates for PPMs in is in #8185. 

## Verification Steps for Author

These are to be checked by the author.

- [ ] Have the Jira acceptance criteria been met for this change?

## Verification Steps for Reviewers

These are to be checked by a reviewer.

### Frontend

- [ ] User facing changes have been reviewed by design.
- [ ] There are no aXe warnings for UI.
- [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
- [ ] There are no new console errors in the browser devtools
- [ ] There are no new console errors in the test output

## Screenshots

Desktop:
image

Mobile:
image
image
image

PR #8251, opened 2022-03-08
MB-11597: Cypress test for estimated weights and  progear

pearl-truss:
## Summary

This should be merged after https://github.com/transcom/mymove/pull/8216/. And I'm temporarily pointing this PR to MB-10675 so that it's easier to review and see which file changes are particular to this PR


cypress/integration/milmove/ppms/estimatedWeightsProgear.js
```diff
@@ -0,0 +1,76 @@
...
+    customerChoosesAPPMMove();
+    submitsDateAndLocation();
+    invalidInputs();
+  });
+
+  it('can continue to next page', () => {
+    cy.apiSignInAsUser(userId);
+    // submitsDateAndLocation();
+    submitsEstimatedWeights();
+  });
```
felipe-lee, line 25: Should there also be another happy path where the user submits with pro-gear weights? Since right now that part is only in the unhappy path.

src/components/Customer/PPMBooking/EstimatedWeightsProGearForm/EstimatedWeightsProGearForm.jsx
```diff
@@ -38,10 +38,10 @@ const validationSchema = Yup.object().shape({
 
 const EstimatedWeightsProGearForm = ({ orders, serviceMember, mtoShipment, onSubmit, onBack }) => {
   const initialValues = {
-    estimatedWeight: mtoShipment.ppmShipment.estimatedWeight?.toString() || '',
-    hasProGear: mtoShipment.ppmShipment.hasProGear?.toString() || 'false',
-    proGearWeight: mtoShipment.ppmShipment.proGearWeight?.toString() || '',
-    spouseProGearWeight: mtoShipment.ppmShipment.spouseProGearWeight?.toString() || '',
+    estimatedWeight: mtoShipment?.ppmShipment?.estimatedWeight?.toString() || '',
```
felipe-lee, line 41: I've fixed this in my branch so you should be able to pull in the latest changes and this diff will go away.

cypress/integration/milmove/ppms/estimatedWeightsProgear.js
```diff
@@ -0,0 +1,76 @@
...
+  before(() => {
+    cy.prepareCustomerApp();
+  });
+
+  beforeEach(() => {
+    cy.logout();
+  });
+
+  // estimated_weights@ppm.unsubmitted
+  const userId = '13f3949d-0d53-4be4-b1b1-ae4314793f34';
```
felipe-lee, line 13: I hadn't realized your branch was based on mine when I replied to [your comment in my PR](https://github.com/transcom/mymove/pull/8216#discussion_r821812372). But now knowing that, I think you can merge the latest changes already then and start calling that new testdatagen function from the e2ebasic file.

felipe-lee (commented):
LGTM so far! just left a few comments

cypress/integration/milmove/ppms/estimatedWeightsProgear.js
pearl-truss: Sorry, I should've called that out! And thanks, I called the new function in teh e2ebasic file but I did have to update the subscenario to generate an access code to prevent the 401 error from happening when logging in with cypress

cypress/integration/milmove/ppms/estimatedWeightsProgear.js
pearl-truss: Yeah that makes since, the latest push includes this

pkg/testdatagen/scenario/subscenarios.go
```diff
@@ -95,6 +95,12 @@ func subScenarioPPMOnboarding(appCtx appcontext.AppContext, userUploader *upload
 		createTXOUSMC(appCtx)
 
 		// Onboarding
+		testdatagen.MakeAccessCode(appCtx.DB(), testdatagen.Assertions{
+			AccessCode: models.AccessCode{
+				Code:     "ABC123",
+				MoveType: models.SelectedMoveTypePPM,
+			},
+		})
```
felipe-lee, line 103: Nice, I didn't realize this was needed, thanks for adding it.

cypress/integration/milmove/ppms/estimatedWeightsProgear.js
felipe-lee: Thanks for adding that. I'm wondering if there's value in having both as separate cases though. Like one case that only tests putting in an estimated weight, and then another case that tests both estimated weight and pro gear weights. I'm a bit between wanting the separate things and being good with a single case that tests it all like it is now, I'll leave it up to you.

cypress/integration/milmove/ppms/estimatedWeightsProgear.js
```diff
@@ -0,0 +1,90 @@
...
+  cy.get('input[name="hasProGear"][value="true"]').check({ force: true });
+  cy.get('input[name="proGearWeight"]').type(500).blur();
+  cy.get('button').contains('Save & Continue').should('be.enabled');
+
+  // TODO: once the Estimated Weights Page is updated to be able to navigate to the next page
+  // uncomment out the lines below
+  //   cy.get('button').contains('Save & Continue').click();
+  //   cy.location().should((loc) => {
+  //     expect(loc.pathname).to.match(/^\/moves\/[^/]+\/shipments\/[^/]+\/estimated-incentive/);
+  //   });
```
felipe-lee, line 89: I think this should be able to work now, because the estimated incentive work is in this branch now.

felipe-lee (approved):
LGTM, left a few comments, but only the one about uncommenting the estimated incentive url check is really needed before merging (well that and the other pr merging haha)

cypress/integration/milmove/ppms/dateAndLocation.js
```diff
@@ -83,3 +62,24 @@ function invalidInputs() {
...
+}
+
+function submitsDateAndLocation() {
+  cy.get('input[name="pickupPostalCode"]').clear().type('90210').blur();
+
+  cy.get('input[name="destinationPostalCode"]').clear().type('76127');
+  cy.get('input[name="expectedDepartureDate"]').clear().type('01 Feb 2022').blur();
+
+  cy.get('button').contains('Save & Continue').click();
+
```
duncan-truss, line 81: In other Cypress tests we define the API requests as cy.intercept() and call cy.wait() on them.  What do you think about adding them into this file and the estimated weights file?  I know sometimes those are needed to avoid timeouts but it also works as an assert that the API was called.

duncan-truss, line 81: In other Cypress tests we define the API requests as cy.intercept() and call cy.wait() on them.  What do you think about adding them into this file and the estimated weights file?  I know sometimes those are needed to avoid timeouts but it also works as an assert that the API was called.

duncan-truss (approved):
Thanks for adding those API intercepts in

PR #8254, opened 2022-03-08
[MB-11486] Add assertions around details on customer reviewing NTS shipment info

hannaliebl:
## Summary

I could be wrong, but I think the NTS shipment customer tests already covered basically everything we needed, except I did add a few more assertions around the rows on the shipment cards on the shipment review page. I think this is what was missing according to the e2e discover doc: https://dp3.atlassian.net/wiki/spaces/MT/pages/1641119778/Discovery+for+e2e+testing+for+PO8 BUT I could be missing something else, in which case, let me know!

### Additional steps

1. `make e2e test`
2. Run the customer nts.js test suite, ensure it passes.

### Frontend


cypress/integration/mymove/nts.js
```diff
@@ -28,6 +29,7 @@ describe('Customer NTS Setup flow', function () {
   it('Submits an NTS shipment from homepage', function () {
     cy.apiSignInAsUser(ntsUser);
     customerVisitsReviewPage();
+    customerReviewsShipmentInfoOnReviewPageAfterEdit();
```
reggieriser, line 32: I assume that the smallest unit of work we can run in Cypress is a file (if we could run an individual `it` block, then this might fail since the NTS move might not have been edited yet).

cypress/integration/mymove/nts.js
```diff
@@ -99,6 +101,52 @@ function customerVisitsReviewPage() {
...
+  cy.get('[data-testid="nts-summary"]').contains('123 Any Street P.O. Box 12345');
+  cy.get('[data-testid="nts-summary"]').contains('Beverly Hills, CA 90210');
+
+  // Releasing agent
+  cy.get('[data-testid="nts-summary"]').contains('Jason Ash');
+  cy.get('[data-testid="nts-summary"]').contains('202-555-9301');
+  cy.get('[data-testid="nts-summary"]').contains('jason.ash@example.com');
+
+  // Remarks
+  cy.get('[data-testid="nts-summary"]').contains('Please treat gently');
```
reggieriser, line 124: The `nts-summary` block is a `div` that covers the whole `NTS` section.  Since all these `contains` tests just look anywhere in that block, it seems like we potentially could have the wrong values aligned with the wrong labels and everything would still pass.  That said, I see that's what some existing tests were doing in this same file, so we're following the same pattern.  It also looks like we don't have a good `data-testid` in the labeled sections yet, so we'd have to add that too if we wanted to get more specific.

Not sure there's anything to change here at this point, but thought I'd mention it.

reggieriser (approved):
LGTM!  I added a couple of comments (mainly for my curiosity than any issue with the code).  I also see that the office integration tests are failing, but I don't see evidence that it has anything to do with this PR, plus these Cypress tests ran fine locally for me.

PR #8256, opened 2022-03-08
Fix the diagnostic command

rogeruiz:
## Summary

It looks like `Docker.raw` location has changed. The **Disk utilization
in Docker for Mac** states that this is in a different place compared to
where it was when this was originally written. This uses `find` to start
at the `Data` path to locate the `Docker.raw` file rather than relying
on a strictly hard-coded file path.

=> Official Docker Documentation https://docs.docker.com/desktop/mac/space/

## Verifying this code

Run the following command in the repository on macOS.

```sh
make diagnostic
```

### DIFF
--- a/scripts/check-docker-size
+++ b/scripts/check-docker-size
@@ -7,7 +7,7 @@
 set -eu -o pipefail
 
 # Location of the reserved docker disk space
-DOCKER_RAW=~/Library/Containers/com.docker.docker/Data/vms/0/Docker.raw
+DOCKER_RAW=$(find ~/Library/Containers/com.docker.docker/Data -name Docker.raw)
 
 # Listing the actual used size by docker requires du
 SIZE_ACTUAL=$(/usr/bin/du "${DOCKER_RAW}" | cut -f1)


rogeruiz:
tagging you on this as you're secondary on-call right now. but it's not a rush.

rogeruiz:
@jacquelineIO tagging you on this as you're secondary on-call, but there's no rush in getting this in.

donaldthai:
When I ran `make diagnostic` the output said `Docker is using 0% of its available disk space!` which doesn't seem correct to me. Am I doing something wrong?

rogeruiz:
Thanks @donaldthai. @michaelvaldes-truss mentioned seeing the same thing even with containers running. We checked via `docker ps`. 

I think it could be beneficial to remove this target entirely, as I don't see the value in this as most of us use Docker GUI to view these types of statistics. It's a very old target in the Makefile.

rogeruiz:
Closing this ticket as we don’t seem to use this. I will possibly remove this command and others in another PR after writing up an ADR around stale or abandoned tooling.

PR #8257, opened 2022-03-09
MB-10677-Advance-DB

NamibiaTorres:
## Summary

 The purpose of this story is to archive the reimbursement table and update the ppm_shipment table so that the Advance column is available.

**Work completed:**

Remove “Advance_requested” field from PPM Shipment table ( Do we actually need 2 fields to indicate an advance?)

Make Advance a nullable field (null for no advance was request, an integer for an advance was requested and it was X amount)

Remove the Advance_ID from PPM Shipment

Unlink Advance to the Reimbursement table

Archive the Reimbursement table by renaming

Remove Advance Worksheet and AdnceWorksheetID field.

## Setup to Run Your Code

<details>
<summary>💻 Some changes were made to the internal yaml to add the Advance definition. </summary>

##### Terminal 1

Regenerate the gen files

```sh
make server_generate
```

##### Terminal 2

Run the server to make sure it is working.

```sh
make server_run
```

##### Terminal 3

Run the handler tests, since there was an update

```sh
go test ./pkg/handlers/internalapi  --testify.m TestCreateMTOShipmentHandler -v
```

</details>

### Additional steps

1. Access the
2. Login as a
3.

- [ x] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
- [x ] Request review from a member of a different team.
- [ x] Have the Jira acceptance criteria been met for this change?

### Backend


migrations/app/schema/20220304192644_archive-reimbursements-table.up.sql
```diff
@@ -0,0 +1,13 @@
+-- Archiving because there is data on this table
+ALTER TABLE IF EXISTS reimbursements
```
duncan-truss, line 2: Maybe we should rum the statement dropping the ppm_shipments.advance_id column first before renaming this table?

It looks like the PersonallyProcuredMove model still references the reimbursements table, we should remove that Advance field if it will no longer work https://github.com/transcom/mymove/blob/18e302fbcdb736e27edd2e4f8c2affc9b62ab6f1/pkg/models/personally_procured_move.go#L76

pkg/models/ppm_shipment.go
```diff
@@ -50,11 +50,7 @@ type PPMShipment struct {
 	ProGearWeight                  *unit.Pound       `json:"pro_gear_weight" db:"pro_gear_weight"`
 	SpouseProGearWeight            *unit.Pound       `json:"spouse_pro_gear_weight" db:"spouse_pro_gear_weight"`
 	EstimatedIncentive             *int32            `json:"estimated_incentive" db:"estimated_incentive"`
-	AdvanceRequested               *bool             `json:"advance_requested" db:"advance_requested"`
-	AdvanceID                      *uuid.UUID        `json:"advance_id" db:"advance_id"`
-	Advance                        *Reimbursement    `belongs_to:"reimbursements" fk_id:"advance_id"`
-	AdvanceWorksheetID             *uuid.UUID        `json:"advance_worksheet_id" db:"advance_worksheet_id"`
-	AdvanceWorksheet               *Document         `belongs_to:"documents" fk_id:"advance_worksheet_id"`
```
duncan-truss, line 57: I am trying to get clarity from design but I think the service member will be asked during requesting payment to upload a cost spreadsheet at some point.  I believe we might still use AdvanceWorksheet for that purpose, whether we end up storing it here or not I don't know.

swagger/internal.yaml
```diff
@@ -2542,22 +2542,10 @@ definitions:
...
-        x-omitempty: false
-      advanceWorksheetId:
-        example: 1f2270c7-7166-40ae-981e-b200ebdf3054
-        format: uuid
-        type: string
-        readOnly: true
+      advance:
+        description: |
+          The amount request for an advance, or null if no advance is requested
+        type: integer
```
duncan-truss, line 2548: Could you add `format: cents` here as well so we don't assume its already converted to dollars?

NamibiaTorres: Yes! thanks for calling this out.

pkg/models/ppm_shipment.go
duncan-truss: I think the answer I got is the service member is still uploading the advance spreadsheet but it is for the purpose of pro gear pricing https://ustcdp3.slack.com/archives/C0250Q8469K/p1646839962178309.  We'll probably want to store it somewhere closer to where we capture the closeout weight tickets data.

migrations/app/schema/20220304192644_archive-reimbursements-table.up.sql
NamibiaTorres: Ok, so if I understand correctly, I will need to for an alter table for the removal of advance_id first, then rename the reimbursements table, and then do another alter table for the removal of the other columns plus addition of the Advance column? Or did you mean the Alter Table with the drop and add columns should go before the renaming of the reimbursements ?table.

duncan-truss: Huh I just read where Postgres updates the table name used in the constraints automatically so maybe it's not gonna be an issue at all.  Feel free to disregard.

ronaktruss: Feel free to wait for Duncan to confirm, but my understanding is that there are two separate, but related parts of this story. One for the PPM_Shipments and updating that to fit our current needs and then another part to update the old PPM table / the reimbursement tables. So the current lines 6-10 hand the ppm_shipments portion. 

Since the old ppm shipment model that Duncan linked above still utilize this mapping, we will either need to change this or archive this similar to the work that was done in https://dp3.atlassian.net/browse/MB-11261. (Though the same technique might not work here). While the code indicates a FK relationship between reimbursements and advance id this fk relationship doesn't actually exist on the DB.

ronaktruss: FYI - this "fk relationship" exists only on the backend code. The DB doesn't seem to have this constraint anywhere. 

Let me know if this is incorrect as we would potentially need to make modification to the `archived_personally_procured_moves` table that we created in an earlier story/migration.

ronaktruss: If we just rename the table, won't we have to update the existing model to point to the right db table so that the existing APIs don't fail.

NamibiaTorres: Update for reviewers: We talked through this and it looks like the change is no longer needed.

pkg/models/ppm_shipment.go
NamibiaTorres: FYI, we had a thread coving many topics and one of them was bout the worksheet. My understanding is that this still needs to be worked out in the counselor flow so we could add the columns back in later. This is the thread: https://ustcdp3.slack.com/archives/CP6F568DC/p1646349639765449

migrations/app/schema/20220304192644_archive-reimbursements-table.up.sql
ronaktruss: I wanted to double check on 

> If we just rename the table, won't we have to update the existing model to point to the right db table so that the existing APIs don't fail

since its not clear to me why this is not needed.

My understanding is that if we are renaming the `reimbursements` table wouldn't pop no longer know how to relate the new table with the reimbursements object in go? Without doing an explicit mapping at the model level pop defaults to reimbursements based on the name of the struct. I am under the impression that we would need to manually override the default tableName similar to what was done in the models/ppm_shipment.go
```golang
// TableName overrides the table name used by Pop. By default it tries using the name `ppmshipments`.
func (p PPMShipment) TableName() string {
	return "ppm_shipments"
}
```

It might be the case the reimbursements object is only used in the old PPM code and that we are planning on removing this altogether so we are okay with the model not working. Is this the rationale? I think I did see some usages of the reimbursement model in the office flow, but this might also be related to the old PPM flow.

felipe-lee: The `archived_personally_procured_moves` only has an id column for the advance, but yeah, isn't a FK (same as the original table). We could update that column to be a foreign key so that at least at the DB level, it makes more sense.

That's a good point about renaming the existing model, though I'm not sure how much effort we want to put into it since we'll likely end up deleting the model in the end. So like, idk that we want to go through and update the model everywhere and relationships. Might be easier if for now we just add something like 

```go
func (r Reimbursement) TableName() string {
	return "archived_reimbursements"
}
```

pkg/handlers/internalapi/office_test.go
```diff
@@ -284,48 +284,49 @@ func (suite *HandlerSuite) TestApprovePPMHandlerForbidden() {
...
-	}
-
-	// And: a reimbursement is approved
-	handler := ApproveReimbursementHandler{handlers.NewHandlerContext(suite.DB(), suite.Logger())}
-	response := handler.Handle(params)
-
-	// Then: expect Forbidden response
-	suite.Assertions.IsType(&officeop.ApproveReimbursementForbidden{}, response)
-}
+// TODO: Determine whether we need to complete remove reimbursements handler from Office handlers
```
ronaktruss, line 287: What's the rationale for removing these tests now? (Asking because I could see arguments on both sides for doing this now / in a different pr).

NamibiaTorres: Hi Ronak, the possibility of keeping worksheets is still so up in the air that I'm not sure I can make a call on whether to keep or to remove now.

Re tests: I commented them out because they were breaking due to the change. They shouldn't break anymore due to the minor edit I made to the personally_procured_move model, so I can uncomment them.

pkg/models/reimbursement.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/transcom/mymove/pkg/unit"
 )
 
+// FLAG: This table is deprecated and has been archived and renamed to archived_reimbursements
```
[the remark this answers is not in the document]
NamibiaTorres: It isn't, I'll change to TODO, but I wanted to avoid using TODO since we have so many TODOs everywhere and they have all been seemingly ignored.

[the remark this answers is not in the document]
ronaktruss: No need to make this a TODO. 

Nit: (this is personal preference and you don't need to make this change) I think I would likely just remove the `FLAG:` and just leave the comment. 

As an aside, I believe goland supports a `DEPRECATED:` comment that will make is so that the function is crossed out in your IDE. (As far as I can tell this temporarily doesn't work on VSCode or requires additional configuration)

pkg/handlers/internalapi/office_test.go
ronaktruss: Gotcha, as you mentioned earlier this change is easily revertible so I think its fine to keep as is.

I'll approve once the tests are added back in!

ronaktruss (approved):

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -952,9 +952,11 @@ func (suite *HandlerSuite) TestListMTOShipmentsHandler() {
 				suite.EqualPoundPointers(expectedShipment.PPMShipment.ProGearWeight, returnedShipment.PpmShipment.ProGearWeight)
 				suite.EqualPoundPointers(expectedShipment.PPMShipment.SpouseProGearWeight, returnedShipment.PpmShipment.SpouseProGearWeight)
 				suite.EqualInt32Int64Pointers(expectedShipment.PPMShipment.EstimatedIncentive, returnedShipment.PpmShipment.EstimatedIncentive)
-				suite.Equal(expectedShipment.PPMShipment.AdvanceRequested, returnedShipment.PpmShipment.AdvanceRequested)
-				suite.EqualUUIDPointers(expectedShipment.PPMShipment.AdvanceID, returnedShipment.PpmShipment.AdvanceID)
-				suite.EqualUUIDPointers(expectedShipment.PPMShipment.AdvanceWorksheetID, returnedShipment.PpmShipment.AdvanceWorksheetID)
+				if expectedShipment.PPMShipment.Advance != nil {
+					suite.Equal(expectedShipment.PPMShipment.Advance.Int64(), returnedShipment.PpmShipment.Advance)
```
duncan-truss, line 956: Could you give Advance a value in your test data setup so that we're hitting this condition and not always the else clause?  I think by default Advance is nil in the MakePPMShipment methods so you would need to set it in the assertions object.

migrations/app/schema/20220304192644_archive-reimbursements-table.up.sql
```diff
@@ -0,0 +1,13 @@
...
+
+-- Dropping these columns because we only need advance to record whether an advance has been requested, and for how much.
+ALTER TABLE ppm_shipments
+    DROP COLUMN IF EXISTS advance_id,
+    DROP COLUMN IF EXISTS advance_worksheet_id,
+	DROP COLUMN IF EXISTS advance_requested,
+	ADD COLUMN advance int;
+
+
+COMMENT on COLUMN ppm_shipments.advance IS 'Advance amount and if no advance requested field is set to NULL';
```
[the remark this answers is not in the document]
NamibiaTorres: I made that note to remind us that if the advance is null then no advance has been requested by the user. Do you think the 60% estimated incentive can go in the service object when we define the business logic?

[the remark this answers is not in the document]
duncan-truss: Yea I think you can leave the 60% part out since that threshold may not be fixed.  I think the more important part is making the connection of advance being a portion of the estimated incentive amount or you may think they are unrelated.

pkg/handlers/internalapi/mto_shipment_test.go
NamibiaTorres: Why is it problematic to hit the else clause every time?

duncan-truss: If we only hit the else clause that means it's possible for the change you added in the model to payload to be wrong when Advance is not nil.  We want that condition to be hit so if someone introduces a change we know this test will break so a bug isn't inadvertently introduced.

duncan-truss (approved):
Approved! The updated test looked good and you're resolving the conflicts.

PR #8260, opened 2022-03-09
[MB-11847] Update all pre-commit hooks; fix any issues that were flagged

reggieriser:
## Summary

This PR upgrades all of our pre-commit hooks to their latest versions (using `pre-commit autoupdate`).  It also fixes some issues in files/code that were flagged as a result (along with a couple of scripts that had pre-existing issues that had slipped in).

## Setup to Run Your Code

1. `pre-commit install`
1. `pre-commit install-hooks` (to install the new hooks)
1. `pre-commit autoupdate` (verify no additional hooks were updated)
1. `make clean server_build client_build build_tools`
1. `pre-commit run -a` (make sure there all hooks passed)

Also ensure all tests seem to be passing in CI.

If you're curious what the linter pointed out prior to my fixes (so you know what I was fixing), then go to `master` and run the following:

1. `make clean server_build client_build build_tools`
1. `pre-commit autoupdate`
1. `pre-commit run -a`


scripts/gen-server
```diff
@@ -82,6 +82,7 @@ generate_client swagger/prime.yaml primemessages primeclient
 generate_client swagger/support.yaml supportmessages supportclient
 
 # Wait for all processes to finish
+# shellcheck disable=SC2048
```
reggieriser, line 85: Without this check being disabled, I get the following:
```
In scripts/gen-server line 85:
for pid in ${pids[*]}; do
           ^--------^ SC2048 (warning): Use "${array[@]}" (with quotes) to prevent whitespace problems.
```

But if I do what it says, then I get:

```
In scripts/gen-server line 85:
for pid in "${pids[*]}"; do
           ^----------^ SC2066 (error): Since you double quoted this, it will not word split, and the loop will only run once.
```

I don't think we want this loop only running once, so it seems like I'm in a circle.  I don't think whitespace problems are going to be an issue here with process IDs (please correct me if you have more context), so it seems reasonable to disable this check.

rogeruiz: I think there's a way to fix this. but it does add more complexity to the script as we will have to change the `$IFS` or **Internal Field Separator** for the `for...in...do` loop to understand how to properly split the text from the array. I'm okay with just having the `shellcheck disable` but could we add a comment? Perhaps something like this?

```suggestion
# We need to the following disable call in order to not have to modify the IFS variable for the for-loop.
# shellcheck disable=SC2048
```

ahobson: @reggieriser did you try `"${pids[@]}"` instead of `${pids[*]}` ? I think that would work

ahobson: Sorry, to explain: `pids` is already declared as an array on line 38, so using `@` to access all the elements of the array is what we want, AFAICT

reggieriser: 🤦‍♂️ @ahobson I got so preoccupied with the linter's request to add quotes that I missed that their suggestion also changed the `*` to `@`.  That seems like it should work and passes the linter without any suppressions.  I've made the adjustment.  Thanks!

ahobson: AFAICT, using `"${pids[@]}"` passes the `shell-lint` pre-commit check and the `gan-server` script also seems to work

ahobson (approved):

LeDeep (approved):
ran the commands, all looks good!

PR #8264, opened 2022-03-10
[MB-11696] PPM modify pricer param lookups to not require MTO service items

duncan-truss:
## Summary

As part of this story Ronak and I weighed the choice of if a PPM shipment needs MTO service items to be created, like they are when HHG shipments are approved, in order to calculate the PPM estimated incentive (and potentially also the final incentive).

We chose to not create the default MTO service items (DLH, FSC, DOP, DDP, DPK, and DUPK) in the database at least for the estimated incentive.  Going this route we identified the distance zip3/zip5 lookups that queried for the existing service item shipment fields that would need changed for future stories to price service items.

SIT service items have some of the most complex lookups and we may need to take a different approach there for the final incentive calculation but not for the estimate.

We wrote up some of the tradeoff decisions in this Confluence document https://dp3.atlassian.net/wiki/spaces/MT/pages/1672609793/PPM+Pricing+Creating+MTO+Service+Items

Jacqui did propose combining these pricer lookups in the past because they are so similar https://github.com/transcom/mymove/pull/6368.  I don't remember where we landed on shorthaul pricing if the origin and destination have the same zip3 digits.  For HHG there are actually different shipment types and service items for shorthaul.

##### Terminal 1

Run all the server tests.

```sh
make server_test
```

or

Run all of the service_param_vallue_lookups tests

```sh
make db_test_reset && make db_test_migrate
go test ./pkg/payment_request/service_param_value_lookups
```

### Frontend

### Backend


pkg/payment_request/service_param_value_lookups/contract_code_lookup_test.go
```diff
@@ -24,21 +24,15 @@ func (suite *ServiceParamValueLookupsSuite) TestContractCodeLookup() {
...
 		})
 
-		// MS
-		mtoServiceItem1 := testdatagen.MakeMTOServiceItem(suite.DB(), testdatagen.Assertions{
-			ReService: reService1,
-		})
-
 		// ContractCode
-		serviceItemParamKey1 := testdatagen.MakeServiceItemParamKey(suite.DB(), testdatagen.Assertions{
+		serviceItemParamKey1 := testdatagen.FetchOrMakeServiceItemParamKey(suite.DB(), testdatagen.Assertions{
```
duncan-truss, line 35: I couldn't run the tests on this package directly without running into database unique key issues.  I will see if I can work through these errors but may not find a solution.

pkg/payment_request/service_param_value_lookups/distance_zip3_lookup.go
```diff
@@ -24,28 +24,23 @@ func (r DistanceZip3Lookup) lookup(appCtx appcontext.AppContext, keyData *Servic
...
-		Eager("MTOShipment", "MTOShipment.PickupAddress", "MTOShipment.DestinationAddress").
-		Find(&mtoServiceItem, mtoServiceItemID)
+	// Make sure there's an MTOShipment since that's nullable
+	mtoShipmentID := keyData.mtoShipmentID
+	if mtoShipmentID == nil {
+		return "", apperror.NewNotFoundError(uuid.Nil, "looking for MTOShipmentID")
+	}
+
+	var mtoShipment models.MTOShipment
+	err := db.Find(&mtoShipment, keyData.mtoShipmentID)
```
duncan-truss, line 34: We don't need the service item to exist to query for the shipment that is really only functioning as a not found guard.  We use the PickupAddress and DestinationAddress instance values set on the DistanceZip3Lookup struct and not the ones coming back from this query anyway.

pkg/payment_request/service_param_value_lookups/distance_zip3_lookup_test.go
```diff
@@ -49,6 +55,30 @@ func (suite *ServiceParamValueLookupsSuite) TestDistanceZip3Lookup() {
...
+		}
+
+		distance, err := distanceZip3Lookup.lookup(suite.AppContextForTest(), &ServiceItemParamKeyData{
+			planner:       suite.planner,
+			mtoShipmentID: &ppmShipment.ShipmentID,
+		})
+		suite.NoError(err)
+
+		planner := suite.planner.(*mocks.Planner)
+		planner.AssertCalled(suite.T(), "Zip3TransitDistance", mock.Anything, ppmShipment.PickupPostalCode, ppmShipment.DestinationPostalCode)
```
duncan-truss, line 73: I noticed we weren't asserting that the mock was called with the right arguments (it accepts any if you look in the testsuite) so thought that was a better test even though the mocked distanced returned is constant.

felipe-lee: Wait sorry, I don't think I fully understand the call args for this. What is the `mock.Anything` filling in?

pkg/payment_request/service_param_value_lookups/eia_fuel_price_lookup_test.go
```diff
@@ -109,7 +129,10 @@ func (suite *ServiceParamValueLookupsSuite) TestEIAFuelPriceLookup() {
 	suite.Run("No MTO shipment pickup date found", func() {
 		setupTestData()
 
-		mtoServiceItem := testdatagen.MakeDefaultMTOServiceItem(suite.DB())
+		// create a service item that has a shipment without an ActualPickupDate
+		mtoServiceItem := testdatagen.MakeMTOServiceItem(suite.DB(), testdatagen.Assertions{
+			MTOShipment: testdatagen.MakeMTOShipmentMinimal(suite.DB(), testdatagen.Assertions{}),
```
felipe-lee, line 134: Small thing, but you can use `testdatagen.MakeDefaultMTOShipmentMinimal(suite.DB())` so you don't need to pass the empty assertions. Not important though, just small shortcut.

pkg/payment_request/service_param_value_lookups/eia_fuel_price_lookup_test.go
```diff
@@ -21,14 +21,34 @@ func (suite *ServiceParamValueLookupsSuite) TestEIAFuelPriceLookup() {
 
 		firstGHCDieselFuelPrice.PublicationDate = time.Date(2020, time.July, 06, 0, 0, 0, 0, time.UTC)
 		firstGHCDieselFuelPrice.FuelPriceInMillicents = unit.Millicents(243699)
+
+		var existingFuelPrice1 models.GHCDieselFuelPrice
+		err := suite.DB().Where("ghc_diesel_fuel_prices.publication_date = ?", firstGHCDieselFuelPrice.PublicationDate).First(&existingFuelPrice1)
+		if err == nil {
+			firstGHCDieselFuelPrice.ID = existingFuelPrice1.ID
+		}
```
felipe-lee, line 29: Why is this needed? Like my understanding is that each test is run in a transaction so they'll be created each time the function is called and there shouldn't be pre-existing ones.

felipe-lee (commented):
A few questions and comments on what's done so far. I am glad we're making the changes in this direction though, I agree with y'alls points in the confluence doc.

pkg/payment_request/service_param_value_lookups/distance_zip3_lookup_test.go
duncan-truss: Oh the mock.Anything is for the appContext when calling planner.Zip3TransitDistance(appCtx, pickupPostalCode, destinationPostalCode).  I can check for that type if you want.

pkg/testdatagen/make_service_param.go
```diff
@@ -70,11 +50,61 @@ func MakeServiceParam(db *pop.Connection, assertions Assertions) models.ServiceP
...
+	// fmt.Printf("serviceParam param key id %s\n", serviceParam.ServiceItemParamKeyID)
 	mustCreate(db, &serviceParam, assertions.Stub)
 
 	return serviceParam
 }
 
+func setServiceParamIDs(db *pop.Connection, serviceParam *models.ServiceParam, assertions Assertions) {
+	// Make sure we have a ServiceID
+	var reServiceItem models.ReService
+	if assertions.ServiceParam.ServiceID == uuid.Nil && assertions.ReService.ID == uuid.Nil {
```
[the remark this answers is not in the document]
duncan-truss: Ahh yes I didn't notice the difference, I'll update the first one to be like the second.

pkg/testdatagen/make_service_param.go
```diff
@@ -70,11 +50,61 @@ func MakeServiceParam(db *pop.Connection, assertions Assertions) models.ServiceP
...
+	if (assertions.ServiceParam.ServiceID == uuid.Nil && assertions.ReService.ID == uuid.Nil) ||
+		(assertions.ServiceParam.ServiceItemParamKeyID == uuid.Nil &&
+			assertions.ServiceParam.ServiceItemParamKey.ID == uuid.Nil &&
+			assertions.ServiceItemParamKey.ID == uuid.Nil) {
+		return MakeServiceParam(db, assertions)
+	}
+
+	serviceParam := models.ServiceParam{}
+
+	setServiceParamIDs(db, &serviceParam, assertions)
```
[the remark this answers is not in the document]
duncan-truss: Yea I considered that but I wouldn't know after the IDs are set if they were passed in as existing ones or we just created them.  If I change those MakeDefaultReService and MakeDefaultServiceItemParamKey calls to the Fetch versions then yes I think I could make the change you're suggesting.  But right now if we create those records, we know that there isn't going to be a match in the ServiceParam table.

pkg/payment_request/service_param_value_lookups/eia_fuel_price_lookup_test.go
duncan-truss: I think all this data is populated by the migrations, if you are running `go test ./pkg/payment_request/service_param_value_lookups` then the data is not truncated as it is with `make server_test` I believe.

pkg/testdatagen/make_service_param.go
```diff
@@ -70,11 +50,61 @@ func MakeServiceParam(db *pop.Connection, assertions Assertions) models.ServiceP
...
+	} else if assertions.ServiceParam.ServiceItemParamKey.ID != uuid.Nil {
+		serviceParam.ServiceItemParamKeyID = assertions.ServiceParam.ServiceItemParamKey.ID
+	} else if assertions.ServiceItemParamKey.ID != uuid.Nil {
+		serviceParam.ServiceItemParamKeyID = assertions.ServiceItemParamKey.ID
+	}
+}
+
+func FetchOrMakeServiceParam(db *pop.Connection, assertions Assertions) models.ServiceParam {
+	// ServiceID and ServiceItemParmKeyID are unique to the ServiceParam and must be set unless creating with defaults
+	if (assertions.ServiceParam.ServiceID == uuid.Nil && assertions.ReService.ID == uuid.Nil) ||
```
pearl-truss, line 88: I'm not entirely sure but I think this might causing the issues that you're seeing with the server test with the `panic: sql: no rows in result set` error. Because in the code when `FetchOrMakeServiceParam` is being called (like[ here](https://github.com/transcom/mymove/pull/8264/files#diff-ea26fa718c123290eeb16d2ed4571cb833ec575de9b31fbde6c001bc26cdeb3bR171)), the assertions:

```
ServiceParam: models.ServiceParam{
	ServiceID:             mtoServiceItemDLH.ReServiceID,
	ServiceItemParamKeyID: serviceItemParamKey1.ID,
	ServiceItemParamKey:   serviceItemParamKey1,
}
```
doesnt meet the conditional here so `MakeServiceParam(db, assertions)` isn't being run

duncan-truss: Yep you're right, I have this fixed locally just hadn't pushed up my changes yet.

pkg/testdatagen/make_service_param.go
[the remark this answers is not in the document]
felipe-lee: Ah ok, good point

pkg/payment_request/service_param_value_lookups/eia_fuel_price_lookup_test.go
felipe-lee: Oh, I see, that makes sense, thanks!

pkg/payment_request/service_param_value_lookups/distance_zip3_lookup_test.go
felipe-lee: Oh 🤦🏼 I missed that. I think it might be good if possible, just so that we're not accidentally letting something through incorrectly. I know there's other parts of the code where the mocks have led to the tests not catching changes/issues. I don't think that's an issue here, but might be good practice to avoid `mock.Anything`, when possible. I'll leave that up to you though.

duncan-truss: I replaced it with the real test appcontext arg when I couldn't get the type assertion to work.

felipe-lee (approved):
Awesome, lgtm, thanks for making these changes!

I'm glad that these tests can now be run more easily without worrying about truncating the DB or running all the tests.

PR #8275, opened 2022-03-14
MB-11037: Advance Form for PPM booking on storybook

pearl-truss:
## Summary

This PR creates the Advance form for PPM Booking and adds it to storybook under: `Customer Components /PPM Booking`

It can be tested by running storybook locally or navigating to here: https://617557-114694829-gh.circle-artifacts.com/0/storybook/index.html?path=/story/customer-components-ppm-booking-advance-form--blank-advance-form
## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

### Frontend

## Screenshots
Screen Shot 2022-03-14 at 2 42 49 PM
Screen Shot 2022-03-14 at 2 42 57 PM
Screen Shot 2022-03-14 at 2 43 04 PM


src/components/Customer/PPMBooking/Advance/AdvanceForm.jsx
```diff
@@ -0,0 +1,143 @@
...
+
+const validationSchema = Yup.object().shape({
+  advanceRequested: Yup.boolean().required('Required'),
+  amountRequested: Yup.number()
+    .min(1, "The minimum advance request is $1. If you don't want an advance, select No.")
+    .when('advanceRequested', {
+      is: true,
+      then: (schema) =>
+        schema.required(
+          `Enter a weight into at least one pro-gear field. If you won't have pro-gear, select No above.`,
```
gidjin, line 28: Does this message need to be updated? Wondering if it should be updated to `Enter an advance amount. If you do not want an advance select No above.` or something.

src/components/Customer/PPMBooking/Advance/AdvanceForm.jsx
```diff
@@ -0,0 +1,143 @@
...
+                      id="amountRequested"
+                      mask={Number}
+                      scale={0} // digits after point, 0 for integers
+                      signed={false} // disallow negative
+                      thousandsSeparator=","
+                      lazy={false} // immediate masking evaluation
+                      prefix="$"
+                      error={
+                        mtoShipment?.ppmShipment &&
+                        values.amountRequested > mtoShipment.ppmShipment.estimatedIncentive / 100
```
duncan-truss, line 96: I think you should be able to move this into your validationSchema, you'll just need to make it dynamic instead of static.

I think it could become something like:
```javascript
const validationSchema = (estimatedIncentive) => {
    return Yup.object().shape({
    amountRequested: Yup.number().max(estimatedIncentive, `Your advance is limited to ${maxAdvanceToRequest}`)
    ...
    })
}
```

If you're concerned about this getting called a bunch you could put it inside of a useMemo or useEffect hook so it only changes when the mtoShpment.ppmShipment.estimatedIncentive prop value changes.

duncan-truss: That's right, I forgot to add that in my function definition example. The validation schema doesn't have to be outside of the component but that's why I converted it to a function where you can pass any data you need in.

src/components/form/fields/MaskedTextField/MaskedTextField.jsx
```diff
@@ -54,8 +55,9 @@ const MaskedTextField = ({
       )}
       {showWarning && <Hint data-testid="textInputWarning">{warning}</Hint>}
       {/* eslint-disable react/jsx-props-no-spreading */}
-      {suffix ? (
-        <div className={suffix && styles.hasSuffix}>
+      {suffix || prefix ? (
+        <div className={suffix ? styles.hasSuffix : styles.hasPrefix}>
```
felipe-lee, line 59: Should this maybe use `classnames` to combine things? Like what if a field has both a prefix and a suffix?

src/components/Customer/PPMBooking/Advance/AdvanceForm.test.jsx
```diff
@@ -0,0 +1,100 @@
...
+      expect(screen.getByText('Would you like to request an advance on your incentive?')).toBeInstanceOf(
+        HTMLLegendElement,
+      );
+    });
+  });
+
+  describe('displays conditional inputs', () => {
+    it('displays input for amount requested when advance requested is true', async () => {
+      render(<AdvanceForm {...defaultProps} />);
+      const requestAdvance = await screen.getByLabelText('Yes');
```
felipe-lee, line 50: iirc the `get*` queries aren't async. If you want to `await` it, you'd need to use the `find*` query, I _think_.

src/components/Customer/PPMBooking/Advance/AdvanceForm.test.jsx
```diff
@@ -0,0 +1,122 @@
...
+    id: '123',
+    ppmShipment: {
+      id: '123',
+      estimatedIncentive: 1000000,
+      advance: 30000,
+    },
+  },
+};
+
+describe('AdvanceForm component', () => {
```
felipe-lee, line 32: I'm not sure which test it is, but I'm getting formik updates not wrapped in act errors:

image

felipe-lee (commented):
Overall looks good! Just a few comments, but the only one that I think needs to be addressed for sure is the test error messages.

Also, I think it might be good to pull in master because there's some CSS that might be useful for the error state. Currently it moves the field label around a lot and I had to fix that in another PR, so _maybe_ those changes will help here too.

src/components/Customer/PPMBooking/Advance/AdvanceForm.jsx
```diff
@@ -0,0 +1,143 @@
...
+    }),
+    agreeToTerms: Yup.boolean().when('advanceRequested', {
+      is: true,
+      then: (schema) => schema.oneOf([true], 'Required'),
+    }),
+  });
+};
+
+const AdvanceForm = ({ mtoShipment, onSubmit, onBack }) => {
+  const ppmShipmentAdvance = mtoShipment?.ppmShipment?.advance;
```
duncan-truss, line 39: If you want to cut down on all the optional chaining, I've found destructuring like this to work well:
```javascript
const { advance, estimatedIncentive } = mtoShipment?.ppmShipment || {}
```

src/components/Customer/PPMBooking/Advance/AdvanceForm.jsx
```diff
@@ -0,0 +1,143 @@
...
+  const initialValues = {
+    amountRequested: ppmShipmentAdvance ? formatCentsTruncateWhole(ppmShipmentAdvance) : '',
+    advanceRequested: ppmShipmentAdvance ? 'true' : 'false',
+    agreeToTerms: false,
+  };
+
+  const maxAdvanceToRequest = maxAdvance(mtoShipment?.ppmShipment?.estimatedIncentive);
+  const formatedIncentive = formatCentsTruncateWhole(mtoShipment?.ppmShipment?.estimatedIncentive);
```
duncan-truss, line 47: Should we add some null case checking or use some defaults in case the shipment data hasn't resolved yet so we aren't doing division on undefined values?

dinabeigelman (changes requested):
It looks like the error state for advance amount is pushing the whole field down. I believe this is what @duncan-truss and @felipe-lee also encountered. Might there be a more systemic way to address this issue from popping up again?

error state issue

scarequotes (changes requested):
The Jira ticket says the customer should get a warning if they enter over 60% of their incentive, and a hard block if they enter over 100%. 

Instead, they don't get any message for entering 60-99%, and a hard block for over 100%.

Screen Shot 2022-03-15 at 4 10 55 PM
Screen Shot 2022-03-15 at 4 10 46 PM
Screen Shot 2022-03-15 at 4 10 35 PM

Expected behavior: 

- Should get a warning message if I enter between $6,001 and $10,000, but Save & continue button should be active
- Should get a different warning message if I enter $10,001 or more, and Save & continue should not be active

I don't think we gave specific language for those scenarios, so I recommend:

- for soft validation: "Reminder: your advance can not be more than $X,XXX" (number is 60% of incentive)
- for hard validation: "Enter an amount less than $X,XXX " (where the number is 60% of incentive) 

Also discovered a related error while tracking that down. 

Screen Shot 2022-03-15 at 4 03 56 PM

I got that by entering $555,555, then moving my cursor to the beginning of the line and deleting the second 5. When the amount switched to $55,555, I got that odd error message.

(This but is less critical than the one about hard and soft errors.)

felipe-lee:
> Also discovered a related error while tracking that down.
> 
> Screen Shot 2022-03-15 at 4 03 56 PM
> 
> I got that by entering $555,555, then moving my cursor to the beginning of the line and deleting the second 5. When the amount switched to $55,555, I got that odd error message.
> 
> (This but is less critical than the one about hard and soft errors.)

I think this one should be fixed after master gets merged in.

pearl-truss:
> > Also discovered a related error while tracking that down.
> > Screen Shot 2022-03-15 at 4 03 56 PM
> > I got that by entering $555,555, then moving my cursor to the beginning of the line and deleting the second 5. When the amount switched to $55,555, I got that odd error message.
> > (This but is less critical than the one about hard and soft errors.)
> 
> I think this one should be fixed after master gets merged in.

Thanks for working on the fix for this Felipe! I saw that it was resolved once I pulled in master 🎉

pearl-truss:
> * Reminder: your advance can not be more than

This was resolved when I pulled in master which had Felipe's changes

pearl-truss:
> The Jira ticket says the customer should get a warning if they enter over 60% of their incentive, and a hard block if they enter over 100%.
> 
> Instead, they don't get any message for entering 60-99%, and a hard block for over 100%.
> 
> Screen Shot 2022-03-15 at 4 10 55 PM Screen Shot 2022-03-15 at 4 10 46 PM Screen Shot 2022-03-15 at 4 10 35 PM
> 
> Expected behavior:
> 
> * Should get a warning message if I enter between $6,001 and $10,000, but Save & continue button should be active
> * Should get a different warning message if I enter $10,001 or more, and Save & continue should not be active
> 
> I don't think we gave specific language for those scenarios, so I recommend:
> 
> * for soft validation: "Reminder: your advance can not be more than $X,XXX" (number is 60% of incentive)
> * for hard validation: "Enter an amount less than $X,XXX " (where the number is 60% of incentive)
> 
> Also discovered a related error while tracking that down.
> 
> Screen Shot 2022-03-15 at 4 03 56 PM
> 
> I got that by entering $555,555, then moving my cursor to the beginning of the line and deleting the second 5. When the amount switched to $55,555, I got that odd error message.
> 
> (This but is less critical than the one about hard and soft errors.)

Thanks, this was super helpful and has been addressed

src/components/Customer/PPMBooking/Advance/AdvanceForm.jsx
pearl-truss: Nice, thanks!

duncan-truss (approved):
Code looks good to me, I think you can clean up the warning vars to use your new destructed ones.  We probably want to format the max advance number in the warning too with the $ sign and thousands comma separator.

scarequotes (approved):
The warning message should have a `$` before the amount — so $6,000 in the screenshot below. 

Looks like we're still getting the warning pushing text down the page a little, as well

LGTM otherwise. I'll `Approve` so it's not a blocker but definitely want the dollar sign fixed.

Screen Shot 2022-03-16 at 8 21 48 AM

pearl-truss:
> The warning message should have a `$` before the amount — so $6,000 in the screenshot below.
> 
> Looks like we're still getting the warning pushing text down the page a little, as well
> 
> LGTM otherwise. I'll `Approve` so it's not a blocker but definitely want the dollar sign fixed.
> 
> Screen Shot 2022-03-16 at 8 21 48 AM

This was updated in the latest commit

Screen Shot 2022-03-16 at 11 27 01 AM

pearl-truss:
> Code looks good to me, I think you can clean up the warning vars to use your new destructed ones. We probably want to format the max advance number in the warning too with the $ sign and thousands comma separator.

Good catch, I just pushed a commit that addresses the formatting and cleans up the vars

dinabeigelman (commented):
Looks like the field still seems to jump a bit (tho less than before) when it goes into the error state. Also jumps in the warning states as well.

error state issue 2

dinabeigelman (approved):

PR #8280, opened 2022-03-15
[MB-11940] Adding Cypress tests for NTS-release as SC

reggieriser:
## Summary

This PR adds Cypress tests for manipulating an NTS-release shipment as a services counselor.

__Note:__ MB-11940 is an epic; this PR hopefully covers all 7 stories under this epic.  Please confirm as part of your review.

## Setup to Run Your Code

- `make clean server_build client_build`
- `make e2e_test_fresh`
- Run the `servicesCounselingNTSR.js` tests and verify they succeed locally.

### Frontend

### DIFF
--- a/cypress/integration/office/servicescounseling/servicesCounselingNTSR.js
+++ b/cypress/integration/office/servicescounseling/servicesCounselingNTSR.js
@@ -114,15 +114,15 @@ describe('Services counselor user', () => {
       .within(($form) => {
         cy.get('[data-icon="chevron-down"]').click();
 
-        cy.get('div[class*="warning"] > [data-testid="ntsRecordedWeight"]').should('exist');
-        cy.get('div[class*="missingInfoError"] > [data-testid="storageFacilityName"]').should('exist');
+        cy.get('div[class*="warning"] [data-testid="ntsRecordedWeight"]').should('exist');
+        cy.get('div[class*="missingInfoError"] [data-testid="storageFacilityName"]').should('exist');
         cy.get('[data-testid="storageFacilityName"]').contains('Missing');
-        cy.get('div[class*="warning"] > [data-testid="serviceOrderNumber"]').should('exist');
-        cy.get('div[class*="missingInfoError"] > [data-testid="storageFacilityAddress"]').should('exist');
+        cy.get('div[class*="warning"] [data-testid="serviceOrderNumber"]').should('exist');
+        cy.get('div[class*="missingInfoError"] [data-testid="storageFacilityAddress"]').should('exist');
         cy.get('[data-testid="storageFacilityAddress"]').contains('Missing');
-        cy.get('div[class*="warning"] > [data-testid="counselorRemarks"]').should('exist');
-        cy.get('div[class*="warning"] > [data-testid="tacType"]').should('exist');
-        cy.get('div[class*="warning"] > [data-testid="sacType"]').should('exist');
+        cy.get('div[class*="warning"] [data-testid="counselorRemarks"]').should('exist');
+        cy.get('div[class*="warning"] [data-testid="tacType"]').should('exist');
+        cy.get('div[class*="warning"] [data-testid="sacType"]').should('exist');
       });
 
     editNTSReleaseShipment();
--- a/cypress/integration/office/servicescounseling/servicesCounselingNTSR.js
+++ b/cypress/integration/office/servicescounseling/servicesCounselingNTSR.js
@@ -1,0 +1,278 @@
+import { ServicesCounselorOfficeUserType } from '../../../support/constants';
+
+describe('Services counselor user', () => {
+  before(() => {
+    cy.prepareOfficeApp();
+  });
+
+  beforeEach(() => {
+    cy.intercept('**/ghc/v1/swagger.yaml').as('getGHCClient');
+    cy.intercept('**/ghc/v1/queues/counseling?page=1&perPage=20&sort=submittedAt&order=asc').as('getSortedMoves');
+    cy.intercept('**/ghc/v1/move/**').as('getMoves');
+    cy.intercept('**/ghc/v1/orders/**').as('getOrders');
+    cy.intercept('**/ghc/v1/move_task_orders/**/mto_shipments').as('getMTOShipments');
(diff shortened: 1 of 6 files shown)


cypress/integration/office/servicescounseling/servicesCounselingNTSR.js
```diff
@@ -0,0 +1,278 @@
...
+        cy.get('div[class*="warning"] > [data-testid="counselorRemarks"]').should('exist');
+        cy.get('div[class*="warning"] > [data-testid="tacType"]').should('exist');
+        cy.get('div[class*="warning"] > [data-testid="sacType"]').should('exist');
+      });
+
+    editNTSReleaseShipment();
+  });
+});
+
+function navigateToMove(moveLocator) {
```
reggieriser, line 132: I believe this function is identical to the one @jenniferlynparsons had in her PR (#8261).  Glad to abstract this out if we want so it can be shared -- where's a good place to put it in the Cypress world?

LeDeep: I think you could abstract out to `support/commands`.  https://docs.cypress.io/api/cypress-api/custom-commands#Syntax

reggieriser: I just tried that and it worked, but ... the code waits on a bunch of routes that are setup in the test's `beforeEach`.  So, if someone uses this command but hasn't set up those specific routes, then the command fails (which makes it feel less generic).  Maybe it can just be made a shared function for just these SC tests?  I'll play around with that.

jenniferlynparsons: Good call! We can move this to the `commands.js` file!

jenniferlynparsons: I've been thinking about abstracting those `intercept`s out as well since we use them all over the place. Not something for this PR but it's a thought.

cypress/integration/office/servicescounseling/servicesCounselingNTSR.js
```diff
@@ -0,0 +1,278 @@
...
+
+  it('Services Counselor can see errors/warnings for missing data, then make edits', () => {
+    navigateToMove('NTSRMN');
+
+    cy.get('[data-testid="ShipmentContainer"]')
+      .last()
+      .within(($form) => {
+        cy.get('[data-icon="chevron-down"]').click();
+
+        cy.get('div[class*="warning"] > [data-testid="ntsRecordedWeight"]').should('exist');
```
jenniferlynparsons, line 117: I doubt it will change but was the `>` necessary to target these fields? In theory `div[class*="warning"] [data-testid="ntsRecordedWeight"]` would have been sufficient and be a little less brittle if the dom changes.

jenniferlynparsons (approved):
Approval with a few non-blocking suggestions/questions.

cypress/integration/office/servicescounseling/servicesCounselingNTSR.js
reggieriser: I believe what you're describing would work if the `div` had both of those matching attributes, but that's not the case here (I just tried it to make sure and it failed).  The DOM for this looks like:
```html
<div class="descriptionList_row__18vxY ShipmentDefinitionLists_warning__3TiPa">
    <dt>Previously recorded weight</dt>
    <dd data-testid="ntsRecordedWeight">—</dd>
</div>
```

The fact it's a warning is noted in the CSS class for the `div` (parent), but I need to also match the right one of those by making sure its child (via the `>`) has the expected `data-testid` value.  I'm not an expert on CSS selectors, so glad to simplify if possible.

jenniferlynparsons: Okay, I think I understand your approach better. so `>` indicates the element is a direct child, where leaving that out indicates the element would be any nested child. 

`div[class*="warning"] > [data-testid="ntsRecordedWeight"]` means that the exact DOM structure must be a `div[class*="warning"]` with a direct child `[data-testid="ntsRecordedWeight"]` 

`div[class*="warning"] [data-testid="ntsRecordedWeight"]` would be a DOM structure where a `div[class*="warning"]` has a child somewhere within it that is `[data-testid="ntsRecordedWeight"]` 

In the first case, if we added an additional `div` around the `dt` /`dd` combination (or wrapped them in anything else), the tests would break.

In the second case, we can change the DOM structure within `div[class*="warning"]` as much as we like and it will still locate the `[data-testid="ntsRecordedWeight"]`.

cypress/integration/office/servicescounseling/servicesCounselingNTSR.js
reggieriser: I played around with an exported/shared function but that seemed to have its quirks too.  If I leave the function in the NTS file and just export/import it, all the tests for both run when I run the NTS-R file.  If I put it in a separate file like `shared.js` in the same directory, it shows up in the Cypress UI as if it has tests to run.  Based on the [Cypress docs](https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests#Support-file), it seems that putting it in a command or the support file may be the way to go, but it still feels too specific to these tests to me.  Rather than establish a new (perhaps wrong) pattern here, I'm going to leave the little bit of duplication as is for this PR and we can abstract out later if desired (perhaps with JP's suggestion about intercepts).

cypress/integration/office/servicescounseling/servicesCounselingNTSR.js
reggieriser: Oh, you know what?  In your original message, I missed that you had a space in there since github put a line break between the two `[]` parts.  That may make a difference.  I'll try that out.

reggieriser: Removing the `>` but keeping the space worked, so I've made that change and agree it'd be a little less brittle.  Thanks!  Sorry for the confusion.

jenniferlynparsons: Awesome! Thanks github for providing some extra confusion. lol.

PR #8303, opened 2022-03-18
MB-10676 Advance Page - frontend

pearl-truss:
## Summary

This PR wires the Advance page into the PPM onboarding flow. It follows the Estimated Incentive page and precedes the Review page

### Open Questions

- **DESIGN**: I noticed that when you fill out an advance, and you come back to the page, the amount requested value inputed persists, however the legalesee checkbox that was previously selected in order to go to the next page is unselected again when the page is revisited. Please let me know what the desired behavior should be. I was thinking if anyone comes to the page with an advance amount already set on the backend (meaning they already would've had to click the checkbox in order to submit the Advance form) then the checkbox could appear as selected?
- **ENGINEERING**: After submitting the advance form and being sent to the Review page, if you click the browsers back button you end up on the Advance page but you have to reload in order to load any of the previously prefilled values. I noticed this also happens on all of the preceding pages in the form. Should I try to address that in this ticket or is that captured somewhere else already?

## Setup to Run Your Code

<details>
<summary>💻 You will need to use 3 separate terminals to test this locally.</summary>
##### Terminal 1

Populate the new devseed user

```sh
make db_dev_e2e_populate
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the local dev environment
2. Login as advance_requested@ppm.unsubmitted (milmove)
3. Click the pen icon on the PPM shipment card to be entered into the PPM onboarding flow. Click through from the Dates and Location page until you get to the Advance page
4. Any valid value added to the Amount Requested field should persist when the page is revisted 

### Frontend

## Screenshots
Screen Shot 2022-03-18 at 2 02 45 PM


pkg/services/validation.go
```diff
@@ -66,6 +66,21 @@ func SetOptionalPoundField(newWeight *unit.Pound, oldWeight *unit.Pound) *unit.P
...
 
+// SetOptionalCentsField sets the correct new value for the updated cents field. Can be nil.
+func SetOptionalCentsField(newValue *unit.Cents, oldValue *unit.Cents) *unit.Cents {
+	// check if the user wanted to keep this field the same:
+	if newValue == nil {
+		return oldValue
+	}
+
+	// check if the user wanted to nullify the value in this field:
+	if *newValue == 0 {
```
duncan-truss, line 77: To make this work for clearing previously set values you would need to do some post processing in your onSubmit handler of the Advance page to set Advance to zero if advance requested is "No".  Doing it this way though means that our swagger payload will allow zero values, which won't match our client side validations.  I think that's why I was leaning to storing advanceRequested in the database because then we could restrict the swagger advance value with a min property.

And we need a way to know that the user has reached this screen to know if a PPM shipment is incomplete.  If advance is null we can't distinguish between saying they don't want an advance and they haven't progressed to the advance screen page yet.

duncan-truss, line 77: To make this work for clearing previously set values you would need to do some post processing in your onSubmit handler of the Advance page to set Advance to zero if advance requested is "No".  Doing it this way though means that our swagger payload will allow zero values, which won't match our client side validations.  I think that's why I was leaning to storing advanceRequested in the database because then we could restrict the swagger advance value with a min property.

And we need a way to know that the user has reached this screen to know if a PPM shipment is incomplete.  If advance is null we can't distinguish between saying they don't want an advance and they haven't progressed to the advance screen page yet.

dinabeigelman (approved):
Looks good! Only thing of note is the wonkiness with the jumping when the amount requested field is in a warning or error state that's been called out previously.

To answer your open question about "should we have the customer recheck the legalese checkbox" yes, we should. cc @scarequotes

scarequotes:
> Looks good! Only thing of note is the wonkiness with the jumping when the amount requested field is in a warning or error state that's been called out previously.
> 
> To answer your open question about "should we have the customer recheck the legalese checkbox" yes, we should. cc @scarequotes

Agree. Dina and I discussed on Slack and went with this approach.

scarequotes (approved):
Looks good to me.

swagger-def/internal.yaml
```diff
@@ -2655,6 +2652,13 @@ definitions:
         type: integer
         example: 4300
         x-nullable: true
+      advance:
```
felipe-lee, line 2655: Since #8317 also has some of this, are you going to merge it into this one? Mostly wondering about the overlap

pkg/handlers/internalapi/internal/payloads/payload_to_model.go
```diff
@@ -148,6 +148,7 @@ func UpdatePPMShipmentModel(ppmShipment *internalmessages.UpdatePPMShipment) *mo
 		ProGearWeight:                  handlers.PoundPtrFromInt64Ptr(ppmShipment.ProGearWeight),
 		SpouseProGearWeight:            handlers.PoundPtrFromInt64Ptr(ppmShipment.SpouseProGearWeight),
 		EstimatedIncentive:             &estimatedIncentive,
+		Advance:                        handlers.FmtInt64PtrToPopPtr(ppmShipment.Advance),
```
felipe-lee, line 151: I know you didn't set the hard-coded `estimatedIncentive` above, but I wonder if we should move it from here to instead be in the `services/ppmShipment/validation.go` file, when we merge. Like set it if it's not already set. The reason I bring it up is that currently, it keeps setting the estimated incentive to be $100, which immediately clashes with the data that is set up in the `testdatagen` and so as soon as you land on the advance page, there's an error (advance too high). Also since `estimatedIncentive` isn't something that can even come in from the API, probably cleaner to not have it in this file at all.

Given that this branch is more focused on the FE though, I'm fine with not making this change. It could even be part of the ticket that updates `EstimatedIncentive` to use cents.

src/pages/MyMove/PPMBooking/Advance/Advance.test.jsx
```diff
@@ -0,0 +1,278 @@
...
+    ...mockMTOShipment.ppmShipment,
+    estimatedWeight: 4000,
+    hasProGear: false,
+    proGearWeight: null,
+    spouseProGearWeight: null,
+    eTag: btoa(new Date()),
+    estimatedIncentive: 1000000,
+  },
+  eTag: btoa(new Date()),
+};
```
felipe-lee, line 61: Minor nit, but this doesn't seem to be setting any new fields that `mockMTOShipment` doesn't already have so you can probably sway any usages of `mockPreExistingShipment` with `mockMTOShipment`

src/pages/MyMove/PPMBooking/Advance/Advance.test.jsx
```diff
@@ -0,0 +1,278 @@
...
+  ...mockMTOShipment,
+  ppmShipment: {
+    ...mockMTOShipment.ppmShipment,
+    estimatedWeight: 4000,
+    advance: 40000,
+    hasProGear: false,
+    proGearWeight: null,
+    spouseProGearWeight: null,
+    eTag: btoa(new Date()),
+    estimatedIncentive: 1000000,
```
felipe-lee, line 73: Another small nit, but I think the only two that probably need to be defined are `advance` and `eTag` since the rest are set to the same values by `...mockMTOShipment.ppmShipment`.

src/pages/MyMove/PPMBooking/Advance/Advance.test.jsx
```diff
@@ -0,0 +1,278 @@
...
+    expect(backButton).toBeInTheDocument();
+    expect(backButton).not.toBeDisabled();
+
+    const saveButton = screen.getByRole('button', { name: /save & continue/i });
+    expect(saveButton).toBeInTheDocument();
+    expect(saveButton).not.toBeDisabled();
+  });
+
+  it.each([[mockPreExistingShipment], [mockPreExistingShipmentWithAdvance]])(
+    'renders the form pre-filled when advance info has been entered previously',
```
felipe-lee, line 138: Minor nit on the name of this test in that the first mock shipment doesn't have advance info. I think the test itself is good though, just the name might need to be something else. What that is, I have no idea...maybe something like without/with advance info? Idk

src/pages/MyMove/PPMBooking/Advance/Advance.test.jsx
```diff
@@ -0,0 +1,278 @@
...
+    const advanceInput = screen.getByLabelText('Amount requested');
+    userEvent.type(advanceInput, String(advance));
+
+    const saveButton = screen.getByRole('button', { name: /save & continue/i });
+    userEvent.click(saveButton);
+
+    await waitFor(() => expect(mockDispatch).toHaveBeenCalledWith(updateMTOShipment(mockPreExistingShipment)));
+  });
+
+  it('routes to the estimated incentive page when the user clicks save & continue', async () => {
```
felipe-lee, line 240: Small tweak to match expected path
```suggestion
  it('routes to the review page when the user clicks save & continue', async () => {
```

felipe-lee (approved):
Mostly lgtm! I think my main question is the BE changes that are in this one and in #8317 

Also as for
> ENGINEERING: After submitting the advance form and being sent to the Review page, if you click the browsers back button you end up on the Advance page but you have to reload in order to load any of the previously prefilled values. I noticed this also happens on all of the preceding pages in the form. Should I try to address that in this ticket or is that captured somewhere else already?

It's a strange issue, but I can confirm I can re-create it. Strangely, if you then move forward again with the forward button, the data loads. I'd say that that should be a separate PR, whether that's part of this ticket though I'll leave up to you.

swagger-def/internal.yaml
pearl-truss: Yeah good point, I'm going to merge @NamibiaTorres 's PR into this one, there just a few BE things I needed to add in order to get  this PR up and working

pkg/handlers/internalapi/internal/payloads/payload_to_model.go
pearl-truss: @felipe-lee I increased the hardcoded `estimatedIncentive` to be $10,000 as a temp work around for this PR

src/pages/MyMove/PPMBooking/Advance/Advance.jsx
```diff
@@ -0,0 +1,71 @@
...
+    patchMTOShipment(mtoShipment.id, payload, mtoShipment.eTag)
+      .then((response) => {
+        setSubmitting(false);
+        dispatch(updateMTOShipment(response));
+        history.push(generatePath(customerRoutes.MOVE_REVIEW_PATH, { moveId }));
+      })
+      .catch((err) => {
+        setSubmitting(false);
+
+        setErrorMessage(getResponseError(err.response, 'Failed to update MTO shipment due to server error.'));
```
duncan-truss, line 47: I see we're using this in a number of places across the app, I wonder what the possible error messages could be here and how helpful is the body.detail or statusText message?  They could end up seeing a not found error, an eTag error, or maybe a shipment status error.

src/pages/MyMove/PPMBooking/Advance/Advance.jsx
```diff
@@ -0,0 +1,71 @@
...
+    history.push(generatePath(customerRoutes.SHIPMENT_PPM_ESTIMATED_INCENTIVE_PATH, { moveId, mtoShipmentId }));
+  };
+
+  const handleSubmit = async (values, { setSubmitting }) => {
+    setErrorMessage(null);
+
+    const payload = {
+      ppmShipment: {
+        id: mtoShipment.ppmShipment.id,
+        advance: Number(values.amountRequested * 100),
```
duncan-truss, line 33: Can we make sure to send null in the case of advanceRequested being false here or at least omit it from the payload? I'm pretty sure the advance field doesn't get cleared if No is selected for advanceRequested if it had a previous value. I would avoid doing numeric operations if the advance field has no value either even if if the backend ignores it.

src/pages/MyMove/PPMBooking/Advance/Advance.jsx
```diff
@@ -0,0 +1,71 @@
...
+        advanceRequested: Boolean(values.advanceRequested),
```
duncan-truss, line 34: This is not actually parsing the strings value to the boolean type so it's always passing true.

Screen Shot 2022-03-23 at 1 52 33 PM

felipe-lee: Oh good catch. I think something more like this right?

https://github.com/transcom/mymove/blob/628e07fca821c2a5b5d0635ed7e2bda4b081dd48/src/pages/MyMove/PPMBooking/DateAndLocation/DateAndLocation.jsx#L38

src/pages/MyMove/PPMBooking/Advance/Advance.jsx
felipe-lee: Yeah I think sending null or leaving it out works. The BE code will have a statement to set it to null if `advanceRequested` is set to `false`.

Agree on the avoiding numeric operations if the var in undefined or null

src/pages/MyMove/PPMBooking/Advance/Advance.jsx
felipe-lee: yeah, tbh we probably need a more thorough review of error handling across the app because we handle this differently in several spots. Some pages catch the error and then display a generic message, while others do this, but like you said, the messages (and utility of those messages) can vary.

pkg/testdatagen/scenario/shared.go
```diff
@@ -700,6 +700,60 @@ func createUnsubmittedMoveWithPPMShipmentThroughEstimatedWeights(appCtx appconte
...
+	})
+
+	testdatagen.MakeMinimalPPMShipment(appCtx.DB(), testdatagen.Assertions{
+		Move: move,
+		PPMShipment: models.PPMShipment{
+			ID:                 uuid.Must(uuid.FromString("9160a396-9b60-41c2-af7a-aa03d5002c71")),
+			EstimatedWeight:    models.PoundPointer(unit.Pound(4000)),
+			HasProGear:         models.BoolPointer(false),
+			EstimatedIncentive: models.Int32Pointer(int32(10000000)),
+			Advance:            models.CentPointer(unit.Cents(30000)),
```
felipe-lee, line 752: We need to also set `AdvanceRequested` here, otherwise this PPMShipment will have invalid data.

pkg/testdatagen/scenario/shared.go
```diff
@@ -700,6 +700,60 @@ func createUnsubmittedMoveWithPPMShipmentThroughEstimatedWeights(appCtx appconte
...
+		UserUploader: userUploader,
+		Move: models.Move{
+			ID:               uuid.Must(uuid.FromString("fe322fae-c13e-4961-9956-69fb7a491ad4")),
+			Locator:          "PPMADV",
+			SelectedMoveType: &ppmMoveType,
+		},
+	})
+
+	testdatagen.MakeMinimalPPMShipment(appCtx.DB(), testdatagen.Assertions{
+		Move: move,
```
felipe-lee, line 746: Can you also set the UUID for the MTOShipment? I realized as we were testing the nil pointer dereference error you comment about here: https://github.com/transcom/mymove/pull/8317/files#r834303291

That the MTOShipment ID changes each time this gets re-generated, but that's what is easiest to grab from the network requests on the FE.

felipe-lee (commented):
Found two things related to the nil pointer dereference error that could use updates here.

pkg/testdatagen/scenario/shared.go
felipe-lee: Mmm nevermind on this. To get this to work well it would require updates to the PPM generation functions. Since this bit isn't necessary, and those changes are out of scope, we can leave this out. But the other change to set `AdvanceRequested` is still needed.

pkg/testdatagen/scenario/shared.go
felipe-lee: This was one of the bits that was leading to the nil pointer dereference error. Namibia is also making changes to the BE though to make sure this type of thing is caught more gracefully.

pearl-truss: Ahh okay, thanks for tracking this down

gidjin (approved):
DB changes look good. 🚢

cypress/integration/milmove/ppms/estimatedIncentive.js
```diff
@@ -44,5 +44,5 @@ function navigateToEstimatedIncentivePage() {
 }
 
 function verifyShipmentSpecificInfo() {
-  cy.get('.container h2').contains('$100');
+  cy.get('.container h2').contains('$10000');
```
felipe-lee, line 47: Looks like it might need a comma

```suggestion
  cy.get('.container h2').contains('$10,000');
```

image

image

src/pages/MyMove/PPMBooking/Advance/Advance.jsx
```diff
@@ -0,0 +1,73 @@
...
+        advance: values.amountRequested ? values.amountRequested * 100 : null,
+        advanceRequested,
+      },
+    };
+
+    patchMTOShipment(mtoShipment.id, payload, mtoShipment.eTag)
+      .then((response) => {
+        setSubmitting(false);
+        dispatch(updateMTOShipment(response));
+        history.push(generatePath(customerRoutes.MOVE_REVIEW_PATH, { moveId }));
```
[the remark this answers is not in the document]
pearl-truss: Great catch! I think this counts as AC for this ticket so I updated it in this PR

Screen Shot 2022-03-30 at 9 40 52 AM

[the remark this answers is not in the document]
duncan-truss: Super, thanks for including it. I can mess with the grid container and spacing on the Review Cypress PR if needed.

PR #8305, opened 2022-03-18
MB-11972: 3.14.4 addresses CVE-2022-0778

ahobson:
Use 3.14.4 for now while 3.15.x gets sorted out

### DIFF
--- a/Dockerfile.e2e
+++ b/Dockerfile.e2e
@@ -1,4 +1,4 @@
-FROM alpine:3.15.0
+FROM alpine:3.14.4
 
 # hadolint ignore=DL3017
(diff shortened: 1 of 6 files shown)


rogeruiz:
It looks like the update to `3.15.x` was done automatically with [at]Dependabot. If we would like this in sooner, rather than later, then maybe it makes sense to downgrade to `3.14.x`. I'm not sure on what the difference is between those minor versions, but if it's not important to our project, I'm all for the path of least resistance here.

[=> Dependabot upgrading to 3.15.0 from 3.14.2](https://github.com/transcom/mymove/pull/7788)

PR #8313, opened 2022-03-21
MB-11598 cypress ppm estimated incentive

ronaktruss:
## Summary

Adding basic cypress tests for the estimated incentive page.

Planning on having a secondary PR that adds some additional data-testids and uses them as selectors for the other cypress tests and begins an entire ppm flow while utilizing the back button to navigate back to previous pages.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the server locally.

```sh
make server_run
```

##### Terminal 2

Start cypress locally.

```sh
make e2e_test_fresh
```

</details>

### Frontend

## Screenshots

image


pearl-truss:
I saw your note in the description and from my understanding we're actually trying to move away from using data-testids. And @duncan-truss may know more about this

cypress/integration/milmove/ppms/estimatedIncentive.js
```diff
@@ -0,0 +1,38 @@
+describe('PPM Onboarding - Estimated Incentive', function () {
+  before(() => {
+    cy.prepareCustomerApp();
+  });
+
+  beforeEach(() => {
+    cy.intercept('GET', '**/internal/moves/**/mto_shipments').as('getShipment');
+    cy.intercept('PATCH', '**/internal/mto-shipments/**').as('patchShipment');
+
+    const userId = '4512dc8c-c777-444e-b6dc-7971e398f2dc';
```
pearl-truss, line 10: Other cypress files have included a comment above the user to indicate which user it is, i.e 
```
// estimated_weights@ppm.unsubmitted
```

cypress/integration/milmove/ppms/estimatedIncentive.js
```diff
@@ -0,0 +1,38 @@
...
+  });
+
+  it('mobile - go to estimated incentives page', () => {
+    cy.viewport(479, 875);
+    navigateToEstimatedIncentivePageAndVerifyAmount();
+    cy.get('[data-testid="ppm-estimated-incentive-next"]').should('not.be.disabled').should('have.css', 'order', '1');
+  });
+});
+
+function navigateToEstimatedIncentivePageAndVerifyAmount() {
```
pearl-truss, line 27: It might also be worth testing the the button on this page is enabled and navigates to the correct url

cypress/integration/milmove/ppms/estimatedIncentive.js
```diff
@@ -0,0 +1,38 @@
...
+    cy.prepareCustomerApp();
+  });
+
+  beforeEach(() => {
+    cy.intercept('GET', '**/internal/moves/**/mto_shipments').as('getShipment');
+    cy.intercept('PATCH', '**/internal/mto-shipments/**').as('patchShipment');
+
+    const userId = '4512dc8c-c777-444e-b6dc-7971e398f2dc';
+    cy.apiSignInAsUser(userId);
+    cy.wait('@getShipment');
```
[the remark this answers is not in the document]
ronaktruss: I was running into an issue where the `GET` call for getting mto_shipments details was being called twice and would occasionally cause ETag issues if they both didn't resolve before the first `PATCH` call. The two GET calls happen on the initial landing page (when `apiSignInAsUser` is called) and when we navigate to edit an existing PPM (clicking on `shipment-list-item-container`).

I put this in the beforeEach because it would be duplicate code for the two tests. We could have this code exist as part of the `navigateToEstimatedIncentivePageAndVerifyAmount` function but I initially didn't because I wasn't necessarily testing the the initial sign in portion using the mobile view. There likely isn't an issue with using the mobile view for the entirety of the testing so I think I will be moving this to the `navigateToEstimatedIncentivePageAndVerifyAmount` function.

felipe-lee:
> I saw your note in the description and from my understanding we're actually trying to move away from using data-testids. And @duncan-truss may know more about this

Yeah, seconding what @pearl-truss said. It was discussed in [#g-testing](https://ustcdp3.slack.com/archives/CNTE04R5Z) (wish I could link to a thread, but it was just being discussed at a channel level). Seems to be the general sentiment is to not use `data-testid`s unless they already exist, or help with flakiness. In the cases used here, I think we can probably skip adding more `data-testid`s since we've been adding other cypress tests for the same forms that haven't needed the ids.

ronaktruss:
Fwiw, my understanding of the discussion was that there was an indication that folks would be okay not use `data-testid` and that we weren't required to use it. I did not read it as we shouldn't be using or adding `data-testid` or as a definitive we should be using the RTL philosophy on cypress tests.

I ended up adding them in locations where we were relying on text or where they were tightly coupled with the dom to find elements (as they seemed potentially brittle), but didn't update the previous queries. I was planning on updating past usages of these to utilize the same queries.

Is the ask here to remove the usages of data-testid that I added (when another query would suffice) or just a double check to not overzealously replacing the existing queries with data-testid?

cypress/integration/milmove/ppms/estimatedIncentive.js
```diff
@@ -0,0 +1,38 @@
...
+  cy.get('[data-testid="shipment-list-item-container"]').click();
+  cy.wait('@getShipment');
+
+  cy.get('[data-testid="ppm-date-and-location-submit"]').should('not.be.disabled').click();
+  cy.wait('@patchShipment');
+
+  cy.get('[data-testid="ppm-estimated-weights-submit"]').should('not.be.disabled').click();
+  cy.wait('@patchShipment');
+
+  cy.get('[data-testid="ppm-estimated-incentive-amount-sentence"]').contains(/\$\d{1,3}(?:,\d{3})*? is/);
```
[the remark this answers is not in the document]
duncan-truss: I know the pricing work is left to do but this value should be what's already stored in the shipment database if we're not updating inputs on the prior pages right?  I would say we're better off with using the expected value than matching some possibly incorrect value that passes because it is still in the right format.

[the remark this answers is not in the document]
ronaktruss: Yeah, you're probably right. For some reason I thought there might be the chance for this value to change when we update (even without changing the previous info) based on some time related fields, but this is not the case. 

I'll make the change.

felipe-lee:
> Fwiw, my understanding of the discussion was that there was an indication that folks would be okay not use `data-testid` and that we weren't required to use it. I did not read it as we shouldn't be using or adding `data-testid` or as a definitive we should be using the RTL philosophy on cypress tests.
> 
> I ended up adding them in locations where we were relying on text or where they were tightly coupled with the dom to find elements (as they seemed potentially brittle), but didn't update the previous queries. I was planning on updating past usages of these to utilize the same queries.
> 
> Is the ask here to remove the usages of data-testid that I added (when another query would suffice) or just a double check to not overzealously replace the existing queries with data-testid?

I think to remove the `data-testid`s that were added and use the other available queries instead. Like yeah, the idea behind `data-testid`s is to reduce how brittle tests are, but as was pointed out in the channel, our structure and wording don't really change, so that's less of a concern for us. And it seems that in general projects at Truss focus more on RTL's point of view of testing what the user interacts with, rather than relying on something that user's won't interact with.

pkg/testdatagen/scenario/shared.go
```diff
@@ -695,7 +695,7 @@ func createUnsubmittedMoveWithPPMShipmentThroughEstimatedWeights(appCtx appconte
 			ID:                 uuid.Must(uuid.FromString("65eea403-89ac-4c2d-9b1c-0dcc8805258f")),
 			EstimatedWeight:    models.PoundPointer(unit.Pound(4000)),
 			HasProGear:         models.BoolPointer(false),
-			EstimatedIncentive: models.Int32Pointer(int32(1000000)),
+			EstimatedIncentive: models.Int32Pointer(int32(10000)),
```
ronaktruss, line 698: Modified this because, this is the value that we are setting on the backend so that this test doesn't break when we navigate to the page. This and the actual amount verified in the test should be modified if #8303 is merged prior.

We will run into some of these minor inconveniences if we are checking the actual value prior to setting up the actual logic for this.

cypress/integration/milmove/ppms/estimatedIncentive.js
```diff
@@ -0,0 +1,52 @@
...
+function navigateToEstimatedIncentivePageAndVerifyFields(isMobile = false) {
+  // estimated_weights@ppm.unsubmitted
+  const userId = '4512dc8c-c777-444e-b6dc-7971e398f2dc';
+  cy.apiSignInAsUser(userId);
+  cy.wait('@getShipment');
+
+  cy.get('[data-testid="shipment-list-item-container"]').click();
+  cy.wait('@getShipment');
+
+  cy.get('h1').should('contain', 'PPM date & location');
```
ronaktruss, line 30: These should assertions are needed to increase the stability of the tests. I was running into instances where the queries would attempt to run on the page prior to the UI being updated even though there was a wait. 

If there is a better way to handle this please let me know.

cypress/integration/milmove/ppms/estimatedIncentive.js
```diff
@@ -0,0 +1,52 @@
...
+  cy.get('button').contains('Save & Continue').should('not.be.disabled').click();
+  cy.wait('@patchShipment');
+
+  cy.get('h1').should('contain', 'Estimated weight');
+  cy.get('button').contains('Save & Continue').should('not.be.disabled').click();
+  cy.wait('@patchShipment');
+
+  cy.get('h1').should('contain', 'Estimated incentive');
+  cy.get('.container h2').contains(/\$\d{1,3}(?:,\d{3})*? is/);
+  cy.get('.container h2').contains('$100');
```
ronaktruss, line 40: We are querying the same element here, but I am planning on refactoring this in the #8321 to separate out a generic check (regex) and the specific check(actual value).

cypress/integration/milmove/ppms/estimatedIncentive.js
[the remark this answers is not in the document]
pearl-truss: Thanks! That makes sense :)

pearl-truss (approved):
LGTM

pkg/testdatagen/scenario/shared.go
felipe-lee: Yeah, I'm looking forward to us having this be a little less brittle in the future haha

felipe-lee (approved):
Awesome, thanks for making those updates, LGTM

PR #8315, opened 2022-03-21
MB-11970 hide deleted shipments

leia-sefkin:
## Summary

Quick query fix to remove deleted shipments from our shipment list fetcher. 

Note that the bug filed some time ago mentions a use case for showing deleted shipments via admin or other APIs. Since this currently is not an issue, for simplicity's sake I chose this path. If in the future we need to return deleted shipments for any reason we can create a new service endpoint (or decide at that time what is best to do). 

## Setup to Run Your Code

- [ ] Populate test data (to have moves to work with, or alternatively create your own manually) - `make db_dev_fresh`, `make db_dev_e2e_populate` 

- [ ] Login as a customer, create shipments if necessary, remove one and it should no longer show up. 
- [ ] As a service counselor remove a shipment from from a move and it should no longer show up.


pkg/services/query/query_builder.go
```diff
@@ -219,8 +222,17 @@ func filteredQuery(query *pop.Query, filters []services.QueryFilter, t reflect.T
 
 		// Column lookup should always adhere to SQL injection input validations
 		// https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/SQL_Injection_Prevention_Cheat_Sheet.md#defense-option-3-whitelist-input-validation
-		columnQuery := fmt.Sprintf("%s %s ?", f.Column(), f.Comparator())
-		query = query.Where(columnQuery, f.Value())
+
+		var columnQuery string
+
+		if f.Comparator() == isNull {
```
leia-sefkin, line 228: not the prettiest, but i didn't want to spend a long time over engineering here when we are looking at refactoring this service out

src/components/Customer/Review/Summary/Summary.jsx
```diff
@@ -272,15 +271,8 @@ function mapStateToProps(state) {
   };
 }
 
-function mapDispatchToProps(dispatch, ownProps) {
```
leia-sefkin, line 275: the call here to loadMove was overriding shipment data and other props that get loaded already in the lines above so i removed it altogether.

mboren (approved):
Looks good, works for me. I don't know enough Redux to understand the `mapDispatchToProps` part, so you might want to get a second opinion on that though.

PR #8317, opened 2022-03-21
MB-11814-advance-backend

NamibiaTorres:
## Summary
This PR is complete when:

- [x] The properties advanceRequested and advance should be added to the UpdatePPMShipment request definition in internal.yaml.

- [x] The property advance should be added to the PPMShipment response definition in internal.yaml

- [x] The payload to model functions and model to payload functions for PPM shipments should be updated to transform values from one to the other.

- [x] Add a boolean pointer field to the PPMShipment model for the advance_requested field that exists in the database but is not defined on the model.

- [x] In the ppm_shipment_updater service the [mergePPMShipment](https://github.com/transcom/mymove/blob/2b64113c86611a3445383b331781fa3ac751abcc/pkg/services/ppmshipment/validation.go#L57) method is updated to incorporate the advance fields 

- [x] clear an existing advance value if the advanceRequested is marked as false

- [x] New validation rules areadded to prevent an advance from being set above the estimated incentive amount, an advance with a value less than 1, or advanceRequested being true and with no advanced amount provided.
- [x] A migration for the new field advance_requested and updated comments for advance

## Setup to Run Your Code

<details>
##### Terminal 1

Regerate the server if it is not up to date

```sh
make server_generate
```

##### Terminal 2

Run tests

```sh
go test ./pkg/services/ppmshipment --testify.m TestMergePPMShipment -v

```

</details>

### Backend


pkg/services/ppmshipment/validation.go
```diff
@@ -65,15 +65,17 @@ func mergePPMShipment(newPPMShipment models.PPMShipment, oldPPMShipment *models.
...
-
 	ppmShipment.HasProGear = services.SetNoNilOptionalBoolField(newPPMShipment.HasProGear, ppmShipment.HasProGear)
-
 	ppmShipment.EstimatedWeight = services.SetNoNilOptionalPoundField(newPPMShipment.EstimatedWeight, ppmShipment.EstimatedWeight)
 	ppmShipment.NetWeight = services.SetNoNilOptionalPoundField(newPPMShipment.NetWeight, ppmShipment.NetWeight)
 	ppmShipment.ProGearWeight = services.SetNoNilOptionalPoundField(newPPMShipment.ProGearWeight, ppmShipment.ProGearWeight)
 	ppmShipment.SpouseProGearWeight = services.SetNoNilOptionalPoundField(newPPMShipment.SpouseProGearWeight, ppmShipment.SpouseProGearWeight)
-
 	ppmShipment.EstimatedIncentive = services.SetNoNNilOptionalInt32Field(newPPMShipment.EstimatedIncentive, ppmShipment.EstimatedIncentive)
+	ppmShipment.Advance = services.SetNoNilOptionalCentField(newPPMShipment.Advance, ppmShipment.Advance)
```
pearl-truss, line 74: Does `ppmShipment.AdvanceRequested` also need to be set in this function?

pearl-truss (commented):
What I saw looks good! I just left a comment about the `validation.go` file. 

And I also believe the `AdvanceRequested` field should be added in `payload_to_model.go` and `model_to_payload` under `handlers/internalapi/internal/payloads`

pkg/services/ppmshipment/validation.go
duncan-truss: Yea I think this needs to take into account clearing a previously requested advance too.  So if advanceRequested was previously set to true and advance had a value but now we want advanceRequested to be false and the advance value gets cleared (set back to nil)

migrations/app/migrations_manifest.txt
```diff
@@ -711,3 +711,4 @@
 20220304192644_archive-reimbursements-table.up.sql
 20220308211442_update_origin_duty_location_to_gbloc_view.up.sql
 20220308221248_fix_duty_station_triggers.up.sql
+20220321233308_add_advance_requested.up.sql
```
felipe-lee, line 714: Can you commit the migration?

pkg/services/ppmshipment/validation.go
```diff
@@ -65,15 +65,17 @@ func mergePPMShipment(newPPMShipment models.PPMShipment, oldPPMShipment *models.
...
+
+	if newPPMShipment.Advance != nil {
+		ppmShipment.Advance = newPPMShipment.Advance
+	}
```
felipe-lee, line 78: I think this might be partially duplicating what line 74 is doing above.

pkg/services/ppmshipment/validation_test.go
```diff
@@ -35,6 +36,8 @@ func setupShipmentData() (data dataSetup) {
 		HasProGear:          models.BoolPointer(true),
 		ProGearWeight:       &proGearWeight,
 		SpouseProGearWeight: &spouseProGearWeight,
+		Advance:             &advanceCents,
```
felipe-lee, line 39: The function Pearl added in #8303 here: https://github.com/transcom/mymove/pull/8303/files#diff-d6b58d2dc6179aa48c22df923c9cb4d145428d1f699b64f05e439c80a2caeb88R60-R64 could be used here in your PR. Just letting you know you could copy that function if you wanted to be able to use it on this line, but it's not something that actually needs to happen.

swagger-def/internal.yaml
```diff
@@ -2669,6 +2675,18 @@ definitions:
       estimatedIncentive:
         type: integer
         x-nullable: true
+      advance:
+        description: >
+          The amount request for an advance, or null if no advance is requested
+        type: integer
+        format: cents
+        x-nullable: true
+        x-omitempty: false
```
felipe-lee, line 2684: I think you can leave the `x-omitempy` out of this since we don't want to require the key be passed in when the value is not being updated

migrations/app/migrations_manifest.txt
NamibiaTorres: Yes, it was hiding in the Unversioned FIles section

pkg/handlers/converters.go
```diff
@@ -190,6 +190,15 @@ func FmtCost(c *unit.Cents) *int64 {
...
 
+// FmtCostToInt64 converts pop type to go-swagger type
+func FmtCostToInt64(c *int64) *unit.Cents {
+	if c == nil {
+		return nil
+	}
+	cost := unit.Cents(*c)
+	return &cost
+}
+
```
felipe-lee, line 201: We have a function for this in the `handlers/converters_strfmt.go` file: https://github.com/transcom/mymove/blob/c60e4594ed1c060060c119844d39de3858af0eae/pkg/handlers/converters_strfmt.go#L45-L53

migrations/app/migrations_manifest.txt
felipe-lee: Awesome, thanks!

pkg/services/ppmshipment/validation.go
```diff
@@ -65,15 +75,22 @@ func mergePPMShipment(newPPMShipment models.PPMShipment, oldPPMShipment *models.
...
-
 	ppmShipment.EstimatedIncentive = services.SetNoNNilOptionalInt32Field(newPPMShipment.EstimatedIncentive, ppmShipment.EstimatedIncentive)
+	ppmShipment.Advance = services.SetNoNilOptionalCentField(newPPMShipment.Advance, ppmShipment.Advance)
+	ppmShipment.AdvanceRequested = services.SetNoNilOptionalBoolField(newPPMShipment.AdvanceRequested, ppmShipment.AdvanceRequested)
+
+	if newPPMShipment.Advance != nil {
+		ppmShipment.Advance = newPPMShipment.Advance
+	}
+
+	if !*newPPMShipment.AdvanceRequested {
```
pearl-truss, line 91: I believe this is causing an `invalid memory address or nil pointer dereference` I'm seeing when the frontend is integrated in the backend in this PR https://github.com/transcom/mymove/pull/8303

pkg/services/ppmshipment/validation.go
```diff
@@ -74,7 +74,7 @@ func mergePPMShipment(newPPMShipment models.PPMShipment, oldPPMShipment *models.
 	ppmShipment.Advance = services.SetNoNilOptionalCentField(newPPMShipment.Advance, ppmShipment.Advance)
 	ppmShipment.AdvanceRequested = services.SetNoNilOptionalBoolField(newPPMShipment.AdvanceRequested, ppmShipment.AdvanceRequested)
 
-	if ppmShipment.AdvanceRequested == nil && !*ppmShipment.AdvanceRequested {
+	if ppmShipment.AdvanceRequested == nil || !*ppmShipment.AdvanceRequested {
```
felipe-lee, line 77: Ah, I think this might need to be 
```go
if ppmShipment.AdvanceRequested != nil && !*ppmShipment.AdvanceRequested
```

pkg/services/ppmshipment/rules.go
```diff
@@ -76,3 +76,36 @@ func checkRequiredFields() ppmShipmentValidator {
...
+
+		if newPPMShipment.AdvanceRequested == nil && newPPMShipment.Advance == nil {
+			return verrs
+		}
+
+		if !*newPPMShipment.AdvanceRequested && newPPMShipment.Advance == nil {
+			return verrs
+		}
+
+		if !*newPPMShipment.AdvanceRequested && newPPMShipment.Advance != nil {
```
pearl-truss, line 93: This creates a nil pointer dereference error

pkg/handlers/converters_strfmt.go
```diff
@@ -66,3 +66,13 @@ func FmtUUIDToPop(u strfmt.UUID) uuid.UUID {
...
+
+// PoundPtrFromInt64Ptr converts go-swagger type to pop type
+func PoundPtrFromInt64Ptr(num *int64) *unit.Pound {
+	if num == nil {
+		return nil
+	}
+	value := int(*num)
+	pound := unit.Pound(value)
+	return &pound
+}
```
felipe-lee, line 78: A note for reviewers, this function is still the same, we just moved it because it was in the wrong converters file. It's still the same package so no other code change was needed related to this.

pkg/services/ppmshipment/rules.go
```diff
@@ -76,3 +76,42 @@ func checkRequiredFields() ppmShipmentValidator {
...
+
+		if float64(*newPPMShipment.Advance) > float64(*newPPMShipment.EstimatedIncentive)*0.6 {
+			verrs.Add("advance", "Advance can not be greater than 60% of the estimated incentive")
+		}
+
+		if float64(*newPPMShipment.Advance) < float64(1) {
+			verrs.Add("advance", "Advance can not be  value less than 1")
+		}
+
+		if *newPPMShipment.AdvanceRequested && *newPPMShipment.Advance == 0 {
```
ronaktruss, line 111: Do we need this check? Is the check above not sufficient? 

Also shouldn't we be checking `newPPMShipment.Advance` to see if this is nil before dereferencing? (While the UI doesn't allow this, we should generally make the server validation agnostic of the validations on the FE.)

duncan-truss:
When I integrate with Pearl's PR I get this validation error when going over the 60% of the estimated incentive.  My understanding was this is supposed to be a soft validation to match the frontend and that we would only return an error if it exceeded the estimated incentive value itself. @pearl-truss 

Screen Shot 2022-03-29 at 11 50 51 AM

pearl-truss:
> When I integrate with Pearl's PR I get this validation error when going over the 60% of the estimated incentive. My understanding was this is supposed to be a soft validation to match the frontend and that we would only return an error if it exceeded the estimated incentive value itself. @pearl-truss
> 
> Screen Shot 2022-03-29 at 11 50 51 AM

I'm not seeing that error in my branch that has some of the changes from this branch pulled in
Screen Shot 2022-03-29 at 12 28 41 PM

pkg/services/ppmshipment/rules.go
```diff
@@ -76,3 +76,42 @@ func checkRequiredFields() ppmShipmentValidator {
...
+		if !*newPPMShipment.AdvanceRequested && newPPMShipment.Advance == nil {
+			return verrs
+		}
+
+		if !*newPPMShipment.AdvanceRequested && newPPMShipment.Advance != nil {
+			verrs.Add("advance", "Advance must be nil if advance requested is false")
+		}
+
+		if float64(*newPPMShipment.Advance) > float64(*newPPMShipment.EstimatedIncentive)*0.6 {
+			verrs.Add("advance", "Advance can not be greater than 60% of the estimated incentive")
```
pearl-truss, line 104: I tracked down the issue that @duncan-truss flagged (I was able to reproduce it by pressing the submit button)

Seems be triggered by this line. I think we don't want an explicit rule here because it should be possible for them to submit with a requested advance over 60%

Screen Shot 2022-03-29 at 12 36 03 PM

felipe-lee: Ah right, the hard limit is 100% right?

pearl-truss (approved):
LGTM

pkg/services/ppmshipment/rules.go
pearl-truss: yeah

gidjin (approved):
I reviewed the DB changes in the other PR #8303 so marking the DB work good to go.

duncan-truss (approved):
🎉 Re-tested the validation rule and all looks well.

PR #8321, opened 2022-03-22
MB-10714: Initial Refactor ppm cypress

ronaktruss:
## Summary

This is just proposing one way of handling the future refactors so that we can adopt the patterns a bit earlier. The pattern for mobile tests should likely be reviewed with more scrutiny as if we need different viewports for specific tests we would need to override the current value throughout our tests. If this happens frequently it may warrant a modified approach.   

## Setup to Run Your Code

<details>

##### Terminal 1

Start the server locally.

```sh
make server_run
```

##### Terminal 2

Start cypress locally.

```sh
make e2e_test_fresh
```

</details>

### Frontend

### Backend


pearl-truss:
> @pearl-truss In [this commit](https://github.com/transcom/mymove/pull/8321/commits/f4521f7a32ebc050766f862a299dc83909b34de2) I ended up removing the logout because it didn't seem necessary for the flow and seemed to make the tests take longer to complete. It seems like just navigating back to the base url should suffice. Let me know if the logouts were needed for any other reason. (I also added some additional waits to resolve ETag issues that were similar to the ones I was seeing in my [PR](https://github.com/transcom/mymove/pull/8313).

Sounds good to me, I don't think the logging out is necessary with the new set up

felipe-lee:
@ronaktruss I think something went wrong with the latest merges because it's showing a bunch of changes that I'm pretty sure were from PRs merged to master.

ronaktruss:
@felipe-lee should be resolved. Thanks for calling that out!

cypress/integration/milmove/ppms/entireShipment.js
```diff
@@ -0,0 +1,117 @@
...
+  it('happy path with new shipment', () => {
+    // profile_full_ppm@move.draft
+    const userId = '9b9ce6ed-70ba-4edf-b016-488c87fc1250';
+    navigateHappyPath(userId, false);
+  });
+
+  it('mobile - happy path with new shipment', () => {
+    // full_ppm_mobile@complete.profile
+    const userId = '4fd6726d-2d05-4640-96dd-983bec236a9c';
+    cy.viewport(479, 875);
```
pearl-truss, line 28: I wonder if it's worth making these constants i.e `mobileWidth` and `mobileHeight` that can be used outside of this test as well

ronaktruss: +1 it's probably worth considering. I was holding off on making this as I wasn't sure how other mobile test would want to be run. I chose the 479 because <480 is when we modify the order on the buttons on the estimated incentive page. If we end up doing this for the rest of the tests we should go the route of having these variables be constants,

cypress/integration/milmove/ppms/dateAndLocation.js
```diff
@@ -1,11 +1,12 @@
+import { customerChoosesAPPMMove, submitsDateAndLocation } from '../../../support/ppmShared';
```
felipe-lee, line 1: Could this be an absolute import? Same question about the other files, but don't want to spam the PR with the same comment over and over.

cypress/integration/milmove/ppms/entireShipment.js
```diff
@@ -0,0 +1,117 @@
...
+  cy.apiSignInAsUser(userId);
+  cy.wait('@getShipment');
+  customerChoosesAPPMMove();
+  submitsDateAndLocation();
+  submitsEstimatedWeightsAndProgear();
+  generalVerifyEstimatedIncentivePage(isMobile);
+}
+
+function navigateHappyPathWithEditsAndBacks(isMobile = false) {
+  // TODO: need to change id to be unique + add email associated with user
```
felipe-lee, line 52: It might be good to do this since now. It would also simplify the `submitAndVerifyUpdateDateAndLocation` since you could re-use `submitsDateAndLocation` for part of it at least.

pkg/testdatagen/scenario/e2ebasic.go
```diff
@@ -1523,48 +1523,70 @@ func serviceMemberWithOrdersAndPPMMove04(appCtx appcontext.AppContext, userUploa
...
-		ServiceMember: models.ServiceMember{
-			ID: uuid.FromStringOrNil("a5cc1277-37dd-4588-a982-df3c9fa7fc20"),
-		},
-	})
-	orders := testdatagen.MakeOrder(appCtx.DB(), testdatagen.Assertions{
-		Order: models.Order{
-			ID:              uuid.FromStringOrNil("42f9cd3b-d630-4762-9762-542e9a3a67e4"),
-			ServiceMemberID: customer.ID,
-			ServiceMember:   customer,
+	email := "profile_full_ppm@move.draft"
```
felipe-lee, line 1526: I wonder if these functions should have more specific names. I feel like 05, 06, etc. aren't very indicative. Maybe for a future PR though.

felipe-lee (approved):
LGTM, though I think that TODO would be good to address since now, but I'm fine if you want to address it in a future PR.

pkg/testdatagen/scenario/e2ebasic.go
ronaktruss: Yeah, I was thinking the same. It seems like some of these are used for non PPM related items as well, I'll make some adjustments here before merging.

cypress/integration/milmove/ppms/dateAndLocation.js
ronaktruss: Yup will do.

cypress/integration/milmove/ppms/entireShipment.js
ronaktruss: I left the TODO since in my eyes this would test an existing PPM being updated on each page and then navigating back to the page where the changes were made to ensure that the page persisted these changes, and we were still waiting on the advance fields to be finalized.

The reason that there are two functions right now is because one is a POST and the other is for a PATCH. Right now they have different aliases so the line where we wait on them would be different. I should be able to refactor this a bit to reduce the duplicate information.

cypress/integration/milmove/ppms/dateAndLocation.js
ronaktruss: It doesn't look like cypress supports absolute paths out of the box.

felipe-lee: 😞

PR #8322, opened 2022-03-23
[MB-10956] Duty Station Renaming Cleanup Part 4: Rename duty station in backend

mboren:
## Summary
I renamed just about every permutation of `duty station` that I could find in the backend.

Most of the heavy lifting was done in prior PRs so this is mostly local variables, comments, and stuff like that.
  
There are two things I didn't get to
- the code for loading new duty stations from  a CSV and generating duty station migrations.
  - I don't think this stuff would even be useful anymore without heavy modification. It's maybe still worth updating just for the sake of consistency, but I'll leave that for another PR since it's not a part of the milmove application and would inflate this  PR a lot
- the query parameter for sorting office queues by destination duty station `destinationDutyStation`
  - This requires changes done in both the frontend and backend simultaneously, and I'd like to keep this PR at least a little focused. Also, it doesn't look to me like we allow sorting by this anymore, so might just delete it.

## Setup to Run Your Code
This touches a lot of stuff. Fortunately, I think the compiler and automated tests will catch most of the mistakes that would have come up from renaming things.
There are a few things that still probably a good idea to walk through manually though.

### Make sure e2e populate still works:
```sh
make db_dev_e2e_populate
```
### Walk a move through the different stages and make sure nothing weird happens
```
make server_run
```
```
make client_run
```
- Create a move as a customer
- Approve as TOO
- Submit payment request as the Prime
- Approve the payment request as a TIO.


leia-sefkin (approved):
Another beast of a PR 🐻  - code changes look good, everything works find on the customer and TOO/TIO/Prime side of things when testing locally. 🚢  🚢  🚢

PR #8324, opened 2022-03-23
[MB-8450] Edit office app config do that refetching data on window refocus is disabled

hannaliebl:
## Summary

I'm tagging the senior eng. group as well as Roci since this is a very small config change, but does change how data gets refetched in the office app.

This change is based on numerous bug reports in the office app that refocusing a browser tab results in the generic error page. I think this is happening because our app was configured to have refetching data set to true on window refocus (see: https://react-query.tanstack.com/guides/window-focus-refetching). We discussed this at the Milmove frontend working group a few weeks ago, and we didn't really see any benefit to this behavior in our app, but I do think it's worth thinking about this change and if it has any unintended consequences.

Here are other similar reports that I think are all the same issue:

https://dp3.atlassian.net/browse/MB-8473
https://dp3.atlassian.net/browse/MB-8448
https://dp3.atlassian.net/browse/MB-8447

This doesn't address the network failures themselves, but hopefully this will give a little more visibility into the actual failures that do happen (because the app won't just refetch data to mask it, or the moment you try to pull up the network tab to see what failed.) I, personally, wasn't able to find any consistent steps to reproduce any network failures in staging, for what it's worth.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Just ensure the app works as expected. You should not see any more network requests on window refocus.


rogeruiz (approved):
lgtm 🌈

let's see how this affects bug reporting and UAT testing going forward.

hannaliebl:
@mergifyio refresh

hannaliebl:
@Mergifyio requeue

hannaliebl:
@Mergifyio unqueue

PR #8330, opened 2022-03-23
[MB-11452] update rds certs for dp3 sites

sheenamt:
## Summary
AWS is requiring an update the RDS certificates. This adds the new certificate to the dockerfiles for the dp3 environments (`exp`, `demo`, `loadtest`)

As written, this is only _adding_ a new cert, not removing anything. Once the new dockerfiles are built, the change to the new cert will be made to the `exp` environment in AWS by infra and app connectivity will be tested before old certs are removed or this change is rolled out to any other environments.

### DIFF
--- a/Dockerfile.dp3
+++ b/Dockerfile.dp3
@@ -1,8 +1,9 @@
 # hadolint ignore=DL3007
 FROM gcr.io/distroless/base:latest
 
-COPY bin/rds-ca-2019-root.pem /bin/rds-ca-2019-root.pem
-COPY bin/rds-ca-us-gov-west-1-2017-root.pem /bin/rds-ca-us-gov-west-1-2017-root.pem
+COPY bin/rds-ca-rsa4096-g1.pem /bin/rds-ca-rsa4096-g1.pem
+# COPY bin/rds-ca-2019-root.pem /bin/rds-ca-2019-root.pem
+# COPY bin/rds-ca-us-gov-west-1-2017-root.pem /bin/rds-ca-us-gov-west-1-2017-root.pem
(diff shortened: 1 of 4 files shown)


sheenamt:
Whoops, forgot that I had to tell circle to do things for exp specifically.

.circleci/config.yml
```diff
@@ -40,7 +40,7 @@ references:
   #
   # In addition, it's common practice to disable acceptance tests and
   # ignore tests for exp deploys. See the branch settings below.
-  exp-branch: &exp-branch placeholder_branch_name
+  exp-branch: &exp-branch MB-11452-rds-certs
```
ahobson, line 43: This should be the `placeholder_branch_name` when merging into `master`

.circleci/config.yml
```diff
@@ -54,7 +54,7 @@ references:
   # set integration-mtls-ignore-branch to the branch if you want to
   # IGNORE mtls integration tests, or `placeholder_branch_name` if you
   # do want to run them
-  integration-mtls-ignore-branch: &integration-mtls-ignore-branch placeholder_branch_name
+  integration-mtls-ignore-branch: &integration-mtls-ignore-branch MB-11452-rds-certs
```
ahobson, line 57: This should be the `placeholder_branch_name` when merging into `master`

ahobson:
Thank you for doing this! I think if you update the branches, the branch protection checks will run for the right branch and we can get this merged

.circleci/config.yml
sheenamt: Fixed

ahobson (approved):
🚀

PR #8341, opened 2022-03-24
[ MB-12041 ] Refactor AuditableAppCtx

rogeruiz:
## Summary

This change is only adding a new `AuditableAppContextFromRequestWithErrors` method and interface. There are also some `go fmt` changes introduced here in a separate commit. 

[this conversation](https://github.com/transcom/mymove/pull/8304#discussion_r833397487) explains more about the approach used.

## Setup to Run Your Code

Open the file `pkg/handlers/contexts.go` and read through the new code. I will also add comments to specific parts of the code changes to better explain what's happening here.

**It's important to get this merged into the default branch because it is blocking multiple tickets that will need to be updated to use this new method**.


pkg/handlers/contexts.go
```diff
@@ -135,6 +139,40 @@ func (hctx *handlerContext) AuditableAppContextFromRequest(r *http.Request, hand
 	return resp
 }
 
+// AuditableAppContextFromRequestWithErrors creates a transaction and sets local
+// variables for use by the auditable trigger and also allows handlers to return errors.
+func (hctx *handlerContext) AuditableAppContextFromRequestWithErrors(
```
rogeruiz, line 151: This is the new method being added. And eventually, this is the method that will exist after everything is refactored.

pkg/handlers/contexts.go
```diff
@@ -135,6 +139,40 @@ func (hctx *handlerContext) AuditableAppContextFromRequest(r *http.Request, hand
...
+	r *http.Request,
+	handler func(appCtx appcontext.AppContext) (middleware.Responder, error),
```
rogeruiz, line 153: The `handler` function now returns a tuple of both the `middleware.Responder` and an `error`.

pkg/handlers/contexts.go
```diff
@@ -135,6 +139,40 @@ func (hctx *handlerContext) AuditableAppContextFromRequest(r *http.Request, hand
...
+		if err != nil {
+			return err
+		}
+		eventName := audit.EventNameFromContext(r.Context())
+		err = txnAppCtx.DB().RawQuery("SET LOCAL audit.current_event_name = '" + eventName + "'").Exec()
+		if err != nil {
+			return err
+		}
+		resp, err = handler(txnAppCtx)
+		return err
```
rogeruiz, line 175: The error is returned here rather than `nill` which will allow for Pop to automatically rollback database transactions if there is an error.

rogeruiz (commented):
Leaving some comments for further scrutiny and review.

rogeruiz:
I need to run `make server_generate mock_generate` to get the tests to pass.

YanZ777 (approved):

PR #8347, opened 2022-03-25
MB-12031 Add ADR for fake data generators and factories

felipe-lee:
## Summary

This adds an ADR around using fake data generators and factories. The ADR itself should have all the details hopefully. 

Some things to note:

* I'm not sure if this should be a single ADR, or if we should have separate ones for the BE vs the FE. Or maybe it should be more generic? Though making it generic would make it a bit vaguer about pros/cons for each option (and would change the options to some degree).
* @ahobson brought up at a backend meeting was that one thing to look into would be to generate data based on our API spec.
  * I think this would be great to have, and is related to this, but not the only bit to this. Like this would be good for testing things at an integration level, but when testing for example that the service objects do what is expected with new and old model instances, I personally think that is better served using factories for the models themselves. I do plan on looking more into this though, but wanted to open up the ADR for feedback in the meantime.


docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
```diff
@@ -0,0 +1,157 @@
...
+1. Leave things as they are.
+
+Each of the options that has us creating factories for the FE could also have an alternative of us creating the
+factories from scratch rather than using an existing package, but if we're starting from a blank slate, we might as well
+use a supported package that comes with features we'll want out of the box, rather than coding it ourselves and possibly
+making mistakes or having feature gaps. I.e. the argument for using open source code in general.
+
+## Decision Outcome
+
+Chosen Alternative -> Option 2: Create factories on the FE and use fake data generators for both the BE and the FE.
```
ronaktruss, line 68: Do you have any factory packages in mind for the FE? It might be nice to link out to a couple in the resources even if we don't end up using them

docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
```diff
@@ -0,0 +1,157 @@
...
+
+I'm not 100% sure what this section means...like does this mean how would others keep in line with this ADR? If so, I
+think that would look like people creating/updating factories when models are created/updated. Also updating factories
+for different states of things, e.g. what a service member looks like a different stages of on-boarding, or what a
+shipment looks like at different stages of the move process.
+
+### Ownership
+
+AppEng would own this since it's related to the way we test our application code.
+
```
felipe-lee, line 52: Note that this section is based on a branch @rogeruiz has for changing the ADR template: https://github.com/transcom/mymove/compare/rogeruiz/refactor-adr-template

I figured I could try it out and see what people thought.

docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
felipe-lee: Ah good point, I can add those

felipe-lee: added in https://github.com/transcom/mymove/pull/8347/commits/b9deb12591756d95c3c59358dd51fb813f3349d9

docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
```diff
@@ -0,0 +1,160 @@
...
+semi-realistic state, but they have flaws and inconsistencies. For example, you can't easily call the `testdatagen`
+function for creating an address more than once without passing in `assertions` (overrides), because otherwise you'll
+create the same exact address created twice. To help with that, we actually have 4 different address creation functions
+so that you don't need to always pass in assertions, but that's not a great solution. In a sense, these are already
+factories, but could use some fixing up to generate realistic fake data without needing overrides every time you want
+non-hard-coded data.
+
+As for the frontend, we tend to create fake objects that kind of look like the data within each test file, which again
+goes back to the issue of needing to understand what the data _should_ look like. We end up with many tests that have
+fake objects that don't really look like the data will, e.g. many tests have IDs for objects that are just a few digits
```
YanZ777, line 27: > that don't really look like the data will

I think you meant to say "that don't really look like what the data actually looks like" or "that don't really look like what the data will look like"

docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
```diff
@@ -0,0 +1,160 @@
...
+rather than a UUID. There are plenty of times that it doesn't matter too much, but it can make it that much harder for
```
YanZ777, line 28: > but the new tests need more realistic data.

This part of the sentence doesn't make sense in the context of the overall sentence.

docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
```diff
@@ -0,0 +1,160 @@
...
+someone that is looking at existing tests as examples in order to write new tests, but the new tests need more realistic
+data.
+
+## Measuring success
+
+Initial success would probably look like us having a decent number of core factories set up and team members feeling
```
YanZ777, line 34: I'd just take out the "probably" here.
Can "decent number" be defuzzed here? 
I'd like to hear a better definition of core definition as well. At this point in the ADR, I have a vague idea of what you mean.
I also think a better phrasing for "team members feeling comfortable" would be:
"team members know to use the core factories when available in their tests"

docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
```diff
@@ -0,0 +1,160 @@
...
+comfortable using them in their tests.
+
+Long-term, we would ideally have replaced our standalone fake objects in FE tests as well as improved our `testadagen`
```
YanZ777, line 37: I'd rephrase
"ideally have replaced" 
to 
"ideally replace"
and change "improved" to "improve"

docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
```diff
@@ -0,0 +1,160 @@
...
+functions that we use for BE and e2e tests.
+
+### Observability
+
+_How will this change be observed by other team members?_
```
YanZ777, line 42: I'm not sure what this section means either. I interpret it to mean that how do we make sure that other people are aware and can see the change happening.

docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
YanZ777: I think another aspect would be to how do we make sure that people actually use this new way of generating data.

docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
```diff
@@ -0,0 +1,160 @@
...
+for different states of things, e.g. what a service member looks like a different stages of on-boarding, or what a
+shipment looks like at different stages of the move process.
+
+### Ownership
+
+AppEng would own this since it's related to the way we test our application code.
+
+## Considered Alternatives
+
+1. Create factories on the FE (using a package), replace the factories on the BE with a package, and use fake data generators for both the BE and FE.
```
YanZ777, line 55: These are all 1.s? 
Maybe they should be bulleted?

YanZ777 (commented):
Left some initial review comments. 
I've only reviewed up to the considered alternatives portion.

docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
felipe-lee: Let me know if the updates I made to this section make sense.

docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
felipe-lee: Ah, that makes sense. I've updated my response then. 

Also since the text changed, for future reviewers Yan was replying this because I'd put I wasn't 100% sure what this section meant. My original text in this section was about how we would keep in line with this ADR (creating/updating factories, including appropriate traits).

docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
felipe-lee: I'll add something about this.

docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
felipe-lee: Github auto-numbers them: image

I mostly gave them numbers to correspond to the numbers in the headings below, but would also be fine with them just being bullets.

YanZ777 (approved):
I mostly have grammar feedback, but the content itself looks good to me.

docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
ahobson: Thank you so much for this ADR!

I wonder if we could use this section to discuss the implementation plan. If this is a best effort migration, then we'll be in both worlds for some time right? And we won't have any automated way (like a linter) to catch folks accidentally using the old pattern, right?

Can you think if any mitigations we might use? I'm brainstorming here, but what if we added a comment above/in all the places that work the old way saying it should be changed?

felipe-lee: Yeah, I don't think we'll be able to use linters easily for this work.

I do like that idea of the comments. Like, it'll be a lot of comments, but it'll serve as a good flag for people looking at existing tests. At least this way they'll know not to just copy that test, or that if they are updating that test, they should update the use of fake data

felipe-lee: I added an implementation plan section separate from the observability section. Mainly because I'm still a bit unsure of what that section means, but I feel like implementation is slightly different from the current text in there. Maybe... 🤷🏼

duncan-truss:
I'm not entirely in the loop on this topic but wanted to share a few thoughts that come to mind that may or may not be helpful:

1. I think some reasons why frontend data is mocked in an ad-hoc way is because every React component is defining a different interface with their props which doesn't always map directly to a data model.  So in practice there is going to be a lot of transformation of data that's modeled after the API for non-page components that do the majority of data fetching.  You're going to also find transformations that happen as well like normalization and selectors in Redux that change payload shapes from regular response data.  We also still have discrepancies like snake case in some places of the internal and ghc APIs that might complicate/confuse things.

2. I do think people get tripped up not knowing what data is or isn't present in different states of the system but I don't know that generating faked data necessarily assists in their understanding or helps them write more robust tests.  I think the required/optional properties of the API level is going to be a bit different than what prop types are required at the component level.  The faker library and api parser were useful in the load testing repo for fuzzing API input and catching some bugs but I think it had limitations when you wanted to write semantically correct load tests that say had deterministic results such as an address that made sense across dependent fields or not sending 2 mto agents of the same type.

Would client tests now have random prop inputs and lead to flakier tests making it harder to reason about their purpose?

Could I just use the functions to generate test data that is pre-populated but that I can still visibly see in the tests and massage into the right shape?

docs/adr/0068-use-fake-data-factories-with-fake-data-generators-to-create-test-data.md
```diff
@@ -0,0 +1,181 @@
...
+This ADR will be announced in meetings where most of the MilMove team is present. If approved, as core factories are set
+up, they could be announced at demos or FE check in meetings.
+
+### Implementation Plan
+
+1. Mark existing fake data with comments indicating it's the old way of doing things. BE/FE order not important.
+   1. Mark existing `testdatagen` functions with comments indicating that they should be updated to use fake data.
+   1. Mark existing fake data in FE tests with comments indicating they they should be updated to use factories if possible.
+      1. This way anyone that looks at it for an example will see that it's using an older pattern.
+      1. Anyone working on those tests will see it and can update it as they edit the tests.
```
ronaktruss, line 61: I could see this being laborious and error prone (unless the plan is to search through files based on file names/extensions and just prepending these comments to all files that match the criteria)

felipe-lee: Luckily all the FE test files should have `.test` in the file name, so that helps for finding them all, but I agree that it would likely be laborious. I think the problem is that it'd be ideal to have some way of indicating what is left to work on, but automated linters wouldn't necessarily work.

gidjin (approved):
LGTM. I think implementation of this will need to be coordinated so we can fit it in with what ever team picks it up. But I like the idea and appreciate the Measuring success and observability considerations.

jenniferlynparsons (approved):

felipe-lee:
> Just my two cents:
> 
> Overall option 2 does seem like the most promising path forward given the current knowledge and state of things (available libraries), but its difficult for me to give this an approval without knowing/seeing the implementation details or getting a sense of the level of effort needed to actually implement these changes. Additionally, the benefit to engineers is not tangible (without seeing actual examples).
> 
> I'd advocate for having this ADR stay in a pending state while we continue to looking into the details to implement the suggested decision outcome. To me the discussion of this ADR is more about deciding what outcome we should spend some time looking into and coming up with a proof of concept PR. Once we do that and can see the actual changes we should actually review the decision and see if it makes sense more generally for the project.

@ronaktruss I haven't gotten to the BE portion, but I have a draft PR showing what the FE factories/builders could look like and updated a test file to use one. 

#8388

PR #8359, opened 2022-03-30
[MB-10956] part 5: Remove sort by destination duty station in backend

mboren:
## [Jira ticket](https://dp3.atlassian.net/browse/MB-10956) for this change
I started to rename this, and then realized that we don't use sort by destination duty location anywhere.
All of the office queues have been switched over to sorting by origin duty location.
So I deleted this dead code instead of renaming the parameter.

## Setup to Run Your Code
Open up the TOO, TIO, and Service Counselor queues and make sure there are no errors.

### DIFF
--- a/pkg/handlers/ghcapi/orders_test.go
+++ b/pkg/handlers/ghcapi/orders_test.go
@@ -379,7 +379,7 @@ func (suite *HandlerSuite) TestUpdateOrderHandlerWithAmendedUploads() {
 		subtestData := suite.makeUpdateOrderHandlerSubtestData()
 		move := subtestData.move
 		order := subtestData.move.Orders
-		destinationDutyStation := order.NewDutyLocation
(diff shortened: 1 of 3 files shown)


pkg/handlers/ghcapi/orders_test.go
```diff
@@ -379,7 +379,7 @@ func (suite *HandlerSuite) TestUpdateOrderHandlerWithAmendedUploads() {
 		subtestData := suite.makeUpdateOrderHandlerSubtestData()
 		move := subtestData.move
 		order := subtestData.move.Orders
-		destinationDutyStation := order.NewDutyLocation
+		destinationDutyLocation := order.NewDutyLocation
```
mboren, line 382: Changes in this file are unrelated to everything else, it's just a use of `destinationDutyStation` that I caught and renamed.

PR #8364, opened 2022-03-31
MB-11350 Allow customers to edit PPM shipments

felipe-lee:
## Summary

Most of the logic for this was implemented in past PRs. This one is mainly fixing up a few things we noticed along the way.

If the PR is hard to review in the "Files changed" tab, you can try going through the commits in order. I try keeping them somewhat atomic. Though there are 2 or so that change what a past commit had done a bit.

#### Main Changes

* Make `PPMShipment.SitExpected` a pointer. 
    * This makes it easier to determine when the API caller is trying to change the value of `SitExpected` vs not.
* Add check in `rules.go` to ensure we do have a non-nil value for `SitExpected`
* Update `mergePPMShipment` to handle `SitExpected` changes.
    * Had been left alone previously because there was no easy way to know what the value should be.
* Update customer home page to properly show edit button for PPM shipments only when the move hasn't  been submitted. Currently, if it's a PPM Shipment, it is always editable.
* Add cypress tests to ensure edit icons are only visible for shipments on the homepage before move submission.
* Add a bare-bones `MakeBaseMTOShipment` creator function.
    * We have a "minimal" one, but it's not really minimal anymore now that we have `PPMShipment`s. The minimal one is used in a lot of places though (~90) so I didn't want to mess with what it did. 
    * `MakeMTOShipmentMinimal` is now a thin wrapper around `MakeBaseMTOShipment`.
* Update `MakePPMShipment` to:
    * account for advances.
    * be a wrapper around `MakeMinimalPPMShipment` to make it easier to see what is actually different between the two. This is an experiment, lmk what y'all think.
* Update `Make*PPMShimpent` funcs to:
    * always use `MakeBaseMTOShipment`.
    * be able to create an `MTOShipment` with a given ID.

#### Quality-of-life Changes

* Add generic functions to `testdatagen/scenario` for creating an un-submitted move w/o any shipments yet, and one for creating an un-submitted move with a PPM shipment (minimal or not).
    * Both can take assertions to make allow more customization.
    * With these, it makes it so that if you need a PPMShipment, you can call one of these functions and only have to define the unique fields, like UUIDs.
* Updated some of the existing PPM funcs in `testdatagen/scenario` to use the new generic functions.
* Update `ShipmentList/index` -> `ShipmentList/ShipmentList` to match changes in previous PRs.
* Update `ShipmentList` tests to use RTL (some were already, just converted the rest).
* Add testdatagen helper to convert a uuid string to a UUID
* Add models helper to convert a UUID literal to a pointer

### Questions

* Should there be validation on the BE to ensure shipments are only edited by customers when the move hasn't been submitted?

### Other possible work
Unlikely this will actually be done in this PR. It's already big as is...

* Remove old PPM logic from `src/pages/MyMove/Home/index.jsx`. Could be done as part of fixing edit logic on Home page.
* Move `mto_shipment` validation functions into `rules.go file` (need to create the `rules.go` file)
* See if removing internal api update validation of status is ok…seems to be duplicating status check in `validation.go`
* Fix `mtoShipmentUpdater.updateShipmentRecord` to use original shipment’s `UpdatedAt` field for e-tag check. It’s using the combined shipment of old and new currently, which works because we don’t overwrite the `UpdatedAt` tag elsewhere, but is a bit weird.

#### Server Tests

<details>
<summary>💻 Run Server Tests</summary>

##### Terminal

```sh
make server_test
```

</details>

#### E2E Tests

<details>
<summary>💻 Run E2E Tests</summary>

##### Terminal

Start the Storybook locally.

```sh
make e2e_test_fresh
```

</details>

##### Additional steps

1. You can then either run all the e2e tests, or filter down using "homepage" in the filter box and then running the mymove spec.

### Checking Things Manually

<details>
<summary>💻 Terminal Setup</summary>

#### Dev DB Data

Populate Dev DB with Data.

```sh
make db_dev_fresh
```

#### Client/Customer UI

Start the UI locally.

```sh
make client_run
```

#### Server

Start the Go server locally.

```sh
make server_run
```

</details>

#### Additional steps

1. Access the [local client app](http://milmovelocal:3000/) or the [deployed client app](https://my-milmove-pr-8364.mymove.sandbox.truss.coffee/)
2. Login as `finished@ppm.unsubmitted`.
3. Select the PPM shipment from the home screen and try editing fields through the edit flow. 
4. Ensure your edits are saved. Can do this either by checking on the review screen, or going back to the previous edit page after it takes you to the next page.
5. Submit the move.
    1. Alternatively, log out and then log in as `new@ppm.submitted`.
7. Ensure you can't see the edit button anymore.

### Frontend

### Backend

## Screenshots

Un-submitted move with finished PPM: 
image

Submitted move with finished PPM:
image


ronaktruss:
> Should there be validation on the BE to ensure shipments are only edited by customers when the move hasn't been submitted?

Similar to the above, another open question is if we should have validation on the BE to ensure that a deleted shipment cannot be edited?

pkg/testdatagen/make_ppm_shipment.go
```diff
@@ -64,46 +59,44 @@ func getDefaultValuesForRequiredFields(db *pop.Connection, shipment models.MTOSh
...
-		EstimatedIncentive:    &estimatedIncentive,
+	// We only want to set a SubmittedAt time if there is no status set in the assertions, or the one set matches our
+	// default of submitted.
+	if assertions.PPMShipment.Status == "" || assertions.PPMShipment.Status == models.PPMShipmentStatusSubmitted {
+		fullAssertions.PPMShipment.SubmittedAt = models.TimePointer(time.Now())
 	}
 
-	// Overwrite values with those from assertions
-	mergeModels(&ppmShipment, assertions.PPMShipment)
+	if assertions.PPMShipment.AdvanceRequested != nil && *assertions.PPMShipment.AdvanceRequested {
```
duncan-truss, line 88: Are you guarding against the case here where someone marks advance requested as true but doesn't provide an advance value in their assertion?

duncan-truss, line 88: Are you guarding against the case here where someone marks advance requested as true but doesn't provide an advance value in their assertion?

felipe-lee: It serves as that, but also as a lazy way of being able to say you want an advance but don't want to have to set a realistic value in the caller.

duncan-truss:
> Should there be validation on the BE to ensure shipments are only edited by customers when the move hasn't been submitted?

I think the Services Counselor will still be going in and potentially editing the shipment information after submission and creating their own PPM shipments. Although that will come from a different handler, so we could apply different validators based on the status/user doing the editing.

The customer will be coming back to edit shipment fields like the actual move date in closeout which is on the same model as the onboarding fields.  Not sure if it makes sense to introduce different endpoints/payload shapes to make it easier to distinguish updates coming at different times.

I'm not sure we want to get ahead of the Services Counselor work, seems like we'll know more once some of that lands.  I know orders has validators split out by user that is doing the updating if we're looking for an existing pattern.

felipe-lee:
> > Should there be validation on the BE to ensure shipments are only edited by customers when the move hasn't been submitted?
> 
> I think the Services Counselor will still be going in and potentially editing the shipment information after submission and creating their own PPM shipments. Although that will come from a different handler, so we could apply different validators based on the status/user doing the editing.
> 
> The customer will be coming back to edit shipment fields like the actual move date in closeout which is on the same model as the onboarding fields. Not sure if it makes sense to introduce different endpoints/payload shapes to make it easier to distinguish updates coming at different times.
> 
> I'm not sure we want to get ahead of the Services Counselor work, seems like we'll know more once some of that lands. I know orders has validators split out by user that is doing the updating if we're looking for an existing pattern.

Ah that's a good point. I'm fine with waiting on this one until we know more how the rest will look. I agree auth per user is likely what we'd need for this.

duncan-truss (approved):
Approved, really appreciate the great PR description write up too.

scarequotes (approved):
Looks great — worked as noted and as expected.

PR #8370, opened 2022-04-01
MB-11697-estimator-service-object

NamibiaTorres:
## Summary

This code sets up the estimated incentive for ppms. Team Roci has a ticket to add the estimator to the creator service, and so it was not added in this PR. The reason is that the pricer has not been added yet as part of the estimator and once it is, Team Roci may choose a different approach to how they will add the estimator to the creator service. 
https://dp3.atlassian.net/browse/MB-11865

##### Terminal 1

Run tests.

```sh
go test ./pkg/services/ppmshipment --testify.m TestEstimatedIncentive -v
```

</details>

### Backend


pkg/services/ppmshipment/ppm_shipment_creator.go
```diff
@@ -13,17 +13,19 @@ import (
 type ppmShipmentCreator struct {
 	mtoShipmentCreator services.MTOShipmentCreator
 	checks             []ppmShipmentValidator
+	ppmEstimator       services.PPMEstimator
```
felipe-lee, line 16: Do we need the estimator for the ppmShipmentCreator? At this point in time we won't really have the data we need to run the estimator right?

NamibiaTorres: Yes, that's right. We will need the estimator in the creator and the updater.

duncan-truss: I think we were anticipating possibly the services counselor filling in all the required data at creation time but probably not in the customer flow.

felipe-lee: Yeah, Namibia and I ended up talking about that yesterday. I think it's far enough out (Roci is working on the office routing phase 1 epic rn,  SC adding a PPM shipment is phase 2) that I think it would be better to hold off on putting this part in. I did add a comment to that story though about how we'll need to make updates to the creator if SCs will be adding the info in all at once: https://dp3.atlassian.net/browse/MB-11865?focusedCommentId=22015

pkg/services/ppmshipment/ppm_shipment_updater.go
```diff
@@ -49,6 +51,19 @@ func (f *ppmShipmentUpdater) updatePPMShipment(appCtx appcontext.AppContext, ppm
...
+	estimator, err := newPPMEstimator.EstimateIncentiveWithDefaultChecks(appCtx, *oldPPMShipment, updatedPPMShipment)
+	// HOW DO YOU CALL THE ESTIMATOR?
+	verrs, _ := appCtx.DB().ValidateAndUpdate(estimator)
+	// Add check for if the estimator encounters an err, then do not proceed with update
+	if err != nil {
+		return nil, err
+	}
+
+	if verrs != nil && verrs.HasAny() {
+		return apperror.NewNotImplementedError("Cannot update the Estimated Incentive."), nil
```
pearl-truss: Regarding the first server test error:
Screen Shot 2022-04-11 at 1 22 00 PM

The error is coming up because the function expects the return statement to be `(*models.PPMShipment, error)` and currently it's `(error, *models.PPMShipment)`
```suggestion
		return nil, apperror.NewNotImplementedError("Cannot update the Estimated Incentive.")
```

NamibiaTorres: Hi there! Thanks for the review. I fixed the change right around when you wrote this, but my PR should be up to date now with all of the most recent changes.

pkg/services/ppmshipment/ppm_estimator.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if newPPMShipment.ExpectedDepartureDate == oldPPMShipment.ExpectedDepartureDate && newPPMShipment.PickupPostalCode == oldPPMShipment.PickupPostalCode && newPPMShipment.DestinationPostalCode == oldPPMShipment.DestinationPostalCode && newPPMShipment.AdvanceRequested == oldPPMShipment.AdvanceRequested && newPPMShipment.EstimatedWeight == oldPPMShipment.EstimatedWeight {
+		return oldPPMShipment.EstimatedIncentive, nil
+	}
+	newPPMShipment.AdvanceRequested = nil
+	newPPMShipment.Advance = nil
+	// TODO: Call the pricer to calculate the incentive
+	newPPMShipment.EstimatedIncentive = models.Int32Pointer(int32(1000000))
+	// Saving Esimtated Incentive to DB
+	verrs, err := appCtx.DB().ValidateAndSave(newPPMShipment)
```
duncan-truss, line 54: I think because we are saving the PPM shipment to the database after this in the updater service that we shouldn't need to make an extra ValidateAndSave call here.  You've updated the `newPPMShipment.EstimatedIncentive` field with the value correctly so now in the PPM shipment updater you'll need to just call `appCtx.DB().ValidateAndUpdate(updatedPPMShipment)`.  Like we talked about because you passed that in as a pointer variable to this function that any updates you make to it will be available to the parent function that called it.

pkg/services/ppmshipment/ppm_shipment_updater.go
```diff
@@ -49,6 +51,19 @@ func (f *ppmShipmentUpdater) updatePPMShipment(appCtx appcontext.AppContext, ppm
 		return nil, err
 	}
 
+	newPPMEstimator := NewEstimatePPM()
+	estimator, err := newPPMEstimator.EstimateIncentiveWithDefaultChecks(appCtx, *oldPPMShipment, updatedPPMShipment)
+
+	verrs, _ := appCtx.DB().ValidateAndUpdate(estimator)
```
duncan-truss, line 57: You are returning the integer estimate from the call to the estimator not the model, so this should really be:

`verrs, _ := appCtx.DB().ValidateAndUpdate(updatedPPMShipment)`

pkg/services/ppmshipment/ppm_estimator.go
NamibiaTorres: Great! I was actually debating whether I needed that Validate and Save. I'll remove it.

pkg/services/ppmshipment/ppm_shipment_creator.go
duncan-truss: That's fair I didn't know how that SC create a shipment screen is supposed to work either.

pkg/services/ppmshipment/ppm_shipment_updater.go
NamibiaTorres: Thank you! I ended up removing the line because the updatedPPMShipment is validated and updated in the transaction.

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -0,0 +1,115 @@
...
+func (suite *PPMShipmentSuite) TestEstimatedIncentive() {
+	suite.Run("Estimated Incentive - Success", func() {
+		oldPPMShipment := testdatagen.MakeDefaultPPMShipment(suite.DB())
+		ppmEstimator := NewEstimatePPM()
+
+		estimatedWeight := unit.Pound(5000)
+		newPPM := models.PPMShipment{
+			ID:                    uuid.FromStringOrNil("575c25aa-b4eb-4024-9597-43483003c773"),
+			ShipmentID:            oldPPMShipment.ShipmentID,
+			Status:                "DRAFT",
```
duncan-truss, line 22: I did notice that oldPPMShipment has a status of SUBMITTED and we are setting it back to DRAFT here.  For creating oldPPMShipment I think you can call MakeMinimalPPMShipment instead and set stub to true because we're not testing any database functionality here.

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -0,0 +1,115 @@
...
+			DestinationPostalCode: "94040",
+			EstimatedWeight:       &estimatedWeight,
+			SitExpected:           models.BoolPointer(false),
+		}
+
+		ppmEstimate, err := ppmEstimator.EstimateIncentiveWithDefaultChecks(suite.AppContextForTest(), oldPPMShipment, &newPPM)
+		suite.NilOrNoVerrs(err)
+		suite.NotEqualValues(oldPPMShipment.PickupPostalCode, newPPM.PickupPostalCode)
+		suite.NotEqualValues(oldPPMShipment.EstimatedWeight, newPPM.EstimatedWeight)
+		suite.Equal(int32(1000000), *ppmEstimate)
```
duncan-truss, line 34: Because you're ignoring the return value of the estimate incentive in the ppm updater service, let's test that the newPPM.EstimatedIncentive is equal to 100000 here as well

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -0,0 +1,115 @@
...
+	})
+	suite.Run("Estimated Incentive - does not change when required fields are the same", func() {
+		oldPPMShipment := testdatagen.MakePPMShipment(suite.DB(), testdatagen.Assertions{
+			PPMShipment: models.PPMShipment{
+				EstimatedIncentive: models.Int32Pointer(int32(1000000)),
```
duncan-truss, line 39: Could you set this to a different value than 1000000 so we know it didn't just get overwritten with the same value and actually hit that conditional that checks that all of the fields are unchanged?

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -0,0 +1,115 @@
...
+			SitExpected:           oldPPMShipment.SitExpected,
+		}
+
+		ppmEstimate, err := ppmEstimator.EstimateIncentiveWithDefaultChecks(suite.AppContextForTest(), oldPPMShipment, &newPPM)
+		suite.NilOrNoVerrs(err)
+		suite.Equal(oldPPMShipment.PickupPostalCode, newPPM.PickupPostalCode)
+		suite.Equal(oldPPMShipment.EstimatedWeight, newPPM.EstimatedWeight)
+		suite.Equal(oldPPMShipment.DestinationPostalCode, newPPM.DestinationPostalCode)
+		suite.Equal(oldPPMShipment.ExpectedDepartureDate, newPPM.ExpectedDepartureDate)
+		suite.Equal(*oldPPMShipment.EstimatedIncentive, *ppmEstimate)
```
duncan-truss, line 61: Same comment here about checking to see that newPPM.EstimatedIncentive is the value you expect since that's what we're ultimately saving to the database.

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -0,0 +1,115 @@
...
+	})
+	suite.Run("Estimated Incentive - does not change when required fields are the same", func() {
```
duncan-truss, line 63: This test has the same title as the test above, did you mean for it to test something different?

Other tests I think you should include:
- Test for updating a shipment in a non draft status does not update the EstimatedIncentive field
- Test for updating a shipment without an estimated weight does not update the EstimatedIncentive field
- Test that updating a shipment clears a previously set Advance and AdvanceRequested values if a new estimated incentive was calculated

pkg/services/ppmshipment/ppm_estimator_test.go
duncan-truss: You could also add a value for advanceRequested and advance here to the old and new shipments and confirm those values aren't cleared.

pkg/services/ppmshipment/ppm_estimator.go
```diff
@@ -0,0 +1,61 @@
...
+	// Saving Esimtated Incentive to DB
+	//verrs, err := appCtx.DB().ValidateAndSave(newPPMShipment)
+	//if verrs != nil && verrs.HasAny() {
+	//	return nil, apperror.NewInvalidInputError(newPPMShipment.ID, err, verrs, "Invalid input found while creating the Estimated Incentive.")
+	//} else if err != nil {
+	//	return nil, apperror.NewQueryError("EstimatedIncentive", err, "")
+	//}
+
+	return newPPMShipment.EstimatedIncentive, nil
+}
```
felipe-lee, line 61: We're both setting the value on the `newPPMShipment` model and returning it. I think we should probably do one or the other, and do it both times. Like right now, if the value hasn't changed, we aren't setting it on the `newPPMShipment`, we're just returning the `oldPPMShipment.EstimatedIncentive` (line 46). 

I personally think it's cleaner to return the value we expect, and let the caller (the update service object) set that value on the model, but I'm fine if we chose to set the value here. If we do set the value here, then I don't think we need to return the value though.

pkg/services/ppmshipment/ppm_estimator_test.go
felipe-lee: **todo**: Can you update any statuses to use the constants we have in https://github.com/transcom/mymove/blob/7a99496cd2590770b2edc2b0b187f4dc44c0749a/pkg/models/ppm_shipment.go#L17-L30

pkg/services/ppmshipment/ppm_estimator.go
```diff
@@ -0,0 +1,58 @@
...
+
+	if newPPMShipment.Status != models.PPMShipmentStatusDraft {
+		return nil, err
+	}
+	// Check that all the required fields we need are present.
+	err = validatePPMShipment(appCtx, *newPPMShipment, &oldPPMShipment, &oldPPMShipment.Shipment, checks...)
+	if err != nil {
+		switch err.(type) {
+		case apperror.InvalidInputError:
+			return nil, nil
```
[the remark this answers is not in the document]
NamibiaTorres: Yes! I updated with comments.

.pre-commit-config.yaml
```diff
@@ -8,7 +8,7 @@ repos:
         types: [go]
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.1.0
+    rev: v4.2.0
```
duncan-truss, line 11: This is what triggered the Truss-IS3 code reviewer, did we need to keep this for your autoupgrade to work?

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -0,0 +1,144 @@
...
+		newPPM := models.PPMShipment{
+			ID:                    uuid.FromStringOrNil("575c25aa-b4eb-4024-9597-43483003c773"),
+			ShipmentID:            oldPPMShipment.ShipmentID,
+			Status:                models.PPMShipmentStatusDraft,
+			ExpectedDepartureDate: time.Time(time.Date(2022, 12, 11, 0, 0, 0, 0, time.UTC)),
+			PickupPostalCode:      "20636",
+			DestinationPostalCode: "94040",
+			EstimatedWeight:       &estimatedWeight,
+			SitExpected:           models.BoolPointer(false),
+		}
```
duncan-truss, line 34: As I'm reading this is hard for me to know what values are changing from the old shipment without diving into what testdatagen.MakeMinimalPPMShipment() is doing.  I might propose calling `mergePPMShipment()` directly with only the fields being updated to make that obvious.  The other alternative is calling something like `deepcopy.Copy()` and then updating the fields with values that you want to change.

.pre-commit-config.yaml
NamibiaTorres: I'm not sure. I can revert the line, however do most folks have version 4.2?

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -0,0 +1,144 @@
...
+		suite.NotEqualValues(oldPPMShipment.PickupPostalCode, newPPM.PickupPostalCode)
+		suite.NotEqualValues(oldPPMShipment.EstimatedWeight, newPPM.EstimatedWeight)
+		suite.Equal(oldPPMShipment.Advance, newPPM.Advance)
+		suite.Equal(oldPPMShipment.AdvanceRequested, newPPM.AdvanceRequested)
+		suite.Equal(int32(1000000), *ppmEstimate)
+		suite.Equal(int32(1000000), *newPPM.EstimatedIncentive)
+	})
+
+	suite.Run("Estimated Incentive - Success - clears advance and advance requested values", func() {
+		oldPPMShipment := testdatagen.MakeDefaultPPMShipment(suite.DB())
```
duncan-truss, line 47: I don't think MakeDefaultPPMShipment (at least on master) sets advance and advanceRequested fields, you may need to call MakePPMShipment and pass in those values in the assertions.

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -0,0 +1,144 @@
...
+	suite.Run("Estimated Incentive - does not change when required fields are the same", func() {
+		oldPPMShipment := testdatagen.MakePPMShipment(suite.DB(), testdatagen.Assertions{
+			PPMShipment: models.PPMShipment{
+				EstimatedIncentive: models.Int32Pointer(int32(500000)),
+			},
+		})
+		ppmEstimator := NewEstimatePPM()
+
+		newPPM := models.PPMShipment{
+			ID:                    uuid.FromStringOrNil("575c25aa-b4eb-4024-9597-43483003c773"),
```
duncan-truss, line 78: This test looks good, you should be able to use the oldPPMShipment.ID field as well unless you're stubbing it then you might need to provide one.

pkg/services/ppmshipment/ppm_estimator_test.go
NamibiaTorres: Would it be sufficient to add a comment. All the values were changing.

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -0,0 +1,144 @@
...
+			ExpectedDepartureDate: oldPPMShipment.ExpectedDepartureDate,
+			PickupPostalCode:      oldPPMShipment.PickupPostalCode,
+			DestinationPostalCode: "94040",
+			EstimatedWeight:       oldPPMShipment.EstimatedWeight,
+			SitExpected:           oldPPMShipment.SitExpected,
+		}
+
+		ppmEstimate, err := ppmEstimator.EstimateIncentiveWithDefaultChecks(suite.AppContextForTest(), oldPPMShipment, &newPPM)
+		suite.NilOrNoVerrs(err)
+		suite.Nil(ppmEstimate)
```
duncan-truss, line 118: Could you add a check here that newPPM.EstimatedIncentive is still 500000 like in the oldPPMShipment?

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -0,0 +1,144 @@
...
+	})
+
+	suite.Run("Estimated Incentive - Failure - is not created when Estimated Weight is missing", func() {
+		oldPPMShipment := testdatagen.MakePPMShipment(suite.DB(), testdatagen.Assertions{
+			PPMShipment: models.PPMShipment{
+				EstimatedIncentive: models.Int32Pointer(int32(1000000)),
```
duncan-truss, line 124: I think you are testing the scenario where a customer hasn't yet filled in the estimated weight in the ppm booking flow.  In that case EstimatedIncentive would not be calculated yet so we could use MakeMinimalPPMShipment here.

pkg/services/ppmshipment/ppm_estimator_test.go
duncan-truss: Yep at least a comment about intent would help.

pkg/services/ppmshipment/ppm_estimator_test.go
NamibiaTorres: I'm confused, `MakeDefaultPPMShipment` returns `MakePPMShipment`, and it looks like that does pass in advance and advance requested.

duncan-truss: You're right! Sorry I don't know what I was looking at 😅

duncan-truss: mergePPMShipment is getting called so advance and advanceRequested should have values on newPPM though prior to calling the estimator.

pkg/services/ppmshipment/ppm_estimator.go
NamibiaTorres: I have a question about this. Would it be alright if we set up a time to zoom later @felipe-lee ?

pkg/services/ppmshipment/ppm_estimator.go
```diff
@@ -0,0 +1,60 @@
...
+		default:
+			return nil, err
+		}
+	}
+
+	if newPPMShipment.ExpectedDepartureDate == oldPPMShipment.ExpectedDepartureDate && newPPMShipment.PickupPostalCode == oldPPMShipment.PickupPostalCode && newPPMShipment.DestinationPostalCode == oldPPMShipment.DestinationPostalCode && newPPMShipment.EstimatedWeight == oldPPMShipment.EstimatedWeight {
```
duncan-truss, line 49: This is a sneaky one that Cypress surfaced where the advance was being cleared even though the values didn't change.

Turns out the expectedDepartureDate will not be equal when comparing using `===` because the timezone location format is slightly different when comparing a date coming back from the database and one coming from the api.

`2020-03-15 00:00:00 +0000 +0000`
`2020-03-15 00:00:00 +0000 UTC`

The fix would be to use `newPPMShipment.ExpectedDepartureDate.equal(oldPPMShipment.ExpectedDepartureDate)` instead.

duncan-truss: And then we'll need to compare the estimatedWeight values instead of the pointer addresses `*newPPMShipment.EstimatedWeight == *oldPPMShipment.EstimatedWeight`

felipe-lee (commented):
Thanks for addressing my comments! I have a few more, and this:

**issue**: Now that the estimator is setting the hard-coded incentive, can you remove the hard-coded amount from the `payload_to_model.go` file: 

https://github.com/transcom/mymove/blob/958b67467ee6433f1c7d99d0ca6581f904178cb2/pkg/handlers/internalapi/internal/payloads/payload_to_model.go#L136-L137

With the new service object, the `payload_to_model.go` shouldn't even set the estimated incentive on the model.

NamibiaTorres:
> https://github.com/transcom/mymove/pull/8313/files#r833618231

I see! I'll play around with this value to see if that fixes the tests. Thanks for pointing this one out! I suspected it had to do with passing in a wrong value.

pkg/services/ppmshipment/rules_test.go
```diff
@@ -476,4 +476,44 @@ func (suite *PPMShipmentSuite) TestValidationRules() {
...
+				EstimatedWeight: nil,
+			}
+
+			newPPMShipment := models.PPMShipment{
+				ShipmentID:      shipmentID,
+				EstimatedWeight: &estimatedWeight,
+			}
+
+			err := checkEstimatedWeight().Validate(suite.AppContextForTest(), newPPMShipment, &oldPPMShipment, nil)
+			suite.Error(err)
```
duncan-truss, line 516: Your validation looks right but this test is silently succeeding because verrs is not nil but is empty.  The newPPMShipment should have EstimatedWeight set to nil.

In this case you could change your test to make sure the error message matches:
`suite.Equal("cannot be empty", err.Error())`

duncan-truss (approved):
🎉 I think this is in a good place now.  Last comment was just about the failure test for estimatedWeight test data needing to be changed https://github.com/transcom/mymove/pull/8370/files#r856407947

NamibiaTorres:
Merge estimated weight ticket now that all code has been approved.

PR #8372, opened 2022-04-01
MB-11673 Add Delete Shipment

pearl-truss:
## Summary

This PR adds delete shipment functionality to the internal api. 

### Open Questions
- Should I add a check to ensure that the deleter is the logged in customer? I wasn't sure if this was needed or not
- This PR makes the assumption that we want to delete the shipment at the `MTOShipment` level and not at the subtable `PPM_Shipment` level 

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the
2. Login as a
3.

### Backend


pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -326,3 +326,34 @@ func (h ListMTOShipmentsHandler) Handle(params mtoshipmentops.ListMTOShipmentsPa
...
+// Handle soft deletes a shipment
+func (h DeleteShipmentHandler) Handle(params mtoshipmentops.DeleteShipmentParams) middleware.Responder {
+	return h.AuditableAppContextFromRequestWithErrors(params.HTTPRequest,
+		func(appCtx appcontext.AppContext) (middleware.Responder, error) {
+			shipmentID := uuid.FromStringOrNil(params.MtoShipmentID.String())
+			_, err := h.DeleteShipment(appCtx, shipmentID)
+			if err != nil {
+				appCtx.Logger().Error("internalapi.DeleteShipmentHandler", zap.Error(err))
+
+				switch err.(type) {
```
felipe-lee, line 349: I know we should likely not hit the `apperror.ForbiddenError` error that the `DeleteShipment` service can raise, but it'd probably be good to catch it in case we do hit it, or the logic it uses changes.

felipe-lee (commented):
LGTM so far!

> Should I add a check to ensure that the deleter is the logged in customer? I wasn't sure if this was needed or not

I feel like we probably should start adding in auth checks. It keeps coming up but we haven't done it, so maybe it's time to start implementing that logic in our handlers since now if we have time.

> This PR makes the assumption that we want to delete the shipment at the MTOShipment level and not at the subtable PPM_Shipment level

We went back and forth on this a bit, but I was looking at the [soft delete docs](https://transcom.github.io/mymove-docs/docs/backend/guides/how-to/soft-delete/) we have and they seem to indicate that we should be adding the soft delete to child models as well. So maybe we should add that in.

pkg/handlers/internalapi/mto_shipment.go
pearl-truss: this is resolved now

pearl-truss:
> LGTM so far!
> 
> > Should I add a check to ensure that the deleter is the logged in customer? I wasn't sure if this was needed or not
> 
> I feel like we probably should start adding in auth checks. It keeps coming up but we haven't done it, so maybe it's time to start implementing that logic in our handlers since now if we have time.
> 
> > This PR makes the assumption that we want to delete the shipment at the MTOShipment level and not at the subtable PPM_Shipment level
> 
> We went back and forth on this a bit, but I was looking at the [soft delete docs](https://transcom.github.io/mymove-docs/docs/backend/guides/how-to/soft-delete/) we have and they seem to indicate that we should be adding the soft delete to child models as well. So maybe we should add that in.

Thanks for the review! My fixes for these points are in progress. I noticed that the current `DeleteShipment` service isn't using `utilities.SoftDestroy` so unless. there's objections I'm planning on adding that in for both `mto_shipment` and `ppm_shipment`

swagger-def/internal.yaml
```diff
@@ -2445,6 +2445,10 @@ definitions:
         format: date-time
         type: string
         readOnly: true
+      deletedAt:
+        format: date-time
+        type: string
+        readOnly: true
```
felipe-lee, line 2451: I wonder if we do need to add this or not to the endpoint. Like I think at least for the customer side, we don't want to return deleted shipments at all, so I think leaving this out of the response would be fine, what do you think?

felipe-lee:
> > LGTM so far!
> > > Should I add a check to ensure that the deleter is the logged in customer? I wasn't sure if this was needed or not
> > 
> > 
> > I feel like we probably should start adding in auth checks. It keeps coming up but we haven't done it, so maybe it's time to start implementing that logic in our handlers since now if we have time.
> > > This PR makes the assumption that we want to delete the shipment at the MTOShipment level and not at the subtable PPM_Shipment level
> > 
> > 
> > We went back and forth on this a bit, but I was looking at the [soft delete docs](https://transcom.github.io/mymove-docs/docs/backend/guides/how-to/soft-delete/) we have and they seem to indicate that we should be adding the soft delete to child models as well. So maybe we should add that in.
> 
> Thanks for the review! My fixes for these points are in progress. I noticed that the current `DeleteShipment` service isn't using `utilities.SoftDestroy` so unless. there's objections I'm planning on adding that in for both `mto_shipment` and `ppm_shipment`



> > LGTM so far!
> > > Should I add a check to ensure that the deleter is the logged in customer? I wasn't sure if this was needed or not
> > 
> > 
> > I feel like we probably should start adding in auth checks. It keeps coming up but we haven't done it, so maybe it's time to start implementing that logic in our handlers since now if we have time.
> > > This PR makes the assumption that we want to delete the shipment at the MTOShipment level and not at the subtable PPM_Shipment level
> > 
> > 
> > We went back and forth on this a bit, but I was looking at the [soft delete docs](https://transcom.github.io/mymove-docs/docs/backend/guides/how-to/soft-delete/) we have and they seem to indicate that we should be adding the soft delete to child models as well. So maybe we should add that in.
> 
> Thanks for the review! My fixes for these points are in progress. I noticed that the current `DeleteShipment` service isn't using `utilities.SoftDestroy` so unless. there's objections I'm planning on adding that in for both `mto_shipment` and `ppm_shipment`

Yeah that makes sense to update the service to use the utility.

swagger-def/internal.yaml
gidjin: I had a similar thought to Felipe.

gidjin (commented):
Changes look good, I'm ready to approve the DB changes as a db-reviewer but I'll wait pending the discussion about deleted-at in the customer api yaml.

swagger-def/internal.yaml
pearl-truss: Makes sense, this is resolved now

pkg/services/mto_shipment/shipment_deleter.go
```diff
@@ -33,9 +33,24 @@ func (f *shipmentDeleter) DeleteShipment(appCtx appcontext.AppContext, shipmentI
 		return uuid.Nil, err
 	}
 
-	now := time.Now()
-	shipment.DeletedAt = &now
-	err = appCtx.DB().Save(shipment)
+	if shipment.PPMShipment != nil {
+		err := models.DeletePPMShipment(appCtx.DB(), shipment.PPMShipment)
```
duncan-truss, line 37: This seems to be a separate transaction, could we move it into the same transaction as the one below?

pkg/models/ppm_shipment.go
```diff
@@ -82,3 +85,10 @@ func FetchPPMShipmentFromMTOShipmentID(db *pop.Connection, mtoShipmentID uuid.UU
 	}
 	return &ppmShipment, nil
 }
+
+// DeletePPMShipment will soft delete the PPMShipment
+func DeletePPMShipment(db *pop.Connection, ppmShipment *PPMShipment) error {
```
duncan-truss, line 90: This would make more sense to me as a function in the shipment deleter service, was there a reason you wanted this on the model?

pkg/services/mto_shipment/shipment_deleter.go
pearl-truss: I didn't do this to avoid nesting transactions

pkg/models/ppm_shipment.go
pearl-truss: Okay. I'm seeing that `utilities.softDestroy` also destroys a model's child associations. So I'm going to remove this and rely on the top level delete like I had previously

pearl-truss:
Regarding the open question: Should I add a check to ensure that the deleter is the logged in customer? I wasn't sure if this was needed or not?

I found that:
- There's  `UserAuthMiddleware` function in https://github.com/transcom/mymove/blob/master/pkg/auth/authentication/auth.go#L58 
- This function currently only works for the office and admin apps. It has this comment regarding the customer app
- > // DO NOT CHECK MILMOVE SESSION BECAUSE NEW SERVICE MEMBERS WON'T HAVE AN ID RIGHT AWAY
			// This must be the right type of user for the application
-The `UserAuthMiddleware` function gets used in `serve.go` for each API (including the internalAPI even though it doesn't check milmove sessions...?)
- The auth.Session is passed into `AppContext` which is available to the handlers

What I'm taking away is the current state of the internal app is that auth isn't happening unless it's being handled ad hoc in the handlers. I can add an auth check for the delete ppm handler cc: @transcom/truss-a-team if there are other ideas or thoughts?

pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -326,3 +326,48 @@ func (h ListMTOShipmentsHandler) Handle(params mtoshipmentops.ListMTOShipmentsPa
...
+			shipmentID := uuid.FromStringOrNil(params.MtoShipmentID.String())
+
+			sm, err := models.GetCustomerFromShipment(appCtx.DB(), shipmentID)
+			if err != nil {
+				return mtoshipmentops.NewDeleteShipmentNotFound(), err
+			}
+
+			if appCtx.Session() != nil {
+				if appCtx.Session().ServiceMemberID != sm.ID {
+					return mtoshipmentops.NewDeleteShipmentInternalServerError(), err
```
duncan-truss, line 353: Should this be a Forbidden error instead of an InternalServerError?

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -1155,3 +1156,158 @@ func (suite *HandlerSuite) TestListMTOShipmentsHandler() {
...
+		shipment := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+			Move: move,
+			MTOShipment: models.MTOShipment{
+				ShipmentType: models.MTOShipmentTypePPM,
+			},
+		})
+
+		deleter := &mocks.ShipmentDeleter{}
+
+		deleter.On("DeleteShipment", mock.AnythingOfType("*appcontext.appContext"), shipment.ID).Return(uuid.Nil, apperror.ForbiddenError{})
```
duncan-truss, line 1259: Similar to the comment above, I think the authorization error would be when GetCustomerFromShipment returns a different ServiceMember.ID. because DeleteShipment never returns a 403 error.  It doesn't look like we generate mocks for our models (which is one reason why having queries in them complicates things).  You may need to setup the real data with factories to test that one.

migrations/app/schema/20220412195727_add_deletedAt_to_storage_facilities_and_mto_agents.up.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE mto_agents
+ADD deleted_at timestamp with time zone;
```
duncan-truss, line 2: Sorry to be a stickler, do you mind throwing comments on this columns too?

duncan-truss (approved):

pkg/handlers/internalapi/mto_shipment.go
```diff
@@ -326,3 +326,52 @@ func (h ListMTOShipmentsHandler) Handle(params mtoshipmentops.ListMTOShipmentsPa
...
+	return h.AuditableAppContextFromRequestWithErrors(params.HTTPRequest,
+		func(appCtx appcontext.AppContext) (middleware.Responder, error) {
+			shipmentID := uuid.FromStringOrNil(params.MtoShipmentID.String())
+
+			sm, err := models.GetCustomerFromShipment(appCtx.DB(), shipmentID)
+			if err != nil {
+				return mtoshipmentops.NewDeleteShipmentNotFound(), err
+			}
+
+			if appCtx.Session() != nil {
```
felipe-lee, line 351: Is it possible to get to this endpoint without a session? I know you showed the comment about SMs not having IDs right away, but I think by the point we get here, they probably should have an ID and session.

pearl-truss: Before I added this check I was getting a nil point dereference error when I ran the server test

pkg/handlers/internalapi/mto_shipment_test.go
pearl-truss: Good catch, this is resolved now

pkg/handlers/internalapi/mto_shipment.go
duncan-truss: I think to Felipe's point you were getting nil because you may need to wrap your test requests in AuthenticateUser to associate the request with a user session https://github.com/transcom/mymove/blob/19cac0997a7a4f21b668813fe2e606e637ae95fe/pkg/handlers/apitests.go#L138

PR #8375, opened 2022-04-04
MB-11601 delete shipment

ronaktruss:
## Summary

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Frontend

## Screenshots
Screen Shot 2022-04-08 at 1 59 40 PM
Screen Shot 2022-04-08 at 2 00 21 PM
Screen Shot 2022-04-08 at 2 01 00 PM
Screen Shot 2022-04-08 at 2 01 40 PM
Screen Shot 2022-04-08 at 2 01 53 PM
Screen Shot 2022-04-08 at 2 02 08 PM


src/components/ShipmentList/index.jsx
```diff
@@ -60,16 +56,29 @@ export const ShipmentListItem = ({
           <span className={styles.warningText}>{isOverweight ? 'Over weight' : 'Missing weight'}</span>
         </div>
       )}
-      {canEdit ? <FontAwesomeIcon icon="pen" className={styles.edit} /> : <div className={styles.noEdit} />}
```
felipe-lee, line 63: I'm glad we're getting rid of the icon-only button in favor of text 🎉

src/components/ShipmentList/index.jsx
```diff
@@ -60,16 +56,29 @@ export const ShipmentListItem = ({
...
-    </button>
+      {canEditOrDelete ? (
+        <span className={styles['shipment-btns']}>
+          <Button className={styles['edit-btn']} onClick={onDeleteClick} type="button">
+            Delete
+          </Button>
+          <div>|</div>
+          <Button className={styles['edit-btn']} onClick={onShipmentClick} type="button">
+            Edit
+          </Button>
```
felipe-lee, line 67: I'm not 100% sure if we should keep using buttons or not for things that are in effect, just links. It's come up in the past but I don't remember the decision in the end. I've started a [thread in #prac-frontend](https://ustcdp3.slack.com/archives/CTQQJD3G8/p1649202414214919) asking if others remember what was decided.

src/components/Customer/Home/Step/Step.module.scss
```diff
@@ -22,6 +22,10 @@
   }
 }
 
+.step-container:last-of-type {
+  margin-bottom: 0;
+}
+
```
ronaktruss, line 28: Remove additional whitespace at the bottom of the home page
image

src/pages/MyMove/Review/Review.jsx
```diff
@@ -65,9 +65,10 @@ Review.propTypes = {
 };
 
 const mapStateToProps = (state, ownProps) => {
+  const mtoShipments = state?.entities?.mtoShipments;
   return {
     ...ownProps,
-    canMoveNext: true, // TODO: prevent incomplete PPM submission?
+    canMoveNext: !!Object.keys(mtoShipments).length, // TODO: prevent incomplete PPM submission?
```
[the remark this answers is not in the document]
duncan-truss: Heads up I removed this property in my PR that's now in master.  I did add a selector to return the shipments similar to what you're doing above.  Probably just move this logic to inside the component in conjunction with the check for an incomplete PPM.

src/components/ShipmentList/ShipmentList.test.jsx
```diff
@@ -42,38 +45,51 @@ describe('ShipmentList component', () => {
...
+    editBtn = queryByRole(screen.getAllByTestId('shipment-list-item-container')[2], 'button', { name: 'Edit' });
 
     await waitFor(() => {
       checkShipmentClick('ID-3', 1, SHIPMENT_OPTIONS.NTS);
     });
 
-    userEvent.click(screen.getByRole('button', { name: /^nts-release /i }));
+    editBtn = queryByRole(screen.getAllByTestId('shipment-list-item-container')[3], 'button', { name: 'Edit' });
 
     await waitFor(() => {
```
ronaktruss, line 77: @felipe-lee Is there a reason that we are using await here as opposed just triggering the click and using expects similar to what is done for the deletes below?

src/containers/FlashMessage/FlashMessage.jsx
```diff
@@ -7,10 +7,13 @@ import { clearFlashMessage as clearFlashMessageAction } from 'store/flash/action
...
-    // Clear this flash message on unmount (this will happen on navigation or if flash state changes)
-    clearFlashMessage(flash?.key);
-  });
+  useEffect(
+    () => () => {
+      // Clear this flash message on unmount (this will happen on navigation or if flash state changes)
+      clearFlashMessage(flash?.key);
+    },
+    [clearFlashMessage, flash?.key],
+  );
```
ronaktruss, line 16: This was changed because the previous implementation made it so that if you modified the message multiple times on the same page it would be cleared on alternating notifications. (eg. If you delete two records on the first you would see a message, but on the second the message would disappear. You would then alternate seeing and not seeing it on subsequent triggers.)

Does anyone have context on the reason for this original implementation? If the comment is correct I think this servers to better mimic that description.

ronaktruss:
@scarequotes There are some existing issues with the modal that causes some potentially jarring scrolling to the top and back to your original position after the modal closes.

src/components/ConfirmationModals/DestructiveShipmentConfirmationModal.jsx
```diff
@@ -0,0 +1,63 @@
...
+import { Button } from '@trussworks/react-uswds';
+
+import Modal, { ModalTitle, ModalClose, ModalActions, connectModal } from 'components/Modal/Modal';
+import { Overlay, ModalContainer } from 'components/MigratedModal/MigratedModal';
+
+export const DestructiveShipmentConfirmationModal = ({
+  onClose,
+  onSubmit,
+  shipmentID,
+  title = 'Are you sure?',
```
jenniferlynparsons, line 12: issue: The default value strings go against our typical pattern on this project.
suggestion: Use `defaultProps` (part of `propTypes`) to hold default values. You can see an example of how `defaultProps` are used with string placeholders on [FileUpload.jsx])https://github.com/transcom/mymove/blob/84f4f1004a9f7031d3e1cce0682e741eab5a9230/src/components/FileUpload/FileUpload.jsx#L92)

ronaktruss: Happy to make that change. Is there a particular reason that is the pattern? 

I saw some indication that defaultProps for functional components may be deprecated in the "future"(https://github.com/reactjs/rfcs/pull/107 and https://github.com/facebook/react/pull/16210/files)

YanZ777 (changes requested):

src/components/Customer/Review/ShipmentCard/PPMShipmentCard/PPMShipmentCard.test.jsx
```diff
@@ -116,9 +120,24 @@ describe('PPMShipmentCard component', () => {
...
+    render(<PPMShipmentCard {...completeProps} showEditAndDeleteBtn={false} />);
 
     expect(screen.queryByRole('button', { name: 'Edit' })).not.toBeInTheDocument();
+    expect(screen.queryByRole('button', { name: 'Delete' })).not.toBeInTheDocument();
+  });
+
+  it('calls onEditClick when edit button is pressed', () => {
+    render(<PPMShipmentCard {...completeProps} />);
+    const editBtn = screen.queryByRole('button', { name: 'Edit' });
+    fireEvent.click(editBtn);
```
YanZ777, line 133: **suggestion (non blocking):** Use `userEvent` rather than `fireEvent`.

src/components/ShipmentList/ShipmentList.test.jsx
```diff
@@ -42,38 +45,51 @@ describe('ShipmentList component', () => {
...
       checkShipmentClick('ID-4', 1, SHIPMENT_OPTIONS.NTSR);
     });
   });
+
+  it('calls onDeleteClick when delete is clicked', () => {
+    render(<ShipmentList {...defaultProps} />);
+    for (let i = 0; i < defaultProps.shipments.length; i += 1) {
```
YanZ777, line 84: **suggestion (non blocking):** It looks like you might want to use a table driven test here instead.

src/components/ShipmentList/ShipmentList.jsx
```diff
@@ -60,16 +56,29 @@ export const ShipmentListItem = ({
...
   );
 };
 
 ShipmentListItem.propTypes = {
   shipment: shape({ id: string.isRequired, shipmentType: string.isRequired }).isRequired,
   onShipmentClick: func,
+  onDeleteClick: func,
   shipmentNumber: number.isRequired,
-  canEdit: bool.isRequired,
+  canEditOrDelete: bool.isRequired,
```
YanZ777, line 81: **thought:** I'm not sure how I feel about having both edit and delete being controlled by the same prop.
**question:** Will edit and delete always go hand in hand or will there ever be a case where a person can edit, but not delete, or vice versa?

src/components/ShipmentList/ShipmentList.module.scss
```diff
@@ -33,12 +30,25 @@
...
-  color: $primary;
+.shipment-btns {
+  display: flex;
+  justify-content: center;
+  align-items: center;
 }
 
-.noEdit {
-  width: 20px;
+:global(.usa-button).edit-btn {
```
YanZ777, line 39: **suggestion:** I think you should be able to pass `secondary` to the button that you are trying to give this styling, since this looks like you're trying to get secondary styling. 
e.g. `<Button secondary type="button" onClick={closeModal}>`

YanZ777 (commented):
A few more comments I have after going through this PR. I don't think I did a very deep review, so I'd prefer others to weigh in as well.

src/components/ShipmentList/ShipmentList.module.scss
YanZ777: **thought:** Scratch this comment. I see what the visual looks like. Secondary won't help here.

src/components/ShipmentList/ShipmentList.jsx
ronaktruss: Our current expectations are that both edit and delete should be accessible as long as the move is not submitted. I didn't want to preemptively optimize for alternative scenarios that might not happen, as this is a fairly small change we can make when it is needed.

Are there any existing scenarios where we would want to separate out the availability of this functionality.

src/components/ConfirmationModals/DestructiveShipmentConfirmationModal.jsx
```diff
@@ -0,0 +1,63 @@
...
+  content = 'You’ll lose all the information in this shipment. If you want it back later, you’ll have to request a new shipment.',
+  submitText = 'Delete shipment',
+  backText = 'Keep shipment',
+}) => (
+  <div>
+    <Overlay />
+    <ModalContainer>
```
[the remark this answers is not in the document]
ronaktruss: Thanks for pointing out the connectedModal. 

Given the current implementation it looks like the usage of a connected Modal will make it so that the modal is top aligned but supports scenarios where the modal text is large/requires scrolling. If we manually use `Overlay` and `ModalContainer` this will make the modal vertically centered but not appropriately handle a large amount of modal text. I have started a [thread](https://ustcdp3.slack.com/archives/CNTDW9Q1Z/p1649691824364099) to see if there is an expected behavior for modals. 

ConnectedModal behavior:
Screen Shot 2022-04-11 at 9 54 12 AM

https://user-images.githubusercontent.com/84742904/162788742-a29ffe3f-35ae-4dbc-90d3-acc7e00637f4.mov

Manual usage of Overlay and ModalContainer:
Screen Shot 2022-04-11 at 10 03 59 AM

https://user-images.githubusercontent.com/84742904/162788599-e6aabae7-2253-4828-8409-9db4ce2584c2.mov

src/containers/FlashMessage/FlashMessage.jsx
YanZ777: **thought:** I don't have any context on this.

src/components/ShipmentList/ShipmentList.jsx
YanZ777: **thought:** I think if the current expectation is that both edit and delete and jointly accessible, then it's fine to leave as it is.

cypress/integration/office/servicescounseling/servicesCounselingNTSR.js
```diff
@@ -40,7 +40,7 @@ describe('Services counselor user', () => {
 
     cy.get('[data-testid="modal"]').should('be.visible');
 
-    cy.get('[data-testid="modal"] button').contains('Delete shipment').click();
+    cy.get('[data-testid="modal"] button').contains('Delete shipment').click({ force: true });
```
pearl-truss, line 43: Thanks for adding this!

src/components/ConfirmationModals/DestructiveShipmentConfirmationModal.stories.jsx
```diff
@@ -0,0 +1,21 @@
...
+  title: 'Components/DestructiveShipmentConfirmationModal',
+  component: DestructiveShipmentConfirmationModal,
+};
+
+export const Basic = () => <DestructiveShipmentConfirmationModal onClose={() => {}} onSubmit={() => {}} />;
+
+export const withOverrides = () => (
+  <DestructiveShipmentConfirmationModal
+    onClose={() => {}}
+    onSubmit={() => {}}
```
duncan-truss, line 15: Try to define these stories using the newer [args pattern](https://storybook.js.org/docs/react/writing-stories/args) so that these props become editable in the Storybook controls panel and you can also use actions so those events show up in the [actions tab](https://github.com/storybookjs/storybook/blob/master/addons/actions/ADVANCED.md#manually-specified-actions) too.

src/components/Customer/Review/ShipmentCard/ShipmentCard.module.scss
```diff
@@ -37,13 +37,17 @@
...
+    .btnContainer {
       display: inline-flex;
       flex: 0;
       align-items: center;
       min-width: auto;
       @include u-padding-x(0);
       @include u-font('body', 'xs');
+
+      button {
+        padding: 12px;
```
duncan-truss, line 49: I think it's good to avoid using pixels and use USWDS tokens whenever possible.  You should be able to rewrite this as `@include u-padding(1.5)` https://designsystem.digital.gov/design-tokens/spacing-units/

ronaktruss:
> I've been testing and things look very good! One minor issue I see is when you delete all of the shipments on the Review page, the next button is still enabled for you to submit the move. I think if you just combine a check for mtoShipments length with the complete PPM conditional in Review.jsx you can fix that disabling.

@duncan-truss Yeah I caught this as well when I was working on the cypress tests. I forgot to add this back in when I merged in your code that added the logic for the incomplete PPMs. See [commit](https://github.com/transcom/mymove/pull/8375/commits/dddeeeba110b2c180d31d669d87f20162af9f5ce)

src/components/Customer/Review/ShipmentCard/PPMShipmentCard/PPMShipmentCard.test.jsx
```diff
@@ -116,9 +121,24 @@ describe('PPMShipmentCard component', () => {
...
+  it('omits the edit button when showEditAndDeleteBtn prop is false', () => {
+    render(<PPMShipmentCard {...completeProps} showEditAndDeleteBtn={false} />);
 
     expect(screen.queryByRole('button', { name: 'Edit' })).not.toBeInTheDocument();
+    expect(screen.queryByRole('button', { name: 'Delete' })).not.toBeInTheDocument();
+  });
+
+  it('calls onEditClick when edit button is pressed', () => {
+    render(<PPMShipmentCard {...completeProps} />);
+    const editBtn = screen.queryByRole('button', { name: 'Edit' });
```
felipe-lee, line 133: The `queryBy*` queries should be used for when you expect something to not exist. Otherwise, it should probably be `getBy*` or `findBy*`

https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence

src/components/ShipmentList/ShipmentList.test.jsx
```diff
@@ -20,18 +20,21 @@ describe('ShipmentList component', () => {
...
 
-    screen.getByRole('button', { name: /^ppm #id-1/i });
-    screen.getByRole('button', { name: /^hhg #id-2/i });
-    screen.getByRole('button', { name: /^nts #id-3/i });
-    screen.getByRole('button', { name: /^nts-release #id-4/i });
+    expect(screen.getAllByTestId('shipment-list-item-container').length).toBe(4);
+    expect(screen.getAllByTestId('shipment-list-item-container')[0]).toHaveTextContent(/^ppm #id-1/i);
+    expect(screen.getAllByTestId('shipment-list-item-container')[1]).toHaveTextContent(/^hhg #id-2/i);
+    expect(screen.getAllByTestId('shipment-list-item-container')[2]).toHaveTextContent(/^nts #id-3/i);
+    expect(screen.getAllByTestId('shipment-list-item-container')[3]).toHaveTextContent(/^nts-release #id-4/i);
```
felipe-lee, line 37: These shouldn't use the testID. I can see needing if for the count since there's not a great way to grab that otherwise, but for each shipment, we can use the `getByText` query. It's better to avoid testIDs with RTL if possible:
* https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-querying-by-text
* https://testing-library.com/docs/queries/about/#priority

src/components/ShipmentList/ShipmentList.test.jsx
```diff
@@ -42,38 +45,51 @@ describe('ShipmentList component', () => {
 
     render(<ShipmentList {...props} />);
 
-    userEvent.click(screen.getByRole('button', { name: /^ppm /i }));
+    let editBtn = queryByRole(screen.getAllByTestId('shipment-list-item-container')[0], 'button', { name: 'Edit' });
```
felipe-lee, line 48: Hmm, I also feel like these shouldn't use the test ID (like my comment above), but this one is harder. I wonder if grabbing the text of the shipment ID and then looking at sibling elements or something. Maybe it's just better like you have it.

src/components/ShipmentList/ShipmentList.test.jsx
```diff
@@ -42,38 +45,51 @@ describe('ShipmentList component', () => {
...
 
     await waitFor(() => {
       checkShipmentClick('ID-4', 1, SHIPMENT_OPTIONS.NTSR);
     });
   });
+
+  it('calls onDeleteClick when delete is clicked', () => {
+    render(<ShipmentList {...defaultProps} />);
+    for (let i = 0; i < defaultProps.shipments.length; i += 1) {
+      const deleteBtn = queryByRole(screen.getAllByTestId('shipment-list-item-container')[i], 'button', {
```
felipe-lee, line 85: Same comment here about using `queryBy*` vs `getBy*`

src/shared/Entities/reducer.js
```diff
@@ -54,6 +54,12 @@ const initialState = {
 //   @@swagger/tag.operationId/SUCCESS
 //   ADD_ENTITIES
 export function entitiesReducer(state = initialState, action) {
+  if (action.type === 'UPDATE_MTO_SHIPMENTS_ENTITIY') {
+    return {
+      ...state,
+      mtoShipments: action.entities['mtoShipments'] || {},
+    };
+  }
```
felipe-lee, line 62: Let me preface this by saying, I don't know much (i.e. next to nothing) about redux or how all these things tie together. But, looking at the existing code, is there a reason we need the new path for updating MTO shipments? It _seems_ like the `ADD_ENTITIES` path is doing the same thing, at least in the end. It's doing a merge of the original and new, but if the new overwrites the whole of `mtoShipments` I feel like that should be fine? Again, I could be totally off and if you tried that and it didn't work, then nvm.

src/sagas/entities.js
```diff
@@ -47,6 +48,12 @@ export function* updateMTOShipment(action) {
   yield put(addEntities(normalizedData));
 }
 
+export function* updateMTOShipments(action) {
+  const { payload } = action;
+
+  yield put(updateMTOShipmentsEntity(payload));
+}
+
```
felipe-lee, line 56: Related to my comment on the reducer, it feels like this could follow the pattern of the other things on here, but going for `mtoShipments` as a whole, unlike the existing singular func above your new one.

src/shared/Entities/reducer.js
duncan-truss: I think would could clarify the purpose some and maybe make it a bit generalized if we called it REPLACE_ENTITY to reflect what it's doing now.

src/sagas/entities.js
ronaktruss: I think my answer above answer this. If not, can you rephrase the question/concern here.

src/pages/MyMove/Review/Review.jsx
```diff
@@ -62,7 +62,7 @@ const Review = ({ currentMove, mtoShipments, push, match }) => {
             <div className={formStyles.formActions}>
               <WizardNavigation
                 onNextClick={handleNext}
-                disableNext={!hasCompletedPPMShipments}
+                disableNext={!hasCompletedPPMShipments || !Object.keys(mtoShipments).length}
```
duncan-truss, line 65: mtoShipments should be an array of objects coming back from the selectMTOShipmentsForCurrentMove selector or assigned to an empty array by default. I don't think calling Object.keys is necessary here.

src/components/ShipmentList/ShipmentList.test.jsx
felipe-lee: Ah I didn't notice that. Ok, sounds good.

scarequotes (approved):
This PR works as expected for design. 

One call-out: Deleted shipments work as expected on this page, but then at least sometimes appear on the review screen. I think there's an existing ticket to fix that as a separate story, but just a heads-up that that will need to be fixed at some point.

src/containers/FlashMessage/FlashMessage.jsx
felipe-lee: I don't know anything about the initial implementation, but I think having a dependency array in `useEffect` is a good practice, and like you said, seems to match what the comment indicates we want, so I think this change is probably for the best.

src/pages/MyMove/Home/index.jsx
```diff
@@ -296,6 +357,16 @@ export class Home extends Component {
...
+            {showDeleteSuccessAlert && (
+              <Alert slim type="success">
+                The shipment was deleted.
+              </Alert>
+            )}
+            {showDeleteErrorAlert && (
+              <Alert slim type="error">
+                Something went wrong, and your changes were not saved. Please try again later or contact your counselor.
+              </Alert>
+            )}
```
felipe-lee, line 369: Sorry, you might have addressed this somewhere and I missed it, but do we need the separate alerts? Like does `<ConnectedFlashMessage />` not work if you're staying on the same page?

src/pages/MyMove/Home/index.test.jsx
```diff
@@ -109,7 +109,7 @@ describe('Home component', () => {
         mtoShipmentId: testProps.mtoShipments[1].id,
       });
 
-      wrapper.find('ShipmentListItem').at(0).simulate('click');
+      wrapper.find('ShipmentListItem').at(0).find('button').at(1).simulate('click');
```
felipe-lee, line 112: I'm getting a missing prop error from one of the tests in this file, but I think it wasn't one you updated so I don't think it's a blocker for this PR. Just putting the info here in case you want to address it:
image

src/components/ShipmentList/ShipmentList.stories.jsx
```diff
@@ -8,7 +8,11 @@ import { SHIPMENT_OPTIONS } from 'shared/constants';
 export const Basic = () => (
   <div className="grid-container">
     <h3>Single Shipment</h3>
-    <ShipmentList shipments={[{ id: '0001', shipmentType: SHIPMENT_OPTIONS.PPM }]} onShipmentClick={() => {}} />
+    <ShipmentList
+      shipments={[{ id: '0001', shipmentType: SHIPMENT_OPTIONS.PPM }]}
+      onShipmentClick={() => {}}
+      onDeleteClick={() => {}}
```
felipe-lee, line 14: This has a prop error from the missing `moveSubmitted` prop:

image

src/components/ShipmentList/ShipmentList.stories.jsx
```diff
@@ -18,6 +22,7 @@ export const Basic = () => (
         { id: '0003', shipmentType: SHIPMENT_OPTIONS.PPM },
       ]}
       onShipmentClick={() => {}}
+      onDeleteClick={() => {}}
```
felipe-lee, line 25: This one also needs the `moveSubmitted` prop

felipe-lee (commented):
Found a few prop errors in stories

src/pages/MyMove/Home/index.jsx
ronaktruss: It's not actually clear to me when we use `Alert` vs `ConnectedFlashMessage`. I'd assume and it does generally seem to be the case that `ConnectedFlashMessage` was used when the parent had the ConnectedFlashMessage and the child element would just dispatch the action to update the message. Alerts seem to be used when the update is happening within the same component. With the update above we should be able to use `ConnectedFlashMessage` if we want.

src/components/ConfirmationModals/DestructiveShipmentConfirmationModal.stories.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+export const Basic = Template.bind({});
+
+export const WithOverrides = Template.bind({});
+WithOverrides.args = {
+  title: 'This is a sample title',
+  content: 'Some sample description',
+  submitText: 'YES!',
+  closeText: 'NO',
+  onClose: { action: 'close button clicked' },
```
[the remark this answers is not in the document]
ronaktruss: Ooops, this wasn't supposed to be checked in 😅. As a general note thanks for being thorough and providing references/sources.

src/pages/MyMove/Home/index.jsx
felipe-lee: I think you're right about the general times we use each. I guess I'd opt for, if the page already has `<ConnectedFlashMessage/>` to use it, but maybe this is a discussion for FE check in. I'm fine with keeping this as is for now and if we decide to change it, it can be a follow-up later on.

felipe-lee: I started [a thread in #prac-frontend about this](https://ustcdp3.slack.com/archives/CTQQJD3G8/p1649882920594469) and plan on bringing it up at tomorrow's check in.

src/shared/Entities/reducer.js
felipe-lee: I started [a semi-related thread in #prac-frontend about this](https://ustcdp3.slack.com/archives/CTQQJD3G8/p1649882920594469) and plan on bringing it up at tomorrow's check in.

felipe-lee (approved):
I think given that some things need to discussed at a FE group level, I'm fine with this as is for now. Well pending those prop error fixes, but I don't want to be a blocker on this once you push those fixes up.

PR #8400, opened 2022-04-08
Update move_history_fetcher.go

kennethchow-truss:
## Summary

Move history page was returning all logs no matter what page you were on. Added a `WHERE` clause to filter by move code 

## Setup to Run Your Code

1. Go to officelocal:3000/
2. Log in as an office user
3. Click on a move that has a shipment you can edit
4. Edit the shipment
5. Look at the history page for that same move and see that the log is there
6. Go back to the move table page
7. Click on a different move
8. Click on history
9. Observe that the earlier change does not appear

### DIFF
--- a/pkg/services/move_history/move_history_fetcher.go
+++ b/pkg/services/move_history/move_history_fetcher.go
@@ -34,6 +34,7 @@ func (f moveHistoryFetcher) FetchMoveHistory(appCtx appcontext.AppContext, param
 			mto_shipments.*
 		FROM
 			mto_shipments
+		WHERE move_id = (SELECT moves.id FROM moves) 
 	),
 	shipment_logs AS (
 		SELECT


YanZ777 (dismissed):
These changes look good to me!

pkg/services/move_history/move_history_fetcher.go
```diff
@@ -34,6 +34,7 @@ func (f moveHistoryFetcher) FetchMoveHistory(appCtx appcontext.AppContext, param
 			mto_shipments.*
 		FROM
 			mto_shipments
+		WHERE move_id = (SELECT moves.id FROM moves)
```
donaldthai, line 37: suggestion (blocking): Nice catch! I do think we need a **unit test** to test the changes here.

donaldthai (changes requested):

pkg/services/move_history/move_history_fetcher.go
michaelvaldes-truss: It looks like there are tests failing related to this change -- the unit tests that count the total number of rows returned by this query.
IMO we shouldn't be testing that, since it's really difficult to know how many rows you should be getting back. I think we could test this better by removing the total count tests and just counting something specific to the unit test, like the number of rows corresponding to the shipment updates made in the test.

kennethchow-truss: Another problem this test has is that it depends on two things 

1. Creating actual rows of data for audit_history
2. Creating the data for related tables like move, mto_shipments, addresses, etc.

The move_history_fetcher query depends on both to form the payload that the test is expecting.

kennethchow-truss (commented):
Unit tests added

kennethchow-truss:
@mergifyio rebase

kennethchow-truss:
@mergifyio rebase

kennethchow-truss:
@mergifyio rebase

kennethchow-truss:
@mergifyio rebase

michaelvaldes-truss (approved):
**praise:** Lgtm -- I think the test changes are super helpful!

PR #8410, opened 2022-04-11
[ MB-11317 ] Add orders and orders_logs to query with tests

rogeruiz:
## Summary

Add `orders` and `orders_logs` to the audit history fetch query. This also
un-comments some tests that already written for this query.

**This patch series does not contain any FE changes. That will be another patch linked to this one.**

## Setup to Run Your Code

<details>
<summary>💻 You will need to use at least one separate terminal to test this locally.</summary>

##### Terminal 1

Run the tests locally

```sh
go test ./pkg/services/move_history -v
```

</details>

### Additional steps

1. Login to the Office App as TOO
1. Edit Orders
1. See the Orders updated in the `/history` page


rogeruiz:
@mergifyio rebase

pkg/services/move_history/move_history_fetcher.go
```diff
@@ -55,6 +55,24 @@ func (f moveHistoryFetcher) FetchMoveHistory(appCtx appcontext.AppContext, param
...
 			AND audit_history.object_id = moves.id
 	),
+	orders AS (
+		SELECT
+			orders.*
+		FROM
+			orders
+		JOIN moves ON moves.orders_id = orders.id
+		WHERE moves.id = (SELECT moves.id FROM moves)
+	),
```
donaldthai, line 65: suggestion (non-blocking): I think since we're looking at a specific move locator we don't have to join the moves with the orders, we can just get the order by looking at the moves sub-query. And then we could just do a WHERE orders.id = (SELECT moves.orders_id FROM moves).

pkg/services/move_history/move_history_fetcher_test.go
```diff
@@ -55,17 +55,15 @@ func (suite *MoveHistoryServiceSuite) TestMoveFetcher() {
...
 
-		suite.Equal(totalCount, int64(6), "total count should be 6")
-
 		// address update
 		verifyOldPickupAddress := false
 		verifyNewPickupAddress := false
 		// orders update
-		// verifyOldSAC := false
-		// verifyNewSAC := false
+		verifyOldSAC := false
```
kennethchow-truss, line 65: Nice now that we can finally add these back in

kennethchow-truss (approved):
LGTM

donaldthai (approved):

pkg/services/move_history/move_history_fetcher.go
```diff
@@ -55,6 +55,24 @@ func (f moveHistoryFetcher) FetchMoveHistory(appCtx appcontext.AppContext, param
 		JOIN moves ON audit_history.table_name = 'moves'
 			AND audit_history.object_id = moves.id
 	),
+	orders AS (
+		SELECT
+			orders.*
+		FROM
+			orders
+		JOIN moves ON moves.orders_id = orders.id
+		WHERE moves.id = (SELECT moves.id FROM moves)
```
rogeruiz, line 64:
```suggestion
		WHERE orders.id = (SELECT moves.orders_id FROM moves)
```

pkg/services/move_history/move_history_fetcher_test.go
```diff
@@ -55,17 +55,15 @@ func (suite *MoveHistoryServiceSuite) TestMoveFetcher() {
 		suite.MustSave(&approvedMove)
 
 		params := services.FetchMoveHistoryParams{Locator: approvedMove.Locator, Page: swag.Int64(1), PerPage: swag.Int64(20)}
-		moveHistory, totalCount, err := moveHistoryFetcher.FetchMoveHistory(suite.AppContextForTest(), &params)
+		moveHistory, _, err := moveHistoryFetcher.FetchMoveHistory(suite.AppContextForTest(), &params)
 		suite.FatalNoError(err)
 
-		suite.Equal(totalCount, int64(6), "total count should be 6")
```
YanZ777, line 61: **question:** Was there a reason this assertion was removed?

**thought:** I'm updating the move_history_fetcher_test.go and I'm seeing this fail too.

michaelvaldes-truss: **thought:** I removed some of these "total count" checks already, and I think this is a good change. it's difficult to know at a glance what the correct number _should_ be for any given test, so I think it would be more beneficial to make our test conditions more specific to the different rows we fetch.

rogeruiz: @kennethchow-truss do you have a good reason for use to be counting the total in our tests? I agree with @michaelvaldes-truss and think that this is going to add some flakiness around these tests as we build on this. But if there's a way for us to know that number rather than hardcoding it every time it would certainly help with the flakiness.

YanZ777: That makes sense @michaelvaldes-truss and @rogeruiz. I think having a hardcoded number in this case makes the tests brittle.

rogeruiz: I added this in my last commit here.

> I think we can add this type of checking once the logging work has
stabilized. Right now this is changing so many times, that it makes
sense to remove them. We could have a new ticket around this for once
PO9 is done to help detect changes to the history log totals. But I
don't believe that this type of count check gives us better tests.

PR #8419, opened 2022-04-12
MB-11992 cypress delete shipment

ronaktruss:
## Summary

Cypress tests were added to test the delete functionality on both the home page and the review page.

The tests for the home page are done within the `entireShipment.js` file and the review page tests are done on the `review.js` file.

With my most recent changes include (see code comments for more details):

- restructuring test datagen a bit and cleaning up comments
- update tests to match work correctly with the above testdatagen changes
- restructuring the entire shipment to match the style of our other tests
- update PR based on feedback
- added sumbit to the entire flow

## Setup to Run Your Code

<details>

##### Terminal 1

Start the server locally.

```sh
make server_run
```

##### Terminal 2

Start cypress locally.

```sh
make e2e_test_fresh
```

</details>

### Backend


cypress/integration/milmove/ppms/entireShipment.js
```diff
@@ -50,6 +51,7 @@ describe('Entire PPM onboarding flow', function () {
 function navigateHappyPath(userId, isMobile = false) {
   cy.apiSignInAsUser(userId);
   cy.wait('@getShipment');
+  customerDeletesExistingShipment();
```
felipe-lee, line 54: I think it might be better if we instead start creating (or fully create, either works) another PPM shipment, and then delete it. Like that would mimic the true customer flow more than already having a PPM shipment to delete (like in terms of this `entireShipment` flow. Just my opinion though and not a hard blocker on this.

cypress/integration/milmove/ppms/review.js
```diff
@@ -110,3 +122,12 @@ function navigateToAgreementAndSign() {
   cy.nextPage();
   signAgreement();
 }
+
+function deleteShipment() {
+  cy.get('[data-testid="ShipmentContainer"]').last().contains('Delete').click();
```
felipe-lee, line 127: small nit, I feel weird about just picking the last shipment. I feel like I'd rather target a shipment with its ID, but again, more of an opinion than a blocker.

pkg/testdatagen/scenario/e2ebasic.go
```diff
@@ -1523,71 +1523,53 @@ func serviceMemberWithOrdersAndPPMMove04(appCtx appcontext.AppContext, userUploa
...
-	loginGovID := uuid.Must(uuid.NewV4())
-	testdatagen.MakeUser(appCtx.DB(), testdatagen.Assertions{
-		User: models.User{
-			ID:            uuid.Must(uuid.FromString(uuidStr)),
-			LoginGovUUID:  &loginGovID,
-			LoginGovEmail: email,
-			Active:        true,
-		},
-	})
+	moveInfo := moveCreatorInfo{
```
felipe-lee, line 1526: 🎉 I hope this is actually making it easier for others too haha, but I personally prefer how this looks.

felipe-lee (commented):
I realize that I wrote the edit cypress tests for the homepage and neglected to have it follow through with actually clicking the button to ensure it works, but do you think that should be added in this PR for edit/delete? Kind of how you tested the delete fully on the review and entire shipment flows. 

It might also be something that can be done in https://dp3.atlassian.net/browse/MB-12157 by @duncan-truss. What do y'all think?

My other comments are just opinions so you can take them or leave them.

cypress/integration/milmove/ppms/entireShipment.js
ronaktruss: I thought about fully creating a PPM and then deleting it, but I ended up deciding against it because I didn't think it would help catch any errors that wouldn't be caught by just deleting an existing shipment and would increase the run time of our tests.

I'm curious to hear other people's opinions on this.

cypress/integration/milmove/ppms/review.js
ronaktruss: We utilize the delete function for multiple tests so if we were to delete based on ID we would need to pass the ID for each user/test. Since we are creating both of these tests using testdatagen we should always be selecting the same test when we utilize last. (For context last is utilize instead of first because the first shipment has an ID that is specified in the backend where the latter is generated.)

ronaktruss:
> I realize that I wrote the edit cypress tests for the homepage and neglected to have it follow through with actually clicking the button to ensure it works, but do you think that should be added in this PR for edit/delete? Kind of how you tested the delete fully on the review and entire shipment flows.

Are there specific edit tests that you want to run? Currently `navigateHappyPathWithEditsAndBacks` does handle the edit flow (or at least partially, it makes modifications to some of the fields). We could be a bit more thorough with the fields that are actually updated with new values.

cypress/integration/milmove/ppms/review.js
felipe-lee: Yeah, though we could specify it if we wanted to on create, but again, I'm fine if we don't.

duncan-truss:
@ronaktruss Do you mind updating from master? The diff is showing a lot of previous commits from when you brought in changes from your other branch.

felipe-lee:
> > I realize that I wrote the edit cypress tests for the homepage and neglected to have it follow through with actually clicking the button to ensure it works, but do you think that should be added in this PR for edit/delete? Kind of how you tested the delete fully on the review and entire shipment flows.
> 
> Are there specific edit tests that you want to run? Currently `navigateHappyPathWithEditsAndBacks` does handle the edit flow (or at least partially, it makes modifications to some of the fields). We could be a bit more thorough with the fields that are actually updated with new values.

It would be more to just make sure that clicking the edit button works from the home page, but you're right, we do test that in the other one more thoroughly, so nvm.

felipe-lee:
I think something weird happened with the last merge because it's showing the changes from both your delete branches on this one PR even though the other one has merged to master. I think you might need to merge master into this one.

cypress/integration/milmove/ppms/entireShipment.js
```diff
@@ -79,6 +81,7 @@ function submitAndVerifyUpdateDateAndLocation() {
   cy.get('input[name="pickupPostalCode"]').clear().type('90210').blur();
 
   cy.get('input[name="destinationPostalCode"]').clear().type('76127');
+  cy.get('input[name="sitExpected"]').last().check();
```
duncan-truss, line 84: To make this more explicit you could include the value attribute like:
`input[name="sitExpected"][value="false"]`

ronaktruss:
Just updated the PR to clear up the extra file changes

cypress/integration/milmove/ppms/entireShipment.js
```diff
@@ -116,3 +118,12 @@ function verifyShipmentSpecificInfoOnEstimatedIncentivePage() {
     .and('contain', '76127')
     .and('contain', '15 Apr 2022');
 }
+
+function customerDeletesExistingShipment() {
+  cy.get('[data-testid="shipment-list-item-container"]').contains('Delete').click();
+  cy.get('[data-testid="modal"]').should('be.visible');
+  cy.get('[data-testid="modal"] button').contains('Yes, Delete').click();
```
duncan-truss, line 125: Could you try using [aliases](https://docs.cypress.io/api/commands/as) here for the selectors we reuse like:
cy.get('[data-testid="shipment-list-item-container"]').as('@shipmentContainer')

Also for the modal I might use a [within](https://docs.cypress.io/api/commands/within) here because it is such a small part of the larger page document as to not re-search it when you located it in the prior step.

cypress/integration/milmove/ppms/review.js
```diff
@@ -110,3 +122,12 @@ function navigateToAgreementAndSign() {
   cy.nextPage();
   signAgreement();
 }
+
+function deleteShipment() {
```
duncan-truss, line 126: I see there is some overlap here with the customerDeletesExistingShipment, maybe they could be combined if they take in the shipment container selector and the expected length? Or look at the prior length and test that it's one less?

cypress/integration/milmove/ppms/entireShipment.js
ronaktruss: Question about aliases:
Since we are deleting the element and removing it from the DOM does the alias actually provide value? Doesn't it internally re-query the original get command?

+1 to using within

cypress/integration/milmove/ppms/review.js
ronaktruss: I thought about doing the above before but it didn't seem worthwhile as we would have to create arguments around the selector and the length. With this only being used twice with only ~4 lines shared it didn't seem that worthwhile.

Since some of the above recommendations require some refactor of the shared code I'll take a second pass.

cypress/integration/milmove/ppms/review.js
```diff
@@ -34,22 +34,33 @@ describe('PPM Onboarding - Review', function () {
...
 
   const viewportType = [
-    { viewport: 'desktop', isMobile: false, userId: '6a7d969a-2347-48c7-9289-0963c447f0a7' }, // complete@ppm.unsubmitted
-    { viewport: 'mobile', isMobile: true, userId: 'fd02a7ac-f9cb-49e0-90ab-93a8443c1fc7' }, // complete2@ppm.unsubmitted
+    { viewport: 'desktop', isMobile: false, userId: 'afcc7029-4810-4f19-999a-2b254c659e19' }, // multiComplete@ppm.unsubmitted
+    { viewport: 'mobile', isMobile: true, userId: '836d8363-1a5a-45b7-aee0-996a97724c24' }, // multiComplete2@ppm.unsubmitted
   ];
 
+  viewportType.forEach(({ viewport, isMobile, userId }) => {
+    it(`deletes shipments on the review page and submits the move -  ${viewport}`, () => {
```
duncan-truss, line 47: I would drop the "submits the move" part of the title since it stops short of doing that on these reused moves until the last set of tests.

cypress/integration/milmove/ppms/entireShipment.js
duncan-truss: This says for stale elements it does replay the original commands https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Stale-Elements.  I don't know specifically how it reacts to a deleted element that no longer exists.

I think defining the selector once helps improve the readability but we're unlikely to see a performance gain unless we extended this test.

cypress/integration/milmove/ppms/entireShipment.js
duncan-truss: Right now we're not actually submitting this move from what I can see (maybe we should be?) , so you could in theory delete the shipment at the end. I know your other test is of the delete on the review page so you would need to navigate back to the homepage.  When I think deletion I don't think happy path so we could break it out of this test to make it more obvious what's happening.  Like if you went looking for that shipment in your database (without creating that spare shipment at the beginning) you might end up really confused 😆

cypress/integration/milmove/ppms/entireShipment.js
```diff
@@ -72,13 +83,18 @@ function navigateHappyPathWithEditsAndBacks(isMobile = false) {
...
+  submitMove('@signedCertifications');
 }
 
 // update the form values by submitting and then return to the page to verify if the values persist and then return to the next page
 function submitAndVerifyUpdateDateAndLocation() {
   cy.get('input[name="pickupPostalCode"]').clear().type('90210').blur();
-
+  cy.get('input[name="secondaryPickupPostalCode"]').clear().type('90212').blur();
   cy.get('input[name="destinationPostalCode"]').clear().type('76127');
+  // TODO: The user has secondary destination zips. We should test clearing this value by selecting the no radio btn. This doesn't work atm
```
ronaktruss, line 96: In the current state, our update API does not work when trying to clear out an existing secondary zip to a null value (eg. setting a secondaryDestinationPostalCode from 90210 to null). We should add that scenario as part of this tests when it does work. I'll be posting on slack and creating a bug with more details about this. 

https://user-images.githubusercontent.com/84742904/163599982-7184c644-99c5-4321-b9d3-a6818c49f159.mov

ronaktruss (commented):
With my most recent changes include (see code comments for more details):

- restructuring test datagen a bit and cleaning up comments
- update tests to match work correctly with the above testdatagen changes
- restructuring the entire shipment to match the style of our other tests
- update PR based on feedback
- added sumbit to the entire flow

pkg/services/ppmshipment/ppm_shipment_updater_test.go
```diff
@@ -29,7 +29,7 @@ func (suite *PPMShipmentSuite) TestUpdatePPMShipment() {
 		suite.NilOrNoVerrs(err)
 		suite.Equal(newPPM.PickupPostalCode, updatedPPMShipment.PickupPostalCode)
 		// suite.True(updatedPPMShipment.SitExpected)
-		suite.Equal(unit.Pound(1150), *updatedPPMShipment.ProGearWeight)
+		suite.Equal(unit.Pound(1987), *updatedPPMShipment.ProGearWeight)
```
ronaktruss, line 32: We should probably update these tests to be more thorough as part of the fix for updates for secondary zips.

cypress/integration/milmove/ppms/entireShipment.js
duncan-truss: Nice find 🎉 !

pkg/testdatagen/make_ppm_shipment.go
```diff
@@ -69,13 +69,16 @@ func getDefaultValuesForRequiredFields(db *pop.Connection, shipment models.MTOSh
...
+			Status:                         models.PPMShipmentStatusSubmitted,
+			SecondaryPickupPostalCode:      models.StringPointer("90211"),
+			SecondaryDestinationPostalCode: models.StringPointer("30814"),
+			EstimatedWeight:                models.PoundPointer(unit.Pound(4000)),
+			HasProGear:                     models.BoolPointer(true),
+			ProGearWeight:                  models.PoundPointer(unit.Pound(1987)),
+			SpouseProGearWeight:            models.PoundPointer(unit.Pound(498)),
+			EstimatedIncentive:             models.Int32Pointer(int32(1000000)),
+			AdvanceRequested:               models.BoolPointer(true),
+			Advance:                        models.CentPointer(unit.Cents(598700)),
```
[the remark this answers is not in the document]
duncan-truss: One thing to keep in mind with adding all of these defaults is that you can't easily set any of these fields to nil anymore because mergeModels won't update any fields set to nil in the assertions struct.  Maybe that's fine and people can always use the minimal version to add in what they want.  I might expect this factory to set the minimum fields necessary to make up a complete PPM shipment.

pkg/testdatagen/scenario/e2ebasic.go
duncan-truss: I am okay with this, we could have also passed in testdatagen.Assertions itself with those fields set already but not sure that's clearly better.

cypress/support/ppmShared.js
```diff
@@ -177,3 +180,23 @@ export function navigateFromReviewPageToHomePage() {
...
+  signAgreement();
+}
+
+export function deleteShipment(selector, expectedLength) {
+  cy.get(selector).contains('Delete').click();
+  cy.get('[data-testid="modal"]').within(($modal) => {
+    expect($modal).to.be.visible;
+    cy.get('button').contains('Yes, Delete').click();
+  });
+  cy.waitFor(['@deleteShipment', '@getShipment']);
```
duncan-truss, line 195: Haha I think you've been writing too many React Testing Library tests.  This should just be `cy.wait(['@deleteShipment', '@getShipment'])`.

Searching the project I see a bunch of other cy.waitFor calls in other tests!  I think it's used for checking for javascript events.

duncan-truss (approved):
Thanks for making those changes 👍  The only remaining comment I had was fixing the `cy.wait`

pkg/testdatagen/make_ppm_shipment.go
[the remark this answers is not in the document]
felipe-lee: **thought**: To your point of needing to pass in values each time, I'm of the opinion that we _should_ have to pass in something (whether it's a value or a trait) to the general `Make*` function for data we need or expect that isn't the common case. In working on the proof-of-concept PR for the fake data generation, I realized just how fragile some of our tests are because they rely on hard-coded values set many levels deep that really should be explicitly stated in the test if the test needs those values. But I realize that's not necessarily something everyone agrees on.

pkg/testdatagen/scenario/e2ebasic.go
felipe-lee: We can't use `testdatagen.Assertions` to pass in IDs for all models because a bunch of our `Make*` functions are set up such that if an ID is present for a related model, it assumes that the model exists and doesn't call the creation function. 

I hadn't really noticed thought about the implications of how our `Make*` functions do that until I'd asked Pearl to set a ID at the shipment level and realized it skipped creating the `MTOShipment` because it saw the ID (and that many other funcs follow that pattern.

pkg/testdatagen/scenario/shared.go
```diff
@@ -761,18 +761,18 @@ func createUnSubmittedMoveWithPPMShipmentThroughAdvanceRequested(appCtx appconte
 	createGenericUnSubmittedMoveWithPPMShipment(appCtx, moveInfo, true, assertions)
 }
 
-func createUnSubmittedMoveWithFinishedPPMShipment(appCtx appcontext.AppContext, userUploader *uploader.UserUploader) {
+func createUnSubmittedMoveWithFullPPMShipment1(appCtx appcontext.AppContext, userUploader *uploader.UserUploader) {
```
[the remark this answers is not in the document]
felipe-lee: Good point. `Finished` would likely be good for the ones when the customer has uploaded all their weight tickets and such and everything is done.

cypress/integration/milmove/ppms/entireShipment.js
felipe-lee: **praise**: Great job catching this! Thorough testing ftw 🎉 

We may end up getting rid of secondary zips, but if we don't it's good to have this in our backlog now to make sure we address it.

felipe-lee (approved):
Mostly suggestions, comments, and responses, but lgtm!

ronaktruss:
+1 for having a function that creates a PPM shipment with only the minimal information needed to get that move to completed state. I'd argue for having the generic make ppm shipments to be a function that has some additional fields as well in this scenario since the additional fields don't fundamentally change how or what business rules are applied.

pkg/testdatagen/scenario/e2ebasic.go
duncan-truss: That's true, I was thinking just reusing assertions as a temporary data structure like you're doing for moveInfo, not that you could use it without the extra added setup work

PR #8430, opened 2022-04-13
[MB-11451] loadtest rds updates

sheenamt:
## Summary
This adds a new variable to the env files for the RDS certificate (DB_SSL_ROOT_CERT). Currently only adding to loadtest.

This is updating how the dp3 sites are deployed by cleaning up the million steps and parameterizing things where possible. Currently, only loadtest is updated. 

Demo and Exp will have these updates done shortly as well. 

[MB-11451]: https://dp3.atlassian.net/browse/MB-11451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -275,27 +275,14 @@ commands:
         description: dp3 env to deploy to (exp, loadtest, demo)
         type: string
         default: *dp3-env
-      dp3-region:
-        type: string
-        default: << parameters.dp3-env >>_REGION
-      dp3-account-id:
-        type: string
-        default: << parameters.dp3-env >>_ACCOUNT_ID
-      dp3-access-key-id:
-        type: string
-        default: << parameters.dp3-env >>_ACCESS_KEY_ID
-      dp3-secret-access-key:
-        type: string
-        default: << parameters.dp3-env >>_SECRET_ACCESS_KEY
-
     steps:
       - run:
           name: 'Setting up AWS environment variables for dp3 env defined in &dp3-env'
           command: |
-            echo "export AWS_DEFAULT_REGION=<< parameters.dp3-region >>" >> $BASH_ENV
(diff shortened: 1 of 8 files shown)


travelar (approved):

.circleci/config.yml
```diff
@@ -266,24 +270,35 @@ commands:
             echo "export AWS_ACCESS_KEY_ID=$DEV_ACCESS_KEY_ID" >> $BASH_ENV
             echo "export AWS_SECRET_ACCESS_KEY=$DEV_SECRET_KEY" >> $BASH_ENV
             source $BASH_ENV
-  aws_vars_loadtest:
+
+  aws_vars_dp3:
+    parameters:
+      dp3-env:
+        description: dp3 env to deploy to (exp, loadtest, demo)
```
ahobson, line 277: I really appreciate all the comments you've added! I see on line 38 it says it must be one of loadtest, demo, experimental but in this comment it uses "exp". I wonder if it might be better to say "experimental" here, too?

.circleci/config.yml
```diff
@@ -266,24 +270,35 @@ commands:
...
+        type: string
+        default: *dp3-env
     steps:
       - run:
-          name: 'Setting up AWS environment variables for loadtest env'
+          name: 'Setting up AWS environment variables for dp3 env defined in &dp3-env'
           command: |
-            echo "export AWS_DEFAULT_REGION=$LOADTEST_REGION" >> $BASH_ENV
```
ahobson, line 274: I wonder if it would catch some errors earlier if we had something like ...

```sh
case "<< parameters.dp3-env >>" in
  demo|experimental|loadtest)
    echo "Using '<< parameters.dp3-env >>'"
    ;;
  *)
    echo "Unknown dp3-env: '<< parameters.dp3-env >>'"
    exit 1
    ;;
esac
```

PR #8431, opened 2022-04-13
Added filter for mto_service_items

kennethchow-truss:
## Summary

Ali was seeing an issue where he was seeing service items that weren't related to his current move. This is because there is a missing filter on the where clause of the mto_service items query. 

## Setup to Run Your Code

1. Go to officelocal:3000/
2. Sign is a TOO user
3. Go to http://officelocal:3000/moves/WTSTAT/history
4. Approve a move
5. Go to the history page
6. Observe that the service items are there
7. Go to a different move's history page
8. Observe that the service items that were on WTSTAT are not on your current page.

### DIFF
--- a/pkg/services/move_history/move_history_fetcher.go
+++ b/pkg/services/move_history/move_history_fetcher.go
@@ -63,7 +63,7 @@ func (f moveHistoryFetcher) FetchMoveHistory(appCtx appcontext.AppContext, param
 			mto_service_items
 		JOIN re_services ON mto_service_items.re_service_id = re_services.id
 		JOIN mto_shipments ON mto_service_items.mto_shipment_id = mto_shipment_id
-		WHERE mto_shipments.move_id = (SELECT moves.id FROM moves)
+		WHERE mto_shipments.move_id = (SELECT moves.id FROM moves) AND mto_service_items.move_id = (SELECT moves.id FROM moves)
 	),
 	service_item_logs AS (
 		SELECT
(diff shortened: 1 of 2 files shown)


pkg/services/move_history/move_history_fetcher.go
```diff
@@ -63,7 +63,7 @@ func (f moveHistoryFetcher) FetchMoveHistory(appCtx appcontext.AppContext, param
 			mto_service_items
 		JOIN re_services ON mto_service_items.re_service_id = re_services.id
 		JOIN mto_shipments ON mto_service_items.mto_shipment_id = mto_shipment_id
-		WHERE mto_shipments.move_id = (SELECT moves.id FROM moves)
+		WHERE mto_shipments.move_id = (SELECT moves.id FROM moves) AND mto_service_items.move_id = (SELECT moves.id FROM moves)
```
donaldthai, line 66: suggestion (blocking): I think the real question might be the join on the mto_shipments table. The `JOIN mto_shipments ON mto_service_items.mto_shipment_id = mto_shipment_id` should be `JOIN mto_shipments ON mto_service_items.mto_shipment_id = mto_shipments.id`.

kennethchow-truss: Wow incredibly sharp eye. Thanks for the lookout

donaldthai (approved):
Looks good! 👍

michaelvaldes-truss (approved):
Thanks for catching this!
This worked for me; I'm seeing a reduction in the number of rows fetched by the query, and the move histories for different move now seem to be unique.

PR #8432, opened 2022-04-13
MB-12275: circleci speedups

ahobson:
## Summary

This PR has 4 main changes

1. It ensures that the babel and terser cache is saved in CI only on builds on the master branch. Using the cache can chop 1-2 minutes off the client build
2. Builds the client code on a xlarge instance with a ramdisk. This saves another 1 minute or so
3. Maximizes the parallelism for every integration test. Every integration test finishes in less than ~6 minutes
4. Reworks the mtls e2e test to work in a manner similar to the other e2e tests. It now finishes in ~2 minutes.

The most recent PR run finishes in just under 14 minutes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -936,8 +936,8 @@ jobs:
       - announce_failure
 
   integration_tests_milmove:
-    # parallelism requires enough files to divide by the # of parallel threads. So you need at least 4 files here
-    parallelism: 4
+    # parallelism requires enough files to divide by the # of parallel threads. So you need at least 6 files here
+    parallelism: 6
     executor: milmove_integration_tester
     steps:
       - attach_workspace:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -688,10 +688,6 @@ jobs:
       - run: rm -rf pkg/assets/assets.go && make pkg/assets/assets.go
       - run: make server_generate mocks_generate
       - run: scripts/check-generated-code pkg/gen/ $(find . -type d -name "*mocks" -exec echo -n '{} ' \;)
-      - save_cache:
-          key: go-mod-sources-v3-{{ checksum "go.sum" }}-{{ checksum "scripts/check-go-version" }}
-          paths:
-            - '~/go'
       - announce_failure
 
   # `anti_virus` uses virus detection software to scan the source code
@@ -1143,10 +1139,10 @@ jobs:
       # see https://www.npmjs.com/package/terser-webpack-plugin/v/1.4.4#cache
       - restore_cache:
           keys:
-            - v1-node-modules-cache
+            - v2-node-modules-cache-{{ checksum "yarn.lock" }}
       - run: make client_build
       - save_cache:
-          key: v1-node-modules-cache
+          key: v2-node-modules-cache-{{ checksum "yarn.lock" }}
           paths:
             - ./node_modules/.cache
       - persist_to_workspace:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -871,8 +871,8 @@ jobs:
       - announce_failure
 
   integration_tests_office:
-    # parallelism requires enough files to divide by the # of parallel threads. So you need at least 8 files here
-    parallelism: 8
+    # parallelism requires enough files to divide by the # of parallel threads. So you need at least 12 files here
+    parallelism: 12
     executor: milmove_integration_tester
     steps:
       - attach_workspace:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,6 +99,18 @@ executors:
         environment:
           GOPATH: /home/circleci/go
 
+  mymove_compiler_xlarge:
+    # large and no ram disk
+    resource_class: xlarge
+    working_directory: ~/transcom/mymove
+    docker:
+      - image: *circleci-docker
+        auth:
+          username: $DOCKERHUB_USERNAME
+          password: $DOCKERHUB_PASSWORD
+        environment:
+          GOPATH: /home/circleci/go
+
   mymove_ramdisk_compiler:
     # large and no ram disk
     resource_class: large
@@ -1097,14 +1109,18 @@ jobs:
   # items are in the workspace, we don't even have to checkout the
   # code again
   #
-  # the babel/terser cache below takes the compile from ~6-7m to ~4-5m !!
+  # the babel/terser cache below takes the compile step from ~6-7m to ~4-5m !!
   compile_app_client:
-    executor: mymove_ramdisk_compiler
-    working_directory: /mnt/ramdisk
+    # ahobson tested this and saw 75% of the 16GB being used
+    # (according to the resources tab on CircleCI)
+    # if it starts running out of memory, we should move away from
+    # using a ramdisk
+    #
+    # This entire step takes ~6 minutes with xlarge ramdisk
+    # This entire step takes ~7 minutes without xlarge ramdisk
+    executor: mymove_ramdisk_compiler_xlarge
     steps:
       - checkout
-      # use the yarn.lock as a key so that PRs with the same
-      # dependencies will use the same cache
       - restore_cache:
           keys:
             - v3-cache-yarn-v3-{{ checksum "yarn.lock" }}
@@ -1118,6 +1134,9 @@ jobs:
       # cacheDirectory
       #
       # see https://www.npmjs.com/package/terser-webpack-plugin/v/1.4.4#cache
+      #
+      # use the yarn.lock as a key so that PRs with the same
+      # dependencies will use the same cache
       - restore_cache:
           keys:
             - v2-node-modules-cache-{{ checksum "yarn.lock" }}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -726,15 +726,6 @@ jobs:
       - restore_cache:
           keys:
             - v1-pre-commit-dot-cache-{{ checksum ".pre-commit-config.yaml" }}
-      - run:
-          name: Output processes and resource usage every 5 seconds while job is running
-          command: |
-            while true; do
-              sleep 5
-              ps auxwwf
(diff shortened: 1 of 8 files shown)


cmd/prime-api-client/utils/shared.go
```diff
@@ -39,6 +39,8 @@ const (
 	PortFlag string = "port"
 	// InsecureFlag indicates that TLS verification and validation can be skipped
 	InsecureFlag string = "insecure"
+	// WaitFlat is how long to wait for the server to respond
```
sandy-wright, line 42: There's a small typo in this comment (says `WaitFlat` instead of `WaitFlag`). Also, an example of the format for this flag could be helpful, since time durations are represented differently all the time. 60s vs 60000, etc

cmd/prime-api-client/main.go
```diff
@@ -24,6 +25,7 @@ func initRootFlags(flag *pflag.FlagSet) {
 	flag.Bool(utils.InsecureFlag, false, "Skip TLS verification and validation")
 	flag.String(utils.FilenameFlag, "", "The name of the file being passed in")
 	flag.String(utils.IDFlag, "", "The UUID of the object being retrieved or updated")
+	flag.Duration(utils.WaitFlag, time.Second*60, "duration to wait for server to respond")
```
sandy-wright, line 28: Oof, did we really have to bump up the wait time to accommodate `listMoves`? That doesn't feel so good 😬

cmd/prime-api-client/prime/list_moves.go
```diff
@@ -77,22 +78,38 @@ func ListMoves(cmd *cobra.Command, args []string) error {
...
-	resp, err := primeGateway.MoveTaskOrder.ListMoves(&params)
-	if err != nil {
-		return utils.HandleGatewayError(err, logger)
-	}
+	startTime := time.Now()
+
+	// this wait retry logic would need to be replicated to all
+	// commands, so start with list moves for now
+	wait := v.GetDuration(utils.WaitFlag)
+	params.SetTimeout(wait)
```
sandy-wright, line 86: I like this! Nice to be flexible instead of hardcoding things

cmd/prime-api-client/prime/list_moves.go
```diff
@@ -77,22 +78,38 @@ func ListMoves(cmd *cobra.Command, args []string) error {
...
-			logger.Fatal(errJSONMarshall)
+			return nil
 		}
-		fmt.Println(string(payload))
-	} else {
-		logger.Fatal(resp.Error())
 	}
 
-	return nil
 }
```
sandy-wright, line 115: I'm out of date with my go best practices for sure, but just to double-check - we won't encounter any issues by not specifying a base-level `return nil`, right? Certainly feels better not having to rewrite it over and over, but I want to make sure there won't be a runtime error about type disagreements or something lousy like that

scripts/run-e2e-mtls-test
```diff
@@ -0,0 +1,45 @@
...
+    --keypath config/tls/devlocal-mtls.key \
+    --insecure \
+    --hostname primelocal \
+    --port 9443 \
+    "$@"
+}
+
+
+# wait for 5 minutes for the server to be ready
+prime_api_client --wait 5m list-moves
```
sandy-wright, line 23: I'm fine with this for now and realize there was a specific purpose for this PR, but is this really how we want to run integration tests? Last time I was on the project, the certs for staging had made it so the Prime API Client was all but deprecated. We could do integration tests within the go server/pkg and use load testing for further acceptance. 

Or do something else! I just had the general vibe that folks didn't enjoy using the PAC, plus having to write tests in bash is not the most fun thing in the world.

But - clearly this PR is for something else and these were moved out to a new file for the sake of the refactor. So this is fine! Just wanting to bring up this convo again so it's top-of-mind.

docker-compose.mtls.yml
```diff
@@ -4,45 +4,25 @@ services:
...
-      - '9443:9443'
+    networks:
+      default:
+        aliases:
+          - primelocal
+    image: ${MILMOVE_MTLS_IMAGE:-milmove_mtls:local}
+    entrypoint:
+      - "/bin/sh"
+      - "-c"
+      - "/bin/milmove migrate && /bin/generate-test-data --named-scenario='e2e_basic' && /bin/milmove serve"
```
sandy-wright, line 25: much more concise 👍

.circleci/config.yml
```diff
@@ -614,10 +626,21 @@ commands:
             docker cp cypress:/home/circleci/cypress/videos cypress/ 2>/dev/null || echo "No cypress videos copied"
             docker cp cypress:/home/circleci/cypress/reports cypress/ 2>/dev/null || echo "No cypress reports copied"
 
+  # this custom step has some duplication with
+  # scripts/run-e2e-mtls-test-docker because we want to run some of the
+  # steps below in parallel
   e2e_tests_mtls:
     steps:
+      - setup_remote_docker:
+          docker_layer_caching: false
```
sandy-wright, line 635: What was your reasoning behind removing the caching here? Just curious, seems like a counter-productive step to speeding up the tests, but I'm sure there are other valid reasons for wanting a fresh build whenever something changes

.circleci/config.yml
```diff
@@ -1141,14 +1134,23 @@ jobs:
...
+      # only save the cache on master builds so that PRs don't pollute
+      # the cache
+      - when:
+          condition:
+            equal: [ master, << pipeline.git.branch >>]
+          steps:
+            save_cache:
+              key: v2-node-modules-cache-{{ checksum "yarn.lock" }}
+              paths:
+                - ./node_modules/.cache
```
sandy-wright, line 1153: This makes a lot of sense and is pretty neat, although I don't totally understand the mechanisms behind this step. It is distinctly different from the docker layer caching though, right? Just curious!

sandy-wright (commented):
This looks good to me so far! I'm not totally clear on how best to test this though. What do you recommend, @ahobson?

.circleci/config.yml
ahobson: We have to pay more for docker layer caching, so that's why this is turned off everywhere we use docker.

cmd/prime-api-client/main.go
ahobson: When the server is first starting up, we need to wait until it's ready. So it's not that listMoves is slow, it's the initial server startup

scripts/run-e2e-mtls-test
```diff
@@ -0,0 +1,45 @@
...
+# now that the server is ready, actually run the tests and parse the JSON response
+
+# Integration test/health check for Prime API
+# Return moves from the list-moves endpoint where available_to_prime_at
+# is NOT NULL
+moves=$(prime_api_client --wait 5m list-moves)
```
ahobson, line 29: Don't need this 5m wait on the second call!

cmd/prime-api-client/main.go
ahobson: This is only for the first request while we wait for the server to be ready. See `scripts/run-e2e-mtls-test` for when it is used

cmd/prime-api-client/prime/list_moves.go
ahobson: In this case, the go compiler is smart enough to see this is a infinite loop without a break statement, so any code after the loop would be unreachable

.circleci/config.yml
ahobson: You'll notice that we're actually just moving this line from below, so it's the same docker setup as before, but in a new spot

cmd/prime-api-client/main.go
sandy-wright: Oh I see, my confusion was that the default in most Prime API Client commands is 30 seconds, not 60. So I thought this was a purposeful change, but it may have been a typo or oversight? Is that correct?

cmd/prime-api-client/prime/list_moves.go
sandy-wright: great, that makes sense! thank you!

.circleci/config.yml
```diff
@@ -626,16 +649,15 @@ commands:
...
-            LOGIN_GOV_MY_CLIENT_ID: urn:gov:gsa:openidconnect.profiles:sp:sso:dod:mymovemillocal
-            LOGIN_GOV_OFFICE_CLIENT_ID: urn:gov:gsa:openidconnect.profiles:sp:sso:dod:officemovemillocal
-            LOGIN_GOV_ADMIN_CLIENT_ID: urn:gov:gsa:openidconnect.profiles:sp:sso:dod:adminmovemillocal
-            LOGIN_GOV_HOSTNAME: idp.int.identitysandbox.gov
-            DOD_CA_PACKAGE: /home/circleci/transcom/mymove/config/tls/Certificates_PKCS7_v5.6_DoD.der.p7b
+            docker-compose -f docker-compose.mtls.yml up
+      - run:
+          name: run e2e mtls tests
+          command: |
+            docker run \
```
shimonacarvalho, line 656: Much tidier this way!

shimonacarvalho (approved):
Thanks for the walkthrough, looks good other than the one comment related thing.

cmd/prime-api-client/main.go
ahobson: Oh, I see now. Yes, the the default should be 30s. Updated!

.circleci/config.yml
```diff
@@ -99,6 +99,18 @@ executors:
...
+      - image: *circleci-docker
+        auth:
+          username: $DOCKERHUB_USERNAME
+          password: $DOCKERHUB_PASSWORD
+        environment:
+          GOPATH: /home/circleci/go
+
   mymove_ramdisk_compiler:
     # large and no ram disk
     resource_class: large
```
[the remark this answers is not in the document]
ahobson: Good catch. Fixed!

sandy-wright (approved):
Thanks for making those updates!

PR #8435, opened 2022-04-14
[MB-10787] Adjust endpoint to allow SC to approve a PPM-only move

reggieriser:
## Summary

This PR updates the existing `PATCH` endpoint for completing services counseling so that it can be called when a service counselor is approving a PPM-only move.  The URL for the endpoint is:
`/move-task-orders/{moveTaskOrderID}/status/service-counseling-completed`

When called with a move that has PPM shipments only, the endpoint should adjust the following statuses (see this [Miro board](https://miro.com/app/board/o9J_lqJFWxk=/?moveToWidget=3458764515331487801&cot=14) and this [Slack conversation](https://ustcdp3.slack.com/archives/C0250Q8469K/p1649358168020879)):

- the `moves` record becomes `APPROVED`
- the `mto_shipments` record for the PPM becomes `APPROVED`
- the `ppm_shipments` record for the PPM becomes `WAITING_ON_CUSTOMER`

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

You should be able to test this in the office app by submitting the move details of a PPM-only move that has been directed to a services counselor.  Verify that the `moves`, `mto_shipments`, and `ppm_shipments` records have had their status updated in the database as described above.

There are tests in this PR that exercise both the handler and the service (as well as some new functionality in the move router).

### Backend


kylehilltruss (approved):
I've tested this on a separate branch off of this one and confirmed that the front end can submit a PPM-only move for approval and interact with the endpoint correctly.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -94,6 +94,106 @@ func (o moveTaskOrderUpdater) UpdateStatusServiceCounselingCompleted(appCtx appc
...
+				ppmOnlyMove = false
+				break
+			}
+		}
+		if !ppmOnlyMove {
+			return apperror.NewConflictError(move.ID, "Move should only contain PPM shipments")
+		}
+
+		// Set move status to APPROVED
+		move.Status = models.MoveStatusAPPROVED
```
duncan-truss, line 136: I know generally across the office app the MoveRouter is responsible for consolidating all move status transitions.  I see another instance in this file not using it but other methods are using it, as it's already included in the struct.

swagger-def/ghc.yaml
```diff
@@ -1394,6 +1394,48 @@ paths:
       description: Changes move (move task order) status to service counseling completed
       operationId: updateMTOStatusServiceCounselingCompleted
       summary: Changes move (move task order) status to service counseling completed
+  '/move-task-orders/{moveTaskOrderID}/status/service-counseling-ppm-approved':
```
duncan-truss, line 1397: Part of me wishes we could use the existing services counseling completed endpoint so that the client doesn't have to dynamically know which API to call.  I know this way leads to simpler assumptions in the service, even though we are still checking that it was called with only PPM shipments.  I'm just wondering how it will grow for PPM + HHG partial moves.

pkg/handlers/ghcapi/move_task_order.go
```diff
@@ -192,6 +192,72 @@ func (h UpdateMTOStatusServiceCounselingCompletedHandlerFunc) Handle(params move
...
+			moveTaskOrderPayload := payloads.Move(mto)
+
+			// Audit
+			_, err = audit.Capture(appCtx, mto, moveTaskOrderPayload, params.HTTPRequest)
+			if err != nil {
+				appCtx.Logger().Error("Auditing service error for transitioning move status to approved and PPM status to waiting on customer.", zap.Error(err))
+				return movetaskorderops.NewUpdateMTOStatusServiceCounselingPPMApprovedInternalServerError(), err
+			}
+
+			_, err = event.TriggerEvent(event.Event{
```
duncan-truss, line 245: I don't think the Prime has any role to play here by notifying of this event. If this is counseled by the services counselor and won't be able to retrieve this move because we haven't set it available to the prime.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -94,6 +94,106 @@ func (o moveTaskOrderUpdater) UpdateStatusServiceCounselingCompleted(appCtx appc
...
+		}
+		if err != nil {
+			return err
+		}
+
+		// Set MTO shipment status to APPROVED and set PPM shipment status to WAITING_ON_CUSTOMER
+		// Note: Avoiding the copy of the element in the range so we can preserve the changes to the
+		// statuses when we return the entire move tree.
+		for i := range move.MTOShipments { // We should only have PPM shipments if we get to here.
+			move.MTOShipments[i].Status = models.MTOShipmentStatusApproved
```
duncan-truss, line 156: We do also have a ShipmentRouter, akin to the MoveRouter, for transitioning statuses but I don't know if it is as well adopted.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -94,6 +94,106 @@ func (o moveTaskOrderUpdater) UpdateStatusServiceCounselingCompleted(appCtx appc
...
 
+func (o moveTaskOrderUpdater) UpdateStatusServiceCounselingPPMApproved(appCtx appcontext.AppContext, moveTaskOrderID uuid.UUID, eTag string) (*models.Move, error) {
+	var err error
+	var verrs *validate.Errors
+
+	searchParams := services.MoveTaskOrderFetcherParams{
+		IncludeHidden:   false,
+		MoveTaskOrderID: moveTaskOrderID,
+	}
+	move, err := o.FetchMoveTaskOrder(appCtx, &searchParams)
```
duncan-truss, line 105: I think it's likely that we'll need to add in the PPM Shipment association here for when the Prime needs to view PPM shipment data, if that influences your thinking in loading the PPM Shipments below.

pkg/services/move_task_order/move_task_order_updater.go
reggieriser: I added some functionality to the move router to support approving a PPM-only move and have incorporated that now.

swagger-def/ghc.yaml
reggieriser: That makes sense -- I think I was a bit confused about how the front-end would change for this story which led me to create a separate endpoint.  I've now merged my new functionality into the existing services counseling completed endpoint.

pkg/handlers/ghcapi/move_task_order.go
reggieriser: Since I refactored this to not create a new endpoint, I'm inclined to leave this as it was already in the existing endpoint.  I think the event will just get ignored for now (in terms of the prime) since the move would still be unavailable to the prime, but perhaps we want to trigger the event anyway in case those rules change.

pkg/services/move_task_order/move_task_order_updater.go
reggieriser: I didn't end up using the shipment router for this.  Given how PPMs are a separate table, it may require more thought and perhaps a separate story if we want to do this.  I also noticed that the `mto_shipments` and `ppm_shipments` records are still in `DRAFT` after they're submitted in the customer app -- I think that could cause issues when using the shipment router until that's fixed (if that's actually an incorrect status).

pkg/services/move_task_order/move_task_order_updater.go
reggieriser: I did go ahead and move the association into the fetcher since we may need it elsewhere.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -38,57 +38,85 @@ func NewMoveTaskOrderUpdater(builder UpdateMoveTaskOrderQueryBuilder, serviceIte
...
 
-	for _, s := range move.MTOShipments {
-		if s.ShipmentType == models.MTOShipmentTypeHHGOutOfNTSDom && s.StorageFacilityID == nil {
-			return &models.Move{}, apperror.NewConflictError(
-				s.ID, "NTS-release shipment must include facility info")
+	transactionError := appCtx.NewTransaction(func(txnAppCtx appcontext.AppContext) error {
+		// Update move status, verifying that move/shipments are in expected state.
+		err := o.moveRouter.CompleteServiceCounseling(appCtx, move)
+		if err != nil {
+			return apperror.NewConflictError(move.ID, err.Error())
```
duncan-truss, line 61: I think these are already conflict errors coming back from the move router so you can probably return them directly.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -38,57 +38,85 @@ func NewMoveTaskOrderUpdater(builder UpdateMoveTaskOrderQueryBuilder, serviceIte
...
+		//   - set PPM shipment status to WAITING_ON_CUSTOMER
+		// TODO: Perhaps this could be part of the shipment router. PPMs are a separate model/table,
+		//   so would need to figure out how they factor in.
+		if ppmOnlyMove {
+			// Note: Avoiding the copy of the element in the range so we can preserve the changes to the
+			// statuses when we return the entire move tree.
+			for i := range move.MTOShipments { // We should only have PPM shipments if we get to here.
+				move.MTOShipments[i].Status = models.MTOShipmentStatusApproved
+
+				verrs, err = appCtx.DB().ValidateAndSave(&move.MTOShipments[i])
```
duncan-truss, line 93: Do you have any insight into why we don't use updateColumns() anywhere in the codebase?  My guess is we'd need to call validate() separately and we tend to already have the full object we need to check for things like not found errors and eTags.  It looks like we'd need to remember to include updated_at every time too.

pkg/services/move_task_order/move_task_order_updater.go
```diff
@@ -38,57 +38,85 @@ func NewMoveTaskOrderUpdater(builder UpdateMoveTaskOrderQueryBuilder, serviceIte
...
+					return apperror.NewInvalidInputError(move.MTOShipments[i].ID, nil, verrs, "")
+				}
+				if err != nil {
+					return err
+				}
+
+				if move.MTOShipments[i].PPMShipment != nil {
+					move.MTOShipments[i].PPMShipment.Status = models.PPMShipmentStatusWaitingOnCustomer
+
+					verrs, err = appCtx.DB().ValidateAndSave(move.MTOShipments[i].PPMShipment)
```
duncan-truss, line 104: So here's a non-obvious question.  Since PPMShipment uses the [newer validator pattern](https://github.com/transcom/mymove/blob/1695b0a9c55dc86d799874a12712d615e6b0ae70/pkg/services/ppmshipment/validation.go#L21), does it feel overkill to you that we should be exporting that and calling that here from outside of the package?

pkg/services/move_task_order/move_task_order_updater.go
reggieriser: Good catch!  I had moved some error checking into the move router and forgot to change the error handling here.

pkg/services/move_task_order/move_task_order_updater.go
reggieriser: Good question.  I'm not sure, unless it's just a case where that's not a well-publicized part of the API (I didn't see any references to it after a quick glance in the Pop docs).  It also looks that it was introduced in [Pop 4.12.1](https://github.com/gobuffalo/pop/releases/tag/v4.12.1) which came out after MilMove had been in development for a while.  I think your point about `updated_at` is a reasonable consideration, too.  Is it worth it to use `UpdateColumns` if it means those calls are now dependent upon remembering `updated_at` as well as explicitly listing column names (which means a column name change now has to touch these calls in a way that it doesn't with a `ValidateAndSave`, for instance)?  I'm not against it, but I think we'd need to be careful and intentional about when we do use it.

duncan-truss: Yea, I don't mean to introduce it now but thought you'd be one of the better people to ask. We could wrap it pretty easily to make sure updated_at was always included.

pkg/services/move_task_order/move_task_order_updater.go
reggieriser: I'm not sure.  It's possible the shipment router is an answer, but not sure if it's worth holding this PR up longer to sort that out (also given that the PPM move/shipment records appear to be `DRAFT`).  That said, I do wonder if I should be using the MTO shipment updater and the PPM shipment updater services here instead (which would force it through the validator, right)?

duncan-truss: Yes that's right, if you used those updater services it would force it through those validators.

reggieriser: @duncan-truss and I talked through this.  I first looked at injecting the MTO shipment updater and PPM shipment updaters.  Turns out there's 28 or so usages of the `MoveTaskOrderUpdater` I'd have to change, then injecting just the `MTOShipmentUpdater` into that would requires seven dependencies to be injected for it for each usage.  We thought about just instantiating the `MTOShipmentUpdater` directly in my service for now (as opposed to injecting it in the handler) but it would probably be tough gathering all the things to inject there like a route planner.  So, for now, we agreed to leave this as is since it's just adjusting a single status field.  Perhaps the validators could be exposed in a different way later on if we want to pursue this further.

pkg/handlers/ghcapi/move_task_order_test.go
```diff
@@ -265,69 +260,114 @@ func (suite *HandlerSuite) TestUpdateMTOStatusServiceCounselingCompletedHandler(
...
+		}
+
 		response := handler.Handle(params)
 		suite.IsNotErrResponse(response)
 		moveTaskOrderResponse := response.(*movetaskorderops.UpdateMTOStatusServiceCounselingCompletedOK)
 		moveTaskOrderPayload := moveTaskOrderResponse.Payload
 
-		suite.Assertions.IsType(&move_task_order.UpdateMTOStatusServiceCounselingCompletedOK{}, response)
-		suite.Equal(strfmt.UUID(moveTaskOrder.ID.String()), moveTaskOrderPayload.ID)
+		suite.IsType(&move_task_order.UpdateMTOStatusServiceCounselingCompletedOK{}, response)
```
duncan-truss, line 281: I think we are supposed to be validating the response types too for load testing parsing now https://transcom.github.io/mymove-docs/docs/api/guides/guide-to-creating-an-endpoint#validating-response-payload

pkg/handlers/ghcapi/move_task_order_test.go
```diff
@@ -265,69 +260,114 @@ func (suite *HandlerSuite) TestUpdateMTOStatusServiceCounselingCompletedHandler(
...
+			handler := UpdateMTOStatusServiceCounselingCompletedHandlerFunc{
+				context,
+				&mockUpdater,
+			}
+
+			response := handler.Handle(params)
+			suite.IsNotErrResponse(response)
+			suite.IsType(testCase.handlerResponse, response)
+		}
 	})
```
duncan-truss, line 371: We don't have any authorization that checks that it's the office user with the Service Counseling role making this request do we?

duncan-truss (approved):

pkg/handlers/ghcapi/move_task_order_test.go
reggieriser: Added the call to validate the response.

pkg/handlers/ghcapi/move_task_order_test.go
reggieriser: I'll try to add this.  I see that was a TODO from the existing endpoint.

pkg/services/move_task_order/move_task_order_fetcher.go
```diff
@@ -133,6 +136,11 @@ func (f moveTaskOrderFetcher) FetchMoveTaskOrder(appCtx appcontext.AppContext, s
 		}
 		mto.MTOShipments[i].Reweigh = reweigh
 
+		loadErr := appCtx.DB().Load(&mto.MTOShipments[i], "PPMShipment")
```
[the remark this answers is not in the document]
reggieriser: I'll go ahead and add a check for a `PPM` shipment type since that should be a good indicator that there should be a `PPMShipment` record.  But I'm not sure I know enough about which shipment types may or may not have a reweigh to safely optimize -- if that's a well-known list, I can add that too.

pkg/handlers/ghcapi/move_task_order_test.go
reggieriser: Added the SC check and modified tests.

duncan-truss (approved):

PR #8438, opened 2022-04-14
MB-11600 Block incomplete PPMs on frontend

pearl-truss:
## Summary

- This PR blocks submission on the move/home page if any ppm shipment is incomplete.
- I removed legacy `current_ppm` logic on this page and in the unit tests

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the 
2. Login as a
3.

### Frontend

## Screenshots
Screen Shot 2022-04-22 at 10 27 59 AM
Screen Shot 2022-04-22 at 10 29 44 AM


src/utils/shipments.js
```diff
@@ -2,7 +2,10 @@
 // information and get to the review screen in an incomplete state from creating another shipment
 // on the move.  Advance requested is the last required field that would mean they're finished.
 export function isPPMShipmentComplete(mtoShipment) {
-  return mtoShipment?.ppmShipment?.advanceRequested != null;
+  if (mtoShipment?.ppmShipment?.advanceRequested === null || mtoShipment?.ppmShipment?.advanceRequested === undefined) {
```
duncan-truss, line 5: The reason I used `advanceRequested != null` is because it gives the same result for undefined as well as null because it doesn't look at type equality the way `!==` or `===` do.

Screen Shot 2022-04-14 at 5 47 03 PM

felipe-lee: **thought**: I think being explicit how this was changed to might be better though so that future folks don't think it was an accident. Either that or adding a comment about it.

felipe-lee:
>     * I noticed that `mto_shipment.ppmShipment.advancedRequested` isn't being automatically set to `null` on the front and instead that field is missing completely. Our existing `isPPMComplete` utility function was checking for a null value and since the field wasn't present (and not set to null) that check was failing. I've added an additional check in `isPPMComplete` to check for `undefined` as a temp fix and will look into why `advanceRequested` isn't being set to null

**thought**: I think it's because we didn't set `x-omitempty` to `false` for `advance` and `advanceRequested`: https://github.com/transcom/mymove/blob/e24b332827774fbe82e51d28b5b5d12999d5b796/swagger-def/internal.yaml#L2540-L2550

**suggestion (non-blocking)**: Since we do set it for other fields on the `PPMShipment` and it's affecting this PR, can you update it as part of this PR?

src/components/ShipmentList/ShipmentList.jsx
```diff
@@ -136,6 +143,11 @@ const ShipmentList = ({ shipments, onShipmentClick, onDeleteClick, moveSubmitted
               }
           }
         }
+        if (shipmentType === SHIPMENT_OPTIONS.PPM) {
+          if (!isPPMShipmentComplete(shipment)) {
+            isIncomplete = true;
+          }
+        }
```
felipe-lee, line 150: **nitpick**: Could you combine these and make it a `const`? 

**note**: Making it a `const` would require a change above to remove the `let isIncomplete` declaration so this suggestion can't be used standalone, sorry 😞 

```suggestion
        const isIncomplete = shipmentType === SHIPMENT_OPTIONS.PPM && !isPPMShipmentComplete(shipment)
```

src/utils/shipments.js
felipe-lee: **suggestion**: Given the change I suggested in another comment to the `internal.yaml` spec, I think we no longer have to check for undefined in this function. So I think we can change this to be 

```js
return mtoShipment?.ppmShipment?.advanceRequested !== null
```

src/components/ShipmentList/ShipmentList.stories.jsx
```diff
@@ -4,6 +4,85 @@ import ShipmentList from './ShipmentList';
...
+    <ShipmentList
+      shipments={[
+        { id: '0001', shipmentType: SHIPMENT_OPTIONS.HHG },
+        { id: '0002', shipmentType: SHIPMENT_OPTIONS.NTS },
+        { id: '0003', shipmentType: SHIPMENT_OPTIONS.PPM },
+      ]}
+      onShipmentClick={() => {}}
+    />
+  </div>
+);
```
felipe-lee, line 22: **issue**: There's two console errors with each of these new stories (only going to comment on this one to save space and comments). One is that the `moveSubmitted` prop is missing from the `ShipmentList` props. The other is that it says it can't find the story with the given ID, though it is loading the page so I'm not sure what the error means tbh.

image

felipe-lee: Huh...I'm getting that second error with every storybook component we have (or at least the ones I tested out that aren't part of this PR). So I guess it's mainly the missing prop that needs to be addressed in this PR. The other error I'm not sure about.

src/pages/MyMove/Home/index.jsx
```diff
@@ -110,22 +110,27 @@ export class Home extends Component {
...
 
-  get hasPPMShipment() {
-    const { currentPpm } = this.props;
-    return !!Object.keys(currentPpm).length;
+  get hasPPMShipments() {
+    const { mtoShipments } = this.props;
+    return mtoShipments?.some((shipment) => shipment.ppmShipment);
+  }
+
+  get hasCompletedPPMShipments() {
```
felipe-lee, line 127: **nitpick**: Could the name be something like `hasCompletedAllPPMShipments`? Mainly because from the name, I could see it meaning all, or as long as there's one PPM Shipment completed it could be true. But the code is functionally doing the first.

src/pages/MyMove/Home/index.jsx
```diff
@@ -503,10 +492,6 @@ Home.propTypes = {
     last_name: string,
   }),
   mtoShipments: arrayOf(MtoShipmentShape).isRequired,
-  currentPpm: shape({
-    id: string,
-    shipmentType: string,
-  }).isRequired,
```
felipe-lee, line 509: **praise**: thank you for cleaning up the old PPM code in this page!

src/components/ShipmentList/ShipmentList.stories.jsx
ronaktruss: This file has been modified quite a bit on master as part of https://github.com/transcom/mymove/pull/8375. That PR resolved the first issue and modified the existing story names.

src/utils/shipments.js
duncan-truss: I'm fine with being more explicit about it 👍 We should be comfortable dealing with undefined on the frontend though, in React props don't always resolve with values immediately and destructuring objects results in undefined values not only null.  By using the optional chaining you are also indicating that the value can be undefined not just null so it's a bit of a mixed message.

duncan-truss: I'm fine with being more explicit about it 👍 We should be comfortable dealing with undefined on the frontend though, in React props don't always resolve with values immediately and destructuring objects results in undefined values not only null.  By using the optional chaining you are also indicating that the value can be undefined not just null so it's a bit of a mixed message.

src/components/ShipmentList/ShipmentList.stories.jsx
```diff
@@ -4,6 +4,85 @@ import ShipmentList from './ShipmentList';
...
+          primeEstimatedWeight: 300,
+          reweigh: { id: '1236', weight: 200 },
+        },
+      ]}
+      showShipmentWeight
+    />
+  </div>
+);
+
+export const ShipmentListWithPPMs = () => (
```
duncan-truss, line 60: Storybook mangled this name
Screen Shot 2022-04-21 at 9 08 59 AM

duncan-truss: It looks like you can override the individual story name this way https://storybook.js.org/docs/react/writing-stories/introduction#rename-stories

src/utils/shipments.js
ronaktruss: Nit (suggestion/non-blocking): Personally prefer the original implementation over the current one.

ronaktruss (approved):
LGTM, outside of Duncan's comment regarding the storybook name.

scarequotes (approved):
Looks good to me!

scarequotes (changes requested):
Screen Shot 2022-04-22 at 1 37 52 PM

This needs tweaking on the mobile view — the incomplete label should not move the delete and edit links off the screen. Label can move left, closer to PPM.

src/pages/MyMove/Home/index.test.jsx
```diff
@@ -149,6 +174,64 @@ describe('Home component', () => {
...
+
+    it('displays incomplete for an incomplete PPM', () => {
+      expect(wrapper.find('ShipmentListItem').at(0).text()).toContain('Incomplete');
+    });
+
+    it('disables the review and submit button', () => {
+      expect(wrapper.find('button[data-testid="review-and-submit-btn"]').prop('disabled')).toBe(true);
+    });
+  });
+
```
felipe-lee, line 234: **thought**: We shouldn't be adding new tests using enzyme

**suggestion (non-blocking)**: Can you update these new tests to use RTL instead? I don't think you should have to update all the tests in the file to use RTL, but I'd rather make any new tests use RTL. Less work in the future if we do switch the rest of the tests.

src/utils/shipments.js
felipe-lee: **thought**: I agree with Ronak, if we're sticking with the check we were doing originally, then I prefer the simpler logic. I mainly meant being explicit about the null vs undefined, but if everyone prefers the `!=`, then I like the shorter version more.

felipe-lee (approved):
LGTM. I added a comment on the new Home page tests, but it's a non-blocking suggestion.

pearl-truss:
> Screen Shot 2022-04-22 at 1 37 52 PM
> 
> This needs tweaking on the mobile view — the incomplete label should not move the delete and edit links off the screen. Label can move left, closer to PPM.

@ronaktruss is this fixed in your PR? otherwise I can address it here

src/pages/MyMove/Home/index.test.jsx
felipe-lee: 🎉 thanks for making this testing change!

swagger/internal.yaml
```diff
@@ -4727,6 +4729,29 @@ paths:
...
+          format: uuid
+          required: true
+          description: UUID of the move
+      responses:
+        '200':
+          description: weight estimate is below allotted entitlement
+        '404':
+          description: personally procured move not found
+        '409':
+          description: Requested weight estimate is above allotted entitlement
```
felipe-lee, line 4754: **question**: I'm not sure if this started showing up because of merging master or something, but it's not meant to be part of your changes right?

pearl-truss: yeah it came after the latest merge from master. I'll look into why

pearl-truss: okay this is resolved, I had to run `make swagger_generate` after the merge

felipe-lee: Nice!

scarequotes (approved):
Thanks for fixing the alignment!

PR #8445, opened 2022-04-15
[MB-12157] Remove deleted shipments from internal handlers

duncan-truss:
## Summary

This PR fixes the FetchMove query used by the customer app to exlcude deleted shipments from being returned and showing up on the review page.

I've introduced our first use of query scopes, with the introduction of the `ExcludeDeletedScope`.  This can be tacked onto any database query and will take an optional list of model names to append as a where clause filtering out deleted columns.  This calls the TableName method for models that override the default snake case pluralized name.

I've gone through and converted some queries that I could to use this method.  It's not a perfect solution in all cases especially when you're fetching another model and loading the association that can be deleted.  You can't apply a filter to just the eager association only the join association itself. In a lot of cases we don't want a move to disappear from the queue entirely or from the Prime's view because the last shipment was deleted either.

[GoBuffalo Pop Scopes documenation](https://gobuffalo.io/documentation/database/scoping/)

One other thing I did explore was using some of Pop's [Callbacks](https://gobuffalo.io/documentation/database/callbacks/) like AfterFind(), but I don't think it would work for filtering out deleted models or eager associations.  You have to define AfterFind on each model as well.

## Setup to Run Your Code

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
```

### Additional steps

To test that the deleted shipments no longer appear on the review page after adding an additional shipment or the homepage after submitting the move
1. Login as the full@ppm.unsubmitted user
2. Delete the existing PPM shipment either from the home screen or from the review page
3. Proceed to add another shipment of any type.
4. After shipment creation you should be brought to the review page and none of the deleted shipments should be displayed.
5. Proceed to sign and submit the move, you should be brought back to the homepage without any deleted shipments being shown.

### Frontend

### Backend

## Screenshots

The API query that was returning the deleted shipments was the PATCH moves/{id} query that sets the selected move type after you make the new shipment type selection.  You can verify after deleting shipments that this is not the case anymore.

Screen Shot 2022-04-20 at 11 06 26 AM
Screen Shot 2022-04-20 at 11 06 49 AM


pkg/handlers/internalapi/entitlements.go
```diff
@@ -49,64 +44,3 @@ func (h IndexEntitlementsHandler) Handle(params entitlementop.IndexEntitlementsP
...
 		})
 }
-
-// ValidateEntitlementHandler validates a weight estimate based on entitlement for a PPM move
-type ValidateEntitlementHandler struct {
-	handlers.HandlerContext
-}
-
-// Handle is the handler
-func (h ValidateEntitlementHandler) Handle(params entitlementop.ValidateEntitlementParams) middleware.Responder {
```
duncan-truss, line 59: I think we removed the last instance of this call on the customer Review move page.  It would show a warning if the ppm estimated weight was over the weight allowance but now tests were breaking because I removed loading the old PersonallyProcuredMove association from a query.

pkg/handlers/internalapi/service_members.go
```diff
@@ -320,12 +320,7 @@ type ShowServiceMemberOrdersHandler struct {
 func (h ShowServiceMemberOrdersHandler) Handle(params servicememberop.ShowServiceMemberOrdersParams) middleware.Responder {
 	return h.AuditableAppContextFromRequestWithErrors(params.HTTPRequest,
 		func(appCtx appcontext.AppContext) (middleware.Responder, error) {
-			serviceMember, err := models.FetchServiceMemberForUser(appCtx.DB(), appCtx.Session(), appCtx.Session().ServiceMemberID)
```
duncan-truss, line 323: We weren't doing anything with the service member query that loaded a bunch of unnecessary information so we can simplify to just fetching the orders with the session's service member.

pkg/models/move.go
```diff
@@ -160,7 +160,8 @@ func FetchMove(db *pop.Connection, session *auth.Session, id uuid.UUID) (*Move,
 		"Orders.ServiceMember",
 		"Orders.UploadedAmendedOrders",
 		"MoveDocuments.Document",
-	).Where("show = TRUE").Find(&move, id)
+	).LeftJoin("mto_shipments", "mto_shipments.id = moves.id AND mto_shipments.deleted_at IS NULL").
```
duncan-truss, line 163: This was the query that would update the Redux store with deleted shipments if you went back and added a new shipment after the fact.

We can't use the new scope I created here because we'll get a 404 not found if no shipments exist yet for the move.

pkg/db/utilities/utilities.go
```diff
@@ -129,3 +131,23 @@ func GetHasManyForeignKeyAssociations(model interface{}) []interface{} {
...
+		for _, model := range models {
+			var tableName string
+			// Some models override the table name instead of the snake case pluralizing of the model name
+			if modelTable, ok := model.(pop.TableNameAble); ok {
+				tableName = modelTable.TableName()
+			} else {
+				modelType := reflect.TypeOf(model)
+				tableName = flect.Pluralize(flect.Underscore(modelType.Name()))
+			}
+			q.Where(fmt.Sprintf("%s.deleted_at IS NULL", tableName))
```
duncan-truss, line 149: One issue not accounted for here is if a table name is aliased in one of the join statements.  We could create a separate method for that if we wanted, I don't think we can get access to that because scopes may be defined first in the query chain.

I didn't know if it was a step to far to actually check that the model has a field deletedAt present.  There's not a great was to error out here, I think you'll get a SQL query error anyway so I don't think it will go silently missed.  The soft destroy code does this already.

pkg/db/utilities/utilities.go
```diff
@@ -129,3 +131,23 @@ func GetHasManyForeignKeyAssociations(model interface{}) []interface{} {
 	}
 	return hasManyForeignKeyAssociations
 }
+
+func ExcludeDeletedScope(models ...interface{}) pop.ScopeFunc {
```
duncan-truss, line 135: I was going to put this in the pkg/services/query package but that creates a cyclical import error in go because of queries in the models and query builder importing models 👎 Happy to move it out of utilities though.

migrations/app/schema/20220415162556_create_indexes_for_mto_shipments_deleted_at.up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX ON mto_shipments (deleted_at);
```
duncan-truss, line 1: For the most part we're usually filtering the shipments already by the move id so this may not make a huge difference.

pkg/handlers/internalapi/personally_procured_move_test.go
```diff
@@ -248,76 +248,6 @@ func (suite *HandlerSuite) TestSubmitPPMHandler() {
 	suite.Require().Equal(submitPPMPayload.Status, internalmessages.PPMStatusSUBMITTED, "PPM should have been submitted")
 }
 
-func (suite *HandlerSuite) TestIndexPPMHandler() {
```
duncan-truss, line 251: I think we removed the last call of this on the customer side prior to submitting a move.  This test was failing because I removed the PPM association from a fetch move query.

pkg/models/move.go
```diff
@@ -170,6 +166,20 @@ func FetchMove(db *pop.Connection, session *auth.Session, id uuid.UUID) (*Move,
 		return nil, err
 	}
 
+	var shipments MTOShipments
```
duncan-truss, line 169: I broke these queries up because we don't want to return a not found error if the move doesn't have any shipments or are all deleted.

pkg/models/order.go
```diff
@@ -178,20 +180,16 @@ func FetchOrderForUser(db *pop.Connection, session *auth.Session, id uuid.UUID)
 	}
 
 	// Eager loading of nested has_many associations is broken
-	err = db.Load(&order.UploadedOrders, "UserUploads.Upload")
+	var userUploads UserUploads
```
duncan-truss, line 183: Added in the scope to exclude deleted orders uploads

pkg/models/service_member.go
```diff
@@ -348,37 +350,28 @@ func (s ServiceMember) FetchLatestOrder(session *auth.Session, db *pop.Connectio
...
+		All(&userUploads)
 	if err != nil {
 		return Order{}, err
 	}
 
-	// Only return user uploads that haven't been deleted
-	userUploads := order.UploadedOrders.UserUploads
-	relevantUploads := make([]UserUpload, 0, len(userUploads))
-	for _, userUpload := range userUploads {
-		if userUpload.DeletedAt == nil {
```
duncan-truss, line 360: Here we were manually looping through the records to filter out the deleted user uploads

pkg/services/move_task_order/move_task_order_fetcher.go
```diff
@@ -122,8 +122,8 @@ func (f moveTaskOrderFetcher) FetchMoveTaskOrder(appCtx appcontext.AppContext, s
 		filteredShipments = models.MTOShipments{}
 	}
 	for i, shipment := range mto.MTOShipments {
-		// Skip any shipments that use an external vendor (if requested)
-		if searchParams.ExcludeExternalShipments && shipment.UsesExternalVendor {
+		// Skip any shipments that are deleted or use an external vendor (if requested)
+		if shipment.DeletedAt != nil || (searchParams.ExcludeExternalShipments && shipment.UsesExternalVendor) {
```
duncan-truss, line 126: This will filter out deleted shipments from the prime that were previously being returned.  This query is a bit too involved for me to refactor it here.

pkg/services/mto_shipment/shipment_cancellation_requester.go
```diff
@@ -54,7 +56,7 @@ func (f *shipmentCancellationRequester) RequestShipmentCancellation(appCtx appco
 
 func (f *shipmentCancellationRequester) findShipment(appCtx appcontext.AppContext, shipmentID uuid.UUID) (*models.MTOShipment, error) {
 	var shipment models.MTOShipment
-	err := appCtx.DB().Q().Find(&shipment, shipmentID)
+	err := appCtx.DB().Q().Scope(utilities.ExcludeDeletedScope()).Find(&shipment, shipmentID)
```
duncan-truss, line 59: I might combine some of these findShipment methods that we've duplicated across service objects in the same package.

duncan-truss: I ended up going through with this refactor, I added a package level method FindShipment to the mto_shipment_fetcher that can be reused.  I didn't add it as part of the fetcher interface because then all other service objects would need an instance of mto shipment fetcher.  A package level method can be harder to mock but I don't think we're normally mocking our database calls in our service tests.

migrations/app/schema/20220415162556_create_indexes_for_mto_shipments_deleted_at.up.sql
reggieriser: I do wonder how much this helps and if it's worth adding or not at this point.  I imagine that we're usually only looking for records where `deleted_at` is null which would be the majority of the records (I'm making an assumption deletes don't happen regularly).  This [StackOverflow post](https://stackoverflow.com/questions/31966218/postgresql-create-an-index-to-quickly-distinguish-null-from-non-null-values) touches on this.  I don't think it's a huge deal either way (we should be able to add/drop the index later if needed), but just thought I'd mention it.

reggieriser (approved):
Reviewed the migration (only) and it ran fine.  I added a comment about whether the index does much for us, but it's not a blocker.  I haven't looked at other parts of the PR, so we'll need some 👀 on that.

migrations/app/schema/20220415162556_create_indexes_for_mto_shipments_deleted_at.up.sql
duncan-truss: Agreed, I didn't think it warranted a partial index either because of the distribution you talked about and how we aren't querying FOR deleted shipments.

pkg/db/utilities/utilities.go
ronaktruss: > One issue not accounted for here is if a table name is aliased in one of the join statements.

Can you clarify what you mean by "table name is aliased". I assume this is referencing if an entity used in a join has a pop tableName alias, as opposed to an alias for a table that is joined when using a raw query.

ronaktruss:
How does `ExcludeDeletedScope` behave with raw queries? Can it be used?

duncan-truss:
> How does `ExcludeDeletedScope` behave with raw queries? Can it be used?

@ronaktruss I didn't try with RawQuery() but I'll write a test and find out.  I know the Scope will need to come after the RawQuery at least because RawQuery operates off of a connection object instead of a chained query.  But RawQuery returns a query object so it's possible it could be used with it let me see.  If not that'll be good to mention in the usage too.  Did you see a RawQuery that was filtering out soft deleted records?

ronaktruss (approved):
LGTM. The previously found issues are now resolved. Thanks for also cleaning up some code in the process. 

It may be nice to spend some time scheduling a short demo going over how this is used and some of the limitations to a broader audience so folks are aware (eg. All together/ BE checkin).

pkg/db/utilities/utilities.go
```diff
@@ -129,3 +131,35 @@ func GetHasManyForeignKeyAssociations(model interface{}) []interface{} {
...
+// If you are filtering the same model that is being returned you can call this with no arguments and it should give
+// you the desired behavior: db.Scope(utilities.ExcludedDeletedScope()).All(&mtoShipments).
+//
+// If you are filtering on a join table(s) or need to disambiguate multiple tables with deleted_at columns, then you
+// will need to pass in the model(s) so we can derive the real table names:
+// db.Scope(utilities.ExcludeDeletedScope(models.MTOShipment{})).
+//     Join("mto_shipments", "mto_shipments.move_id = moves.id").
+//     All(&moves)
+//
+// You won't be able to use this if you have given your table name(s) an alias, so just fall back to a normal where.
```
ronaktruss, line 145: Is this referencing overriding a models default TableName or giving a table name an alias?

ronaktruss:
> @ronaktruss I didn't try with RawQuery() but I'll write a test and find out. I know the Scope will need to come after the RawQuery at least because RawQuery operates off of a connection object instead of a chained query. But RawQuery returns a query object so it's possible it could be used with it, let me see. If not that'll be good to mention in the usage too. Did you see a RawQuery that was filtering out soft deleted records?

Nope, nothing in specific.

pkg/db/utilities/utilities.go
duncan-truss: This is in reference to the latter, it doesn't have to be a raw query it could just be that you used an alias in your join clause like `Q().Join("mto_shipments ms", "ms.move_id = moves.id").All(&moves)`

duncan-truss:
> 

Raw queries don't seem to be compatible to use with Scopes/Joins/Where/Eager features.  Only limit and pagination seem to be possible to combine with RawQuery from the looks of it https://github.com/gobuffalo/pop/blob/c285511d17469aa430ab08202a6f2353c3ae19e5/sql_builder.go#L77-L85

PR #8449, opened 2022-04-18
MB-12186 remove access code

leia-sefkin:
## Summary

The original intent of this work was to refactor validation patterns, but it became apparent Access Code is a deprecated service. So this work removes it and references from the code base. To help in reviewing the changes structured the commits into sections. 

## Setup to Run Your Code

`make db_dev_reset db_dev_migrate`
`make server_run`
`make client_run`

There's not a lot to check here, as it's a full removal. To be safe create and manage a service member from the office app. Also a second check to make sure I didn't miss any references in the code base is welcome. 


cmd/milmove/serve.go
```diff
@@ -132,9 +132,6 @@ func initServeFlags(flag *pflag.FlagSet) {
 	// Logging
 	cli.InitLoggingFlags(flag)
 
-	// Feature Flags
-	cli.InitFeatureFlags(flag)
```
leia-sefkin, line 136: "Feature Flags" as we had implemented was only in use for access code, so it seemed prudent to remove as well for now.

pkg/services/query/query_builder_test.go
```diff
@@ -685,73 +685,3 @@ func (suite *QueryBuilderSuite) TestFetchCategoricalCountsFromOneModel() {
 
 	})
 }
-
```
leia-sefkin, line 688: I removed these tests rather than refactor - as this is a legacy service anyway

PDickeyTruss (approved):
LGTM, I see no change in the app with everything working as before. I played around quite a bit in the office app and saw no issues. Appreciate the specific/granular commits, made it quite easy to follow what was being deleted. 


Non-blocking but I was able to find 'access_code' mentioned in just a few more places that may be worth cleaning up:
* `src/constants/routes.js` has an access_code route
* Docker compose and env files have `FEATURE_FLAG_ACCESS_CODE=false`

Ryan-Koch (approved):

cosburn (approved):

PR #8459, opened 2022-04-19
[MB-12077] Make ServiceItem fields that are unused in responses optional

mboren:
## [Jira ticket](https://dp3.atlassian.net/browse/MB-12077) for this change
These API changes will unblock writing load tests for payment requests.

These service item fields were not always being used in responses, even though they were marked as required in the swagger definition.
The Go app doesn't check response types against their swagger definitions, so inconsistencies like this only come to light when you're working on load tests, which do check response types against the spec.

An alternative way to go about this would be to create another service item type that is just used for responses, instead of having the definition shared. I didn't do this because it seems like we generally tend to err on the side of sharing code vs having super tight definitions.

## Setup to Run Your Code

<details>

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps
I think this should be pretty well covered by automated tests.
To exercise the affected API requests pretty thoroughly you can:
1. create a customer
2. approve as TOO
3. Create additional service items with the prime simulator
4. Approve those service items as a TOO

### DIFF
--- a/pkg/gen/ghcapi/embedded_spec.go
+++ b/pkg/gen/ghcapi/embedded_spec.go
@@ -3984,11 +3984,7 @@ func init() {
         "moveTaskOrderID",
         "reServiceID",
         "reServiceCode",
-        "reServiceName",
-        "mtoShipmentID",
-        "reason",
-        "pickupPostalCode",
-        "description"
+        "reServiceName"
       ],
       "properties": {
         "SITPostalCode": {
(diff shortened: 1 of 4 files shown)


PDickeyTruss (approved):
LGTM! All works for me locally and code changes look good.

PR #8458, opened 2022-04-19
MB-12245 merge formatters files methods that need extra work

YanZ777:
## Summary

Migrating the rest of the functions from `shared/formatters` over into `utils/formatters`. 
All of these functions I've had to adjust in some manner or another. There are comments in this PR that specifically call out what has been changed.
Any imports have been adjusted to now point to `utils/formatters`
I did not migrate `formatCents` into `utils/formatters` because there was an existing `formatCents` inside of `utils/formatters`. The `formatCenters` in `shared/formatters` will be deleted in a separate PR.
`shared/formatters.test.jsx` has been deleted since there are no longer tests in it, those have all been carried over into `utils/formatters.test.jsx`

## Setup to Run Your Code

<details>
<summary>💻 You will need to use three separate terminals to test this locally.</summary>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

Login to all the apps and make sure everything still works as expected.

## Verification Steps for Author

These are to be checked by the author.

- [x] Request review from a member of a different team.
- [ ] Have the Jira acceptance criteria been met for this change?

## Verification Steps for Reviewers

These are to be checked by a reviewer.

PR #8460, opened 2022-04-19
MB-12245 merge formatters files remove unused functions

YanZ777:
## Summary

This PR should not be merged until [this PR](https://github.com/transcom/mymove/pull/8458) is merged first.

All these functions were not being used in the repo. 
The exception to this was `formatCents`. 
I opted to use the `formatCents` inside of `utils/formatters`, which appears to be functionally the same. There is one check missing which I've left out, but let me know if you think it should be put back in.
All the imports for `formatCents` now point to the `formatCents` inside of `utils/formatters`.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

Login to all the applications and make sure everything works as expected 


src/utils/formatters.js
```diff
@@ -454,11 +299,6 @@ export function formatCents(cents, minimumFractionDigits = 2, maximumFractionDig
   return (cents / 100).toLocaleString(undefined, { minimumFractionDigits, maximumFractionDigits });
```
YanZ777, line 299: The difference between this `formatCents` and the one inside `shared/formatters` is that the one inside of `ahared/formatters` has the following code above this line. I am not certain if I should put that in this function or not.

> if (!isFinite(cents)) {
    return '';
  }

michaelvaldes-truss (approved):
lgtm!

donaldthai (approved):
🧹 🚀

src/utils/formatters.js
ronaktruss: Internally this is doing https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L11626. This effectively makes the function return an empty string in the instance this function is called without an appropriate input. 

When we don't have this, we likely get `NaN` when we have an invalid input. 

I don't think we need to include it. ~~That said, from a consistency standpoint it might be better to add the following to the the `formatCents` function.~~ (I see an instance where we are returning null instead of an empty string when it fails this check.)
```suggestion
  if (!Number.isFinite(cents)) {
    return '';
  }
  return (cents / 100).toLocaleString(undefined, { minimumFractionDigits, maximumFractionDigits });
```

ronaktruss: If time permits I'd advocate for looking into why that one function (`convertFromThousandthInchToInch`) returns null and see if we can modify this without breakages and adding the suggestion above.

In general it looks like most of these function (those that are using the Number.isFinite) are called on data from the DB so they probably never go into this if scope.

YanZ777: I've created a new issue based out of this discussion.
[https://dp3.atlassian.net/browse/MB-12306](https://dp3.atlassian.net/browse/MB-12306)

PR #8461, opened 2022-04-20
Copy in the js, not mjs for js-cookie to fix swagger-ui

ahobson:
Tiny little change to try and fix swagger-ui

### DIFF
--- a/public/swagger-ui/admin.html
+++ b/public/swagger-ui/admin.html
@@ -69,7 +69,7 @@
 
 <script src="/swagger-ui/swagger-ui-bundle.js"> </script>
 <script src="/swagger-ui/swagger-ui-standalone-preset.js"> </script>
-<script src="/swagger-ui/dist/js.cookie.mjs"></script>
+<script src="/swagger-ui/dist/js.cookie.js"></script>
 <script>
     window.onload = function() {
         // Intercept the request to add csrf token and only allow requests to the current domain.
--- a/public/swagger-ui/admin.html
+++ b/public/swagger-ui/admin.html
@@ -69,7 +69,7 @@
 
 <script src="/swagger-ui/swagger-ui-bundle.js"> </script>
 <script src="/swagger-ui/swagger-ui-standalone-preset.js"> </script>
-<script src="/swagger-ui/dist/js.cookie.js"></script>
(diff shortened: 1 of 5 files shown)


scripts/copy-swagger-ui
```diff
@@ -5,4 +5,4 @@
 # will need to be manually updated based on node_modules/swagger-ui-dist/index.html
 # if it ever changes.
 cp node_modules/swagger-ui-dist/{*.js,*.css,*.png} public/swagger-ui
-cp node_modules/js-cookie/dist/js.cookie.mjs public/swagger-ui
```
donaldthai, line 8: suggestion: Could we also update the js.cookie.js path of the other html files (internal.html, admin.html, api.html and ghc.html) in the `public/swagger-ui` directory. Noticed that the other html files are using the wrong path for that cookie library. Should be using `<script src="/swagger-ui/js.cookie.js"></script>`, not `<script src="/swagger-ui/dist/js.cookie.js"></script>`.

ahobson: Good catch

public/swagger-ui/admin.html
```diff
@@ -69,7 +69,7 @@
 
 <script src="/swagger-ui/swagger-ui-bundle.js"> </script>
 <script src="/swagger-ui/swagger-ui-standalone-preset.js"> </script>
-<script src="/swagger-ui/dist/js.cookie.mjs"></script>
+<script src="/swagger-ui/dist/js.cookie.js"></script>
```
donaldthai, line 72: I think the path should be `/swagger-ui/js.cookie.js` which is also the same path in `prime.html`

ahobson: 🤦 Thank you, fixed

shimonacarvalho:
What does this fix? It sounds like maybe this takes care of an error in the swagger ui? 
Which url, what is the expected error/change in behavior?
That will help me confirm the fix.

donaldthai:
> 

When you're running the app, you can access the swagger UI page for each API. For example, swagger UI page for the GHC can be access by the path `http://officelocal:3000/ghc/v1/docs`.

donaldthai (approved):
👍

PR #8467, opened 2022-04-21
Moved context to array of objects

kennethchow-truss:
## Summary

Because of two different types of context data objects being used, it was causing an issue for the payment_requests ticket. These changes aim to fix it

## How to run the code
1. Go to officelocal:3000/
2. Login as a user 
3. Choose a move with service items to approve
4. Approve them 
5. Make sure the details match the image below

image

### DIFF
--- a/pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
+++ b/pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
@@ -88,8 +88,8 @@ func MoveAuditHistory(auditHistory models.AuditHistory) *ghcmessages.MoveAuditHi
 		ActionTstampClk:      strfmt.DateTime(auditHistory.ActionTstampClk),
 		ActionTstampStm:      strfmt.DateTime(auditHistory.ActionTstampStm),
 		ActionTstampTx:       strfmt.DateTime(auditHistory.ActionTstampTx),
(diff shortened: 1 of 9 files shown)


pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -123,6 +123,17 @@ func removeEscapeJSON(data *string) map[string]string {
 
 }
 
+func removeEscapeJSONtoArray(data *string) []map[string]string {
```
michaelvaldes-truss, line 126: **suggestion** (non-blocking): Is it worth using generics here? Are we supposed to be using generics yet?

michaelvaldes-truss (approved):
lgtm!

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
kennethchow-truss: You mean the go function return to be generic? I think Drew mentioned that was anti-pattern for Go. So for now, we use strings

michaelvaldes-truss: I was thinking more like using [go generics](https://go.dev/doc/tutorial/generics) to reduce the repeated code between the two 'removeEscapeJSON' methods, but it'd probably only save us like two lines anyway.

PR #8468, opened 2022-04-21
[MB-11863] Allow TOO to see moves with PPM shipments in their queue

hannaliebl:
## Summary

- This creates a migration to drop and then recreate the `move_to_gbloc` view to get GBLOC info from moves with PPM shipments (@reggieriser basically wrote the SQL, so thank you to him for helping me out! He linked me to [this article ](https://weblogs.sqlteam.com/jeffs/2007/10/11/mixing-inner-outer-joins-sql/)which might help with understanding the SQL.) PPM shipments don't contain address information like most other shipment types, so the SQL had to be updated to accomodate looking just at the `pickup_postal_code` column.
- The query now returns null values for some GLBOC values in moves that either don't have shipments or don't have address or postal code info (like NTS-release shipments.) I had to update the handlers/Go model to account for this column now potentially having a null value. The specific NTS-release case for queues is handled elsewhere, so this should be ok. But certainly something to look at closely if I missed a usecase for this view that will be upset now that the GBLOC column can be null.

NOTE: I still need to add some tests, but I thought getting this up to look at earlier would be helpful.

## Setup to Run Your Code
1. Run the migration locally `make db_dev_reset db_dev_migrate`
2. I reset the local data as well using `make db_dev_e2e_populate`
3. On the customer app, use `ppm.test.user1@example.com` to setup a PPM shipment. I used 90210 as the origin and destination ZIPs to keep it all in the KKFA GBLOC.
4. It shouldn't matter what other info you fill in, but go through and save the shipment and then submit the move request, noting the move id (it should be `NTS001`).
5. Go to the office app and login as the TOO. Find the `NTS001` move code in the queue and view it.
6. You won't yet see PPM info on the move details page, but the mto_shipments network request should show one move with the shipmentType set to "PPM."

### Frontend

### Backend


migrations/app/schema/20220420195400_update_move_to_gbloc_view_for_ppm_shipments.up.sql
```diff
@@ -0,0 +1,23 @@
+DROP VIEW move_to_gbloc;
```
hannaliebl, line 1: I *think* this is the only way to successfully get this view (which already exists) to be "updated"...because some types of views are not updatable according to [the Postgres docs](https://www.postgresql.org/docs/current/sql-createview.html) I found. This is outside my normal wheelhouse, though, so let me know if this is wrong!

LeDeep (approved):
looks good. migrations ran successfully.

migrations/app/schema/20220420195400_update_move_to_gbloc_view_for_ppm_shipments.up.sql
```diff
@@ -0,0 +1,23 @@
...
+        JOIN postal_code_to_gblocs pctg ON a.postal_code = pctg.postal_code
+     ) pctg ON pctg.address_id = sh.pickup_address_id
+     -- try the ppm_shipments path
+     LEFT JOIN
+     (
+        SELECT ppm.shipment_id, pctg.gbloc
+        FROM ppm_shipments ppm
+        JOIN postal_code_to_gblocs pctg ON ppm.pickup_postal_code = pctg.postal_code
+     ) pctg_ppm ON pctg_ppm.shipment_id = sh.id
+ORDER BY sh.move_id, sh.created_at;
```
reggieriser, line 23: I'm not sure if there are any performance issues with the way we structured this query.  Interested if any other reviewers have thoughts.

pkg/models/move_to_gblocs.go
```diff
@@ -8,7 +8,7 @@ import "github.com/gofrs/uuid"
 type MoveToGBLOC struct {
 	MoveID uuid.UUID `db:"move_id"`
 	Move   Move      `belongs_to:"moves" fk_id:"move_id"`
-	GBLOC  string    `db:"gbloc"`
+	GBLOC  *string   `db:"gbloc"`
```
reggieriser, line 11: Good catch in making this nullable to match the reality coming out of the view.

reggieriser (commented):
I added a couple of comments about some minor issues, but otherwise looks good.  Since I paired on the new view query, I'd prefer to punt final approval to another DB reviewer.

migrations/app/schema/20220420195400_update_move_to_gbloc_view_for_ppm_shipments.up.sql
hannaliebl: I think you are right, I did this here https://github.com/transcom/mymove/pull/8468/commits/ca66d56638aca039802cc221be132bafc1cdd11e and it seemed to work fine. Thanks!

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -926,7 +928,7 @@ func QueuePaymentRequests(paymentRequests *models.PaymentRequests) *ghcmessages.
 			Age:                math.Ceil(time.Since(paymentRequest.CreatedAt).Hours() / 24.0),
 			SubmittedAt:        *handlers.FmtDateTime(paymentRequest.CreatedAt), // RequestedAt does not seem to be populated
 			Locator:            moveTaskOrder.Locator,
-			OriginGBLOC:        ghcmessages.GBLOC(moveTaskOrder.ShipmentGBLOC[0].GBLOC),
+			OriginGBLOC:        ghcmessages.GBLOC(*moveTaskOrder.ShipmentGBLOC[0].GBLOC),
```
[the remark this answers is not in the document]
hannaliebl: I think it does need a nil check as well, I added it here: https://github.com/transcom/mymove/pull/8468/commits/c2286478fc336821ad94c2df68ae1ae24ef0168c

reggieriser (approved):
Thanks for making the changes.  The migration LGTM!

migrations/app/schema/20220420195400_update_move_to_gbloc_view_for_ppm_shipments.up.sql
duncan-truss: We may need to watch out for shipments that were deleted here

reggieriser: Good catch.  Don't think the previous query for this view did this either.  Looks like we'll need to put a `WHERE` clause to make sure `deleted_at` is null for both the `mto_shipments` records and the `ppm_shipments` records.  It'd be nice if this was covered by a test too.

hannaliebl:
@duncan-truss @reggieriser thanks for your reviews, I've added some tests (many thanks again to Reggie for his debugging skills with this) and removed deleted shipments from the query, if you'd like to take a final look.

pkg/services/order/order_fetcher_test.go
```diff
@@ -570,3 +570,65 @@ func (suite *OrderServiceSuite) TestListOrdersForTOOWithNTSRelease() {
...
+	testdatagen.MakePostalCodeToGBLOC(suite.DB(), ppmShipment.PickupPostalCode, tooOfficeUser.TransportationOffice.Gbloc)
+
+	orderFetcher := NewOrderFetcher()
+	moves, moveCount, err := orderFetcher.ListOrders(suite.AppContextForTest(), tooOfficeUser.ID, &services.ListOrderParams{})
+	suite.FatalNoError(err)
+	suite.Equal(1, moveCount)
+	suite.Len(moves, 1)
+}
+
+func (suite *OrderServiceSuite) TestListOrdersForTOOWithPPMWithDeletedShipment() {
```
duncan-truss, line 601: Thanks for adding this test. Is there a test anywhere that we're actually using the postal code of the first non-deleted shipment if multiple exist on the move?

duncan-truss (approved):
Nice work, would still like to see a test on creation date for a move with multiple shipments if that doesn't exist at some point.

pkg/services/order/order_fetcher_test.go
```diff
@@ -632,3 +632,49 @@ func (suite *OrderServiceSuite) TestListOrdersForTOOWithPPMWithDeletedShipment()
...
+	})
+	// This shipment is created after the first one, but not deleted
+	testdatagen.MakePPMShipment(suite.DB(), testdatagen.Assertions{
+		Move: move,
+		PPMShipment: models.PPMShipment{
+			PickupPostalCode: postalCode,
+			CreatedAt:        time.Now().Add(time.Minute * time.Duration(1)),
+		},
+	})
+	testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
```
duncan-truss, line 660: One thing to note for the future is that MakePPMShipment also creates the parent MTO Shipment necessary.  So there shouldn't be a need to call MakeMTOShipment afterwards, you could pass in MTOShipment fields in the assertions like DeletedAt I think.

pkg/services/order/order_fetcher_test.go
```diff
@@ -632,3 +632,49 @@ func (suite *OrderServiceSuite) TestListOrdersForTOOWithPPMWithDeletedShipment()
...
+		PPMShipment: models.PPMShipment{
+			PickupPostalCode: postalCode,
+			CreatedAt:        time.Now(),
+		},
+	})
+	// This shipment is created after the first one, but not deleted
+	testdatagen.MakePPMShipment(suite.DB(), testdatagen.Assertions{
+		Move: move,
+		PPMShipment: models.PPMShipment{
+			PickupPostalCode: postalCode,
```
duncan-truss, line 656: I would probably make these have unique PickupPostalCodes so we can assert it returned the value from the undeleted shipment.  Here it's a bit ambiguous if they both have the same test data.

PR #8477, opened 2022-04-25
Convert estimated incentive units from int32 to cents

NamibiaTorres:
## Summary

This PR changes the format of Estimated Incentive to unit Cents. 

## Setup to Run Your Code

<details>
<summary>Several backend tests were updated to reflect the new units of estimated incentive, and these tests can be run with the following command</summary>

##### Handler tests

```sh
go test ./pkg/handlers/internalapi --testify.m TestListMTOShipmentsHandler -v
```

##### Services: PPM estimator tests

```sh
go test ./pkg/services/ppmshipment --testify.m TestEstimatedIncentive -v

```

</details>

### Frontend

### Backend


pkg/handlers/internalapi/internal/payloads/payload_to_model.go
```diff
@@ -134,7 +136,7 @@ func UpdatePPMShipmentModel(ppmShipment *internalmessages.UpdatePPMShipment) *mo
 	}
 
 	// Temporarily hard code this value, until we determine this
-	estimatedIncentive := int32(1000000)
+	estimatedIncentive := unit.Cents(1000000)
```
felipe-lee, line 139: **todo**: This line (and the comment above it) should be removed since the hard-coded value is now being set in the ppm estimator service object. `EstimatedIncentive` shouldn't be part of this function anymore since the customer can't really pass that value in when making updates.

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -1018,7 +1018,8 @@ func (suite *HandlerSuite) TestListMTOShipmentsHandler() {
 				suite.Equal(expectedShipment.PPMShipment.HasProGear, returnedShipment.PpmShipment.HasProGear)
 				suite.EqualPoundPointers(expectedShipment.PPMShipment.ProGearWeight, returnedShipment.PpmShipment.ProGearWeight)
 				suite.EqualPoundPointers(expectedShipment.PPMShipment.SpouseProGearWeight, returnedShipment.PpmShipment.SpouseProGearWeight)
-				suite.EqualInt32Int64Pointers(expectedShipment.PPMShipment.EstimatedIncentive, returnedShipment.PpmShipment.EstimatedIncentive)
+				//suite.EqualInt32Int64Pointers(expectedShipment.PPMShipment.EstimatedIncentive, returnedShipment.PpmShipment.EstimatedIncentive)
+				suite.Equal(expectedShipment.PPMShipment.EstimatedIncentive, returnedShipment.PpmShipment.EstimatedIncentive)
```
felipe-lee, line 1022: **thought**: We should check for `nil` as well, and if both are not-nil, then we should dereference the `EstimatedIncentive` pointers before comparing them so that we compare the actual values.

**suggestion (blocking)**: To address my comment, I think we could re-purpose the `EqualInt32Int64Pointers` function to now be something like `EqualCentPointers` or something like that. [Here is the current function](https://github.com/transcom/mymove/blob/c40e61d30050b904fea8b5704b4850ba6ca28283/pkg/handlers/apitests.go#L305-L316). Then you can pass in the pointers to the function and it'll do the comparison. Alternatively, you can keep using the plain `Equal` but you'll need to do the `nil` checks here.

pkg/handlers/internalapi/internal/payloads/model_to_payload.go
```diff
@@ -100,8 +100,8 @@ func PPMShipment(ppmShipment *models.PPMShipment) *internalmessages.PPMShipment
 	}
 
 	if ppmShipment.EstimatedIncentive != nil {
-		int64EstimatedIncentive := int64(*ppmShipment.EstimatedIncentive)
-		payloadPPMShipment.EstimatedIncentive = &int64EstimatedIncentive
+		centEstimatedIncentive := handlers.FmtCost(ppmShipment.EstimatedIncentive)
+		payloadPPMShipment.EstimatedIncentive = centEstimatedIncentive
```
felipe-lee, line 104: **nitpick**: Instead of needing this `if` block, we can shorten it and make it like how we set `Advance` above when we create the `PPMShipment` struct.

felipe-lee (commented):
LGTM so far! I left a few comments, lmk if you have any questions

swagger-def/internal.yaml
```diff
@@ -2496,6 +2496,7 @@ definitions:
         x-omitempty: false
       estimatedIncentive:
         type: integer
+        format: cents
```
felipe-lee, line 2499: **praise**: Nice catch, I just checked that it was an int, but it's great to be specific with the format.

felipe-lee (approved):
Awesome, LGTM! Thanks for making those updates

PR #7296, opened 2021-09-02
MB-9301-connect-Validator-to-Reweigh

NamibiaTorres:
## Description

Created a function called reweighValidator that validates a reweigh and connected the validator to the ReweighCreator and ReweighValidator.

Tests were updated check the connection.

## Notes for Reviewer:
@eamahanna has a PR that contains an update to services/reweigh/reweigh_updater.go that has a new function mergeReweigh. Her updater file will be the final updater file we are using since the mergeReweigh function is necessary in our service. For now, the updateReweigh file here implements the reweighValidator and passes the tests, but keep in mind that Emily's file will eventually replace this one.

## Setup
Check that the tests are passing with the new update and addition or the ReweighValidator:

```go
go test ./pkg/services/reweigh --testify.m TestReweighCreator
```
```go
go test ./pkg/services/reweigh --testify.m TestReweighUpdater -v
``` 

**Double check that the unchanged test files are still passing:**
`go test ./pkg/services/reweigh --testify.m TestValidationRules -v`

## Code Review Verification Steps
* [ ] Have the Jira acceptance criteria been met for this change?

## References

* [Jira story](https://dp3.atlassian.net/browse/MB-9301) for this change

PR #8484, opened 2022-04-27
MB-12208 refactor payment request validation per new pattern

PDickeyTruss:
## Summary

This change done in this PR is very minor and a bit silly, it simply refactors out a single validation check from the payment request service. However, I think this change is still worthwhile as it sets up our [preferred validation pattern](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/service-validation/#the-pattern) for the payment requests service. If we ever add additional validation we have all the files (rules.go, validation.go) and patterns in place to just drop in more validation. Also, it let me write tests to explicitly test the validation that was being done for a payment request (checking the MTOID was not a null uuid). 

Note: There are other types of validation done in payment request that are not validation the payment request itself. I flirted with pulling these out into rules.go as well but it ended up being pretty gross and required abusing the intent of the new validation pattern a bit.

This is the pattern that was put into place: https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/service-validation/#the-pattern

- [ ] The payment-requests endpoint works as expected when a valid moveTaskOrderID is provided:
```
{
    "moveTaskOrderID": "91086334-b9a4-457e-8314-bad944531713",
    "serviceItems": [
        {
            "id": "3f1d996d-816d-4f29-9c4c-343b73393eb3"
        }
    ]
}
```
Screen Shot 2022-04-27 at 8 05 17 AM

- [ ] The payment-requests endpoint returns an error when the nil uuid is provided:
```
{
    "moveTaskOrderID": "00000000-0000-0000-0000-000000000000",
    "serviceItems": [
        {
            "id": "3f1d996d-816d-4f29-9c4c-343b73393eb3"
        }
    ]
}
```
Error should have message: "Invalid Create Input Error: MoveTaskOrderID is required on PaymentRequest create"

### DIFF
--- a/pkg/services/payment_request.go
+++ b/pkg/services/payment_request.go
@@ -13,7 +13,7 @@ import (
 // PaymentRequestCreator is the exported interface for creating a payment request
 //go:generate mockery --name PaymentRequestCreator --disable-version-string
 type PaymentRequestCreator interface {
-	CreatePaymentRequest(appCtx appcontext.AppContext, paymentRequest *models.PaymentRequest) (*models.PaymentRequest, error)
+	CreatePaymentRequestCheck(appCtx appcontext.AppContext, paymentRequest *models.PaymentRequest) (*models.PaymentRequest, error)
 }
 
 // PaymentRequestRecalculator is the exported interface for recalculating a payment request
--- a/pkg/handlers/primeapi/payment_request.go
+++ b/pkg/handlers/primeapi/payment_request.go
@@ -83,7 +83,7 @@ func (h CreatePaymentRequestHandler) Handle(params paymentrequestop.CreatePaymen
 				return paymentrequestop.NewCreatePaymentRequestUnprocessableEntity().WithPayload(errPayload), err
 			}
 
-			createdPaymentRequest, err := h.PaymentRequestCreator.CreatePaymentRequest(appCtx, &paymentRequest)
+			createdPaymentRequest, err := h.PaymentRequestCreator.CreatePaymentRequestCheck(appCtx, &paymentRequest)
 			if err != nil {
 				appCtx.Logger().Error("Error creating payment request", zap.Error(err))
 				switch e := err.(type) {
--- a/pkg/handlers/primeapi/payment_request_test.go
+++ b/pkg/handlers/primeapi/payment_request_test.go
@@ -101,7 +101,7 @@ func (suite *HandlerSuite) TestCreatePaymentRequestHandler() {
 		req = suite.AuthenticateUserRequest(req, subtestData.requestUser)
 
 		paymentRequestCreator := &mocks.PaymentRequestCreator{}
-		paymentRequestCreator.On("CreatePaymentRequest",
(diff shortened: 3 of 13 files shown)


pkg/services/payment_request.go
```diff
@@ -13,7 +13,7 @@ import (
 // PaymentRequestCreator is the exported interface for creating a payment request
 //go:generate mockery --name PaymentRequestCreator --disable-version-string
 type PaymentRequestCreator interface {
-	CreatePaymentRequest(appCtx appcontext.AppContext, paymentRequest *models.PaymentRequest) (*models.PaymentRequest, error)
```
mboren, line 16: Why do we need to rename this?

PDickeyTruss: Good question! I agree that is a little clunky, came from following existing patterns. `CreatePaymentRequestCheck` wraps `CreatePaymentRequest` but provides the validation checks to be run. You can see the same things in files with the refactor done like `services/ppmshipment.go` or `services/reweigh.go` (where I stole the naming pattern from).  Lines 42-44 of `payment_request_creator.go` will probably provide the most clarity on this.

PDickeyTruss: Maybe better said,  the`CreatePaymentRequestCheck` is just `CreatePaymentRequest` with checks included. Prevents us from having to pass the explicit checks we want to do every time we call `CreatePaymentRequest`.  I suspect there are other ways we could do similar but this seems to be the pattern we have been going with so far 🤷

mboren: Thanks for the context! I think in general calling out what the checks are for (like default checks vs a specific API), makes sense. I don't think the reweigh one is the best pattern to follow (looks like Sandy left a similar comment on the [reweigh validations PR]( https://github.com/transcom/mymove/pull/7296#discussion_r704586044))

Sorry about being real nitpicky here, just trying to be careful as we establish the pattern.

mboren: It looks like we don't really have a set naming convention for these though, so I'll leave it up to you.

mboren (approved):

pkg/services/payment_request.go
PDickeyTruss: Good find on that PR comment. 💯  I do agree that the naming pattern is kinda gross, I will play around a bit to see what I can come up with without renaming and we can decide what we prefer. 👍

PDickeyTruss: I spent some time playing with this, ended up just kinda chasing the problem of renaming into changing function signature, which essentially had the same effect throughout the codebase. I am sure there is a way this could be done better but I am not seeing it right off (still a go rookie). I am gonna merge as is since it matches existing patterns and add a comment to the validation refactor epic suggesting we add a ticket to investigate this renaming side effect of the validation pattern and how we might clean out everywhere it already exists with a better pattern.

PDickeyTruss: Added a comment to [the epic](https://dp3.atlassian.net/browse/MB-8885?focusedCommentId=22222) ✅

PR #8485, opened 2022-04-27
[MB-11731] PPM estimated incentive call pricer lookups

duncan-truss:
## Summary

This PR is the first pass at calculating a PPM shipment's estimated incentive from the locations, dates, and weight details during customer booking.

The PPM Estimator is currently only called from the PPM Shipment Updater service but will eventually be called from the PPM Shipment Creator service from when the Service Counselor creates a PPM Shipment all at once.

The PPM estimated incentive is priced exactly as an HHG with the same details would be totaled if the same service items were added to a payment request.

We in effect price the same service items that get auto-created and approved for an HHG:
- Domestic Linehaul
- Fuel Surcharge
- Domestic Origin Price
- Domestic Destination Price
- Domestic Packing
- Domestic Unpacking

We're sure to use the newer GHC planner that is initialized in the appcontext instead of the legacy HERE planner.  Similarly we are using the GHC rate engine as opposed to the older `pkg/rateengine`.

The strategy here was to avoid creating unnecessary database records for all of the service items and associated pricing parameter records because we don't need to present them and have them be a snapshot in time the way a payment request works.  However we want to reuse as much as the pricing code so that this didn't require a ton of new code and we'll potentially benefit from any changes and update to pricing w/o having to update a parallel pricer.

Most of the changes were to avoid getting record not found errors when lookups or pricers attempted to query the database for service items that would normally exist when creating a payment request.

**Changes not added here are to remove the minimums of 500 lbs and 50 miles in the case of estimating a PPM shipment.**

Things left to do:
- Fix Cypress tests
- Clean up some of the comments
- Add setup data that creates real service params in the database for the ppm_estimator_test.go instead of the mocked return values from that fetch query.

## Setup to Run Your Code

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. To test the PPM estimator, login using the devlocal user `dates_and_locations@ppm.unsubmitted` in the customer app
2. Choose to edit the existing PPM Shipment from the Home screen
3. Confirm the date and locations details and proceed to the estimated weights page
4. Enter 5,000 lbs for the estimated weight of the PPM shipment and click next
5. The estimated incentive should appear on the screen as $720,972

To test an HHG with the same details:
1. Login to the office app by creating a new prime simulator office user
2. Find the move with locator `#MISHIP`. This move has an HHG shipment without a destination address, missing dates (It has a requested pickup date of 3-15-2020), and missing weights.
3. We'll need to update the shipment with fields to match our PPM shipment so give it pickup dates of 3-15-2020 and an actual weight of 5,000 lbs and a destination address with the postal code 30813.
4. Create a new payment request for the move and select all of the shipment service items (exclude the basic service items if you want the payment request total to match and not just the summed shipment service items.
5. Logout of the prime simulator user and log in as the TIO.
6. Find the payment request for the move with the locator `#MISHIP` in the TIO queue and select it
7. View that the total matches the previous PPM total for the locations, dates, and weight.

### Frontend

### Backend

## Screenshots
Screen Shot 2022-04-27 at 10 59 11 AM

Screen Shot 2022-05-04 at 11 47 17 AM

Screen Shot 2022-05-04 at 11 47 52 AM

Screen Shot 2022-04-27 at 11 07 35 AM


pkg/payment_request/payment_request_helper.go
```diff
@@ -1,4 +1,15 @@
 package paymentrequest
 
+import (
+	"github.com/transcom/mymove/pkg/appcontext"
+	"github.com/transcom/mymove/pkg/models"
+)
+
+//go:generate mockery --name Helper --disable-version-string
+type Helper interface {
```
duncan-truss, line 9: Making this mockable so tests don't need to create all of the params unless they find it valuable.

pkg/payment_request/service_param_value_lookups/number_days_sit_lookup.go
```diff
@@ -195,9 +195,9 @@ func fetchAndVerifyMTOShipmentSITDates(mtoShipmentSITPaymentServiceItems models.
 		}
 	}
 
-	if isDomesticOrigin(mtoServiceItem) && !originSITEntryDate.IsZero() && mtoServiceItem.SITEntryDate != nil && originSITEntryDate != *mtoServiceItem.SITEntryDate {
+	if isDomesticOrigin(mtoServiceItem) && !originSITEntryDate.IsZero() && mtoServiceItem.SITEntryDate != nil && !originSITEntryDate.Equal(*mtoServiceItem.SITEntryDate) {
```
duncan-truss, line 198: This subtle bug has come up a few times where the timezone offset location will cause dates to be unequal because it's doing a string like comparison of `2022-05-04 00:00:00 +0000 +0000` and `2022-05-04 00:00:00 +0000 UTC`.  I think we should introduce a linter rule to catch this elsewhere so we aren't surprised when dates coming back from the database and those created in memory aren't treated as equal.

https://ustcdp3.slack.com/archives/CP6PTUPQF/p1651616044808579?thread_ts=1651614663.674939&cid=CP6PTUPQF

pkg/payment_request/service_param_value_lookups/service_param_value_lookups.go
```diff
@@ -162,304 +207,183 @@ func ServiceParamLookupInitialize(
...
 	serviceItemCode := mtoServiceItem.ReService.Code
 
-	paramKey = models.ServiceItemParamNameActualPickupDate
-	err = s.setLookup(appCtx, serviceItemCode, paramKey, ActualPickupDateLookup{
-		MTOShipment: mtoShipment,
-	})
-	if err != nil {
-		return nil, err
-	}
+	paramKeyLookups := InitializeLookups(mtoShipment, mtoServiceItem)
```
duncan-truss, line 213: I broke out this param to lookup mapping so it can be called independently.  It still sets all possible lookups, even those that won't be used for a given service item but they never get called in the end so I didn't optimized it.

pkg/payment_request/service_param_value_lookups/weight_billed_lookup.go
```diff
@@ -70,7 +70,7 @@ func (r WeightBilledLookup) lookup(appCtx appcontext.AppContext, keyData *Servic
 		if billableWeightInputs.CalculatedBillableWeight == nil {
 			return "", fmt.Errorf("got a nil calculated billable weight from service for MTOShipmentID [%s]", r.MTOShipment.ID)
 		}
-
+		// skip for PPM estimator no minimum
```
duncan-truss, line 73: There is a followup story that is meant to address no weight minimums and distance minimums https://github.com/transcom/mymove/blob/master/pkg/payment_request/service_param_value_lookups/weight_billed_lookup.go

pkg/services/ghcrateengine/fuel_surcharge_pricer.go
```diff
@@ -65,17 +67,21 @@ func (p fuelSurchargePricer) PriceUsingParams(appCtx appcontext.AppContext, para
...
 
 	var paymentServiceItem models.PaymentServiceItem
-	err = appCtx.DB().Eager("MTOServiceItem", "MTOServiceItem.MTOShipment").Find(&paymentServiceItem, params[0].PaymentServiceItemID)
-	if err != nil {
-		switch err {
-		case sql.ErrNoRows:
-			return unit.Cents(0), nil, apperror.NewNotFoundError(params[0].PaymentServiceItemID, "looking for PaymentServiceItem")
-		default:
-			return unit.Cents(0), nil, apperror.NewQueryError("PaymentServiceItem", err, "")
+	mtoShipment := params[0].PaymentServiceItem.MTOServiceItem.MTOShipment
```
duncan-truss, line 70: The PPM pricer gets around querying for PaymentServiceItems that don't exist in the database by adding those associations to the PaymentServiceItemParams argument that gets passed in here that has a value for the MTOShipment field.

pkg/services/ghcrateengine/service_item_pricer.go
```diff
@@ -45,6 +45,10 @@ func (p serviceItemPricer) PriceServiceItem(appCtx appcontext.AppContext, item m
 }
 
 func (p serviceItemPricer) getPricer(serviceCode models.ReServiceCode) (services.ParamsPricer, error) {
+	return PricerForServiceItem(serviceCode)
+}
+
+func PricerForServiceItem(serviceCode models.ReServiceCode) (services.ParamsPricer, error) {
```
duncan-truss, line 51: We don't need to display these calculations as happens in the TIO payment request so this just exposes the pricer lookup by service item code that can be called directly without the rest.

pkg/services/mto_shipment/mto_shipment_updater.go
```diff
@@ -368,9 +367,7 @@ func (f *mtoShipmentUpdater) updateShipmentRecord(appCtx appcontext.AppContext,
 	var autoReweighShipments models.MTOShipments
 	transactionError := appCtx.NewTransaction(func(txnAppCtx appcontext.AppContext) error {
 		// temp optimistic locking solution til query builder is re-tooled to handle nested updates
-		encodedUpdatedAt := etag.GenerateEtag(newShipment.UpdatedAt)
-
-		if encodedUpdatedAt != eTag {
+		if newShipment.UpdatedAt.Equal(dbShipment.UpdatedAt) {
```
duncan-truss, line 370: I was getting timestamps that were the same except for the timezone being encoded differently not unsimilar to the the other date comparison issue I fixed in this PR.

```
existing eTag MjAyMi0wNS0wNFQyMDozMjo1OS45ODg2NzFa 2022-05-04T20:33:00.041175Z
eTag to be updated MjAyMi0wNS0wNFQyMDozMzowMC4wNDExNzVa 2022-05-04 20:33:00.041175 +0000 UTC
```

src/pages/MyMove/PPMBooking/EstimatedWeightsProGear/EstimatedWeightsProGear.jsx
```diff
@@ -58,6 +59,10 @@ const EstimatedWeightsProGear = () => {
       });
   };
 
+  if (!serviceMember || !orders || !mtoShipment) {
```
duncan-truss, line 62: Formik was not rerendering the initial values like expected because mtoShipment may not resolve immediately from Redux.  This is similar to how we handle things on the office side, waiting on React-query.

Another way to do it would be to make initialValues part of the form component state and add the prop `enableReinitialize` to the `<Formik />` component.  When I tried this though it was showing a required validation error even after the values were populated properly.

pkg/services/mto_shipment/mto_shipment_updater.go
duncan-truss: I fixed this conditional by the way to be when not equal

pkg/services/mto_shipment/mto_shipment_updater.go
```diff
@@ -368,9 +369,12 @@ func (f *mtoShipmentUpdater) updateShipmentRecord(appCtx appcontext.AppContext,
 	var autoReweighShipments models.MTOShipments
 	transactionError := appCtx.NewTransaction(func(txnAppCtx appcontext.AppContext) error {
 		// temp optimistic locking solution til query builder is re-tooled to handle nested updates
-		encodedUpdatedAt := etag.GenerateEtag(newShipment.UpdatedAt)
+		updatedAt, err := etag.DecodeEtag(eTag)
+		if err != nil {
+			return StaleIdentifierError{StaleIdentifier: eTag}
```
ronaktruss, line 374: Nit: I don't think this is the most accurate error message. Would we consider a wrong etag that doesn't decode into a time a stale identifier error?

pkg/payment_request/payment_request_helper.go
pearl-truss: nice!

pkg/payment_request/service_param_value_lookups/weight_billed_lookup.go
pearl-truss: nit: the wording of this comment and its meaning isn't clear to me. And since I'm assuming this is a temporary skip, could we add a TODO here with the follow story number?

pkg/services/ghcrateengine/fuel_surcharge_pricer.go
```diff
@@ -65,17 +67,21 @@ func (p fuelSurchargePricer) PriceUsingParams(appCtx appcontext.AppContext, para
...
+
+	if mtoShipment.ID == uuid.Nil {
+		err = appCtx.DB().Eager("MTOServiceItem", "MTOServiceItem.MTOShipment").Find(&paymentServiceItem, params[0].PaymentServiceItemID)
```
pearl-truss, line 73: Is this a place where `EagerPreload` could be used to reduce db hits?

pkg/services/ppmservices.go
```diff
@@ -9,7 +9,8 @@ import (
 	"github.com/transcom/mymove/pkg/models"
 )
 
-// EstimateCalculator is the exported interface for calculating the PPM estimate
+// DEPRECATED
+// EstimateCalculator is the exported interface for calculating the legacy Personally Procured Move estimate
```
pearl-truss, line 13: Nit: could be worth adding the name of the new service

pkg/testdatagen/scenario/shared.go
```diff
@@ -3519,16 +3519,16 @@ func createApprovedMoveWithMinimalShipment(appCtx appcontext.AppContext, userUpl
 		Move: move,
 	})
 
-	requestedPickupDate := time.Now().AddDate(0, 3, 0)
-	requestedDeliveryDate := requestedPickupDate.AddDate(0, 1, 0)
+	// requestedPickupDate := time.Now().AddDate(0, 3, 0)
```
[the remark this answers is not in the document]
duncan-truss: I use it more of as an optional toggle if I wanted it to be relative or have certain values.  I guess I could convert it to be overridable from the outside but I'm not sure it's being used anywhere else.

pkg/services/mto_shipment/mto_shipment_updater.go
ronaktruss: (Non-blocking): It's unclear to me what specific user facing errors we would want to have, but I know there are some restrictions on what we want to expose. Since the code is open sourced, I'm not quite sure if we are really "exposing" anything.

I'd probably advocate for a generic system error on the UI for the customer. (Having different errors here also makes it more apparent at which point the code fails if we look at logs.)

ronaktruss (approved):
This looks good to me outside of the nits above and the pending merge conflicts. Approving in advance. (If you don't get to any of the changes feel free to let me or the team know and we can help get this merged in your absence.)

pkg/services/mto_shipment/mto_shipment_updater.go
duncan-truss: I think the main concern with some client errors are they could reveal PII or other environmental secrets that maybe aren't checked into the code per se.  There was a POAM a while back to not reveal versions and stack information of the server.  I don't know that that's generally happening here but is a good rule of thumb.

I think it's reasonable to throw a different error here like a bad request if we want to keep this.  It will just mean updating our eTag tests to use well formatted data.  A lot of time we log at the handler level the error itself for developers before stripping that information from the payload and maybe only returning a static error code and message.

There's a lot of room to improve around logging and pricer errors that I think could follow up on this initial PR to make it better.

duncan-truss:
> This looks good to me outside of the nits above and the pending merge conflicts. Approving in advance. (If you don't get to any of the changes feel free to let me or the team know and we can help get this merged in your absence.)

@ronaktruss Thanks for the review.  I am pretty much pencils down at this point, so I'll let you all decide what you want to do with it.

pearl-truss (approved):
Looks good to me :)

pkg/services/ppmservices.go
NamibiaTorres: I think this is fine, because I will delete the file.

gidjin (approved):
LGTM!

PR #8488, opened 2022-04-27
MB-12307 Convert supportapi tests to run within a transaction

sandy-wright:
## Summary

Converts all remaining tests in `/pkg/handlers/supportapi` to use transactions.

## Setup to Run Your Code

## Verification Steps

- [ ] No subtests are run with `suite.T().Run(...)` - they all use `suite.Run(...)`
- [ ] There is no unjustified usage of `Truncate` in the tests
- [ ] There is no unjustified usage of `suite.AppContextForTest().DB()` - instead `suite.DB()` is used directly
- [ ] `Fatalf` is deprecated in favor of other assertions/checks
- [ ] Go's `testing` package is only imported in the setup test file

### DIFF
--- a/pkg/handlers/supportapi/api_test.go
+++ b/pkg/handlers/supportapi/api_test.go
@@ -17,8 +17,8 @@ type HandlerSuite struct {
 }
 
 // AfterTest completes tests by trying to close open files
-func (suite *HandlerSuite) AfterTest() {
-	for _, file := range suite.TestFilesToClose() {
+func (hs *HandlerSuite) AfterTest() {
+	for _, file := range hs.TestFilesToClose() {
 		//RA Summary: gosec - errcheck - Unchecked return value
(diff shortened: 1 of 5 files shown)


pkg/handlers/supportapi/api_test.go
```diff
@@ -17,8 +17,8 @@ type HandlerSuite struct {
 }
 
 // AfterTest completes tests by trying to close open files
-func (suite *HandlerSuite) AfterTest() {
-	for _, file := range suite.TestFilesToClose() {
+func (hs *HandlerSuite) AfterTest() {
+	for _, file := range hs.TestFilesToClose() {
```
sandy-wright, line 21: `suite` conflicts with an imported testing package in this file, so it needed to be changed here. The other test files do not import the third-party `suite` package, so they're fine as-is

shimonacarvalho (approved):
Tested. Checked for suite.T().Run or other red flags. Looks good to me.

PR #8513, opened 2022-04-28
Add sort-imports rule

rogeruiz:
## Summary

This change adds the `sort-imports` rule to ESLint to help mitigate some of the issues that @YanZ777 brought up in our Confluence discussion. I have copied over the defaults from the documentation, but I reordered the `memberSyntaxSortOrder` by reversing it. Please feel free to change it while reviewing this.

[=> The documentation on _sort-imports_](https://eslint.org/docs/rules/sort-imports)

## Setup to Run Your Code

Load up a file, and test that `sort-imports` is correctly showing errors on `jsx` files.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -37,6 +37,16 @@ module.exports = {
     'react/jsx-curly-brace-presence': 'error',
     'react/require-default-props': 'warn',
     'no-unused-vars': 'warn',
+    'sort-imports': [
+      'error',
+      {
+        ignoreCase: false,
+        ignoreDeclarationSort: false,
+        ignoreMemberSort: false,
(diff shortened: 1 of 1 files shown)


rogeruiz:
ay ay ay, adding this rule as `error` gives us this.

```sh
✖ 2280 problems (2280 errors, 0 warnings)
[3623](https://github.com/transcom/mymove/runs/6218793721?check_suite_focus=true#step:5:3623)
  331 errors and 0 warnings potentially fixable with the `--fix` option.
```

We probably want to make this a `warning` instead. And add the `--fix` in general to our Pre-Commit maybe

rogeruiz:
Closing this for now but I'll keep the branch around. Not sure where we landed on this, but the work on the branch is only a small part of it that is easily reproducible.

PR #8483, opened 2022-04-27
MB-11319 document when the too updates the shipment details

YanZ777:
## Summary

I didn't need to add a migration to add a trigger to the entitlements table. 
I did need to add a migration to add a trigger to the mto_agents table.
The trigger for the addresses table already existed.

formatters for formatting the address and agents were added into the `formatters.js` file.

The events for updating and creating agents, updating address have special logic inside of their `getDetailsLabeledDetails` to handle formatting addresses. I pull from `oldValues` to backfill the information so the text being displayed looks whole since what gets returned in new values is only the information that got changed.

## Setup to Run Your Code

<details>
<summary>💻 You will need to use three separate terminals to test this locally.</summary>

##### Terminal 2

Start the UI locally.

```sh
make office_client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the office application
2. Login as a TOO
3. Select a move
4. On the move details page, select a shipment to edit. 
5. Edit as many or as little fields as you desire and save.
6. Go to the move history tab and see the updates to the shipments.

## Verification Steps for Author

These are to be checked by the author.
- [ ] Request review from a member of a different team.
- [ ] Have the Jira acceptance criteria been met for this change?

## Verification Steps for Reviewers

These are to be checked by a reviewer.

### Frontend

- [ ] User facing changes have been reviewed by design.
- [ ] There are no aXe warnings for UI.
- [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
- [ ] There are no new console errors in the browser devtools
- [ ] There are no new console errors in the test output

### Backend

- [ ] Code follows the guidelines for [Logging](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide#logging)
- [ ] The requirements listed in [Querying the Database Safely](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide/#querying-the-database-safely) have been satisfied.

#### Any new migrations/schema changes:

- [ ] Follows our guidelines for [Zero-Downtime Deploys](https://transcom.github.io/mymove-docs/docs/dev/contributing/database/Database-Migrations#zero-downtime-migrations)
- [ ] Have been communicated to #g-database
[...]

PR #8518, opened 2022-04-28
[ MB-11406 ] Document TIO updates shipment billable weight

rogeruiz:
## Summary

When a TIO reviews the billable weight of shipments, their actions will be captured in the history log with the following details:

- [ remarks ] tio_jusitification
- [shipment type ] shipmentBillable
- weight: [ value ] lbs

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. As a **TIO**, click on Review Shipment Billable Weight and modify the weight and add TIO Remarks
1. See the new weight and the TIO Remarks updated. 
1. See the history log and find new log entries for shipment type and weight values being updated.


src/constants/historyLogUIDisplayName.js
```diff
@@ -22,7 +22,8 @@ export const HistoryLogRecordShape = PropTypes.shape({
 export const dbFieldToDisplayName = {
   updated_at: 'Updated at',
   diversion: 'Diversion',
-  billable_weight_cap: 'Billable weight cap',
+  billable_weight_cap: 'Billable weight',
```
rogeruiz, line 25: @juliemoon In the description of MB-11406, it says that the the field should be **Billable weight** and not **Billable weight cap**. Is this still accurate?

src/pages/Office/MoveHistory/LabeledDetails.jsx
```diff
@@ -54,14 +54,12 @@ const LabeledDetails = ({ changedValues, context, getDetailsLabeledDetails }) =>
 };
 
 LabeledDetails.propTypes = {
-  changedValues: HistoryLogValuesShape,
-  context: PropTypes.arrayOf(PropTypes.object),
+  historyRecord: HistoryLogRecordShape,
```
rogeruiz, line 57: I've updated the props we pass into LabeledDetails because I couldn't think of a better way to handle all the edge cases around history records. It's also because we trigger the callback from inside this component. Is there a better way to do this @YanZ777? This is related to my question over in Truss Slack about passing props around.

src/pages/Office/MoveHistory/LabeledDetails.test.jsx
```diff
@@ -5,36 +5,41 @@ import LabeledDetails from './LabeledDetails';
...
-      new_duty_location_name: 'New duty location name',
-      orders_number: '1111',
-      tac: '2222',
-      sac: '3333',
-      nts_tac: '4444',
-      nts_sac: '5555',
-      department_indicator: 'AIR_FORCE',
-      grade: 'E_1',
+    const historyRecord = {
+      changedValues: {
```
rogeruiz, line 9: Since Labeled Details now leverages history record data rather than only changed values, these tests needed to change a bunch.

src/pages/Office/MoveHistory/MoveHistory.jsx
```diff
@@ -27,7 +27,11 @@ const columns = [
   createHeader(
     'Details',
     (row) => {
-      return <MoveHistoryDetailsSelector historyRecord={row} />;
+      return (
+        <div className={styles.details}>
+          <MoveHistoryDetailsSelector historyRecord={row} />
+        </div>
```
rogeruiz, line 33: I wrapped this in a `<div>` in order to give it the `.details` styles. But I am curious if there's a better way to do this? Like I would think it would be better to pass the `styles.details` down into the components, but this was the easier thing to do. But it doesn't sit well with me. I'm curious what's the best practice around this?

src/constants/historyLogUIDisplayName.js
juliemoon: Hi Roger, seeing this now, and thanks for calling this out. It should say **max billable weight and remarks**. I'd like to use what is in the UI to continue the consistency for the users and the application.

src/pages/Office/MoveHistory/LabeledDetails.jsx
YanZ777: I think at some point we were just going to converge on passing historyRecord around. I probably would have just done this eventually as well.

YanZ777:
I merged [my PR](https://github.com/transcom/mymove/pull/8483), which probably caused a lot of conflicts. Let me know if you need help resolving them. 
I suspect a lot of the merge conflicts inside of `moveHistoryEventTemplate` are because I alphabetized the events. 
`LabeledDetails` might be tricky to merge.
I'm not sure about how the others will go.

src/pages/Office/MoveHistory/MoveHistory.jsx
YanZ777: I think wrapping it in a `div` to give it the `.details` styles is reasonable if __all__ the children need those styles (which I think they probably do).

juliemoon (commented):
@rogeruiz This is what I'm seeing when I edit the shipment billable weight and I enter remarks. 
Screen Shot 2022-05-05 at 10 24 18 AM

YanZ777 (approved):

michaelvaldes-truss (approved):
lgtm!

juliemoon (approved):
This looks great!!

PR #8530, opened 2022-04-29
MB-12315 Convert adminapi tests to run in transaction

sandy-wright:
## Summary

Converts all remaining tests in `/pkg/handlers/adminapi` to use transactions.

## Setup to Run Your Code

## Verification Steps

- [x] Go's `testing` package is only imported in the setup test file


pkg/handlers/adminapi/admin_users_test.go
```diff
@@ -285,27 +224,21 @@ func (suite *HandlerSuite) TestCreateAdminUserHandler() {
 }
 
 func (suite *HandlerSuite) TestUpdateAdminUserHandler() {
-	adminUserID, _ := uuid.FromString("00000000-0000-0000-0000-000000000000")
+	adminUserID, _ := uuid.FromString("00000000-0000-0000-0000-000000000001")
```
shimonacarvalho, line 227: For a real non-nil uuid, I like to use the following which says "create a v4 version uuid"
`adminUserID := uuid.Must(uuid.NewV4())`

pkg/handlers/adminapi/admin_users_test.go
```diff
@@ -322,41 +255,31 @@ func (suite *HandlerSuite) TestUpdateAdminUserHandler() {
...
+		params := adminuserop.UpdateAdminUserParams{
+			HTTPRequest: suite.setupAuthenticatedRequest("PUT", fmt.Sprintf("/admin_users/%s", adminUserID)),
+			AdminUser: &adminmessages.AdminUserUpdatePayload{
+				FirstName: &adminUser.FirstName,
+				LastName:  &adminUser.LastName,
+			},
+		}
+
 		adminUserUpdater := &mocks.AdminUserUpdater{}
+		err := validate.NewErrors()
```
shimonacarvalho, line 268: Does this do extra validation? or create an empty verrs? 

I'm a little confused about the changes in this test, since there are some functional changes, especially the mocked call below. 
Was it broken before? Could you maybe add the test comment? 

```
		// TESTCASE SCENARIO
		// Under test: Function under test
		// Mocked:     Mocked objects
		// Set up:     
		// Expected outcome:
```

shimonacarvalho (approved):
Some minor comments but looks good, ran and tested.

PR #8559, opened 2022-05-06
MB-12120 Update shipment service objects to be managed by an orchestrator service object

felipe-lee:
## Summary

This adds new orchestration service objects to manage the `MTOShipment` and `PPMShipment` service objects. 

One thing to note is that this is a new type of service object. I ended up putting them in the same folder as the existing service objects and adding `orchestrator` to the name. I'm not really set on this though an would gladly take input from y'all on naming and placement.

If the full PR is too much to review at once, it might be easier to go commit by commit since I try to keep related changes together.

<details>
<summary>Change Log</summary>

Things that should be addressed by this in the end:

- [x] Update `ppmShipmentCreator` to not depend on `MTOShipmentCreator` anymore. This will bring it more in line with the PPM Updater.
- [x] Add more robust tests for PPM Creator
- [x] Add more robust tests for PPM Updater
- [x] Add service object to orchestrate MTOShipment creator and PPMShipment creator.
- [x] Update internalapi create shipment handler to use new shipment creation orchestrator service object (instead of calling the MTOShipment or PPMShipment ones directly).
- [x] Add service object to orchestrate MTOShipment updater and PPMShipment updater.
- [x] Update internalapi update shipment handler to use new shipment update orchestrator service object (instead of calling the MTOShipment or PPMShipment ones directly).
- [x] Update PPM customer pages to pass shipmentType in payload. It was missing from a few update payloads which was causing issues with the latest changes.

Other things I addressed while I was working on this:
- [x] Fix a bug with PPMShipment pro gear fields. If a user selected no after having saved yes, the old pro gear weights weren't cleared out before.
- [x] Update internalapi create and update tests to follow [newer testing patterns](https://transcom.github.io/mymove-docs/docs/backend/testing/running-server-tests-inside-a-transaction/#how-to-update-a-test-to-the-new-pattern).
- [x] Update `mtoShipmentCreator` struct and initializing function. We had a field we didn't need and were unnecessarily creating a new fetcher even though one gets passed in so I removed the unused field and started using the fetcher that gets passed in.

</details>

Another thing to note is that for this ticket I'd ideally also update docs to include info on orchestrator service objects. I'll do my best to get to that, but if I don't, I'll create a ticket to follow up with that.

## Setup to Run Your Code

You can test this by running server tests and/or by interacting with the app as a customer and ensuring things still work for PPM Shipments (creating/updating).

### Run server tests

<details>
<summary>💻</summary>

```sh
make server_test
```

</details>

### Use the app

<details>
<summary>💻 Set Up</summary>

#### Server

Start the Go server locally.

```sh
make server_run
```

#### Client UI

Start the UI locally.

```sh
make client_run
```

</details>

#### Steps

1. Access the [local client app](milmovelocal:3000/)
    1. Alternatively, you can use the [ephemeral deploy](https://my-milmove-pr-8559.mymove.sandbox.truss.coffee/sign-in)
1. Login as a customer with an unsubmitted move.
    1. You could use `dates_and_locations@ppm.unsubmitted` or really any of the ones that end with `@ppm.unsubmitted`
1. Try adding a PPM and editing a PPM. Both should still work like before.

### Backend


pkg/services/shipmentorchestrator/shipment_creator.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+}
+
+// CreateShipment creates a shipment, taking into account different shipment types and their needs.
+func (s *shipmentCreator) CreateShipment(appCtx appcontext.AppContext, shipment *models.MTOShipment) (*models.MTOShipment, error) {
+	if err := validateShipment(appCtx, *shipment, s.checks...); err != nil {
+		return nil, err
+	}
+
+	isPPMShipment := shipment.ShipmentType != "" && shipment.ShipmentType == models.MTOShipmentTypePPM
```
ronaktruss, line 31: I don't think we need the first part of this check.

felipe-lee: Ah good point. I had that from before I made the `rules.go`. I'll update it

ronaktruss:
For the most part this looks good to me (still finishing up the review on the last commit and want to review the tests again).

Question: Do we have a standard for folder naming? I see both snake_case and all lowercase within the application. I couldn't find any documentation around it. (Depending on that answer I would potentially advocate for snake_case for the `shipmentorchestrator` folder. )

Thought: It might be worth having an additional folder for all orchestrators within the services folder.

ronaktruss (approved):
Changes look good to me. Thanks for adding to the existing tests!

felipe-lee:
> Question: Do we have a standard for folder naming? I see both snake_case and all lowercase within the application. I couldn't find any documentation around it. (Depending on that answer I would potentially advocate for snake_case for the `shipmentorchestrator` folder. )

So, we don't really have any ourselves. We just link to a go presentation on naming:
* [Our docs](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/golang-guide/#style-and-conventions)
* [Go presentation](https://talks.golang.org/2014/names.slide#1)

Originally I'd made it snake case, but go (I think it was actually one of our linters 🤔) doesn't allow snake case package names so I opted to make the folder name match the package name. I'm not a fan of the joined name though.

> Thought: It might be worth having an additional folder for all orchestrators within the services folder.

Yeah, that's a good idea. Could even help with the naming from your question above since it could then just become `orchestrators/shipment` so it wouldn't need to be part of the package name. Though that might mess with imports, but those can be aliased if needed. I'll test it out and see how it looks.

pkg/services/shipmentorchestrator/shipment_updater_test.go
```diff
@@ -0,0 +1,325 @@
...
+		"error updating PPMShipment": {
+			shipmentType:              models.MTOShipmentTypePPM,
+			returnErrorForMTOShipment: false,
+			returnErrorForPPMShipment: true,
+		},
+	}
+
+	for name, tc := range serviceObjectErrorTestCases {
+		name := name
+		tc := tc
```
ronaktruss, line 259: Nit: I assume these are reassigned because of the potential Implicit memory aliasing issue when using range? I don't think these are needed in this instance since we aren't referencing the addresses of these values.

felipe-lee: It's more to do with testing in parallel. We don't do it currently, and I don't think we can even do it currently. But If we wanted to, every table-driven test we have would have to be updated to re-declare the variables like this.

Some reference docs:
* https://github.com/golang/go/wiki/TableDrivenTests
* https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables
* https://gopherguides.com/articles/table-driven-testing-in-parallel
    * This one is longer and doesn't really cover it until almost the very end, but it is more detailed.

ronaktruss: Ahh okay, thanks for the references! 🎉

PR #8575, opened 2022-05-09
[MB-12327] GHC API can create PPM shipments

reggieriser:
## Summary

This PR adds the ability for a PPM shipment to be created via the GHC API.  As part of this work, a migration was included to [add some SIT-related fields/types](https://ustcdp3.slack.com/archives/C0250Q8469K/p1651519128275709) that will be exposed in the office UI.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

There are tests to exercise this functionality, but the UI that would invoke this is not available yet.

### Backend


donaldthai (approved):
DB review. Ran the migration locally and seeing the changes 👍

felipe-lee:
I can take a deeper look tomorrow, but depending on which merges first, either this one or #8559 is going to have conflicts to resolve. I'm changing how the PPM Creator works in mine, and creating a higher level shipment orchestrators that manage the MTOShipment and PPMShipment objects. I'd hoped to get to that before y'all started on this, but we just weren't able to pull it in sooner 😞 but at least it doesn't seem like the merge conflicts will be too much to resolve 🤞🏼

swagger-def/ghc.yaml
```diff
@@ -3334,10 +3334,84 @@ definitions:
       storageFacility:
         x-nullable: true
         $ref: 'definitions/StorageFacility.yaml'
+      ppmShipment:
```
mlm55, line 3337: Should this have x-nullable: true?

mlm55:
I was able to use postman to create ppm shipment, but the shipment from the response didn't return the associated ppmShipment object

felipe-lee:
> I can take a deeper look tomorrow, but depending on which merges first, either this one or #8559 is going to have conflicts to resolve. I'm changing how the PPM Creator works in mine, and creating a higher level shipment orchestrators that manage the MTOShipment and PPMShipment objects. I'd hoped to get to that before y'all started on this, but we just weren't able to pull it in sooner 😞 but at least it doesn't seem like the merge conflicts will be too much to resolve 🤞🏼

@reggieriser so there were more conflicts than I'd hoped. I ended up fixing them in #8594 and updating to use the new shipment orchestrator. Unfortunately that PR diff with this one looks huge because of all the changes from master 😞 I'm not sure if you want to merge that in or if you'd rather resolve the conflicts directly on your branch. I'm also happy to pair or help out however is best for you

swagger-def/ghc.yaml
reggieriser: It looks like the [shared definition has the `x-nullable: true`](https://github.com/transcom/mymove/blob/47f004be7a6de5847d4a8e33d2f2b62e3933ce0d/swagger-def/definitions/PPMShipment.yaml#L2) on it.

reggieriser:
> I was able to use postman to create ppm shipment, but i don't think the shipment in the response returned the associated ppmShipment object

I was able to duplicate this and see that the PPMShipment was not in the model that we were passing to the MTOShipment model-to-payload functions when the request came in via Postman.  My tests were passing, though, and I'm not sure what's different about how things are being set up.  That said, I was able to associate the payload explicitly in the service and that seemed to fix the issue via Postman.

reggieriser:
@mlm55 @felipe-lee I *think* I've completed the merge with the latest master and made the needed adjustments to code in my PR to use the new orchestration.  Thanks a ton to @felipe-lee for PR #8594 which I referenced quite a bit in doing this.  I ended up merging master into my branch directly as both master and my branch had progressed a bit since the PR, plus it helped me understand the new pattern better.  Ready for a re-review.

pkg/services/ppmshipment/ppm_shipment_creator_test.go
```diff
@@ -3,18 +3,22 @@ package ppmshipment
...
+	"github.com/transcom/mymove/pkg/services/mocks"
 	"github.com/transcom/mymove/pkg/testdatagen"
 )
 
 func (suite *PPMShipmentSuite) TestPPMShipmentCreator() {
 
 	// One-time test setup
-	ppmShipmentCreator := NewPPMShipmentCreator()
+	ppmEstimator := mocks.PPMEstimator{}
+	ppmShipmentCreator := NewPPMShipmentCreator(&ppmEstimator)
```
felipe-lee, line 21: **suggestion (non-blocking)**: Could you add a success test case for creating a PPMShipment with the info you'd expect to come in from the GHC side? I think based on everything else it should work, but I think it'd be good to have a unit test at this level to make sure it does work as expected here.

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -228,6 +229,12 @@ func (suite *HandlerSuite) TestCreateMTOShipmentHandler() {
 			ShipmentType: &ppmShipmentType,
 		}
 
+		ppmEstimator.On("EstimateIncentiveWithDefaultChecks",
+			mock.AnythingOfType("*appcontext.appContext"),
+			mock.AnythingOfType("models.PPMShipment"),
+			mock.AnythingOfType("*models.PPMShipment")).
+			Return(models.CentPointer(unit.Cents(estimatedIncentive)), nil).Once()
```
felipe-lee, line 236: **thought**: Ah, I realize now I saw this when I was doing the merge but didn't think about how this should actually be set to return `nil, nil`. When the customer is creating the shipment, the estimator shouldn't be able to return a value for estimated incentive until later on in the process.

**todo**: Can you update this to be `Return(nil, nil).Once()`? I mainly don't want this to be misleading. That would also mean you can take out the estimated incentive check below (and the var declaration on 219).

pkg/handlers/ghcapi/mto_shipment_test.go
```diff
@@ -2222,6 +2239,204 @@ func (suite *HandlerSuite) TestCreateMTOShipmentHandler() {
 	})
 }
 
+func (suite *HandlerSuite) TestCreateMTOShipmentHandlerUsingPPM() {
+	suite.Run("Successful POST - Integration Test (PPM, all fields)", func() {
+		// Make a move along with an attached minimal shipment. Shouldn't matter what's in them.
+		move := testdatagen.MakeDefaultMove(suite.DB())
+		hhgShipment := testdatagen.MakeMTOShipmentMinimal(suite.DB(), testdatagen.Assertions{
+			Move: move,
+		})
```
felipe-lee, line 2248: **thought**: Since it seems like we're mostly using this to use its `RequestedPickupdate`, I'd vote for instead using `time.Time{}` with some value. Mainly because our handler tests can already be slow so if we don't need to create more data in the DB, it can be nice to skip it. I'm not strongly opposed in this case though so I'll leave it up to you.

pkg/handlers/ghcapi/mto_shipment_test.go
```diff
@@ -2222,6 +2239,204 @@ func (suite *HandlerSuite) TestCreateMTOShipmentHandler() {
...
+			suite.NotZero(ppmPayload.UpdatedAt)
+		}
+	})
+
+	suite.Run("Successful POST - Integration Test (PPM, minimal fields)", func() {
+		// Make a move along with an attached minimal shipment. Shouldn't matter what's in them.
+		move := testdatagen.MakeDefaultMove(suite.DB())
+		hhgShipment := testdatagen.MakeMTOShipmentMinimal(suite.DB(), testdatagen.Assertions{
+			Move: move,
+		})
```
felipe-lee, line 2359: **thought**: Same as above, like if we don't need to create the data in the DB, I'd prefer to skip it.

pkg/handlers/ghcapi/mto_shipment_test.go
```diff
@@ -2222,6 +2239,204 @@ func (suite *HandlerSuite) TestCreateMTOShipmentHandler() {
...
+			suite.Equal(sitExpected, ppmPayload.SitExpected)
+			suite.Equal(&sitLocation, ppmPayload.SitLocation)
+			suite.Equal(handlers.FmtPoundPtr(&sitEstimatedWeight), ppmPayload.SitEstimatedWeight)
+			suite.Equal(handlers.FmtDate(sitEstimatedEntryDate), ppmPayload.SitEstimatedEntryDate)
+			suite.Equal(handlers.FmtDate(sitEstimatedDepartureDate), ppmPayload.SitEstimatedDepartureDate)
+			suite.Equal(handlers.FmtPoundPtr(&estimatedWeight), ppmPayload.EstimatedWeight)
+			suite.Equal(&hasProGear, ppmPayload.HasProGear)
+			suite.Equal(handlers.FmtPoundPtr(&proGearWeight), ppmPayload.ProGearWeight)
+			suite.Equal(handlers.FmtPoundPtr(&spouseProGearWeight), ppmPayload.SpouseProGearWeight)
+			suite.Equal(ghcmessages.PPMShipmentStatusDRAFT, ppmPayload.Status) // TODO: Should this be submitted?
```
felipe-lee, line 2347: **thought**: To answer your question, I think it probably should be submitted when a SC creates the shipment. At least, that was my understanding before y'all started implementation work, but I could be wrong. Might be good to double check with Julie.

pkg/handlers/ghcapi/internal/payloads/payload_to_model.go
```diff
@@ -237,6 +240,64 @@ func MTOShipmentModelFromCreate(mtoShipment *ghcmessages.CreateMTOShipment) *mod
...
+			model.SITEstimatedDepartureDate = sitEstimatedDepartureDate
+		}
+	}
+
+	model.EstimatedWeight = handlers.PoundPtrFromInt64Ptr(ppmShipment.EstimatedWeight)
+
+	if model.HasProGear != nil && *model.HasProGear {
+		model.ProGearWeight = handlers.PoundPtrFromInt64Ptr(ppmShipment.ProGearWeight)
+		model.SpouseProGearWeight = handlers.PoundPtrFromInt64Ptr(ppmShipment.SpouseProGearWeight)
+	}
```
felipe-lee, line 299: **nit**: I _think_ these could probably go in the model initialization on line 257 without needing to be broken out here. Kind of like [we do for internal updates](https://github.com/transcom/mymove/blob/5ddeaab9e6aba25e4b096bae2d7673e45bb42cb0/pkg/handlers/internalapi/internal/payloads/payload_to_model.go#L141-L145). Not super important though, just a small thing.

felipe-lee (approved):
This is great, thanks for dealing with all those merge conflicts, and sorry about causing quite a few of them...

I left a few comments, some I do think need to be done before merging, but they're mostly small so I'm leaving my approval regardless. There's a few others that are just thoughts or comments, but aren't blocking.

I will add though that I didn't test with postman and approving based on looking at the code seeing the tests pass in CI

mlm55 (approved):
used postman to create a ppm shipment and everything worked as expected

pkg/handlers/ghcapi/mto_shipment_test.go
reggieriser: My thinking here was to create a move that already had an existing HHG since I don't think the SC would likely get a move that didn't have any associated shipments.  Then, I add a new PPM to it.  May be overkill, but was just making sure the existence of a shipment on the move didn't cause any issues when adding a PPM.  Other than that, you're right that I could just use some `time.Time` value instead.

felipe-lee: Ah ok. That's a good point, we haven't gotten to where we explicitly make sure combo moves work yet, but no harm in having test cases that show parts of the system work with combos since now.

pkg/handlers/ghcapi/internal/payloads/payload_to_model.go
reggieriser: My thinking here with the `SitExpected` and `HasProGear` flags was to not even set fields (at least on create) that shouldn't be applicable if the flag was `false`.  So if somebody tries to set them in the incoming payload for some reason, they're just ignored instead of potentially ending up in the record/payload (which probably won't cause issues, but might be a little odd looking).

pkg/services/ppmshipment/ppm_shipment_creator_test.go
reggieriser: Added a test case with expected GHC data.

pkg/handlers/internalapi/mto_shipment_test.go
reggieriser: I fixed the mock to return `nil` instead of an incentive value and got rid of the check/variable for an incentive.

pkg/handlers/ghcapi/internal/payloads/payload_to_model.go
felipe-lee: Ah ok. I was seeing this func as more of a direct translation and leaving weird states to be addressed in validation, but am good with this approach too.

felipe-lee (approved):
Thanks for making those updates! Reviewed latest changes and it lgmt!

PR #8577, opened 2022-05-09
Removed unneeded imports from tools.go and cleaned up go.mod

reggieriser:
## Summary

I was trying to run a tool to make our import order consistent (which I may try to do in a later PR) but I noticed a couple of things in doing so which I'm addressing here:
- Our `tools.go` has some imports labeled "test packages" for things that don't look like tools to me.  Am I missing something?  I removed them and everything built fine.
- The `go.mod` file had been fixed a while back (I thought) to have just two `require` sections -- one for direct dependencies and one for indirect ones.  But we had regressed back to the three sections we had before with mixed indirect/direct dependencies.  I combined all `require` sections together and ran `go mod tidy` and got the two sections back.

### Additional steps

Does everything still seem to build OK?  Am I missing something about these files?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,8 +24,6 @@ require (
 	github.com/go-openapi/strfmt v0.21.2
 	github.com/go-openapi/swag v0.21.1
 	github.com/go-openapi/validate v0.21.0
-	github.com/go-playground/locales v0.14.0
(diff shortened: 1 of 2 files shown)


ahobson (approved):
🚀 Thank you!

PR #8578, opened 2022-05-09
MB-12237 Permissions POC

PDickeyTruss:
Warning! Do not merge this PR, it contains POC work on permissions with some example restrictions applied to the app that we do not want in the codebase. 

## Summary
This is a POC PR that explores how we could better manage user access to specific buttons, pages, components in the application. This is prompted by our access requirements growing in complexity with new roles being introduced that have significantly different access requirements (QAC/CSR). This PR is intended to both spur discussion as well as provide an example of how we could better manage access. 

There are 3 aspects to this work:
1) **Extend our current 'roles' implementation by adding the concept of permissions**. I have created an ADR to capture this decision and the details around it. It is implemented and working in this PR (see first few commits). 

2) **Make use of user permissions on the frontend to manage what a user can see and do (hide or change buttons, pages, components, etc).** This is implemented and working in this PR (see last few commits). 

3) **Make use of user permissions on the backend to manage/restrict API access**. This is being done by Leia [MB-12238](https://dp3.atlassian.net/browse/MB-12238) and _no work related to this is included in this PR_ (though we of course plan to make use of the same permissions tables/rules on both FE and BE).

### Background and Context
I read many different articles and threads around roles and changing UI based on roles. They almost all scolded our current approach of checking a users roles as not scalable and hard to maintain. Most everything I read suggested the use of **permissions** in addition to roles to solve the common problems and pain points of roles alone. I created [ADR-0070-use-permissions-to-manage-access](TODO) to provide more info on this and support the decision to introduce permissions.

When it came to using these roles and/or permissions on the frontend, the approaches varied somewhat (mostly just syntactically) but did all clearly favor using permissions rather than roles alone.  The frontend approach I found most intriguing came from patterns suggested in [this](https://medium.com/geekculture/how-to-conditionally-render-react-ui-based-on-user-permissions-7b9a1c73ffe2) article.  I felt this approach was relatively easy to read/understand in code and not overly complicated to implement. It also required no external packages and felt maintainable as roles/permissions grow in complexity and number.  I was able to get this flushed out and working, it can be seen in this PR.

## So what is actually in this PR? 
This PR includes the work needed to setup and retrieve permissions from the database and make use of those permissions (for the current user) in the office app. I added some example access restrictions to a few different buttons in the office app with the primary purpose of demonstrating the new patterns. I created 4 'permissions' in the permissions table ('edit.shipment', 'edit.orders', 'edit.max_billable_weight', and 'edit.financial_review_flag') and assigned each of these 4 permissions to the TOO and services counselor but **NOT** to the TIO (for testing purposes). In the FE, I used the new `Restricted` component to demonstrate hiding, modifying, or showing components based on these permissions.  For specifics, see the Implementation Overview below. 

## Setup to Run Your Code

A new migration is included in this PR, be sure to run the migrations first (`make db_dev_fresh` or similar). 

### Validation/Review steps

- [ ] Run migrations and start server and client
- [ ] Access the office application
- [ ] Login as a TOO (has all 4 permissions)
    - [ ] The Flag for Financial Review button is shown
    - [ ] The 'Edit Orders' button is labeled 'Edit Orders'
    - [ ] The 'Review Weights' button is shown
    - [ ] The 'Edit Shipment' button is shown 
- [ ] Login as a TIO (has none of the 4 permissions)
    - [ ] The Flag for Financial Review button is not shown
    - [ ] The button in the orders section is now labeled 'View Orders' rather than 'Edit Orders'
    - [ ] The 'Review Weights' button is not shown
    - [ ] The 'Edit Shipment' button is not shown
- [ ] Logging as a user with both TIO/TOO roles (permissions based on all roles a user has)
    - [ ] The Flag for Financial Review button is shown
    - [ ] The 'Edit Orders' button is labeled 'Edit Orders'
    - [ ] The 'Review Weights' button is shown
    - [ ] The 'Edit Shipment' button is shown (since the TOO role has permission). 

Screens of restricted TIO access:
Screen Shot 2022-05-10 at 9 25 35 AM
Screen Shot 2022-05-10 at 9 26 43 AM

### Implementation Overview:
I added 2 new tables, 'Permissions' and 'Role-Permissions'. The Permissions table contains permissions like 'add.shipment' and 'edit.orders'. The Role-Permissions table defines which roles have which permissions, it links roles (role_id) to permissions (permission_id).

Using these 2 tables, I updated the backend auth to include all permissions a user has (for all their roles) in the users session. Then I wrapped the Office app in a context provider that provides the user specific permissions with the intention that it will be consumed by the new `Restricted` component.  This component abstracts away all interactions with permissions and context into a nice, readable API. 

```javascript
	<Restricted to="edit.orders">
		// Children like OrderComponent will only be rendered if the user has permission to edit orders
		<OrdersComponent ... />  
	</Restricted>
```

The `Restricted` component can be used anywhere in the office app to limit access to child components without having to know anything about a users roles, we only need to know what permission a user should have to be granted access (which should be much more intuitive).

There is also an optional prop, `fallback`, to provide an alternative to be shown if the user does not have the required permission. By default nothing is shown when the user does not have access. 

### Important Note on Security: 
Frontend 'permissions' are just UX, they provide no real 'security'. The frontend code is run client side and could easily be modified to bypass any illusion of security. The actual security/protection must be enforced on the backend when API requests are made. We do want to use the same rules on both front and back end so the UI permits only what the backend allows.


PDickeyTruss:
A good analogy for anyone not super familiar with roles/permissions that has helped me in the past is to think about it in the context a school. There are several roles in a school setting: Student, Teacher, Principle, etc. If you were in charge of restricting access to somewhere in the school like the teachers lounge, you could ask each person coming in if they were either a teacher or a principal or a janitor or any other role that can access the teachers lounge. This is obviously tedious and with more roles and restrictions would be quite painful. An alternative way to restrict access to the teachers lounge would be to simply ask each person if they have permission to access teachers lounge. Assuming they are honest, this would mean I no longer have to care about the persons role (or any roles for that matter) and can instead get right to the point by asking if they have permission to do the action they are wanting to perform (enter the teachers lounge).

PR #8590, opened 2022-05-11
Added generated mock that was not in source control

reggieriser:
## Summary

I noticed when doing a clean build that we had one mock -- `CustomerSupportRemarksFetcher.go` -- that was generated but not in source control.


Ryan-Koch (approved):
Thanks for this PR, this was an oversight on my part

PR #8595, opened 2022-05-11
MB-12269 Add new PPMShipment fields to track actual final values

felipe-lee:
## Summary

We'll need new fields to track a PPM shipment's actual pickup and destinatination zips as well as whether an advance was received and how much it was. These need to be separate from the original fields the customer fills out during PPM onboarding.

This adds the new columns to the `ppm_shipments` table and the corresponding fields to the `PPMShipment` model. I'm open to any feedback, but am especially interested in making sure the names of the columns and the comments make sense.

I also had to fix a few issues in unrelated files that `golangci-lint` was flagging. Not sure why they started being issues, but hopefully once this merges it won't be an issue anymore.

## Verifying changes

<details>
<summary>💻</summary>

### Run migrations

```sh
make db_dev_migrate
```

</details>

### Check table for new fields

1. Access the DB however is best for you. I use the goland DB connection, but we also have other options documented
    1. [alternative UI options for viewing the DB](https://transcom.github.io/mymove-docs/docs/backend/setup/configure-postico-or-tableplus-to-connect-to-mymove-db)
1. Check the `ppm_shipments` table and verify the new columns

## Verification Steps for Author

These are to be checked by the author.

- [ ] Request review from a member of a different team.
- [ ] Have the Jira acceptance criteria been met for this change?

## Verification Steps for Reviewers

These are to be checked by a reviewer.

#### Any new migrations/schema changes:

- [ ] Follows our guidelines for [Zero-Downtime Deploys](https://transcom.github.io/mymove-docs/docs/dev/contributing/database/Database-Migrations#zero-downtime-migrations)
- [ ] Have been communicated to #g-database

PR #8604, opened 2022-05-12
MB-12270 Use new PPM zip and advance fields in service objects and API

felipe-lee:
## Summary

We need new fields to track the final/actual values for zip codes and advances once a customer has moved their PPM shipment. #8595 adds the new fields `PPMShipment` fields to the table and model. This PR wires up the BE to account for the new fields.

One thing I'd like feedback on is how y'all feel about the way I re-structured the `mergePPMShipment` tests. It combines a few patterns we have in the code already, but they aren't common patterns, and one bit isn't something I've seen in our code (though could of course be there somewhere). I was hoping to reduce duplication while still making it easy to define checks should run per each test case, but idk if in the end this is more confusing than just having separate sub-tests (with more duplication). I implemented a similar, but slightly different version for the `internalapi` `mto_shipment_test.go` file for the PPM update handler tests. Lmk if you like either one, and if so which. Saying neither is fine too and I can break them down into subtests.

<details>
<summary>Change Log</summary>

I'll mark these off as I go.

- [x] Add new fields to `mergePPMShipment`. I think this is really the only place in the service objects that needs to directly know about the fields.
- [x] Expand tests for `mergePPMShipment` to include more cases.
    - I also added test cases for adding/removing secondary zips and they pass so this isn't where the [secondary zip bug](https://dp3.atlassian.net/browse/MB-12249) is.
- [x] Update internal.yaml to use the shared PPMShipment.yaml.
- [x] Update shared `PPMShipment.yaml` to add new fields. Also update to define required fields and add missing `x-omitempty` attribute for a couple of fields.
- [x] Add new fields to `model_to_payload.go` and `payload_to_model.go` files for internal, ghc, and prime.
    - Only added to `payload_to_model.go` for internal because the other two don't have definitions yet for PPM updates.
- [x] Update handler tests to account for new fields
- [x] Add testdatagen function with new fields pre-filled. This will make testing the changes easier at this stage since we won't have the UI to fill them in yet.
- [x] Start writing to the fields that will replace the existing advance request fields.
    - This is step 2/6 of the renaming process
    - `AdvanceRequested` -> `HasRequestedAdvance`
    - `Advance` -> `AdvanceAmountRequested`

</details>

### Running Server Tests

<details>
<summary>💻</summary>

#### Run Tests

```sh
make server_test
```

</details>

### Looking at Data in Network Tab

<details>
<summary>💻 Setup</summary>

#### Terminal 1

Start the UI locally.

```sh
make client_run
```

#### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

#### Additional steps

1. Access the [local customer app](http://milmovelocal:3000/)
2. Open your dev tools and go to the network tab.
3. Login as a customer with a PPM shipment.
    1. Any of the ones that end in `@ppm.unsubmitted` or `@ppm.submitted` should work for this.
    2. `actualPPMDateZIPAdvanceDone@ppm.approved` will have a PPM shipment that has info in the new fields.
4. Check the `mto_shipments` request and look at the response data. Look for the PPM shipment and verify that you can see the 4 new fields in the response.
    1. You won't be able to see the info on the regular UI until [MB-12272](https://dp3.atlassian.net/browse/MB-12272) is completed.

### Backend


swagger-def/internal.yaml
```diff
@@ -2379,64 +2379,14 @@ definitions:
...
-    readOnly: true
-    enum:
-      - DRAFT
-      - SUBMITTED
-      - WAITING_ON_CUSTOMER
-      - NEEDS_ADVANCE_APPROVAL
-      - NEEDS_PAYMENT_APPROVAL
-      - PAYMENT_APPROVED
-  PPMShipment:
+  CreatePPMShipment:
```
felipe-lee, line 2382: Idk why the diff for this file looks so bad...I removed the `PPMShipmentStatus` and `PPMShipment` definitions in favor of the shared definitions in `swagger-def/definitions`

pkg/services/ppmshipment/validation_test.go
```diff
@@ -1,185 +1,469 @@
...
+		suite.Equal(oldShipment.SpouseProGearWeight, mergedShipment.SpouseProGearWeight)
+	}
+
+	mergeTestCases := map[string]struct {
+		oldState    PPMShipmentState
+		oldFlags    flags
+		newShipment models.PPMShipment
+		runChecks   runChecksFunc
+	}{
+		"Doesn't set invalid data for required fields": {
```
duncan-truss, line 108: Would this even be possible to get in this state given the swagger endpoint requirements?

pkg/services/ppmshipment/validation_test.go
```diff
@@ -1,185 +1,469 @@
...
+
+			// these should never change
+			suite.Equal(oldShipment.ID, mergedShipment.ID)
+			suite.Equal(oldShipment.ShipmentID, mergedShipment.ShipmentID)
+			suite.Equal(oldShipment.Status, mergedShipment.Status)
+
+			// now run test case specific checks
+			tc.runChecks(*mergedShipment, oldShipment, tc.newShipment)
+		})
+	}
```
duncan-truss, line 468: I think this is definitely an improvement in the coverage of mergePPMShipment's functionality.  I like the experiment and maybe it could be applied elsewhere to other models we have with merge functions.  I think it may become a little hairy to debug potentially if something did start to fail with all of the shared setup and common reuse of functions.

I think there are limits to this unit test though because of all of the pieces involved in the real request and how that may not be accurately what the data looks like at the time of calling mergePPMShipment.  Like I could see handler tests or Cypress tests failing and these still passing but it's not necessarily an either/or choice.

pkg/handlers/ghcapi/mto_shipment_test.go
```diff
@@ -2330,12 +2330,12 @@ func (suite *HandlerSuite) TestCreateMTOShipmentHandlerUsingPPM() {
 		if suite.NotNil(ppmPayload) {
```
duncan-truss, line 2330: I've seen this mistakenly used elsewhere, I don't think it works as a conditional because it will cause a failure while the assumption is it would skip this code block.

duncan-truss (approved):

pkg/handlers/ghcapi/mto_shipment_test.go
reggieriser: I believe I added this in an earlier PR (and have used something similar in a few other places).  My goal in using it as a conditional here is just to prevent a panic on the following lines if the `ppmPayload` does happen to be `nil` (since the `NotNil` assertion failing does not automatically stop the rest of the assertions from being checked).  Is there a better way to prevent the panic?  Or maybe we don't care if it panics as long as the test fails.

duncan-truss: You're right I was oversimplifying a bit.  I don't much care about the panic if there's also a test failure but maybe that's a good reason. There's always `fatalNil` too. I caught a bug in a test recently that had something like this so that's where my head was at.

reggieriser: I remember seeing our `FatalNil` helper but I think that function actually fails if the object is not nil, not the other way around (maybe it's poorly named).

pkg/services/ppmshipment/validation_test.go
felipe-lee: Regarding the debugging with shared setup and re-use of functions, when I had a few failures while setting it up, the traceback showed me each of the steps well enough to be able to trace which test case was the one failing. But, I do agree that it's not as clear as if it's all in a single function.

To your point about there being limits to what this can catch, you're totally right. One very good example of that is how I added a test case for being able to remove secondary zips. It works, but if you actually try that from the UI right now, that fails. It's a problem in the `payload_to_model` though (and thankfully should soon be addressed in #8638, I know you've seen that PR Duncan, mostly mentioning for anyone else seeing this thread). So there are limits. One thing I like about having both levels testing though, is if one fails, it's a lot faster to see where the issue likely is. I did actually add a test for the handler level, but it's not quite correct because we don't send empty strings from the FE, we send null. I was going to either fix that if Namibia's pr merged first, or ask her to update this test in hers. 
https://github.com/transcom/mymove/pull/8604/files#diff-0dc4f49ea27f9cbe2b9257753c04679e143009d5db990c8997cd4e0894aca26fR701-R713
Note that you may have to expand `pkg/handlers/internalapi/mto_shipment_test.go` to show the diff and then paste the link into your url bar again before it'll show you the lines. By default that file diff isn't show due to size.

felipe-lee: Also, a follow-up question, I implemented this test pattern in this `validation_test.go` and the `internalapi/mto_shipment_test.go`, with slight differences. In `validation_test.go`, I used the state and flags in a setup function to determine what the original shipment should look like, while in the `mto_shipment_test.go` I had the setup function be one of the test case fields. I did the different versions as an experiment to see which was better to work with. 

I find the one in `mto_shipment_test.go` nice in that there's a little less overhead imo for having to know what states/flags mean and being explicit for each test case, but it does lead to more verbose test case setup. I think it might be better to pick one or the other and update them to both follow the same pattern, but maybe it's fine as is.

Do you have a preference for either one? Or do you think it's fine to have both?

felipe-lee: I'm going to merge this to continue the renaming process. One of my PRs next week will also be touching these tests so I can align them then though so I'd still like your feedback on what you think about the different methods.

PR #8609, opened 2022-05-13
Rely on Cypress to get Assertions correct

rogeruiz:
This attempts to remove the .then() and rather rely on the Assertions
that Cypress uses to block, wait, and retry while things change state.

### DIFF
--- a/cypress/integration/office/servicescounseling/servicesCounselingFlows.js
+++ b/cypress/integration/office/servicescounseling/servicesCounselingFlows.js
@@ -184,11 +184,7 @@ describe('Services counselor user', () => {
       cy.get('input[name="dependentsAuthorized"]').siblings('label[for="dependentsAuthorizedInput"]').click();
 
       // Edit allowances page | Save
-      cy.get('button')
-        .contains('Save')
-        .should('be.enabled')
-        .click()
-        .then(() => cy.get('button').should('be.disabled'));
+      cy.get('button').contains('Save').should('be.enabled').click().should('be.disabled');
     });
 
     cy.wait(['@patchAllowances']);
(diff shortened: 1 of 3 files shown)


cypress/integration/office/servicescounseling/servicesCounselingFlows.js
```diff
@@ -184,11 +184,7 @@ describe('Services counselor user', () => {
       cy.get('input[name="dependentsAuthorized"]').siblings('label[for="dependentsAuthorizedInput"]').click();
 
       // Edit allowances page | Save
-      cy.get('button')
-        .contains('Save')
-        .should('be.enabled')
-        .click()
-        .then(() => cy.get('button').should('be.disabled'));
+      cy.get('[data-testid="scAllowancesSave"]').should('be.enabled').click().should('be.disabled');
```
rogeruiz, line 187: > from my commit message

This attempts to remove the .then() and rather rely on the Assertions
that Cypress uses to block, wait, and retry while things change state.

[=> Here's where I read this from the Cypress documentation.](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress#Assertions)

src/pages/Office/ServicesCounselingMoveAllowances/ServicesCounselingMoveAllowances.jsx
```diff
@@ -105,7 +105,7 @@ const ServicesCounselingMoveAllowances = () => {
       storageInTransit: Number(storageInTransit),
       organizationalClothingAndIndividualEquipment,
     };
-    mutateOrders({ orderID: orderId, ifMatchETag: order.eTag, body });
+    return mutateOrders({ orderID: orderId, ifMatchETag: order.eTag, body });
```
rogeruiz, line 108: > from my commit message

[Return `mutateOrders`] Which I believe is a Promise. In my research I saw that sometimes
isSubmitting is never changed which causes issues in our tests on
CircleCI but not locally. I stumbled across this GitHub Issue which
explained the problem pretty well. https://github.com/jaredpalmer/formik/issues/2442

jenniferlynparsons (approved):
This looks good!

Thought: This would be great to share out at the BE/FE check in or demo.

PR #8613, opened 2022-05-13
MB-12511: Remove usage of pkg/cli/vault and unused cmd/report-ecs

ahobson:
## Summary

The only uses of aws-vault as a dependency are in legacy parts of the code. Looks like we can nuke it use entirely

## Setup to Run Your Code

`make server_test`

I pushed this to experimental and was able to log in and browser around.


cmd/webhook-client/main.go
```diff
@@ -120,10 +121,8 @@ func InitRootConfig(v *viper.Viper) (*pop.Connection, *zap.Logger, error) {
 
 	var session *awssession.Session
 	if v.GetBool(cli.DbIamFlag) {
-		verbose := cli.LogLevelIsDebug(v)
```
gidjin, line 123: Question: Did you intend to delete the `verbose` variable that we used to pass to the AWSConfig? Or is there no equivalent?

pkg/cli/vault.go
```diff
@@ -1,164 +0,0 @@
-package cli
```
gidjin, line 1: Question: What was this file used for? 
I'm fine deleting things that have no purpose any longer just curious.

cmd/webhook-client/main.go
ahobson: I tried to follow the same pattern as in 2a642e62dd9f608174761d9d41ab66ed899c8450 and so I don't believe the verbose setting applies any more. Am I mistaken?

pkg/cli/vault.go
ahobson: I haven't been on the project long enough to know! 😉  As far as I can tell, we tried to build in some `aws-vault` functionality into the go code itself instead of relying on an external command.

cmd/webhook-client/main.go
rogeruiz: Yeah I see what you're doing here @ahobson. No `verbose` makes sense as we weren't using it for anything besides sending it to `GetAWSConfig()`.

rogeruiz (approved):
lgtm 🌈

Ran `make server_test` and started things up locally and everything was copacetic.

PR #8621, opened 2022-05-16
MB-11407 - Document when a TIO updates max billable weight

michaelvaldes-truss:
Since the TIO update to max billable weight may update two tables (entitlements for the weight update and moves for the tio_remarks update) there will be at least two rows in the move history table for this event and two corresponding event templates.

## Summary

After further investigation, changes to the acceptance criteria have been proposed, which should also be reviewed along with the feature itself.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the office app as a TIO
2. Navigate to a move with a payment request (e.g. locator = 'REWAYD')
3. Navigate to the 'Payment requests' tab and click the 'Review weights' button 
4. Edit the Max billable weight (e.g. reduce it by 1 lb), add a remark, and click 'Save changes'
5. Navigate to the 'Move history' tab
6. Verify that two rows are created and that they match the acceptance criteria of the story

### Frontend

## Screenshots
image


src/constants/moveHistoryEventTemplate.test.js
```diff
@@ -601,6 +603,34 @@ describe('moveHistoryEventTemplate', () => {
     });
   });
 
+  describe('when given an update billable weights as tio history record', () => {
+    const item = {
+      action: 'UPDATE',
+      changedValues: { authorized_weight: '7999' },
+      eventName: moveHistoryOperations.updateBillableWeightAsTIO,
+      tableName: 'entitlements',
```
YanZ777, line 611: **suggestion (non-blocking):** Lets use constants for the table names too here.

YanZ777 (approved):

src/constants/moveHistoryEventTemplate.test.js
```diff
@@ -601,6 +603,34 @@ describe('moveHistoryEventTemplate', () => {
     });
   });
 
+  describe('when given an update billable weights as tio history record', () => {
+    const item = {
+      action: 'UPDATE',
```
rogeruiz, line 608:
```suggestion
      action: dbActions.UPDATE,
```

src/constants/moveHistoryEventTemplate.test.js
rogeruiz: @YanZ777 does it make sense to do this for the `.test.js` file too? I noticed that the `historyLogUIDisplayName.js` module doesn't get imported into the main test file. I'm don't mind having these written out manually. But it's also possible that #8626 could address things like this so that we standardize on using the constants everywhere.

juliemoon (approved):
This looks good to me. I can see the authorized weight & max billable weight fields are updated. I've included a screen shot to document the ongoing conversation surrounding the detail column's Status: APPROVALS REQUESTED.
Screen Shot 2022-05-17 at 4 12 06 PM

src/constants/moveHistoryEventTemplate.test.js
YanZ777: I think we could go either way. I slightly lean towards using constants in the tests since I feel like those are not as important to decouple. Lets lump it in with #8626.

PR #8625, opened 2022-05-16
Convert remaining package tests to transactions

shimonacarvalho:
### Summary 

All packages that were found to clone the database converted, except those that needed substantial changes. Those were written into separate tickets and added to the epic. 

### Testing

Running all server tests should suffice

For more about the changes, please see https://transcom.github.io/mymove-docs/docs/backend/testing/running-server-tests-inside-a-transaction/

### DIFF
--- a/pkg/awardqueue/awardqueue.go
+++ b/pkg/awardqueue/awardqueue.go
@@ -97,13 +97,6 @@ func (aq *AwardQueue) assignPerformanceBandsForTSPPerformanceGroup(appCtx appcon
 	return nil
 }
 
-// waitForLock MUST be called within a transaction!
-func waitForLock(appCtx appcontext.AppContext, id int) error {
-
-	// obtain transaction-level advisory-lock
-	return appCtx.DB().RawQuery("SELECT pg_advisory_xact_lock($1)", id).Exec()
-}
-
 // NewAwardQueue creates a new AwardQueue
 func NewAwardQueue() *AwardQueue {
 	return &AwardQueue{}
--- a/pkg/migrate/migrate_test.go
+++ b/pkg/migrate/migrate_test.go
@@ -12,11 +12,12 @@ type MigrateSuite struct {
 	testingsuite.PopTestSuite
 }
 
+// This suite uses the high privileged role.
 func TestMigrateSuite(t *testing.T) {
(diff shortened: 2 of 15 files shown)


pkg/awardqueue/awardqueue.go
```diff
@@ -97,13 +97,6 @@ func (aq *AwardQueue) assignPerformanceBandsForTSPPerformanceGroup(appCtx appcon
 	return nil
 }
 
-// waitForLock MUST be called within a transaction!
-func waitForLock(appCtx appcontext.AppContext, id int) error {
-
-	// obtain transaction-level advisory-lock
-	return appCtx.DB().RawQuery("SELECT pg_advisory_xact_lock($1)", id).Exec()
-}
-
```
ahobson, line 106: I'm fine with this, but AFAICT, I don't think `AwardQueue` is even used anywhere any more. I wonder if we shouldn't remove it entirely?

pkg/migrate/migrate_test.go
```diff
@@ -12,11 +12,12 @@ type MigrateSuite struct {
 	testingsuite.PopTestSuite
 }
 
+// This suite uses the high privileged role.
 func TestMigrateSuite(t *testing.T) {
 	ms := &MigrateSuite{
 		PopTestSuite: testingsuite.NewPopTestSuite(
 			"migrate",
-			testingsuite.WithHighPrivPSQLRole(),
+			testingsuite.WithHighPrivPSQLRole(), // WithPerTestTransaction cannot be used in conjunction with the HighPrivPSQLRole
```
ahobson, line 20: Nice!

ahobson (approved):

pkg/awardqueue/awardqueue.go
shimonacarvalho: The whole package is unused? If so, I'll put in a ticket to remove entirely.

ahobson: I tried to search for usage of `AwardQueue` and I didn't find anything other than the award queue test

shimonacarvalho: Okay will create a separate pr to delete the whole pkg. Thanks!

PR #8631, opened 2022-05-17
[ MB-16007 ] Open Telemetry improvements for AWS XRay

ahobson:
## Summary

* Add some additional metrics for the db connection and database table size
* Configure otel logger and error handler so we can see errors
* Switch to the cumulative metric aggregator since the one we were using is no longer supported for many metrics
* Upgrade to a more recent aws-otel-collector
* If possible, add log annotation in AWS X-Ray format to allow for log correlation

## Setup to Run Your Code

Most of these changes are really only useful in AWS, and have been tested pretty extensively in the load test environment

- [x] Tested in the Loadtest environment (for changes to containers, app startup, or connection to data stores)


rogeruiz:
After talking to @ahobson about this, I would like to take this on to review this work with @transcom/truss-platform-team folks during Woo Team. Drew's happy to pair and help if we get stuck.

- This is sending more metrics to AWS 🎉.
- It would be nice to have our OpenTel Sidecar be deployed everywhere (lower priority).
  - We will eventually need to enable this in our ATO environments.
- The main goal of this OpenTel is to help debug things in our application logs.

go.mod
```diff
@@ -220,3 +221,6 @@ require (
 	pault.ag/go/othername v0.0.0-20190316144542-859caba4369b // indirect
 	pault.ag/go/piv v0.0.0-20190320181422-d9d61c70919c // indirect
 )
+
+// fix for https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3166
+replace go.opentelemetry.io/contrib/detectors/aws/ecs => github.com/ahobson/opentelemetry-go-contrib/detectors/aws/ecs v0.0.0-20230214135558-6055125933d0
```
rogeruiz, line 226: Do you think we should have this fork live under the Transcom organization rather than your user @ahobson? I'm hoping this is only temporary but your PR has been sitting around for about a month or so. 

It's not an issue for it to live on your fork. Another option I see is maybe having it live in the Trussworks organization instead 🤷🏼.

ahobson: > Do you think we should have this fork live under the Transcom organization rather than your user @ahobson? I'm hoping this is only temporary but your PR has been sitting around for about a month or so.
> 
> It's not an issue for it to live on your fork. Another option I see is maybe having it live in the Trussworks organization instead 🤷🏼.

That's a great point. I keep thinking they are going to merge it in

ahobson: @rogeruiz hey hey! It's merged! PR updated

LeDeep (approved):
Looks good. Thanks for the walkthrough, yesterday. That was really helpful.

PR #8636, opened 2022-05-18
Use testing.T#Cleanup to simplify running tests in transactions

ahobson:
## Summary

By using the [Cleanup](https://pkg.go.dev/testing#T.Cleanup) function whenever a database connection is established for per test transaction, we can ensure it is closed and the transaction is rolled back. We no longer need to convert suite.T().Run to suite.Run!

The docs for how to convert/write transaction tests needs to be updated if we like this approach

## Setup to Run Your Code

```sh
make server_test
```

### DIFF
--- a/pkg/handlers/supportapi/move_task_order_test.go
+++ b/pkg/handlers/supportapi/move_task_order_test.go
@@ -72,11 +72,10 @@ func (suite *HandlerSuite) TestHideNonFakeMoveTaskOrdersHandler() {
 	params := move_task_order.HideNonFakeMoveTaskOrdersParams{
 		HTTPRequest: request,
 	}
-	handlerConfig := handlers.NewHandlerConfig(suite.DB(), suite.Logger())
 
 	suite.Run("successfully hide fake moves", func() {
 		handler := HideNonFakeMoveTaskOrdersHandlerFunc{
-			handlerConfig,
+			handlers.NewHandlerConfig(suite.DB(), suite.Logger()),
 			movetaskorder.NewMoveTaskOrderHider(),
 		}
 		var moves models.Moves
@@ -111,7 +110,7 @@ func (suite *HandlerSuite) TestHideNonFakeMoveTaskOrdersHandler() {
 		}
 		mockHider := &mocks.MoveTaskOrderHider{}
 		handler := HideNonFakeMoveTaskOrdersHandlerFunc{
-			handlerConfig,
+			handlers.NewHandlerConfig(suite.DB(), suite.Logger()),
 			mockHider,
 		}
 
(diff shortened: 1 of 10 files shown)


pkg/testingsuite/pop_suite.go
```diff
@@ -388,17 +387,25 @@ func (suite *PopTestSuite) DB() *pop.Connection {
 	if suite.usePerTestTransaction {
 		if suite.lowPrivConn == nil {
 			suite.lowPrivConn = suite.openTxnPopConnection()
+
+			// Any time a database connection is established, make sure we
+			// close it at the end of the test so that the transaction
+			// is rolled back. See txdb for more info
+			suite.T().Cleanup(func() {
```
shimonacarvalho, line 394: Interesting so does this mean
* suite.Run → rolls back the transaction after every subtest
* suite.T().Run → rolls back the transaction after every test (but not every subtest) (similar to RunWithRollback

pkg/handlers/supportapi/move_task_order_test.go
```diff
@@ -72,11 +72,10 @@ func (suite *HandlerSuite) TestHideNonFakeMoveTaskOrdersHandler() {
 	params := move_task_order.HideNonFakeMoveTaskOrdersParams{
 		HTTPRequest: request,
 	}
-	handlerConfig := handlers.NewHandlerConfig(suite.DB(), suite.Logger())
 
 	suite.Run("successfully hide fake moves", func() {
 		handler := HideNonFakeMoveTaskOrdersHandlerFunc{
-			handlerConfig,
+			handlers.NewHandlerConfig(suite.DB(), suite.Logger()),
```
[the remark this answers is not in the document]
ahobson: Yes! If you notice, the `handlerConfig` was being initialized outside the `suite.Run`

pkg/testingsuite/pop_suite.go
ahobson: First, let's agree on terminology :)

```go
func (suite *MySuite) TestMyThing() {
  t1 := initThingOne()
  suite.Run("It should do something") {
    suite.NoError(t1.Doit())
  }
}
```

I think the terminology is:
1. `TestMyThing` is a test
2. `It should do something` is a subtest

The `Cleanup` function runs when the `*testing.T` scope completes. If you look at the code for [suite.Run](https://github.com/stretchr/testify/blob/master/suite/suite.go#L82) you can see it creates a new `*testing.T` for each subtest, so that means the transaction is rolled back for each subtest.

Note that it also possible to do ...

```go
func (suite *MyPopSuite) TestMyThing() {
  db := suite.DB()
  // do things with the db
}
```

In this case, there's a `*testing.T` scope for the test and so the transaction will be rolled back at the end of the test.

I guess it does mean that if someone does ...

```go
func (suite *MyPopSuite) TestMyThing() {
  db := suite.DB()
  suite.Run("It should do something") {
    subtestDb := suite.DB()
  }
}

```

It looks like in that situation, the rollback would happen at the end of the test and the subtest would share the db connection. That's because we only open a new connection if one doesn't exist.

I can think of some things we might do to ensure that each `*testing.T` gets its own DB connection. It would be a good safety measure to make sure tests aren't accidentally re-using connections.

Thank you for asking such good questions!

shimonacarvalho: First up I agree with your terminology for test and subtest.

Your description seems to cover what `suite.Run()` does. 

My question is, what does `suite.T().Run` do once you make this change? Because that's something that lingers in the code base.

PR #8626, opened 2022-05-16
[ MB-12515 ] Add ADR about Move History Events

rogeruiz:
## Summary

This ADR is for Move History Event Templates. It does a number of
things.

- Introduce the idea of an Event Template Object for Move History.
- Discuss the current state of things
- Propose solutions which split the Event Templates into ES6 modules
  which have a single test file and default export.

<!-- More to come in the future, but for now commented out

## Setup to Run Your Code

<details>
<summary>💻 You will need to use three separate terminals to test this locally.</summary>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

<!-- Fill out the next section as you see fit, these are just suggestions. 

1. Access the
2. Login as a
3.

## Verification Steps for Author

These are to be checked by the author.

- [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
- [ ] Request review from a member of a different team.
- [ ] Have the Jira acceptance criteria been met for this change?

## Verification Steps for Reviewers

These are to be checked by a reviewer.

<!-- Fill out the next sections as you see fit, these are just suggestions. 

### Frontend

- [ ] User facing changes have been reviewed by design.
- [ ] There are no aXe warnings for UI.
- [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
- [ ] There are no new console errors in the browser devtools
- [ ] There are no new console errors in the test output
- [ ] If this PR adds a new component to Storybook, it ensures the component is fully responsive, OR if it is intentionally not, a wrapping div using the `officeApp` class or custom `min-width` styling is used to hide any states the would not be visible to the user.

### Backend

- [ ] Code follows the guidelines for [Logging](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide#logging)
- [ ] The requirements listed in [Querying the Database Safely](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide/#querying-the-database-safely) have been satisfied.

#### Any new migrations/schema changes:

[...]

PR #8646, opened 2022-05-19
[ MB-12518 ] Refactoring of MoveHistory Events

rogeruiz:
## Summary

This large refactor is described in #8626. Please read through the ADR to understand the refactor. Notes will be provided near lines of code in this PR that augment the ADR

## Setup to Run Your Code

<details>
<summary>💻 You will need to use one terminal to test this locally.</summary>

##### Terminal 1

Start the Client tests locally.

```sh
make client_test
```

</details>

### Additional steps

1. Make sure all the tests for the client pass locally


src/constants/MoveHistory/EventTemplates/undefinedEvent.js
```diff
@@ -0,0 +1,34 @@
...
+      default:
+        return 'Updated move';
+    }
+  },
+  getDetailsPlainText: () => {
+    return '-';
+  },
+};
+
+export default () => buildTemplate(t);
```
rogeruiz, line 34: So it seems like Webpack is expecting the default export here to be a function? @michaelvaldes-truss and I noticed that in the last part of our pairing session.

src/constants/MoveHistory/TemplateManager.js
```diff
@@ -0,0 +1,59 @@
...
+    return (
+      propertiesMatch(eventType.action, other?.action) &&
+      propertiesMatch(eventType.eventName, other?.eventName) &&
+      propertiesMatch(eventType.tableName, other?.tableName)
+    );
+  };
+
+  allMoveHistoryEventTemplates.push(eventType);
+
+  return eventType;
```
michaelvaldes-truss, line 51: **suggestion (non-blocking):** with this refactor, can we remove the return here and rename this function to "registerTemplate" or something like that?

rogeruiz: Sure. I'm not working in this file. so please checkout the branch and add any changes here. I'm hoping for us to swarm on this as it's blocking all other tickets until it's done.

rogeruiz: Also this would be a great addition to the ADR around what we end up calling this in the Proposal section.

michaelvaldes-truss (approved):
Looks great, and testing passed on my machine!

PR #8651, opened 2022-05-20
MB-9097: Remove DPS 

ahobson:
## Summary

Remove dead code related to DPS.

Really this was grepping for DPS and SDDC and chasing the string. Lots of files changed, but not terribly complicated

## Setup to Run Your Code

```
make server_test
```


sheenamt:
@ahobson The things I can review (docker and env files) look good.

config/tls/devlocal-not-dps.key
```diff
@@ -1,28 +0,0 @@
------BEGIN PRIVATE KEY-----
```
reggieriser, line 1: **nitpick:** There's a reference to this file still in `.pre-commit-config.yaml`.

reggieriser (approved):
I started to review on behalf of DB reviewers, but went ahead and reviewed the whole PR (well, as much as makes sense for a big code pruning like this).  Things LGTM!  I did do a search for "dps" on the branch and found a few nitpicks that I added.  There's also a couple of references to `DPS users` in the top-level `README.md` which wasn't touched, so you may want to get rid of them while pruning.  But those aren't big deals, so going ahead and approving.  Thanks for taking care of this! 🚀

pkg/handlers/apitests.go
```diff
@@ -169,19 +169,7 @@ func (suite *BaseHandlerTestSuite) AuthenticateOfficeRequest(req *http.Request,
...
-		ApplicationName: auth.MilApp,
-		UserID:          serviceMember.UserID,
-		IDToken:         "fake token",
-		DpsUserID:       dpsUser.ID,
-	}
-	ctx := auth.SetSessionInRequestContext(req, &session)
-	return req.WithContext(ctx)
-}
-
-// AuthenticateAdminRequest authenticates DPS users
```
gidjin, line 184: Praise: Nice catch on the copy pasta comment

pkg/models/client_cert_test.go
```diff
@@ -21,13 +21,11 @@ func (suite *ModelSuite) Test_FetchClientCert() {
 	digest := "abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789"
 	subject := "/C=US/ST=DC/L=Washington/O=Test/OU=Test Cert/CN=localhost"
 	certNew := models.ClientCert{
-		Sha256Digest:    digest,
-		Subject:         subject,
-		AllowDpsAuthAPI: true,
-		AllowOrdersAPI:  true,
```
gidjin, line 27: Question: Did you intend to delete the `AllowOrdersAPI` attribute here?

gidjin (changes requested):
This looks good to me, I left one question for a line that looks like it shouldn't have been deleted so flagging this as requested changes since I'd like to make sure that line is correctly removed.

Second is a general question on this. Is there a plan/story to remove the environment vars from chamber after this PR is merged?

pkg/models/client_cert_test.go
ahobson: The test doesn't depend on the setting of any particular capability, so I thought I'd only add the minimum necessary to the model. Concerns with that approach?

ahobson:
> This looks good to me, I left one question for a line that looks like it shouldn't have been deleted so flagging this as requested changes since I'd like to make sure that line is correctly removed.

I responded to the question as I think the line removed in the test was not necessary for the test and was arguably confusing.

> Second is a general question on this. Is there a plan/story to remove the environment vars from chamber after this PR is merged?

We have a ticket with infrasec related changes, but thank you for flagging that! I'll ask

pkg/models/client_cert_test.go
gidjin: I missed that this was a test! Thank for answering

gidjin (approved):

PR #8658, opened 2022-05-21
MB-12491 Converts pkg/services/move and all its tests to use transactions

shimonacarvalho:
## Summary

Converts the package and tests in `./pkg/services/move` to use transactions.

## Setup to Run Your Code

## Verification Steps

- [x] Go's `testing` package is only imported in the setup test file. (In this case there was no separate setup file)


ahobson (approved):
👏 🚀

PR #8662, opened 2022-05-23
[ MB-9033 ] Update React-USWDS and USWDS;

rogeruiz:
## Summary

React-USWDS is updated to version 3.0.x.
USWDS is updated to version 2.11.2.


rogeruiz:
Here's a breakdown of the Yarn changes between this branch and what's currently on the default branch. https://gist.github.com/rogeruiz/368fcf98a39e845e808ad1d579c36b4a?permalink_comment_id=4178051#gistcomment-4178051

rogeruiz:
The application can now build with what is currently on this branch. 🎉 and also tests are passing using the `yarn test:debug:react-uswds` command as well locally thanks to @YanZ777. 🎉

src/components/ScrollToTop.jsx
```diff
@@ -15,8 +16,10 @@ const ScrollToTop = ({ otherDep = null }) => {
   return null;
 };
 
+const otherDepShape = PropTypes.any;
+
 ScrollToTop.propTypes = {
-  otherDep: null,
+  otherDep: otherDepShape,
```
duncan-truss, line 22: this tends to be an alert component passed here, I would think it could be `PropTypes.node` or `PropTypes.element`

rogeruiz: There are also times where it can be a `PropTypes.func` or `PropTypes.string` from the errors I found in the unit tests. Does it make sense to make this `PropTypes.oneOfType([/*...*/])` with all those different PropTypes?

duncan-truss: PropTypes.node is pretty pervasive, I think it covers anything renderable (strings, components, ...).  I think oneOfType of PropTypes.node and PropTypes.func might cover it.  I don't see any functions when I look at the usage but I didn't look that closely.

rogeruiz: ah thank you. that makes a lot of sense.

rogeruiz:
Now that I have a Dockerfile update, I'm going to continue iterating and testing on it. I think it makes sense to perhaps start this whole branch off from scratch locally with the update Node version to see if it resolves some of our dependency issues. I also believe that to update _react-uswds_ we also need to update _react-scripts_ to version `5.x` since that would bring in a lot of the updates we're doing here around Jest, Webpack, and Babel. But in a more measured and tested way. I hadn't thought of that approach last week.

rogeruiz:
Closing this as I've updated USWDS and React-USWDS across these two PRs #8713 and #8724.

PR #8673, opened 2022-05-24
MB-12447 Permissions Middleware POC

leia-sefkin:
## Summary

This is an initial pass at creating a permissions middleware, building off of previous implementations and research. 

Permissions are at their simplest level a string composed of `action.object` wherein the action is the approved action a user can take on a given object. Actions should follow standard rest CRUD terminology and both actions and objects should align with the endpoint in question. 

Examples:
- `update.move`
- `create.shipment`

Endpoints can have multiple permissions set on them, and a user must have all permissions listed in order to be granted access. Permissions should be defined on the [swagger-definition operation object](https://swagger.io/docs/specification/paths-and-operations/) as an array of `x-permissions` 

A user's role is already determined, so this middleware looks up the permissions already assigned to that role vs. the permissions needed on a given route in order to determine access. Permissions are added onto the session object and LoggedInUserPayload so that the FE can read these. 

At this time the definition of which permissions are assigned to which roles are stored inline at the top of `pkg/auth/authentication/permissions.go`, TBD is a better resting place for the data.

## Testing out this PR ##

- Run migrations, regen swagger files, run the server tests
- Run the server and client, login as TOO to approve a move - everything should go through successfully

## Next steps

- Determine what an appropriate store for permissions should be and update the implementation so that it's no longer hardcoded. Under consideration should be feasibility for maintenance and upkeep as we continue to define permissions in the application. 
- Clean up code base to use permissions middleware instead of checking in-line for roles

### ADR and previous implementations ###
- [ADR-0040](https://github.com/transcom/mymove/blob/c90ae337459b0dafaede81e8b684f59bf5aa1465/docs/adr/0040-role-base-authorization.md)
-  [Authorization POC](https://github.com/transcom/mymove/pull/2824/files#diff-be793c31d522efea61c667cf49ed11f865eb47751287243daf9d0a4383372eb4) 
- [Permissions POC](https://github.com/transcom/mymove/pull/8578/files#diff-b7c08fec7647d998d5244c3117fee9b0710e74dccb44c7cc1c33fd96cf41b16d) 

### External Resources ###

- [OWASP recommendation for implementing role based access control](https://owasp.org/www-project-proactive-controls/v3/en/c7-enforce-access-controls)
- [Swagger Authentication Documentation](https://swagger.io/docs/specification/2-0/authentication/): Built in support in swagger for auth
- [Swagger composed Auth](https://goswagger.io/tutorial/composed-auth/): how to do more advanced security configurations in Go-Swagger
- [A useful blog post about middleware in general](https://www.alexedwards.net/blog/making-and-using-middleware)
- [Go swagger recommendations for how to use middleware](https://goswagger.io/use/middleware.html)
- [How Go + OpenAPI + Swagger all work together to serve the API](https://eli.thegreenplace.net/2021/rest-servers-in-go-part-4-using-openapi-and-swagger/)


pkg/auth/authentication/auth_test.go
```diff
@@ -246,6 +250,108 @@ func (suite *AuthSuite) TestRequireAuthMiddleware() {
 	suite.Equal(handlerSession.UserID, user.ID, "the authenticated user is different from expected")
 }
 
+// Test permissions middleware with a user who will be ALLOWED POST access on the endpoint: ghc/v1/shipments/:shipmentID/approve
+// role must have update.shipment permissions
+func (suite *AuthSuite) TestRequirePermissionsMiddlewareAuthorized() {
```
leia-sefkin, line 255: These two tests are basically carbon copies of each other, but trying to make the setup of the router and middleware more generic so as to be reusable proved a bit of a tail-chasing exercise.

pkg/auth/authentication/permissions.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/transcom/mymove/pkg/auth"
+
+	"github.com/gofrs/uuid"
+
+	"github.com/transcom/mymove/pkg/models/roles"
+)
+
+// TODO: placeholder until we figure out where these should be stored
+type RolePermissions struct {
```
leia-sefkin, line 17: this was easiest for just getting this thing up and running - open to input on a better longer term solution here.

swagger-def/internal.yaml
```diff
@@ -41,6 +41,10 @@ definitions:
         items:
           $ref: '#/definitions/Role'
         x-nullable: true
+      permissions:
```
leia-sefkin, line 44: this adds permissions to the logged in user payload so that the frontend has access

pkg/handlers/internalapi/users.go
```diff
@@ -141,6 +148,7 @@ func (h ShowLoggedInUserHandler) Handle(params userop.ShowLoggedInUserParams) mi
 				Email:         appCtx.Session().Email,
 			}
 			decoratePayloadWithRoles(appCtx.Session(), &userPayload)
+			decoratePayloadWithPermissions(appCtx.Session(), &userPayload)
```
leia-sefkin, line 151: adding in the permissions data to the payload so the FE can access

cmd/milmove/serve.go
```diff
@@ -21,13 +21,14 @@ import (
...
+	"github.com/gobuffalo/pop/v6"
+
 	"github.com/alexedwards/scs/redisstore"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	awssession "github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/sts"
-	"github.com/gobuffalo/pop/v6"
 	"github.com/golang-jwt/jwt/v4"
```
PDickeyTruss, line 32: Very minor nit: This import looks lonely up above the others 😁

swagger-def/ghc.yaml
```diff
@@ -795,6 +795,8 @@ paths:
       description: Approves a shipment
       operationId: approveShipment
       summary: Approves a shipment
+      x-permissions:
+        - update.shipment
```
PDickeyTruss, line 799: Really like how clean and readable this ends up being.... very clear what permissions are needed to use each endpoint.

PDickeyTruss (approved):
LGTM - All works as expected wrt restricting access based on permissions. Very nice that we can add it piecewise as endpoints without x-permissions defined are not restricted. I also see the appropriate permissions coming across in the session for the FE to make use of. Approved! 🎉

cmd/milmove/serve.go
leia-sefkin: hah weird! i'll put it back

pkg/auth/authentication/permissions.go
```diff
@@ -0,0 +1,106 @@
...
+
+var QAECSR = RolePermissions{
+	RoleType:    roles.RoleTypeQaeCsr,
+	Permissions: []string{"read.move"},
+}
+
+var AllRolesPermissions = []RolePermissions{TOO, TIO, QAECSR}
+
+// check if a [user.role] has permissions on a given object
+func checkUserPermission(appCtx appcontext.AppContext, session *auth.Session, permission string) (bool, error) {
```
gidjin, line 41: Question: Is there an easy way to add unit tests for these functions?

swagger/ghc.yaml
```diff
@@ -2074,6 +2079,9 @@ paths:
           $ref: '#/responses/PermissionDenied'
         '500':
           $ref: '#/responses/ServerError'
+      x-permissions:
+        - update.move
+        - create.serviceItem
```
gidjin, line 2084: Question: Does this cover all the api's that need permission or only the ones related to your current story/outcome? If the latter are there stories for updating other endpoints with the correct permissions?

swagger-def/ghc.yaml
gidjin: Praise: I agree with @PDickeyTruss on the read-ability in the yaml files

gidjin (approved):
I left a few questions and a couple suggestions, but this looks good to me otherwise. I tested it out locally too everything seemed fine and I saw the permissions returned by the login response.

pkg/auth/authentication/permissions.go
Ryan-Koch: I agree with @gidjin on the longer term. I think this is good for now, especially since we're still in the experiment phase of doing this. Later I suspect we'll want to follow up with something like moving it to the DB (eventually it'd be nice to be able to manage these in the admin app).

Ryan-Koch (approved):

pkg/auth/authentication/auth.go
```diff
@@ -54,6 +56,62 @@ func IsLoggedInMiddleware(globalLogger *zap.Logger) http.HandlerFunc {
...
+
+			route, r, _ := api.Context().RouteInfo(r)
+			if route == nil {
+				// If we reach this error, something went wrong with the swagger router initialization, in reality will probably never be an issue except potentially in local testing
+				logger.Error("Route not found on permission lookup")
+				http.Error(w, http.StatusText(400), http.StatusBadRequest)
+				return
+			}
+
+			permissionsRequired, exists := route.Operation.VendorExtensible.Extensions["x-permissions"]
```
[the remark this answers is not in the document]
leia-sefkin: @gidjin - yes it is something we came up with, but the `x-` format is standard for swagger extensions, so it's a bit of a yes AND answer to your question. Good call on documenting on the developer portal site, I'll make sure to add a chore for info about how this all works there. cc @Ryan-Koch

swagger/ghc.yaml
leia-sefkin: @gidjin so far it has only been the APIs that are covered in our current outcome which is adding in functionality for QAE/CSR users. We talked in planning that we would need to identify stories for cleaning up other endpoints and role checks throughout the code - but I don't think those have been written out yet. cc @Ryan-Koch

pkg/auth/authentication/auth.go
```diff
@@ -54,6 +56,62 @@ func IsLoggedInMiddleware(globalLogger *zap.Logger) http.HandlerFunc {
...
+	return func(next http.Handler) http.Handler {
+		mw := func(w http.ResponseWriter, r *http.Request) {
+
+			logger := logging.FromContext(r.Context())
+			session := auth.SessionFromRequestContext(r)
+
+			route, r, _ := api.Context().RouteInfo(r)
+			if route == nil {
+				// If we reach this error, something went wrong with the swagger router initialization, in reality will probably never be an issue except potentially in local testing
+				logger.Error("Route not found on permission lookup")
```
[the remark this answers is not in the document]
leia-sefkin: hah i mostly ran into it when i was trying to figure out how the go-server and swagger-api initialization works. A painful learning process indeed 😂

pkg/auth/authentication/auth.go
```diff
@@ -625,6 +683,7 @@ var authorizeKnownUser = func(appCtx appcontext.AppContext, userIdentity *models
 		return
 	}
 	appCtx.Session().Roles = append(appCtx.Session().Roles, userIdentity.Roles...)
+	appCtx.Session().Permissions, _ = getPermissionsForUser(appCtx, userIdentity.ID)
```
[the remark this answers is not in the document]
leia-sefkin: 🤔 good question - i'm pretty sure the main ramifications would be that a user may not have access to the components or endpoints that they would expect. If there's an error the Permissions set should be empty. I'm fairly sure any errors should get logged out before here - but I'll double check!

duncan-truss:
I tried a few quick experiments to see if I could get a shared YAML definition to share across APIs but ran into some limitations I think.  Maybe someone can work some magic with `$ref`s that I'm unaware of but in each API file at least we could define all the permissions only once and use aliases in the x-permissions block or something simliar:

```yaml
x-endpoint-permissions:
    - &updateMove update.move 
paths:
...
    x-permissions:
        - *updateMove
```

swagger-def/internal.yaml
```diff
@@ -41,6 +41,11 @@ definitions:
         items:
           $ref: '#/definitions/Role'
         x-nullable: true
+      # Here we are returning permissions on the user session to the FE, rather than requiring them on the endpoint. This is why this is permissions rather than x-permissions
+      permissions:
+        type: array
+        items:
+          type: string
```
duncan-truss, line 48: Could we define an string enum type here maybe similar to how affiliation is defined in definitions/Affiliation.yaml?

cmd/milmove/serve.go
```diff
@@ -992,10 +992,13 @@ func serveFunction(cmd *cobra.Command, args []string) error {
 		}
 
 		// Mux for GHC API that enforces auth
+		api := ghcapi.NewGhcAPIHandler(handlerConfig)
 		ghcAPIMux := ghcMux.PathPrefix("/").Subrouter()
 		ghcAPIMux.Use(userAuthMiddleware)
 		ghcAPIMux.Use(middleware.NoCache(logger))
-		api := ghcapi.NewGhcAPIHandler(handlerConfig)
+
+		permissionsMiddelware := authentication.PermissionsMiddleware(appCtx, api)
```
ahobson, line 1000: Tiny typo here? `permissionsMiddleware` ?

pkg/auth/authentication/permissions.go
ahobson: How will this scheme support situations where a service member can see their own move but no one else's, while an office worker may be able to see moves related to e.g. their GBLOC and another office worker can see all moves?

Using the [ABAC](https://en.wikipedia.org/wiki/Attribute-based_access_control) model, you generally need to provide the subject, action, object and context

leia-sefkin:
> I tried a few quick experiments to see if I could get a shared YAML permissions definition to share across APIs but ran into some limitations I think. Maybe someone can work some magic with `$ref`s that I'm unaware of. In each API file at least we could define all the permissions only once and use aliases in the x-permissions block or something simliar:
> 
> ```yaml
> x-endpoint-permissions:
>     - &updateMove update.move 
> paths:
> ...
>     x-permissions:
>         - *updateMove
> ```

Thanks for looking at this @duncan-truss - i went with operation level here rather than endpoint since i wanted to focus more granularly for now, but it may be the case it makes more sense to move these up to the global endpoint level once we get more defined and start to see the patterns

swagger-def/internal.yaml
leia-sefkin: 🤔 we could, i'm a little wary of doing so right now as it would mean another place to list the current permission definitions and keep them in sync with each other. i'll add a note under the current comment to list that as a TODO as we keep working on this

cmd/milmove/serve.go
leia-sefkin: how DID YOU CATCH THAT? zomg thanks for finding what my dyslexia brain always glosses over

PR #8676, opened 2022-05-25
MB-12592 Convert ghcapi tests

shimonacarvalho:
## Summary

Converts all tests in `./pkg/handlers/ghcapi` to use transactions.

## Setup to Run Your Code

## Verification Steps

- [x] Go's `testing` package is only imported in the setup test file. (In this case there was no separate setup file)


felipe-lee:
@shimonacarvalho can you say more about this:

> There is no unjustified usage of suite.AppContextForTest().DB() - instead suite.DB() is used directly

I thought we wanted to be using the DB from the appctx

shimonacarvalho:
> @shimonacarvalho can you say more about this:
> 
> > There is no unjustified usage of suite.AppContextForTest().DB() - instead suite.DB() is used directly
> 
> I thought we wanted to be using the DB from the appctx

Been using the same PR text that Sandy came up with. I believe in the tests, it makes no difference within the test and suite.DB() is cleaner. We definitely want to use suite.AppContextForTest if we need an appcontext, though.

felipe-lee:
> > @shimonacarvalho can you say more about this:
> > > There is no unjustified usage of suite.AppContextForTest().DB() - instead suite.DB() is used directly
> > 
> > 
> > I thought we wanted to be using the DB from the appctx
> 
> Been using the same PR text that Sandy came up with. I believe in the tests, it makes no difference within the test and suite.DB() is cleaner. We definitely want to use suite.AppContextForTest if we need an appcontext, though.

Ah ok, could you add that to the docs? At least, I don't think I saw it. I did find this: https://transcom.github.io/mymove-docs/docs/backend/guides/use-stateless-services-with-app-context#testing-with-appcontext
but that's actually a bit misleading since we don't actually have to add that per test suite since we add it at the top level for all test suites.

shimonacarvalho:
> Ah ok, could you add that to the docs? At least, I don't think I saw it. I did find this: https://transcom.github.io/mymove-docs/docs/backend/guides/use-stateless-services-with-app-context#testing-with-appcontext
> but that's actually a bit misleading since we don't actually have to add that per test suite since we add it at the top level for all test suites.

I could add to docs but I think we realized over time that it's pretty rare we've seen that so not that important of a pattern to correct. Just might as well watch out for it as we are cleaning up. 

That doc though does need to be updated because it seems to refer to TestAppContext? When really that function (I _think_) is now called `AppContextForTest`. Not part of this ticket I think but I'll try to get that in soon.

shimonacarvalho:
@felipe-lee anything further you'd like fixed on this pr?

PR #8688, opened 2022-05-26
MB-11344 Added lbs to prime actual weight

kennethchow-truss:
## Summary

Prime actual weight (labeled as shipment weight) was missing lbs. Now lbs is added.


juliemoon (approved):
Thanks for getting this in!

donaldthai (approved):
Looks good. 

FYI, looks like there's a failing client_test?

PR #8690, opened 2022-05-26
MB-9033 Fix Tests

YanZ777:
## [Jira ticket](https://dp3.atlassian.net/browse/MB-9033) for this change

[This gist](https://gist.github.com/rogeruiz/8d8f827c8b0182c1da7540f36a5d770d) has a summary of failing tests that need to be fixed


package.json
```diff
@@ -143,6 +144,7 @@
     "cypress-wait-until": "^1.7.2",
```
rogeruiz, line 144: Looks like Danger is getting triggered to mention Cypress because this is included in the Diff, but it's not modified. We need to fix how Danger is checking for file differences better. Not necessarily in this ticket, but I think it would be a quick fix after researching how to better check for actual file changes in Danger's documentation.

rogeruiz: It looks like we can leverage `JSONDiffValue` rather than `diffForFile` to check for what is added to the diff to better ensure if any of the `cypressPackageNames` have been modified.

https://danger.systems/js/reference.html#JSONDiffValue

rogeruiz:
Relates to #8662

PR #8711, opened 2022-06-01
MB-12619 FE permissions setup

PDickeyTruss:
## Summary

This PR builds on @leia-sefkin's permissions middleware [PR](https://github.com/transcom/mymove/pull/8673/files). It adds the frontend components and setup needed to hide or modify what is rendered in the office app based on the current users permissions. It also includes use of the permission restriction components/pattern to hide the `Flag for financial review` button for users without permission to flag for review (QAE/CSR users). 

### Restricting access
To hide or modify the component that should be rendered based on the current users permissions, simply wrap the component in the new `Restricted` component:
```javascript
<Restricted to="permission">
    <SomeComponent />
</Restricted />
```
The `to` prop is the permission required to see the child of the restricted component. Optionally, a `fallback` prop can be passed to provide control over what should be shown when a user does not have the required permissions.

```javascript

<Restricted to={permissionTypes.updateShipment} fallback={<Button>View Shipment</Button>}>
    <Button>Edit Shipment</Button>
</Restricted />
```
### How it works? 
I added a `PermissionsProvider` component to the office index.jsx file. This component wraps the office app in a context provider and gets passed in the user permissions (from the session) as a prop. The value of this context is an object with a 'isAllowedTo' function. This function can be used to check a given permission string against the permissions that were passed to the provider, which is exactly what the `Restricted` component does to conditionally render children or null (of a fallback). The `Restricted` component is the consumer of the context setup by the `PermisionsProvider`.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Validation Steps
- [ ] As a TXO user nothing in the office app is changed, especially with the `Flag for financial review` functionality. 
- [ ] As a QAE/CSR user the `Flag for financial review` button is not shown.
- [ ] The pattern for hiding/modifying what renders based on user permission is reasonable.

https://user-images.githubusercontent.com/62263329/171727843-57cda8a9-9a61-4f57-a0bc-b0440f07141d.mov


pkg/auth/authentication/permissions.go
```diff
@@ -0,0 +1,111 @@
...
+// TODO: placeholder until we figure out where these should be stored
+type RolePermissions struct {
+	RoleType    roles.RoleType
+	Permissions []string
+}
+
+var TOO = RolePermissions{
+	RoleType: roles.RoleTypeTOO,
+	Permissions: []string{"update.move", "create.serviceItem",
+		"update.shipment", "update.financial_review_flag"},
```
leia-sefkin, line 25: should we use snake case or camel? this new flag is in snake, but our old ones camel. i'm open either way - if you have a preference for snake i can update the other flags

PDickeyTruss: Fixed ✅

leia-sefkin (approved):
Looks great, overall nice clean logic we can use and extend as needed for future components as we work through the rest of the QAE/CSR outcome. No other specific feedback here since we've already been working together and this is a nice evolution of the previous PR. 

I ran through and tested locally and everything was working as expected. I'd say 🚢 :it: (provided we can get another independent review)

src/components/Restricted/Restricted.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import PermissionContext from './PermissionContext';
+
+// Render the children if the user is allowed to access the permission, otherwise render the fallback or null.
+// This component is meant to be used everywhere a restriction based on user permission is needed
+const Restricted = ({ to, fallback, children }) => {
```
Ryan-Koch, line 8: I really like the use of fallback here empowering folks an easy choice between rendering something else or nothing at all.

PR #8722, opened 2022-06-03
[MB-12446] search for move by customer name

mboren:
## Summary
I added a customer name option to the search on the QAE/CSR landing page.
I also refactored out the search inputs into their own component and put it in storybook.

I don't have a lot of confidence that I'm making the right choice for removing diacritics, and I would love more feedback about this.
I laid out a lot of the details over here: https://dp3.atlassian.net/wiki/spaces/MT/pages/1735557148/Discovery+Searching+for+customers+by+name
the tl;dr is:
- postgres has an [unaccent](https://www.postgresql.org/docs/current/unaccent.html
) function that's great, but we can't use it in an index because it is not `IMMUTABLE` (it depends on a file, so the same arguments to the function don't necessarily result in the same return value)
- lots of people recommend wrapping `unaccent` in your own function and declaring that function `IMMUTABLE`. This feels sketchy to me and would mean that we would have to remember to rebuild this index every time the `unaccent` module gets updated)
- the main alternative to this is rolling your own function to do roughly the same thing (eg [example using python](https://wiki.postgresql.org/wiki/Strip_accents_from_strings), [example using TRANSLATE](https://wiki.postgresql.org/wiki/Strip_accents_from_strings,_and_output_in_lowercase)). I kind of like the simplicity of the `TRANSLATE` approach, and I think it'll handle the vast majority of our cases while being very stable over time. There are lots of things it won't handle (eg you can put an unbounded number of diacritics on a character, this only handles a single one), but for these cases we can hopefully rely on our search being fuzzy.

- We could also save the unaccented full name in its own column and do this operation in `go` code, but I would prefer to keep this in the database so that we don't have to worry about keeping different name columns in sync.

This SO answer explains why we can't use unaccent easily and discusses wrapping it: https://stackoverflow.com/a/11007216

## Setup to Run Your Code
You will need at least one customer with a move to test that this feature works.
To test whether it works well, you'll need a lot of customers, ideally with realistic names.
If you have load testing set up, you can use that to build up a bigger dataset
```sh
# in milmove_load_testing
locust -f locustfiles/queue.py --host local
```
but if you don't, you can trust that I tested with a bunch of people / or try it when it goes to staging

1. Log in as any office user other than QAE/CSR and find a move to search for by customer name
2. Log in to the office app as a QAE/CSR user.
3. Select "Customer name", search for the name, and verify that the move you picked is in the results.

### Frontend

### Backend

## Screenshots
https://www.figma.com/file/5AhuEkaH3bJVrtffU3wp22/Milmove-Office-App
image


reggieriser (commented):
Took a look at the migration and the query and it looks good from what I can tell (and based on what I read in the linked articles).  I think I like the wrapper function approach a little better (at least for now) than rolling our own function.  Thanks for all the references as they helped me gain context.

I did notice that when I searched for a name like "Ready Finish" on the E2E data that the exact match was last of the three results that matched.  Looks like the front-end is ordering the results by status (ascending).  I would assume if searching by name we'd prefer to have the top match first since presumably that's what the user is looking for, then let the user sort in the table UI from there if there were multiples.

Ryan-Koch (approved):

reggieriser (approved):
LGTM!  Thanks for making the changes to the default sorting.  I mainly looked at the database and back-end code, but did try it out and the UI looks good now (searching for "Ready Finish" now leads to the exact match showing up first).

cosburn (approved):
Looks great!!

PR #8713, opened 2022-06-01
[ MB-9033 ] Update node v16.15.0

rogeruiz:
## Summary

Update Node to `16.15.0` from `14.18.1`. This work is being done in anticipation of updating _react-uswds_ to version `3.0.1`. That upgrade is being tracked in #8662 separately.

## Setup to Run Your Code

Load up the application locally and make sure that it builds. I way of doing this is by running `make clean build` which will clean out your local dependencies and then build the client and server. This should exit cleanly without errors. You may also fire up the application with `make client_run` and `make server_run` in separate terminals to ensure that the application is running. Lastly, run `make client_test` and `make e2e_test_fresh` to ensure the testing commands still run after this Node update.
## Verification Steps for Author

These are to be checked by the author.

- [x] Request review from a member of a different team.
- [x] Have the Jira acceptance criteria been met for this change?

See https://github.com/transcom/circleci-docker/pull/267 for updates to the underlying Docker image changes.

PR #8724, opened 2022-06-03
[ MB-9033 ] Update react-scripts v5.0.1, Storybook 6.5.8, USWDS 2.13.1, & React-USWDS 3.0.2

rogeruiz:
## Summary

This patch updates _react-scripts_ to version `v5.0.1` in order to update the
build tool chain for this repository. This include various changes to our build
and test system and updates _Webpack_ to version 5 among other changes.

## Testing this code

Checkout this branch and run `yarn install --force` to bring in the latest dependencies. All of the tests in CI/CD must pass in Circle and GitHub as well. You can verify the application is running locally as well. Also, all test commands for the client application must pass as well. 

- `yarn test` or `make client_test`
- `yarn storybook` or `make storybook`
- `yarn lint:production`
- `make e2e_test_fresh e2e_test_docker`


rogeruiz:
The client test failures happening here seem related to the new regression tests around code coverage introduced in #12600. I'll have to investigate a bit more tomorrow I may be wrong here.

rogeruiz:
On second thought it doesn't seem like that's the case. Something else is happening here.

rogeruiz:
Now that I've pulled in `master`, I've realized that now the Client Test Coverage is causing me to be stuck. Will tackle this more tomorrow, but this is now ready to merge in after #8713 but we'll need to address the test coverage issue 😄 
https://app.circleci.com/pipelines/github/transcom/mymove/42713/workflows/25ffe84a-2935-4055-96ba-a5ce0f6449e7/jobs/668667

rogeruiz:
@juliemoon good news. I don't believe there are any more Happo diffs beyond the ones in #8713 even though we are on USWDS 2.13.1. This means less Happo diffs to focus on. 😌

package.json
```diff
@@ -92,12 +92,12 @@
...
   "scripts": {
     "analyze": "source-map-explorer 'build/static/js/*.js'",
-    "build": "react-scripts build",
-    "eject": "react-scripts eject",
-    "start": "react-scripts start",
-    "test": "react-scripts test --env=jsdom",
-    "test:debug": "react-scripts --inspect-brk test --runInBand --env=jsdom",
-    "test:coverage": "react-scripts test --coverage --testResultsProcessor=jest-junit  --env=jsdom --watchAll=false --coverageDirectory=coverage --runInBand",
+    "build": "react-app-rewired build",
+    "eject": "react-app-rewired eject",
```
YanZ777, line 96: **suggestion:** It looks like `react-app-rewired` explicitly says not to change the eject script here in their documentation.
[https://github.com/timarney/react-app-rewired](https://github.com/timarney/react-app-rewired)

YanZ777, line 96:
```suggestion
    "eject": "react-scripts eject",
```

rogeruiz:
Had to rebase and force push due to conflicts in the `yarn.lock` file

rogeruiz:
@juliemoon there aren't any actual design changes in this PR. The files that I modified here are importing other libraries due to how **webpack** 5 works. Take a look at the Happo diff on this but otherwise, you should be able to give us the Approval without needing to dig too deep. These updates caused me to update a large number of files and our `CODEOWNERS` file had those listened under Design Review.

eslint-plugin-ato/no-unapproved-annotation.js
```diff
@@ -44,6 +44,7 @@ const approvedBypassableRules = new Set([
   'import/no-named-as-default',
   'import/order',
   'import/prefer-default-export',
+  'import/no-relative-packages',
```
rogeruiz, line 47: @transcom/truss-is3 I had to modify this in order to get the project to compile. I don't see any reason why this would inherently unsafe.

src/ghc_index.scss
```diff
@@ -1,5 +1,5 @@
 @import './shared/styles/settings';
-@import 'uswds';
+@import '../node_modules/uswds/dist/scss/uswds.scss';
```
rogeruiz, line 2: This is the reason why I needed to disable `import/no-relative-packages`. As mentioned, I didn't see it written anywhere that this would be unsafe. My opinion here is that it could cause the project to be able to use relative paths to import things, but that is mostly a stylistic change.

I could see relative paths being a concern about stability. Where a relative path doesn't exist in a given environment. But I would assume our automated and manual testing would catch issues like that.

rogeruiz (commented):
Helpful comments for the @transcom/truss-is3 review.

juliemoon (approved):

src/ghc_index.scss
Ryan-Koch: I appreciate the comment here pointing out the reason for disabling the rule. I agree with your assessment that this is more about stability than security. My experience with this sort of thing aligns very much with yours with it being about not breaking deploys.

That consistency is quite useful and important in those cases though. I suspect you've had to venture away from that pattern here because of the need for a path out into `node_modules.` It might be possible to use the `~` operator to get around this, which could allow us to keep the rule. It looks like that's what we're doing with the line below: `@import '~@trussworks/react-uswds/lib/index.css';` (I may also be missing something).

rogeruiz: I attempted to use that operator but _USWDS_ does not export this path and _webpack_ 5 is strict about importing only the items in a dependencies `exports` stanza in their `package.json`. 

https://github.com/uswds/uswds/blob/974f976c85e06c37ca9641b68f308864136bc3f9/package.json#L21

I'll give it another try though to see if it works. I may have been doing it on a version prior to `2.13.3` which does include that directory in its exports.

ahobson: AFAICT, `@import '~uswds/scss/uswds.scss';` works for me when I run `make client_build`

rogeruiz: Strange. So if I run `yarn lint:production`, I'm seeing this error.
```
>_  yarn lint:production
yarn run v1.22.11
$ NODE_ENV=production yarn lint
$ eslint --ext .js,.jsx --max-warnings=0 src

/Users/rsr/Developer/transcom/mymove/src/pages/Office/index.jsx
  9:8  error  Unable to resolve path to module '~uswds/dist/css/uswds.css'  import/no-unresolved

✖ 1 problem (1 error, 0 warnings)
```

here's the diff of what I have changed locally:

```diff
diff --git a/src/pages/Office/index.jsx b/src/pages/Office/index.jsx
index 4e382c0fe9..67ea8ac3cb 100644
--- a/src/pages/Office/index.jsx
+++ b/src/pages/Office/index.jsx
@@ -6,7 +6,7 @@ import { connect } from 'react-redux';
 import classnames from 'classnames';
 
 import styles from './Office.module.scss';
-import 'uswds/dist/css/uswds.css';
+import '~uswds/dist/css/uswds.css';
 import 'scenes/Office/office.scss';
 
 // API / Redux actions
```

rogeruiz: I also realized that I started this thread on the wrong line number 🤦🏽

src/pages/Office/index.jsx
```diff
@@ -6,7 +6,8 @@ import { connect } from 'react-redux';
 import classnames from 'classnames';
 
 import styles from './Office.module.scss';
-import 'uswds/dist/css/uswds.css';
+// eslint-disable-next-line import/no-relative-packages
+import '../../../node_modules/uswds/dist/css/uswds.css';
```
rogeruiz, line 10: Its this import @Ryan-Koch and @ahobson that cannot be resolved using the ~ tilde character. It breaks the client_build and lint commands.

Ryan-Koch (approved):
Giving the IS3 👍

src/ghc_index.scss
ahobson: Ok, I have some changes that allow the app to compile for me locally that don't require the relative imports AFAICT.

What would be the best way to share them? Would you like a PR to your PR? A gist diff? Another branch?

Broadly, it means adding the settings below before importing the `uswds.scss` file.

```
$theme-font-path: '~uswds/fonts';
$theme-image-path: '~uswds/img';
```

PR #8734, opened 2022-06-07
Mb 12273 ppm add cypress tests testdatagen nt

NamibiaTorres:
## Summary
This ticket adds two cypress tests for the About page:
1. The first flow tests when an advance has been received
2. The second flow tests when an advance has not been received

## Setup to Run Your Code

<details>

##### Terminal 1

Start the server locally.

```sh
make server_run
```

##### Terminal 3

Run cypress

```sh
yarn test:e2e
```

</details>

### Additional steps

1. Access the file ppmshipment/about.js
3. Watch cypress run and see if the tests pass

### Frontend

### Backend


cypress/support/ppmCustomerShared.js
```diff
@@ -32,6 +47,38 @@ export function navigateFromHomePageToReviewPage(isMoveSubmitted = false) {
...
+
+export function navigateToAboutPageWithoutAdvance() {
+  cy.get('button[data-testid="button"]').contains('Upload PPM Documents').click();
+  cy.get('input[name="actualMoveDate"]').clear().type('01 Feb 2022').blur();
+  cy.get('input[name="actualPickupPostalCode"]').clear().type('90210').blur();
+
+  cy.get('input[name="actualDestinationPostalCode"]').clear().type('76127');
+  cy.get('input[name="hasReceivedAdvance"][value="false"]').check({ force: true });
+  cy.get('button').contains('Save & Continue').should('be.enabled').click();
+}
```
felipe-lee, line 69: **thought**: I think we can split navigation from the filling out of data on the page. That would follow the pattern of the other tests a bit more. For example, the `submitsEstimatedWeightsAndProGear` works by the parent navigating to the page first, then the function fills out the info on the page and checks that the Save button isn't disabled, and then it uses the `navigateFromEstimatedWeightsPageToEstimatedIncentivePage` function to submit the page and go to the next page. 

So that would mean something like having 3 functions:

* `navigateFromHomePageToAboutPage`
    * Waits for the `@getShipment` call to finish
    * Checks to make sure the upload button is there and clicks it
    * Checks to make sure the new location is the expected one (i.e. check the url to see we're on the about page)
* `submitAboutPage`
    * This could take a parameter that it uses to determine if advance should be set or not and use an if statement for the different paths.
    * Fills out the page
    * Checks that the Save button isn't disabled
* `navigateFromAboutPageToWeightTicketsPage`
    * Clicks the submit button on the about page
    * Waits for the `@patchShipment` call to finish
    * Checks that the new location is the weight tickets page

Then you can either have the `about.js` file do the sign in and call each of these in order, or you can have a fourth function that does that. You could also lump the sign in to be in that first navigation function, but would still need something to then call the three functions in order.

The reasoning for this is that we'll re-use the two navigation functions when we set up the weight ticket cypress tests, but we don't really need to re-fill out the data. In fact may not want it fill out the data again to because we may have set up the data how we want it in the testdatagen.

cypress/support/ppmCustomerShared.js
```diff
@@ -20,6 +20,21 @@ export function signInAndNavigateFromHomePageToReviewPage(userId, isMoveSubmitte
...
+  navigateToAboutPageWithAdvance(isMoveSubmitted);
+}
+
+export function signInAndNavigateToAboutPageWithoutAdvance(userId, isMoveSubmitted = true) {
+  cy.apiSignInAsUser(userId);
+
+  cy.wait('@getShipment');
+
+  navigateToAboutPageWithoutAdvance(isMoveSubmitted);
+}
```
felipe-lee, line 37: **thought**: Both of these take `isMoveSubmitted` and pass it along but the functions they pass it to don't take args so it's not doing anything. I think it was probably left over from using another one of the functions in this file as a template for the new ones.

**todo**: Since I don't think it's needed, can you take out this param?

cypress/integration/milmove/ppms/about.js
```diff
@@ -0,0 +1,87 @@
...
+describe('About Your PPM - With Advances', function () {
+  before(() => {
+    cy.prepareCustomerApp();
+  });
+
+  beforeEach(() => {
+    cy.intercept('GET', '**/internal/moves/**/mto_shipments').as('getShipment');
+    cy.intercept('PATCH', '**/internal/mto-shipments/**').as('patchShipment');
+    cy.intercept('DELETE', '**/internal/mto-shipments/**').as('deleteShipment');
+    cy.intercept('GET', '**/internal/moves/**/signed_certifications').as('signedCertifications');
```
felipe-lee, line 17: **todo**: We aren't going to hit either of these endpoints so they can be removed from the intercepts.

cypress/integration/milmove/ppms/about.js
```diff
@@ -0,0 +1,87 @@
...
+describe('About Your PPM - Without Advances', function () {
+  before(() => {
+    cy.prepareCustomerApp();
+  });
+
+  beforeEach(() => {
+    cy.intercept('GET', '**/internal/moves/**/mto_shipments').as('getShipment');
+    cy.intercept('PATCH', '**/internal/mto-shipments/**').as('patchShipment');
+    cy.intercept('DELETE', '**/internal/mto-shipments/**').as('deleteShipment');
+    cy.intercept('GET', '**/internal/moves/**/signed_certifications').as('signedCertifications');
```
felipe-lee, line 72: **todo**: We aren't going to hit either of these endpoints so they can be removed from the intercepts.

cypress/integration/milmove/ppms/about.js
```diff
@@ -0,0 +1,87 @@
...
+  });
+  // Get another user for desktop
+  const viewportType = [
+    { viewport: 'desktop', isMobile: false, userId: 'c28b2eb1-975f-49f7-b8a3-c7377c0da908' }, // readyToFinish2@ppm.approved
+    { viewport: 'mobile', isMobile: true, userId: 'c28b2eb1-975f-49f7-b8a3-c7377c0da908' }, // readyToFinish2@ppm.approved
```
felipe-lee, line 22: **question**: the `userId` for these is the same but the comment about which email is tied to them is different. Did you mean to use the same user for both of these, or did you mean to use different users? I think either is probably fine for these tests, but just need to align the `userId`s and the emails to make sure they match each other.

**suggestion**: If you do intend to use the same `userId` for both, you can also pull it out from this object and define it in the `it` block below. That would also mean you could combine the two `describe` blocks and re-use the `viewportType` since the second `describe` block is also using the same `userId` for its tests. So something like:

```js
const viewportType = [
    { viewport: 'desktop', isMobile: false}
    { viewport: 'mobile', isMobile: true}
]

viewportType.forEach(({ viewport, isMobile }) => {
    it(`can upload with an advance - ${viewport}`, () => {
      if (isMobile) {
        setMobileViewport();
      }

     // readyToFinish2@ppm.approved
     const userId = 'c28b2eb1-975f-49f7-b8a3-c7377c0da908'

      signInAndNavigateToAboutPageWithAdvance(userId);
    });

   it(`can upload without an advance - ${viewport}`, () => {
      if (isMobile) {
        setMobileViewport();
      }

     // readyToFinish@ppm.approved
     const userId = 'cde987a1-a717-4a61-98b5-1f05e2e0844d'

      signInAndNavigateToAboutPageWithoutAdvance(userId);
    });
});
```

cypress/support/ppmCustomerShared.js
```diff
@@ -32,6 +47,38 @@ export function navigateFromHomePageToReviewPage(isMoveSubmitted = false) {
...
+  cy.get('dt').contains('Weight allowance');
+  cy.get('button').contains('Upload PPM Documents').click();
+  cy.get('input[name="hasReceivedAdvance"][value="false"]').check({ force: false });
+  // cy.get('input[name="advanceAmountReceived"]').clear().type('5000');
+  cy.get('button').contains('Save & Continue').should('be.enabled').click();
+  cy.location().should((loc) => {
+    expect(loc.pathname).to.match(/^\/moves\/[^/]+\/shipments\/[^/]+\/weight-tickets/);
+  });
+}
+
```
felipe-lee, line 81: **todo**: since this isn't being used (for others reading this, we'd talked about it in a chat), it can be removed.

felipe-lee (commented):
Thanks for putting this together! I left a few comments. Tbh, I don't think any of my comments will necessarily help with the flaky tests, but will help once we get to weight tickets.

It looks like the latest flaky test can't find the "Upload PPM Documents" text, so maybe if that gets taken out and it's just grabbing by the `data-testid` and clicking, it'll work? I just don't like that personally because then we won't match what the user sees...but I guess if that's what makes `cypress` happy, 🤷🏼

cypress/support/ppmCustomerShared.js
NamibiaTorres: Hi @felipe-lee! Thanks for your feedback and suggestions. I applied and push most of them. WOrking on a workout for the datatestid suggestion. The problem with the datatestid is that both buttons on the page have an id of "button" so the only way to tell them apart was to identify what the user sees. Maybe there is another way around this?

pearl-truss: @NamibiaTorres a workaround could be to use `first` and or `last`. Here's some documentation on [first[(https://docs.cypress.io/api/commands/first#Usage) and [last](https://docs.cypress.io/api/commands/last#Usage)

NamibiaTorres: @pearl-truss , I forgot to followup, looks like cypress really likes it when I add the button name that the user sees. But the workaround was instead to have circle load the whole page, and then look for the button (it found it then).

felipe-lee (commented):
Overall, LGTM! The only reason I didn't approve yet is I want to be sure we are purposely changing the config. If we are, I think it might be good to document why and possibly get someone else to look at it.

.circleci/config.yml
```diff
@@ -942,8 +942,8 @@ jobs:
       - announce_failure
 
   integration_tests_milmove:
-    # parallelism requires enough files to divide by the # of parallel threads. So you need at least 6 files here
-    parallelism: 6
+    # parallelism requires enough files to divide by the # of parallel threads. So you need at least 4 files here
+    parallelism: 4
```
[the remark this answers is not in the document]
NamibiaTorres: I thought I pushed up this change, I'll change push it up again.

cypress/support/ppmCustomerShared.js
```diff
@@ -32,6 +55,34 @@ export function navigateFromHomePageToReviewPage(isMoveSubmitted = false) {
   }
 }
 
+export function fillOutAboutPageWithAdvance() {
+  cy.get('input[name="actualMoveDate"]').clear().type('01 Feb 2022').blur();
+  cy.get('input[name="actualPickupPostalCode"]').clear().type('90210').blur();
+
+  cy.get('input[name="actualDestinationPostalCode"]').clear().type('76127');
+  cy.get('input[name="hasReceivedAdvance"][value="true"]').check({ force: true });
+  cy.get('input[name="advanceAmountReceived"]').clear().type('5000');
```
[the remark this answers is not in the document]
NamibiaTorres: Makes sense, it's a good refactor idea. I'll implement it.

cypress/support/ppmCustomerShared.js
```diff
@@ -29,20 +29,22 @@ export function signInAndNavigateToAboutPageWithAdvance(userId, isMoveSubmitted
...
+//
+//   cy.wait('@getShipment');
+//   cy.screenshot();
+//   cy.get('button[data-testid="button"]').contains('Upload PPM Documents').click();
+//   cy.location().should((loc) => {
+//     expect(loc.pathname).to.match(/^\/moves\/[^/]+\/shipments\/[^/]+\/about/);
+//   });
+//   cy.get('input[name="hasReceivedAdvance"][value="false"]').check({ force: true });
+//   fillOutAboutPage();
+// }
```
felipe-lee, line 47: **todo**: Can you remove the commented out bits?

felipe-lee (approved):
LGTM, just left a comment about cleaning up the commented out code

cypress/support/ppmCustomerShared.js
NamibiaTorres: And I merged before the change to remove the comments got merged... I'll create a new PR to clean this up.

PR #8758, opened 2022-06-13
[MB-1200] PPM closeout Weight Tickets page component

duncan-truss:
## Summary

This PR creates the WeightTicket form and page components for the customer's PPM closeout flow.

The page component is incomplete and has some placeholder methods to work around not having fully implemented the data APIs on the backend or the client side work in Redux.  **Note if you are testing the page component it updates the URL and may behave weirdly on reload because the weight ticket is not really created**

We're limited in how we can test uploads in our jest unit tests so I had to work around that limitation.  I have implemented mocks in Storybook though that should give you a real sense of how the form component is working.

I have added the weight estimator test files to the repo under the `pkg/testdatagen` folder so we can use the xls and xlsx version in testing going forward.

[Ephemeral testing environment deploy](https://my-milmove-pr-8758.mymove.sandbox.truss.coffee/)

[Storybook link for this PR](https://output.circle-artifacts.com/output/job/cf3f0271-5098-4b8b-8b7e-dd58bbfbdca5/artifacts/0/storybook/index.html?path=/story/customer-components-ppm-closeout-weight-ticket-form--blank)

[Zeplin designs](https://zpl.io/Am9ndZ6)

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Login as the readyToFinish@ppm.approved user to the customer app
2. Click on the Upload PPM Documents button for the PPM Shipment
3. Proceed from the About Your PPM screen to the Weight Tickets page
4. Observe that the page loads the form and updates the URL with the fake created weight ticket record.
5. Upload functionality will fail because without a real weight ticket we can't attach uploads to document ids at the moment.

### Frontend

## Screenshots

Screenshot 2022-06-28 at 11-01-55 customer-components-ppm-closeout-weight-ticket-form--trailer-ownership
Screenshot 2022-06-28 at 11-01-08 customer-components-ppm-closeout-weight-ticket-form--trailer-ownership
Screen Shot 2022-06-28 at 11 07 21 AM


src/components/Customer/PPM/Closeout/WeightTicketForm/WeightTicketForm.jsx
```diff
@@ -0,0 +1,397 @@
...
+    hasOwnTrailer: hasOwnTrailer === true ? 'true' : 'false',
+    hasClaimedTrailer: hasClaimedTrailer === true ? 'true' : 'false',
+    trailerOwnershipDocs: trailerOwnershipDocs || [],
+  };
+
+  const constructedWeightDownload = (
+    <>
+      <p>Download the official government spreadsheet to calculate constructed weight.</p>
+      <Link
+        className={classnames('usa-button', 'usa-button--secondary', styles.constructedWeightLink)}
```
duncan-truss, line 102: This semantically should be a link even though we style it to be a button 🙃

src/components/Customer/PPM/Closeout/WeightTicketForm/WeightTicketForm.jsx
```diff
@@ -0,0 +1,397 @@
...
+    return showConstructedWeight ? 'Upload constructed weight spreadsheet' : 'Upload full weight ticket';
+  };
+
+  const weightTicketUploadHint = (showConstructedWeight) => {
+    return showConstructedWeight ? SpreadsheetUploadInstructions : DocumentAndImageUploadInstructions;
+  };
+
+  const emptyWeightTicketsRef = createRef();
+  const fullWeightTicketsRef = createRef();
+  const trailerOwnershipDocsRef = createRef();
```
duncan-truss, line 129: We need to create references to the underlying FileUpload elements so that we can call the removeFile() method after an upload succeeds to not have it sit in the container.

src/components/Customer/PPM/Closeout/WeightTicketForm/WeightTicketForm.jsx
```diff
@@ -0,0 +1,397 @@
...
+                <CheckboxField
+                  id="missingEmptyWeightTicket"
+                  name="missingEmptyWeightTicket"
+                  label="I don't have this weight ticket"
+                />
+                <div>
+                  {values.missingEmptyWeightTicket && constructedWeightDownload}
+                  <UploadsTable
+                    className={styles.uploadsTable}
+                    uploads={values.emptyWeightTickets}
```
duncan-truss, line 164: This differs from how orders maintains their uploads, they are always requerying for the orders model in the API and then re-rendering the entire component props wholesale.

Here I'm giving us the option of just modifying the form values and not causing a form rerender from the parent's props changing.

src/components/FileUpload/FileUpload.jsx
```diff
@@ -37,8 +37,8 @@ const FileUpload = forwardRef(
     },
     ref,
   ) => {
-    const handleOnChange = () => {
-      if (onChange) onChange();
+    const handleOnChange = (err, file) => {
```
duncan-truss, line 40: We need to pass back the file upload so that we can set the values of the field as well as remove it from the file pond component after successfully adding it to the uploads table.

src/components/UploadsTable/UploadsTable.module.scss
```diff
@@ -50,6 +50,9 @@
 
 .fileInfo {
   @include u-padding-left(2);
+  word-break: normal;
```
duncan-truss, line 53: Always good to test with long file names for wrapping behavior in the mobile layout.  Should fix it for display on the orders pages too.

src/pages/MyMove/PPM/Closeout/WeightTickets/WeightTickets.jsx
```diff
@@ -0,0 +1,102 @@
...
+
+const WeightTickets = () => {
+  const [errorMessage, setErrorMessage] = useState();
+
+  const history = useHistory();
+  const { moveId, mtoShipmentId } = useParams();
+  const { search } = useLocation();
+  const dispatch = useDispatch();
+
+  const { tripNumber } = qs.parse(search);
```
duncan-truss, line 28: Because the trip number is derived from the creation order (and could change if trips are deleted) and we're not passing props into the component we'll likely need to pull this value from the URL query param `?tripNumber=2`

src/components/Customer/PPM/Closeout/WeightTicketForm/WeightTicketForm.jsx
```diff
@@ -0,0 +1,438 @@
...
+        <FileUpload
+          name={fieldName}
+          labelIdle={UploadDropZoneLabel}
+          createUpload={(file) => onCreateUpload(fieldName, file)}
+          onChange={(err, upload) => {
+            setFieldTouched(fieldName, true);
+            onUploadComplete(upload, err, fieldName, values, setFieldValue);
+            fileUploadRef.current.removeFile(upload.id);
+          }}
+          acceptedFileTypes={acceptableFileTypes}
```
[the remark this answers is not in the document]
duncan-truss: The guidance I heard from @pmohal and @scarequotes was to not restrict per upload input if the user selects I'm missing a weight ticket or not.  I believe we still want to guide them with this labeling is my understanding. I think in the scenario where they previously uploaded a constructed weight excel file and come back to the form, they could still upload an image without having to worry about unchecking that box first.

But yea to your point the labeling indicates more restrictions than we enforce with the accepted file types we validate.

src/content/uploads.js
```diff
@@ -0,0 +1,6 @@
+export const DocumentAndImageUploadInstructions =
+  'PDF, JPG, or PNG only. Maximum file size 25 MB. Each page must be clear and legible.';
+
+export const SpreadsheetUploadInstructions = 'XLS or XLSX only. Maximum file size 25 MB.';
+
+export const UploadDropZoneLabel = 'Drag files here or <span class="filepond--label-action">choose from folder</span>';
```
pearl-truss, line 6: Concern: I noticed that this label language appears regardless of if the screen is mobile or not. The FileUpload component has a `labelIdleMobile` prop and logic to display it on mobile screens. 

From the designs it should say 'Upload files' on mobile

Screen Shot 2022-06-28 at 2 12 19 PM


```
labelIdleMobile: '<span class="filepond--label-action">Upload</span>'
```

src/components/Customer/PPM/Closeout/WeightTicketForm/WeightTicketForm.jsx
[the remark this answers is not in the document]
pearl-truss: I found in the designs that the label should change for constructed weight (and I'm assuming the accepted file types should also change)

Screen Shot 2022-06-28 at 2 13 54 PM

src/content/uploads.js
duncan-truss: Thank you I forgot that, let me override the mobile label defaults to add files.

I think it does show the right desktop or mobile label but it only calculates it initially on load based on the viewport and doesn't do any sort of responsive calculation if you resize the window.  I would set the window width and do a hard refresh to see the different labels.

pearl-truss: FilePond has made some choices! I'm glad that it's working locally since I think that matters more (and fwiw I did the initial code review just with storybook)

pearl-truss (approved):
changes look good

src/components/Customer/PPM/Closeout/WeightTicketForm/WeightTicketForm.test.jsx
```diff
@@ -0,0 +1,416 @@
...
+      await userEvent.upload(uploadEmptyWeight, emptyWeightFile);
+
+      expect(uploadEmptyWeight.files[0]).toBe(emptyWeightFile);
+      expect(uploadEmptyWeight.files.item(0)).toBe(emptyWeightFile);
+      expect(uploadEmptyWeight.files).toHaveLength(1);
+
+      const uploadFullWeight = uploadFileHints[1].nextSibling.firstChild;
+
+      const fullWeightFile = new File(['full weight'], 'full weight.png', { type: 'image/png' });
+      await userEvent.upload(uploadFullWeight, fullWeightFile);
```
felipe-lee, line 315: **question**: Since this doesn't work, should we keep the commented out code? Like is this something you think will eventually work? I think it is helpful to have the comment about why we're not uploading files as part of the test, but not sure if the code should stick around.

src/types/shipment.js
```diff
@@ -109,6 +110,24 @@ export const ShipmentShape = shape({
...
+  missingEmptyWeightTicket: bool,
+  emptyWeight: number,
+  emptyWeightDocumentId: string,
+  emptyWeightTickets: ExistingUploadsShape,
+  fullWeight: number,
+  missingFullWeightTicket: bool,
+  fullWeightDocumentId: string,
+  fullWeightTickets: ExistingUploadsShape,
+  hasOwnTrailer: bool,
+  hasClaimedTrailer: bool,
```
felipe-lee, line 126: **suggestion**: `hasClaimedTrailer` might be misleading because we ask them if their trailer meets the criteria in order to be claimable, and one of the disqualifying criteria is if they've claimed the trailer on another shipment or trip already. The field that will likely come back from the API is `trailerMeetsCriteria`

src/pages/MyMove/PPM/Closeout/WeightTickets/WeightTickets.test.jsx
```diff
@@ -0,0 +1,241 @@
...
+    ).toBeInTheDocument();
+    expect(
+      screen.getByText('Weight tickets must be certified, legible, and unaltered. Files must be 25MB or smaller.'),
+    ).toBeInTheDocument();
+    expect(screen.getByText('You must upload at least one set of weight tickets to get paid for your PPM.'));
+
+    // renders form content
+    expect(screen.getByRole('heading', { level: 2 })).toHaveTextContent('Trip 2');
+  });
+
```
felipe-lee, line 171: **suggestion**: I know I mentioned in another comment that we might not want to create the ticket on page load but rather on navigation from the about page, but if we stick with the current implementation, I think it might be good to have a test that checks that we don't try to create a weight ticket on page load if we already have an id for the weight ticket.

felipe-lee:
Ah, I meant to put in my review comment that none of my comments or questions are blocking, but mainly because some will maybe be easier to clear out once we do the full integration piece with the BE changes in place.

src/pages/MyMove/PPM/Closeout/WeightTickets/WeightTickets.jsx
```diff
@@ -0,0 +1,175 @@
...
+
+  // TODO remove when replaced by Redux call
+  const [weightTicket, setWeightTicket] = useState();
+
+  const mtoShipment = useSelector((state) => selectMTOShipmentById(state, mtoShipmentId));
+  // TODO add selector for selecting weight ticket from Redux store when data changes are solidified
+
+  useEffect(() => {
+    if (!weightTicketId) {
+      createWeightTicket(mtoShipmentId)
```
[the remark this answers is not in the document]
duncan-truss: I think we said we'd preemptively create the database record but I don't think that means we have to do it on those prior screens that lead into this component.  Remember you would need that review add an item screen to be creating all of the different types of closeout docs in addition to the About Your PPM page knowing how to create and store weight tickets.

I think this is more self contained honestly and doesn't bleed responsibility into the other page components and you don't need to go searching for behavior elsewhere in the flow.

src/components/Customer/PPM/Closeout/WeightTicketForm/WeightTicketForm.test.jsx
duncan-truss: Yea I guess I wanted you all to see the attempt and what didn't work.  Honestly leaving it might save someone trying to write the test again for a new upload screen or trying to re-add the test here.  I can't say I really got to the bottom of it because I feared it was a limitation and time sink so there's hope someone might find the time to run with it.

src/components/Customer/PPM/Closeout/WeightTicketForm/WeightTicketForm.jsx
```diff
@@ -0,0 +1,441 @@
...
+  fieldName: string.isRequired,
+  missingWeightTicket: bool,
+  onCreateUpload: func.isRequired,
+  onUploadComplete: func.isRequired,
+  onUploadDelete: func.isRequired,
+  fileUploadRef: shape({ current: shape({}) }).isRequired,
+  values: shape({}).isRequired,
+  formikProps: shape({
+    touched: shape({}),
+    errors: shape({}),
```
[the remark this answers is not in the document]
duncan-truss: The object prop type is disallowed by eslint and this is the workaround to have it not be flagged.

src/types/shipment.js
duncan-truss: Thanks for the suggestion, I didn't love that name either.

src/pages/MyMove/PPM/Closeout/WeightTickets/WeightTickets.jsx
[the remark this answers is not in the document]
ronaktruss: This seems like a reasonable and potentially better approach. 🥳 

**Thought(out of scope)**: Curious how this interacts in the scenario when they click back to get to the about page and then click `Save & Continue` I assume it'll create a new weight ticket. I think we should probably add some logic for how the about page handles routing. I think editing the about page from the review page may also require some additional logic wrt routing from the about page.

src/components/Customer/PPM/Closeout/WeightTicketForm/WeightTicketForm.jsx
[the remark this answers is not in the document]
felipe-lee: Ah, didn't know that, thanks!

scarequotes (approved):
Looks good!

src/pages/MyMove/PPM/Closeout/WeightTickets/WeightTickets.jsx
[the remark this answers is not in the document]
felipe-lee: Ah good point about the Add an item page potentially needing to do the same. Sounds good.

For the case you're saying Ronak, maybe the about page should route to the add an item page if there are any weight tickets on the PPM shipment already.

src/components/Customer/PPM/Closeout/WeightTicketForm/WeightTicketForm.test.jsx
felipe-lee: Ah ok, that's a good point, maybe we'll have time in the future to circle back to this and it'll be good to know what was attempted.

src/pages/MyMove/PPM/Closeout/WeightTickets/WeightTickets.jsx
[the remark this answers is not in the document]
duncan-truss: Yea we'll have to address that behavior with the About Your PPM routing anyway when they are coming back to edit their information from the Review page.  Like you said we can either derive the state from the data that the component has access too (although we may have to load more data than we want).  Another way might be looking in the router history, if it exists, to know where to send them to next.

felipe-lee (approved):
Thanks for making those updates, and for putting this all together!

PR #8766, opened 2022-06-14
MB-12001 FileUpload component to accept file types as a prop

pearl-truss:
## Summary

This PR refactored the fileupload component to have an acceptedFileTypes prop so that this component can be used with different file types in other places throughout the app.

This PR was tested in Chrome, FF and Safari 

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Run using storybook
2. Navigate to the Components/FileUpload stories
3. Upload an excel spreadsheet 

## Screenshots
Screen Shot 2022-06-14 at 1 39 26 PM

### DIFF
--- a/src/components/FileUpload/FileUpload.jsx
+++ b/src/components/FileUpload/FileUpload.jsx
@@ -1,5 +1,5 @@
 import React, { forwardRef } from 'react';
-import PropTypes from 'prop-types';
+import { func, string, arrayOf } from 'prop-types';
 import isMobile from 'is-mobile';
 import { FilePond, registerPlugin } from 'react-filepond';
 import 'filepond-polyfill/dist/filepond-polyfill';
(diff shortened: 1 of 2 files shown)


src/components/FileUpload/FileUpload.jsx
```diff
@@ -20,80 +20,90 @@ registerPlugin(
...
-    maxFileSize: '25MB',
-    credits: false,
-  };
+    /**
+     * Default FilePond instance props
+     * If these need to be overwritten, they can be exposed as a prop on this
+     * component and passed through (like labelIdle)
+     */
+    const filePondProps = {
+      allowMultiple: true,
```
duncan-truss, line 67: Do you mind making this property configurable while you're making changes here? I think Ronak is thinking that would simplify upload document creation if the user can't upload multiple docs in parallel. https://ustcdp3.slack.com/archives/C0250Q8469K/p1655310264319049

duncan-truss: Whoops I think that `allowMultiple` property would limit us to 1 upload total not force them to upload one file at a time.  It's a little ambiguous but we probably also need `maxParralelUploads`, which could let us create the first document record on the first upload and associate it with the second upload as well if they are done synchronously. https://pqina.nl/filepond/docs/api/instance/properties/

felipe-lee: Would we want to set that here directly? It kind of sounds like from this other thread that we'd want it to be configurable: https://ustcdp3.slack.com/archives/CNTDW9Q1Z/p1655314135153309

duncan-truss: No disagreement here, I just commented where this current value is hardcoded and the filepond component props are currently organized.

pearl-truss: Thanks for the context everyone! I'll see if I can get the 1 upload at a time to work and be configurable

pearl-truss: @ronaktruss were you thinking something like this?

Screen Shot 2022-06-16 at 11 20 18 AM

src/components/FileUpload/FileUpload.jsx
```diff
@@ -20,80 +20,108 @@ registerPlugin(
...
-  createUpload: PropTypes.func,
-  onChange: PropTypes.func,
-  onAddFile: PropTypes.func,
+  name: string,
+  createUpload: func,
+  onChange: func,
+  onAddFile: func,
+  acceptedFileTypes: arrayOf(string),
+  allowMultiple: bool,
+  maxParralelUploads: int,
```
ronaktruss, line 111: I think this should be number. We would also need to update the import
```suggestion
  maxParralelUploads: number,
```
```suggestion
  maxParralelUploads: int,
```

src/components/FileUpload/FileUpload.stories.jsx
```diff
@@ -8,13 +8,45 @@ export default {
...
+FileUploadWithLimitOneFile.args = {
+  createUpload: mockCreateUploadSuccess,
+  allowMultiple: false,
+};
+
+export const FileUploadWithNoParallelUploads = Template.bind({});
+FileUploadWithNoParallelUploads.args = {
+  createUpload: mockCreateUploadSuccess,
+  maxParallelUploads: 1,
+};
```
ronaktruss, line 47: I didn't have any luck with using the maxParallelUploads argument. 

In the ideal scenario we would want the first `createUpload` make an API call to create a new document ID and then have state be updated to capture this change. We'd want the second instance of `createUpload` call to then based on state determine if we have a document ID and then do some other logic.

More generically we want to do something the first time and then update state and then do something different the second time. 

Even when I change `maxParallelUploads` to 1 I am not able to achieve this. Have you been able to get this working?

ronaktruss, line 47: I didn't have any luck with using the maxParallelUploads argument. 

In the ideal scenario we would want the first `createUpload` make an API call to create a new document ID and then have state be updated to capture this change. We'd want the second instance of `createUpload` call to then based on state determine if we have a document ID and then do some other logic.

More generically we want to do something the first time and then update state and then do something different the second time. 

Even when I change `maxParallelUploads` to 1 I am not able to achieve this. Have you been able to get this working?

duncan-truss: I see what you're saying when this is tested on the upload orders form, it seems to queue up the second file but eventually call remove all files after the first succeeds.  It could be a bug in our implementation of the callbacks as well, if we've never tried using it a different way.

I don't like some of the silent errors or warnings in the component like if you disallowed multiple and you dragged 2 files it will drop one without explanation.

src/components/FileUpload/FileUpload.jsx
```diff
@@ -20,80 +20,108 @@ registerPlugin(
...
+      <FilePond
+        ref={ref}
+        {...filePondProps}
+        allowMultiple={allowMultiple}
+        maxParralelUploads={maxParralelUploads}
+        acceptedFileTypes={acceptedFileTypes}
+        name={name}
+        onprocessfile={handleProcessFile}
+        onaddfilestart={onAddFile}
+      />
```
duncan-truss, line 98: I had to dig in the source code because it's not in the obvious documentation but it would be great to expose the invalid type error message label props so that we can make them more user friendly and not just default to showing the user confusing mime types:

https://github.com/pqina/filepond/blob/3b8b58147703d895848999aafe1aa25e9154ce18/locale/en-en.js#L28-L29

duncan-truss:
We could also check-in the real weight estimator spreadsheet to the repository, so folks have something to test with https://www.ustranscom.mil/dp3/weightestimator.cfm

If not on this PR we can add it in during Cypress and testdatagen PRs

src/components/FileUpload/FileUpload.stories.jsx
pearl-truss: Yeah I think I was able to achieve this:

On the orders page, I temporarily added 2 document uploader components. Each with a maxParallelUplads of 1 and allowMultiple's set to false

```
<div className="uploader-box">
              <FileUpload
                ref={this.filePondEl}
                createUpload={this.handleUploadFile}
                onChange={this.onChange}
                maxParallelUploads={1}
                allowMultiple={false}
                labelIdle={'Drag & drop or <span class="filepond--label-action">click to upload orders</span>'}
              />
              <FileUpload
                ref={this.filePondEl}
                allowMultiple={false}
                maxParallelUploads={1}
                createUpload={this.handleUploadFile}
                onChange={this.onChange}
                labelIdle={'Drag & drop or <span class="filepond--label-action">click to upload orders</span>'}
              />
              <div className="hint">(Each page must be clear and legible.)</div>
            </div>
```

Then I uploaded 3 files and clicked Next. And on the frontend and backend it looks like the uploads are associated with the same document

Screen Shot 2022-06-21 at 12 22 22 PM

`select * from user_uploads where document_id='474d7fd8-b75a-4f3e-8d95-be54d0204669';` 
Screen Shot 2022-06-21 at 12 28 58 PM

ronaktruss: I think the above demonstrates the regular uploads working because a document id already exists. When we don't have a document id on the amend page we try to have the backend create a document id and save the upload. We want the second file to depend on the results of the first file upload.

Do you know if this works on the AmendUpload page where the `createUpload` function tries to create a document id.

pearl-truss:
> Drag & drop or click to upload (1 file at a time)

this has been updated

src/components/FileUpload/FileUpload.jsx
duncan-truss: Ugg maybe I was wrong about the undefined behavior, filepond doesn't seem to fallback to their defaults?

```
labelFileTypeNotAllowed: 'File of invalid type',
fileValidateTypeLabelExpectedTypes: 'Expects {allButLastType} or {lastType}',
```
Screen Shot 2022-06-22 at 9 45 39 AM

duncan-truss (approved):

scarequotes (approved):
Thanks! Looks good.

PR #8777, opened 2022-06-16
MB-12621: Limit recalculation of ZipTransitDistance to PPMs

pearl-truss:
[Jira ticket](tbd) for this change

## Summary

Recent changes to the `DistanceZipLookup` function removed the restriction to not recalculate distance if the shipment had the distance field set. This can expectedly result in additional calls for service items that rely on distance: DLH, DSH, and FSC. This PR limits the previous change to only apply to PPMs since the `ppm_estimator` is explicitly setting the shipment distance and running with new zips wasn't resulting in distance recalculation 

Without this PR the `db_dev_fresh` and was failing, and tT(https://ustcdp3.slack.com/archives/CP6PTUPQF/p1655329802041779?thread_ts=1655328463.425639&cid=CP6PTUPQF)

## Setup to Run Your Code

1. `make db_dev_fresh` and `make_dev_dev_e2e_populate` should run without error

### DIFF
--- a/pkg/payment_request/service_param_value_lookups/distance_zip_lookup.go
+++ b/pkg/payment_request/service_param_value_lookups/distance_zip_lookup.go
@@ -52,6 +52,11 @@ func (r DistanceZipLookup) lookup(appCtx appcontext.AppContext, keyData *Service
 	if len(destinationZip) < 5 {
 		return "", apperror.NewInvalidInputError(*mtoShipmentID, fmt.Errorf(errorMsgForDestinationZip), nil, errorMsgForDestinationZip)
 	}
+
+	if mtoShipment.Distance != nil && mtoShipment.ShipmentType != models.MTOShipmentTypePPM {
+		return strconv.Itoa(mtoShipment.Distance.Int()), nil
+	}
+
 	distanceMiles, err := planner.ZipTransitDistance(appCtx, pickupZip, destinationZip)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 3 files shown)


pkg/payment_request/service_param_value_lookups/distance_zip_lookup.go
```diff
@@ -52,6 +52,11 @@ func (r DistanceZipLookup) lookup(appCtx appcontext.AppContext, keyData *Service
 	if len(destinationZip) < 5 {
 		return "", apperror.NewInvalidInputError(*mtoShipmentID, fmt.Errorf(errorMsgForDestinationZip), nil, errorMsgForDestinationZip)
 	}
+
+	if mtoShipment.Distance != nil && mtoShipment.ShipmentType != models.MTOShipmentTypePPM {
```
gidjin, line 56: Suggestion: Could you add a unit test case for this change?

gidjin (commented):
Change looks good. I left one suggestion

pkg/payment_request/service_param_value_lookups/distance_zip_lookup.go
pearl-truss: Also, I'm out tomorrow. I pushed changes to include more test. If it looks good, feel free to merge this. If not, I can make updates when I'm back on Tuesday

YanZ777 (approved):
Checked out the branch and things look good to me!

PR #8779, opened 2022-06-16
MB-12264 read only move details for QAE/CSR

leia-sefkin:
## Summary

This sets up the move task order page to be read only for QAE/CSR users, defining the following permissions for the necessary components. I wasn't sure on how the permissions should be broken down between TIO/TOO/SCs, so I defaulted to giving them all permissions, except for SIT extension creation/modification which I was fairly sure was TOO only. It's quite simple to update these so please feel free to comment where things should be changed. 

TOO/TIO/Service counselors (all have):

* update.billableWeight
* create.shipmentCancellation
* create.shipmentDiversionRequest
* create.reweighRequest
* update.MTOServiceItem

TOO only:

* create.SITExtension
* update.SITExtension

## Setup to Run Your Code

* After pulling down this branch and running the app you should be able to login as a QAE/CSR and have read only access as defined in the attached screenshots, which are taken from [the PO 10/11 QAE/CSR Miro board](https://miro.com/app/board/uXjVOJRSnug=/?moveToWidget=3458764524967260264&cot=14)
* Switch roles to TOO and you should still have access to the necessary components.
* Storybook components and tests should still be working as expected, using mock providers to add permissions

## Screenshots
Screen Shot 2022-06-16 at 3 52 17 PM
Screen Shot 2022-06-16 at 3 52 23 PM
Screen Shot 2022-06-16 at 3 52 41 PM


cypress/integration/office/txo/tooFlows.js
```diff
@@ -415,7 +415,7 @@ describe('TOO user', () => {
     cy.wait(['@requestShipmentCancellation']);
     // After updating, the button is disabeld and an alert is shown
     cy.get('[data-testid="request-cancellation-modal"]').should('not.exist');
-    cy.get('.shipment-heading').find('button').should('be.disabled').and('contain', 'Cancellation Requested');
+    cy.get('.shipment-heading').contains('Cancellation Requested');
```
leia-sefkin, line 418: this changed because the component is no longer a button but a tag (see the ShipmentHeading component)

pkg/auth/authentication/permissions.go
```diff
@@ -22,17 +22,17 @@ type RolePermissions struct {
 var TOO = RolePermissions{
 	RoleType: roles.RoleTypeTOO,
 	Permissions: []string{"update.move", "create.serviceItem",
-		"update.shipment", "update.financialReviewFlag", "update.orders", "update.allowances"},
+		"update.shipment", "update.financialReviewFlag", "update.orders", "update.allowances", "update.billableWeight", "create.shipmentCancellation", "create.SITExtension", "update.SITExtension", "create.shipmentDiversionRequest", "create.reweighRequest", "update.MTOServiceItem"},
```
leia-sefkin, line 25: only the TOO has SIT extension creation and update permissions, not sure if there are other differences to map out here

src/components/Office/ShipmentContainer/ShipmentContainer.stories.jsx
```diff
@@ -23,121 +25,127 @@ const shipmentInfoNoReweigh = {
 };
 
 export const HHG = () => (
```
leia-sefkin, line 27: the below changes look big, but it is just being wrapped in <MockProviders> so we can pass permissions

Ryan-Koch (approved):

PR #8780, opened 2022-06-16
remove-comments-from-FE

NamibiaTorres:
##There is no Jira ticket associated with this change

## Summary

Cleaning up some code that was not merged in the previous PR.

### Files changed
1. `ppmCustomerShared.js`

### DIFF
--- a/cypress/support/ppmCustomerShared.js
+++ b/cypress/support/ppmCustomerShared.js
@@ -33,18 +33,6 @@ export function signInAndNavigateToAboutPage(userId, selectAdvance) {
   fillOutAboutPage(selectAdvance);
 }
 
-// export function signInAndNavigateToAboutPageWithoutAdvance(userId, selectAdvance = false) {
-//   cy.apiSignInAsUser(userId);
-//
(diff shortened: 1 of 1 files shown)


cypress/support/ppmCustomerShared.js
```diff
@@ -33,18 +33,6 @@ export function signInAndNavigateToAboutPage(userId, selectAdvance) {
   fillOutAboutPage(selectAdvance);
 }
```
felipe-lee, line 35: I think like 32 above can also be removed

NamibiaTorres: You got it! I also added a comment above one of the functions. I could rename the function though. What do you think? Rename or keep the comment?

felipe-lee: Thanks!

I'm fine with either. I guess we have a few other functions named `NavigateFrom<Page1>To<Page2>` so it would fit those better, but I'll leave it up to you.

felipe-lee (approved):

PR #8794, opened 2022-06-18
[MB-12616] pagination, sorting, and filtering for move search

mboren:
## Summary
This PR adds pagination, filtering, and sorting to the move search results table used by QAE/CSR users.
Most of the changes are heavily based on the `TableQueue` component.

I had to update `SearchResultsTable` to take a query instead of the table data because it needs to be able to 
re-query the data when there are changes to column filters, column sort ordering, or pagination.

I also had to make a lot of changes on the backend (`move_searcher.go`) to handle all the different filters and sorting options.

I flattened the response format a bit to just include the fields we need from `ServiceMember` and `DutyLocation` instead of nesting the entire objects, because I thought I would be able to get these fields directly from my query without needing to EagerPreload these relationships and pull them out in Go code (see `model_to_payload.go`). I never quite got that to work, so these API changes were not necessary. I'm open to changing the API format back, but I'm not sure it makes sense to take the time to do that.

The search query feels unnecessarily expensive, but it's the best I could figure out how to do with Pop without doing a raw query with a lot of string concatenation, which I would like to avoid.
Eager preloading the shipments to count them, and then ALSO counting them in the query irks me. There's gotta be a better way.
One easy way to make this query faster would be use a different query for each kind of search. Move code search should only return one result, so filtering/sorting don't make any sense, and we could pare down the query a lot.

## Setup to Run Your Code

<details>

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

- Log in to the office app as a QAE/CSR user.
- Search for a name that will give you multiple results.
- Attempt to sort each column both ascending and descending by clicking on the column header.
- Try filtering each column.

### Frontend

### Backend

### column filtering
https://user-images.githubusercontent.com/2627844/175131344-fa0636f1-794b-4491-bf7b-73655cade589.mov
### column sorting
https://user-images.githubusercontent.com/2627844/175131534-1ba6469f-8e55-4356-815f-0b1755268ac0.mov
### pagination
https://user-images.githubusercontent.com/2627844/175131898-b3da1172-b161-452c-9c7e-4f8ad21409c0.mov


pkg/services/move/move_searcher.go
```diff
@@ -39,26 +47,132 @@ func (s moveSearcher) SearchMoves(appCtx appcontext.AppContext, locator *string,
...
+}
+
+func sortOrder(sort *string, order *string, customerNameSearch *string) QueryOption {
+	return func(query *pop.Query) {
+		if sort != nil && order != nil {
+			sortTerm := parameters[*sort]
+			if *sort == "customerName" {
+				orderName(query, order)
+			} else {
+				query.Order(fmt.Sprintf("%s %s", sortTerm, *order))
```
mboren, line 165: Building a query by manually concatenating a user-supplied string (`order` comes from request body) gives me the heebie jeebies. However, go-swagger will only let it past if it is `"asc"` or `"desc"`. So right now, I think we're safe from SQL injection here. maybe it would make sense to double check anyway though?

src/pages/Office/QAECSRMoveSearch/QAECSRMoveSearch.jsx
```diff
@@ -106,17 +36,11 @@ const QAECSRMoveSearch = ({ history }) => {
...
-  };
+  }, []);
 
-  const { searchResult, isLoading, isError } = useQAECSRMoveSearchQueries({
-    moveCode: search.moveCode,
-    dodID: search.dodID,
-    customerName: search.customerName,
-  });
+  const isLoading = false;
+  const isError = false;
```
mboren, line 42: remove these

pkg/services/move/move_searcher.go
mboren: also, I should add that this follows a pattern that we've used in other places

Ryan-Koch: I can understand that feeling. The enumeration you mentioned along with pop forcing the string into a parameterized query *should* keep things safe, though it does mean trusting each of these tooling layers to do what they're expected to do.

pkg/services/move/move_searcher.go
```diff
@@ -19,16 +21,30 @@ func NewMoveSearcher() services.MoveSearcher {
...
 		verrs.Add("search key", "search by multiple keys is not supported")
-		return models.Moves{}, apperror.NewInvalidInputError(uuid.Nil, nil, verrs, "")
+		return models.Moves{}, 0, apperror.NewInvalidInputError(uuid.Nil, nil, verrs, "")
+	}
+
+	// The SQL % operator filters out strings that are below this similarity threshold
+	// We have to set it here because other areas of the code that do a trigram search
+	// (eg Duty Location search) may set a different threshold.
+	// If the threshold is too high, we may filter out too many results and make searching harder.
+	// If it's too low, the query will get slower/more memory intensive.
```
Ryan-Koch, line 44: This is a great comment block. The commentary about the tradeoff behind the threshold value is easy to understand and gives future contributors a peak into the 'why' for what it's set to.

Ryan-Koch (approved):

PR #8810, opened 2022-06-23
MB-12926 add funcs to make weight tickets

felipe-lee:
## Summary

This isn't strictly tied to the DB changes (which are being made in #8809), but will hopefully make it easier to create WeightTickets for tests. I made this a separate PR to not slow down the DB PR.

## Testing Changes

1. Open any test file
2. Update a test to use any of the new functions.
3. [Run tests](https://transcom.github.io/mymove-docs/docs/backend/testing/run-go-tests)


pkg/testdatagen/make_weight_ticket.go
```diff
@@ -0,0 +1,110 @@
...
+		assertions.Document.ServiceMember = serviceMember
+	}
+
+	return assertions
+}
+
+// getOrCreateDocument checks if a document exists. If it does, it returns it, otherwise, it creates it
+func getOrCreateDocument(db *pop.Connection, document models.Document, assertions Assertions) models.Document {
+	if assertions.Stub && document.CreatedAt.IsZero() || document.ID.IsNil() {
+		upload := MakeUserUpload(db, assertions)
```
duncan-truss, line 104: Is the rationale here for using MakeUserUpload so that we can pass uploads through the assertions at creation time?

For the minimal weight ticket I think we won't have any uploads associated with the document when first saved so that makes me think we still need to have a way to call `MakeDocument`.

Keep in mind that this document object you're returning is going to have an empty array of uploads for the field UserUploads unless you explicitly assign them back to it.  It's probably okay for usage in the app because they are linked with the foreign keys but can be misleading if it gets passed in to another assertion or something.

pkg/testdatagen/make_weight_ticket.go
```diff
@@ -0,0 +1,110 @@
...
+		assertions.Order.ServiceMember = serviceMember
+		assertions.Document.ServiceMemberID = serviceMember.ID
+		assertions.Document.ServiceMember = serviceMember
+	}
+
+	return assertions
+}
+
+// getOrCreateDocument checks if a document exists. If it does, it returns it, otherwise, it creates it
+func getOrCreateDocument(db *pop.Connection, document models.Document, assertions Assertions) models.Document {
```
duncan-truss, line 102: The other thing to be aware of here is making sure to pass in the assertions.UserUploader here for creating the real files (local/s3)

pkg/testdatagen/make_weight_ticket.go
felipe-lee: Ah, it was because this used to be called in the non-minimal one, and I moved the call to the minimal one after talking with the others about what fields should be required. But I forgot to update it to only create the document, not the upload anymore. Thanks for pointing it out, I'll make updates to this.

pkg/testdatagen/make_weight_ticket.go
felipe-lee: I think that's for the thing calling the `Make<thing>` function to do right? Like I'm just passing through whatever assertions were sent in at the beginning.

pkg/testdatagen/make_weight_ticket.go
```diff
@@ -0,0 +1,143 @@
...
+func ensureServiceMemberIsSetUpInAssertions(db *pop.Connection, assertions Assertions) Assertions {
+	if !assertions.Stub && assertions.ServiceMember.CreatedAt.IsZero() || assertions.ServiceMember.ID.IsNil() {
+		serviceMember := MakeExtendedServiceMember(db, assertions)
+
+		assertions.ServiceMember = serviceMember
+		assertions.Order.ServiceMemberID = serviceMember.ID
+		assertions.Order.ServiceMember = serviceMember
+		assertions.Document.ServiceMemberID = serviceMember.ID
+		assertions.Document.ServiceMember = serviceMember
+	} else {
```
felipe-lee, line 108: I wasn't sure if to just set the fields like I did, or if I should check the other fields too and raise an error if there are conflicts.

pkg/testdatagen/make_weight_ticket.go
```diff
@@ -0,0 +1,143 @@
...
+	if assertions.Stub && document.CreatedAt.IsZero() || document.ID.IsNil() {
+		return MakeDocument(db, assertions)
+	}
+
+	return document
+}
+
+// getOrCreateDocumentWithUploads checks if a document exists. If it doesn't, it creates it. Then checks if the document
+// has any uploads. If not, creates an upload associated with the document. Returns the document at the end.
+func getOrCreateDocumentWithUploads(db *pop.Connection, document models.Document, assertions Assertions) models.Document {
```
duncan-truss, line 129: I'm not quite sure this works for all the cases where you pass in UserUploads the way someone seeing it might think.  If I wanted to create a new document with 2 user uploads then we need to check the length of `assertions.Document.UserUploads` or `document.UserUploads` and  call `MakeUserUpload()` on each of them with the newly created `doc.id` and assign them back to `doc.UserUploads`.

Let me know if that's not what you intended for this function to do or I'm misreading something.

pkg/testdatagen/make_weight_ticket.go
```diff
@@ -0,0 +1,143 @@
...
+	// so we'll look at assertions.WeightTicket.<Document>
+	emptyDocument := getOrCreateDocumentWithUploads(db, assertions.WeightTicket.EmptyDocument, assertions)
+	fullDocument := getOrCreateDocumentWithUploads(db, assertions.WeightTicket.FullDocument, assertions)
+
+	emptyWeight := unit.Pound(14500)
+	fullWeight := emptyWeight + unit.Pound(4000)
+
+	fullAssertions := Assertions{
+		WeightTicket: models.WeightTicket{
+
```
duncan-truss, line 60: I think we're missing VehicleDescription here zoom zoom 🚗

felipe-lee: Ah, thanks for catching that.

pkg/testdatagen/make_weight_ticket.go
felipe-lee: Hmm I think I had originally been envisioning that if they created uploads that were passed in as assertions, they'd have also created the document themselves. 

You're right though that this isn't accounting for them passing in assertions for those UserUploads rather than having created them themselves. I'll update to account for that.

duncan-truss (approved):
Thanks, I think you addressed most of my feedback.

pkg/testdatagen/make_weight_ticket.go
felipe-lee: Ok, I think I addressed this in the last commit. We should now be able to handle if:

1. Doc passed into assertion w/o upload assertions (will create default upload)
2. doc passed in w/upload assertions that haven't been created (will loop over assertions and create uploads)
3. doc passed in w/upload assertions that exist (will loop over to double check they exist, meaning a mix of case 2 and this one would be handled too)
4. upload assertion that needs to be created passed in w/o existing doc (will create doc and then set up upload assertions properly and create upload)
5. no doc or upload assertions passed in (create docs for each one and a default upload)

There is another case I was trying to account for and then realized it isn't really possible: the caller creating an upload not associated with a doc, meaning we'd have had to create the doc and update the upload to point at the doc. This can't be done by the db, and while it's possible to pass in the assertions incorrectly (pass in the upload assertion w/o specifying the doc), I didn't think it'd be a good idea to code that since this is already catching quite a few cases and I feel like that's more of making sure the user uses it correctly.

PR #8829, opened 2022-06-27
MB-12927: weight ticket service objects

ronaktruss:
## Summary

Adds weight ticket services objects.

Please give additional attention to the comments in code and tests.

- [x] weight ticket creator
- [x] weight ticket updater
- [x] mto shipment fetcher

## Setup to Run Your Code

<details>
<summary>💻 You will need to use 1 separate terminals to test this locally.</summary>

##### Terminal 1

Reset your test setup

```sh
make server_test_setup
```

Run the specific test in question 

```sh
go test ./pkg/services/weight_tickets/ -testify.m TestValidationRules
```
OR

Run all of the tests
```sh
go test ./pkg/services/weight_tickets
```
</details>

### Backend


pkg/models/weight_ticket.go
```diff
@@ -51,3 +54,44 @@ func (w *WeightTicket) Validate(_ *pop.Connection) (*validate.Errors, error) {
...
+func filterDeletedValued(userUploads UserUploads) UserUploads {
+	if userUploads != nil {
+		index := 0
+		for _, userUpload := range userUploads {
+			if userUpload.DeletedAt == nil {
+				userUploads[index] = userUpload
+				index++
+			}
+		}
+		userUploads = userUploads[:index]
```
ronaktruss, line 94: We may want to set userUploads to nil rather than an empty slice.

pkg/models/weight_ticket.go
```diff
@@ -51,3 +54,44 @@ func (w *WeightTicket) Validate(_ *pop.Connection) (*validate.Errors, error) {
...
+	}
+
+	weightTicket.EmptyDocument.UserUploads = filterDeletedValued(weightTicket.EmptyDocument.UserUploads)
+	weightTicket.FullDocument.UserUploads = filterDeletedValued(weightTicket.FullDocument.UserUploads)
+	weightTicket.ProofOfTrailerOwnershipDocument.UserUploads = filterDeletedValued(weightTicket.ProofOfTrailerOwnershipDocument.UserUploads)
+
+	return &weightTicket, nil
+}
+
+func filterDeletedValued(userUploads UserUploads) UserUploads {
```
ronaktruss, line 85: Are folks okay with this being done on the server side and not the DB?

pkg/models/weight_ticket.go
```diff
@@ -51,3 +54,44 @@ func (w *WeightTicket) Validate(_ *pop.Connection) (*validate.Errors, error) {
 		&validators.UUIDIsPresent{Name: "ProofOfTrailerOwnershipDocumentID", Field: w.ProofOfTrailerOwnershipDocumentID},
 	), nil
 }
+
+func FetchWeightTicketByIDExcludeDeletedUploads(db *pop.Connection, weightTicketID uuid.UUID) (*WeightTicket, error) {
```
pearl-truss, line 58: Praise: naming things is hard, and I like how this name doubles as documentation for this function

pkg/models/weight_ticket.go
pearl-truss: I don't feel I know enough to have strong feelings either way. But I would be curious to know more about tradeoffs for either option (if applicable)

pkg/services/weight_ticket/rules.go
```diff
@@ -0,0 +1,95 @@
...
+		if originalWeightTicket.ProofOfTrailerOwnershipDocumentID.IsNil() {
+			verrs.Add("ProofOfTrailerOwnershipDocumentID", "ProofOfTrailerOwnershipDocumentID must exist")
+		}
+
+		if newWeightTicket.VehicleDescription == nil || *newWeightTicket.VehicleDescription == "" {
+			verrs.Add("VehicleDescription", "VehicleDescription must exist")
+		}
+
+		if newWeightTicket.EmptyWeight == nil || *newWeightTicket.EmptyWeight < 1 {
+			verrs.Add("EmptyWeight", "EmptyWeight must have a value of at least 1")
```
pearl-truss, line 52: Nit/Question: is the minimum set to 1 confirmed? I could see a case for 0, but not sure if that has been considered already

pkg/services/weight_ticket/rules.go
```diff
@@ -0,0 +1,95 @@
...
+			if *newWeightTicket.OwnsTrailer && *newWeightTicket.TrailerMeetsCriteria &&
+				len(originalWeightTicket.ProofOfTrailerOwnershipDocument.UserUploads) < 1 {
+				verrs.Add("ProofOfTrailerOwnershipDocument", "At least 1 proof of ownership file is required")
+			}
+		}
+
+		return verrs
+	})
+}
+
```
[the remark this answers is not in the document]
pearl-truss: I think that's a good idea (just so that we're not creating test data that'll eventually be invalid but that's only a minor concern since we could always update our test data as part of adding the new check)

pkg/services/weight_ticket/rules.go
ronaktruss: You are right, we should allow 0 here.

pkg/models/weight_ticket.go
ronaktruss: The alternative would be to write a query that explicitly joins all of the tables that we are concerned about and add the appropriate where clause to filter out the soft deleted data. I don't think any of the functions that pop provides, help us achieve this in a more streamlined way. 

The current approach seems a bit more readable, but relies on Pop. The alternative would mean less reliance on pop, but would likely be a bit harder to read. (In reality this is context dependent, reading a larger SQL query vs reading go code that modifies fetched data). 

Filtering on the DB side is likely faster when we have a large number of uploads, but that's not really applicable in this scenario.

duncan-truss: Yea i think I broke it out into a [separate query](https://github.com/transcom/mymove/blob/58d3ff5c31c68353cb5c208e40b19dcf237dd512/pkg/models/service_member.go#L364-L374) when fixing soft deleted orders uploads but it required another query anyway because of an eager loading bug.  I think you could generalize this to make it work for other non-upload models too.

My one ask would be to move this into the services package, we shouldn't have database logic in the models for new code.  I know we haven't gotten rid of all of the tech debt yet for some customer models having database queries.

pkg/services/weight_ticket/rules.go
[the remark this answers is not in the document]
felipe-lee: I think it might be good. Also serves to make sure that the PPM shipment actually exists, otherwise we'll get a DB error when we try creating the weight ticket.

pkg/services/weight_ticket/rules.go
```diff
@@ -0,0 +1,95 @@
...
+			verrs.Add("VehicleDescription", "VehicleDescription must exist")
+		}
+
+		if newWeightTicket.EmptyWeight == nil || *newWeightTicket.EmptyWeight < 1 {
+			verrs.Add("EmptyWeight", "EmptyWeight must have a value of at least 1")
+		}
+
+		if newWeightTicket.MissingEmptyWeightTicket == nil {
+			verrs.Add("MissingEmptyWeightTicket", "MissingEmptyWeightTicket is required")
+		} else if !*newWeightTicket.MissingEmptyWeightTicket && len(originalWeightTicket.EmptyDocument.UserUploads) < 1 {
```
felipe-lee, line 57: **question**: Should this check for uploads regardless of the value of `MissingEmptyWeightTicket`? Because if they are missing it, they still need to upload the constructed weights file right? Or is that not required?

Same question for the full weight doc.

pkg/services/weight_ticket/rules.go
```diff
@@ -0,0 +1,95 @@
...
+			verrs.Add("MissingFullWeightTicket", "MissingFullWeightTicket is required")
+		} else if !*newWeightTicket.MissingFullWeightTicket && len(originalWeightTicket.FullDocument.UserUploads) < 1 {
+			verrs.Add("FullWeightDocument", "At least 1 full weight file is required")
+		}
+
+		if newWeightTicket.OwnsTrailer == nil {
+			verrs.Add("OwnsTrailer", "OwnsTrailer is required")
+		}
+
+		if newWeightTicket.TrailerMeetsCriteria == nil {
```
felipe-lee, line 75: **thought**: I think this field is only required if they say yes to owning the trailer, at least that was my understanding, but I could be wrong.

pkg/models/weight_ticket.go
ronaktruss: 🤦 For some reason I thought it was the opposite and I had moved it from services into the model object.

pkg/services/weight_ticket/weight_ticket_creator.go
```diff
@@ -0,0 +1,87 @@
...
+		checks: basicChecks(),
+	}
+}
+
+func (f *weightTicketCreator) CreateWeightTicket(appCtx appcontext.AppContext, ppmShipmentID uuid.UUID) (*models.WeightTicket, error) {
+	err := validateWeightTicket(appCtx, nil, nil, f.checks...)
+
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
ronaktruss: Yeah just added that incase we add some other tests that can actually fail.

pkg/services/weight_ticket/rules.go
ronaktruss: Hmm, I think they are probably required to have at least 1 upload (ie. they have a constructed weight ticket only, they could theoretically upload it to either the empty or full weight ticket and it should get approved), but I'm not sure if we want to enforce that. Let me confirm with James

pkg/services/weight_ticket/rules.go
ronaktruss: Your understanding is accurate. I made this required because I assume the FE will be always sending this value (to avoid the scenario where we edit a yes/yes to just a no and have a partial update)

pkg/services/weight_ticket/weight_ticket_creator.go
```diff
@@ -0,0 +1,87 @@
...
+
+	return &weightTicket, nil
+}
+
+func (f *weightTicketCreator) addNewDocument(appCtx appcontext.AppContext) (*models.Document, error) {
+	document := &models.Document{
+		ServiceMemberID: appCtx.Session().ServiceMemberID,
+	}
+	verrs, err := appCtx.DB().ValidateAndCreate(document)
```
duncan-truss, line 78: Something I've had in the back of my mind to start using, is that you can call different Pop methods with a slice of models in addition to just one model at a time.  If it doesn't hinder error handling I'd be curious if it results in 1 statement instead of the 3 individual ones we end up doing here.

ronaktruss: Looks to still result in 3 separate insert statements, though the code seems a bit simpler.

pkg/models/weight_ticket.go
```diff
@@ -51,3 +54,44 @@ func (w *WeightTicket) Validate(_ *pop.Connection) (*validate.Errors, error) {
...
+
+	return &weightTicket, nil
+}
+
+func filterDeletedValued(userUploads UserUploads) UserUploads {
+	if userUploads != nil {
+		index := 0
+		for _, userUpload := range userUploads {
+			if userUpload.DeletedAt == nil {
+				userUploads[index] = userUpload
```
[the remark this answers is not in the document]
felipe-lee: I think both not having side effects, and making it so future peeps don't accidentally change this in a way that doesn't do what they expect because they misunderstood which variable was being edited.

pkg/services/mto_shipment/mto_shipment_fetcher.go
```diff
@@ -44,6 +45,7 @@ func (f mtoShipmentFetcher) ListMTOShipments(appCtx appcontext.AppContext, moveI
 			"DestinationAddress",
 			"SecondaryDeliveryAddress",
 			"MTOServiceItems.Dimensions",
+			"PPMShipment.WeightTickets.EmptyDocument.UserUploads.Upload",
```
duncan-truss, line 48: Did you consider having a separate fetcher for weight tickets with documents?  I have a feeling this might get too unwieldy to work and will be over fetching too much data in many cases.

duncan-truss: Yea I'm just thinking by the end of PPM closeout you will have ended up adding:
```
PPMShipment.WeightTickets.EmptyDocument.UserUploads.Upload
PPMShipment.WeightTickets.FullDocument.UserUploads.Upload
PPMShipment.WeightTickets.TrailerOwenshipDocument.UserUplaods.Upload
PPMShipment.ProGearTickets.EmptyDocument.UserUploads.Upload
PPMShipment.ProGearTickets.FullDocument.UserUploads.Upload
PPMShipment.Expenses.RecieptDocument.UserUploads.Upload
```

Now maybe there are more efficient ways to not blindly query all of those records but that would be the default approach. My instincts say we want to normalize these entitites (WeightTickets, ProGear, and Expenses) in Redux so they are easier to deal with.  Are you planning for the individual pages to return the full PPM shipment record every time when say just a single weight ticket/pro-gear/expense is modified?

pkg/services/weight_ticket/rules.go
ronaktruss: This should now be resolved. We generally expect at least 1 document to exists for both empty and full weights but if the user mentions that they don't have either weight docs then they require at least 1 doc between the two.

pkg/services/weight_ticket/rules.go
```diff
@@ -0,0 +1,95 @@
...
+
+		if newWeightTicket.MissingEmptyWeightTicket == nil {
+			verrs.Add("MissingEmptyWeightTicket", "MissingEmptyWeightTicket is required")
+		} else if !*newWeightTicket.MissingEmptyWeightTicket && len(originalWeightTicket.EmptyDocument.UserUploads) < 1 {
+			verrs.Add("EmptyWeightDocument", "At least 1 empty weight file is required")
+		}
+
+		if newWeightTicket.FullWeight == nil || *newWeightTicket.FullWeight < 1 {
+			verrs.Add("FullWeight", "FullWeight must have a value of at least 1")
+		}
```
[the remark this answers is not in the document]
ronaktruss: The validation for this has been added

pkg/services/weight_ticket/weight_ticket_creator.go
```diff
@@ -0,0 +1,70 @@
...
+		} else if err != nil {
+			return apperror.NewQueryError("Document for WeightTicket", err, "")
+		}
+
+		weightTicket = models.WeightTicket{
+			EmptyDocumentID:                   allDocs[0].ID,
+			FullDocumentID:                    allDocs[1].ID,
+			ProofOfTrailerOwnershipDocumentID: allDocs[2].ID,
+			PPMShipmentID:                     ppmShipmentID,
+		}
```
duncan-truss, line 51: Could you also assign the association fields of EmptyDocument, FullDocument, and ProofOfTrailerOwnershipDocument?  Then in the weight_ticket_creator_test.go check that the service member id is the one associated with the session?

pkg/services/weight_ticket/weight_ticket_creator_test.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+func (suite *WeightTicketSuite) TestWeightTicketCreator() {
+	suite.Run("Can successfully create a WeightTicket", func() {
+		serviceMember := testdatagen.MakeDefaultServiceMember(suite.DB())
+		session := &auth.Session{
+			ServiceMemberID: serviceMember.ID,
+		}
+
+		ppmShipment := testdatagen.MakeMinimalDefaultPPMShipment(suite.DB())
```
duncan-truss, line 15: This is minor but there will be a different service member created for this PPM Shipment than the one that you created above unless you pass it in as an assertion to the factory.  I'm more concerned with testing that the service member id is populated in the documents of the weigh ticket like in the comment above.

pkg/services/weight_ticket/weight_ticket_creator_test.go
```diff
@@ -0,0 +1,26 @@
+package weightticket
+
+import (
+	"github.com/transcom/mymove/pkg/auth"
+	"github.com/transcom/mymove/pkg/testdatagen"
+)
+
+func (suite *WeightTicketSuite) TestWeightTicketCreator() {
```
duncan-truss, line 8: To test that your creator returns errors successfully you could try when a service member id has a value of uuid.Nil or when ppmShipmentId is uuid.Nil.

pkg/services/weight_ticket/weight_ticket_creator_test.go
ronaktruss: The issues above should be resolved now. Yeah there isn't a good way to verify that the serviceMemberId that is passed as an assertion into PPMShipment is actually the same so I just excluded this. We'd likely need to pass something like
```go
ppmShipment := testdatagen.MakeMinimalPPMShipment(suite.DB(), testdatagen.Assertions{
	Order: models.Order{
	ServiceMember: serviceMember,
	},
})
```

pkg/services/weight_ticket/weight_ticket_updater_test.go
```diff
@@ -0,0 +1,205 @@
...
+			suite.IsType(apperror.PreconditionFailedError{}, updateErr)
+
+			suite.Equal(
+				fmt.Sprintf("Precondition failed on update to object with ID: '%s'. The If-Match header value did not match the eTag for this record.", originalWeightTicket.ID.String()),
+				updateErr.Error(),
+			)
+		}
+	})
+
+	suite.Run("Successfully updates", func() {
```
duncan-truss, line 135: Could you add a test that fails when there are no uploads associated with any of the documents during update?  I think most of the errors are captured in the rules test but this would make sure the error is bubbling up somewhere.

duncan-truss (approved):
Great work 🎉  I think this is in a good place just had a few untested path comments.

PR #8840, opened 2022-06-29
Weight Ticket Create and Update Handlers

NamibiaTorres:
## Summary

This ticket creates the handlers for weight tickets.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the
2. Login as a
3.

### Frontend

### Backend

### DIFF
--- a/pkg/handlers/internalapi/weight_ticket.go
+++ b/pkg/handlers/internalapi/weight_ticket.go
@@ -1,0 +1,43 @@
+package internalapi
+
+import (
+	"github.com/go-openapi/runtime/middleware"
+	"github.com/transcom/mymove/pkg/appcontext"
+	"github.com/transcom/mymove/pkg/apperror"
+	weightticketops "github.com/transcom/mymove/pkg/gen/internalapi/internaloperations/documents"
+	"github.com/transcom/mymove/pkg/handlers"
+	"github.com/transcom/mymove/pkg/handlers/primeapi/payloads"
+	"go.uber.org/zap"
+)
+
+// Create weightTicketHandler
+type CreateWeightTicketHandler struct {
+	handlers.HandlerConfig
+	weightTicketCreator services.WeightTicketCreator
+}
+
+// Handle creates a weight ticket
+// Depending on the SO, may need to change the doument params to weight ticket params
+func (h CreateWeightTicketHandler) Handle(params weightticketops.CreateWightTicketParams) middleware.Responder {
+	return h.AuditableAppContextFromRequestWithErrors(params.HTTPRequest,
+		func(appCtx appcontext.AppContext) (middleware.Responder, error) {
+			// Get payload
+			payload := params.Body
+			if payload == nil {
+				missingBodyErr := apperror.NotFoundError{}
+				appCtx.logger().Error(missingBodyErr.Error())
+				return weightticketops.NewCreateWightTicketBadRequest().WithPayload(payloads.ClientError(handlers.BadRequestErrMessage, "Wight Ticket body cannot be empty", h.GetTraceIDFromRequest(params.HTTPRequest))), missingBodyErr
+			}
+			weightTicket := payloads.WeightTicketModelFromCreate(payload)
+			var err error
+
+			weightTicket, err = h.weightTicketCreator.CreateWightTicket(appCtx, weightTicket)
+
+			if err != nil {
+				appCtx.Logger().Error("internalapi.CreateWeightTicketHandler", zap.Error(err))
+				//TODO: maybe add a switch statement here?
+			}
+			returnPayload := payloads.WeightTicket(weightTicket)
+			return weightticketops.NewCreateWeightTicketOK().WithPayload(returnPayload), nil
+		})
+}
--- a/pkg/handlers/internalapi/weight_ticket_test.go
+++ b/pkg/handlers/internalapi/weight_ticket_test.go
@@ -1,0 +1,1 @@
+package internalapi


swagger-def/internal.yaml
```diff
@@ -1000,6 +1000,91 @@ definitions:
...
       - empty_weight_ticket_missing
       - trailer_ownership_missing
+  CreateWeightTicket:
+    type: object
+    properties:
+      id:
+        type: string
+        format: uuid
+        example: c56a4180-65aa-42ec-a945-5fd21dec0538
+      ppm_shipment_id:
```
duncan-truss, line 1010: These fields should be in camel case `ppmShipmentId` I think only some really old entities still use snake case.

swagger-def/internal.yaml
```diff
@@ -1000,6 +1000,91 @@ definitions:
...
+      proof_of_trailer_ownership_document_id:
+        title: Trailer Document ID
+        type: string
+        format: uuid
+        example: c56a4180-65aa-42ec-a945-5fd21dec0538
+      proof_of_trailer_ownership_document:
+        title: Proof of Trailer Ownership Document
+        $ref: '#/definitions/DocumentPayload'
+    required:
+      - ppm_shipment_id
```
duncan-truss, line 1078: I think if you look at Ronak's PR the only required information for the create endpoint request is going to be the PPM Shipment id.  That may not need a request body at all if it or the mtoShipmentId is passed in the the path.

For the UpdateWeightTicket request payload you'll need some of these:
- vehicleDescription
- emptyWeight
- missingEmptyWeightTicket
- fullWeight
- missingFullWeightTicket
- ownsTrailer
- trailerMeetsCriteria

For the response payload you can include the documentIds and document objects but we aren't going to let the customer alter those when submitting the form data.  UserUploads will have their own endpoint that adds them to the proper document. We can't reuse a single shape for the request and response payloads here for that reason and should split them out.

swagger-def/internal.yaml
```diff
@@ -4332,6 +4446,84 @@ paths:
           description: not authorized to modify this move
         '500':
           description: server error
+  /{ppmShipmentId}/weight_ticket:
```
duncan-truss, line 4449: It is convention to still prefix all resource ids by their type so we know what it is an id of in REST APIs.  So I would be expecting this to be `ppm-shipments/{ppmShipmentId}/weight-tickets`. Note the pluralized names and kebab case instead of underscores.  If you were just looking at the request URL in the logs or network console you wouldn't know what that uuid was by just looking at it without the resource name.

pkg/handlers/internalapi/weight_ticket.go
```diff
@@ -0,0 +1,142 @@
...
+			}
+
+			// NO NEED FOR payload_to_model, will need for Update
+			ppmShipmentID, err := uuid.FromString(params.PpmShipmentID.String())
+			if err != nil {
+				appCtx.Logger().Error("missing PPM Shipment ID", zap.Error(err))
+				return weightticketops.NewCreateWeightTicketBadRequest(), nil
+			}
+
+			var weightTicket *models.WeightTicket
```
duncan-truss, line 46: if you use `weightTicket, err :=` on the next line you should be able to remove this.  You were probably seeing a warning because `err` is not a new variable.

pkg/handlers/internalapi/weight_ticket.go
```diff
@@ -0,0 +1,142 @@
...
+			weightTicket, err = h.weightTicketCreator.CreateWeightTicket(appCtx, ppmShipmentID)
+
+			if err != nil {
+				appCtx.Logger().Error("internalapi.CreateWeightTicketHandler", zap.Error(err))
+				// Can get a status error
+				// Can get an DB error - does the weight ticket, doc create?
+				// Can get an error for whether the PPM exist
+				switch err.(type) {
```
duncan-truss, line 54: You might be missing a not found 404 error here.

pkg/handlers/internalapi/weight_ticket.go
```diff
@@ -0,0 +1,142 @@
...
+				case apperror.ForbiddenError:
+					return weightticketops.NewCreateWeightTicketForbidden(), err
+				case apperror.UnprocessableEntityError:
+					return weightticketops.NewCreateWeightTicketUnauthorized(), err
```
duncan-truss, line 58: Unprocessable entity should have it's own response code of 422, unauthorized would be a 401.

swagger-def/internal.yaml
```diff
@@ -1000,6 +1000,120 @@ definitions:
...
+        type: boolean
+      proofOfTrailerOwnershipDocumentId:
+        title: Trailer Document ID
+        type: string
+        format: uuid
+        example: c56a4180-65aa-42ec-a945-5fd21dec0538
+        readOnly: true
+      proofOfTrailerOwnershipDocument:
+        title: Proof of Trailer Ownership Document
+        $ref: '#/definitions/DocumentPayload'
```
duncan-truss, line 1079: We'll need to return an eTag in this response so the form update has one to send with it on submission.  Most of the time we don't return the updatedAt because that's what the encoded eTag value represents.

swagger-def/internal.yaml
NamibiaTorres: I am changing this definition to "WeightTicket" instead of CreateWeightTicket. What are your thoughts? The handler for the create should not be returning a body.

swagger-def/internal.yaml
NamibiaTorres: Oh I see! Thanks. I will make this change.

pkg/handlers/internalapi/weight_ticket.go
```diff
@@ -0,0 +1,142 @@
...
+// Depending on the SO, may need to change the document params to weight ticket params
+func (h CreateWeightTicketHandler) Handle(params weightticketops.CreateWeightTicketParams) middleware.Responder {
+	return h.AuditableAppContextFromRequestWithErrors(params.HTTPRequest,
+		func(appCtx appcontext.AppContext) (middleware.Responder, error) {
+			if appCtx.Session() == nil {
+				noSessionErr := apperror.NewSessionError("No user session")
+				return weightticketops.NewCreateWeightTicketUnauthorized(), noSessionErr
+			}
+
+			if !appCtx.Session().IsMilApp() && appCtx.Session().ServiceMemberID == uuid.Nil {
```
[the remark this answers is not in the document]
NamibiaTorres: That is a good question. I borrowed this from mto_shipments handler since I knew I wanted to check the service member ID was being used in the correct app.

swagger-def/internal.yaml
```diff
@@ -1000,6 +1000,120 @@ definitions:
...
+        format: uuid
+        example: c56a4180-65aa-42ec-a945-5fd21dec0538
+        readOnly: true
+      ppmShipmentId:
+        type: string
+        format: uuid
+        example: c56a4180-65aa-42ec-a945-5fd21dec0538
+        readOnly: true
+      ppmShipment:
+        $ref: 'definitions/PPMShipment.yaml'
```
ronaktruss, line 1017: **Question/thought (non-blocking)**: Do we want/need this relationship to exist on this side?
This is similar to the relationship between mtoShipments and ppmShipments where we generally have the ppmShipment information as part of the  mtoShipment object but not vice versa.

pkg/handlers/internalapi/weight_ticket_test.go
```diff
@@ -0,0 +1,227 @@
+package internalapi
```
pearl-truss, line 1: **Non blocking suggestion**: I noticed there's some extra work happening in the tests because the `weightTicketCreator` and `updateWeightTicket` services are mostly being used directly. I think if you wanted to avoid extra code and needing the extra parameter in the `makeCreateSubsetData` function you could use the mocked services. There are examples in the [mto_service_item](https://github.com/transcom/mymove/blob/master/pkg/handlers/primeapi/mto_service_item_test.go) test suite

pkg/handlers/internalapi/weight_ticket.go
```diff
@@ -0,0 +1,142 @@
...
+			weightTicket, err := h.weightTicketCreator.CreateWeightTicket(appCtx, ppmShipmentID)
+
+			if err != nil {
+				appCtx.Logger().Error("internalapi.CreateWeightTicketHandler", zap.Error(err))
+				// Can get a status error
+				// Can get an DB error - does the weight ticket, doc create?
+				// Can get an error for whether the PPM exist
+				switch err.(type) {
+				case apperror.InvalidInputError:
+					return weightticketops.NewCreateWeightTicketUnauthorized(), err
```
pearl-truss, line 54: `apperror.InvalidInputError` should return an UnprocessableEntity error. There's some documentation on this https://github.com/transcom/mymove-docs/blob/main/docs/api/guides/api-errors.md#error-types-and-error-responses

duncan-truss:
It looks like the test coverage dropped with the added files causing the CI server_tests to fail.

Here is the [coverage artifact that was saved](https://output.circle-artifacts.com/output/job/d5305466-d054-4443-b7e0-43d559029e69/artifacts/0/test-results/gotest/app/go-coverage.html#file172), you can see the untested paths in red.

I think if you add tests for some of the error cases that should bring that file percentage up from 68%

pearl-truss (approved):
Thanks for working on this! I'm putting approve now since all of the blocking comments have been addressed and this should be ready for merging after the additional test are added to fix the ci issue

swagger-def/internal.yaml
duncan-truss: It's probably not hurting anything to keep it the main point is the eTag. I know the createdAt field is used primarily for sort order sometimes.

PR #8842, opened 2022-06-30
[MB-12949] add aws logging mechanism

chtakahashi:
Adds a mechanism to log failures to a cloudwatch log group called `aws/lambda/circleci-logging`

### DIFF
--- a/scripts/circleci-announce-broken-branch
+++ b/scripts/circleci-announce-broken-branch
@@ -45,6 +45,11 @@ echo
 curl -X POST --data-urlencode payload="$slack_payload" "$SLACK_WEBHOOK_URL"
 
 #####
+## Log failure to AWS. Failure logs can be found in the log group `aws/lambda/circleci-logging`
+#####
+aws events put-events --entries '[{ "Source": "circleci.logging", "DetailType": "CircleCI failure", "Detail": "{ \"message\": \"'"$message"'\" }"}]'
+
+#####
 ## Exit and do nothing else unless a deploy job
 #####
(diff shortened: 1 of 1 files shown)


sheenamt (approved):

scripts/circleci-announce-broken-branch
```diff
@@ -44,6 +44,11 @@ echo
 
 curl -X POST --data-urlencode payload="$slack_payload" "$SLACK_WEBHOOK_URL"
 
+#####
+## Log failure to AWS. Failure logs can be found in the log group `aws/lambda/circleci-logging`
+#####
+aws events put-events --entries '[{ "Source": "circleci.logging", "DetailType": "CircleCI failure", "Detail": "{ \"message\": \"'"$message"'\" }"}]'
```
ahobson, line 50: It seems this is getting the AWS credentials from the environment. But sometimes we override those variables when deploying to particular environment, so it seems like the logs would go to different places depending on which step was being executed. Is that what we want?

chtakahashi: Since the workflow on which the circleCI run fails corresponds to a specific environment, we've set up the same infrastructure with the same resource names in each environment to handle the logging. So if you choose exp as your dp3 environment, the failure will be logged to that log group in the exp environment, and for demo, it will be in demo.

ahobson: But what if the part that fails is the office integration test, before the environment has been chosen? Which AWS environment is the default?

ahobson (approved):

PR #8857, opened 2022-07-06
[ MB-12535 ] Fix the broken test steps

rogeruiz:
## Summary

Tied to the work from #8836. It looks like this was missed when I was
testing things. I believe it happened because I had not rebased my work
onto the current state of the default branch at the time.

## Reviewing this patch

Check that all the tests are passing.


reggieriser (approved):
LGTM!

PR #8868, opened 2022-07-07
MB-12334 Remove zeplin and abstract from repo

felipe-lee:
## Summary

I was trying to fix an issue with a zeplin plugin downstream dependency but I [talked with design](https://ustcdp3.slack.com/archives/CNTDW9Q1Z/p1657225757658969) to see if I could just remove the plugin since we're moving away from zeplin anyway. The decision was that we're good to go with removing this so I removed this and abstract while I was at it since it looks like we're not ever using that plugin anyways (was commented out in storybook config).

## Testing the changes

The zeplin part was a cli tool and since it was removed, there's not really something to test. For abstract we weren't using the plugin anyways, so also nothing to really test.


felipe-lee:
I realized after shutting down for the day that I forgot to remove some commands for zeplin from the package.json. I'll push up a change with that tomorrow. The rest of this should still be reviewable though.

AnnaGingle (approved):
Looks good to me!

PR #8882, opened 2022-07-11
[MB-12493] backend for evaluation reports page

mboren:
## Summary
Added an endpoint to get a list of evaluation reports for a move.
This endpoint will show submitted reports created by any user, and draft reports created by the current office user.

we have a lot of inconsistency between paths, eg `/move/...` , `/moves/...`, `/move-task-orders/...`, `/move_task_orders/...`
And in some paths we use locator and in others we use move id.
I went with `/moves/{moveID}/...` for this one,

I feel like `moves` would be the right one to standardize on, and it seems like it probably makes more sense to use IDs instead of locators for non-user-facing URLs, so we don't have to do the extra query to the moves table.

Another thing that's up for debate here is the response format. I have it just returning all the reports together, which will require the frontend to figure out which ones are which type since the shipment and counseling reports are displayed in separate sections. It wouldn't be unreasonable to have a response format that breaks them out by type, for example:
```json
{ "counselingReports": [],
  "shipmentReports": []
}
```
I think my current format is a bit RESTier and more consistent with other endpoints, but it wouldn't be hard to persuade me otherwise.

I'm also not sure what exactly to be doing for permissions for this. Should I be using the `x-permissions` header? checking roles in the handler? Not sure what the current pattern is.

## Setup to Run Your Code

- Generate devseed data:
```
make db_dev_e2e_populate
```

- Then start the client and server:
```sh
make client_run
```
```sh
make server_run
```

- Log in to the office app
- Request the evaluation reports for move EVLRPT by pasting the following into your browser console (or use postman or something):

```js
// for counseling reports
fetch("http://officelocal:3000/ghc/v1/move/EVLRPT").then(res => res.json()).then(({id}) => fetch("http://officelocal:3000/ghc/v1/moves/" + id + "/counseling-evaluation-reports-list").then(res => res.json()).then(console.log))
```
```js
// for shipment reports
fetch("http://officelocal:3000/ghc/v1/move/EVLRPT").then(res => res.json()).then(({id}) => fetch("http://officelocal:3000/ghc/v1/moves/" + id + "/shipment-evaluation-reports-list").then(res => res.json()).then(console.log))
```
- There are multiple reports created in the devseed data, but some of them are drafts created by another user, so you should only see the three submitted reports printed to the console:
image

### Backend


pkg/testdatagen/scenario/subscenarios.go
```diff
@@ -213,6 +213,63 @@ func subScenarioCustomerSupportRemarks(appCtx appcontext.AppContext) func() {
 	}
 }
 
+func subScenarioEvaluationReport(appCtx appcontext.AppContext) func() {
```
mboren, line 216: todo: update this with ppm, nts, and ntsr shipments.

swagger-def/ghc.yaml
```diff
@@ -165,6 +165,37 @@ paths:
       description: Returns the history for a given move for a unique alphanumeric locator string
       summary: Returns the history of an identified move
       operationId: getMoveHistory
+  '/moves/{moveID}/evaluation-reports':
```
leia-sefkin, line 168: i like your reasoning behind choosing moveID rather than locator, and the choice of moves seems good too.

r.e. the question of response structure, i would recommend a few tweaks:
* split the route into two, one for get a list of counseling reports, one for get a list of shipment reports
* from a FE standpoint for this story we don't need to load the entire report data, we only need enough to populate the table. this means you could simplify your code a little just to retrieve that subset of info, and update the route name to something like `report-list` or `report-info`.

leia-sefkin (commented):
Left some comments and suggestions inline, interested to hear what you think. Not fully decided but two separate routes feels more restful, we can also convo with Patrick when he's back in action since he'll bring a good FE view. 

Otherwise the code is looking good so far

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -103,6 +103,57 @@ func CustomerSupportRemarks(customerSupportRemarks models.CustomerSupportRemarks
...
+
+	var inspectionType *ghcmessages.EvaluationReportInspectionType
+	if evaluationReport.InspectionType != nil {
+		foo := ghcmessages.EvaluationReportInspectionType(*evaluationReport.InspectionType)
+		inspectionType = &foo // :(
+	}
+	var location *ghcmessages.EvaluationReportLocation
+	if evaluationReport.Location != nil {
+		foo := ghcmessages.EvaluationReportLocation(*evaluationReport.Location)
+		location = &foo // :(
```
leia-sefkin, line 123: looks like these foos 🤪 need updating

pkg/services/evaluation_report/evaluation_report_list_fetcher.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	if officeUserID == uuid.Nil {
+		return nil, apperror.NewBadDataError("officeUserID must be provided")
+	}
+
+	err := appCtx.DB().
+		Scope(utilities.ExcludeDeletedScope()).
+		Where("move_id = ?", moveID).
+		Where("type = ?", reportType).
+		Where("(submitted_at IS NOT NULL OR office_user_id = ?)", officeUserID).
```
leia-sefkin, line 33: question: does submitted_at get set on initial creation or only when a report is saved? 

concern: according to our figma designs a report can't be edited once it's submitted. would this query exclude reports in draft state? e.g. if they are set to `submitted_at IS NULL`

swagger-def/ghc.yaml
```diff
@@ -165,6 +165,68 @@ paths:
...
+    parameters:
+      - description: Code used to identify a move in the system
+        in: path
+        name: moveID
+        required: true
+        type: string
+        format: uuid
+    get:
+      produces:
+        - application/json
```
leia-sefkin, line 178: TODO: add a response body ref here describing the response structure

leia-sefkin (changes requested):
Did another review after splitting out the endpoints and like the changes overall. I saw a few places where there's some cleanup needed with temp variables and documentation. 

I am also curious to hear from @Ryan-Koch or @PDickeyTruss what they think about the response structure for these endpoints [The design](https://www.figma.com/file/5AhuEkaH3bJVrtffU3wp22/Milmove-Office-App?node-id=1938%3A36651) shows only a small subset of report info is needed (report ID, date submitted, location, violations (bool), serious incident (bool)). Should we return the full report data to the FE or just query for what we need and only load the full set when we go to resume or view?

pkg/services/evaluation_report/evaluation_report_list_fetcher.go
Ryan-Koch: Eventually we'll want this to exclude draft reports that the user didn't create. We can address that extra layer of logic later though (likely in: https://dp3.atlassian.net/browse/MB-12887), for now I'd be okay with spitting out all of the reports including those in draft that anyone has created.

mboren: submitted_at should only get set when a report is submitted.
The query is supposed to exclude all queries in draft state unless you are logged in as the same office user that created them.

pkg/services/evaluation_report/evaluation_report_list_fetcher_test.go
```diff
@@ -0,0 +1,159 @@
...
+		fetcher := NewEvaluationReportListFetcher()
+		move := testdatagen.MakeDefaultMove(suite.DB())
+		officeUser := testdatagen.MakeOfficeUser(suite.DB(), testdatagen.Assertions{})
+		reports, err := fetcher.FetchEvaluationReports(suite.AppContextForTest(), models.EvaluationReportTypeCounseling, move.ID, officeUser.ID)
+		suite.NoError(err)
+		suite.Empty(reports)
+	})
+	suite.Run("fetch for nonexistent move or office user should not error", func() {
+		// if we want to detect this error, it will take another query to the moves table
+		// would that be worth it?
```
Ryan-Koch, line 24: I think this is okay. Returning the empty list seems like a reasonable approach here. If this were an external API it might be worth the extra feedback around what caused the empty list, but since it's only our client using this I don't think it's necessary.

swagger-def/ghc.yaml
mboren: I'm not totally sure what you're referring to here

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
mboren: I couldn't see a way to avoid using the temporary variables, so i renamed these to something more reasonable and got rid of my frowny face comments.

pkg/services/evaluation_report/evaluation_report_list_fetcher_test.go
mboren: Thanks. I agree. I reworded this comment to be more of an explanation than a question.

mboren:
re: how much info we should include in the response:
I'm happy to go in either direction with this.
- I have a strong personal preference for including just the minimum that we need for the frontend
- I think on milmove we generally include whole objects so that we can share swagger definitions between endpoints, so this is what I went with. But I'm not sure if that's an explicit convention, or more of a vibe that I've inferred.

swagger-def/ghc.yaml
mboren: I implemented the suggestion to split into two routes, I think it is nicer. Still not totally sure how much to include in the response, I left another comment about that down below.

leia-sefkin (approved):

leia-sefkin:
thanks for all the hard work and conversation around how to approach this sensibly. changes look great, let's 🚢 IT

PR #8893, opened 2022-07-13
Add info msg about how to solve to point people in the right direction

ronaktruss:
## Summary
When test coverage fails this adds a message linking to our documentation about test coverage so folks know how to go about diagnosing and resolving the issue.

Added because this is a relatively new check and I don't think everyone knows that there is relevant documentation.


ahobson (approved):
👏

PR #8917, opened 2022-07-18
MB-12919 weight ticket integration

ronaktruss:
## Summary

This integrates the previous weight ticket stories. You should now be able to see the following working as expected on the weight ticket page:
- file upload
- file delete
- hard refreshing
- the mto_shipment api should return weight tickets
- state management

There are a decent number of changes though the commits should help give clarity around the specific changes that were made. I'd suggest starting from the commits on Jul 20, 2022 and moving to the present day. (The changes prior to that were some minor clean up and a few merges).

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the customer application
2. Login as `actualPPMDateZIPAdvanceDone@ppm.approved`
3. Navigate to the weight ticket page and test uploading a file, deleting a file, submitting, and hard refreshing (hard refreshing should persist the document changes).

### Backend


swagger-def/prime.yaml
```diff
@@ -1826,7 +1826,7 @@ definitions:
         example: MTO Shipment not good enough
         x-nullable: true
       ppmShipment:
-        $ref: 'definitions/PPMShipment.yaml'
```
ronaktruss, line 1829: Having weight tickets here causes yaml issues (spectral linting errors) and from a business rule/security perspective we probably don't want to expose the uploads to the prime.

pkg/handlers/internalapi/internal/payloads/model_to_payload.go
```diff
@@ -161,14 +104,15 @@ func PPMShipment(ppmShipment *models.PPMShipment) *internalmessages.PPMShipment
...
+		WeightTickets:                  WeightTickets(storer, ppmShipment.WeightTickets),
 		ETag:                           etag.GenerateEtag(ppmShipment.UpdatedAt),
 	}
 
 	return payloadPPMShipment
 }
 
 // MTOShipment payload
-func MTOShipment(mtoShipment *models.MTOShipment) *internalmessages.MTOShipment {
+func MTOShipment(storer storage.FileStorer, mtoShipment *models.MTOShipment) *internalmessages.MTOShipment {
```
duncan-truss, line 115: Yea this is where deeply nested models gets a bit bloated passing down the storer.

Maybe one way alternative possibility would be:
```go
// mto_shipment.go
WithWeightTickets(mtoShipment.ppmShipment.weightTickets, payloads.MTOShipment(mtoShipment), storer)
```

```go
WithWeightTickets(weightTickets models.WeightTickets, payload internalmessages.MTOShipment, storer storage.FileStorer) {
    return payload.ppmShipment.WeightTickets: WeightTickets(storer, weightTickets)
}
```

swagger-def/definitions/PPMShipment.yaml
```diff
@@ -179,6 +179,10 @@ properties:
     format: cents
     x-nullable: true
     x-omitempty: false
+  weightTickets:
+    items:
+      $ref: '../internal.yaml#/definitions/WeightTicket'
```
duncan-truss, line 184: I think I know why you are doing this but it might be good to explain if the reasoning is related to how it would get compiled into the other apis and require more refactoring?

Ideally we would define the WeightTicket definition and Document definitions at this level?

pkg/handlers/internalapi/weight_ticket_test.go
```diff
@@ -173,8 +174,11 @@ func (suite *HandlerSuite) TestUpdateWeightTicketHandler() {
 			IfMatch:        eTag,
 		}
 
+		handlerConfig := handlers.NewHandlerConfig(appCtx.DB(), appCtx.Logger())
```
duncan-truss, line 177: We'll probably need this in other test files when we do pro gear and expenses, you can define it in the handler suite like the ghc api does https://github.com/transcom/mymove/blob/6e027a8eb2318676a89a414dfc1e39475af0473f/pkg/handlers/ghcapi/api_test.go#L36

swagger-def/definitions/PPMShipment.yaml
ronaktruss: >I think I know why you are doing this but it might be good to explain if the reasoning is related to how it would get compiled into the other apis and require more refactoring?

Yeah moving the weightTicket objects into the `PPMShipment.yaml` file would require additional rework.

> Ideally we would define the WeightTicket definition and Document definitions at this level?

They seem to fit better within the ppmShipment context and I probably would place them there if it wasn't for the above.

>There's definitely work here to strip the Payload name from the Document definition and convert the property names to camel case. A first step may be to keep Orders using the old DocumentPayload while we introduce a new definition if we don't want to take on that orders refactoring work ATM. At least we won't be proliferating it's usage further.

I think even that rework requires a decent amount of effort because there is shared/existing document code that we are leveraging in the model to payload conversions.

src/components/Customer/PPM/Closeout/WeightTicketForm/WeightTicketForm.jsx
```diff
@@ -41,7 +41,7 @@ const validationSchema = Yup.object().shape({
     }),
```
duncan-truss, line 41: I think I didn't quite get this schema validation right because you can enter zeros for both the empty and full weight and it doesn't trigger the validation because zero is falsey so the min schema isn't applied.  The API then responds with a validation error.  I thought it was possible to enter zero for an empty weight but the API is expecting a value greater than zero.  Maybe reconfirm with design and product.

I think it should be something like:

```javascript
.when('emptyWeight', (emptyWeight, schema) => {
  return emptyWeight != null
     ? schema.min(emptyWeight + 1, 'The full weight must be greater than the empty weight')
: schema;
}
```

ronaktruss: We should allow 0 for the the empty weight. I'll update the rule (should be just a `<` instead of `<=` in the validation). Nice catch!

I do want to followup on how we want to handle the document uploads. On the backend we allow one upload between the full and empty docs if both have the "I don't have this weight ticket" checked. On the FE we are always forcing the user to have both uploads.

src/pages/MyMove/PPM/Closeout/WeightTickets/WeightTickets.jsx
```diff
@@ -120,20 +101,20 @@ const WeightTickets = () => {
 
   const handleSubmit = async (values, { setSubmitting }) => {
     setErrorMessage(null);
-    const hasOwnTrailer = values.hasOwnTrailer === 'true';
-    const trailerMeetsCriteria = hasOwnTrailer ? !!values.trailerMeetsCriteria : false;
+    const ownsTrailer = values.ownsTrailer === 'true';
+    const trailerMeetsCriteria = ownsTrailer ? !!values.trailerMeetsCriteria : false;
```
duncan-truss, line 105: I think I bungled this logic too as I saved it initially as yes and when I tried to come back and switch it to no and it didn't work.

`const trailerMeetsCriteria = ownsTrailer ? values.trailerMeetsCriteria === 'true' : false` is what we want I think.

pkg/handlers/internalapi/moves.go
```diff
@@ -46,7 +46,7 @@ func payloadForMoveModel(storer storage.FileStorer, order models.Order, move mod
 	var hhgPayloads internalmessages.MTOShipments
 	for _, hhg := range move.MTOShipments {
 		copyOfHhg := hhg // Make copy to avoid implicit memory aliasing of items from a range statement.
-		payload := payloads.MTOShipment(&copyOfHhg)
+		payload := payloads.MTOShipment(storer, &copyOfHhg)
```
duncan-truss, line 49: just pointing out that the naming here of HHG is inaccurate now that PPMs and other shipment types could be mixed in here.

src/pages/MyMove/PPM/Closeout/WeightTickets/WeightTickets.jsx
```diff
@@ -85,33 +72,27 @@ const WeightTickets = () => {
...
-    const newValue = {
-      id: uuidv4(),
-      created_at: '2022-06-22T23:25:50.490Z',
-      bytes: upload.file.size,
-      url: 'a/fake/path',
-      filename: upload.file.name,
-      content_type: upload.file.type,
-    };
-
-    setFieldValue(fieldName, [...values[`${fieldName}`], newValue]);
```
duncan-truss, line 103: Was this no longer necessary?

I am noticing that the file gets added to the uploads table while it's still showing an upload in progress in the File Pond component.  Is that what's happening here that the shipment data is being updated in Redux and the form is re-rendering?

ronaktruss: Yeah, on file upload we enter the `handleCreateUpload` function and more specifically when the API call that is triggered from the `createUploadForDocument` func is successful we are updating the redux state that causes the form to re-render.

duncan-truss:
@ronaktruss One thing is we don't remove and of the existing uploads if the user changes their answer on trailer ownership/meets criteria.  I guess it's good they could change their mind and the uploads would still exist.  Maybe on them actually completing closeout we could remove them otherwise we'll have to filter them out in the fetcher backend so the office user doesn't review them.

src/pages/MyMove/PPM/Closeout/WeightTickets/WeightTickets.jsx
duncan-truss: Thanks, I think adding and deleting uploads in storybook is currently broken then.  We may have to live with that, I don't know that we should spend any time on a workaround but that was one nice to have.

duncan-truss (approved):

ronaktruss:
> @ronaktruss One thing is we don't remove and of the existing uploads if the user changes their answer on trailer ownership/meets criteria. I guess it's good they could change their mind and the uploads would still exist. Maybe on them actually completing closeout we could remove them otherwise we'll have to filter them out in the fetcher backend so the office user doesn't review them.

Yeah that's definitely one situation that we don't "gracefully" handle. I was under the impression that we would still be sending that information and have the counselor modify/reject it if needed. Personally I'd probably advocate for preventing the user from submitting with a proof document if they have no, but I recall that being not part of the initial roll out.

duncan-truss:
You can also probably delete this hardcoded data I added for review screen acceptance testing https://github.com/transcom/mymove/blob/227fbf14bd906ca9dd1383ac1fa83c002c26826e/src/pages/MyMove/PPM/Closeout/Review/Review.jsx#L65-L79

scarequotes (approved):
Thumbs up!

PR #8918, opened 2022-07-18
[MB-12728] Cypress test for Weight Tickets

pearl-truss:
## Summary

This PR adds cypress test for the weight ticket screen.


cypress/integration/milmove/ppms/weightTickets.js
```diff
@@ -0,0 +1,62 @@
...
+
+describe('Weight Tickets', function () {
+  before(() => {
+    cy.prepareCustomerApp();
+  });
+
+  beforeEach(() => {
+    cy.intercept('GET', '**/internal/moves/**/mto_shipments').as('getShipment');
+    cy.intercept('PATCH', '**/internal/mto-shipments/**').as('patchShipment');
+  });
```
felipe-lee, line 15: **thought**: I think we might also want to intercept the weight ticket patch call so we know when the submission of that data is done. Maybe also for the creation if we want to know when we can start filling out data on the page, though idk if that one is fully needed or not.

felipe-lee (commented):
**question**: I saw you added and then removed the `scrollIntoVIew` method call on the "Upload PPM Documents" button. I'm not sure if it wasn't working, but I was going to say that if we could get that working in general in our cypress tests that would be awesome. There's so many screenshots that aren't that useful because the element we're looking for is off screen.

cypress/support/ppmCustomerShared.js
```diff
@@ -55,17 +55,79 @@ export function fillOutAboutPage(selectAdvance) {
...
+    cy.get('input[name="emptyWeight"]').clear().type('1000').blur();
+    cy.get('input[name="missingEmptyWeightTicket"]').check({ force: true });
+    cy.intercept('/internal/uploads').as('uploadFile');
+    cy.upload_file('.emptyDocument.filepond--root', 'constructedWeight.xls');
+    cy.wait(10000);
+    cy.get('input[name="fullWeight"]').clear().type('3000');
+    cy.get('input[name="missingFullWeightTicket"]').check({ force: true });
+    cy.intercept('/internal/uploads').as('uploadFile');
+    cy.upload_file('.fullDocument.filepond--root', 'constructedWeight.xls');
+    cy.wait(10000);
```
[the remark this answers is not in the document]
pearl-truss: I tried this initially and the uploads were failing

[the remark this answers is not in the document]
ronaktruss: I was playing around with this a bit and we can get rid of the explicit wait by doing something like
```js
cy.intercept('/internal/uploads**').as('uploadFile');
cy.upload_file('.emptyDocument.filepond--root', 'sampleWeightTicket.jpg');
cy.wait('@uploadFile');
``` 
the regular `cy.intercept('/internal/uploads').as('uploadFile');` doesn't work because we append a querystring to the end of this api call with the document id.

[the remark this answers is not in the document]
pearl-truss: My latest commit doesn't have the wait

cypress/support/ppmCustomerShared.js
```diff
@@ -55,17 +56,82 @@ export function fillOutAboutPage(selectAdvance) {
...
+  navigateFromWeightTicketPage();
+}
+
+export function fillOutWeightTicketPage(options) {
+  cy.get('input[name="vehicleDescription"]').clear().type('Kia Forte').blur();
+
+  if (options?.useConstructedWeight) {
+    cy.get('input[name="emptyWeight"]').clear().type('1000').blur();
+    cy.get('input[name="missingEmptyWeightTicket"]').check({ force: true });
+    cy.intercept('/internal/uploads').as('uploadFile');
```
ronaktruss, line 90: Are these lines `cy.intercept('/internal/uploads').as('uploadFile');` still needed?

pearl-truss: Yeah I think so, removing them passed locally but in ci the removal had the test fail and flagged as flakey

ronaktruss: Yeah I saw that. I think the errors were unrelated to this change though. Most of the errors were about the urls not matching which happens before we even get to the intercept part of the tests.

cypress/support/ppmCustomerShared.js
```diff
@@ -55,17 +56,82 @@ export function fillOutAboutPage(selectAdvance) {
...
+        1,
+      );
+    } else {
+      cy.get('input[name="trailerMeetsCriteria"][value="false"]').check({ force: true });
+    }
+  }
+}
+
+export function navigateFromWeightTicketPage() {
+  cy.get('button').contains('Save & Continue').click();
```
ronaktruss, line 128: I wonder/think the current errors might be resolved if we do this
```suggestion
  cy.get('button').contains('Save & Continue').should('be.enabled').click();
```

pearl-truss: Okay I'll try this. I was looking at the mp4 on circle and looks like the upload isn't finished before cypress tries to click this button, I wonder if we do need a wait for upload.

cypress/support/ppmCustomerShared.js
```diff
@@ -55,17 +56,82 @@ export function fillOutAboutPage(selectAdvance) {
...
+
   cy.location().should((loc) => {
     expect(loc.pathname).to.match(/^\/moves\/[^/]+\/shipments\/[^/]+\/weight-tickets/);
   });
 }
 
+export function signInAndNavigateToWeightTicketPage(userId) {
+  cy.apiSignInAsUser(userId);
+  cy.wait('@getShipment');
+  cy.screenshot();
```
ronaktruss, line 74: Do we still need this / do we know why we need to wait here?

I think this might be because we have the loading component being displayed at some point (ie. the Loading, please wait... screen). If this is the case we might be able to just add a `should` that verifies some of the actual content of the page post the loading component.

cypress/support/ppmCustomerShared.js
pearl-truss: Yeah now it's failing `expected '<button.usa-button.PPM_saveButton__n3eZE>' to be 'enabled'`

I think I need to add the wait back since it was passing before I took that out

cypress/support/ppmCustomerShared.js
pearl-truss: Based on the error I got when the wait was removed and replaced with should, I don't think this would work

ronaktruss (approved):
This looks good to me outside of the usage of the screenshot. Since it's a bit of a workaround, it would be nice to try and resolve the root issue. 

I'm happy to spend some time trying to look into that, if you want to work on your other story. Or we could create a ticket to capture this work

pearl-truss:
> This looks good to me outside of the usage of the screenshot. Since it's a bit of a workaround, it would be nice to try and resolve the root issue.
> 
> I'm happy to spend some time trying to look into that, if you want to work on your other story. Or we could create a ticket to capture this work

Oh okay, your comment mentioned the wait statement (which we do need). We may be able to get rid of the screenshot. I'll look into it

ronaktruss:
> > > This looks good to me outside of the usage of the screenshot. Since it's a bit of a workaround, it would be nice to try and resolve the root issue.
> > > I'm happy to spend some time trying to look into that, if you want to work on your other story. Or we could create a ticket to capture this work
> > 
> > 
> > Oh okay, your comment mentioned the wait statement (which we do need). We may be able to get rid of the screenshot. I'll look into it
> 
> I haven't been able to figure this out. You can spend time on this if you'd like otherwise agreed that we can capture this in another ticket

I'll spend some time looking into it this afternoon.

PR #8935, opened 2022-07-21
MB-13177 add moving expenses table and model

felipe-lee:
## Summary

This adds a new `MovingExpense` model and a corresponding `moving_expenses` table that will be used to track moving expenses for PPMs.

Do the field names and comments on the table columns make sense?

One thing to call out:
There is an existing type called `MovingExpenseType`, which was used for the old `MovingExpenseDocument` model, so we can't use that name in the new type for our new model. I thought about renaming the old one since we're going to get rid of it, but it has a lot of references, so for the sake of getting this model done, I created the new type with a slightly different name (`MovingExpenseReceiptType`). I still used the shorter name at the table level and for the model field itself because I figure when we get rid of the old type, we can rename the new type more easily. If it's preferred though, we can just use the new type name for the field itself too at both the table and model level.

## Checking Changes

<details>
<summary>💻 Terminal setup</summary>

### Run migrations on top of existing db

You can run this the first time you pull it down to see it applies the migration properly on top of the existing DB. Not required though, you can also run the other command instead.

```sh
make db_dev_migrate
```

### Run all migrations

```sh
make db_dev_fresh
```

</details>

### Additional steps 

1. Access the DB however you usually do and see if the new table is there correctly


migrations/app/schema/20220720214641_create_moving_expenses_table.up.sql
```diff
@@ -0,0 +1,53 @@
...
+	'OIL',
+	'PACKING_MATERIALS',
+	'RENTAL_EQUIPMENT',
+	'STORAGE',
+	'TOLLS',
+	'WEIGHING_FEES',
+	'OTHER'
+	);
+
+CREATE TYPE moving_expense_status AS enum (
```
duncan-truss, line 12: I wonder if we should make this a common type across weight tickets/pro-gear/expenses instead of them all having their own statuses?  If you're adding status we should probably add `reason` while we're at it.

migrations/app/schema/20220720214641_create_moving_expenses_table.up.sql
```diff
@@ -0,0 +1,53 @@
...
+		CONSTRAINT moving_expenses_document_id_fkey
+			REFERENCES documents,
+	moving_expense_type moving_expense_type,
+	description varchar,
+	paid_with_gtcc bool,
+	amount int,
+	missing_receipt bool,
+	status moving_expense_status,
+	sit_start_date timestamptz,
+	sit_end_date timestamptz,
```
duncan-truss, line 34: We could make these `date` types if they have no time component.

migrations/app/schema/20220720214641_create_moving_expenses_table.up.sql
felipe-lee: Good point, I'll update it to be more generic. Ah thanks for catching that, I totally forgot to add that field.

pkg/models/moving_expense.go
```diff
@@ -0,0 +1,98 @@
...
+type MovingExpenses []MovingExpense
+
+// Validate gets run every time you call a "pop.Validate*" (pop.ValidateAndSave, pop.ValidateAndCreate,
+// pop.ValidateAndUpdate) method. This should contain validation that is for data integrity. Business validation should
+// occur in service objects.
+func (m *MovingExpense) Validate(_ *pop.Connection) (*validate.Errors, error) {
+	var movingExpenseType *string
+
+	if m.MovingExpenseType != nil {
+		movingExpenseType = StringPointer(string(*m.MovingExpenseType))
```
duncan-truss, line 65: I think you can simplify this to be:

`movingExpenseType = (*string)(m.MovingExpenseType)`

pkg/models/moving_expense.go
```diff
@@ -0,0 +1,98 @@
...
+		&OptionalStringInclusion{Name: "MovingExpenseType", Field: movingExpenseType, List: []string{
+			string(MovingExpenseReceiptTypeContractedExpense),
+			string(MovingExpenseReceiptTypeOil),
+			string(MovingExpenseReceiptTypePackingMaterials),
+			string(MovingExpenseReceiptTypeRentalEquipment),
+			string(MovingExpenseReceiptTypeStorage),
+			string(MovingExpenseReceiptTypeTolls),
+			string(MovingExpenseReceiptTypeWeighingFees),
+			string(MovingExpenseReceiptTypeOther),
+		}},
```
duncan-truss, line 87: This may be a good use case for converting to generics going forward.

Maybe you can define the allowed list once so it's not recalculated everytime validate is called?  Here's an example of how it's done for EDI https://github.com/transcom/mymove/blob/8b70122799d4b1ed8e76ea5079b9685120df6fce/pkg/models/edi_type.go#L17-L22

duncan-truss (approved):

pkg/models/moving_expense.go
felipe-lee: Oh nice, I didn't know you could do this. I'll update!

pkg/models/moving_expense.go
felipe-lee: Yeah, I feel like there's several things like this that would benefit from generics. I've updated now to declare both of the allowed lists at the pkg level. I made them exportable so they're available in tests too (since we have our model tests as a separate package)

PR #8940, opened 2022-07-21
[MB-13510] Add gci linter to make go import order/grouping more consistent

reggieriser:
## Summary

_Note: I had a similar draft PR #8593 a few months ago (now closed in lieu of this one) that attempted to do something similar.  At the time, I ran into some `gci` issues that caused me to abandon it until some upstream fixes were made._

I've noticed that the order of our go imports has seemed a little random and getting more so over time.  Ideally, the imports would be organized as follows (alphabetically within a section):

```
import (
    [system packages]

    [third-party packages]

    [project packages]
)
```

However, when a blank line is added between imports, `goimports` won't regroup the grouped imports to be with other similar groups.  In this PR, I used the [gci linter](https://github.com/daixiang0/gci) via `golangci-lint` to do the reorganization which eliminates extra sections and groups like I think we want.

A lot of files have changed, but you should be able to follow the commits as so:

1. **Update all pre-commit hooks; fix issues discovered by updates:** this updates all of our pre-commit hooks (via `pre-commit autoupdate`) to their latest versions, then fixes a few new issues that popped up from the updated versions of `golangci-lint` and `markdownlint`.  I needed the latest `golangci-lint` for `gci` to work properly.
1. **Update nix config to use pre-commit 2.20:** don't think that was required, but I noticed the `nix` config was several versions behind a brew'd version as well as what we use in the `circleci-docker` repository.
1. **Removed some import duplication that was causing inconsistent import formatting by linter (and just confusing):** fixed some import duplication and import comments that were causing `gci` to flip import order back and forth when run.  I assume it was unintentional (although legal) to have the import duplication.
1. **Add gci linter to golangci-lint to format order/grouping of go imports:** added the `gci` linter and settings to `golangci-lint` and included the go files where `gci` fixed the import order (appears to be most of them).

To test it, try changing the order of imports on a file or two, do a `git add` on them, then run:
`pre-commit run golangci-lint`

To try it on all files (it shouldn't touch generated files),
`pre-commit run -a golangci-lint`

## Special Notes

- one of the new linter issues I had to fix involved setting a `ReadHeaderTimeout` in addition to the existing `IdleTimeout` (`gosec` rule `G112: Potential slowloris attack`).  Since it wasn't set previously, I think there was no timeout.  I set it to 60 seconds (half of the 120 second idle timeout), but I could use some advice on if another value makes more sense.
- ~~there is a performance problem when running the `revive` linter under `golangci-lint` in the latest 1.47.*+ versions.  Temporarily, `golangci-lint` [has turned off some `revive` rules](https://github.com/golangci/golangci-lint/issues/2997#issuecomment-1190681977) as a workaround until it can be fixed upstream.  I'm hoping this might be fixed by the time we would consider merging this, so I'm making this a draft PR for now.  If not, we'll have to assess if it makes sense to upgrade now and have some rules turned off for a bit.~~
- the performance problem noted above has been fixed in `revive` and merged into `golangci-lint` (along with reversing the rules that had been temporarily turned off).  However, a new tagged version of `golangci-lint` has not been released yet as of this writing.  This PR is now using a specific hash of `golangci-lint` that is several commits beyond the latest `1.48.0` tagged version.  Leaving this on draft still for the moment to see if a tagged version shows up.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,12 +1,12 @@
 # Contributing to MyMove
 
-Anyone is welcome to contribute code changes and additions to this project. If you'd like your changes merged into the master branch, please read the following document before opening a [pull request](https://github.com/transcom/mymove/pulls).
+Anyone is welcome to contribute code changes and additions to this project. If you'd like your changes merged into the master branch, please read the following document before opening a [pull request][pulls].
 
 There are several ways in which you can help improve this project:
 
-1. Fix an existing [issue](https://github.com/transcom/mymove/issues) and submit a [pull request](https://github.com/transcom/mymove/pulls).
-1. Review open [pull requests](https://github.com/transcom/mymove/pulls).
-1. Report a new [issue](https://github.com/transcom/mymove/issues). _Only do this after you've made sure the behavior or problem you're observing isn't already documented in an open issue._
+1. Fix an existing [issue][issues] and submit a [pull request][pulls].
+1. Review open [pull requests][pulls].
+1. Report a new [issue][issues]. _Only do this after you've made sure the behavior or problem you're observing isn't already documented in an open issue._
 
 ## Table of Contents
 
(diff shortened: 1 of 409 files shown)


pkg/server/server.go
```diff
@@ -141,23 +142,18 @@ func CreateNamedServer(input *CreateNamedServerInput) (*NamedServer, error) {
...
-			Handler:        input.HTTPHandler,
-			IdleTimeout:    idleTimeout,
-			MaxHeaderBytes: maxHeaderSize,
-			TLSConfig:      tlsConfig,
+			Addr:              address,
+			ErrorLog:          newStandardLogger(input.Logger),
+			Handler:           input.HTTPHandler,
+			IdleTimeout:       idleTimeout,
+			MaxHeaderBytes:    maxHeaderSize,
+			ReadHeaderTimeout: readHeaderTimeout,
```
reggieriser, line 155: Original linter issue: `G112: Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server (gosec)`

cmd/prime-api-client/utils/connection.go
```diff
@@ -60,7 +60,6 @@ func CreatePrimeClientWithCACStoreParam(v *viper.Viper, store *pksigner.Store) (
 			MinVersion:         tls.VersionTLS12,
 			MaxVersion:         tls.VersionTLS12,
 		}
-		tlsConfig.BuildNameToCertificate()
```
reggieriser, line 63: Original linter issue (several of these): `SA1019: tlsConfig.BuildNameToCertificate has been deprecated since Go 1.14: NameToCertificate only allows associating a single certificate with a given name. Leave that field nil to let the library select the first compatible chain from Certificates. (staticcheck)`

CONTRIBUTING.md
```diff
@@ -37,17 +37,17 @@ We use a number of tools to evaluate the quality and security of this project's
...
 You may submit contributions anonymously or under a pseudonym if you'd like, but we need to be able to reach you at the email address you provide when agreeing to the DCO. Contributions you make to this public Department of Defense repository are completely voluntary. When you submit a pull request, you're offering your contribution without expectation of payment and you expressly waive any future pay claims against the U.S. Federal Government related to your contribution.
 
-[contributors]: https://github.com/transcom/move.mil/blob/master/CONTRIBUTORS.md
-[issues]: https://github.com/transcom/move.mil/issues
-[license]: https://github.com/transcom/move.mil/blob/master/LICENSE.md
-[pulls]: https://github.com/transcom/move.mil/pulls
+[contributors]: https://github.com/transcom/mymove/blob/master/CONTRIBUTORS.md
+[issues]: https://github.com/transcom/mymove/issues
+[license]: https://github.com/transcom/mymove/blob/master/LICENSE.txt
+[pulls]: https://github.com/transcom/mymove/pulls
```
reggieriser, line 53: Example original linter issue (several in this file): `MD053/link-image-reference-definitions Link and image reference definitions should be needed [Unused link or image reference definition: "contributors"] [Context: "[contributors]: https://github..."]`

reggieriser:
Renamed branch using GitHub UI which auto-closed this PR.  Will open a new one.

PR #8947, opened 2022-07-25
MB-12542 Move and shipment info for evaluation report forms

Ryan-Koch:
## Summary

This adds the ShipmentDisplay component to the ShipmentEvaluationReport page. This also meant needing to add handlers for getting a single evaluation report, a single MTOShipment, and making sure the services infrastructure supported those things effectively. There will be another PR coming to cover adding this information to a counseling evaluation report screen.

## Setup to Run Your Code

<details>
<summary>💻 You will need to use three separate terminals to test this locally.</summary>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the office app.
2. Login as a QAE/CSR.
3. Find a move, if you're looking locally or at an ephemeral deploy you can search for `SITEXT` (move code).
5. Click on the quality assurance tab.
6. Click to create or view a report.
7. Revel in the shipment/move info you see.

## Verification Steps for Author

These are to be checked by the author.

- [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
- [ ] Request review from a member of a different team.
- [ ] Have the Jira acceptance criteria been met for this change?

## Verification Steps for Reviewers

These are to be checked by a reviewer.

### Frontend

- [ ] User facing changes have been reviewed by design.
- [ ] There are no aXe warnings for UI.
- [ ] This works in [Supported Browsers and their phone views](https://github.com/transcom/mymove/tree/master/docs/adr/0016-Browser-Support.md) (Chrome, Firefox, IE, Edge).
- [ ] There are no new console errors in the browser devtools
- [ ] There are no new console errors in the test output
- [ ] If this PR adds a new component to Storybook, it ensures the component is fully responsive, OR if it is intentionally not, a wrapping div using the `officeApp` class or custom `min-width` styling is used to hide any states the would not be visible to the user.

### Backend

- [ ] Code follows the guidelines for [Logging](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide#logging)
- [ ] The requirements listed in [Querying the Database Safely](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide/#querying-the-database-safely) have been satisfied.

#### Any new migrations/schema changes:

[...]

PR #8949, opened 2022-07-25
MB-13147  Create/Delete Shipment Evaluation Reports

PDickeyTruss:
## Summary

This PR adds functionality to allow creation and cancelation (soft delete) of shipment evaluation reports. New reports can be created using the button(s) on the existing quality assurance reports page.  Reports can be canceled using the Cancel button that is present when viewing/editing a single report (`/moves/<moveCode>/evaluation-reports/<reportId>/`). See the below video for a brief example creating/deleting shipment reports.

Note: The balance of when to split out shipment vs counseling reports is pretty hard to get right upfront. I suspect some aspects of this will change when get deeper into the reports and better grok their similarities/differences. 

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Validation
- [x] Shipment eval reports can be created using the button(s) on the quality assurance reports page/tab.
- [x] Reports are created in a draft state as soon as the create button is clicked with no additional save click (or similar) needed
- [x] After creating a report, user is redirected to the report page for editing (no form exists yet)
- [x] Reports can be canceled using the cancel button
- [x] Confirmation modal is shown before canceling a report
- [x] Exiting or not confirming a cancel in the cancel modal does not end up canceling the report
- [x] Canceled reports are not shown in the quality assurance shipments reports table(s)
- [x] An alert is shown on the top of the quality assurance reports page after being redirected to it after canceling a report

https://user-images.githubusercontent.com/62263329/180891133-24c9064d-bdab-45fb-90f1-1ea102525872.mov


swagger-def/ghc.yaml
```diff
@@ -4219,6 +4278,75 @@ definitions:
...
+      submittedAt:
+        type: string
+        format: date-time
+        x-nullable: true
+      createdAt:
+        type: string
+        format: date-time
+  CreateShipmentEvaluationReport:
+    type: object
+    description: 'todo: add description'
```
mboren, line 4341: Missed a todo here

swagger-def/ghc.yaml
```diff
@@ -4219,6 +4278,75 @@ definitions:
...
+      travelTimeMinutes:
+        type: integer
+        minimum: 0
+        x-nullable: true
+      location:
+        $ref: '#/definitions/EvaluationReportLocation'
+        x-nullable: true
+      locationDescription:
+        type: string
+        example: asdfasdfsd
```
mboren, line 4316: here's an example for this field that was in one of the designs: "Route 66 at crash inspection site 3"

swagger-def/ghc.yaml
```diff
@@ -4219,6 +4278,75 @@ definitions:
...
+        type: boolean
+        x-nullable: true
+      remarks:
+        type: string
+        x-nullable: true
+      submittedAt:
+        type: string
+        format: date-time
+        x-nullable: true
+      createdAt:
```
mboren, line 4336: Do we need to include this field? We generally let Pop fill this field automatically with the current time on the backend, but i think it is fine to include if we need to use a different time than that.

mboren: oh this is the response, not the create payload, please ignore me.

swagger-def/ghc.yaml
PDickeyTruss: Good spot and thanks for the example, will update! 🎉

src/components/ConfirmationModals/DeleteEvaluationReportConfirmationModal.stories.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+import ConnectedDeleteEvaluationReportConfirmationModal, {
+  DeleteEvaluationReportConfirmationModal,
+} from 'components/ConfirmationModals/DeleteEvaluationReportConfirmationModal';
+
+export default {
+  title: 'Components/DeleteEvaluationReportConfirmationModal',
+  component: DeleteEvaluationReportConfirmationModal,
+  // args: {
```
Ryan-Koch, line 10: This might be a left over

src/pages/Office/ShipmentEvaluationReport/ShipmentEvaluationReport.jsx
```diff
@@ -0,0 +1,87 @@
...
+import { useMutation } from 'react-query';
+
+import styles from '../TXOMoveInfo/TXOTab.module.scss';
+
+import shipmentEvaluationReportStyles from './ShipmentEvaluationReport.module.scss';
+
+import ConnectedDeleteEvaluationReportConfirmationModal from 'components/ConfirmationModals/DeleteEvaluationReportConfirmationModal';
+import { deleteEvaluationReport } from 'services/ghcApi';
+
+const mtoRefId = 'TODO'; // move?.referenceId
```
Ryan-Koch, line 15: I added taking care of this to https://github.com/transcom/mymove/pull/8947/files#diff-ab1fb1e9fe03e0ba098d3e432b92a32863e4fa7f409aa5bcf97fd59d973fd8e7R98 so I think we can leave the TODO here and catch the reference ID in that PR after this one merges.

src/components/ConfirmationModals/DeleteEvaluationReportConfirmationModal.stories.jsx
PDickeyTruss: Good eye, is for sure not needed ✅

Ryan-Koch (approved):
I see there's a couple of minor comments, but more or less this LGTM

pkg/services/evaluation_report/evaluation_report_creator_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (suite *EvaluationReportSuite) TestEvaluationReportCreator() {
+	creator := NewEvaluationReportCreator()
+
+	suite.Run("Can create customer support report successfully", func() {
+		move := testdatagen.MakeDefaultMove(suite.DB())
+		shipment := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+			MTOShipment: models.MTOShipment{
+				MoveTaskOrderID: move.ID,
+			},
```
mboren, line 19: I think what you've got should work fine, but will create and discard an extra `Move`.
If you pass the move through `Assertions.Move`, it should be a bit more efficient.

```suggestion
			Move: move,
```

pkg/services/evaluation_report/evaluation_report_deleter.go
```diff
@@ -0,0 +1,61 @@
...
+			return apperror.NewNotFoundError(reportID, "attempting to delete evaluationReport")
+		default:
+			return apperror.NewQueryError("evaluationReport", err, "")
+		}
+	}
+
+	sessionUserID := appCtx.Session().OfficeUserID
+
+	if report.OfficeUserID != sessionUserID {
+		appCtx.Logger().Warn("Evaluation reports may only be edited by the user who created them.", zap.String("Evaluation Report ReportID", reportID.String()))
```
mboren, line 39: nitpick: our error logging guidelines say to always use `Error` instead of `Warn`: https://transcom.github.io/mymove-docs/docs/backend/guides/golang-guide#5-dont-fmt-errors-log-instead

leia-sefkin (approved):
Code, tests and storybook all are looking good. I went through the acceptance flow and everything is working well. A huge chunk of work!

pkg/services/evaluation_report/evaluation_report_deleter.go
mboren: actually, upon reading more closely, maybe it should be `Info`

pkg/services/evaluation_report/evaluation_report_deleter_test.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+func (suite *EvaluationReportSuite) TestEvaluationReportDeleter() {
+	setupTestData := func() (services.EvaluationReportDeleter, models.EvaluationReport, appcontext.AppContext) {
+		deleter := NewEvaluationReportDeleter()
+		move := testdatagen.MakeDefaultMove(suite.DB())
+		shipment := testdatagen.MakeMTOShipment(suite.DB(), testdatagen.Assertions{
+			MTOShipment: models.MTOShipment{
+				MoveTaskOrderID: move.ID,
+			},
```
mboren, line 21: we can simplify this a bit and avoid creating an extra move
```suggestion
			Move: move,
```

pkg/services/evaluation_report/evaluation_report_deleter_test.go
```diff
@@ -0,0 +1,70 @@
...
+		})
+		officeUser := testdatagen.MakeOfficeUser(suite.DB(), testdatagen.Assertions{})
+		report := testdatagen.MakeEvaluationReport(suite.DB(), testdatagen.Assertions{
+			EvaluationReport: models.EvaluationReport{
+				ShipmentID: &shipment.ID,
+				Type:       models.EvaluationReportTypeShipment, OfficeUserID: officeUser.ID,
+			},
```
mboren, line 28: This is another case where the testdatagen function is intended to work with flattened assertions. I think your code will work just fine, but it will create an extra office user and move behind the scenes.

```suggestion
			OfficeUser: officeUser,
			MTOShipment: shipment,
			Move: move,
```

pkg/services/evaluation_report/evaluation_report_creator.go
```diff
@@ -0,0 +1,38 @@
...
+	if err != nil {
+		if errors.Cause(err).Error() == models.RecordNotFoundErrorString {
+			return nil, models.ErrFetchNotFound
+		}
+		return nil, err
+	}
+	report.MoveID = shipment.MoveTaskOrderID
+
+	verrs, err := appCtx.DB().ValidateAndCreate(report)
+	if verrs.Count() != 0 || err != nil {
```
mboren, line 33: The model doesn't currently have validations so this is not a big deal, but if it did, it would be nice to pass `verrs` up to the handler, so we can log em and return the appropriate response.

I was looking for a good example to recommend a pattern for this, but it looks like we are pretty inconsistent about it.

I think the simplest way to do it is to return `verrs` and `err` separately, and then in the handler return an `Unprocessable Entity` error if there are verrs, but I'm not sure if that's the Best Way.

Also, you can use `verrs.HasAny()` instead of checking the count.

pkg/handlers/ghcapi/evaluation_report.go
```diff
@@ -67,3 +72,58 @@ func (h GetCounselingEvaluationReportsHandler) Handle(params moveop.GetMoveCouns
...
+	return h.AuditableAppContextFromRequestWithErrors(params.HTTPRequest,
+		func(appCtx appcontext.AppContext) (middleware.Responder, error) {
+
+			payload := params.Body
+			shipmentID := uuid.FromStringOrNil(payload.ShipmentID.String())
+			report := &models.EvaluationReport{
+				ShipmentID: &shipmentID,
+				Type:       models.EvaluationReportTypeShipment,
+				CreatedAt:  time.Now(),
+				UpdatedAt:  time.Now(),
```
mboren, line 96: I don't think you need to set `CreatedAt` or `UpdatedAt` because they get set automatically by Pop.

swagger-def/ghc.yaml
PDickeyTruss: Fixed ✅

swagger-def/ghc.yaml
PDickeyTruss: Fixed ✅

src/components/ConfirmationModals/DeleteEvaluationReportConfirmationModal.stories.jsx
PDickeyTruss: Fixed ✅

mboren:
Looks great! I left a zillion minor comments, but I don't think any of them are blocking.

pkg/handlers/ghcapi/evaluation_report.go
PDickeyTruss: Good spot, makes sense Pop would set those... and probably better not to get in Pops way. 😁 
Fixed ✅

pkg/services/evaluation_report/evaluation_report_creator_test.go
PDickeyTruss: Great suggestion, that not only works better (no extra create/discard) but it also reads way more clearly. 💯 

Fixed ✅

pkg/services/evaluation_report/evaluation_report_deleter.go
PDickeyTruss: Good 👁️, fixed ✅

pkg/services/evaluation_report/evaluation_report_deleter_test.go
PDickeyTruss: Fixed ✅

pkg/services/evaluation_report/evaluation_report_deleter_test.go
PDickeyTruss: Another great suggestion, fixed ✅

cosburn (approved):

PR #8956, opened 2022-07-28
MB-13293: add testdatagen functions for moving expense

ronaktruss:
## Summary

This adds the testdatagen functions for making a moving expense. Make moving expense implicitly creates a ppmshipment that is approved and a completed weight ticket.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

To test this app please reset the db and populate the latest dev seed data. You can do that by:
```sh
make db_dev_reset
make db_dev_e2e_populate
```

In the database you should then verify that a record has been created for the moving_expenses table that matches the data that is created in https://github.com/transcom/mymove/pull/8956/files#diff-e24b52a37c80dbc5e9d03237bb69a26607a061212b3ecf3e17636d9cae57448cR1060-R1101

### Backend


pkg/testdatagen/shared.go
```diff
@@ -304,3 +305,79 @@ func CurrentDateWithoutTime() *time.Time {
 
 	return &currentDate
 }
+
```
felipe-lee, line 308: **praise**: thanks for moving these to the shared file! Makes it less confusing later than if we're using functions from another model creation file haha.

pkg/testdatagen/make_moving_expenses.go
```diff
@@ -0,0 +1,68 @@
...
+	packingMaterialType := models.MovingExpenseReceiptTypePackingMaterials
+	amountPaid := unit.Cents(2345)
+
+	fullAssertions := Assertions{
+		MovingExpense: models.MovingExpense{
+			DocumentID:        document.ID,
+			Document:          document,
+			MovingExpenseType: &packingMaterialType,
+			Description:       models.StringPointer("Packing Peanuts"),
+			PaidWithGTCC:      models.BoolPointer(true),
```
felipe-lee, line 45: **question**: is this more likely than `false`? Not hugely important, but I'm curious which would be a better default.

felipe-lee (approved):
Thanks for getting these up. I left a question, but it's not really blocking

pkg/testdatagen/make_moving_expenses.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"github.com/gobuffalo/pop/v6"
+
+	"github.com/transcom/mymove/pkg/models"
+	"github.com/transcom/mymove/pkg/unit"
+)
+
+func MakeMinimalMovingExpense(db *pop.Connection, assertions Assertions) models.MovingExpense {
+	weightTicket := checkOrCreateWeightTicket(db, assertions)
```
duncan-truss, line 11: Did you mean for this to reuse or create the PPM Shipment?

pkg/testdatagen/make_moving_expenses.go
ronaktruss: I know that some branches will tell folks to use their GTCC ([ref](https://dp3.atlassian.net/wiki/spaces/MT/pages/1070956545/Outcome+7+Personally+Procured+Moves+PPM#Description)), but let me ask.

pkg/testdatagen/make_moving_expenses.go
ronaktruss: Can you clarify/elaborate?

I'm not sure if you mean this to be either the following or something else:
1. should this existing call reuse or create the ppmShipment - yeah, that's the intention. Is there a reason that we should not do this?
2. should this call checkOrCreatedPPMShipment - the existing call eventually calls `checkOrCreatedPPMShipment`

pkg/testdatagen/make_moving_expenses.go
```diff
@@ -0,0 +1,68 @@
...
+
+	document := GetOrCreateDocument(db, assertions.MovingExpense.Document, assertions)
```
duncan-truss, line 13: Should we by default assign the service member from the PPM being used instead of forcing them to provide it in the assertions?

pkg/testdatagen/make_moving_expenses.go
duncan-truss: I wouldn't expect this to create any additional weight ticket records. I would be surprised if I was using this and my shipment ended up with a weight ticket record when I was only creating a moving expense.

I think the place to create a PPM shipment that has actuals + weight ticket + moving expense is in the seed files and keep these methods composable.  I think it's most likely that a PPM shipment has one trip weight ticket and many moving expenses in reality.

ronaktruss: On a second look, I think I agree. I'll make the changes to not create the weight ticket.

I was originally thinking in terms of the customer perspective and not the api/db perspective.

felipe-lee: I was thinking this was good because we want our make functions to create realistic data. So like if we're creating an expense, we would have a ppm shipment with a weight ticket already. It also makes it easier to handle some of the service member logic so we don't have to repeat it, though I guess that func can get moved to the shared file like the document functions.

ronaktruss: I'm planning on moving that logic into the scenarios as it seems like logic that can be separated out from the actual make function. We should be able to test moving expenses without a weight ticket. While they are required from the customer workflow perspective, there are no hard business rules that prevents the API or DB from creating these.

Alternatively I may keep some of the existing functionality and rename the function to makeMovingExpenseWithWeightTicket.

pkg/testdatagen/make_moving_expenses.go
ronaktruss: Looks to be 50/50. [source](https://ustcdp3.slack.com/archives/C0250Q8469K/p1659453958530529)

pkg/testdatagen/make_moving_expenses.go
ronaktruss: I don't think there is a good way of doing this, since the service member is a bit removed from the PPM. If there's a better way that I am not thinking of, please let me know. We can still generate a moving expense without providing a service member id, but then the service member id for the moving expense documents won't match that of the orders.

pkg/testdatagen/make_moving_expenses.go
ronaktruss: @felipe-lee It doesn't look like `ensureServiceMemberIsSetUpInAssertions` is necessary here based on some tests I ran locally.

felipe-lee: When I was working on the weight ticket make functions, it was mainly needed because of the possible chain of calls if no assertions are passed in. Both a `PPMShipment` and multiple `Documents` would need to be created. The ppm shipment creator would call `MakeBaseMTOShipment', which would then call `MakeMove`, which would then call `MakeOrder`. That one looks in `assertions.Order.ServiceMember` for the service member. Meanwhile, `MakeDocument` looks for the service member in `assertions.Document.ServiceMember`. So when I was testing, I was ending up with multiple service members instead of just one. Like it could have been solved without that function I wrote by making the first call (document or ppm creation) and then adding the service member created there to assertions as needed, but I wanted to solve it up front rather than figuring it out for each call.

That said, if you aren't having the same issue once you take out the weight ticket creation (at least I think your other comment said you'll either take it out or have that as a `makeMovingExpenseWithWeightTicket` function), then no need to include the service member logic.

PR #8957, opened 2022-07-28
[MB-13178] PPM moving expenses service object

duncan-truss:
## Summary

This PR adds the MovingExpense creator and updater service objects with validation rules.  It also updates the mto shipments fetcher to include the moving expenses association in the list mto shipments response.

There is a new use of Eager creation for the moving expense document association to make that a one step process.  https://gobuffalo.io/documentation/database/relations/#flat-nested-creation

## Setup to Run Your Code

In your terminal run:

```make server_test```

### Frontend

### Backend

### CODE REVIEW

File: pkg/models/moving_expense.go [hunk 1]
```diff
@@ -72,7 +72,6 @@ type MovingExpenses []MovingExpense
 func (m *MovingExpense) Validate(_ *pop.Connection) (*validate.Errors, error) {
 	return validate.Validate(
 		&validators.UUIDIsPresent{Name: "PPMShipmentID", Field: m.PPMShipmentID},
-		&validators.UUIDIsPresent{Name: "DocumentID", Field: m.DocumentID},
```
duncan-truss, line 75: Because I'm using the eager creator this fails on ValidateAndCreate.  I'm wondering if we should separate out ValidateAndUpdate and ValidateAndCreate instead of overloading Validate for both.

pkg/services/moving_expense/moving_expense_creator.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	err := validateMovingExpense(appCtx, newMovingExpense, nil, f.checks...)
+
+	if err != nil {
+		return nil, err
+	}
+
+	txnErr := appCtx.NewTransaction(func(txnCtx appcontext.AppContext) error {
+		verrs, err := txnCtx.DB().Eager().ValidateAndCreate(newMovingExpense)
```
duncan-truss, line 37: So pop will let you create nested associations if you chain Eager() beforehand.  It did mean I had to remove DocumentID from being required on create in the model Validate() but the check is there in the update rules.  Since the service object is the one doing the document creation and it's not being sent from the user we can trust it I think.

pkg/services/moving_expense/moving_expense_updater_test.go
```diff
@@ -0,0 +1,321 @@
...
+
+		expectedMovingExpense := &models.MovingExpense{
+			ID:                originalMovingExpense.ID,
+			MovingExpenseType: &contractedExpenseType,
+			Description:       models.StringPointer("Dumpster rental"),
+			PaidWithGTCC:      models.BoolPointer(true),
+			MissingReceipt:    models.BoolPointer(true),
+			Amount:            models.CentPointer(unit.Cents(67899)),
+			Status:            &rejectedStatus,
+			Reason:            models.StringPointer("Learn to recycle"),
```
ronaktruss, line 128: 😂

pkg/services/moving_expense/moving_expense_updater.go
```diff
@@ -0,0 +1,140 @@
...
+		case sql.ErrNoRows:
+			return nil, apperror.NewNotFoundError(movingExpenseID, "while looking for MovingExpense")
+		default:
+			return nil, apperror.NewQueryError("MovingExpense fetch original", err, "")
+		}
+	}
+
+	// we aren't actively deleting the document stub but it's possible someone cascaded delete
+	// starting with the document instead of the user uploads.
+	if movingExpense.Document.DeletedAt == nil {
```
ronaktruss, line 85: Don't we always want to filter out the deleted records?

If the document had a deletedAt entry the pop query above would still return the entire set of deleted records and we wouldn't enter this if condition.

duncan-truss: Yea this was sort of a placeholder half finished thought that I had, while a Document can be deleted our Moving Expenses model doesn't really allow that because DocumentID can't be null.  Having this doesn't really save us from doing any real extra work or avoid any possible nil errors so I will probably get rid of it.

pkg/services/moving_expense/moving_expense_updater.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func FetchMovingExpenseByID(appContext appcontext.AppContext, movingExpenseID uuid.UUID) (*models.MovingExpense, error) {
+	var movingExpense models.MovingExpense
+
+	err := appContext.DB().Scope(utilities.ExcludeDeletedScope()).
+		EagerPreload("Document.UserUploads.Upload").Find(&movingExpense, movingExpenseID)
+
+	if err != nil {
+		fmt.Println(err.Error())
```
ronaktruss, line 72: What's the standard for this? Should we be using the logger from appContext?

pkg/services/moving_expense/rules.go
```diff
@@ -0,0 +1,113 @@
...
+			verrs.Add("MissingReceipt", "MissingReceipt is required")
+		}
+
+		if len(originalMovingExpense.Document.UserUploads) < 1 {
+			verrs.Add("Document", "At least 1 receipt file is required")
+		}
+
+		if newMovingExpense.Status != nil {
+			if newMovingExpense.Reason != nil {
+				fmt.Printf("newMovingExpense reason %s\n", *newMovingExpense.Reason)
```
ronaktruss, line 89: What's the standard for this? Should we be using the logger from appContext?

pkg/services/moving_expense/moving_expense_updater_test.go
```diff
@@ -0,0 +1,321 @@
...
+		suite.Len(updatedMovingExpense.Document.UserUploads, 1)
+		suite.Equal(*expectedMovingExpense.MovingExpenseType, *updatedMovingExpense.MovingExpenseType)
+		suite.Equal(*expectedMovingExpense.Description, *updatedMovingExpense.Description)
+		suite.Equal(*expectedMovingExpense.PaidWithGTCC, *updatedMovingExpense.PaidWithGTCC)
+		suite.Equal(*expectedMovingExpense.Amount, *updatedMovingExpense.Amount)
+		suite.Equal(*expectedMovingExpense.MissingReceipt, *updatedMovingExpense.MissingReceipt)
+		suite.Equal(*expectedMovingExpense.Status, *updatedMovingExpense.Status)
+		suite.Equal(*expectedMovingExpense.Reason, *updatedMovingExpense.Reason)
+		// Only the storage type receipt should be able to set these fields, would we rather reject
+		// the update outright than fail silently?
```
ronaktruss, line 148: Rejecting might be nice, in that It would potentially help us catch errors where the UI doesn't clear out the values when switching from a storage type receipt to something else. It would also add clarity to the business rule that is happening on the post. (Ie. I could see someone looking at the API documentation and not realizing there is a business rule that prevents them from adding sit info to a non SIT expense.

ronaktruss (approved):
LGTM. Nice job on the tests!

pkg/services/moving_expense/moving_expense_updater_test.go
duncan-truss: I guess we aren't failing silently we just ignore those SIT field values if it's not type storage.  If they update the storage type from storage to something else then we nil out those fields in the updater https://github.com/transcom/mymove/pull/8957/files/e8ee85847d78da729041febb3274edd7c5e0d1dd#diff-f517feb1649debad15e67e21636e285955d87f62523e7ea50e04693de0871bc0R99-R104

PR #8958, opened 2022-07-29
Add PopTestSuite#PreloadData, use suite.T().Cleanup

ahobson:
This change seems to greatly simplify the PopTestSuite code so we no
longer need to override suite.Run and no longer need a separate
RunWithPreloadedData

### DIFF
--- a/pkg/handlers/ghcapi/mto_service_items_test.go
+++ b/pkg/handlers/ghcapi/mto_service_items_test.go
@@ -164,7 +164,10 @@ func (suite *HandlerSuite) TestUpdateMTOServiceItemStatusHandler() {
 
 	req := httptest.NewRequest("PATCH", fmt.Sprintf("/move_task_orders/%s/mto_service_items/%s/status",
 		moveTaskOrderID, serviceItemID), nil)
-	requestUser := testdatagen.MakeStubbedUser(suite.DB())
+	var requestUser models.User
+	suite.PreloadData(func() {
+		requestUser = testdatagen.MakeStubbedUser(suite.DB())
+	})
 	req = suite.AuthenticateUserRequest(req, requestUser)
 
 	params := mtoserviceitemop.UpdateMTOServiceItemStatusParams{
--- a/pkg/handlers/ghcapi/mto_service_items_test.go
+++ b/pkg/handlers/ghcapi/mto_service_items_test.go
@@ -161,26 +161,28 @@ func (suite *HandlerSuite) createServiceItem() (models.MTOServiceItem, models.Mo
 func (suite *HandlerSuite) TestUpdateMTOServiceItemStatusHandler() {
 	moveTaskOrderID, _ := uuid.NewV4()
 	serviceItemID, _ := uuid.NewV4()
-
-	req := httptest.NewRequest("PATCH", fmt.Sprintf("/move_task_orders/%s/mto_service_items/%s/status",
-		moveTaskOrderID, serviceItemID), nil)
 	var requestUser models.User
-	suite.PreloadData(func() {
-		requestUser = testdatagen.MakeStubbedUser(suite.DB())
-	})
-	req = suite.AuthenticateUserRequest(req, requestUser)
 
-	params := mtoserviceitemop.UpdateMTOServiceItemStatusParams{
-		HTTPRequest:      req,
-		IfMatch:          etag.GenerateEtag(time.Now()),
-		Body:             &ghcmessages.PatchMTOServiceItemStatusPayload{Status: "APPROVED"},
-		MoveTaskOrderID:  moveTaskOrderID.String(),
-		MtoServiceItemID: serviceItemID.String(),
+	setupTestData := func() mtoserviceitemop.UpdateMTOServiceItemStatusParams {
+		requestUser = testdatagen.MakeStubbedUser(suite.DB())
+		req := httptest.NewRequest("PATCH", fmt.Sprintf("/move_task_orders/%s/mto_service_items/%s/status",
+			moveTaskOrderID, serviceItemID), nil)
+
+		req = suite.AuthenticateUserRequest(req, requestUser)
+		params := mtoserviceitemop.UpdateMTOServiceItemStatusParams{
+			HTTPRequest:      req,
+			IfMatch:          etag.GenerateEtag(time.Now()),
+			Body:             &ghcmessages.PatchMTOServiceItemStatusPayload{Status: "APPROVED"},
+			MoveTaskOrderID:  moveTaskOrderID.String(),
+			MtoServiceItemID: serviceItemID.String(),
+		}
+		return params
 	}
 
 	// With this first set of tests we'll use mocked service object responses so that we can make sure the handler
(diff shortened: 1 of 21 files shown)


pkg/testingsuite/pop_suite_test.go
```diff
@@ -19,10 +19,13 @@ func TestPopTestSuite(t *testing.T) {
 	ps.TearDown()
 }
 
-func (suite *PopTestSuite) TestRunWithPreloadedData() {
-	address := testdatagen.MakeAddress(suite.DB(), testdatagen.Assertions{})
+func (suite *PopTestSuite) TestRunWithPreloadData() {
+	var address models.Address
```
shimonacarvalho, line 23: Hey Drew so the issue we need fixed is having the SetupTest and/or SetupSuite function preload data. I had a sample test here that we looked at - test-pop-suite-bug

What do you think about whether this new design will be able to solve that situation?

pkg/handlers/ghcapi/mto_service_items_test.go
```diff
@@ -166,7 +166,10 @@ func (suite *HandlerSuite) TestUpdateMTOServiceItemStatusHandler() {
 
 	req := httptest.NewRequest("PATCH", fmt.Sprintf("/move_task_orders/%s/mto_service_items/%s/status",
 		moveTaskOrderID, serviceItemID), nil)
-	requestUser := testdatagen.MakeStubbedUser(suite.DB())
+	var requestUser models.User
```
shimonacarvalho, line 169: Why would a stubbed user need to be preloaded? Doesn't touch db right?

pkg/handlers/ghcapi/payment_request_test.go
```diff
@@ -230,11 +230,15 @@ func (suite *HandlerSuite) TestGetPaymentRequestsForMoveHandler() {
...
 	paymentRequestID, _ := uuid.FromString("00000000-0000-0000-0000-000000000001")
 	officeUserUUID, _ := uuid.NewV4()
-	officeUser := testdatagen.MakeOfficeUser(suite.DB(), testdatagen.Assertions{Stub: true, OfficeUser: models.OfficeUser{ID: officeUserUUID}})
-	officeUser.User.Roles = append(officeUser.User.Roles, roles.Role{
-		RoleType: roles.RoleTypeTIO,
-	})
+	var officeUser models.OfficeUser
+
+	suite.PreloadData(func() {
+		officeUser = testdatagen.MakeOfficeUser(suite.DB(), testdatagen.Assertions{Stub: true, OfficeUser: models.OfficeUser{ID: officeUserUUID}})
```
shimonacarvalho, line 236: Stubbed user

pkg/handlers/primeapi/mto_shipment_test.go
```diff
@@ -70,27 +70,32 @@ func (suite *HandlerSuite) TestCreateMTOShipmentHandler() {
...
-		StreetAddress1: &newAddress.StreetAddress1,
-		StreetAddress2: newAddress.StreetAddress2,
-		StreetAddress3: newAddress.StreetAddress3,
-	}
+	var pickupAddress primemessages.Address
+	var destinationAddress primemessages.Address
+
+	suite.PreloadData(func() {
+		// Make stubbed addresses just to collect address data for payload
+		newAddress := testdatagen.MakeStubbedAddress(suite.DB())
```
shimonacarvalho, line 78: Stubbed addresses - I think all these stubs shouldn't go in preloads.

pkg/testingsuite/pop_suite_test.go
```diff
@@ -3,26 +3,38 @@ package testingsuite
...
 	"github.com/stretchr/testify/suite"
 
 	"github.com/transcom/mymove/pkg/models"
 	"github.com/transcom/mymove/pkg/testdatagen"
 )
 
-func (suite *PopTestSuite) SetupTest() {
+type SimplePopSuite struct {
+	*PopTestSuite
+	models.ReServices
```
ahobson, line 15: @shimonacarvalho As far as I can tell, this isn't used in the `SimplePopSuite` ?

pkg/testingsuite/pop_suite_test.go
```diff
@@ -72,3 +103,124 @@ func (suite *PopTestSuite) TestRun() {
...
+		},
+		{
+			ID:   uuid.Must(uuid.NewV4()),
+			Code: models.ReServiceCodeMS,
+		},
+		{
+			ID:   uuid.Must(uuid.NewV4()),
+			Code: models.ReServiceCodeDCRT,
+		},
+	}
```
ahobson, line 156: @shimonacarvalho I would have put the initialization of `ReServices` in `SetupSuite` so all of the initialization code was in the same place. Is there a reason you prefer to initialize it here?

pkg/handlers/ghcapi/payment_request_test.go
ahobson: It is a stubbed user, but it references `suite.DB` and that call needs to be in a `PreloadData` block

pkg/handlers/primeapi/mto_shipment_test.go
ahobson: I mentioned this elsewhere, but because it references `suite.DB`, even though it is a stubbed user, it needs to be in `PreloadData`

pkg/testingsuite/pop_suite_test.go
shimonacarvalho: Hmm, I guess I just felt like this function was for initializing the PreloadedPopSuite "object" so I put it there (and originally both SetupTest and SetupSuite were using the data). But equally I could see good reason for putting it in SetupSuite. So whichever you prefer is fine.

pkg/testingsuite/pop_suite_txn.go
```diff
@@ -96,62 +84,17 @@ func (store *PopSuiteTxnStore) TransactionContextOptions(ctx context.Context, op
...
-	if suite.lowPrivConn == nil {
-		return
-	}
-	err := suite.lowPrivConn.Close()
-	if err != nil {
-		log.Panicf("Error closing lowPrivConn: %v", err)
-	}
-	suite.lowPrivConn = nil
+	// if suite.lowPrivConn == nil {
+	// 	return
```
shimonacarvalho, line 93: Deadcode? Could delete if so.

pkg/handlers/ghcapi/payment_request_test.go
shimonacarvalho: Ah i see. Bit weird but okay, understood - we might want to update documentation with this nugget of info.

shimonacarvalho (approved):

pkg/handlers/ghcapi/mto_service_items_test.go
ahobson: To reiterate from another conversation: any call to `suite.DB()` would need to be in `PreloadData` or in a subtest. If we can change/fix the `MakeStubbedThing` calls so they never actually hit the db, we can then change this.

shimonacarvalho:
This needs documentation + broadcast the change to the team.

PR #8971, opened 2022-08-01
[MB 13164] PPM Closeout Expense Dropdown

pearl-truss:
## Summary
This PR created the ExpenseForm component and displays different scenarios of its use in storybook under Customer Components / PPM Closeout / Expense Form

Storybook https://output.circle-artifacts.com/output/job/4d27fb3e-efb8-4c2f-8b3e-1608258c59bb/artifacts/0/storybook/index.html?path=/story/customer-components-ppm-closeout-expenses-ppm-form--sit-expenses

Designs https://app.zeplin.io/project/625896e702018f8ff82451c0/screen/625897c910f1e1907c1b2151

## Setup to Run Your Code

Start the Storybook locally.

```sh
make storybook
```

## Screenshots
Screen Shot 2022-08-02 at 3 48 07 PM
Screen Shot 2022-08-02 at 3 48 18 PM
Screen Shot 2022-08-02 at 3 48 27 PM
Screen Shot 2022-08-03 at 11 00 34 AM


dinabeigelman (commented):
Took a look and found a few things:
Should there be a char limit for the text input "What did you buy?"? I can just keep pasting more words in and I guess it's fine?

When you change the dates on the SIT options the calculation gets really wonky
Screen Shot 2022-08-02 at 2 53 18 PM
Screen Shot 2022-08-02 at 2 53 26 PM

If you delete the already uploaded file and upload a new PDF, it looks like it was successful but gives a very strange error.
Screen Shot 2022-08-02 at 2 54 06 PM

Additionally, if you add a new PDF the styling is very different between what was uploaded previously and what was just uploaded. I'm not the most familiar with our file upload component so not sure if this is as designed.
Screen Shot 2022-08-02 at 3 01 26 PM

Lastly, I don't know if this is within the scope of this ticket, but I think @scarequotes and @AnnaGingle created an alert for when the user checks "I don't have my receipt" here: https://www.figma.com/file/tRkjtoWjT8727XH5m9k7mK/Milmove-Customer-App?node-id=305%3A1411

pearl-truss:
Thanks Dina! 

I've addressed your feedback, changes can be viewed with the [updated url](https://output.circle-artifacts.com/output/job/4d27fb3e-efb8-4c2f-8b3e-1608258c59bb/artifacts/0/storybook/index.html?path=/story/customer-components-ppm-closeout-expenses-ppm-form--sit-expenses)
- Quite a few of what you listed here are just storybook specific interaction issues. I went ahead and mocked out the interactions for file uploading and deleting since storybook can't use our actual APIs for handling these events. So if you were to click on the file upload component, you'll just see the event registered in the actions tab
- The extra decimal issue has been addressed 
-  FWIW we have other text fields in the PPM closeout flow (like the vehicle description field on the Weight Ticket screen) that aren't implementing a character limit
- Thanks for the link to the alert for what happens when the missing receipt button is checked. That was missing from this ticket. I went ahead and added it here. However, I noticed the blue used on figma is different than what's used for our Alert component. I'm not sure if that's an intentional change that should be introduced to all info type Alerts?

Screen Shot 2022-08-03 at 11 55 55 AM

src/utils/dates.js
```diff
@@ -0,0 +1,10 @@
+export default function numOfDaysBetweenDates(date1, date2) {
+  const jsDate1 = new Date(date1);
+  const jsDate2 = new Date(date2);
```
duncan-truss, line 3: It looks like the javascript Date constructor is setting a value for the hours and the difference in the daylight savings is causing the non-round value.  Moment might be easier to use here or [setting the hours value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setHours) to zero might help reduce the variability.

Screen Shot 2022-08-03 at 12 48 32 PM

pearl-truss: I thought there was an ADR to not use moment so I did this instead. But if it's okay to use moment I'll give that I try next (that would of been my first choice otherwise)

duncan-truss: I think if we had an alternative already introduced like DayJS that would be the preference but the "standard" Javascript date library (or non-standard browser implementations) has it's own issues and the API is lacking.

src/utils/dates.js
```diff
@@ -1,10 +1,8 @@
-export default function numOfDaysBetweenDates(date1, date2) {
-  const jsDate1 = new Date(date1);
-  const jsDate2 = new Date(date2);
+import moment from 'moment';
 
-  // To calculate the time difference of two dates
-  const timeDiff = jsDate2.getTime() - jsDate1.getTime();
+export default function numOfDaysBetweenDates(date1, date2) {
+  const mDate1 = moment(date1);
+  const mDate2 = moment(date2);
```
duncan-truss, line 5: I think we may want `moment.utc(date)` so no local time conversion is applied https://momentjs.com/docs/#/parsing/utc/

src/components/Customer/PPM/Closeout/ExpenseForm/ExpenseForm.jsx
```diff
@@ -0,0 +1,238 @@
...
+                  </>
+                )}
+                {values.expenseType === 'storage' && (
+                  <FormGroup>
+                    <h3>Dates</h3>
+                    <DatePickerInput name="sitStartDate" label="Start date" />
+                    <DatePickerInput name="sitEndDate" label="End date" />
+                    <h3>
+                      Days in storage:{' '}
+                      {values.sitStartDate && values.sitEndDate
```
duncan-truss, line 195: You may want to also check that there are no errors for those SIT date fields. You can still end up with a NaN for the number of days if one of those fields has a type parse error.

src/utils/dates.js
ronaktruss: Just sharing this because this tripped me up a bit:

Looking at the example screenshot that Duncan shared I was a little surprised that this wasn't working. The reason I would expect that to work is that the underlying time (that is retrieved when using getTime()) represents the beginning of the day in UTC time so the local timezones shouldn't have any impact.

After a closer look, it looks like the UI actually passes the dates using a slightly different format. For example the dates are passed in as "02 Sep 2022" or "07 Nov 2022" which when calling new Date on returns the beginning of the day in your local time.

duncan-truss: Ahh good catch about the api vs field formats, I did print the app values out locally but then did something different when trying to illustrate the console example.

Screen Shot 2022-08-03 at 3 22 05 PM

src/components/Customer/PPM/Closeout/ExpenseForm/ExpenseForm.test.jsx
```diff
@@ -0,0 +1,223 @@
...
+      expect(screen.getByText('200KB')).toBeInTheDocument();
+      expect(screen.getByText('Uploaded 10 Aug 2022 12:59 PM')).toBeInTheDocument();
+
+      expect(screen.getByLabelText('No')).toBeChecked();
+      expect(screen.getByRole('heading', { level: 3, name: 'Dates' })).toBeInTheDocument();
+      expect(screen.getByLabelText('Start date')).toBeInstanceOf(HTMLInputElement);
+      expect(screen.getByLabelText('Start date')).toHaveDisplayValue('24 Sep 2022');
+      expect(screen.getByLabelText('End date')).toBeInstanceOf(HTMLInputElement);
+      expect(screen.getByLabelText('End date')).toHaveDisplayValue('26 Dec 2022');
+
```
duncan-truss, line 179: Could you add a check for the days in storage value being displayed?

src/components/Customer/PPM/Closeout/ExpenseForm/ExpenseForm.jsx
```diff
@@ -0,0 +1,238 @@
...
+    .when('expenseType', {
+      is: 'storage',
+      then: (schema) => schema.required('Required'),
+    }),
+  sitEndDate: Yup.date()
+    .typeError('Enter a complete date in DD MMM YYYY format (day, month, year).')
+    .when('expenseType', {
+      is: 'storage',
+      then: (schema) => schema.required('Required'),
+    }),
```
duncan-truss, line 43: I think you could chain another when clause on here for the sitEndDate needing to be equal to or after the sitStartDate https://github.com/jquense/yup#schemawhenkeys-string--string-builder-object--values-any-schema--schema-schema

duncan-truss (approved):

dinabeigelman:
@pearl-truss I took another look and I think it looks good!

I will admit I don't know what changes were made to the upload functionality but understand that there are limitations because of Storybook so will assume that when this is on a page in Milmove we can take another look there.

I'm checking with @AnnaGingle about a copy thing in the alert [here](https://www.figma.com/file/tRkjtoWjT8727XH5m9k7mK?node-id=305:1411#240505116) so once I hear back we can call this done. With regard to the Alert styling, yeah, it's very different in Storybook than in our designs. What we have in our designs should be the source of truth but because it takes engineer time to update the component in Storybook it's not a blocker at the moment.

pearl-truss:
@dinabeigelman does copy match? I pushed a recent change
Screen Shot 2022-08-05 at 11 22 27 AM

dinabeigelman:
@pearl-truss looks good to me!

dinabeigelman (approved):

PR #8972, opened 2022-08-01
MB-13390: Remove screenshot workaround

ronaktruss:
## Summary

Remove the usage of `cy.screenshot()` for PPMs.


felipe-lee (approved):
Thanks for updating this, I'm glad this works as a replacement for the screenshot.

ronaktruss:
I think the reason that we "needed" the screenshot was because we have the loading screen so sometimes the cypress commands were trying to run on that version of the page instead of the page post loading.

PR #9023, opened 2022-08-05
MB-13444: Resolve upload validation for weight tickets

ronaktruss:
## Summary

This should, resolve the issue we were seeing in demo where we would see the validation error of needing at least 1 file even after a file has been uploaded. The root cause seemed to be related with the fact that the `onChange` function would sometimes trigger before the `createUpload`.

## Setup to Run Your Code / Verification steps

To recreate the issue please run the main branch with the modified environment variable so that you are using the dev s3 to store files instead of local. Steps to do this can be found on https://dp3.atlassian.net/wiki/spaces/MT/pages/1470955567/How+to+test+storing+data+in+S3+locally. You should use the file that is linked [here](https://dp3.atlassian.net/browse/MB-13444?focusedCommentId=23290).

Once you have verified the issue please switch back to this branch and using the same modified environment variables described above, verify if you are still seeing this issue.


scottmries (approved):

scottmries:
LGTM! I was able to reproduce this on `master` but not on this branch.

PR #9035, opened 2022-08-08
MB-11853 Update ppm upload document button routing

felipe-lee:
## Summary

This PR updates what happens when a customer with an approved PPM shipment clicks on the "Upload PPM Documents" button on the customer home page. Routing will depend on where they are in the process. 

If the user has submitted a move with a ppm shipment that has been approved, then when the "Upload PPM Documents" button is clicked, the user will be routed to the:

* About Your PPM page if no actual shipment info has been input
* Weight Ticket page if weight ticket info is missing
* Weight Ticket page if weight ticket info is incomplete
* Weight Ticket page for the first weight ticket if there are multiple but none are complete
* Review page if weight ticket info is complete
* Review page if at least one weight ticket is completely filled out

Asked in [#feat-ppm-integration](https://ustcdp3.slack.com/archives/C0250Q8469K/p1660002826332799) to verify the above routing is correct. 

<details>

* Add `isPPMAboutInfoComplete` util function to help know when the "About your PPM" section has been completed.
* Add `isWeightTicketComplete` util function to test when a weight ticket is complete.
    * This takes and expands on the logic that was being used in the previously named `hasCompleteWeightTickets` to make it possible to re-use the logic in different scenarios (testing for a single complete weight ticket vs testing that all are complete).
    * The rules roughly follows the same criteria as what we check for when they're trying to submit the weight ticket info.
* Update routing from the Home page when the "Upload PPM Documents" button is clicked.
* Updated "About Your PPM" cypress tests to use unique users for each test (desktop vs mobile and advance vs no advance, so 4 total). Now that the routing has changed, we can't re-use users for these tests because on the second pass-through, they don't end up where they used to. I think it's better to test each one from scratch anyway.
</details>

<details>
These are other changes that helped with this ticket, but weren't strictly changed because of this ticket.

* Added a few factories for creating uploads, documents, and weight tickets.
    * None of them are super robust, but they can serve until we truly implement factories in the future.
    * I purposely didn't start adding factories like the ones called out in the [fake data factory ADR](https://transcom.github.io/mymove-docs/docs/adrs/use-fake-data-factories-with-fake-data-generators-to-create-test-data) and shown in the [proof of concept PR](https://github.com/transcom/mymove/pull/8388) because I didn't want to make this PR too hard to review (since it'd be new tools people would have to get familiar with). If y'all think I should just start using them though, I'm more than happy to update.
* Modify `.eslintrc.js` to allow `'testUtils.jsx'` to spread `jsx` props (rule: `'react/jsx-props-no-spreading'`)
    * This is what triggered the `Truss-IS3` review. 
    * This was in order to make the mem history & `MockProviders` function mentioned next.
* Move logic for prepping mem history & MockProviders to a test util function
    * Makes it easier to re-use, which I hope to do in my new tests for another component.
    * Replaced the code I copied from with the new function, both to minimize duplication and to ensure the new function does work as intended.
* Reduce data duplication in `src/pages/MyMove/Home/index.test.jsx` and update to use some more realistic data (e.g. using uuids for ids vs hard-coded short strings).
* Rename `hasCompleteWeightTickets` to `hasCompletedAllWeightTickets`. 
    * This matches its function more, and becomes important to disambiguate because there is a difference between having some weight tickets completed vs all.
</details>

### Local Setup 

<details>
<summary>💻</summary>

#### Client UI

Start the UI locally.

```sh
make client_run
```

#### Server

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the [local client app](http://milmovelocal:3000/sign-in) or the [ephemeral deployment](https://my-milmove-pr-9035.mymove.sandbox.truss.coffee/sign-in).
1. Now log in as the client that matches what you want to test. To see routing when:
    1. no "About your PPM" info has been input, log in as `readyToFinish@ppm.approved`
        1. Should route to "About your PPM" page 
    1. "About Your PPM" info is filled out, but there is no weight ticket info, log in as `actualPPMDateZIPAdvanceDone@ppm.approved`
        1. Should route to weight ticket page 
    1. Weight ticket info is complete, log in as `weightTicketPPM@ppm.approved`
        1. Should route to review page 

There are a few other scenarios you can test out, but there's no pre-made data that will get you there. You can edit an existing one that is close (I'd suggest starting with `actualPPMDateZIPAdvanceDone2@ppm.approved` and the related 3-6 since they're pretty close).

The other three scenarios to test are:

* Weight Ticket page if weight ticket info is incomplete (e.g. only has empty weight info)
* Weight Ticket page for the first weight ticket if there are multiple but none are complete
* Review page if at least one weight ticket is completely filled out

### Frontend


scottmries (approved):
LGTM. I confirmed that the routing works as expected in the ephemeral environment, and I don't see issues with any of the code. 
It might be worth getting a second approval considering the size of this PR and my inexperience with the codebase.

src/testUtils.jsx
```diff
@@ -49,10 +49,23 @@ MockProviders.propTypes = {
...
+  initialEntries: DEFAULT_INITIAL_ENTRIES,
   history: null,
   permissions: [],
   currentUserId: null,
 };
+
+export const setUpProvidersWithHistory = (initialEntries = DEFAULT_INITIAL_ENTRIES) => {
+  const memoryHistory = createMockHistory(initialEntries);
+
+  const mockProviderWithHistory = (props) => <MockProviders history={memoryHistory} {...props} />;
```
duncan-truss, line 65: Nice, thanks for extracting this out of the review tests 👏

src/utils/shipments.js
```diff
@@ -8,14 +8,45 @@ export function isPPMShipmentComplete(mtoShipment) {
...
-export function hasCompleteWeightTickets(weightTickets) {
+// isPPMAboutInfoComplete - checks if all the "About your ppm" fields have data in them.
+export function isPPMAboutInfoComplete(ppmShipment) {
+  return [
+    'actualMoveDate',
+    'actualPickupPostalCode',
+    'actualDestinationPostalCode',
+    'hasReceivedAdvance',
+    'advanceAmountReceived',
+  ].every((fieldName) => ppmShipment[fieldName] !== null);
```
duncan-truss, line 19: If hasReceivedAdvance is false, I believe advanceAmountRecieved could still be null right or are you seeing it set to zero?

felipe-lee: Ah, good point. I'll add a test for that and update the function.

src/utils/shipments.test.js
```diff
@@ -80,28 +95,118 @@ describe('shipments utils', () => {
...
+    it.each([
+      [false, 'all about your ppm info is null', approvedPPMShipment],
+      [false, 'actual move date is null', { ...ppmShipmentWithAboutInfo, actualMoveDate: null }],
+      [false, 'actual pickup postal code is null', { ...ppmShipmentWithAboutInfo, actualPickupPostalCode: null }],
+      [
+        false,
+        'actual destination postal code is null',
+        { ...ppmShipmentWithAboutInfo, actualDestinationPostalCode: null },
+      ],
+      [false, 'has received advance is null', { ...ppmShipmentWithAboutInfo, hasReceivedAdvance: null }],
```
duncan-truss, line 157: I don't think we can get into a scenario with the current form where only one of these fields is null but it's probably fine to keep them in in-case something changes.

duncan-truss:
@felipe-lee One thing I'm seeing that we need to fix is that if you go back and edit the About Your PPM we're still taken to the create new Weight Ticket page.  Are you able to use your new complete functions or some other means to know when to take us back to the Review page?

duncan-truss (approved):
Works as expected outside of the re-editing the About page that is more of a loose end leftover from other work.

src/utils/shipments.test.js
felipe-lee: Ah good point. I think it's best to match the data to what we expect. Like in this case, it doesn't matter too much what the amount is since we'd ignore it based on the bool field, but still.

felipe-lee:
> @felipe-lee One thing I'm seeing that we need to fix is that if you go back and edit the About Your PPM we're still taken to the create new Weight Ticket page. Are you able to use your new complete functions or some other means to know when to take us back to the Review page?

Good question. I can mess with it a bit. I realized I forgot I need an IS3 review anyways before I can merge.

LeDeep (approved):
looks good!

felipe-lee:
> > @felipe-lee One thing I'm seeing that we need to fix is that if you go back and edit the About Your PPM we're still taken to the create new Weight Ticket page. Are you able to use your new complete functions or some other means to know when to take us back to the Review page?
> 
> Good question. I can mess with it a bit. I realized I forgot I need an IS3 review anyways before I can merge.

@duncan-truss I'm going to go ahead and merge this PR and make a followup PR for the routing from the about page

PR #9042, opened 2022-08-09
[MB-13307] SC can approve/deny an advance (PPMs)

pearl-truss:
## Summary

This PR adds the ability for a SC to approve, deny or adjust a requested advance. When the SC rejects the advance a new status field called `AdvanceStatus` is updated on the ppm shipment on the backend. It's set to `EDITED` when the SC adjusted the amount and `APPROVED` when no changes are made.

## Setup to Run Your Code
Can be viewed here: https://office-milmove-pr-9042.mymove.sandbox.truss.coffee/sign-in 

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Login the office app as a SC `too_tio_services_counselor_role@office.mil (PPM office)`
2. Find a move with an approved PPM, `ABTPPM`
3. Click Edit Shipment
4. Continue to the Advance and Incentive page (page 2)

## Screenshots
Screen Shot 2022-08-12 at 3 31 16 PM
Screen Shot 2022-08-12 at 3 31 21 PM
Screen Shot 2022-08-12 at 3 59 10 PM
Screen Shot 2022-08-12 at 4 08 02 PM
Screen Shot 2022-08-12 at 4 06 47 PM


migrations/app/schema/20220810115142_add_indication_of_advance_adjustment_to_ppm_shipments.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE ppm_shipments
```
felipe-lee, line 1: **question**: should this be a `bool` or some kind of status enum that can be like `APPROVED`, `EDITED`, or `REJECTED`? 

I'm thinking the scenarios for each status would be:

If a customer created the PPM
* with and advance requested
    * SC goes through and doesn't change anything, then it's `APPROVED`
    * SC goes through and edits the amount, then it's `EDITED`
    * SC changes it to not have an advance, then it's `REJECTED`
* w/o an advance
    * SC goes through and doesn't change anything, then it's `null`
    * SC goes through and adds an advance, then it's `EDITED`

SC creates the PPM
* This one is hard because we don't know that the SC created the PPM so when they're on the second page, any changes they make look the same as if they made those changes on a PPM that the customer creates. So maybe we should treat them the same, where we can. It would mean that with my breakdown above, there wouldn't really be a way to get an `APPROVED` status.

src/components/Office/ShipmentForm/ppmShipmentSchema.js
```diff
@@ -3,7 +3,12 @@ import * as Yup from 'yup';
 import { getFormattedMaxAdvancePercentage } from 'utils/incentives';
 import { InvalidZIPTypeError, ZIP5_CODE_REGEX } from 'utils/validation';
 
-const ppmShipmentSchema = ({ estimatedIncentive = 0, weightAllotment = {} }) => {
+const ppmShipmentSchema = ({
+  estimatedIncentive = 0,
+  weightAllotment = {},
+  requestedAdvanceAmount = 0,
+  isAdvanceRequested,
```
felipe-lee, line 10: **nit**: Could you update these to match the field names that come back from the endpoint? Or is there a particular reason to change them?

src/components/Office/ShipmentForm/ShipmentForm.jsx
```diff
@@ -165,10 +167,16 @@ const ShipmentForm = (props) => {
...
-      const ppmShipment = formatPpmShipmentForAPI(formValues);
-
+      const ppmShipmentBody = formatPpmShipmentForAPI(formValues);
+      if (
+        formValues.counselorRemarks &&
+        (formValues.advance !== mtoShipment.ppmShipment.advanceAmountRequested / 100 ||
+          formValues.advanceRequested !== mtoShipment.ppmShipment.hasRequestedAdvance)
+      ) {
+        ppmShipmentBody.ppmShipment.hasOfficeAdjustedAdvance = true;
+      }
```
felipe-lee, line 177: **thought**: I feel like this might be better to manage fully on the server side. I see it more as a status that should be set by the server based on what the user does, rather than something that we should rely on the FE to send.

src/components/Office/ShipmentForm/ShipmentForm.jsx
```diff
@@ -551,6 +561,8 @@ const ShipmentForm = (props) => {
                     shipmentType={shipmentType}
                     customerRemarks={mtoShipment.customerRemarks}
                     counselorRemarks={mtoShipment.counselorRemarks}
+                    showHint={false}
```
felipe-lee, line 564: **praise**: nice catch on this not matching the designs previously and making it easier to toggle on/off as needed.

src/components/form/fields/TextField/TextField.jsx
```diff
@@ -58,8 +59,13 @@ const TextField = ({
...
 
       {showWarning && <Hint data-testid="textInputWarning">{warning}</Hint>}
-      {/* eslint-disable-next-line react/jsx-props-no-spreading */}
-      <TextInput id={id} name={name} disabled={isDisabled} {...fieldProps} {...inputProps} />
+      {/* eslint-disable react/jsx-props-no-spreading */}
+      {display === 'input' ? (
+        <TextInput id={id} name={name} disabled={isDisabled} {...fieldProps} {...inputProps} />
+      ) : (
+        <Textarea id={id} name={name} disabled={isDisabled} {...fieldProps} {...inputProps} />
+      )}
```
felipe-lee, line 67: **praise**: 🎉 thanks for making this component usable in more use cases now.

src/components/Office/ShipmentForm/ShipmentForm.jsx
pearl-truss: Yeah that makes 👍

swagger-def/internal.yaml
```diff
@@ -2613,6 +2613,11 @@ definitions:
           Indicates whether PPM shipment has pro gear.
         type: boolean
         x-nullable: true
+      hasOfficeAdjustedAdvance:
+        description: >
+          Indicates whether an advance has been adjusted or denied by the office.
+        type: boolean
+        x-nullable: true
```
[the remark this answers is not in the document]
pearl-truss: Oh yeah, nice catch!

migrations/app/schema/20220810115142_add_indication_of_advance_adjustment_to_ppm_shipments.up.sql
pearl-truss: I was thinking a bool because I wasn't sure if we needed to differentiate beyond if an advance was changed by the SC or not. It sounds like we might? So I can go ahead and make this change

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -629,12 +629,13 @@ func PPMShipment(ppmShipment *models.PPMShipment) *ghcmessages.PPMShipment {
 		HasRequestedAdvance:            ppmShipment.HasRequestedAdvance,
 		AdvanceAmountRequested:         handlers.FmtCost(ppmShipment.AdvanceAmountRequested),
 		HasReceivedAdvance:             ppmShipment.HasReceivedAdvance,
-		AdvanceAmountReceived:          handlers.FmtCost(ppmShipment.AdvanceAmountReceived),
-		SitEstimatedWeight:             handlers.FmtPoundPtr(ppmShipment.SITEstimatedWeight),
-		SitEstimatedEntryDate:          handlers.FmtDatePtr(ppmShipment.SITEstimatedEntryDate),
-		SitEstimatedDepartureDate:      handlers.FmtDatePtr(ppmShipment.SITEstimatedDepartureDate),
-		SitEstimatedCost:               handlers.FmtCost(ppmShipment.SITEstimatedCost),
-		ETag:                           etag.GenerateEtag(ppmShipment.UpdatedAt),
+		// AdvanceStatus:                 ghcmessages.PPMAdvanceStatus(ppmShipment.AdvanceStatus),
```
pearl-truss, line 632: I've tried a few ways of getting this to work. Any ideas would be appreciated!

pearl-truss: This is resolved

duncan-truss (approved):
👍 Reviewed the database migration for the advance status, haven't had a chance to review the rest yet.

scarequotes (commented):
Everything looks good. 

Not sure if this is part of this ticket or not, but one thing I noticed:

Went in to a move with a PPM and advance requested.

Changed "yes" to "no" and added remarks. 

Saved the changes.

The shipments screen 
* Does not update "advance requested" field to `no`
* but does save the comments I put in when I denied the request

Screen Shot 2022-08-12 at 2 49 55 PM
 
If that's separate work, cool, I'll approve, but wanted to call it out.

pearl-truss:
> If that's separate work, cool, I'll approve, but wanted to call it out.

Great catch James! I think that's part of this ticket and will find a fix for

pearl-truss:
> https://office-milmove-pr-9042.mymove.sandbox.truss.coffee/sign-in

@scarequotes I just pushed a fix. It can be re-reviewed here: https://office-milmove-pr-9042.mymove.sandbox.truss.coffee/sign-in 

Screen Shot 2022-08-15 at 10 00 57 AM

scarequotes (approved):
Glad I spotted that! Looks good to go now.

src/utils/formatMtoShipment.js
```diff
@@ -105,7 +105,7 @@ export function formatPpmShipmentForDisplay({ counselorRemarks = '', ppmShipment
     spouseProGearWeight: (ppmShipment.spouseProGearWeight || '').toString(),
 
     estimatedIncentive: ppmShipment.estimatedIncentive,
-    advanceRequested: !!ppmShipment.hasRequestedAdvance,
+    advanceRequested: ppmShipment.hasRequestedAdvance?.toString() || 'false',
     advance: (ppmShipment.advanceAmountRequested / 100 || '').toString(),
 
     counselorRemarks,
```
felipe-lee, line 111: **suggestion**: I found a bug with another line in this file that is masking a bug with this PR. This line: https://github.com/transcom/mymove/blob/pm-mb-13307/src/utils/formatMtoShipment.js#L250 is making it so that if you set `false` for advance requested at some point, going forward your changes on page 1 are never saved because `!!formValues.advanceRequested` evaluates to `true` when `formValues.advanceRequested === 'false'`. I know it's not quite what you're working on this PR, but if you can fix it, it'd be great so that it's easier to see the bug I was trying to test. I did fix it locally to be `hasRequestedAdvance: formValues.advanceRequested === 'true'` so I could test what I was actually trying to test.

I found this because while looking at the changes to `pkg/services/ppmshipment/validation.go`, but I'll put the details of that bug in a comment on that file.

pkg/services/ppmshipment/validation.go
```diff
@@ -76,6 +76,7 @@ func mergePPMShipment(newPPMShipment models.PPMShipment, oldPPMShipment *models.
 	ppmShipment.HasRequestedAdvance = services.SetNoNilOptionalBoolField(newPPMShipment.HasRequestedAdvance, ppmShipment.HasRequestedAdvance)
 	ppmShipment.AdvanceAmountRequested = services.SetNoNilOptionalCentField(newPPMShipment.AdvanceAmountRequested, ppmShipment.AdvanceAmountRequested)
 	ppmShipment.HasReceivedAdvance = services.SetNoNilOptionalBoolField(newPPMShipment.HasReceivedAdvance, ppmShipment.HasReceivedAdvance)
+	ppmShipment.AdvanceStatus = newPPMShipment.AdvanceStatus
```
felipe-lee, line 79: **todo**: There is a bug with this where the `AdvanceStatus` gets nullified if you don't explicitly include it in the `newPPMShipment`. I think this should probably only be set if `newPPMShipment.AdvanceStatus` isn't `nil`.

src/utils/formatMtoShipment.js
felipe-lee: It's fine if this bug gets made into a separate ticket though, I don't want to derail this ticket.

src/components/Office/ShipmentForm/ppmShipmentSchema.js
```diff
@@ -70,6 +76,14 @@ const ppmShipmentSchema = ({ estimatedIncentive = 0, weightAllotment = {} }) =>
...
         then: (schema) => schema.required('Required'),
       }),
+
+    counselorRemarks: Yup.string().when(['advance', 'advanceRequested'], {
+      is: (advance, advanceRequested) =>
+        isAdvancePage &&
+        (Number(advance) !== advanceAmountRequested / 100 ||
+          advanceRequested?.toString() !== hasRequestedAdvance?.toString()),
+      then: (schema) => schema.required('Required'),
+    }),
```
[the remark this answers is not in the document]
pearl-truss: I'll defer to @scarequotes on this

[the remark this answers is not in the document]
scarequotes: I'm going to loop in @pmohal. Basic thoughts:
* We don't want to delete old notes; they're supposed to be a historical record
* But we don't have a way at this point to add new notes as a separate field
* Forcing the notes to be different seems like a nice enhancement but probably complicated to implement well
* So training seems like a reasonable solution for this

So my thought is to leave it as-is and note for enhancements later. Let's confirm with product.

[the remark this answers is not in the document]
pmohal: Is this information stored in the H&AL? That should take care of the historical record portion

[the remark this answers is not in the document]
pearl-truss: @scarequotes forcing the notes to be different is fairly straightforward

src/utils/formatMtoShipment.js
pearl-truss: @felipe-lee I fixed the bug in this PR instead

felipe-lee (approved):
Thanks for making those updates! I'm not sure if the decision is to add a check for the counselor notes being different or not based on the thread, but if not, then this lgtm!

pearl-truss:
@felipe-lee thanks, from what @pmohal shared in sync it sounds like this information is already stored in the H&AL which is okay for mvp

PR #9047, opened 2022-08-10
[MB-13443] Add ability to use mock DTOD

reggieriser:
## Summary

In the non-ATO environments (demo, experimental, loadtest, ephemeral), we get errors when attempting to use the pricer in a way that requires distance calculations via the external DTOD API (because of a certificate error).  There also may be other cases (local dev) where we don't want to hit the real DTOD API, or perhaps the service is down.

This PR adds a new environment variable (`DTOD_USE_MOCK`) that, when set to true, will use a mocked-but-deterministic DTOD.  The default value of this environment variable is false (i.e., use the real DTOD) to match current behavior, but I went ahead and set it to true in the non-ATO environments.

For the mock, a pair of pickup/destination zips should always return the same (made up) distance.  I've tried to create some ranges for zip3 and zip5 distances that seem sort-of plausible without spending a lot of effort on it.

Note that I had to do some refactoring to be able to expose the DTOD interface in a way that made it more easily mockable.

I have tried this out in the ephemeral and experimental environments and the PPM incentive calculated without error; before this PR, the endpoint failed with a 500 server error.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

One straighforward way to see if things are mocked is to go through the customer onboarding process and add a PPM.  After entering the weight of the PPM and submitting the page, you'd previously get a 500 server error in the non-ATO environments.  You should get a mocked distance now and things should proceed fine.

### Backend


pkg/cli/route.go
```diff
@@ -36,6 +39,8 @@ func InitRouteFlags(flag *pflag.FlagSet) {
 	flag.String(DTODApiPasswordFlag, "", "DTOD api auth password")
 	flag.String(DTODApiURLFlag, "", "URL for sending a SOAP request to DTOD")
 	flag.String(DTODApiWSDLFlag, "", "WSDL for sending a SOAP request to DTOD")
+
+	flag.Bool(DTODUseMock, false, "Whether to use a mocked version of DTOD")
 }
 
 // CheckRoute validates Route command line flags
```
ahobson, line 46: Should `CheckRoute` validate that the use mock flag is not enabled in production or staging? We've done that with other flags

reggieriser: Yep, that makes sense.  I'll see how we've done it elsewhere and add that here.

reggieriser: Hopefully fixed now.

ahobson (approved):
Looks great from my perspective

sheenamt (approved):

pearl-truss:
Non blocking question: could we use this mock for Cypress as well? Currently we mock out the API call (which is probably fine) but I took a quick look at this PR and wondered if it could be applied to cypress too

pearl-truss (approved):

scottmries (approved):

reggieriser:
I added some logging to indicate whether a route planner is using the mocked or real DTOD so we can see that in logs.

reggieriser:
> Non blocking question: could we use this mock for Cypress as well? Currently we mock out the API call (which is probably fine) but I took a quick look at this PR and wondered if it could be applied to cypress too

@pearl-truss, that's a good question.  I don't know of any reason we couldn't use this in Cypress tests and/or server tests where we currently mock the DTOD interaction at some other level.  I think I'd prefer to leave that research/implementation as a separate PR after this lands, though, since those environments are not currently broken and a change might require adjustments to tests if we want to switch over from the current mocking strategy.  I've made a new chore ([MB-13499](https://dp3.atlassian.net/browse/MB-13499)) to capture that work.

pearl-truss:
> @pearl-truss, that's a good question. I don't know of any reason we couldn't use this in Cypress tests and/or server tests where we currently mock the DTOD interaction at some other level. I think I'd prefer to leave that research/implementation as a separate PR after this lands, though, since those environments are not currently broken and a change might require adjustments to tests if we want to switch over from the current mocking strategy. I've made a new chore ([MB-13499](https://dp3.atlassian.net/browse/MB-13499)) to capture that work.

Thanks that makes sense to me and thanks for looking into this!

pkg/route/dtod_zip5_distance_mock.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/transcom/mymove/pkg/appcontext"
+)
+
+const (
+	minZip5Distance = 10
+	maxZip5Distance = 3500
+	zip5Range       = 99999
+
+	minZip3Distance = 1
```
duncan-truss, line 16: When testing SIT service items for PPMs I saw that DTOD would return a distance of zero when the origin and destination are the same postal codes.  In that scenario usually our app code will error out, which I think is fine if that's what the real DTOD API is doing as well.

pkg/route/dtod_zip5_distance_mock_test.go
```diff
@@ -0,0 +1,45 @@
...
+	}{
+		{"zip5 distance", "30907", "29212", 69, false, ""},
+		{"zip5 distance reversed", "29212", "30907", 69, false, ""},
+		{"zip5 distance min", "30900", "30899", minZip5Distance, false, ""},
+		{"zip5 distance max", "00000", "99999", maxZip5Distance, false, ""},
+		{"zip3 distance", "30907", "30901", 7, false, ""},
+		{"zip3 distance reversed", "30901", "30907", 7, false, ""},
+		{"zip3 distance min", "30907", "30907", minZip3Distance, false, ""},
+		{"zip3 distance max", "30900", "30999", maxZip3Distance, false, ""},
+		{"too short pickup zip", "3090", "29212", 0, true, "pickup zip must be at least 5 digits"},
```
duncan-truss, line 24: I think DTOD returns a value of -1 for distance when there is an error at least according to this comment https://github.com/transcom/mymove/blob/3c60007df7bc6c246e5d957324a7766a0852882f/pkg/route/dtod_zip5_distance.go#L118.

If there is an error though we shouldn't be using the value anyway.

pkg/route/dtod_zip5_distance_mock.go
reggieriser: Pretty easy fix, so went ahead and took care of it to keep things in sync.

pkg/route/dtod_zip5_distance_mock_test.go
reggieriser: Pretty easy fix, so went ahead and took care of it to keep things in sync.

PR #9060, opened 2022-08-12
[ MB-13183 ] Refactor ECS Deploy task definition

rogeruiz:
We realized in transcom/terraform-aws-app-environment#215 that while we
define container and task definitions in Terraform, we override them in
the deployment of our ECS scheduled tasks within the
`cmd/ecs-deploy/task_def.go`. This means that rather than setting things
in our Terraform, we'll need to define the task and container in the Go
binary.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -366,7 +366,7 @@ commands:
       - announce_failure
       - deploy:
           name: Deploy export db data service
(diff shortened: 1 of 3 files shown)


michaelvaldes-truss (approved):
lgtm! I like that it leaves an obvious place for more storage options in the future

PR #9069, opened 2022-08-15
MB-13179-create-update-movingExpense-handler

NamibiaTorres:
## Summary

This PR enables users to create and update a moving expense

Files that were added or updated:
* internal.yaml
* MovingExpense.yaml
* handlers/internalapi/moving_expense.go
* handlers/internalapi/moving_expense_test.go
* handlers/internalapi/internalapi/payload_to_model
* handlers/internalapi/internalapi/model_to_payload

## Setup to Run Your Code

<details>
##### Terminal 1

Run Create tests

```sh
go test ./pkg/handlers/internalapi --testify.m TestCreateMovingExpenseHandler  -v
```

##### Terminal 2

Run Update Tests:

```sh
go test ./pkg/handlers/internalapi --testify.m TestUpdateMovingExpenseHandler  -v
```

</details>


swagger-def/definitions/MovingExpense.yaml
```diff
@@ -0,0 +1,58 @@
+type: object
```
pearl-truss, line 1: **Praise**: I think it's nice to have this broken down into its own reusable definition

swagger/ghc.yaml
```diff
@@ -5610,6 +5610,10 @@ definitions:
         items:
```
pearl-truss, line 5610: Question: I notice this file was updated but the swagger-def/ghc.yaml was not (normally I would only expect to file changes here after changes were made to swagger-def/ghc.yaml )

pkg/handlers/internalapi/internal/payloads/model_to_payload.go
```diff
@@ -279,6 +279,37 @@ func PayloadForUploadModel(
...
+
+	payload := &internalmessages.MovingExpense{
+		ID:                handlers.FmtUUID(movingExpense.ID),
+		PpmShipmentID:     handlers.FmtUUID(movingExpense.PPMShipmentID),
+		PpmShipment:       &internalmessages.PPMShipment{},
+		DocumentID:        handlers.FmtUUID(movingExpense.DocumentID),
+		Document:          document,
+		CreatedAt:         handlers.FmtDateTime(movingExpense.CreatedAt),
+		UpdatedAt:         handlers.FmtDateTime(movingExpense.UpdatedAt),
+		DeletedAt:         *handlers.FmtDateTime(*movingExpense.DeletedAt),
```
pearl-truss, line 298: Screen Shot 2022-08-16 at 2 23 52 PM

FWIW I noticed that one of the failing server test for the nil issue pointed back to this line. I don't believe you need the `*` in front of handlers and/or movingExpense in this line

swagger/ghc.yaml
felipe-lee: I think it's because `MovingExpense` was added to the shared `PPMShipment` definition, so that got added to the generated `ghc.yaml`

pkg/handlers/internalapi/internal/payloads/model_to_payload.go
NamibiaTorres: Good catch! I narrowed it down to this earlier and haven't pushed up the code yet.

pkg/handlers/internalapi/moving_expense.go
```diff
@@ -0,0 +1,142 @@
...
+		ppmShipmentID, err := uuid.FromString(params.PpmShipmentID.String())
+		if err != nil {
+			appCtx.Logger().Error("missing PPM Shipment ID", zap.Error(err))
+			return movingexpenseops.NewCreateMovingExpenseBadRequest(), nil
+		}
+
+		movingExpense, err := h.movingExpenseCreator.CreateMovingExpense(appCtx, ppmShipmentID)
+
+		if err != nil {
+			appCtx.Logger().Error("internalapi.CreateMovingExpenseHandler", zap.Error(err))
```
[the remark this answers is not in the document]
NamibiaTorres: That's a great suggestion, and I think your instinct is right about it. I can add a check.

swagger-def/internal.yaml
```diff
@@ -1854,6 +1876,17 @@ definitions:
...
+    type: string
+    title: PPM document status
+    enum:
+      - APPROVED
+      - EXCLUDED
+      - REJECTED
+    x-display-value:
+      APPROVED: Approved
+      EXCLUDED: Excluded
+      REJECTED: Rejected
```
felipe-lee, line 1886: **todo**: Can you also make this a its own definition in the shared definitions?

swagger-def/definitions/MovingExpense.yaml
```diff
@@ -0,0 +1,58 @@
...
+  ppmShipment:
+    $ref: 'PPMShipment.yaml'
+  documentId:
+    type: string
+    format: uuid
+    example: c56a4180-65aa-42ec-a945-5fd21dec0538
+  document:
+    type: object
+  movingExpenseType:
+    $ref: '../internal.yaml#/definitions/MovingExpenseType'
```
felipe-lee, line 20: **suggestion**: I know you're re-using the `MovingExpenseType` definition that already existed, but can you move it to the shared definitions folder since it's going to be used in this shared definition?

swagger-def/internal.yaml
```diff
@@ -908,6 +908,25 @@ definitions:
...
+    type: object
+    properties:
+      movingExpenseType:
+        $ref: '#/definitions/MovingExpenseType'
+      description:
+        type: string
+      status:
+        $ref: '#/definitions/PPMDocumentStatus'
+      reason:
+        type: string
```
felipe-lee, line 921: **todo**: Status and reason aren't something that the customer should be able to set. Can you take these out of this definition?

Also, need to be able to set `paidWithGTCC`, `amount`, and `missingReceipt`

swagger-def/internal.yaml
```diff
@@ -908,6 +908,25 @@ definitions:
       - moving_expense_type
       - requested_amount_cents
       - payment_method
+  UpdateMovingExpense:
+    type: object
+    properties:
```
felipe-lee, line 913: **suggestion (non-blocking)**: Since we're updating this all in a single request, we can probably set most of these fields to be required, except the SIT related ones.

swagger-def/definitions/MovingExpense.yaml
```diff
@@ -0,0 +1,58 @@
...
+  createdAt:
+    type: string
+    format: date-time
+  updatedAt:
+    type: string
+    format: date-time
+  deletedAt:
+    type: string
+    format: date-time
+    readOnly: true
```
felipe-lee, line 50: **todo**: Can you remove the `deletedAt` from the payload? We don't really need to have that info on the FE since we'll be filtering out the deleted moving expenses probably.

swagger-def/definitions/MovingExpense.yaml
```diff
@@ -0,0 +1,58 @@
+type: object
+properties:
```
felipe-lee, line 2: **todo**: Can you set the `readOnly: true` attributes for the `id` fields and the `x-nullable: true` & `x-omitempty: false` for any non-required fields?

NamibiaTorres: Should I remove the `readOnly: true` from the DeletedAt field

swagger-def/definitions/MovingExpense.yaml
NamibiaTorres: Yes!

pkg/handlers/internalapi/internal/payloads/model_to_payload.go
```diff
@@ -279,6 +279,36 @@ func PayloadForUploadModel(
...
+		UpdatedAt:         handlers.FmtDateTime(movingExpense.UpdatedAt),
+		MovingExpenseType: internalmessages.MovingExpenseType(*movingExpense.MovingExpenseType),
+		Description:       *movingExpense.Description,
+		PaidWithGtcc:      *movingExpense.PaidWithGTCC,
+		Amount:            *handlers.FmtCost(movingExpense.Amount),
+		MissingReceipt:    *movingExpense.MissingReceipt,
+		Status:            internalmessages.MovingExpense{}.Status,
+		Reason:            *movingExpense.Reason,
+		SitStartDate:      *handlers.FmtDate(*movingExpense.SITStartDate),
+		SitEndDate:        *handlers.FmtDate(*movingExpense.SITEndDate),
```
felipe-lee, line 306: **suggestion**: With the latest changes to the swagger definitions, you'll need to regen the server files because some of these are now pointers on `internalmessages.MovingExpense`, which makes this a bit easier since they'll be pointers both in the model and payload types.

swagger-def/definitions/MovingExpense.yaml
felipe-lee: that [field isn't needed](https://github.com/transcom/mymove/pull/9069/files#r947302506) so it can be removed entirely

felipe-lee (commented):
This is looking good so far and I think it's close! I left comments for some small changes or missing things. One other thing that's missing is that the handlers need to be added to the `pkg/handlers/internalapi/api.go` file, kind of like we did with the weight ticket handlers: https://github.com/transcom/mymove/blob/MB-13179-create-update-moving-expense-handler/pkg/handlers/internalapi/api.go#L96-L97

Let me know if you have any questions or want to pair on any of this.

pkg/handlers/internalapi/moving_expense.go
```diff
@@ -0,0 +1,142 @@
...
+
+			if err != nil {
+				appCtx.Logger().Error("internalapi.UpdateMovingExpenseHandler", zap.Error(err))
+				switch e := err.(type) {
+				case apperror.InvalidInputError:
+					return movingexpenseops.NewUpdateMovingExpenseUnprocessableEntity().WithPayload(payloads.ValidationError(handlers.ValidationErrMessage, h.GetTraceIDFromRequest(params.HTTPRequest), e.ValidationErrors)), err
+				case apperror.PreconditionFailedError:
+					return movingexpenseops.NewUpdateMovingExpensePreconditionFailed().WithPayload(payloads.ClientError(handlers.PreconditionErrMessage, err.Error(), h.GetTraceIDFromRequest(params.HTTPRequest))), err
+				case apperror.ForbiddenError:
+					return movingexpenseops.NewUpdateMovingExpenseForbidden().WithPayload(payloads.ClientError(handlers.PreconditionErrMessage, err.Error(), h.GetTraceIDFromRequest(params.HTTPRequest))), err
```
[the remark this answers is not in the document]
NamibiaTorres: `if etag.GenerateEtag(originalMovingExpense.UpdatedAt) != eTag {
		return nil, apperror.NewPreconditionFailedError(originalMovingExpense.ID, nil)
	}`

After double checking, only the PreconditionedFailed. Do you recommend adding the Forbidden error or removing it from swagger and  the handler?

pkg/handlers/internalapi/internal/payloads/model_to_payload.go
```diff
@@ -279,6 +279,44 @@ func PayloadForUploadModel(
...
+		PpmShipment:    &internalmessages.PPMShipment{},
+		DocumentID:     handlers.FmtUUID(movingExpense.DocumentID),
+		Document:       document,
+		CreatedAt:      handlers.FmtDateTime(movingExpense.CreatedAt),
+		UpdatedAt:      handlers.FmtDateTime(movingExpense.UpdatedAt),
+		Description:    movingExpense.Description,
+		PaidWithGtcc:   movingExpense.PaidWithGTCC,
+		Amount:         handlers.FmtCost(movingExpense.Amount),
+		MissingReceipt: movingExpense.MissingReceipt,
+		Status:         internalmessages.MovingExpense{}.Status,
```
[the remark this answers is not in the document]
NamibiaTorres: I see because the status can be nil in the create up until it is set during the approve or exclusion/rejection process

swagger-def/definitions/MovingExpense.yaml
```diff
@@ -0,0 +1,79 @@
...
+    type: string
+    format: uuid
+    example: c56a4180-65aa-42ec-a945-5fd21dec0538
+    readOnly: true
+  ppmShipmentId:
+    type: string
+    format: uuid
+    example: c56a4180-65aa-42ec-a945-5fd21dec0538
+  ppmShipment:
+    $ref: 'PPMShipment.yaml'
```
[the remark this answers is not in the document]
NamibiaTorres: Can you walk me through this later?

pkg/handlers/internalapi/internal/payloads/model_to_payload.go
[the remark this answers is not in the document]
felipe-lee: Yeah. `reason` can be too, but since that one is a regular string pointer, no conversion is needed so we can pass the pointer directly.

swagger-def/definitions/MovingExpense.yaml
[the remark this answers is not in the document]
felipe-lee: Yeah sure, lmk when you want to talk.

felipe-lee:
@NamibiaTorres I think something went wrong with the [latest commit](https://github.com/transcom/mymove/pull/9069/commits/15f62e0e649bf2a97cc838299b61dce9048d4412). I'm seeing 1233 files changed, and it looks like a bunch of gen files were deleted:

image

image

NamibiaTorres:
> @NamibiaTorres I think something went wrong with the [latest commit](https://github.com/transcom/mymove/pull/9069/commits/15f62e0e649bf2a97cc838299b61dce9048d4412). I'm seeing 1233 files changed, and it looks like a bunch of gen files were deleted:
> 
> image
> 
> image

Fixed!

pkg/handlers/internalapi/moving_expense.go
```diff
@@ -0,0 +1,142 @@
...
+		movingExpense, err := h.movingExpenseCreator.CreateMovingExpense(appCtx, ppmShipmentID)
+
+		if err != nil {
+			appCtx.Logger().Error("internalapi.CreateMovingExpenseHandler", zap.Error(err))
+			// Can get a status error
+			// Can get an DB error - check if moving expense doc creates
+			// Can get an error for whether the PPM exist
+			switch err.(type) {
+			case apperror.ForbiddenError:
+				return movingexpenseops.NewCreateMovingExpenseForbidden(), err
```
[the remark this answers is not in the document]
NamibiaTorres: Checked and it does not, I can remove this check.

pkg/handlers/internalapi/moving_expense.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+
+		movingExpense, err := h.movingExpenseCreator.CreateMovingExpense(appCtx, ppmShipmentID)
+
+		if err != nil {
+			appCtx.Logger().Error("internalapi.CreateMovingExpenseHandler", zap.Error(err))
+			// Can get a status error
+			// Can get an DB error - check if moving expense doc creates
+			// Can get an error for whether the PPM exist
+			switch err.(type) {
```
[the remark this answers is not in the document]
NamibiaTorres: If I've added a `QueryErr`, do I need a `NewQueryError`

pkg/handlers/internalapi/moving_expense_test.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+
+	suite.Run("Successfully Update Moving Expense - Integration Test", func() {
+		appCtx := suite.AppContextForTest()
+
+		subtestData := makeUpdateSubtestData(appCtx, true)
+
+		params := subtestData.params
+		// Add a Description
+		contractedExpense := internalmessages.MovingExpenseType("CONTRACTED_EXPENSE")
```
felipe-lee, line 192: **suggestion (non-blocking)**: Since we have a constant that has this value, it would be better to use it than a hard-coded value

```suggestion
		contractedExpense := internalmessages.MovingExpenseType(models.MovingExpenseReceiptTypeContractedExpense)
```

pkg/handlers/internalapi/moving_expense_test.go
```diff
@@ -0,0 +1,314 @@
...
+		suite.IsType(&movingexpenseops.UpdateMovingExpenseUnprocessableEntity{}, response)
+	})
+
+	suite.Run("PATCH failure - 500 - server error", func() {
+		mockUpdater := mocks.MovingExpenseUpdater{}
+		appCtx := suite.AppContextForTest()
+
+		subtestData := makeUpdateSubtestData(appCtx, true)
+		params := subtestData.params
+		contractedExpense := internalmessages.MovingExpenseType("CONTRACTED_EXPENSE")
```
felipe-lee, line 287: **suggestion (non-blocking)**: Since we have a constant that has this value, it would be better to use it than a hard-coded value
```suggestion
		contractedExpense := internalmessages.MovingExpenseType(models.MovingExpenseReceiptTypeContractedExpense)
```

pkg/handlers/internalapi/moving_expense.go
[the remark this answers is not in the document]
felipe-lee: Ah, no, sorry that was from me copy/pasting from where the error gets set. You are catching the correct one!

felipe-lee (approved):
Thank you for taking all that feedback into account! Lgtm! I left two last comments, but they are by no means blocking so this is good to go.

swagger-def/definitions/MovingExpense.yaml
```diff
@@ -0,0 +1,72 @@
...
+    example: '2018-05-26'
+    format: date
+    x-nullable: true
+    x-omitempty: false
+  createdAt:
+    type: string
+    format: date-time
+  updatedAt:
+    type: string
+    format: date-time
```
duncan-truss, line 65: Could we add an eTag field here, I don't think we usually ask the frontend to generate it from the updatedAt field value.

felipe-lee: Ah, I'm glad you caught that

pkg/handlers/internalapi/moving_expense.go
```diff
@@ -0,0 +1,170 @@
...
+		// Get the MovingExpense ID
+		err = appCtx.DB().Find(movingExpense, movingExpense.ID)
+		if err != nil {
+			switch err {
+			case sql.ErrNoRows:
+				return nil, apperror.NewNotFoundError(movingExpense.ID, "looking for Moving Expense")
+			default:
+				return nil, apperror.NewQueryError("Moving Expense", err, "")
+			}
+		}
```
duncan-truss, line 87: This looks unnecessary to me to query for the moving expense that you just created above on https://github.com/transcom/mymove/pull/9069/files#diff-6b77b29645c288426e58ac4d59962837310e4699111858bfa384ed1f7ac7fe69R44.

I think you can add the NotFoundError to your error block above, if the api was passed a PPMShipmentID that didn't exist.

pkg/handlers/internalapi/moving_expense.go
```diff
@@ -0,0 +1,170 @@
...
+
+type CreateMovingExpenseHandler struct {
+	handlers.HandlerConfig
+	movingExpenseCreator services.MovingExpenseCreator
+}
+
+// Handle creates a moving expense
+func (h CreateMovingExpenseHandler) Handle(params movingexpenseops.CreateMovingExpenseParams) middleware.Responder {
+	return h.AuditableAppContextFromRequestWithErrors(params.HTTPRequest, func(appCtx appcontext.AppContext) (middleware.Responder, error) {
+		if appCtx.Session() == nil {
```
duncan-truss, line 28: This check for a nil session should be happening already in our [user auth middleware](https://github.com/transcom/mymove/blob/42205c4f1af6e3d3e274be6a9f230e813f8911e4/pkg/handlers/authentication/auth.go#L116-L128) that is called prior to our handler so you could remove it.  I think the one below seems still needed though, thanks for adding that one.

pkg/handlers/internalapi/moving_expense.go
```diff
@@ -0,0 +1,170 @@
...
+				return movingexpenseops.NewUpdateMovingExpenseForbidden(), noServiceMemberIDErr
+			}
+
+			payload := params.UpdateMovingExpense
+			if payload == nil {
+				noBodyErr := apperror.NewBadDataError("Invalid moving expense: params UpdateMovingExpense is nil")
+				appCtx.Logger().Error(noBodyErr.Error())
+				return movingexpenseops.NewUpdateMovingExpenseBadRequest().WithPayload(payloads.ClientError(handlers.BadRequestErrMessage,
+					"The moving expense request payload cannot be empty.", h.GetTraceIDFromRequest(params.HTTPRequest))), noBodyErr
+			}
```
duncan-truss, line 121: This isn't hurting anything but I don't think you'll ever hit this logic because the updateMovingExpense payload is marked as required in your [internal.yaml endopoint definition](https://github.com/transcom/mymove/pull/9069/files#diff-d75579e46ab58b0cd40af93d565278cc0643cb0a2de4ea57a065ab1da6b1d5e2R4517).  This is a case where swagger will reject the request if you are using say Postman or the using the real application but because we call Handle() directly in our tests we don't get that pre-validate step.  That's why in our handler tests we should always call `params.Body.Validate(strfmt.Default)` to catch swagger validations.

pkg/handlers/internalapi/moving_expense_test.go
```diff
@@ -0,0 +1,316 @@
...
+		subtestData := makeCreateSubtestData(appCtx, true)
+
+		response := subtestData.handler.Handle(subtestData.params)
+
+		suite.IsType(&movingexpenseops.CreateMovingExpenseOK{}, response)
+
+		createdMovingExpense := response.(*movingexpenseops.CreateMovingExpenseOK).Payload
+
+		suite.NotEmpty(createdMovingExpense.ID.String())
+		suite.NotNil(createdMovingExpense.PpmShipmentID.String())
```
duncan-truss, line 70: I think I would check that this createMovingExpense.PpmShipmentID is equal to the value of subtestData.ppmShipment.ID.  I don't think this string could ever be nil since it's not a pointer but it could be the empty string.

pkg/handlers/internalapi/moving_expense.go
NamibiaTorres: Thanks Duncan. Are you suggesting I take out this entire section?

pkg/handlers/internalapi/moving_expense.go
NamibiaTorres: When I took this check out, however, I saw a failure in the 401 CreateMovingExpense test.

NamibiaTorres:
I updated with PR with suggestions from Duncan, after it was approved. The tests pass and it looks like nothing is breaking after implementing changes.

pkg/handlers/internalapi/moving_expense.go
duncan-truss: That's true calling this handler directly won't return a 401 error for a request without a session anymore if you removed this logic.  In reality though our middleware wraps every request and would return an error first before ever calling the moving expense handler authentication.UserAuthMiddleware -> createMovingExpenseHandler.Handle.  In the handler tests though we aren't adding the middleware so you wouldn't expect it to flag a nil session.

PR #9083, opened 2022-08-17
[ MB-13183 ] Removing export-db-data prepwork from DP3 Dockerfile

rogeruiz:
Since we aren't going to enable the Advana integration in environments
outside of the ATO boundary, we don't need this in our Dockerfile that
is used for dp3.us environments.

### DIFF
--- a/Dockerfile.tasks_dp3
+++ b/Dockerfile.tasks_dp3
@@ -1,13 +1,3 @@
-###########
-# BUILDER-POSTGRES #
-###########
-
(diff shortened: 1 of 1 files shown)


rogeruiz:
Adding the `do-not-merge` label here as we don't want this merged in until #8953 is finally ready to be merged in. I'll be creating a similar PR for removing the Experimental deployment and making sure we have the `ecs-deploy` commands in CircleCI setup correctly.

PR #9102, opened 2022-08-23
MB-13577: Allow Decimals in Expense Amount 

pearl-truss:
## Summary

This PR allows the expense amount to include decimals. 

**NOTE Engineering:** 
- This assumes we are storing the expense amount in cents in the database like we are with the estimated incentive amount

**NOTE Design:**
- If someone leaves the text field after inputting 160.9 a trailing zero will be padded to the end to make it 160.90
- This PR didn't change the styles of anything on the page. Only the behavior of the Amount field on the expense form to allow decimal inputs

## Setup to Run Your Code

Start the Storybook locally.

```sh
make storybook
```
**If you don't want to run storybook locally, changes can be viewed here:** 

## Screenshots
Screen Shot 2022-08-23 at 9 08 16 AM

### DIFF
--- a/src/components/Customer/PPM/Closeout/ExpenseForm/ExpenseForm.jsx
+++ b/src/components/Customer/PPM/Closeout/ExpenseForm/ExpenseForm.jsx
@@ -26,7 +26,7 @@ const validationSchema = Yup.object().shape({
   expenseType: Yup.string().required('Required'),
   description: Yup.string().required('Required'),
   paidWithGTCC: Yup.boolean().required('Required'),
-  amount: Yup.number().required('Required'),
+  amount: Yup.string().required('Required'),
   missingReceipt: Yup.boolean().required('Required'),
   receiptDocument: Yup.array().of(uploadShape).min(1, 'At least one upload is required'),
   sitStartDate: Yup.date()
@@ -123,8 +123,11 @@ const ExpenseForm = ({
                         label="Amount"
                         id="amount"
                         mask={Number}
-                        scale={0} // digits after point, 0 for integers
+                        scale={2} // digits after point, 0 for integers
                         signed={false} // disallow negative
+                        radix="." // fractional delimiter
+                        mapToRadix={['.']} // symbols to process as radix
+                        padFractionalZeros // if true, then pads zeros at end to the length of scale
                         thousandsSeparator=","
                         lazy={false} // immediate masking evaluation
                         prefix="$"
--- a/src/components/Customer/PPM/Closeout/ExpenseForm/ExpenseForm.stories.jsx
+++ b/src/components/Customer/PPM/Closeout/ExpenseForm/ExpenseForm.stories.jsx
@@ -56,7 +56,7 @@ ExistingExpenses.args = {
     description: 'bubble wrap',
     missingReceipt: false,
     paidWithGTCC: false,
-    amount: 600,
+    amount: '600',
     receiptDocument: {
       uploads: [
         {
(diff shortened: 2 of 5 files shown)


duncan-truss:
Do you mind defining these expense types somewhere so that we can also make the propTypes limited to only the valid values as well instead of any string?  I think the backend also expects uppercase constants looking at the models file.

https://github.com/transcom/mymove/blob/c6cd62cc4b0ef75a1e6ad17cda0383d5736d3916/src/components/Customer/PPM/Closeout/ExpenseForm/ExpenseForm.jsx#L71-L79

pearl-truss:
Sorry I was late seeing this. I pushed up the change but one thing is that the options aren't proptypes on the form that are a part of the component. Let me know if you have strong feelings on moving them out of the component

duncan-truss:
> Sorry I was late seeing this. I pushed up the change but one thing is that the options aren't proptypes on the form that are a part of the component. Let me know if you have strong feelings on moving them out of the component

Thanks for pulling out those into a constants. I wasn't thinking exposing the options as props either, I meant that the `ExpenseShape` could use your new constants because right now it's just defined as a string.  So like

```
const ExpenseShape {
...
expenseType: propTypes.oneOf(Object.values(ppmExpenseTypes))
...
}
```

src/components/Customer/PPM/Closeout/ExpenseForm/ExpenseForm.jsx
```diff
@@ -26,19 +28,19 @@ const validationSchema = Yup.object().shape({
   expenseType: Yup.string().required('Required'),
   description: Yup.string().required('Required'),
   paidWithGTCC: Yup.boolean().required('Required'),
-  amount: Yup.number().required('Required'),
+  amount: Yup.string().required('Required'),
```
duncan-truss, line 31: Should we worry about accepting a zero value here?

pearl-truss:
> > Sorry I was late seeing this. I pushed up the change but one thing is that the options aren't proptypes on the form that are a part of the component. Let me know if you have strong feelings on moving them out of the component
> 
> Thanks for pulling out those into a constants. I wasn't thinking exposing the options as props either, I meant that the `ExpenseShape` could use your new constants because right now it's just defined as a string. So like
> 
> ```js
> const ExpenseShape {
> ...
> expenseType: propTypes.oneOf(Object.values(ppmExpenseTypes))
> ...
> }
> ```

:doh: gotcha

src/components/Customer/PPM/Closeout/ExpenseForm/ExpenseForm.jsx
pearl-truss: hmm I think we probably don't want zero here. I'll go ahead and make the change but @scarequotes if you feel otherwise, I can revert it later

pearl-truss: Screen Shot 2022-08-24 at 11 24 01 AM

scarequotes (approved):
Nice work. Looks good to me!

src/components/Customer/PPM/Closeout/ExpenseForm/ExpenseForm.jsx
```diff
@@ -26,19 +28,19 @@ const validationSchema = Yup.object().shape({
   expenseType: Yup.string().required('Required'),
   description: Yup.string().required('Required'),
   paidWithGTCC: Yup.boolean().required('Required'),
-  amount: Yup.number().required('Required'),
+  amount: Yup.string().notOneOf(['0', '0.00'], 'Please enter a non-zero amount').required('Required'),
```
duncan-truss, line 31: Cool I haven't seen `notOneOf` used before.

duncan-truss (approved):
This looks good, you will need to update the storybook args to match the new uppercase expense type constants, because right now the select dropdown shows the default.

pearl-truss:
> This looks good. You will need to update the storybook args to match the new uppercase expense type constants, because right now the select dropdown shows the default.

Thanks, I just pushed an update to use the constants in storybook

PR #9105, opened 2022-08-23
MB-11277: Cypress tests for PPM office flow

ronaktruss:
## Summary

Add end to end flows for the office user reviewing a PPM shipment.

I've added scenarios to cover modifying an existing PPM shipment and adding a second ppm shipment. There were existing test that test the delete functionality. 

There is an existing stand-alone `addShipment.js` test, that attempts to delete an existing shipment and then add a ppmShipment. This test probably isn't necessary, but adds enough value that it's probably not worth removing entirely.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Go server locally.

```sh
make server_run
```

##### Terminal 2

Start the local cypress test.

```sh
make e2e_test_fresh
```

</details>

## Screenshots
image


cypress/integration/office/ppms/happyPathFlows.js
```diff
@@ -0,0 +1,126 @@
...
+
+    cy.intercept('**/ghc/v1/queues/counseling?page=1&perPage=20&sort=submittedAt&order=asc&locator=**').as(
+      'getFilterSortedMoves',
+    );
+    cy.intercept('**/ghc/v1/move/**').as('getMoves');
+    cy.intercept('**/ghc/v1/orders/**').as('getOrders');
+    cy.intercept('**/ghc/v1/move_task_orders/**/mto_shipments').as('getMTOShipments');
+    cy.intercept('**/ghc/v1/move_task_orders/**/mto_shipments/**').as('updateMTOShipments');
+    cy.intercept('**/ghc/v1/mto-shipments').as('createShipment');
+    cy.intercept('**/ghc/v1/move_task_orders/**/mto_service_items').as('getMTOServiceItems');
```
duncan-truss, line 28: I've been bitten by intercepts that match both the GET and PATCH endpoints if we don't explicit declare them.  Do you mind adding them like `cy.intercept('GET', '**/ghc/v1/...'`?

cypress/integration/office/ppms/happyPathFlows.js
```diff
@@ -0,0 +1,126 @@
...
+    cy.get('[data-testid="errorMessage"]').contains('Required');
+    cy.get('[data-testid="counselor-remarks"]').clear().type('Added correct incentive').blur();
+
+    // Submit page two
+    cy.get('[data-testid="submitForm"]').should('be.enabled').click();
+    cy.wait(['@updateMTOShipments', '@getMTOShipments', '@getMoves', '@getOrders', '@getMTOServiceItems']);
+
+    // Expand details and verify information
+    cy.contains('Your changes were saved.');
+    cy.get('[data-prefix="fas"][data-icon="chevron-down"]').last().click();
```
duncan-truss, line 111: Maybe you can scope this using a `within()` to restrict the search to the shipment card you care about.  The move details page has a lot of DOM to search through and these data-testids will have more than one match on the page with 2 PPM shipments.

cypress/integration/office/ppms/happyPathFlows.js
ronaktruss: Yeah I can make this more explicit

cypress/integration/office/ppms/happyPathFlows.js
ronaktruss: Yup, this is still pending

cypress/support/ppmOfficeShared.js
```diff
@@ -20,3 +20,59 @@ export function navigateToShipmentDetails(moveLocator) {
...
+    expectedDepartureDate = '09 Jun 2022',
+    pickupPostalCode = '90210',
+    secondPickupPostalCode = '07003',
+  } = options;
+
+  cy.get('input[name="expectedDepartureDate"]').clear().type(expectedDepartureDate).blur();
+  cy.get('input[name="pickupPostalCode"]').clear().type(pickupPostalCode).blur();
+  if (secondPickupPostalCode) {
+    cy.get('input[name="secondPickupPostalCode"]').clear().type(secondPickupPostalCode).blur();
+  }
```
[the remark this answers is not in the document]
ronaktruss: I added the clear to support the scenario where we call this function during the edit flow. Right now we don't do this so it's not necessary. Personally I think its fine to keep this as is, but it could mean that we are missing the scenario where we are doing something weird with initial values. (ie. if you have the initial value as undefined and then change it to the typed value this cypress test doesn't mimic that 1 to 1 because we are effectively going from the clear value which is "" to the typed value).

cypress/support/ppmOfficeShared.js
```diff
@@ -20,3 +20,59 @@ export function navigateToShipmentDetails(moveLocator) {
...
+};
+
+export const fillOutDestinationInfo = (options = {}) => {
+  const { destinationPostalCode = '76127', secondDestinationPostalCode = '08540' } = options;
```
duncan-truss, line 65: I'm not sure this secondary destination postal code makes much sense but it shouldn't impact anything.

cypress/support/ppmOfficeShared.js
```diff
@@ -20,3 +20,59 @@ export function navigateToShipmentDetails(moveLocator) {
...
+  cy.get('input[name="sitEstimatedWeight"]').clear().type(sitEstimatedWeight).blur();
+  cy.get('input[name="sitEstimatedEntryDate"]').clear().type(sitEstimatedEntryDate).blur();
+  cy.get('input[name="sitEstimatedDepartureDate"]').clear().type(sitEstimatedDepartureDate).blur();
+};
+
+export const fillOutOriginInfo = (options = {}) => {
+  const {
+    expectedDepartureDate = '09 Jun 2022',
+    pickupPostalCode = '90210',
+    secondPickupPostalCode = '07003',
```
duncan-truss, line 54: This secondary pickup postal code is across the country but it's just test data I guess 😛

cypress/integration/office/ppms/addShipment.js
```diff
@@ -51,46 +57,39 @@ describe('Services counselor user', () => {
...
-    // cy.get('[data-testid="expectedDepartureDate"]').contains('09 Jun 2022');
-    // cy.get('[data-testid="originZIP"]').contains('90210');
-    // cy.get('[data-testid="destinationZIP"]').contains('76127');
-    // cy.get('[data-testid="sitPlanned"]').contains('no');
-    // cy.get('[data-testid="estimatedWeight"]').contains('4,000 lbs');
-
-    // TODO uncomment these assertions after display bug has been fixed
-    // cy.get('[data-testid="hasRequestedAdvance"]').contains('Yes, $5,987');
-    // cy.get('[data-testid="counselorRemarks"]').contains('The requested advance amount has been added.');
+    fillOutIncentiveAndAdvance({ advance: '5987' });
```
duncan-truss, line 69: I see you test for the estimated incentive showing after saving page 2 on the shipment card.  It would be good to verify it's shown on page 2 as well, I don't see us testing that anywhere.

duncan-truss (approved):

cypress/support/ppmOfficeShared.js
```diff
@@ -20,3 +20,59 @@ export function navigateToShipmentDetails(moveLocator) {
...
+
+export const fillOutWeight = (options = {}) => {
+  const { estimatedWeight = '4000', hasProGear = false, proGearWeight = '1000', spouseProGearWeight = '500' } = options;
+
+  cy.get('input[name="estimatedWeight"]').clear().type(estimatedWeight).blur();
+  if (hasProGear) {
+    cy.get('input[name="hasProGear"][value="yes"]').check({ force: true });
+    cy.get('input[name="proGearWeight"]').type(proGearWeight).blur();
+    cy.get('input[name="spouseProGearWeight"]').type(spouseProGearWeight).blur();
+  }
```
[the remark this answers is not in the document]
ronaktruss: Yeah it should.

cypress/integration/office/ppms/happyPathFlows.js
```diff
@@ -0,0 +1,129 @@
...
+    cy.get('[data-testid="expectedDepartureDate"]').contains('15 Mar 2020');
+    cy.get('[data-testid="originZIP"]').contains('90210');
+    cy.get('[data-testid="secondOriginZIP"]').contains('90211');
+    cy.get('[data-testid="destinationZIP"]').contains('30813');
+    cy.get('[data-testid="secondDestinationZIP"]').contains('30814');
+    cy.get('[data-testid="sitPlanned"]').contains('Yes');
+    cy.get('[data-testid="estimatedWeight"]').contains('4,000 lbs');
+    cy.get('[data-testid="proGearWeight"]').contains('Yes, 1,987 lbs');
+    cy.get('[data-testid="spouseProGear"]').contains('Yes, 498 lbs');
+    cy.get('[data-testid="estimatedIncentive"]').contains('$10,000');
```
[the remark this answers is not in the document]
ronaktruss: Yeah that's correct.

PR #9110, opened 2022-08-24
MB-12492 convert move_history tests 

mlm55:
## Summary

Converts move_history tests to use transactions.

[this article](https://transcom.github.io/mymove-docs/docs/backend/testing/running-server-tests-inside-a-transaction/) explains more about the approach used.

##### Terminal 1
Run the tests and make sure they pass
```sh
go test -count 1 -v ./pkg/services/move_history
```


michaelvaldes-truss (approved):
lgtm!

PR #9113, opened 2022-08-25
MB-13329 As a Customer I will see a page where I can enter my pro gear details

scottmries:
## Summary

This PR adds a page for the Pro-gear form, where weight tickets will be able to be added for pro-gear, and the first field for this form, including the Storybook component Customer Components > PPM Closeout > Pro Gear > Default, [at this link](https://output.circle-artifacts.com/output/job/e9962dea-26d8-40fe-b67c-7abd44cc706c/artifacts/0/storybook/index.html?path=/story/customer-components-ppm-closeout-pro-gear--default).

Note that this page appears both when adding and editing a Pro-gear ticket, and while editing is not yet hooked up on the front-end, the route is in place.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the[ customer site](https://my-milmove-pr-9113.mymove.sandbox.truss.coffee/)
2. Login as a progearWeightTicket@ppm.approved 
3. Click the "Upload PPM Documents" button
4. Click the "Add Pro-Gear Weight" button

### Frontend


scarequotes (approved):
LGTM!

src/pages/MyMove/PPM/Closeout/ProGear/ProGear.jsx
```diff
@@ -0,0 +1,42 @@
...
+          <Grid col desktop={{ col: 8, offset: 2 }}>
+            <ShipmentTag shipmentType={shipmentTypes.PPM} />
+            <h1>Pro-gear</h1>
+            <div className={closingPageStyles['closing-section']}>
+              <p>
+                If you moved pro-gear for yourself or your spouse as part of this PPM, document the total weight here.
+                Reminder: This pro-gear should be included in your total weight moved.
+              </p>
+            </div>
+            <ProGearForm onBack={handleBack} onSubmit={handleSubmit} />
```
ronaktruss, line 34: (No change needed) Right now the `setNumber` is effectively hard coded. When adding multiple pro-gear weights we will need to make an adjustment ensure that the correct number shows up.

Right now for weightTickets we are trying to leverage the URL, but I don't believe that is currently working as expected. Ref: 

https://user-images.githubusercontent.com/84742904/186957487-47386291-5a62-4f5e-bc33-eefb33cb3b64.mov

ronaktruss (commented):
Looks good so far! Still reviewing, but thought I'd share the comments I have so far.

scottmries:
@ronaktruss 
I pushed updates from your suggestions. 
I went ahead with `&` over `&amp;` since the encoded character isn't used anywhere in the codebase.
I'm also targeting the label on the radio buttons since Cypress complains that the actual `input`s are offscreen, just FYI.

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+import { render, waitFor, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import ProGearForm from 'components/Customer/PPM/Closeout/ProGearForm/ProGearForm';
+
+const defaultProps = {
+  onBack: jest.fn(),
+  onSubmit: jest.fn(),
+  initialValues: {
+    selfProGear: 'true',
```
ronaktruss, line 11: I think the one thing that needs to be changed is that the initial values should likely be boolean values since that is how we will receive the data from the server/DB. In the actual `ProGearForm.jsx` file / component we can continue to use the string values but will need an initial conversion from the boolean that we get from the server to the string value that we use in the rest of the component. Something similar is done in the [ExpenseForm](https://github.com/transcom/mymove/blob/master/src/components/Customer/PPM/Closeout/ExpenseForm/ExpenseForm.jsx#L63).

ronaktruss:
> @ronaktruss I pushed updates from your suggestions. I went ahead with `&` over `&amp;` since the encoded character isn't used anywhere in the codebase. I'm also targeting the label on the radio buttons since Cypress complains that the actual `input`s are offscreen, just FYI.

Thanks for that! 

Apologies for not making this clearer, but I actually think the way you had originally implemented the save and continue button was more correct and that we should potentially look into changing the previous versions that are just using `Save & Continue` directly to something more similar to what you had (though not necessarily in this PR). That said I don't think this has much of a functional impact since most browsers properly translate the `&` to what we expect. Happy to discuss this in more detail if you're interested.

The current cypress test targeting the labels are great! Just FYI I think you might be able to target the inputs doing something like `cy.get('input[name="selfProGear"]');`. This should return both radio options. If you want to select a specific one you can do something like `cy.get('input[name="selfProGear"][value="true"]')` where you replace the value with the respective value that is defined in `ProGearForm.jsx`. To then trigger the "checking" of that radio button you could do something like `cy.get('input[name="selfProGear"][value="false"]').check({ force: true });`

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.test.jsx
scottmries: I made a `ProGearTicketShape` and changed this prop to `proGear`, as I anticipate that will make more sense when we get data from the server. That value's now set to a string based on the incoming value's truthiness, too.

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.module.scss
```diff
@@ -0,0 +1,9 @@
+@import 'shared/styles/_basics';
+
+.ProGearForm {
+  .ownerSectionWrapper {
```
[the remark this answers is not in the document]
scottmries: Removed.

cypress/integration/milmove/ppms/progear.js
```diff
@@ -0,0 +1,34 @@
...
+  viewportType.forEach(({ viewport, isMobile, userId }) => {
+    it(`page loads - ${viewport}`, () => {
+      if (isMobile) {
+        setMobileViewport();
+      }
+
+      signInAndNavigateToWeightTicketPage(userId);
+      navigateFromCloseoutReviewPageToProGearPage();
+      cy.get('#ownerOfProGearSelf').should('exist', 'be.checked');
+      cy.get('#ownerOfProGearSpouse').should('exist', 'not.be.checked');
```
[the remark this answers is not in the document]
scottmries: I updated this to use just the `.should({not.}be.checked)` tests.

src/pages/MyMove/PPM/Closeout/Review/Review.jsx
```diff
@@ -191,7 +191,11 @@ const Review = () => {
                 }
                 contents={proGearContents}
                 renderAddButton={() => (
-                  <Link className="usa-button usa-button--secondary" to={proGearCreatePath}>
+                  <Link
+                    className="usa-button usa-button--secondary"
+                    to={proGearCreatePath}
+                    data-testid="add-progear-weight"
```
ronaktruss, line 197: Since we aren't using this we may want to get rid of this.

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.stories.jsx
```diff
@@ -0,0 +1,31 @@
...
+            <Story />
+          </Grid>
+        </Grid>
+      </GridContainer>
+    ),
+  ],
+  args: {
+    initialValues: {
+      selfProGear: 'true',
+    },
```
ronaktruss, line 23: We may want to change this to an empty args. The way that this is currently written indicates that we are adding an initialValues to the `ProGearForm`. The  `ProGearForm` only takes in the values: proGear, setNumber, onSubmit, onBack. So even if we were to change this initialValues to false or 'false' it won't actually change the output. 

In the future when we want a storybook version for an edit flow or want an example of the radio button selected to My spouse you can do something like:
```
  args: {
    proGear: {
      selfProGear: false,
    },
  },
```

ronaktruss (approved):
LGTM, outside of the minor comments!

cypress/integration/milmove/ppms/progear.js
```diff
@@ -0,0 +1,33 @@
...
+  viewportType.forEach(({ viewport, isMobile, userId }) => {
+    it(`page loads - ${viewport}`, () => {
+      if (isMobile) {
+        setMobileViewport();
+      }
+
+      signInAndNavigateToWeightTicketPage(userId);
+      navigateFromCloseoutReviewPageToProGearPage();
+      cy.get('#ownerOfProGearSelf').should('be.checked');
+      cy.get('#ownerOfProGearSpouse').should('not.be.checked');
```
duncan-truss, line 27: I would say we have a slight preference for `cy.get('[name="selfProGear"][value="true"]')` or adding a `data-testid` attribute and targeting that in Cypress over id selectors.  I think Ronak made a similar comment elsewhere.

There is an ADR that has us preferring to use `data-testid` across jest tests and Cypress, which is why you won't find `data-cy` used https://transcom.github.io/mymove-docs/docs/adrs/use-data-testid

cypress/integration/milmove/ppms/progear.js
```diff
@@ -0,0 +1,33 @@
...
+      cy.get('label[for="ownerOfProGearSpouse"').should('exist').click();
```
duncan-truss, line 28: You may not need the `should('exist')` part, I think you would get a failure if the get() assert failed.

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
```diff
@@ -0,0 +1,88 @@
...
+import { FormGroup } from '@material-ui/core';
+
+import formStyles from 'styles/form.module.scss';
+import ppmStyles from 'components/Customer/PPM/PPM.module.scss';
+import SectionWrapper from 'components/Customer/SectionWrapper';
+import Fieldset from 'shared/Fieldset';
+import { ProGearTicketShape } from 'types/shipment';
+
+const validationSchema = Yup.object().shape({
+  selfProGear: Yup.string().required('Required'),
```
duncan-truss, line 15: You could make this a boolean instead of a string so that you don't need to doing any casting in the submit handler.

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
```diff
@@ -0,0 +1,88 @@
...
+  setNumber: number,
+  onBack: func.isRequired,
+  onSubmit: func.isRequired,
+  proGear: ProGearTicketShape,
+};
+
+ProGearForm.defaultProps = {
+  setNumber: 1,
+  proGear: {
+    selfProGear: true,
```
duncan-truss, line 84: It looks like @scarequotes approved this, just making note that the designs don't have a default selection so seems like the entire form will be expanded by default and not wait for a selection?

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+  });
+  describe('attaches button handler callbacks', () => {
+    it('calls the onSubmit callback', async () => {
+      render(<ProGearForm {...defaultProps} />);
+
+      await userEvent.click(screen.getByRole('button', { name: 'Save & Continue' }));
+
+      await waitFor(() => {
+        expect(defaultProps.onSubmit).toHaveBeenCalled();
+      }, expect.anything());
```
duncan-truss, line 55: I think you started to test the params onSubmit is called with, this would need changed to `toHaveBeenCalledWith()`

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
scottmries: It's a good point. Since the designs with pro-gear weight tickets have fields that depend on some selection, it needs to at least be able to not have a selection. I'll update this now.

src/pages/MyMove/PPM/Closeout/ProGear/ProGear.jsx
```diff
@@ -0,0 +1,43 @@
...
+import ProGearForm from 'components/Customer/PPM/Closeout/ProGearForm/ProGearForm';
+
+const handleSubmit = () => {};
+
+const ProGear = () => {
+  const history = useHistory();
+  const handleBack = () => {
+    history.push(generalRoutes.HOME_PATH);
+  };
+  const initialValues = { selfProGear: true };
```
duncan-truss, line 20: I think you removed this prop from the form component so shouldn't need this anymore if the defaultProps are handling proGear defaults.

PR #9114, opened 2022-08-25
MB-13499: Update config so that cypress tests use the mockDTOD service

ronaktruss:
## Summary

This adds the mockDTOD service for cypress tests in both local and CI related pipelines.

The usage of this mock service are incorporated in #9105.


scripts/run-e2e-test
```diff
@@ -6,7 +6,7 @@ set -eu -o pipefail
 trap "kill %1" SIGINT
 
 CYPRESS_PORT=4000
-DB_ENV=test DB_PORT="$DB_PORT_TEST" LOGIN_GOV_CALLBACK_PORT="$CYPRESS_PORT" NO_TLS_ENABLED=1 NO_TLS_PORT="$CYPRESS_PORT" \
+DB_ENV=test DB_PORT="$DB_PORT_TEST" LOGIN_GOV_CALLBACK_PORT="$CYPRESS_PORT" NO_TLS_ENABLED=1 NO_TLS_PORT="$CYPRESS_PORT" DTOD_USE_MOCK=1 \
```
ronaktruss, line 9: Is there a preference for using `true` over `1`?

ahobson: I don't think so. If you chase down how the package we use (viper) converts string values to booleans, you come across that it uses

https://pkg.go.dev/strconv#ParseBool

> It accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False

ronaktruss: Yeah I saw that as well. Was mainly asking if there was a set standard / if want to be consistent with the env file.

duncan-truss (approved):
I tested this locally by removing the Cypress fixture for the office/ppms/addShipment.js create shipment intercept and running `make e2e_test` and it seems to have called the mock DTOD service where before it would have failed.

One thing you could do is add `DTOD_USE_MOCK=0` to `.envrc` so people know it's an available flag to override if they choose and that it is enabled for other environments.

duncan-truss (approved):

PR #9118, opened 2022-08-26
MB-13351 Update ppm shipments and signed certification tables

pearl-truss:
## Summary
Migrations in this PR:

- The database can record the customer’s W-2 address
- The address has a relationship with the PPM shipment
- The database can record the calculated final incentive
- Signed certifications table has relationship with the PPM Shipment

## Setup to Run Your Code

This is a task ticket with nothing visual to review except the code itself


pkg/models/signed_certification.go
```diff
@@ -42,6 +42,7 @@ type SignedCertification struct {
 	SubmittingUserID         uuid.UUID                `json:"submitting_user_id" db:"submitting_user_id"`
 	MoveID                   uuid.UUID                `json:"move_id" db:"move_id"`
 	PersonallyProcuredMoveID *uuid.UUID               `json:"personally_procured_move_id" db:"personally_procured_move_id"`
+	PpmID                    *uuid.UUID               `json:"ppm_id" db:"ppm_id"`
```
pearl-truss, line 45: Is the casing here correct?

pkg/models/signed_certification.go
```diff
@@ -81,6 +82,8 @@ func (s *SignedCertification) ValidateUpdate(tx *pop.Connection) (*validate.Erro
 	return validate.NewErrors(), nil
 }
 
+// DEPRECATED - This can be removed when the PPM Shipment Summary Worksheet is updated
+// to use the new PPM shipment table
 // FetchSignedCertificationsPPMPayment Fetches and Validates a PPM Payment Signature
 func FetchSignedCertificationsPPMPayment(db *pop.Connection, session *auth.Session, id uuid.UUID) (*SignedCertification, error) {
```
pearl-truss, line 88: I added this comment as a placeholder. I believe to be able to reuse this model for PPMs this function will need to be updated to pass in a ppm shipment id

pkg/models/signed_certification.go
rogeruiz: from the looks of the other files, I believe it could be either `PPMID` but that certainly doesn't read as easily as `PpmID`.

Looking at `payloads/model_to_payload.go` there's references to `PPMType` and `PpmType`. So I'd say for readability, this is the way to go 👍🏽

pkg/models/signed_certification.go
rogeruiz: Do you think we should have a Jira ticket around this to remove it? I like the idea of having the comment, but I'm wondering it there's a better way to bubble up the visibility on this. The comment's good enough though 💯

rogeruiz:
I'm going to set this up so I can see the Swagger documentation here to test something. but this looks good so far @pearl-truss. I may be slow on my review but if others get to it first that's fine too 😃

swagger-def/internal.yaml
```diff
@@ -2599,6 +2609,14 @@ definitions:
...
         pattern: ^(\d{5})$
         x-nullable: true
+      w2Address:
+        x-nullable: true
+        $ref: 'definitions/Address.yaml'
+      finalIncentive:
+        type: integer
+        format: cents
+        x-nullable: true
+        x-omitempty: false
```
rogeruiz, line 2619:
```suggestion
      finalIncentive:
        description: >
          The final calculated incentive for the PPM shipment. This does not include **SIT** as it is a reimbursement.
        type: integer
        format: cents
        x-nullable: true
        x-omitempty: false
```

swagger-def/ghc.yaml
```diff
@@ -3753,6 +3753,14 @@ definitions:
...
         pattern: ^(\d{5})$
         x-nullable: true
+      w2Address:
+        x-nullable: true
+        $ref: 'definitions/Address.yaml'
+      finalIncentive:
+        type: integer
+        format: cents
+        x-nullable: true
+        x-omitempty: false
```
rogeruiz, line 3763:
```suggestion
      finalIncentive:
        description: >
          The final calculated incentive for the PPM shipment. This does not include **SIT** as it is a reimbursement.
        type: integer
        format: cents
        x-nullable: true
        x-omitempty: false
```

swagger-def/definitions/PPMShipment.yaml
```diff
@@ -130,6 +133,11 @@ properties:
     format: cents
     x-nullable: true
     x-omitempty: false
+  finalIncentive:
+    type: integer
+    format: cents
+    x-nullable: true
+    x-omitempty: false
```
rogeruiz, line 140:
```suggestion
      finalIncentive:
        description: >
          The final calculated incentive for the PPM shipment. This does not include **SIT** as it is a reimbursement.
        type: integer
        format: cents
        x-nullable: true
        x-omitempty: false
```

swagger-def/internal.yaml
```diff
@@ -1600,6 +1605,11 @@ definitions:
         type: string
         format: uuid
         x-nullable: true
+      ppm_id:
+        type: string
+        format: uuid
+        example: c56a4180-65aa-42ec-a945-5fd21dec0538
+        x-nullable: true
```
rogeruiz, line 1612: Similar to Final Incentive, I've borrowed this from our database comments on the table.

```suggestion
      ppm_id:
        description: >
          The PPM Shipment ID to associate with the assigned certificate.
        type: string
        format: uuid
        example: c56a4180-65aa-42ec-a945-5fd21dec0538
        x-nullable: true
```

swagger-def/internal.yaml
```diff
@@ -1574,6 +1574,11 @@ definitions:
...
```
rogeruiz, line 1581:
```suggestion
      ppm_id:
        description: >
          The PPM Shipment ID to associate with the assigned certificate.
        type: string
        format: uuid
        example: c56a4180-65aa-42ec-a945-5fd21dec0538
        x-nullable: true
```

rogeruiz (changes requested):
❌ This PR needs changes around descriptions for `finalIncentives` and `ppm_id` defined in a couple of places. Great work on this @pearl-truss. I don't see any technical issues here. From the documentation perspective, I do see some things that would really be great to add here even though our Swagger documentation does not currently meet the expectations I'm trying to shift our API docs to. 

> 📣 **Please feel free to dismiss this review**. I'm adding it as request changes so aid discussion on the topic, but I'm not trying to dictate here.

Can we get a description section here that describes this a bit better? I'm not familiar with this outcome at all, so I'm not sure what **Final Incentive** is here and _I personally think_ it'd be good for the project to have better descriptive text for new definitions in our API. I pulled what **Final Incentive** is from our Slack history in DP3 and from our Migration files that added the column.

Also, I'd like to suggest adding a new `final_incentive.yaml` file in our `swagger-def/definitions/` directory to centralize where this definition is being defined since I'm suggestion adding a `description: >` stanza to the Yaml and my current flow has me suggesting it in three different places across `ghc.yaml` and `internal.yaml`.

I've also added some suggestions to `ppm_id` definitions and since there's two in the `internal.yaml` file, I think it's a good idea to centralize this inside of `swagger-def/definitions/` as well.

rogeruiz:
Also, you'll notice in my suggestions in my review that I'm using Markdown in the `description: >` stanza and I tested this locally with our Docusaurus repository to ensure that Markdown does get processed correctly in our documentation.

Captura de Pantalla 2022-08-26 a la(s) 1 41 46 p m

pearl-truss:
@rogeruiz thanks for the review! I believe all requested changes have been made

rogeruiz (approved):
lgtm 🌈 

Great work on this @pearl-truss!

rogeruiz:
Looks like the `pkg/gen` code I mentioned above wasn't committed to the branch. 👀

pearl-truss:
> Looks like the `pkg/gen` code I mentioned above wasn't committed to the branch.
> 
> 👀 https://app.circleci.com/pipelines/github/transcom/mymove/44762/workflows/b490be6a-f23f-4e4f-a2f8-38fd5cd47a17/jobs/708646/parallel-runs/0/steps/0-106
> 
> I believe the fix is to run
> 
> ```
> make swagger_generate server_generate
> ```



> Looks like the `pkg/gen` code I mentioned above wasn't committed to the branch.
> 
> 👀 https://app.circleci.com/pipelines/github/transcom/mymove/44762/workflows/b490be6a-f23f-4e4f-a2f8-38fd5cd47a17/jobs/708646/parallel-runs/0/steps/0-106
> 
> I believe the fix is to run
> 
> ```
> make swagger_generate server_generate
> ```

👍 I will push a fix for that as soon as I resolve swagger conversion issues in the payload_to_model and model_to_payload files

migrations/app/schema/20220824133409_add_w2_address_final_incentive_to_ppm_shipment.up.sql
```diff
@@ -0,0 +1,8 @@
+ALTER table ppm_shipments
+    ADD COLUMN w2_address_id uuid CONSTRAINT ppm_shipments_address_id_fkey REFERENCES addresses (id),
+    ADD COLUMN final_incentive int;
+
+CREATE INDEX on ppm_shipments (w2_address_id);
+
+COMMENT ON COLUMN ppm_shipments.w2_address_id IS 'Customer address to receive their W2 tax form';
```
duncan-truss, line 7: When I read this it made it seem like we were managing where they would get their W2 sent.  In the Zeplin mock they ask "What is the address on your W2?" https://app.zeplin.io/project/625896e702018f8ff82451c0/screen/625897c6c3a6839268f33d89

pkg/models/ppm_shipment.go
```diff
@@ -98,6 +100,7 @@ type PPMShipment struct {
 	ProGearWeight                  *unit.Pound          `json:"pro_gear_weight" db:"pro_gear_weight"`
 	SpouseProGearWeight            *unit.Pound          `json:"spouse_pro_gear_weight" db:"spouse_pro_gear_weight"`
 	EstimatedIncentive             *unit.Cents          `json:"estimated_incentive" db:"estimated_incentive"`
+	FinalIncentive                 *unit.Cents          `json:"final_incentive" db:"final_incentive"`
 	HasRequestedAdvance            *bool                `json:"has_requested_advance" db:"has_requested_advance"`
 	AdvanceAmountRequested         *unit.Cents          `json:"advance_amount_requested" db:"advance_amount_requested"`
 	HasReceivedAdvance             *bool                `json:"has_received_advance" db:"has_received_advance"`
```
duncan-truss, line 106: You could also add a placeholder association here for the signed certifications to make sure it's wired up:

```go
SignedCertifications SignedCertifications`has_many:"signed_certifications" fk_id:"ppm_id" order_by:"created_at desc"`
```

swagger-def/definitions/FinalIncentive.yaml
```diff
@@ -0,0 +1,6 @@
+description: >
+    The final calculated incentive for the PPM shipment. This does not include **SIT** as it is a reimbursement.
+type: integer
+format: cents
+x-nullable: true
+x-omitempty: false
```
duncan-truss, line 6: I think this will also be a read-only property set by the system not modified by any API.

swagger-def/ghc.yaml
```diff
@@ -3753,6 +3753,11 @@ definitions:
         example: '90210'
         pattern: ^(\d{5})$
         x-nullable: true
+      w2Address:
+        x-nullable: true
+        $ref: 'definitions/Address.yaml'
+      finalIncentive:
```
duncan-truss, line 3759: I am pretty sure this won't get set by the office user but will just be the value of running the PPM incentive pricer on the actual values during closeout instead of the estimated/expected values during the initial incentive.

migrations/app/schema/20220824133409_add_w2_address_final_incentive_to_ppm_shipment.up.sql
pearl-truss: I switched to say, "Address for where a customer received their W2 tax form"

duncan-truss (approved):

migrations/app/schema/20220826142534_associate_signed_certifications_with_ppms.up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER table signed_certifications
+    ADD COLUMN ppm_id uuid CONSTRAINT ppm_id_fkey REFERENCES ppm_shipments (id);
```
felipe-lee, line 2: **question**: I looked at the DDL for this table and noticed that signature is required but we aren't asking for signature on the page. Are we going to re-use their signature from when they submitted the move? Or modify this table to allow signature to be nullable?

```abbreviated signed_certifications DDL
-- auto-generated definition
create table signed_certifications
(
    id                          uuid      not null
        primary key,
    submitting_user_id          uuid      not null
        references users,
    move_id                     uuid      not null
        constraint signed_certifications_moves_id_fk
            references moves,
    certification_text          text      not null,
    signature                   text      not null,
    date                        timestamp not null,
    created_at                  timestamp not null,
    updated_at                  timestamp not null,
    personally_procured_move_id uuid
        references personally_procured_moves,
    certification_type          text,
    ppm_id                      uuid
        constraint ppm_id_fkey
            references ppm_shipments
);
```

pearl-truss: I think using the signature from the move makes the most sense since we probably don't want to allow nullable signatures for this table across the board. I made a note of this in https://dp3.atlassian.net/browse/MB-13354

pearl-truss: @felipe-lee @duncan-truss do you have any thoughts on this?

duncan-truss: That's a good question Felipe, I think we could weigh our options in the follow up PRs.  We could also insert something like "N/A" honestly.

I'm thinking more about the certification_text and how we may need to re-prompt if it has changed.  I think for move submission they have the frontend send the whole block of certification text to the API kind of naively.

felipe-lee: Yeah I'd be fine with those options. I think we'll want to verify with Danny, but agree we can do that in future tickets. Thanks for noting it in that ticket Pearl!

PR #9142, opened 2022-09-01
MB-12189 set up fe fake data generator and factory tools

felipe-lee:
## Summary

This PR sets up our FE to be able to have factories to more easily generate test data. 

Document with some findings and thoughts: 🔒 https://dp3.atlassian.net/wiki/spaces/MT/pages/1895071898/Test+factories+and+fake+data+generators

<details>
<summary>Main changes</summary>

* [Added faker and test-data-bot to js dev deps](https://github.com/transcom/mymove/pull/9142/commits/d87606a63098d70b3e64afa964b51ef96b6c2ee3)
* [Updated eslintrc w/changes needed for factories](https://github.com/transcom/mymove/blob/felipe-lee/MB-12189-set-up-fe-fake-data-generator-and-factory-tools/.eslintrc.js)
    * This is what triggered the IS3 review.
    * Set up stricter checks for extraneous deps so that we won't just allow any extraneous deps, but still allow dev dependencies where needed.
    * Also allow param-reassignment so that factories can more easily work with objects in the `postBuild` step.
* [Add faker setup for FE factories](https://github.com/transcom/mymove/pull/9142/commits/c2590b4860224492906959d2068a0f53f737a8dd)
    * This sets up faker to use `en_US` as the locale for generated data
    * Also sets up the `fake` function that will be used by factories to generate fake data.
* [Add factory for addresses](https://github.com/transcom/mymove/blob/felipe-lee/MB-12189-set-up-fe-fake-data-generator-and-factory-tools/src/utils/test/factories/address.js)
    * This serves as a simple factory to showcase how factories work, and to make it easier to now build out other factories that will need addresses, e.g. service members, shipments, etc.
    * It includes setting field attributes per build, `postBuild` function, `traits`, and `traits` with `postBuild`.
*  [Update a set of tests to use addressFactory to showcase using factories](https://github.com/transcom/mymove/pull/9142/commits/ca6f33625e1d0e7ce2eb0292ae6b77263d12480a)

</details>

<details>
<summary>Other changes</summary>

* [Add @storybook/addon-actions as an explicit dev dependency](https://github.com/transcom/mymove/pull/9142/commits/240d91b23c21b11b0800b2667a41aceb0a94daab)
    * This was needed because of the `eslint` config changes I made. We already have the package because of other dependencies, but since we import it in some files, it's better to list it as an explicit dependency.

</details>

## Testing the changes

<details>
<summary>💻</summary>

### Running Tests

```sh
yarn test src/pages/MyMove/Profile/ResidentialAddress.test.jsx
```

</details>


.eslintrc.js
```diff
@@ -69,7 +83,6 @@ module.exports = {
     {
       files: ['*.stories.js', '*.stories.jsx', 'setupTests.js'],
       rules: {
-        'import/no-extraneous-dependencies': 'off',
```
felipe-lee, line 72: I pulled this out of here and put it above so that we only allow dev dependencies but still check for other extraneous dependencies.

src/utils/test/factories/base.js
```diff
@@ -0,0 +1,10 @@
+import { faker } from '@faker-js/faker';
+import { perBuild } from '@jackfranklin/test-data-bot';
+
+faker.setLocale('en_US');
+
+const fake = (callback) => {
+  return perBuild(() => callback(faker));
+};
```
felipe-lee, line 8: This is partially syntactic sugar. It makes it so than in a factory, instead of having to write:

```js
...
fields: {
    streetAddress1: perBuild(() => faker.address.streetAddress())
}
...
```

you can instead write: 

```js
...
fields: {
    streetAddress1: fake((f) => f.address.streetAddress()),
}
...
```

which, tbh, imo isn't a huge win for this alone.

But, one other advantage is that it lets us set the locale once and then everything that uses `fake` will automatically get it. In version 3.0.0 of `faker`, [there will be per localization packages](https://fakerjs.dev/api/localization.html#individual-localization-packages), which will make this a little less necessary since we could just make sure we always import from the same package everywhere. Though I still slightly feel it might be beneficial to have it here so that we don't have to be ensuring we import from the same place and just leave that here.

src/utils/test/factories/address.js
```diff
@@ -0,0 +1,51 @@
+import { faker } from '@faker-js/faker';
+import { build } from '@jackfranklin/test-data-bot';
+
+import fake from 'utils/test/factories/base';
+
+export const BLANK_ADDRESS = 'blank';
+export const ADDRESS_WITHOUT_COUNTRY = 'omitCountry';
+
+const addressFactory = build({
```
pearl-truss, line 9: **Question:** Will engineers still be able to add specific values for certain fields? For instance with pricing wanting to choose zipcodes that are a certain distance from each other

src/utils/test/factories/base.js
pearl-truss: That seems worth it

pearl-truss:
**Request**: can a link to this PR be added to the [ADR](https://transcom.github.io/mymove-docs/docs/adrs/use-fake-data-factories-with-fake-data-generators-to-create-test-data#resources) for this concept?

src/utils/test/factories/address.js
```diff
@@ -0,0 +1,51 @@
...
+      address.postalCode = address.postalCode.padStart(5, '0');
+    } else {
+      address.postalCode = '';
+    }
+
+    return address;
+  },
+});
+
+export default addressFactory;
```
ronaktruss, line 51: What's the rationale behind having this be a default import over a named import?

src/utils/test/factories/address.js
ronaktruss: Yeah I think we should be able to do that. If you want to just pass some override values you can do that by doing something like:
```js
const fakeAddress = addressFactory({ overrides: { streetAddress1: '1600 Washington Ave' } });
```

For something like the postalCode that is set as part of the `postBuild`, I don't think there is a way to pass in a value and set the value. Though we can just create a value using the factory and then do a manual override by doing something like:
```js
const fakeAddress = addressFactory({ overrides: { streetAddress1: '1600 Washington Ave' } });
fakeAddress.postalCode = '90210';
```

Ryan-Koch (approved):
Giving a 👍 for the eslintrc change

src/utils/test/factories/address.js
scottmries: Do we think it's worth turning the override keys into constants? I'm working on the ServiceWorker factory, and the keys down in the `weight_allotment` can be tricky to get correct in an override. Constantizing them would at least make them available to Intellisense.

src/utils/test/factories/address.js
felipe-lee: Ah, that was more because I started with only one export and our eslint config requires it to be default. Then I added more things, but by then I wasn't sure if it was being used in other branches so I didn't want to cause merge issues. I'm happy to change it though and just communicate out on the channel in case anyone has a branch and is using this.

felipe-lee:
> **Request**: can a link to this PR be added to the [ADR](https://transcom.github.io/mymove-docs/docs/adrs/use-fake-data-factories-with-fake-data-generators-to-create-test-data#resources) for this concept?

Yeah, great idea! I can open up a PR for the ADR.

src/utils/test/factories/address.js
felipe-lee: Scott and I met earlier and I learned from him that in order for an override to work, it needs to be defined as one of the base fields. So if I add `postalCode` to the field definitions, we can pass in an override. The `postBuild` can still overwrite it, but I can add a conditional to leave it as is if it's already been set. I can push up a commit with those changes.

felipe-lee: We also talked about having constants for the fields. I agree it would be good to do that because intellisense is nice and the point of these factories is to speed up test writing. We talked about a few options for how that could work. I'm going to push up one version, but plan on talking more about it during our PR review time on Tuesday, and bringing things up at the BE/FE checkin.

felipe-lee:
> > **Request**: can a link to this PR be added to the [ADR](https://transcom.github.io/mymove-docs/docs/adrs/use-fake-data-factories-with-fake-data-generators-to-create-test-data#resources) for this concept?
> 
> Yeah, great idea! I can open up a PR for the ADR.

@pearl-truss here's the other PR: https://github.com/transcom/mymove-docs/pull/205

src/utils/test/factories/address.js
```diff
@@ -0,0 +1,58 @@
...
+export const ADDRESS_WITHOUT_COUNTRY = 'omitCountry';
+
+export const ADDRESS_FIELDS = {
+  STREET_ADDRESS_1: 'streetAddress1',
+  STREET_ADDRESS_2: 'streetAddress2',
+  CITY: 'city',
+  STATE: 'state',
+  POSTAL_CODE: 'postalCode',
+  COUNTRY: 'country',
+};
```
felipe-lee, line 16: Note that this is just one potential way we could set up these constants. This decreases how easy it is for intellisense to know what field you want without you first using the parent name, but makes it so things are categorized.

An alternative would be to have them as standalone constants like the trait ones above.

Scott also suggested we could maybe do something like `ADDRESS__STREET_ADDRESS_1` to still make them easy to grab via intellisense, whlie having a part define what category they belong to.

I plan to bring this up for discussion on Tuesday.

src/utils/test/factories/base.js
```diff
@@ -0,0 +1,49 @@
...
+import { perBuild } from '@jackfranklin/test-data-bot';
+import yaml from 'js-yaml';
+
+faker.setLocale('en_US');
+
+const fake = (callback) => {
+  return perBuild(() => callback(faker));
+};
+
+const loadSpec = (fileName) => {
```
felipe-lee, line 15: Got the idea for this from https://github.com/swagger-api/swagger-js/issues/1410

Came up while I was looking for a way to use the swagger client with local files. I didn't want to point to a URL since the swagger files could change with a PR and then we'd need to update the path for those. An alternative is linked to in that thread to spin up a local server just to serve the file, but that seems like more work than is needed. Especially because once the yaml is parsed and converted to json, we don't really need the swagger client, we can access all of our definitions through the json this outputs.

felipe-lee: So I tried the codegen tool a bit running:

```shell
docker run --rm -v ${PWD}:/local swaggerapi/swagger-codegen-cli generate \
    -i local/swagger/internal.yaml \
    -l javascript \
    -o /local/src/utils/test/swagger-client/internal \
    -DmodelDocs=false -DmodelTests=false -DapiDocs=false -DapiTests=false
```

It then needs to be installed with:

```shell
yarn add --dev file:./src/utils/test/swagger-client/internal
```

This seems to give us access to the definitions. One nice thing about this is that it gives us code completion since there are actual variables for properties like the state enum, rather than just working with json. I haven’t gotten a chance to mess with it too much though.

src/utils/test/factories/address.js
```diff
@@ -0,0 +1,63 @@
...
+    [ADDRESS_FIELDS.CITY]: fake((f) => f.address.city()),
+    [ADDRESS_FIELDS.STATE]: perBuild(() => {
+      const spec = getInternalSpec();
+      return oneOf(...spec.definitions.Address.properties.state.enum).call();
+    }),
+    [ADDRESS_FIELDS.POSTAL_CODE]: '',
+    [ADDRESS_FIELDS.COUNTRY]: 'US', // Likely change once we support more than just CONUS moves.
+  },
+  traits: {
+    [ADDRESS_TRAITS.BLANK]: {
```
duncan-truss, line 34: This isn't really a factory question, is there a reason we would have an address with blank fields?  I see that streetAddress2 gets saved as a blank string in the customer app.  Or is there some null/undefined behavior where this is useful?

It looks like we have to define all of the fields on the model and then make a trait if we ever want to remove fields down to say what is only required?

src/utils/test/factories/address.js
```diff
@@ -0,0 +1,63 @@
...
+  STREET_ADDRESS_1: 'streetAddress1',
+  STREET_ADDRESS_2: 'streetAddress2',
+  CITY: 'city',
+  STATE: 'state',
+  POSTAL_CODE: 'postalCode',
+  COUNTRY: 'country',
+};
+
+export const addressFactory = build({
+  fields: {
```
duncan-truss, line 21: Should `id` be included as an option here to represent an already created record?

src/utils/test/factories/address.js
```diff
@@ -0,0 +1,63 @@
...
+        [ADDRESS_FIELDS.COUNTRY]: '',
+      },
+    },
+    [ADDRESS_TRAITS.NO_COUNTRY]: {
+      postBuild: (address) => {
+        delete address.country;
+
+        return address;
+      },
+    },
```
duncan-truss, line 50: It could be a good pattern to start where there is a 'required' trait that doesn't automatically fill in all of the optional fields.  This may not be perfect across forms but usually we are worried about validations in our tests.

duncan-truss (approved):
Seems like a good start to me.  I am curious how it will work with nested objects but this was a good first demonstration case.

src/utils/test/factories/address.js
felipe-lee: Ah, yeah I need to add that one and `eTag`. I can push that up today.

src/utils/test/factories/address.js
felipe-lee: Good point, that would probably be a good thing to have, I can add that.

src/utils/test/factories/address.js
felipe-lee: Ah, that came up because when I initially started using this in the residential address tests, there had been an empty address. Though after looking at the internal yaml, it looked like the address as a whole was nullable, and could be omitted when empty, so I ended up not using this. I should probably just remove the trait since there isn't a current use case for it. 

To your point about having to define all the fields and then removing them, that is what needs to be done because if you don't define a field, you can't override it at build-time, which is a bit annoying. Thanks for reminding me, I'll add that to the [things that would make it easier to work with factories](https://dp3.atlassian.net/wiki/spaces/MT/pages/1895071898/Test+factories+and+fake+data+generators#Features-That-Would-Make-It-Easier-To-Work-on-Factories)

src/utils/test/factories/base.js
```diff
@@ -0,0 +1,49 @@
...
+  const yamlPath = path.join(process.cwd(), `./swagger/${fileName}`);
+
+  // RA Summary: eslint - security/detect-non-literal-fs-filename - OWASP Path Traversal
```
felipe-lee, line 18: https://transcom.github.io/mymove-docs/docs/backend/guides/guide-to-static-analysis-security-workflow/

NamibiaTorres (commented):
Nice job so far! I was just learning about factories in React so it was really interesting seeing how they are implemented!

felipe-lee:
Closing in favor of #9233

PR #9149, opened 2022-09-02
Update demo certs

chtakahashi:
- Updating the demo certs to the most recent out of SSLMate
- Made the naming for the demo certificates consistent with the loadtest and exp environments
-


rogeruiz:
@mergifyio rebase

rogeruiz:
Trying to rebase here since the added test coverage could be in the default branch? Trying to reduce variables here.

rogeruiz:
Looks like Mergify isn't using the MyMove Robot and we should look into that. But not now. Just noting it again.

ahobson:
> To be clear here, @$ahobson **did not interact with this PR at all**. This was Mergify acting on his behalf.

I am EVERYWHERE 😀

ahobson (approved):

duncan-truss (changes requested):
When I tried to download the migrations from all environments, I got a 404 for loadtest.  I think we need a matching migration in each environment even if it's a no-op right?

Has this PR been deployed to demo to test that the Postman user can authenticate?  I see that the secure migration inserts a new record but I feel like that would result in 2 client_cert records for the same api.demo.dp3.us subject.  I didn't see a user_id value specified either, which I think may be an issue.

duncan-truss:
I think I'm wrong about needing migrations in the other environments, it will fallback to the base secure migration I think in the case there isn't one first found in s3.

duncan-truss:
It looks like we want to use the update version of the `gen-certs-migration` command and pass in the existing id `2fbfb740-cc36-4eeb-a6c6-1e800c74b348` from the last prior migration.

https://github.com/transcom/mymove/blob/master/cmd/milmove/gen_certs_migration.go#L101-L105
https://github.com/transcom/mymove/blob/master/migrations/app/secure/20210909171753_add_demo_tls_cert.up.sql

chtakahashi:
paired with @duncan-truss to try to get this working. Needed to run gen certs-migration with the --update flag since its an existing entry, then upload

duncan-truss (approved):
Tested the local migration and re-downloaded the one from the demo s3 bucket and tested it ran locally.  Thanks for working through this @chtakahashi!

PR #9154, opened 2022-09-02
Update references to api.demo.dp3.us.chain.der.p7b after rename

ahobson:
## Summary

The path to the demo cert chain was renamed and we accidentally missed a few spots that used the old name


chtakahashi (approved):
Looks good! just was about ot push these exact changes. :shipit:

PR #9162, opened 2022-09-06
MB-13757: Adds .skip to flaky integration tests

kylehilltruss:
## Summary

This pull request temporarily skips a number of known flaky integration tests (and any other tests in the same test run that rely on them). As the rate of integration test failures in the `master` branch has been increasing lately, these failures have presented a blocker to developers merging their own work and/or dependency updates, so temporarily disabling them while we search for a solution to any underlying race conditions is warranted.

There are no visual or behavioral changes in this pull request. It can be tested by running the office integration tests locally (in Docker) or just by observing that they pass on CircleCI's run.


jenniferlynparsons (approved):
LGTM and glad we're able to remove this blocker for everyone.

PR #9192, opened 2022-09-13
MB-12577 Refactor SQL query

kctruss:
## Summary

Separate move history SQL query into its own file.
In the go code, call a new function
which fetches the query string.
This pattern allows future queries to be organized in a separate directory
and opens up new opportunities for linting and formatting improvements.

[ADR for SQL templating](https://github.com/transcom/mymove-docs/pull/208)

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the office app
2. Login as an office user
3. Navigate to move history


rogeruiz:
Taking a look at this now @kctruss should have a review done in the next 20 minutes.

pkg/services/query/sql_scripts/move_history_fetcher.sql
```diff
@@ -0,0 +1,387 @@
+-- name: move_history_fetcher
```
rogeruiz, line 1: all the lines from the comment above have moved here.

pkg/services/query/query_fetcher.go
```diff
@@ -0,0 +1,34 @@
...
+	"os"
+	"path"
+
+	"github.com/qustavo/dotsql"
+)
+
+// GetQueryStringFromFile accepts a query name,
+// returns SQL query as a string
+func GetQueryString(queryName string) (string, error) {
+	sqlDir := "services/query/sql_scripts"
```
rogeruiz, line 14: This is a new folder that is generated with this work. So I think in another PR, we should have a `README.md` file that simply points to the ADR in the MilMove Documentation Portal.

rogeruiz, line 14: This needs to be defined as a constant in the query package rather than in the function that uses it since it's not going to change.

```suggestion
const sqlDir string = "services/query/sql_scripts"

// GetQueryStringFromFile accepts a query name,
// returns SQL query as a string
func GetQueryString(queryName string) (string, error) {

```

rogeruiz (changes requested):
❌ This PR needs changes because of the constant being defined in the wrong place.

But otherwise, this is great work @kctruss. I will add a PR for adding a readme to `pkg/services/query/sql_scripts` that points to your ADR around this in cause folks stumble on to the folder casually browsing the repository.

rogeruiz (approved):
lgtm 🌈

pkg/services/query/query_fetcher.go
kctruss: Good call, I've made this change!

PR #9200, opened 2022-09-15
Make the psql-wrapper script more user friendly

rogeruiz:
## Summary

My idea here is to check for where the `psql` binary is installed and fail with
a better message of what's happening so folks know what to do. Also, I want
to make sure it doesn't negatively effect Nix users.


scripts/psql-wrapper
```diff
@@ -7,14 +7,22 @@ db_port="${DB_PORT:-5432}"
 
 command="${*:-}"
 
+# Check if Homebrew is installed and fail with an error message if it isn't for
+# some reason.
+{ type -p brew > /dev/null; } 2>&1 || \
+    echo "You must have Homebrew installed to run this script."
```
ahobson, line 13: Do we need `brew`? As a nix user, I don't have brew installed. Could/should this check go into the if/else with `NIX_PROFILE` below?

rogeruiz: ooh good point. Thank you @ahobson I can relax this check then.

ahobson (approved):
This looks right to me. Thank you

PR #8327, opened 2022-03-23
[MB-11991] Upgrade to go 1.18

reggieriser:
## Summary

This PR updates the project from Go 1.17.7 to Go 1.18.

Go 1.18.x version history:
https://golang.org/doc/devel/release.html#go1.18

Here's the [corresponding PR from circleci-docker](https://github.com/transcom/circleci-docker/pull/227) for the docker image hash referenced in this PR.  Please verify that it seems correct and that all relevant references have been updated.

After this PR has been initially vetted, I will merge in the circleci-docker PR and update this PR with the new hash.

Note that we had a couple of [new deprecations in 1.18](https://go.dev/doc/go1.18) that had to be dealt with:

> The [Title](https://go.dev/pkg/strings/#Title) function is now deprecated. It doesn't handle Unicode punctuation and language-specific capitalization rules, and is superseded by the [golang.org/x/text/cases](https://golang.org/x/text/cases) package.

For the above, I was able to adjust the code to use the recommended `cases` package instead.

> [CertPool.Subjects](https://go.dev/pkg/crypto/x509/#CertPool.Subjects) is deprecated. On Windows, macOS, and iOS the [CertPool](https://go.dev/pkg/crypto/x509/#CertPool) returned by [SystemCertPool](https://go.dev/pkg/crypto/x509/#SystemCertPool) will return a pool which does not include system roots in the slice returned by Subjects, as a static list can't appropriately represent the platform policies and might not be available at all from the platform APIs.

There was no recommended option with the above deprecation, but it only seems to apply to a CertPool that was returned by `SystemCertPool`.  We are not doing that here.  So, it seems like we can ignore the deprecation for now since there's no obvious alternative.  I've added annotations noting that and will need the approval of the IS3 group for that.

I also had to make some adjustments around `golangci-lint`.  From my inline comment in the `.golangci.yml` config file:
```
  # Setting this to 1.17 for now instead of letting golangci-lint auto-detect 1.18 since golangci-lint turns off
  # several linters we use due to 1.18 compatibility issues. Once those are sorted out in later versions of
  # golangci-lint and the corresponding linters, we should be able to remove this line. We may also experience linter
  # issues if we start using generics before then. Follow https://github.com/golangci/golangci-lint/issues/2649 for
  # more info. To see which of our linters are turned off, remove this line and run `pre-commit run -a --verbose
  # golangci-lint`.
  go: '1.17'
```

If you happen to run into issues when interacting with `pre-commit`, then try to clear your cache and install again:
[...]

PR #9201, opened 2022-09-15
[MB-13908] Update golangci-lint config

reggieriser:
## Summary

This PR updates and cleans up our `golangci-lint` config.  Some of this work was deferred from #9064 when we added the `gci` linter.  Specifically, the following changes are being made:

- three linters that are now deprecated are being removed: `deadcode`, `structcheck`, and `varcheck`.  [Per the documentation](https://golangci-lint.run/usage/linters/#disabled-by-default), these linters were abandoned by the author and the `unused` linter should be used as the replacement for all three.
- the list of linters noted in the `enable` section was not the complete list that were being run.  There are several others that were [enabled by default](https://golangci-lint.run/usage/linters/#enabled-by-default), including the `unused` one noted above.  So, I've added `gosimple` and `unused` in explicitly here to avoid any confusion about what linters are actually running.  If you want to see a list of all enabled linters, get a local version of `golangci-lint` and run `golangci-lint linters`.
- the `deadline` setting under `run` has been deprecated and [`timeout` should be used in its place](https://github.com/golangci/golangci-lint/pull/793).
- instead of just skipping `mocks` directories, I skip any directory that ends in `mocks` since we do have a mocks directory called `ghcmocks` that we had to create to get around some circular dependency issues.
- the workaround of setting a `go` version to `1.17` that we introduced when [moving to go 1.18](https://github.com/transcom/mymove/pull/8327) has been removed because the problem linters have either been fixed or deprecated.
- in removing the `1.17` workaround, I realized that [an annotation and linter suppression added/approved in a previous PR](https://github.com/transcom/mymove/pull/8327/files#r835386427) was inadvertently removed in a subsequent PR's refactor.  This PR adds it back.
- the `CODEOWNERS` file was referencing the wrong filename for the golangci-lint config, so I've fixed that.

## Setup to Run Your Code

How to test:
- `make clean server_build client_build build_tools`
- `pre-commit run -a` to make sure there are no linter issues
- `pre-commit run -a --verbose golangci-lint` to see if there are any warnings about linters being turned off (there shouldn't be anymore)
- `golangci-lint linters` to see a list of enabled/disabled linters for our configuration


ahobson (approved):
🚀

Ryan-Koch (approved):

PR #9204, opened 2022-09-15
Let's also prefix pg_dump to use Brew and Nix

rogeruiz:
## Summary

The work done in #9200 was useful for M1 macs, so I figured it should also be
done for our `pg_dump` commands as well.


ahobson (approved):
🚀

michaelvaldes-truss (approved):
lgtm, and works on my machine (m1 and no nix)

PR #9207, opened 2022-09-16
[ MB-12577 ] Adding a simple doc for the new QueryFetcher service

rogeruiz:
## Summary

While pairing with @micahelvaldes-truss on the review for #9192, we thought it
would be a great idea to leave a small document as a bread crumb to inform
engineers of the ADR that added this change. I didn't want to hold up the review
of #9192 so I created this small PR for this change. 

I'm also changing the code owners rule to trigger on changes to this directory
so folks can get the appropriate notification when SQL files are being added
with this new method. 

After @kctruss' [post about this in #g-database][slack-post], if we find a
better place for documentation we can probably remove this file. But for now, I
think this is a great addition to the repository for knowledge sharing.

[slack-post]: https://ustcdp3.slack.com/archives/CSGDM3NUW/p1663351683035619


kctruss (approved):
Love it!

rogeruiz:
Interesting that `@robot-mymove` is getting slower on the updates of the PR comments. Shouldn't be anything to worry about _right now_, but just a small observation if you look at the time stamps in this PR.

PR #9208, opened 2022-09-16
[ MB-12577 ] Refactor QueryFetcher to use another function

rogeruiz:
## Summary

When testing this further, @kctruss and I found some improvements to the
work in #9192. This patch updates things to make it a bit easier to work
with in the future.


kctruss (approved):
We tested this together and it is a much needed improvement. Thank you!

rogeruiz:
This is closed by the work done in #9217

PR #9219, opened 2022-09-20
MB-13654 Save draft of violations on eval reports

PDickeyTruss:
## Summary

This PR adds saving of violations to a report using the 'save draft' button on the 2nd page of the Evaluation Report form. Saving of the other items on this page (4 dates and 'serious incident') will come in a follow on PR.

Engineering-wise this PR adds 2 new endpoints: 1) A `post` to add/update associations between reports/violations and 2) a `get` endpoint to fetch all violations associated with a report.  The fetch endpoint is pretty straightforward. The 'associate' endpoint does a little more and is worth describing. It accepts a POST request to `/report-violations/{reportID}` that includes a body with violation IDs for all violations that should be associated with the report. The unique thing here is that when you add new violation associations to a report the existing violation associations for that report are replaced. This allows for 'unselecting' of existing violation associations. I added some documentation around this in swagger but am more than willing to add more elsewhere if anyone has ideas about where might be a better place. 

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

## Verification Steps
- [x] As a QAE/CSR user access the `Quality assurance reports`, create a new report, and fill out the first page of the report form being sure to select 'yes' for Violations Observed.
- [ ]  From the 2nd page (violations page) of the Eval report form:
    - [x] Violations that are selected are retained when the 'Save Draft' button is clicked
    - [x] Violations that were un-selected that had been previously saved are not retained when the 'Save Draft' button is clicked
    - [x] After saving violations to the report, the 'Save Draft' button redirects to the evaluation reports page and shows the 'Draft was saved successfully' alert. 
    - [x] Using the 'Save Draft' button on the first page of the eval report has no impact on the violations that have been previously saved to the report from the second page.
    - [ ] No errors or console warnings are shown

## Screenshots

https://user-images.githubusercontent.com/62263329/192324509-e4521d16-1fb7-4a92-94db-d4a15a0adcf0.mov


pkg/services/report_violation/report_violation_creator.go
```diff
@@ -0,0 +1,48 @@
...
+
+	txnErr := appCtx.NewTransaction(func(txnCtx appcontext.AppContext) error {
+
+		// Delete all existing report_violations for this report
+		existingReportViolations := models.ReportViolations{}
+		err := appCtx.DB().Where("report_id in (?)", reportID).All(&existingReportViolations)
+		if err != nil {
+			return err
+		}
+		err = appCtx.DB().Destroy(existingReportViolations)
```
leia-sefkin, line 29: same question as above comment on your migrations: did you figure out if it is a requirement to soft delete?

leia-sefkin: looking at this more i'm questioning whether we may run into trouble by deleting all existing report_violation entries and recreating them with the ones passed in. if i were to use this API my default would be to expect it to only add the new relationships i passed in.

interested to hear your thoughts, in particular on the FE side of things what makes more sense. if we want to stick with the current pattern i'd just recommend documenting explicitly in swagger and other places

leia-sefkin (commented):
just did a quick once over and wanted to start the convo on how we're handling update and delete of report violation relationships

pkg/handlers/ghcapi/report_violations.go
```diff
@@ -0,0 +1,76 @@
...
+//Handle is the handler for associating violations with reports
+func (h AssociateReportViolationsHandler) Handle(params reportViolationop.AssociateReportViolationsParams) middleware.Responder {
+	return h.AuditableAppContextFromRequestWithErrors(params.HTTPRequest,
+		func(appCtx appcontext.AppContext) (middleware.Responder, error) {
+
+			reportID := handlers.FmtUUIDToPop(params.ReportID)
+
+			var reportViolations models.ReportViolations
+			for _, violation := range params.Body.Violations {
+				violatinoID, err := uuid.FromString(violation.String())
```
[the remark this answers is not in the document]
PDickeyTruss: Good 👁️, fixed

pkg/services/report_violation/report_violation_creator.go
PDickeyTruss: Very good questions/points. This is something I went back and forth a lot on. I have not asked about soft-delete requirement-wise but will do so right now in slack. I spent some time adding soft delete but eventually talked myself into to hard/real delete being the best choice. I am of course very open for discussion around this. My thought was we don't retain the history of every field value in the form from each time it is saved and this is essentially a form value (even though it is joined from another table). Will post the question in slack now and go from there. 👍

PDickeyTruss: Question raised in slack here: https://ustcdp3.slack.com/archives/CP4979J0G/p1663707369130839

PDickeyTruss: Sounds like hard delete will be ok but I realized I didn't respond fully to your comment around deleting all existing report_violation entrees for a report before re-creating the new ones. I struggled with how best to do that and tried to use naming to help (`AssociateReportViolations` rather than `CreateReportViolations`) but I agree it still feels a little clunky. The main motivation behind deleting all reportViolations for a report before applying the new ones was to handle unselecting (need to 'unselect' any violations that may have been saved in the past but are not included in the current save). Another way to do this would be to have two distinct endpoints for delete (unselect) and create (select). This would certainly work but would require the FE do more to keep track of what needs to be unselected (deleted) when saved and make an additional delete request... it may be a bit less cryptic though. 🤷 Happy to zoom chat with ya to discuss more if you think it worthwhile. Either way we go I agree with your suggestion that documentation around it is needed. 😄

leia-sefkin: Hmm 🤔 more to think on! I more or less expected that this model was serving ease of use on the FE. I think what you have implemented makes sense given our existing requirements, hearing what you said and from Claire and Ryan - it feels more like two endpoints would be a little extraneous at this point (even if it's more technically correct). 

I say go with good documentation, just to clarify for anyone who might not dig down into the service files to see what's going on. The current name feels appropriate too. 

Thanks for entertaining my questions and thoughts :D I'll let you know if there's anything else that comes up as I dig into the submission workflow

AnnaGingle (approved):
Works as expected. Nice job!

pkg/models/report_violation.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"github.com/gobuffalo/pop/v6"
+	"github.com/gobuffalo/validate/v3"
+	"github.com/gofrs/uuid"
+)
+
+type ReportViolation struct {
+	ID          uuid.UUID        `json:"id" db:"id"`
+	Report      EvaluationReport `belongs_to:"evaluation_reports" fk_id:"report_id"`
```
duncan-truss, line 11: Hey Patrick, full disclosure I'm not very familiar with the work on this outcome.  I don't know how you all plan on querying this model but to me this looks like a join table where you likely want to define a `has_many` association on EvaluationReport to hold a slice of ReportViolations?  

I'm thinking it might not be needed to have this circular association where every ReportViolation will need the same copy of the EvaluationReport data to be populated.

Ryan-Koch (approved):

pkg/models/report_violation.go
PDickeyTruss: Great catch, you are 100% correct, there is no need for this circular association. I initially did the joining a bit different and I think this association is an artifact of that. I will clean that up, thanks!

migrations/app/schema/20220919160852_create_report_violations_table.up.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE IF NOT EXISTS report_violations (
+	id uuid PRIMARY KEY,
+	report_id uuid NOT NULL
+		CONSTRAINT report_violations_report_id_fkey
+		REFERENCES "evaluation_reports",
+	violation_id uuid NOT NULL
+		CONSTRAINT report_violations_violation_id_fkey
+		REFERENCES "pws_violations"
```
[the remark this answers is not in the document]
PDickeyTruss: Good question! They are always going to be unique, so a unique constraint would be a good idea here. I will add that in, thanks!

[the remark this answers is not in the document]
PDickeyTruss: Fixed ✅

pkg/models/report_violation.go
PDickeyTruss: Fixed ✅

PR #8417, opened 2022-04-12
MB-4001: Use pop schema for schema dump/restore to squash migrations

ahobson:
## Summary

Pop supports dumping the schema (using pg_dump). Use this to support dumping/loading from schema files.

Unfortunately, the existing migrations contain schema changes, data loading, and role/permission changes. Split those into 3 different files.

## Setup to Run Your Code

In theory, all interactions will remain the same, but just be faster

### Additional steps

This is still draft and needs more documentation and testing

## Verification Steps for Author

These are to be checked by the author.

- [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
- [ ] Request review from a member of a different team.
- [ ] Have the Jira acceptance criteria been met for this change?

## Verification Steps for Reviewers

These are to be checked by a reviewer.

### Backend

- [ ] Code follows the guidelines for [Logging](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide#logging)
- [ ] The requirements listed in [Querying the Database Safely](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide/#querying-the-database-safely) have been satisfied.

#### Any new migrations/schema changes:

- [ ] Follows our guidelines for [Zero-Downtime Deploys](https://transcom.github.io/mymove-docs/docs/dev/contributing/database/Database-Migrations#zero-downtime-migrations)
- [ ] Have been communicated to #g-database
- [ ] Secure migrations have been tested following the instructions in our [docs](https://transcom.github.io/mymove-docs/docs/dev/contributing/database/Database-Migrations#secure-migrations)

PR #9223, opened 2022-09-21
[ MB-13652, MB-14051, MB-14151 ] Add Advana Integration PSQL User

rogeruiz:
## Summary

This work is tied to the Advana integration and follows the
documentation on Secure Migrations.

[=> Secure migrations documentation](https://transcom.github.io/mymove-docs/docs/backend/setup/database-migrations#secure-migrations)

We will be actively testing this in Experimental.


migrations/app/secure/20220921160013_add_advana_dms_integration_psql_user.up.sql
```diff
@@ -0,0 +1,36 @@
...
+RESET ROLE;
+
+-- Set a password when running this locally.
+DO
+$do$
+BEGIN
+  IF NOT EXISTS (
+    SELECT -- SELECT list can stay empty for this
+    FROM   pg_catalog.pg_roles
+    WHERE  rolname = 'rdssuperuser') THEN
```
rogeruiz, line 21: @Ryan-Koch since you were the approver on #5567 I'm wondering if you remember what this `rolname` is doing here? 

Basically we want to set up a new crud user just for our DMS work and want to follow the conventions in the project to do so. 

cc: @reggieriser

rogeruiz (commented):
Leaving a comment for @transcom/truss-db to help out with on this asynchronously.

migrations/app/secure/20220921160013_add_advana_dms_integration_psql_user.up.sql
ahobson: When I was fuzting around with a change to how we might set up the database, I created a single file with all of the roles in it (with all the related comments):

https://github.com/transcom/mymove/pull/8417/files#diff-b6cc11bf54033fb50b8089a9813ddb53a6e86c2a37a7596c3ff675c4a71d01b6

In those comments, you can see that the roles are from RDS. I think this link in the section about "Understanding the rds_superuser role" has some additional relevant information.

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.PostgreSQL.CommonDBATasks.Roles.html

migrations/app/schema/20220921194349_add_advana_dms_integration_psql_user.up.sql
```diff
@@ -0,0 +1,36 @@
+-- Local test migration.
+-- This will be run on development environments.
+-- It should mirror what you intend to apply on loadtest/demo/exp/stg/prd
+-- DO NOT include any sensitive data.
+
+SET ROLE master;
+
+CREATE ROLE dms_crud WITH LOGIN NOINHERIT;
+
+GRANT rds_superuser TO dms_crud;
```
ahobson, line 10: https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.PostgreSQL.html suggests you also need `rds_replication` ?

Actually that doc suggests if you don't want to use the master account, you need to do a bit more https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.PostgreSQL.html#CHAP_Source.PostgreSQL.RDSPostgreSQL.NonMasterUser and I guess maybe you'd want to do that in this migration?

ahobson: I wonder if this migration should point to another location (like maybe in mymove docs) with all the relevant information on the requirements for DMS?

ahobson: After noodling on this overnight, do we have something like a mini-ADR describing why we think a separate user is the right approach? It does seem like it might be easier if we didn't create the separate user, but I'm not sure what the downsides would be

rogeruiz: We do need a separate user because we can't use IAM authentication for our DMS solution. DMS does not fully-support it an RDS IAM authenticated user which is what `crud` and `ecs_user` are in our Terraform.

We also need a separate user to be able to rotate and control access for this specific Advana integration. We're still in the planning phase here but an ADR could be useful in the future.

rogeruiz: > https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.PostgreSQL.html suggests you also need `rds_replication` ?

We shouldn't need to assign `rds_replication` as `rds_superuser` is part of that group. Another reason why we need a user like `dms_crud` is because `rds_superuser` can't login. The docs in the following picture are from your comment at this link.

[=> This link](https://github.com/transcom/mymove/pull/9223#discussion_r976930802) 

Captura de Pantalla 2022-09-22 a la(s) 10 28 04 a m

ahobson: Thank you! Do we use the `master` user when connecting from the application? I didn't think we did. My understanding is that the `master` user does not require IAM authentication, but maybe I'm mistaken?

ahobson: My understanding is that the `master` user is what we've set up in terraform.

https://github.com/transcom/terraform-aws-app-environment/blob/main/database/main.tf#L88-L89

That password is already managed by SSM and thus we could rotate it, right?

Because the DMS docs say it's possible to user the `master` user, I don't know if we have other settings that are preventing us from user it. Again, hope these are helpful comments.

rogeruiz: > Thank you! Do we use the `master` user when connecting from the application? I didn't think we did.

you're right! we don't do that. But once IAM authentication is used for the `MasterUser` we aren't able to log into the DB with the master user credentials. Also, those credentials are managed by SSM.

=> [Will enabling RDS IAM auth disable the password auth automatically?](https://stackoverflow.com/questions/61898308/will-enabling-rds-iam-auth-disable-the-password-auth-automatically)

[=> Slack thread where @chtakahashi mentions this](https://trussworks.slack.com/archives/C02KKS13KQE/p1661945423622429?thread_ts=1661889850.514619&cid=C02KKS13KQE) 🔒 

> That password is already managed by SSM and thus we could rotate it, right?

> Because the DMS docs say it's possible to use the master user, I don't know if we have other settings that are preventing us from using it. Again, hope these are helpful comments.

From @ahobson: The `MasterUser` may not use RDS IAM authentication, so that's why we could use `MasterUser`. 

We still feel like having a separate user here is the security posture we want. 

We are going to be managing the `dms_crud` user in Terraform using SSM to manage just the password. But we still need a secure migration here.

rogeruiz: We're going to have secure migrations but they will be empty for non-ATO environments. But we will be adding the triggers for ATO environments for now. Said another way, our migrations will be no-op for all non-ATO including local as we don't need to have these extra triggers on anything where DMS is not running.

jesset-truss:
Closing this PR out as part of the MB-14051 AC. We've captured and documented the changes. Please see MB-14051 for details. After closing this PR, we will remove the branch.

PR #9226, opened 2022-09-22
MB-6682 create and use new 404 page design

YanZ777:
## Summary

The new 404 page design has been implemented to display a 404 when a user attempts to reach a page that isn't found. 
The `spa_handler.go` has also been changed to return a 404 when a user attempts to access a directory.
I've also added tests for the `spa_handler`.

### Open questions and requests
Please scrutinize what I've done in `spa_handler.go` and the new `spa_handler_test.go`

Why is the behavior different between when I locally run things and the ephemeral deploy?

How do I get the frontend to display the 404 page rather than just a page with the 404 error on it? e.g. I'll see
image

[This article](https://www.alexedwards.net/blog/disable-http-fileserver-directory-listings#using-a-custom-filesystem) explains more about the approach used.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

Would recommend running this as an ephemeral deploy rather than locally. There's differing behavior between the two environments, but I'm not sure why.

1. Access the customer or office app
2. Attempt to access pages that don't exist, in particular [http://officelocal:3000/static/media](ttp://officelocal:3000/static/media) and [http://milmovelocal:3000/downloads](http://milmovelocal:3000/downloads)
3. The page returned should be the new error page.

### Frontend

### Backend

### Customer desktop view
image

### Customer phone view
image

### Office view
image


pkg/handlers/spa_handler.go
```diff
@@ -43,6 +44,11 @@ func (h SpaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if strings.HasSuffix(r.URL.Path, "/") {
+		http.NotFound(w, r)
+		return
+	}
```
ahobson, line 50: I *think* the go way to do this would be to add a wrapper that would return 404 if the requested path is not a file.  The go documentation has an [example for hiding dotfiles](https://pkg.go.dev/net/http#example-FileServer-DotFileHiding) that I think we could modify.

I'm the one that added spa_handler, I think and I didn't add a test for it 😢 . I think we should add one now, though.

Thoughts?

ahobson: I found an [example that we could crib from](https://www.alexedwards.net/blog/disable-http-fileserver-directory-listings), I think

src/pages/Office/index.jsx
```diff
@@ -345,6 +346,8 @@ export class OfficeApp extends Component {
                         }
                       }}
                     />
+                    {/* 404 */}
+                    <Route render={(routeProps) => <NotFound {...routeProps} handleOnClick={history.goBack} />} />
```
ahobson, line 350: Should we add a test for this in the office app? Can we do that?

pkg/handlers/spa_handler.go
YanZ777: I also found the article in your second comment. I tried cribbing from that one, in particular, the section on "Using MIddleware", but let me also look at the hiding dotfiles example too. 
And yes, let's add tests.

src/pages/Office/index.jsx
YanZ777: Yes, yes we should (and can)!

pkg/handlers/spa_handler.go
ahobson: I think the custom filesystem approach is the best one instead of middleware. Here's why I think that: with middleware, there's still a redirect and then a 404. I think that means an attacker could figure out a directory exists and then try to guess filenames. With the custom filesystem, there's no redirect, so only an immediate 404.

Does that make sense? I'm happy to talk it through with you if you'd like.

ahobson: If I'm thinking about things correctly, we wouldn't need a middleware if implemented a custom filesystem.

YanZ777: I've been able to replace the `spa_handler` when managing the `/static` and `/downloads` folders, but I'm having trouble doing that when requests are made to `/internal/swagger.yaml` since the `spa_handler` gets used when `routingConfig.serveAPIInternal` and the like set to true. 
Was the intention to use the custom filesystem to replace those requests too?

ahobson: I think the only change you need to make in the go code is to the SpaHandler. If I use the [code here](https://gist.github.com/ahobson/92936d30adbffcadf207f5be8bbd73ea), it prevents showing directory listings

juliemoon (commented):
Hi @YanZ777 this looks great! The functionality works as expected. I saw the font size and spacing looks different from the specs for the desktop customer & office application. I linked the figma files in here but I'm also happy to jump on a call as well. 

customer application : https://www.figma.com/file/tRkjtoWjT8727XH5m9k7mK/Milmove-Customer-App?node-id=1030%3A15806

office application : https://www.figma.com/file/5AhuEkaH3bJVrtffU3wp22/Milmove-Office-App?node-id=4932%3A74818

juliemoon (approved):
@YanZ777 thank you for making the changes. It looks great in mobile and desktop!
Screen Shot 2022-10-26 at 4 12 31 PM
Screen Shot 2022-10-26 at 4 13 29 PM

pkg/handlers/routing/routing_init.go
```diff
@@ -140,9 +140,16 @@ func InitRouting(appCtx appcontext.AppContext, redisPool *redis.Pool,
 	clientCertMiddleware := authentication.ClientCertMiddleware(appCtx)
 
 	// Serves files out of build folder
+	cfs := handlers.NewCustomFileSystem(
+		http.Dir(routingConfig.BuildRoot),
+		"index.html",
+		appCtx.Logger(),
+	)
+
```
rogeruiz, line 148: Loving this. Good work on discussing this @ahobson and @YanZ777

pkg/handlers/spa_handler.go
```diff
@@ -34,33 +80,19 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
-	if os.IsNotExist(err) {
-		// file does not exist, serve index.html
-		http.ServeFile(w, r, filepath.Join(h.staticPath, h.indexPath))
-		return
-	} else if err != nil {
-		// if we got an error (that wasn't that the file doesn't exist) stating the
-		// file, return a 500 internal server error and stop
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-		return
-	}
```
rogeruiz, line 60: Glad to see this code go where it served a redirect instead for a 404 or a 500 for finding the file but stating the file. ㊙️

pkg/handlers/spa_handler_test.go
```diff
@@ -0,0 +1,155 @@
...
+			request:            "test/a",
+			expectedStatusCode: http.StatusOK,
+			expectedBody:       "file a",
+		},
+		{
+			name:               "A file that exists in a directory that does not have an index.html",
+			request:            "test/noIndexDir/b",
+			expectedStatusCode: http.StatusOK,
+			expectedBody:       "file b",
+		},
```
rogeruiz, line 120: These tests are interesting? are they covering the cases where we'd like to serve certain files like Swagger Yaml files and static images and more?

src/pages/Office/index.test.jsx
```diff
@@ -503,5 +503,16 @@ describe('Office App', () => {
...
+    describe('page not found route', () => {
+      it('handles a nonexistent route by returning a 404 page', () => {
+        render(
+          <MockProviders initialEntries={['/pageNotFound']}>
+            <ConnectedOffice />
+          </MockProviders>,
+        );
+        expect(screen.getByText('Error - 404')).toBeInTheDocument();
+      });
+    });
```
rogeruiz, line 516: Loving this straightforward test.

rogeruiz (approved):
lgtm 🌈

great work on this @YanZ777 and helpful reviews on this @ahobson and @juliemoon. 🎉

pkg/handlers/spa_handler_test.go
YanZ777: I think they are covering those cases since the tests are testing that individual files are able to be served from directories that either do have or don't have an index.html. I don't know if there are any particularities with Swagger Yaml files and static images that would change the behavior enough from a file created this way though.

PR #9230, opened 2022-09-22
Mb 13703 move details page scroll bug nt

NamibiaTorres:
## Summary

Getting some pre-commit errors related to the changes in `ScrollToTop.jsx`

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make office_client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

Before starting make sure that there are test users for the office client.

1. Access the [office app](http://officelocal:3000/sign-in) in experimental
2. Login as a the service_counselor
3. select a move with a PPM
4. Edit the PPM and save changes
5. You will likely go to a page where you are asked to leave comments as the Service Counselor, do this and `save and continue`.
6. On the Move Details page, make sure that you see that the page starts at the top, and not in the middle of the page, so that you have to scroll to the top.
7. check out of my branch and check into master, then replicate steps 1-6, you should see that on the move details page, the page scrolling is off and does not bring the user to the top of the page. This is the bug that was fixed.

- [ X] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
- [X ] Have the Jira acceptance criteria been met for this change?

### Frontend

### DIFF
--- a/src/components/ScrollToTop.jsx
+++ b/src/components/ScrollToTop.jsx
@@ -4,13 +4,13 @@
 
 import { useEffect } from 'react';
 import { useLocation } from 'react-router-dom';
-
-export default function ScrollToTop({ otherDep = null }) {
+// TODO: Make this function only take otherDeps, but change uses everywhere else
+export default function ScrollToTop({ otherDep = null, otherDeps = [] }) {
   const { pathname } = useLocation();
 
   useEffect(() => {
     window.scrollTo(0, 0);
-  }, [pathname, otherDep]);
+  }, [pathname, otherDep, ...otherDeps]);
 
   return null;
 }
--- a/src/components/ScrollToTop.jsx
+++ b/src/components/ScrollToTop.jsx
@@ -5,12 +5,12 @@
 import { useEffect } from 'react';
 import { useLocation } from 'react-router-dom';
 // TODO: Make this function only take otherDeps, but change uses everywhere else
-export default function ScrollToTop({ otherDep = null, otherDeps = [] }) {
+export default function ScrollToTop({ otherDep = null }) {
   const { pathname } = useLocation();
 
   useEffect(() => {
     window.scrollTo(0, 0);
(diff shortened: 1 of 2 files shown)


src/components/ScrollToTop.jsx
```diff
@@ -4,13 +4,13 @@
...
-
-export default function ScrollToTop({ otherDep = null }) {
+// TODO: Make this function only take otherDeps, but change uses everywhere else
+export default function ScrollToTop({ otherDep = null, otherDeps = [] }) {
   const { pathname } = useLocation();
 
   useEffect(() => {
     window.scrollTo(0, 0);
-  }, [pathname, otherDep]);
+  }, [pathname, otherDep, ...otherDeps]);
```
NamibiaTorres, line 13: Ok, apprarently, React doesn't like when I do this, and I found a thread talking about this so I'm going to try something different: https://stackoverflow.com/questions/56262515/how-to-handle-dependencies-array-for-custom-hooks-in-react

ronaktruss: If multiple dependencies are common throughout the application, it probably makes sense to abstract this so that it accepts more than one argument/dependency. 

Though, if this is a one off thing we may be able to get away with doing an `||` or some other logic to the variables we care about before we send them as props. For example may be able to just do:
```js
<ScrollToTop otherDep={alertMessage || infoSavedAlert} />
```

NamibiaTorres: Hmm, ok. I was debating whether it would be more common. From what I could tell, it only happens in this file, but theoretically it could happen in others. For now, I can do this, and we can consider changing the ScrollToTop function later.

rileyorr:
@NamibiaTorres I followed your outlined steps and the page scroll worked for me 👍

pearl-truss (approved):
LGTM

rileyorr (approved):
Design changes look good to me - nice job!

PR #9233, opened 2022-09-22
MB-12189 Set up FE data factories and implement duty locations

scottmries:
## Summary

This PR sets up baseline factories for generating test data on the front end. I've provided the `dutyLocation` factory and implemented it in various places as a real-world example.

## Setup to Run Your Code

<details>
<summary>💻 You will need to run a docker command to generate a file, and then run standard client tests.

##### Run tests

`make client_test`

- [ ] Tests pass without additional errors
- [ ] Documentation is sufficient to implement further work
- [ ] Since data is being loaded from the file system, ensure that it is doing so safely


src/utils/test/factories/README.md
```diff
@@ -0,0 +1,199 @@
...
+  }
+}
+```
+
+### Traits
+
+Use traits by setting the `useTraits` key to an array of traits. Traits are applied in order, overriding any default values, before postBuild is executed.
+
+`{ [BASE_FIELDS.USE_TRAITS]: [MY_OBJECT_TRAITS.TRAIT_1, MY_OBJECT_TRAITS.TRAIT_2] }`
+
```
ronaktruss, line 194: While this can be inferred from reading the description of all three options, It might be helpful to have an explicit sentence indicating that the precedence is overrides < traits < lazy overrides.

src/utils/test/factories/README.md
```diff
@@ -0,0 +1,199 @@
...
+```sh
+
+docker run -it -v ${PWD}:/local "parsertongue/swagger-codegen-cli:latest" generate \
+-i /local/swagger/internal.yaml \
+-l javascript \
+-o /local/src/utils/test/swagger-client/internal \
+-DmodelDocs=false -DmodelTests=false -DapiDocs=false -DapiTests=false
+
+```
```
ronaktruss, line 147: We can probably get rid of these lines and update the title since we are not using swagger codegen and instead using js-yaml.

src/utils/test/factories/base.js
```diff
@@ -0,0 +1,126 @@
...
+const baseFactory = (params) => {
+  const { fields, postBuild, lazyOverrides, overrides, traits, useTraits } = params;
+
+  const appliedFields = applyOverrides(fields, overrides);
+  const formattedFields = camelCaseFields(appliedFields);
+
+  const builder = build({
+    fields: formattedFields,
+    postBuild: basePostBuild(lazyOverrides, postBuild),
+    traits,
```
kylehilltruss, line 120: The `BASE_FIELDS` properties aren't used anywhere else in this file. Do we want to worry about consistency w/r/t using the exported property names here? (`{ fields: ` instead of `{ [BASE_FIELDS.FIELDS]: ` for instance) Or is this just an optional, use-if-you-want setup?

src/utils/test/factories/helpers.js
```diff
@@ -0,0 +1,23 @@
...
+export const gblocHelper = (f) => f.random.alpha({ count: 4, casing: 'upper' });
+export const idHelper = () => uuidv4();
+export const phoneHelper = (f) => f.phone.number(PHONE_FORMAT);
+export const placeNameHelper = (f) => f.lorem.words(Math.random(4) + 1);
+export const dutyLocationNameHelper = (f) => `${f.lorem.words(Math.random(4) + 1)} AFB`;
+export const stateHelper = () => oneOf(...spec.definitions.Address.properties.state.enum).call();
+
+export const updatedAtFromCreatedAt = (createdAt) => faker.date.between(createdAt, Date.now());
+
+export default {};
```
kylehilltruss, line 23: Why are we exporting a default empty object? (It might make sense to export default an object containing all of the individually exported helper functions.)

src/utils/test/factories/transportationOffice.js
```diff
@@ -0,0 +1,38 @@
...
+  ADDRESS: 'address',
+  PHONE_LINES: 'phone_lines',
+  GBLOC: 'gbloc',
+  LATITUDE: 'latitude',
+  LONGITUDE: 'longitude',
+};
+
+const transportationOfficeFactory = (params) => {
+  const getPhoneLines = () => {
+    const numberOfLines = Math.random(4);
```
kylehilltruss, line 17: `Math.random` doesn't accept any parameters. What's the 4 doing?

src/utils/test/factories/helpers.js
```diff
@@ -0,0 +1,23 @@
...
+
+const PHONE_FORMAT = '###-###-####';
+const spec = getInternalSpec();
+
+export const agencyHelper = () => oneOf(...Object.keys(serviceMemberAgencies)).call();
+export const createdAtHelper = (f) => f.date.recent();
+export const gblocHelper = (f) => f.random.alpha({ count: 4, casing: 'upper' });
+export const idHelper = () => uuidv4();
+export const phoneHelper = (f) => f.phone.number(PHONE_FORMAT);
+export const placeNameHelper = (f) => f.lorem.words(Math.random(4) + 1);
```
kylehilltruss, line 17: See below for comment on Math.random accepting params

src/utils/test/factories/discovery/factory.js
```diff
@@ -0,0 +1,41 @@
...
+export const TEST_FACTORY_FIELDS = {
+  DEFAULT: 'default',
+  SUBFACTORY: 'subfactory',
+  POST_BUILD_TOUCHED_FIELD: 'postBuildTouchedField',
+  SNAKE_CASE_FIELD: 'case_field',
+  CAMEL_CASE_FIELD: 'caseField',
+};
+
+export const TEST_FACTORY_TRAITS = {
+  TEST_TRAIT: 'testTrait',
```
kylehilltruss, line 16: I hate to ask "what is this" in a code review, but, well... the contents of this file feels like testing/discovery code, and not something that we should commit to the project's code base (at the very least, not without explanatory comments).

kylehilltruss (changes requested):

src/components/Customer/FactoryTest/FactoryTest.test.jsx
```diff
@@ -0,0 +1,92 @@
+import { BASE_FIELDS } from 'utils/test/factories/base';
```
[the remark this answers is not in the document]
scottmries: Updated with the new filename and path.

src/utils/test/factories/base.js
scottmries: Updated to use constants here too.

src/utils/test/factories/helpers.js
scottmries: Fixed.

src/utils/test/factories/transportationOffice.js
scottmries: Good eye, I don't know why I think this method works this way. This and the other instances are fixed.

src/utils/test/factories/discovery/factory.js
scottmries: This is set up to work exactly the way the fields do, so user of the factory can just look up the name of the trait in the IDE. Noted that the trait test was not using this constant, which I fixed.

src/utils/test/factories/address.js
```diff
@@ -0,0 +1,37 @@
...
+const addressFactory = build({
+  fields: {
+    [ADDRESS_FIELDS.ID]: fake(idHelper),
+    [ADDRESS_FIELDS.STREET_ADDRESS_1]: fake((f) => f.address.streetAddress()),
+    [ADDRESS_FIELDS.STREET_ADDRESS_2]: fake((f) => f.address.secondaryAddress()),
+    // left out streetAddress3 since we don't even let users input that line...
+    [ADDRESS_FIELDS.CITY]: fake((f) => f.address.city()),
+
+    [ADDRESS_FIELDS.STATE]: fake(stateHelper),
+    [ADDRESS_FIELDS.COUNTRY]: 'US', // Likely change once we support more than just OCONUS moves.
```
YanZ777, line 28: **suggestion:** Could you make a Jira ticket and put the ticket number in a comment here too?

src/utils/test/factories/README.md
```diff
@@ -0,0 +1,197 @@
+# Factories
```
YanZ777, line 1: **question:** Is this intended to live here? Should it live inside the [MilMove developer documentation](https://transcom.github.io/mymove-docs/) instead?

src/utils/test/factories/base.js
```diff
@@ -0,0 +1,124 @@
...
+  // RA: system.
+  // RA: In this case, this doesn't contain user input and serves to be able to grab the local swagger files.
+  // RA Developer Status: Mitigated
+  // RA Validator Status: {RA Accepted, Return to Developer, Known Issue, Mitigated, False Positive, Bad Practice}
+  // RA Modified Severity:
+  /* eslint-disable-next-line security/detect-non-literal-fs-filename */
+  return yaml.load(fs.readFileSync(yamlPath, 'utf8'));
+};
+
+let internalSpec = null;
```
YanZ777, line 30: **question:** Do `internalSpec` (and similarly, `ghcSpec` below) need to be declared outside their respective functions? I'm not seeing where else they're being used besides those places.

YanZ777 (commented):
Aside from my questions and suggestions, the code is looking good!

src/utils/test/factories/README.md
felipe-lee: **issue**: Yeah, I think it should live in the mm dev docs like Yan said. Having this doc here is nice in that it lives next to where people will use the code, but it deviates from how the rest of our docs work (other than the top level readme, which tbh should also probably be migrated at some point).

src/utils/test/factories/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Writing a new factory
+
+Create and export a const called `{OBJECT_}FIELDS`. This constant should be an object whose keys are CAPITAL_SNAKE_CASE and whose values are the actual, camelCase names of the fields, e.g. `MY_FIELD: myField`. While this step is not strictly necessary, it will make using your factory much easier, since the keys in question will be available to your IDE.
```
felipe-lee, line 5: **thought**: While it's true that it isn't strictly necessary, we should probably decide if we do want to do this, and if so, then we should say to do it. I feel like our docs saying something is optional will mean we won't be consistent. Maybe it won't be a big issue if we aren't consistent, but I think the docs should probably not leave that open.

felipe-lee: **suggestion**: I think it might be good to show an example that is more ironed out than the inline one we have currently. Especially once this doc gets move to be in our dedicated docs repo.

src/utils/test/factories/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+Create a const called `{object}Factory`, giving it a camelCase name. This const is a function that
+
+- takes params
+- returns a `baseFactory`, into which is passed an object containing
+  - `fields`: the fields your factory returns
+  - any other configuration passable to the `build()` functions [here](https://www.npmjs.com/package/@jackfranklin/test-data-bot), such as `postBuild` or `traits`.
+  - the spread params: `...params`.
```
felipe-lee, line 13: **suggestion**: I think an example of this, even as a simple case, would be great to have in the docs.

src/utils/test/factories/README.md
```diff
@@ -0,0 +1,197 @@
...
+  }
+```
+
+#### Literals
+
+If your field's value is a literal, the factory will always return that value. For example:
+
+`field: 'my value'`
+
+will always set the value of `field` to `'my value'` by default (i.e., unless)
```
felipe-lee, line 41: **suggestion**: the `i.e.` bit at the end is a bit confusing to me so I'd suggest removing it, but maybe I'm missing something.

```suggestion
will always set the value of `field` to `'my value'` by default.
```

src/utils/test/factories/README.md
felipe-lee, line 1: **praise**: Thanks for getting the docs started for this!

felipe-lee: **question**: are these docs meant to cover the work for https://dp3.atlassian.net/browse/MB-12838? Or are they more of a baseline and when we get to that ticket, we'll iron things out more and add more details/examples? I ask to know how much to request in terms of details and such vs only focusing on making sure things make sense with the knowledge that we'll add more later.

felipe-lee (commented):
Thank you for continuing the work on this!

I'm still looking through this, but I've got a few questions, comments, etc. I'll continue looking on Monday!

src/utils/test/factories/base.js
felipe-lee: **thought**: good question, I think the way I'd written this locally at first did require them to be defined at a top level, but then I switched to the `get<api>Spec` functions and so they do seem to remove the need for the declaration at this level. Thanks for calling that out.

src/utils/test/factories/address.js
scottmries: I created the ticket and added the number to the comment. It's [here](https://dp3.atlassian.net/browse/MB-13996), for the record.

src/utils/test/factories/README.md
scottmries: I removed this file from this repo and moved it to a PR in the [docs](https://github.com/transcom/mymove-docs/pull/214). I'll make any remaining changes in this PR to the code in that repo, and then maybe we move docs discussion there?

scottmries: @felipe-lee I had only really intended these as baseline for anyone coming into this fresh, but I'm assuming we'll also want to back-fill each of [these](https://github.com/transcom/mymove-docs/blob/main/docs/frontend/testing/writing-frontend-tests-for-milmove.md#address). I'll include a request to change addresses in my PR in the docs.

src/utils/test/factories/base.js
scottmries: thought: I refactored these to be just aliases of their respective calls. I don't believe the falsy check is necessary anymore.

src/utils/test/factories/README.md
felipe-lee: Ah good catch, I'm not sure I'd seen those, or if I had, I'd forgotten, but they would be good to swap out.

src/utils/test/factories/README.md
scottmries: I removed the "strictly necessary" language and added an example over in the [docs PR](https://github.com/transcom/mymove-docs/pull/214).

src/utils/test/factories/README.md
scottmries: I removed this over in the [docs PR](https://github.com/transcom/mymove-docs/pull/214).

src/utils/test/factories/README.md
scottmries: I added an example in the [docs PR](https://github.com/transcom/mymove-docs/pull/214).

src/utils/test/factories/base.js
```diff
@@ -0,0 +1,113 @@
...
+const applyOverrides = (object, overrides) => {
+  const appliedFields = object;
+  Object.entries(object).forEach(([field, value]) => {
+    let appliedValue;
+    switch (typeof value) {
+      case 'function':
+        if (overrides && overrides[field]) {
+          appliedValue = value({ [BASE_FIELDS.OVERRIDES]: overrides[field] });
+          // overrides are delegated to the function; don't handle them here:
+          delete overrides[field];
```
felipe-lee, line 55: **question**: Should we be deleting the field from the override in every section? If we do, is there a reason to continue passing the overrides to the builder down on line 110?

src/utils/test/factories/base.js
```diff
@@ -0,0 +1,113 @@
...
+
+const BASE_FIELDS = {
+  FIELDS: 'fields',
+  OVERRIDES: 'overrides',
+  LAZY_OVERRIDES: 'lazyOverrides',
+  POST_BUILD: 'postBuild',
+  TRAITS: 'traits',
+};
+
+const applyOverrides = (object, overrides) => {
```
felipe-lee, line 46: **suggestion**: can you add docstrings to these functions? Especially to make it easier to understand the different things `object` can be.

src/utils/test/factories/discovery/factory.js
```diff
@@ -0,0 +1,41 @@
+import { perBuild } from '@jackfranklin/test-data-bot';
```
felipe-lee, line 1: **question**: are the files in the `discovery` dir meant to be examples of how to make factories? Or were they meant to be removed now that this pr is trying to be merged? If we do want to keep them around, it might be good to add a comment or something stating they're mainly examples so people don't delete them or think they're something else.

src/utils/test/factories/address.js
```diff
@@ -0,0 +1,38 @@
...
+      [ADDRESS_FIELDS.COUNTRY]: 'US', // Likely change once we support more than just CONUS moves: JIRA ticket MB-13996
+    },
+    postBuild: (address) => {
+      address.postalCode = faker.address.zipCodeByState(address.state);
+
+      return address;
+    },
+    ...params,
+  });
+};
```
felipe-lee, line 36: **suggestion**: Can you add the `ONLY_BASIC_ADDRESS` trait from the other pr? https://github.com/transcom/mymove/pull/9142/files#diff-b25a459b02fa21a858ddd1db0f12409d125b99c6b1c983d2e107845cf82f7162R39-R47

Mainly since in that one it did come up as one that can be useful in tests. The `ONLY_REQUIRED_FIELDS` didn't get used so that one can get left out, though we may eventually add it.

src/utils/test/factories/base.js
```diff
@@ -0,0 +1,113 @@
...
+import yaml from 'js-yaml';
+import { _ } from 'lodash';
+
+faker.setLocale('en_US');
+
+const fake = (callback) => {
+  return perBuild(() => callback(faker));
+};
+
+const loadSpec = (fileName) => {
```
scottmries, line 16: This function uses the file system and is called in two locations, within `getInternalSpec` and `getGHCSpec` just below. In both cases, it's called with a hardcoded path, and it's not exported, so it cannot be called with an arbitrary path.

src/utils/test/factories/base.js
```diff
@@ -0,0 +1,113 @@
...
+      applyOverrides(object, lazyOverrides);
+    }
+    return object;
+  };
+};
+
+const camelCaseFields = (fields) => {
+  const formattedFields = {};
+  Object.entries(fields).forEach(([field, value]) => {
+    const formattedField = _.camelCase(field);
```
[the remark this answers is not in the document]
scottmries: Oh makes sense, I didn't realize we were phasing out lodash. I'll push up a fix.

src/utils/test/factories/base.js
scottmries: This was a really good catch. I pushed up a refactor to this function that doesn't need to delete these overrides (which seemed odd), and doesn't pass them into the builder. I think the function is much easier to understand now.
The issue was that `preBuild`s actually become objects with a `call` function, and we weren't properly handling that case.

src/utils/test/factories/base.js
scottmries: That's a good suggestion. It's definitely hard to get a handle on how this works by just glancing over it. 
I pushed up a commit with docstrings.

src/utils/test/factories/discovery/factory.js
scottmries: I think we should keep them here to make sure the tests run and they work as we expect. I added a comment to the top of each file.

src/utils/test/factories/address.js
scottmries: Done!

kylehilltruss (approved):

PR #9234, opened 2022-09-22
loadtest cert update migration https://dp3.atlassian.net/browse/MB-13808

CoryAtTruss:
Updated the certificate in all the places in AWS and CircleCI
Updated the certificates in here
Ran the following commands:
bin/milmove gen certs-migration --update -f **** -s **** -n **** --certid ****

ran (UPDATED):

ENVIRONMENTS="loadtest" upload-secure-migration tmp/20220923135119_app_loadtest_cert_migration.up.sql

ENVIRONMENTS="demo stg prd exp" upload-secure-migration migrations/app/secure/20220923135119_app_loadtest_cert_migration.up.sql


duncan-truss:
@CoryAtTruss Running `bin/milmove gen certs-migration` should have updated the migrations/app/migrations_manifest.txt with a unique migration entry and added the empty SQL file to the migrations/app/secure folder.  I would expect those 2 changes to be committed in this PR.

I used the migration name in your description to try downloading it but got a not found error when trying `scripts/download-secure-migration 20220908163100_app_loadtest_cert_migration.up.sql`

The timestamp of that migration looks like the prior one for experimental, I would expect this one to be newer although it shouldn't matter as long as some other part of the name has changed.

CoryAtTruss:
> 

Definitely saw that before - something weird happened when I was committing things with VSC the first time. The two files should be there now, and I corrected the timestamps. Thank you for your help! Let me know how it looks now @duncan-truss

duncan-truss (approved):
Changes look good now!

PR #9270, opened 2022-09-29
[MB-13817] Fix missing "create report" button on first login

mboren:
## Summary
When office users were signing in with login.gov for the very first time, they didn't have access to actions that they should have had permission for. For example, new QAE users wouldn't see "create report" buttons on the quality assurance page.
However, when they signed out and signed in again, they would have the correct permissions and things would work fine.

### why did this happen?
There are two code paths that an office user can go through when they log in:
- if it's their first time logging in, we won't have their login.gov UUID saved yet, so they go through `authorizeUnknownUser`
- if they've logged in before, we can look them up in our database by login.gov UUID, and they go through `authorizeKnownUser`

When we were setting up the new permission system, we added code to stick their permissions in their session in `authorizeKnownUser`, but forgot to put it in `authorizeUnknownUser`, so they wouldnt have any permissions on their first login.

This has gone unnoticed for a while because we typically reuse the same office user accounts when we're testing on staging, so it didn't pop up until the user exercise.

In future work, we should refactor `authorizeKnownUser` and `authorizeUnknownUser` to avoid having permissions applied in multiple places.
It would make more sense split this up into an authentication step and an authorization step. Authentication can take into account the first time user login stuff, and authorization should be the same for everybody.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps
- Create an account on login.gov
- then go to admin interface, (you can use local sign-in or login.gov sign-in)
- as this admin user, create a new office user using your login.gov email
- log in to office app using login.gov
- go to quality assurance page for some move
- if all is well, you should see "create report" buttons

When testing locally, you can reuse one login.gov account.
To simulate its first login, just wipe your database:
```sh
make db_dev_reset
make db_dev_e2e_populate
```
 and recreate an office user for it in the admin interface


ahobson (approved):
🚀

leia-sefkin (approved):
Tested locally and it works like a charm. Thank you for catching this fix I completely missed on our first round of development of this. 

+1 to refactoring this authorization/authentication flows overall so they're more future proof and less redundant.

ahobson:
@mboren This really is great stuff! I was working on something else when I noticed, however, that the current permissions middleware doesn't seem to use the `Permissions` on the session.

https://github.com/transcom/mymove/blob/master/pkg/handlers/authentication/permissions.go#L87-L91

I fixed/changed that in something else I'm working on. Do you think it would be appropriate to include that fix here?

mboren:
> @mboren This really is great stuff! I was working on something else when I noticed, however, that the current permissions middleware doesn't seem to use the `Permissions` on the session.
> 
> https://github.com/transcom/mymove/blob/master/pkg/handlers/authentication/permissions.go#L87-L91
> 
> I fixed/changed that in something else I'm working on. Do you think it would be appropriate to include that fix here?

I'm definitely happy to through an additional fix in here if that's helpful.
If the permissions middleware isn't using the permissions from the session, I guess the session permissions must just be used by the frontend for visibility?

mboren:
@ahobson I took a stab at incorporating the middleware change and ran into some test issues that I haven't had time to sort out, so I think I'll probably have to leave that out to get this merged tomorrow.

PR #9277, opened 2022-09-29
MB-14156: global scroll to top

ronaktruss:
## Summary

This refactors the existing `ScrollToTop` component with a similar component that is only used at the top level of the App. We have also created a `NotificationScrollToTop` component that is used for when we need to scroll to the top of a page when a notification (usually a server error/success) occurs.

While there are a decent number of file changes looking through the commits should give you a better understanding of the actual changes that happened.

Note: The behavior when using the default browser navigation (forward and back buttons) is a bit weird. On chrome the general behavior is to maintain the positioning of where you were when you were last on this page. Generally this seems to be the case, but the behavior is not consistent. This inconsistency is existed prior to these changes. I time boxed looking into this issue, but was not able to find the root cause.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

To test the scenario in the screen capture that is shown in the ticket
1. Access the customer app
2. Login as `hhg@only.unsubmitted`
3. Click `Add another shipment`
4. Select `HHG` and scroll down.
5. Click `Next`
6. You should see the page starting at the top rather than the previous scroll positioning.

You should also test other portions of the customer and office application where the url changes or when the server sends a message that is displayed to the user.


src/components/ScrollToTop.jsx
```diff
@@ -5,12 +5,12 @@
...
 import { useLocation } from 'react-router-dom';
 
-export default function ScrollToTop({ otherDep = null }) {
+export default function ScrollToTop() {
   const { pathname } = useLocation();
 
   useEffect(() => {
-    window.scrollTo(0, 0);
-  }, [pathname, otherDep]);
+    window.scroll({ top: 0, left: 0, behavior: 'instant' });
```
ronaktruss, line 12: We don't need this, but an instant scroll made more sense for this than a smooth scroll.

Note: instant is not supported in the [editor draft documentation](https://drafts.csswg.org/cssom-view/#enumdef-scrollbehavior) (so many sites like mdn don't have documentation for this), but this is supported in the [working draft documentation](https://www.w3.org/TR/cssom-view-1/#enumdef-scrollbehavior). Also based on my testing chrome, firefox, safari, and edge all support this.

felipe-lee, line 12: **praise**: Thanks for changing this to have named args, makes it easier to know what it is actually doing vs `0, 0`

src/components/NotificationScrollToTop.jsx
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useRef } from 'react';
+
+export default function NotificationScrollToTop({ dependency }) {
+  const isMounted = useRef(false);
```
felipe-lee, line 4: **question**: was this causing issues on the first mount without the `ref`? Mostly curious, not asking for this to change.

src/containers/FlashGridContainer/FlashGridContainer.test.jsx
```diff
@@ -82,19 +82,18 @@ describe('FlashGridContainer component', () => {
...
     );
-
-    // ScrollToTop should fire at the initial mount
-    expect(global.scrollTo).toHaveBeenCalledTimes(1);
+    // ScrollToTop should not fire at the initial mount
+    expect(global.scrollTo).toHaveBeenCalledTimes(0);
 
     mockStore.store.dispatch(setFlashMessage('TEST_SUCCESS_FLASH', 'success', 'This is a successful message!'));
 
     // Re-render after changing state with new message and ScrollToTop should fire again
```
felipe-lee, line 90: **suggestion**: Based on the other changes you made, I think this comment should change to something like:

```suggestion
    // Re-render after changing state with new message and ScrollToTop should fire for the first time
```

felipe-lee (approved):
LGTM, thanks for making this change! 

I didn't go through the app to test every page you changed, but I tested all the customer ppm pages as well as a few other scattered pages like the customer agreement page and the SC move details page.

src/components/NotificationScrollToTop.jsx
ronaktruss: I don't think a comment is necessary for the reason that you mentioned

src/components/NotificationScrollToTop.jsx
```diff
@@ -0,0 +1,15 @@
...
+
+  useEffect(() => {
+    if (isMounted.current) {
+      window.scrollTo(0, 0);
```
YanZ777, line 8: **suggestion:** Since we're using named args in `ScrollToTop.jsx`, should this one have named args too?
```suggestion
      window.scrollTo({ top: 0, left: 0, behavior: 'instant' });
```

ronaktruss: I'm primarily using named args because we need to override the scroll-behavior. In this instance we aren't overriding it and the values are 0 so I don't think modifying this provides much value.

duncan-truss:
@ronaktruss The flows I worked through on the Customer app all looked really good.

The only edge case I came across was in the office app on the Move Task Order page where we replace notifications.  So if you request a reweigh on a shipment it displays a notifciation, then if you scroll to request a diversion the notification changes but we aren't scrolled back up to the top to see the change.

ronaktruss:
> @ronaktruss The flows I worked through on the Customer app all looked really good.
> 
> The only edge case I came across was in the office app on the Move Task Order page where we replace notifications. So if you request a reweigh on a shipment it displays a notification, then if you scroll to request a diversion the notification changes but we aren't scrolled back up to the top to see the change.

Is there a specific user/scenario that you tried this on? This seems to be working on my end in both chrome and firefox.

PR #9259, opened 2022-09-28
Mb 13859 db sc review weight tickets nt

NamibiaTorres:
## Summary

After the service member submits their closeout documentation, including at least one weight ticket record, the Services Counselor will review the documentation and provide a status of approve/exclude/reject in the final PPM calculations.

This ticket adds status and reason columns to the DB table for Weight Tickets, and it adds those fields to the models as well. Finally the swagger file is updated with the new fields.

## Files and Folders touched:
* `[migrations/app/migrations_manifest.txt](https://github.com/transcom/mymove/pull/9259/files#diff-3bfbe8b71332fe4dc24d752cb37d3463e9d60e2d6250b20e3c5da7f196040995)`
* `migrations/schema`
* `internal.yaml`
* `ghc.yaml`
* `internalapi/internal/model_to_payload.go`
* `handler/weight_ticket.go`
* `handler/weight_ticket_test.go`
* `swagger-def/definitions`

## Setup to Run Your Code

<details>
<summary>💻 You will need to use 2 separate terminals to test this locally.</summary>

##### Terminal 1

Run the migration if the new migration has not run locally.

```sh
make db_dev_migrate
```

##### Terminal 2

Make sure changes to the yaml are reflected in the autogen files.

```sh
make swagger_generate
```

### Additional steps

1. Check that the DB has the 2 new columns
2. Check that the validations for the updated model pass.

## Verification Steps for Author

These are to be checked by the author.

- [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
- [ ] Request review from a member of a different team.
- [ ] Have the Jira acceptance criteria been met for this change?

## Verification Steps for Reviewers

These are to be checked by a reviewer.

### Backend

- [ ] Code follows the guidelines for [Logging](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide#logging)
- [ ] The requirements listed in [Querying the Database Safely](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide/#querying-the-database-safely) have been satisfied.

#### Any new migrations/schema changes:

- [ ] Follows our guidelines for [Zero-Downtime Deploys](https://transcom.github.io/mymove-docs/docs/dev/contributing/database/Database-Migrations#zero-downtime-migrations)
- [ ] Have been communicated to #g-database
[...]

PR #9312, opened 2022-10-06
MB-13859-weight-ticket-status-and-reason-cleanPR

NamibiaTorres:
## Summary

After the service member submits their closeout documentation, including at least one weight ticket record, the Services Counselor will review the documentation and provide a status of approve/exclude/reject in the final PPM calculations.

This ticket adds status and reason columns to the DB table for Weight Tickets, and it adds those fields to the models as well. Finally the swagger file is updated with the new fields.

## Files and Folders touched:
* `[migrations/app/migrations_manifest.txt](https://github.com/transcom/mymove/pull/9259/files#diff-3bfbe8b71332fe4dc24d752cb37d3463e9d60e2d6250b20e3c5da7f196040995)`
* `migrations/schema`
* `internal.yaml`
* `ghc.yaml`
* `internalapi/internal/model_to_payload.go`
* `handler/weight_ticket.go`
* `handler/weight_ticket_test.go`
* `swagger-def/definitions`

## Setup to Run Your Code

<details>

##### Terminal 1

Run the migration if the new migration has not run locally.

```sh
make db_dev_migrate
```

##### Terminal 2

Make sure changes to the yaml are reflected in the autogen files.

```sh
make swagger_generate
```

### Additional steps

1. Check that the DB has the 2 new columns
2. Check that the validations for the updated model pass.

### Backend


pearl-truss (approved):
LGTM

swagger-def/internal.yaml
```diff
@@ -1022,6 +1022,11 @@ definitions:
       proofOfTrailerOwnershipDocument:
         title: Proof of Trailer Ownership Document
         $ref: '#/definitions/DocumentPayload'
+      status:
+        $ref: 'definitions/PPMDocumentStatus.yaml'
+      reason:
+        description: The reason the services counselor has excluded or rejected the weight ticket.
+        $ref: 'definitions/PPMDocumentStatusReason.yaml'
```
duncan-truss, line 1029: When using a `$ref` to include a definition, swagger ignores any other top level fields like the description you added for status.  You can verify this if you run the `mymove-docs` API documentation locally or look at the swagger-ui docs.  There is another way using `allOf` that you will see but then you will have to use that in multiple places.

I don't think we'll need different descriptions per model so you can just add a description field to both PPMDocumentStatus.yaml and PPMDocumentStatusReason.yaml

Screen Shot 2022-10-06 at 10 00 52 AM

swagger-def/definitions/PPMDocumentStatusReason.yaml
```diff
@@ -0,0 +1,3 @@
+type: string
```
duncan-truss, line 1: This isn't blocking but I don't know that we're getting a lot of value from having a shared string field definition if there's nothing special about the type or properties.

This means in the swagger go payload now we have to use a special type internalmessages.PPMDocumentStatusReason instead of a basic string pointer.

https://github.com/transcom/mymove/pull/9312/files#diff-c2cba06b3937dd4e36cd5b8f7096e17cfc6e56e00a038d11ad1e7f57c51b5d8fR315

duncan-truss (approved):
Database migration and model changes look good, my comments are mostly swagger and documentation callouts.

swagger-def/internal.yaml
NamibiaTorres: Ah ok! I did not realize that. I will add generic descriptions to both those definitions since they are shared.

swagger-def/definitions/PPMDocumentStatusReason.yaml
NamibiaTorres: Hmm, are you saying to remove the shared definition? I got some feedback to add it so now I'm torn.

swagger-def/internal.yaml
NamibiaTorres: @duncan-truss , it looks like I'm not able to add a definition field to a shared definition?: 

```
The swagger spec at "swagger/internal.yaml" is invalid against swagger specification 2.0. see errors:

definitions.PPMDocumentStatusReason.definition in body is a forbidden property


```

NamibiaTorres: Option 1: remove definition so we won't have one for the reason field.
Option 2: remove the shared definition for the reason field, and each objects that has this definition will need:
```
type: string
x-nullable: true
x-omitempty: false
```

duncan-truss: The field name should be `description` not `definition`

swagger-def/definitions/PPMDocumentStatusReason.yaml
duncan-truss: It's just an opinion I think it's okay to leave it.  I'm just pointing out making something shared doesn't always make things simpler/better.

swagger-def/internal.yaml
NamibiaTorres: Well, that explains the error. Thanks Duncan!

PR #9328, opened 2022-10-11
Add usage of status and reason in service object + related tests

ronaktruss:
## Summary

This adds logic to the service object to handle both the status and reason.

This also adds rules to ensure that the customer is unable to modify the status and reason value while ensuring that the office user sets these values to valid ones (ensures reason is required when status is excluded/rejected AND ensures reason is empty when status is approved).

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

To test this you have a few options:

1. You can run the existing tests by running the following. For general information on how to run go test you can refer to [this](https://transcom.github.io/mymove-docs/docs/backend/testing/run-go-tests).
    - `make server_test_setup` to setup the test data
    - `go test ./pkg/services/weight_ticket`to run all of the tests within the folder
    - OR something like `go test ./pkg/services/weight_ticket/ -testify.m TestValidationRules` to run a specific test
1. You can mock out some API calls using postman. Instructions can be found [here](https://transcom.github.io/mymove-docs/docs/tools/postman/intercepting-cookies-for-postman). `weightTicketPPM@ppm.approved` is a sample user that you can use.

### Backend


pkg/testdatagen/make_weight_ticket.go
```diff
@@ -67,6 +67,7 @@ func MakeWeightTicket(db *pop.Connection, assertions Assertions) models.WeightTi
 			FullDocumentID:           fullDocument.ID,
 			FullDocument:             fullDocument,
 			OwnsTrailer:              models.BoolPointer(false),
+			TrailerMeetsCriteria:     models.BoolPointer(false),
```
ronaktruss, line 70: Added because we have a validation on that requires this.

pkg/services/weight_ticket/weight_ticket_updater.go
```diff
@@ -72,6 +70,28 @@ func (f *weightTicketUpdater) UpdateWeightTicket(appCtx appcontext.AppContext, w
...
+	mergedWeightTicket := originalWeightTicket
+
+	mergedWeightTicket.VehicleDescription = services.SetOptionalStringField(weightTicket.VehicleDescription, mergedWeightTicket.VehicleDescription)
+	mergedWeightTicket.EmptyWeight = services.SetNoNilOptionalPoundField(weightTicket.EmptyWeight, mergedWeightTicket.EmptyWeight)
+	mergedWeightTicket.MissingEmptyWeightTicket = services.SetNoNilOptionalBoolField(weightTicket.MissingEmptyWeightTicket, mergedWeightTicket.MissingEmptyWeightTicket)
+	mergedWeightTicket.FullWeight = services.SetNoNilOptionalPoundField(weightTicket.FullWeight, mergedWeightTicket.FullWeight)
+	mergedWeightTicket.MissingFullWeightTicket = services.SetNoNilOptionalBoolField(weightTicket.MissingFullWeightTicket, mergedWeightTicket.MissingFullWeightTicket)
+	mergedWeightTicket.OwnsTrailer = services.SetNoNilOptionalBoolField(weightTicket.OwnsTrailer, mergedWeightTicket.OwnsTrailer)
+	mergedWeightTicket.TrailerMeetsCriteria = services.SetNoNilOptionalBoolField(weightTicket.TrailerMeetsCriteria, mergedWeightTicket.TrailerMeetsCriteria)
+	mergedWeightTicket.Reason = services.SetOptionalStringField(weightTicket.Reason, mergedWeightTicket.Reason)
```
ronaktruss, line 83: If we want to update an existing weightTicket that has been rejected and now is becoming approved we will need to explicitly pass in an empty reason string.

pkg/services/weight_ticket/weight_ticket_updater_test.go
```diff
@@ -228,4 +228,168 @@ func (suite WeightTicketSuite) TestUpdateWeightTicket() {
 		suite.IsType(apperror.InvalidInputError{}, updateErr)
 		suite.Equal("Invalid input found while validating the weight ticket.", updateErr.Error())
 	})
+
+	suite.Run("Status and reason related", func() {
+		suite.Run("successfully", func() {
```
ronaktruss, line 233: When should we be using nested tests? If a single test fails you get failures all the way up.

pkg/services/weight_ticket/weight_ticket_updater.go
ronaktruss: This approach differs from the code that has been implemented on moving_expenses. For moving expenses we automatically set the reason to nil if the status is approved rather than explicitly require you to pass an empty string.

If there are no objections to this, I may make some modifications to moving_expenses to have consistent approaches. 

@duncan-truss Just tagging you since you wrote that original code and since this was brought up during demo while you were out.

duncan-truss: Does this rely on the frontend passing the empty string or will we get the empty string based on how we define the weight ticket API definition and payload to model transformation?

pkg/services/weight_ticket/rules_test.go
```diff
@@ -175,4 +177,248 @@ func (suite *WeightTicketSuite) TestValidationRules() {
...
+						OwnsTrailer:              models.BoolPointer(false),
+						TrailerMeetsCriteria:     models.BoolPointer(false),
+					},
+					existingWeightTicket,
+				)
+				suite.NilOrNoVerrs(err)
+			})
+
+			// Should pass this check, but fails at the verifications in the model
+			suite.Run("invalid status", func() {
```
duncan-truss, line 250: Should we just remove this test then if it's captured in the model validate() and doesn't test anything else?

ronaktruss: Yeah this should be removed. This was meant to be a placeholder until I added an equivalent test on the updater.

pkg/services/weight_ticket/weight_ticket_updater.go
duncan-truss: I think that does probably align with the form reason field being blank when a status and reason doesn't exist yet.

On the case where we are switching from a non-approved to an approved status setting that reason field to a blank string instead of the service working for nil and omitted values is a little inconsistent to me.  I don't know if we want to clear the reason value immediately in the form either so it isn't lost for the user (I think this happens for the TIO because status is updated immediately on radio element click).  So we'd have to specifically remember to clear that reason field in the submit handler and not just pass through the field value. 

I'm fine with it if that's the route you want to go though.  I think the way the TIO updating the status of a service item behaves like moving expenses was implemented and disregards the reason FWIW.

duncan-truss (approved):
Code looks good to me, we can't test the office side of things until there are APIs and handlers really.  Using Postman for the internal API just ignores any status and reason fields because they aren't present in the update payload definition.

pkg/services/weight_ticket/weight_ticket_updater.go
ronaktruss: > (I think this happens for the TIO because status is updated immediately on radio element click).

Oh interesting! Yeah just verified and this does appear to be the behavior.

ronaktruss: Since we have other bits of information on the review page, it likely makes sense to update everything all at once on the click of the next button rather than do it instantaneously like the service items where just the status is being changed. I do wonder though if that causes confusion for the user because we have one review page where we update immediately and another where things only change if you click next.

duncan-truss: I am pretty sure it came up in the user exercise usability testing that the TIO review service items was confusing because participants would select Reject but until they entered a reason and saved it specifically it wouldn't persist.  They would choose Reject and then hit the button for Next I think.  I wouldn't be surprised if that flow changes in behavior.

PR #9341, opened 2022-10-12
MB-14227: independent session managers for each app

ahobson:
## Summary

Instead of having a single session store for for all apps, create separate stores for each. This ensures we can't accidentally have session collisions or accidentally find the wrong session for the wrong app.

In addition, refactor the session managers to use a struct instead of an array.

## Setup to Run Your Code

There should be no visible changes to how the app works. There are new assertions in the session_test that ensures independent session stores.

### Backend


pkg/handlers/apitests.go
```diff
@@ -25,6 +23,29 @@ import (
...
+	PrimeTestHost string = "prime.example.com"
+)
+
+// ApplicationTestServername is a collection of the test servernames
+func ApplicationTestServername() auth.ApplicationServername {
+	return auth.ApplicationServername{
+		MilServername:    MilTestHost,
+		OfficeServername: OfficeTestHost,
+		OrdersServername: OrdersTestHost,
+		AdminServername:  AdminTestHost,
```
mboren, line 45: is `PrimeTestHost` intentionally missing here?

pkg/handlers/apitests.go
```diff
@@ -46,27 +67,16 @@ func NewBaseHandlerTestSuite(sender notifications.NotificationSender, packageNam
...
-	adminSession.Cookie.Name = "admin_session_token"
-
-	officeSession = scs.New()
-	officeSession.Store = store
-	officeSession.Cookie.Name = "office_session_token"
-
-	return [3]*scs.SessionManager{milSession, adminSession, officeSession}
+func setupSessionManagers() auth.AppSessionManagers {
+	return auth.SetupSessionManagers(nil, false, time.Duration(180*time.Second),
+		time.Duration(180*time.Second))
```
mboren, line 79: extreme nitpick: I think wrapping with `time.Duration(...)` here is redundant.
```suggestion
		180*time.Second
```

pkg/handlers/authentication/devlocal.go
```diff
@@ -959,7 +955,26 @@ func createSession(h devlocalAuthHandler, user *models.User, userType string, w
...
+	// ctx, err := sessionManager.Load(r.Context(), session.IDToken)
+	// if err != nil {
+	// 	appCtx.Logger().Error("Cannot load session from context", zap.Error(err))
+	// 	return nil, err
+	// }
+	// _, _, err = sessionManager.Commit(ctx)
+	// if err != nil {
+	// 	appCtx.Logger().Error("Cannot commit session to context", zap.Error(err))
+	// 	return nil, err
+	// }
```
mboren, line 975: Is this still useful? I don't totally understand what this snippet would do for us when debugging.

mboren:
I think this is looking good, I only found a couple of small things I had questions about. I will test on experimental later today

pkg/handlers/apitests.go
ahobson: Thank you!

pkg/handlers/authentication/devlocal.go
ahobson: No, it's not! I've removed it

pkg/handlers/apitests.go
ahobson: 😊 Nope!

mboren:
Tested on experimental by logging into each of the apps with login.gov accounts simultaneously, and logging in and out a few times. Didn't observe any strange behavior with sessions.

mboren (approved):

PR #9343, opened 2022-10-12
MB-14201: forbidden landing page

ahobson:
## Summary

In cases where a person is not authorized for the office app, redirect them to a invalid permissions page

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

https://user-images.githubusercontent.com/2627844/196532373-478b3678-4b5e-4f84-b2ae-4802aebcb0a9.mov

1. Sign into the office app with login.gov. You'll want to use a user account that is not an office user.
2. Verify that you get redirected to the invalid permissions landing page, and don't end up in the office app.
3. Click the log out button
4. You should be redirected to an error page on login.gov 

Logging out is broken right now due to a change to Login.gov's API. This is why we end up at an error page, and aren't properly signed out upon clicking the log out button. 
This will be fixed in a separate PR.

You may also notice that the Admin sign-in page has been updated. It now uses the same styles as the sign in pages for the office and customer apps, with the caveat that the page content is not centered and we're not sure why. We're going to create a separate bug ticket to fix the centering to avoid holding up this work.

Old version of the admin sign in page:
image
New version of the admin sign in page:
image

### Frontend

### Backend


src/pages/InvalidPermissions/InvalidPermissions.jsx
```diff
@@ -0,0 +1,89 @@
+import React from 'react';
+import qs from 'query-string';
+import { bool, shape, string } from 'prop-types';
+import { Button, ButtonGroup } from '@trussworks/react-uswds';
+import { useHistory } from 'react-router-dom';
+
+import '../../styles/office.scss';
```
mboren, line 7: we can tidy up this import

pkg/handlers/authentication/devlocal_test.go
```diff
@@ -1,11 +1,3 @@
-// RA Summary: gosec - errcheck - Unchecked return value
-// RA: Linter flags errcheck error: Ignoring a method's return value can cause the program to overlook unexpected states and conditions.
-// RA: Functions with unchecked return values in the file are used to generate test data for use in the unit test
-// RA: Creation of test data generation for unit test consumption does not present any unexpected states and conditions
-// RA Developer Status: Mitigated
-// RA Validator Status: Mitigated
-// RA Modified Severity: N/A
-// nolint:errcheck
```
mboren, line 8: love to see this annotation removed

src/scenes/MyMove/index.jsx
```diff
@@ -237,6 +238,7 @@ export class CustomerApp extends Component {
                       <h2>We are experiencing an internal server error</h2>
                     </div>
                   </Route>
+                  <Route exact path="/invalid-permissions" component={InvalidPermissions} />
```
mboren, line 241: what's the difference between `/forbidden` and `/invalid-permissions`?

src/scenes/SystemAdmin/index.jsx
```diff
@@ -25,24 +25,19 @@ class AdminWrapper extends Component {
...
+    return (
+      <div id="app-root">
+        <CUIHeader />
+        <Switch>
+          {/* no auth */}
+          <Route path="/sign-in" component={SignIn} />
+          <Route path="/invalid-permissions" component={InvalidPermissions} />
+          <Route path="/" component={defaultComponent} />)
+        </Switch>
+      </div>
```
mboren, line 38: this seems to "break" the signin page's styling. But I'm not sure if it's a bug or a feature.
On `master`, our styling for this page is pretty different than the rest of the app because it doesn't inherit most of our styles. So it's got serif fonts and stuff.
image

After these changes, we end up including additional styles, which mostly makes the sign in look better, but also bumps it off center. 
image

mboren: Oh I misunderstood the issue here, it's not so much related to the markup, this PR just imports a different `SignIn` component. Which probably makes more sense to use here if we can, rather than having a unique sign in page for admin. Not sure whether that is in-scope for this work though.

mboren: I've dug into this some more and can't figure out why this ends up off-center, when it seems to be using essentially the same stuff as the sign in pages for the other apps.

mboren:
I think this works pretty well, aside from a styling issue. I haven't come up with a better way to get to the invalid permissions page in the admin app. Using routes seems straightforward and, based on the react-admin docs, it doesn't seem like we're doing anything too crazy.

src/scenes/MyMove/index.jsx
ahobson: Excellent question! I'm not sure!

You can see that `/forbidden` just displays a relatively opaque error message. I could see that working for when the user is signed into the app, but tried to hit an endpoint they aren't allowed to see.

But I was thinking we needed a page that indicated the permissions were wrong for the entire app and allowed the user to sign out. So maybe the name isn't right. Maybe it's more like "unauthorized for application" or something like that?

src/scenes/SystemAdmin/index.jsx
mboren: Yeah I think it's fine.

mboren (approved):

PR #9344, opened 2022-10-12
MB-8674 Move transaction logic to service object

mlm55:
## Summary
This PR moves DB saves/transaction logic from the handler level (pkg/handlers/internalapi/moves.go) to service objects. For the submit handler, saves were moved to the move router. ~~For the patch to update `SelectedMoveType`, a new method to the `moveTaskOrderUpdater`.~~

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Make sure `make server_test` and `make e2e_test` run cleanly

### Frontend

### Backend

### DIFF
--- a/pkg/handlers/internalapi/moves.go
+++ b/pkg/handlers/internalapi/moves.go
@@ -8,7 +8,9 @@ import (
 
 	"github.com/go-openapi/runtime/middleware"
 	"github.com/go-openapi/swag"
+	"github.com/gobuffalo/validate/v3"
 	"github.com/gofrs/uuid"
+	"github.com/pkg/errors"
 	"go.uber.org/zap"
 
 	"github.com/transcom/mymove/pkg/appcontext"
@@ -103,7 +105,6 @@ func (h ShowMoveHandler) Handle(params moveop.ShowMoveParams) middleware.Respond
 // PatchMoveHandler patches a move via PATCH /moves/{moveId}
 type PatchMoveHandler struct {
 	handlers.HandlerConfig
-	services.MoveTaskOrderUpdater
 }
 
 // Handle ... patches a Move from a request payload
@@ -130,12 +131,13 @@ func (h PatchMoveHandler) Handle(params moveop.PatchMoveParams) middleware.Respo
 
 			if newSelectedMoveType != nil {
 				stringSelectedMoveType := models.SelectedMoveType(*newSelectedMoveType)
-				move, err = h.UpdateSelectedMoveType(appCtx, moveID, &stringSelectedMoveType)
-				if err != nil {
-					return handlers.ResponseForError(logger, err), err
-				}
+				move.SelectedMoveType = &stringSelectedMoveType
 			}
 
+			verrs, err := appCtx.DB().ValidateAndUpdate(move)
+			if err != nil || verrs.HasAny() {
+				return handlers.ResponseForVErrors(logger, verrs, err), err
+			}
 			movePayload, err := payloadForMoveModel(h.FileStorer(), orders, *move)
 			if err != nil {
 				return handlers.ResponseForError(logger, err), err
@@ -354,6 +356,18 @@ func (h SubmitAmendedOrdersHandler) Handle(params moveop.SubmitAmendedOrdersPara
 			err = h.MoveRouter.Submit(appCtx, move, models.SignedCertification{})
 			if err != nil {
 				return handlers.ResponseForError(logger, err), err
+			}
+
+			responseVErrors := validate.NewErrors()
+			var responseError error
+
+			if verrs, saveErr := appCtx.DB().ValidateAndSave(move); verrs.HasAny() || saveErr != nil {
+				responseVErrors.Append(verrs)
+				responseError = errors.Wrap(saveErr, "Error Saving Move")
+			}
+
+			if responseVErrors.HasAny() {
+				return handlers.ResponseForVErrors(logger, responseVErrors, responseError), responseError
 			}
 
 			movePayload, err := payloadForMoveModel(h.FileStorer(), move.Orders, *move)
--- a/pkg/handlers/internalapi/moves.go
+++ b/pkg/handlers/internalapi/moves.go
@@ -13,7 +13,6 @@ import (
 
 	"github.com/transcom/mymove/pkg/appcontext"
 	"github.com/transcom/mymove/pkg/assets"
-	certop "github.com/transcom/mymove/pkg/gen/internalapi/internaloperations/certification"
 	moveop "github.com/transcom/mymove/pkg/gen/internalapi/internaloperations/moves"
 	"github.com/transcom/mymove/pkg/gen/internalmessages"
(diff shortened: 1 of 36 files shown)


StuTruss (approved):

pkg/handlers/internalapi/moves.go
```diff
@@ -132,13 +131,12 @@ func (h PatchMoveHandler) Handle(params moveop.PatchMoveParams) middleware.Respo
 
 			if newSelectedMoveType != nil {
 				stringSelectedMoveType := models.SelectedMoveType(*newSelectedMoveType)
-				move.SelectedMoveType = &stringSelectedMoveType
+				move, err = h.UpdateSelectedMoveType(appCtx, moveID, &stringSelectedMoveType)
+				if err != nil {
+					return handlers.ResponseForError(logger, err), err
+				}
```
felipe-lee, line 137: **thought**: Instead of updating this handler, can we get rid of it entirely maybe? `selectedMoveType` isn't really something that seems to be used anymore, nor make sense with how the system works now. [I brought this up in slack in february](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1643944479746139) (my old user account was deactivated so that's why it says deactivated user 😞). The [access code stuff mentioned in the thread was removed](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1650399645486159). There are still uses in the FE code, but none that really need it as far as I can tell. I don't think removing this should be part of this pr probably because it would be a sizeable PR on its own, but I do wonder whether that could be done instead of updating it and adding the function to the move task order updater.

pkg/services/move/move_router.go
```diff
@@ -45,12 +45,27 @@ func (router moveRouter) Submit(appCtx appcontext.AppContext, move *models.Move)
...
+				return err
+			}
+
+			if newSignedCertification != (models.SignedCertification{}) {
+				verrs, err = txnAppCtx.DB().ValidateAndCreate(&newSignedCertification)
+				if err != nil || verrs.HasAny() {
+					txnAppCtx.Logger().Error("error saving signed certification: %w", zap.Error(err))
+					return err
+				}
+			}
```
felipe-lee, line 63: **question**: Can we sync on this tomorrow or next week? I've got a [ticket this sprint to set up a signed certs api for ppm shipments](https://dp3.atlassian.net/browse/MB-14263) because we can't re-use what already exists, but as part of that, I plan on trying to create a service object for signed certs and was thinking of seeing if I could update the existing code to use that service object instead of having the save logic in multiple places. I think we could also each do our work separately and then have a followup PR that syncs up how signed certs are created.

pkg/handlers/internalapi/moves.go
```diff
@@ -166,19 +164,32 @@ func (h SubmitMoveHandler) Handle(params moveop.SubmitMoveForApprovalParams) mid
...
 			logger := appCtx.Logger().With(zap.String("moveLocator", move.Locator))
-			err = h.MoveRouter.Submit(appCtx, move)
-			if err != nil {
-				return handlers.ResponseForError(logger, err), err
-			}
 
 			certificateParams := certop.NewCreateSignedCertificationParams()
 			certificateParams.CreateSignedCertificationPayload = params.SubmitMoveForApprovalPayload.Certificate
 			certificateParams.HTTPRequest = params.HTTPRequest
 			certificateParams.MoveID = params.MoveID
```
felipe-lee, line 171: **nit**: I know you didn't write this part, but I don't really think it's buying us anything since we don't use the `certificateParams` for anything that couldn't also be achieved from accessing the values in `params` instead or if we want a shorter variable, then doing something like 

```go
certificate := params.SubmitMoveForApprovalPayload.Certificate
```
 
and using `certificate` below instead of `certificateParams.CreateSignedCertificationPayload`

pkg/handlers/internalapi/moves.go
```diff
@@ -166,19 +164,32 @@ func (h SubmitMoveHandler) Handle(params moveop.SubmitMoveForApprovalParams) mid
...
+				SubmittingUserID:         appCtx.Session().UserID,
+				CertificationText:        *certificateParams.CreateSignedCertificationPayload.CertificationText,
+				Signature:                *certificateParams.CreateSignedCertificationPayload.Signature,
+				Date:                     date,
+			}
+
+			if certificateParams.CreateSignedCertificationPayload.PersonallyProcuredMoveID != nil {
+				ppmID := uuid.FromStringOrNil(certificateParams.CreateSignedCertificationPayload.PersonallyProcuredMoveID.String())
+				newSignedCertification.PersonallyProcuredMoveID = &ppmID
+			}
```
felipe-lee, line 188: **suggestion**: can you move this logic to the `payload_to_model` file? It's the type of stuff that file normally handles.

pkg/handlers/internalapi/moves.go
```diff
@@ -197,84 +208,6 @@ func (h SubmitMoveHandler) Handle(params moveop.SubmitMoveForApprovalParams) mid
 		})
 }
 
-// SaveMoveDependencies safely saves a Move status, ppmShipment status, mtoShipment status, orders statuses, signed certificate,
-// and shipment GBLOCs.
-func (h SubmitMoveHandler) saveMoveDependencies(appCtx appcontext.AppContext, move *models.Move, certificateParams certop.CreateSignedCertificationParams, userID uuid.UUID) (*validate.Errors, error) {
```
felipe-lee, line 202: **praise**: I know the point of the ticket _is_ to get rid of this logic being here, but it's still really exciting to see it removed, thanks for doing this! 🥳

pkg/services/move.go
```diff
@@ -45,7 +45,7 @@ type MoveRouter interface {
 	Cancel(appCtx appcontext.AppContext, reason string, move *models.Move) error
 	CompleteServiceCounseling(appCtx appcontext.AppContext, move *models.Move) error
 	SendToOfficeUser(appCtx appcontext.AppContext, move *models.Move) error
-	Submit(appCtx appcontext.AppContext, move *models.Move) error
+	Submit(appCtx appcontext.AppContext, move *models.Move, newSignedCertification models.SignedCertification) error
```
felipe-lee, line 48: **thought**: I think this is one thing that syncing could be good for. If we do switch to using a service object for signed certs, would we need to pass it in to the Submit function here? We can pass the service object as a dependency when the move router is set up, but I guess we would still need to pass the signed cert info somewhere, so we could still need this here 🤔

pkg/handlers/internalapi/moves.go
mlm55: that makes sense. i'll remove the updates and can write up a ticket to actually remove the code

shimonacarvalho (approved):
Looks good to me. Test coverage looks good too.

PR #9345, opened 2022-10-13
MB-13773 add update and endpoint definitions to ghc api for ppm documents

felipe-lee:
## Summary

This PR adds endpoint definitions to the GHC API for updating PPM shipment weight tickets, moving expenses, and pro-gear weight tickets.

This builds on some pre-work done in #9323

Some things to note:

* I tried to make things a bit more re-usable across internal and ghc, so to that end:
    * [I moved a few PPM shipment-related parameters to be shared](https://github.com/transcom/mymove/commit/6750ead0d98989592fae5bda2bb0e950d3aa97ef)
    * Added a shared definition of pro-gear weight tickets for responses.
        * Weight tickets and moving expenses were already being shared, but we haven't done the pro-gear API work for the internal API, so there wasn't an existing thing that could be shared.
    * Tried aligning the descriptions of the fields across the shared response definitions as well as the separate update definitions.
* [I added pro gear weight tickets to ppm shipment definition](https://github.com/transcom/mymove/pull/9345/commits/0a7c5d38ecf4441c657c5f639cd22c11495307b7)
* I had to create a second shared copy of `PPMDocumentStatus`. One has the `x-nullable` and `x-omitempty` keys (`OmittablePPMDocumentStatus`), while the other does not (`PPMDocumentStatus`). I tried seeing about having only one definition (without those properties) and using `allOf` to compose the field to have both the base things plus the added properties, but that doesn't work. It'll parse it and generate, but it generates the incorrect code. I'm thinking this is likely a bug in `go-swagger`, but I didn't look too far into it because I couldn't really find many things talking about this.


swagger-def/ghc.yaml
```diff
@@ -3981,6 +4090,93 @@ definitions:
         type: integer
         format: cents
         x-nullable: true
+  UpdateWeightTicket:
```
ronaktruss, line 4093: For these update objects we may not need to include description, because based on the [designs](https://www.figma.com/file/5AhuEkaH3bJVrtffU3wp22/Milmove-Office-App?node-id=1279%3A23184) they seem like they won't be updatable. The same may be true for `movingExpenseType`.

ronaktruss:
> I had to create a shared copy of PPMDocumentStatus that doesn't have the x-nullable and x-omitempty keys. I tried seeing about taking them off the other definition and then using allOf to compose the properties, but that doesn't work. It'll parse it and generate, but it generates the incorrect code. I'm thinking this is likely a bug in go-swagger, but I didn't look too far into it because I couldn't really find many things talking about this.

If we aren't actually extending the the definitions using the allOf property, I think the names `BasePPMDocumentStatus` and `PPMDocumentStatus` are a bit confusing/misleading. Would a name indicating their use case (office/customer) or something describing what the differences are (requiring the status / it not be a pointer) be more helpful?

swagger-def/ghc.yaml
felipe-lee: Thanks for catching this. [I've asked to confirm](https://ustcdp3.slack.com/archives/C0250Q8469K/p1666047966577759), but I think you're right and I'll update to remove the fields we don't need.

ronaktruss (approved):
This LGTM! We may also want to take a look at `w2Address` since its using a ref and x-nullable which to my understanding doesn't work.

swagger-def/definitions/ProGearWeightTicket.yaml
```diff
@@ -0,0 +1,106 @@
...
+  emptyDocumentId:
+    description: The ID of the document that is associated with the user uploads containing the empty vehicle weight.
+    type: string
+    format: uuid
+    example: c56a4180-65aa-42ec-a945-5fd21dec0538
+    readOnly: true
+  emptyDocument:
+    allOf:
+      - description: Document that is associated with the user uploads containing the empty vehicle weight.
+      - $ref: 'Document.yaml'
```
duncan-truss, line 54: Nice, I think we can do this elsewhere for other shared definitions to not use a generic description.

swagger-def/definitions/ProGearWeightTicket.yaml
```diff
@@ -0,0 +1,106 @@
...
+    example: c56a4180-65aa-42ec-a945-5fd21dec0538
+    readOnly: true
+  constructedWeightDocument:
+    allOf:
+      - description: Document that is associated with the user uploads containing the constructed weight.
+      - $ref: 'Document.yaml'
+  status:
+    $ref: 'OmittablePPMDocumentStatus.yaml'
+  reason:
+    $ref: 'PPMDocumentStatusReason.yaml'
```
duncan-truss, line 90: This is nitpicky but having a shared definition for basic string type is a bit pointless.  It just makes it harder to work with in the generated type, we need a special pointer instead of a string.

swagger-def/ghc.yaml
```diff
@@ -1265,6 +1266,114 @@ paths:
       summary: Denies a SIT extension
       x-permissions:
         - update.SITExtension
+  /ppm-shipments/{ppmShipmentId}/weight-ticket/{weightTicketId}:
+    parameters:
+      - $ref: 'parameters/ppmShipmentId.yaml'
+      - $ref: 'parameters/weightTicketId.yaml'
```
duncan-truss, line 1272: cool I always forget these can have shared definitions too

swagger-def/ghc.yaml
```diff
@@ -3977,6 +4086,75 @@ definitions:
...
+      ownsTrailer:
+        description: Indicates if the customer used a trailer they own for the move.
+        type: boolean
+      trailerMeetsCriteria:
+        description: Indicates if the trailer that the customer used meets all the criteria to be claimable.
+        type: boolean
+      status:
+        $ref: 'definitions/PPMDocumentStatus.yaml'
+      reason:
+        description: The reason the services counselor has excluded or rejected the item.
```
duncan-truss, line 4109: I think elsewhere you used the shared definition for this field, I would consider getting rid of it but just pointing out for consistency.

swagger-def/ghc.yaml
```diff
@@ -3977,6 +4086,75 @@ definitions:
...
+        example: '2022-04-26'
+        format: date
+      sitEndDate:
+        description: The date the shipment exited storage, applicable for the `STORAGE` movingExpenseType only
+        type: string
+        example: '2018-05-26'
+        format: date
+      status:
+        $ref: 'definitions/PPMDocumentStatus.yaml'
+      reason:
```
duncan-truss, line 4129: Ditto, shared definition was used elsewhere

swagger-def/definitions/ProGearWeightTicket.yaml
felipe-lee: yeah I agree, I was mainly re-using what we had, but I don't know that it's really buying us much for the tradeoff of having to mess with types more.

swagger-def/ghc.yaml
felipe-lee: Same, mainly noticed that ghc did this more and though it'd be good to do for things that we have in both files.

swagger-def/ghc.yaml
felipe-lee: So the reason for the difference is that the shared definition is `x-nullable: true` and `x-omitempty: false`, which means you'd always have to include it as part of the request body. That's why I didn't re-use that shared definition here since we don't usually require all fields to be sent with patch requests.

PR #9351, opened 2022-10-13
[MB-13333] Add backend file upload support for PPM document spreadsheets

duncan-truss:
## Summary

PPMs have added a new requirement that we allow spreadsheet files to be uploaded when a weight ticket or pro-gear closeout document is used instead of an image or pdf of the supporting proof.

We made the frontend client changes to our upload component previously but our backend code wasn't yet modified to allow and detect spreadsheet file types.

I added content types for `.xls` and `.xlsx` on the backend to match what we have on the client side but our use of the standard library [DetectContentType](https://pkg.go.dev/net/http#DetectContentType) would always return `application/octet-stream` for excel uploads (or other non-supported file types).  This is because the standard isn't supposed to be a robust MIME type check but is more used for limited API request purposes.  I went looking and found a fuller featured library and found a good option https://github.com/gabriel-vasile/mimetype.

The other problem I was trying to address once files were being accurately identified, was to have our PPM specific pages use a new endpoint separate from the one being shared with uploading orders for the customer (that should only allow images and PDFs).  While it was nice that we could reuse this endpoint to this point, we would have made it more complex by trying to verify which page was making the request and manage the different business rules.

[Transcom Weight Estimator download](https://www.ustranscom.mil/dp3/weightestimator.cfm)

https://dp3.atlassian.net/wiki/spaces/MT/pages/1470955567/How+to+test+storing+data+in+S3+locally

## Setup to Run Your Code

To test using local storage:

`make server_generate && make server_run`

`make client_run`

To test using AWS S3 storage:

Uncomment the line in your .envrc file:
`export STORAGE_BACKEND=s3`

`direnv reload`
`aws-vault exec transcom-gov-dev -- make server_run`

To login to AWS gov-dev and see the S3 uploads you can run `aws-vault login transcom-gov-dev` and select the S3 service.  You'll find the uploads under the `user/` prefix for the user you logged into the customer app as.

`make client_run`

### Additional steps

1. Login to the customer app with a PPM with an existing weight ticket or at least a PPM ready for closeout like `weightTicketPPM@ppm.approved`
2. Proceed to choose Upload PPM Documents from the home screen
3. Add a new weight ticket or edit an existing weight ticket of the PPM
4. In either of the empty weight or full weight upload fields you should be able to select one of the test files [Weight Estimator.xls](https://github.com/transcom/mymove/blob/master/pkg/testdatagen/testdata/Weight%20Estimator.xls) or [Weight Estimator.xlsx](https://github.com/transcom/mymove/blob/master/pkg/testdatagen/testdata/Weight%20Estimator.xlsx)
5. It's also good to verify that behavior hasn't changed for uploading orders docs or image and pdf documents since they are using the new content detection library.

### Frontend

### Backend

## Screenshots

**Tested on Experimental**

Screen Shot 2022-10-17 at 1 28 12 PM

**Inspect the content type saved to the uploads database table**
Screen Shot 2022-10-17 at 5 10 22 PM


pkg/handlers/internalapi/uploads.go
```diff
@@ -150,3 +157,73 @@ func (h DeleteUploadsHandler) Handle(params uploadop.DeleteUploadsParams) middle
 			return uploadop.NewDeleteUploadsNoContent(), nil
 		})
 }
+
+func (h CreatePPMUploadHandler) Handle(params ppmop.CreatePPMUploadParams) middleware.Responder {
```
duncan-truss, line 161: This was mostly copied from our existing handler so it has some legacy cruft in it still that I couldn't totally address if the work was focused on the handler only.

pkg/services/ppmshipment/ppm_shipment_fetcher.go
```diff
@@ -0,0 +1,86 @@
...
+		switch err {
+		case sql.ErrNoRows: // not ready to return an error unless the document is also not part of pro gear or expenses
+		default:
+			return apperror.NewQueryError("PPMShipment", err, "")
+		}
+	} else {
+		return nil
+	}
+
+	err = appCtx.DB().Q().
```
duncan-truss, line 56: We'll need to remember to update this query when pro-gear columns get consolidated.

ronaktruss: Can we add some documentation either in the code to link to fixing that (I think we should be fixing that in [this story](https://dp3.atlassian.net/browse/MB-14165)) or add something in a ticket pointing to this code?

ronaktruss (commented):
Nice work! Reading through the code, most of this seems good and nothing caught my eye as a 🚩.

Are you planning on coordinating with Scott to also update the movingExpenses page with this new api call/updating the associated cypress tests? Or is that additional work that needs to be captured somewhere?

pkg/services/ppmshipment/ppm_shipment_fetcher.go
duncan-truss: I just pushed up some some tests for the new ppm shipment fetcher that should break on new changes.

duncan-truss:
> Nice work! Reading through the code, most of this seems good and nothing caught my eye as a 🚩.
> 
> Are you planning on coordinating with Scott to also update the movingExpenses page with this new api call/updating the associated cypress tests? Or is that additional work that needs to be captured somewhere?

That's a good question, I don't see any uploads happening in the current moving expenses Cypress tests.  For expenses we don't allow spreadsheet files on the client validation like we do for weight tickets or pro gear.  There is an upcoming story for expenses Cypress testing where we can make sure that gets handled. https://dp3.atlassian.net/browse/MB-12540

pkg/services/ppmshipment/ppm_shipment_fetcher.go
```diff
@@ -0,0 +1,64 @@
...
+		First(&movingExpense)
+
+	if err != nil {
+		switch err {
+		case sql.ErrNoRows:
+			return apperror.NewNotFoundError(documentID, "document does not exist for the given shipment")
+		default:
+			return apperror.NewQueryError("PPMShipment", err, "")
+		}
+	}
```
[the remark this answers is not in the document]
duncan-truss: I think we'll want to keep this for authorization purposes in general, I wanted the one endpoint to service all uploads for PPMs for simplicity.  We can expand this to return what model the document belongs to and then apply specific file allowlists to the uploader service.  Because that expense and pro-gear work is being done in parallel still I think we should keep it how it is for now to account for any PPM document.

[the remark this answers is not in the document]
ronaktruss: So the plan is to utilize the new api for all of the pages, right? If so do we want to add the change to call that API in this story or were you indicating that we can add that in the cypress test ticket in your [earlier comment](https://github.com/transcom/mymove/pull/9351#issuecomment-1283003385).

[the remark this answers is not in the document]
duncan-truss: I would rather not add something that is integration work with Scott's PR getting merged at the same time.  It's fine for it to be on the old uploads endpoint initially.

ronaktruss (approved):
LGTM. The work to update the api call to use the new one introduced on this page for moving expenses is documented in https://dp3.atlassian.net/browse/MB-12540 and work to confirm if the existing cypress tests work for constructed weights is still pending. (Duncan is planning on looking into this, but if this requires additional effort won't be done as part of this story)

PR #9352, opened 2022-10-13
[MB-14038] Manage go-swagger in a better way

reggieriser:
## Summary

[Due to some inconsistencies we've experienced](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1663192043851619) with [go-swagger](https://github.com/go-swagger/go-swagger) depending on how it's installed, this PR uses a `go-swagger` that is installed/managed locally via `go.mod` and `tools.go` as opposed to a version installed in a docker container.

A few additional notes:
- this updates us from `go-swagger` 0.30.2 to 0.30.3 (but that resulted in no changes to generated code)
- see this [circleci-docker PR](https://github.com/transcom/circleci-docker/pull/311) where I remove go-swagger entirely from the docker container.  Note that I'm using the hashes from the PR while we test this, but will switch to the master branch hashes once we merge.
- see inline comments for additional context

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

- `make clean server_generate` -- see if things run as expected and if any local mods exist
- `which swagger` -- to see which go-swagger install is preferred (in case you have multiple)
- `swagger version` -- to see what version of go-swagger is installed (this PR upgrades us to 0.30.3, which doesn't appear to lead to any generated file changes) 


.circleci/config.yml
```diff
@@ -673,16 +673,20 @@ jobs:
...
       - run: scripts/check-generated-code go.sum
+      - run:
+          name: Install go-swagger
+          command: make bin/swagger
       - save_cache:
-          key: go-mod-sources-v3-{{ checksum "go.sum" }}-{{ checksum "scripts/check-go-version" }}
+          key: go-mod-sources-v4-{{ checksum "go.sum" }}-{{ checksum ".go-version" }}
           paths:
             - '~/go'
+            - '~/transcom/mymove/bin/swagger'
```
reggieriser, line 689: Hoping to build go-swagger once and cache it for future jobs.

.pre-commit-config.yaml
```diff
@@ -120,6 +120,7 @@ repos:
         exclude: swagger-def/* # These are partial swagger files that are compiled into the ones in swagger/*.
                                # They will be checked unless explicitly excluded, but they will fail.
                                # By only validating the compiled files, we are effectively checking these files as well.
+        require_serial: true
```
reggieriser, line 123: I was having problems with failures in CircleCI (like it couldn't find the executable on one of the runs) where it looked like this script was being executed twice, each with a set of filenames.  I saw some people mention [require_serial](https://pre-commit.com/#hooks-require_serial) as a possible fix and that seems to be working so far.

Dockerfile.reviewapp
```diff
@@ -39,8 +39,6 @@ RUN mkdir /build/src
 
 RUN set -x \
   && make bin/rds-ca-2019-root.pem \
-  && rm -f pkg/assets/assets.go && make pkg/assets/assets.go \
-  && scripts/gen-server \
```
reggieriser, line 43: After talking with Drew, we determined that there's no real reason for the review app to be building the generated files since they are checked into source control already (we do want to check them in CI, though, to make sure the latest have been checked in).  This keeps me from having to build the `go-swagger` executable here.

pkg/tools/tools.go
```diff
@@ -18,4 +18,6 @@ import (
 	_ "github.com/vektra/mockery/v2"
 	// Possible replacement for go-junit-report
 	_ "gotest.tools/gotestsum"
+	// Install for go-swagger code generation
+	_ "github.com/go-swagger/go-swagger/cmd/swagger"
```
reggieriser, line 22: This will keep `go-swagger` in our `go.mod` and should mean dependabot can flag updates for us.

scripts/gen-server
```diff
@@ -2,19 +2,11 @@
...
-  echo "Using local swagger"
-else
-  # Locally we need to download the docker image to get the binary
-  image=milmove/circleci-docker:milmove-app-dd61e0303f06a916ea1ce1454ee152f563189e42
-  docker pull -q "${image}"
+# Check swagger version
+SWAGGER_VERSION=0.30.3
+if ! swagger version 2>/dev/null | grep -q -E "v?${SWAGGER_VERSION}"; then
+  echo "ERROR: Please install go-swagger version ${SWAGGER_VERSION}"
+  exit 1
```
reggieriser, line 9: I wanted some way to make sure that folks were running with the right version of the `go-swagger` executable locally.  I do this check in two files in this PR.  Not crazy about having an explicit version number, so open to other ideas on how to better handle it.

ahobson: Hmm. Since we know the full path to the swagger executable we want, could we use that? e.g. below explicitly use `bin/swagger` instead of relying on the `PATH` ?

Makefile
```diff
@@ -221,6 +221,10 @@ bin/gotestsum: .check_go_version.stamp .check_gopath.stamp pkg/tools/tools.go
 bin/mockery: .check_go_version.stamp .check_gopath.stamp pkg/tools/tools.go
 	go build -o bin/mockery github.com/vektra/mockery/v2
 
+# No static linking / $(LDFLAGS) because swagger is only used for code generation
+bin/swagger: .check_go_version.stamp .check_gopath.stamp pkg/tools/tools.go
```
ahobson, line 225: I'm very much in nitpick territory, but ...

If the version changes in `go.mod`, none of those files would change and the old version of bin/swagger would be used. This is actually true for all our binaries. I wonder if we should make this change:

```
...
.PHONY: check_gopath
check_gopath: .check_gopath.stamp ## Check that $GOPATH exists in $PATH
.check_gopath.stamp: scripts/check-gopath go.sum
...
```

That way all of our binaries will get rebuilt if the dependencies change

ahobson (approved):
I added some nitpick suggestions. If you feel they are valuable and have the time to incorporate them, great! But none of them are blocking this great change. Thank you

Makefile
ahobson: Here's another possibly crazy idea: what if we always used `go run github.com/go-swagger/go-swagger/cmd/swagger` ? That would always use the version in `go.mod`, we wouldn't have to worry about stale compiled versions or other versions of the `swagger` binary in the `PATH`. As I understand it, `go run` [automatically uses a cache](https://github.com/golang/go/issues/33468#issuecomment-617297504) and should be plenty fast.

scripts/gen-server
reggieriser: That could help isolate it to just the executable in `bin/swagger`, so I may try that.  But I was actually thinking more about the case where the `bin/swagger` executable in a local dev environment is old since I don't think it's guaranteed to be rebuilt right now when the version changes.  Perhaps I could make the make target `bin/swagger` depend on `go.mod` since that's where the version number is updated, although that file has a good bit of churn and will trigger rebuilds pretty often.

.pre-commit-config.yaml
reggieriser: Good point -- I'll add in a comment.

.circleci/config.yml
```diff
@@ -747,7 +751,7 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - go-mod-sources-v3-{{ checksum "go.sum" }}-{{ checksum "scripts/check-go-version" }}}
```
reggieriser, line 750: Just a note here that the old key had an extra curly brace at the end which meant that this cache *never* resolved as far as I can tell.  I've fixed that.

Makefile
reggieriser: I like your idea of the dependency on `go.sum` so I incorporated that and took out the version checks in the scripts.  I'm hoping this means that a dependabot PR could automatically be created now for a new version of go-swagger and can be merged without additional changes if there are no generated code changes, or fail if CI detects changes to generated code (which should require manual review/resolution).

reggieriser:
> I added some nitpick suggestions. If you feel they are valuable and have the time to incorporate them, great!

@ahobson, thanks for the review and good suggestions.  I made a few changes -- see if you think things are still OK.

Ryan-Koch (approved):

ahobson:
>  I made a few changes -- see if you think things are still OK.

Fabulous! Let's get this sucker merged!

PR #9361, opened 2022-10-14
MB-12744 QAE/CSR Search Form Fix and Styling Updates

PDickeyTruss:
## Summary

This PR updates the styling/html related to the QAE/CSR move search to be consistent every time a users visits the page. Previously going back to the search page after the initial viewing resulted in funky styling.

While investigating this styling issue I noticed a few parts of the search page/form that differed from the designs. Since I was tweaking styling anyways it felt natural to fix these design discrepancies as well. 

Designs: https://www.figma.com/file/5AhuEkaH3bJVrtffU3wp22/Milmove-Office-App?node-id=227%3A10746

## What does this PR change?
- All visible changes in this PR apply only to the QAE/CSR Search/Results page (home page when logged in as a qaecsr user).
- Most changes are padding/margin updates (compare w/ designs: https://www.figma.com/file/5AhuEkaH3bJVrtffU3wp22/Milmove-Office-App?node-id=227%3A10746). 
- Changed the behavior of 'no search results' a little bit (empty table -> no results message).
- Swapped out a few components with other components/html elements that came with some styling for free and generally tried to simplify the styling as much as possible. 
- 'Fixed' the weird styling behavior on page revisit by tweaking some existing styles, removing `GridContainer`, and in a few cases using `!important` to force a style to not be overridden.
-  Added the ability to pass a `button` prop to a `textField` to allow buttons (or any html/component) to be displayed after a `textField` but still inside of its `formGroup`.  Having this button outside the `formGroup` causes a number of pain points when getting the validation warning to look correct, this was the most elegant way avoid that pain I could think of. 

## Validation/Testing
- [ ] The QAE/CSR move page should function as before (search input/results behavior, validation, etc)
- [ ] Searches that return no results show 'No results found.'
- [ ] The search form looks the same after clicking on a search result and going back to the search form.
- [ ] The search form and results match the designs.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

## Video
https://user-images.githubusercontent.com/62263329/195916347-ed962318-acae-4ba2-9fff-92ad83a153ef.mov

### DIFF
--- a/cypress/integration/office/qaecsr/moveSearchFlows.js
+++ b/cypress/integration/office/qaecsr/moveSearchFlows.js
@@ -77,7 +77,7 @@ describe('QAE/CSR Move Search', () => {
     // Verify results table contents
     cy.get('tbody > tr').as('results');
     cy.get('@results').should('have.length', 1); // Page size of 20 in table
-    cy.get('@results').first().contains(name); // Should have leo in name
+    cy.get('@results').first().contains(name); // Should have QAECSRTestFirst in name
 
     // Click result to navigate to move details page
     cy.get('@results').first().click();
--- a/cypress/integration/office/qaecsr/moveSearchFlows.js
+++ b/cypress/integration/office/qaecsr/moveSearchFlows.js
@@ -64,7 +64,7 @@ describe('QAE/CSR Move Search', () => {
   });
 
   it('is able to search by customer name', () => {
-    const name = 'Leo';
+    const name = 'QAECSRTestFirst';
 
     // Type dodID into search bar and select DOD ID as search type
     cy.get('[type="radio"]').check('customerName', { force: true });
@@ -76,8 +76,8 @@ describe('QAE/CSR Move Search', () => {
 
     // Verify results table contents
     cy.get('tbody > tr').as('results');
-    cy.get('@results').should('have.length', 20); // Page size of 20 in table
-    cy.get('@results').first().contains(name);
+    cy.get('@results').should('have.length', 1); // Page size of 20 in table
+    cy.get('@results').first().contains(name); // Should have leo in name
 
     // Click result to navigate to move details page
     cy.get('@results').first().click();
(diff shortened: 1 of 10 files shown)


cypress/integration/office/qaecsr/moveSearchFlows.js
```diff
@@ -64,7 +64,7 @@ describe('QAE/CSR Move Search', () => {
   });
 
   it('is able to search by customer name', () => {
-    const name = 'Leo';
+    const name = 'QAECSRTestFirst';
```
PDickeyTruss, line 67: I saw a cypress tests failure in circle recently and figured I would fix it while already in the file making updates for this PR. Previously, this test just clicks the first result for the search 'leo', which returns tons of paged results that could be in all kinds of states depending on what other tests had already run. This just makes sure we are searching for a name that will return a single known result to click on rather than almost all of them as with 'leo'.

src/components/MoveSearchForm/MoveSearchForm.module.scss
```diff
@@ -5,38 +5,49 @@
 .MoveSearchForm {
   @include u-padding-y(3);
   @include u-padding-x(4);
-  @include u-margin-x(1);
   @include u-bg('white');
   @include u-radius(4px);
   border: 1px solid $border-color;
+  max-width: none !important;
```
PDickeyTruss, line 11: Adding this fixed about 60% of the 'weirdness'. We were getting different `max-width` values winning the CSS specificity contest on the first vs subsequent page visits.

src/components/MoveSearchForm/MoveSearchForm.jsx
```diff
@@ -38,7 +38,7 @@ const MoveSearchForm = ({ onSubmit }) => {
             onSubmit={formik.handleSubmit}
             role="search"
           >
-            <p>What do you want to search for?</p>
+            <legend className="usa-label">What do you want to search for?</legend>
```
PDickeyTruss, line 41: This legend w/ className of 'usa-label' came with most the styling needed for the field labels and is what allowed most the css removal/cleanup.

cosburn (approved):
Looks great, thank you for the fix!

mboren (approved):
Looks great, thanks for fixing this!

PR #9363, opened 2022-10-14
MB-10689 customer can upload receipts details on expenses screen

scottmries:
## Summary

This PR integrates the front-end Expenses form with the back-end. 
It also normalizes the "weighing fee" Expenses type to the singular on the front- and back-end.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the customer site
2. Login as movingExpensePPM@ppm.approved
3. Scroll down to click on Upload PPM Documents
4. Click Add Expenses
5. Check that the receipt number increments
6. Select a type other than Storage
7. Fill out the form, including uploading a receipt
8. Check that Saving saves the Expense and routes to the Review page
9. Add another Expense
10. Select Storage for type, and fill out the form, including uploading a receipt
11. Check that Saving saves the Expense and routes to the Review page
12. Edit an existing expense and ensure that it saves
13. From the form, check that clicking Return To Homepage routes to the homepage.

### Frontend


duncan-truss (approved):
Tested the database migration updated existing weighing fee expense type values.

src/pages/MyMove/PPM/Closeout/Expenses/Expenses.jsx
```diff
@@ -1,54 +1,153 @@
...
 
-  const handleSubmit = () => {
-    // TODO: Calls update expense API endpoint
-    history.push(generatePath(customerRoutes.SHIPMENT_PPM_REVIEW_PATH, { moveId, mtoShipmentId }));
+  const handleSubmit = async (values, { setSubmitting }) => {
+    setErrorMessage(null);
+    const payload = {
+      ppmShipmentId: mtoShipment.ppmShipment.id,
+      movingExpenseType: values.expenseType,
+      amount: commaSeparatedStringToFloat(values.amount) * 100,
```
duncan-truss, line 105: I think I'm getting an error when entering a floating point value `65.43` since the API is requiring an integer
Screen Shot 2022-10-17 at 11 17 41 AM
Screen Shot 2022-10-17 at 11 21 09 AM

scottmries: I wrapped that value in `parseInt` in my last commit, which seems to have fixed it.

src/components/Customer/PPM/Closeout/ExpenseForm/ExpenseForm.jsx
```diff
@@ -137,43 +137,35 @@ const ExpenseForm = ({
...
-                      {formikProps.touched?.receiptDocument && formikProps.errors?.receiptDocument && (
-                        <ErrorMessage>{formikProps.errors?.receiptDocument}</ErrorMessage>
+                      {formikProps.touched?.document && formikProps.errors?.document && (
+                        <ErrorMessage>{formikProps.errors?.document}</ErrorMessage>
                       )}
                       <Hint className={styles.uploadInstructions}>
                         <p>{DocumentAndImageUploadInstructions}</p>
                       </Hint>
                       <UploadsTable
                         // className={styles.uploadsTable}
```
duncan-truss, line 151: Do you want to remove this class name, since it doesn't seem to be defined in the SCSS module?

src/components/Customer/PPM/Closeout/ExpenseForm/ExpenseForm.jsx
```diff
@@ -137,43 +137,35 @@ const ExpenseForm = ({
...
-                          )
+                          onUploadDelete(uploadId, 'document', formikProps.setFieldTouched, formikProps.setFieldValue)
                         }
                       />
                       <FileUpload
-                        name="receiptDocument"
-                        className="receiptDocument"
-                        createUpload={(file) => onCreateUpload('receiptDocument', file)}
+                        name="document"
+                        className="document"
```
duncan-truss, line 159: Are we using this className in the parent Expense page component or somewhere else?

duncan-truss:
@scottmries I did see that we don't have a storage date validation in our form to match the one we have on the backend, ensuring that the end date doesn't come before the start.  If you submit with that we get the proper server error but wondering if it's worth making that more user friendly?

Screen Shot 2022-10-17 at 4 27 44 PM
Screen Shot 2022-10-17 at 4 28 04 PM

duncan-truss:
It's also true that we can now create incomplete Expenses but the review page still lets us navigate to the final incentive screen.  We will want to account for that in the Review.jsx component at some point like we do for weight tickets https://github.com/transcom/mymove/blob/4813723fdd7f66b98a81da2fbc2bd8f6fc468c39/src/pages/MyMove/PPM/Closeout/Review/Review.jsx#L115

migrations/app/schema/20221014194001_rename_weighing_fees_to_weighing_fee.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TYPE moving_expense_type RENAME VALUE 'WEIGHING_FEES' TO 'WEIGHING_FEE';
```
felipe-lee, line 1: **praise**: Nice! I didn't know this was a thing, but it's much simpler than what I'd suggested initially.

swagger-def/internal.yaml
```diff
@@ -867,6 +867,9 @@ definitions:
         type: string
         example: '2018-05-26'
         format: date
+      eTag:
+        type: string
+        readOnly: true
```
felipe-lee, line 872: **todo**: This should get removed, we don't pass in eTags via request payloads and are instead meant to be passed in via headers.

src/pages/MyMove/PPM/Closeout/Expenses/Expenses.jsx
duncan-truss: Thinking about this more, I think we're still at risk of getting a different result than what the user entered because of floating point numbers.  I think it's still possible to get a different representation for some numbers like `8976.73` could end up being `897672999999`.

Maybe it's worth using a library like https://github.com/MikeMcl/decimal.js/ even though we aren't doing a bunch of arithmetic on the values themselves and when we are it's the integer cents value so we wouldn't see those issues.  I guess another way is to remove the decimal and parse the number as an int (or parse the whole and decimal parts as integers then combine).

ronaktruss: I think having a separate library for this might be a bit excessive, given our current use cases. I believe we could utilize Math.round to resolve this issue.

ronaktruss: We may be able to use the existing `convertDollarsToCents` function in `shared/utils.js`

scottmries: I changed this to use `convertDollarsToCents`, which uses `Math.round`.

(As it was, `convertDollarsToCents` would only have replaced the first comma, so I updated it to replace them all. I think this better matches what the name of the function claims to do, though I doubt that we currently have use cases for dollars in the millions.)

duncan-truss: I think convertDollarsToCents is probably ok to use in this limited scenario and when we are doing arithmetic operations those happen in integer cents.  It's still a good rule of thumb to avoid floats when dealing with currency.

swagger-def/internal.yaml
scottmries: Removed.

PR #9377, opened 2022-10-17
MB-13792 Cleanup selected/preview violations styling

PDickeyTruss:
## Summary

This PR fixes the somewhat inconsistent styling issue around selected violations. Currently we are using the same component with styling overrides for both the selected violations in the eval report form and in the report preview. This resulted in some competing styles causing layout/spacing issues. I took what felt like the simplest route to fix (dont re-use selected violation component) and was able to simplify and clean up the styling quite a bit as a result.

Note: I realized as I got into this that the styles being removed in this PR were unintentionally effecting a lot of other components in the app. See my below comment and the happo diffs for more detail on this. It is resolved by the changes in this PR. From an eng perspective, I think the learning here is we need to be careful not to add any generic css selectors that are not inside a class or we could end up causing unintended effects to other components/pages (again, see below comment for more).  

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

## To Test/Validate
After login and navigation to the second page of a evaluation form (violations page): 
- [ ] Can select violations
- [ ] Selected violations are displayed and match designs
- [ ] On report preview, selected violations are displayed and match designs. 

## Video

https://user-images.githubusercontent.com/62263329/196496797-efed0be0-9f0f-466a-b080-88286f2ffd7e.mov

### DIFF
--- a/src/components/Office/EvaluationReportPreview/EvaluationReportPreview.module.scss
+++ b/src/components/Office/EvaluationReportPreview/EvaluationReportPreview.module.scss
@@ -68,11 +68,13 @@
   @include u-margin-top(2);
 }
 
-.qaeRemarks, .violationsRemarks  {
+.qaeRemarks,
+.violationsRemarks {
   width: 75%;
 }
 
-.qaeRemarksLabel, .violationsLabel  {
+.qaeRemarksLabel,
+.violationsLabel {
   width: 25%;
 }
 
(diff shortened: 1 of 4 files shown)


src/components/Office/EvaluationReportPreview/EvaluationReportPreview.module.scss
```diff
@@ -81,13 +83,15 @@
...
+.violation {
+  @include u-padding-x(2);
+  @include u-padding-y(1);
+  border: 1px solid $base-light;
+  border-radius: 2px;
+  max-width: 648px;
+  margin-bottom: -1px;
 
-div > p {
-  padding-bottom: 6px;
```
PDickeyTruss, line 92: These were the styles causing issues. They are not inside any class so they seem to be applying globally, this means any `h5` or `p` that is in a `div` is picking up these styles. I can see in the happo diffs that this was effecting quite a few files in the app, so this will be good to get cleaned up.

mandaem (approved):
looks good with this new way

cosburn (approved):
Looks good! Thank you for the fix!

PR #8486, opened 2022-04-27
MB-11317 Fix flaky test by commenting out time assertion and revisit later

donaldthai:
## Summary

Commented out one line of a test assertion regarding time since it was flaky. Will revisit this to make a better test assertion. Time is finicky.

## Setup to Run Your Code

<details>
<summary>💻 You will need to use three separate terminals to test this locally.</summary>

##### Terminal 1

Start the server tests locally.

```sh
make server_test
```

</details>

## Verification Steps for Author

These are to be checked by the author.

- [ ] Request review from a member of a different team.
- [ ] Have the Jira acceptance criteria been met for this change?

## Verification Steps for Reviewers

These are to be checked by a reviewer.

### Backend

- [ ] Code follows the guidelines for [Logging](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide#logging)
- [ ] The requirements listed in [Querying the Database Safely](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide/#querying-the-database-safely) have been satisfied.

PR #9413, opened 2022-10-21
[MB-12340] changed test assertion to fall within 2000 microsecond range

ruizajtruss:
## Summary

I thought of three potential approaches and will include some thoughts for each:

1. Hard coding the date:
    - Seemed like the simplest solution, but, when digging in it seemed like it might be more trouble than its worth since the datagen function uses `time.Now()` to generate a dateTime and this might just end up breaking other tests using that same function.

2. Changing the `suite.Equal()` assertion to `suite.GreaterOrEqual()`:
    - This would assert that the timestamp is equal while still passing if it happens to be greater by a microsecond BUT it won't account for egregious differences, which may be something we want to capture and tend to if it were to happen. 

3. Using `suite.WithinRange()` (my chosen approach for this PR):
    - This asserts whether or not the given timestamp falls within a certain range. I set up the assertion so the range begins at the time we are comparing it against and ending the range with a 2000 microsecond buffer. This way it can properly assert if it is equal while providing a bit of wiggle room incase it were to be ahead by a few microseconds. This approach was also informed a bit by [this article](https://michaeljswart.com/2022/02/five-ways-time-makes-unit-tests-flaky/) and the conversation from the original [PR](https://github.com/transcom/mymove/pull/8486/files#diff-2aaf788a6f0f8978cc3c4b19155cd99a084d553a2d49c70add1fcba5e6dcaf60L161) and more specifically a comment left by Roger suggesting an approach along these lines.

Screen Shot 2022-10-20 at 9 03 50 PM

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Go server locally.

```sh
make server_run
```

##### Terminal 2

Run the Go tests.

```sh
make server_test
```

or

Run the test in the specific file (`-count=1` ensures it fully reruns instead of using cached results if running test back to back).

```sh
go test ./pkg/services/move_history -v -count=1
```

</details>

### Backend

### DIFF
--- a/pkg/services/move_history/move_history_fetcher_test.go
+++ b/pkg/services/move_history/move_history_fetcher_test.go
@@ -285,7 +285,8 @@ func (suite *MoveHistoryServiceSuite) TestMoveFetcher() {
 				suite.Equal(*order.DepartmentIndicator, changedData["department_indicator"])
 
 				//changedData["amended_orders_acknowledged_at"] is being converted from a string to a formatted time.Time type
-				changedDataTimeStamp, _ := time.Parse("2006-01-02T15:04:05.000000", changedData["amended_orders_acknowledged_at"].(string))
+				layout := "2006-01-02T15:04:05.000000"
+				changedDataTimeStamp, _ := time.Parse(layout, changedData["amended_orders_acknowledged_at"].(string))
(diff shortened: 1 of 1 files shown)


YanZ777 (approved):
These changes look reasonable to me!

PR #9424, opened 2022-10-22
MB-14263 Add API to handle PPM Final Closeout Submission

felipe-lee:
## Summary

<details>
<summary>Main Changelog</summary>

- [Add stricter model level validation for SignedCertification](https://github.com/transcom/mymove/pull/9424/commits/5999aed728c4ef770b3afc9bf712f93516b526c9) to match what we actually expect.
- Add service objects for creating and updating `SignedCertifications` 
    - [Add service object to create SignedCertifications](https://github.com/transcom/mymove/pull/9424/commits/183ca3b046fa6adfcc1fe94ab4f2ae6203fd3937)
    - [Add service object to update SignedCertifications](https://github.com/transcom/mymove/pull/9424/commits/2d7cfb83214732fb5255e7bedb58724c90f85a4a)
- [Update PPMShipments/SignedCertification relation to be 1-to-1](https://github.com/transcom/mymove/pull/9424/commits/6c0ad0cd28973c3b9b12163004268b1da0879270)
- [Add PPMShipment make funcs for close out process](https://github.com/transcom/mymove/pull/9424/commits/b0967c8fe153a9ede5bd06d9c67fb3aef89f5e81)
    - a function that sets up a shipment that is ready for the customer to sign and submit their PPM documentation
    - a function that sets up a shipment that needs to be closed out by the counselor
    - a function that sets up a shipment that was sent back to the customer because of a rejected document
- [Update ShipmentRouter.Approve to account for ppm shipments](https://github.com/transcom/mymove/pull/9424/commits/fe43c689299fbf681b6f4527e0aff256fb8b6ee7)
- [Add function to retrieve a PPMShipment with its associations](https://github.com/transcom/mymove/pull/9424/commits/1c80733dd97e4dca0ff2784e734d37f334b8d150)
- [Add PPM shipment router](https://github.com/transcom/mymove/pull/9424/commits/684f74057a0e50ec5153b7742b3760272a1ba4cb)
- [Add service objects to handle customer PPM close out submission](https://github.com/transcom/mymove/pull/9424/commits/e7c5efa35e51dc4cf9d83db0b36a7683fb0d5771)
- [Add handler to submit ppm documentation](https://github.com/transcom/mymove/pull/9424/commits/1352eb6afb65b246e96dbd750ffdecdf1c453089)
- [Add handler to resubmit ppm documentation](https://github.com/transcom/mymove/pull/9424/commits/e9a8824662f79172c6fe4ea7c3ada9e36c6144b5)

</details>

<details>

I'll fill this in as I complete my goals.

- [Add testing suite helpers FatalNotNil and FatalNoVerrs](https://github.com/transcom/mymove/pull/9424/commits/d1a52bf7090dc97086afdc03b02c56f7117e4c91)

</details>

### Resources

* [Service Objects Documentation](https://transcom.github.io/mymove-docs/docs/backend/guides/service-objects/overview)

## Testing Changes

<details>
<summary>💻</summary>

### Terminal 1

Run server tests

```sh
make server_test
```

</details>

### Backend


felipe-lee (commented):
Commenting with things that came up during the a-team walkthrough

pkg/services/signed_certification/rules.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/transcom/mymove/pkg/appcontext"
+	"github.com/transcom/mymove/pkg/models"
+)
+
+// checkSignedCertificationID checks that the ID has not been set if creating, otherwise checks that it hasn't changed
+func checkSignedCertificationID() signedCertificationValidator {
+	return signedCertificationValidatorFunc(func(_ appcontext.AppContext, newSignedCertification models.SignedCertification, originalSignedCertificate *models.SignedCertification) error {
+		verrs := validate.NewErrors()
+
+		if originalSignedCertificate == nil {
```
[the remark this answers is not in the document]
felipe-lee: I tested this and it looks like the DB doesn't care if we set an ID:

```golang
func (suite *ModelSuite) TestCreationWithID() {
	move := testdatagen.MakeDefaultMove(suite.DB())

	signedCertification := models.SignedCertification{
		ID:                uuid.Must(uuid.NewV4()),
		SubmittingUserID:  move.Orders.ServiceMember.User.ID,
		MoveID:            move.ID,
		CertificationText: "Lorem ipsum dolor sit amet...",
		Signature:         "Best Customer",
		Date:              testdatagen.NextValidMoveDate,
	}

	verrs, err := suite.DB().ValidateAndCreate(&signedCertification)
	suite.FatalNoVerrs(verrs)
	suite.FatalNoError(err)

	signedCert := &models.SignedCertification{}

	err = suite.DB().Find(signedCert, signedCertification.ID)

	suite.FatalNoError(err)
}
```

So, given that, I'll keep the rule that checks this, unless anyone really thinks it shouldn't be in this.

felipe-lee:
Tagged @shimonacarvalho and @mlm55 mainly for the ppm shipment router and the signed certification service objects.

swagger-def/definitions/SignedCertification.yaml
```diff
@@ -0,0 +1,64 @@
+description: Signed certification
```
felipe-lee, line 1: Created this definition instead of re-using the existing [SignedCertificationPayload](https://github.com/transcom/mymove/blob/59abae21ee1b98117f866fd97c8e9a8173d563ed/swagger/internal.yaml#L1415-L1455) because the older one isn't following our ideal practices of camel case instead of snake case and is more lax in its requirements.

swagger-def/definitions/SignedCertificationType.yaml
```diff
@@ -0,0 +1,11 @@
+description: >
```
felipe-lee, line 1: Created this type instead of using the existing one to not mess with the valid types according to the other definition, as well as making it `readOnly` because this one is for the response payload and we don't need it for the request payload in the new way.

swagger-def/internal.yaml
```diff
@@ -1462,11 +1462,28 @@ definitions:
...
+    required:
+      - certification_text
+      - signature
+      - date
   SignedCertifications:
     type: array
     items:
       $ref: '#/definitions/SignedCertificationPayload'
-  SignedCertificationType:
+  NullableSignedCertificationType:
```
felipe-lee, line 1486: Renamed this one to be more true to what it allows, and because it can likely eventually be removed with some slight refactoring of the move submission one.

pkg/handlers/internalapi/ppm_shipment_test.go
```diff
@@ -0,0 +1,985 @@
...
+				returnedSubmittedAt := handlers.FmtDateTimePtrToPop(returnedPPMShipment.SubmittedAt).Truncate(time.Second)
+
+				suite.True(
+					originalSubmittedAt.Equal(returnedSubmittedAt),
+					fmt.Sprintf(
+						"SubmittedAt should not have changed: was %s, now %s",
+						ppmShipment.SubmittedAt,
+						returnedPPMShipment.SubmittedAt,
+					),
+				)
```
felipe-lee, line 949: This stumped me. I originally wrote the assertion like this:

```go
suite.True(
	ppmShipment.SubmittedAt.Equal(handlers.FmtDateTimePtrToPop(returnedPPMShipment.SubmittedAt)),
	fmt.Sprintf(
		"SubmittedAt should not have changed: was %s, now %s",
		ppmShipment.SubmittedAt,
		returnedPPMShipment.SubmittedAt,
	),
)
```

which passes locally, but kept failing in circleci. I ran with a debugger locally and I can see that there seems to be a difference, but tbh, I'm not familiar enough with time objects to fully understand what that difference is or why it's there.

image
image

I think it has to do with timezones, as one is UTC while the other is Etc/UTC, which from what I found online seems to be equivalent. And they are, locally, but not in CI 😞

duncan-truss: Yea I think the postgres circleci image sets the default time zone name to "Etc/UTC".  You could get the Location() from the returnedSubmittedAt timestamp and then call the Date() constructor with your originalSubmittedAt timestamp values to set the matching location.

We could also override the postgres configuration to use our provided timezone info to match.

You can see the timezone being initialized in the beginning of the logs for the Postgres image starting up

```
fixing permissions on existing directory /var/lib/postgresql/data ... ok
creating subdirectories ... ok
selecting dynamic shared memory implementation ... posix
selecting default max_connections ... 100
selecting default shared_buffers ... 128MB
selecting default time zone ... Etc/UTC
creating configuration files ... ok
running bootstrap script ... ok
performing post-bootstrap initialization ... ok
syncing data to disk ... ok
```

https://app.circleci.com/pipelines/github/transcom/mymove/46874/workflows/b6af6c65-d134-4ec6-926d-682a59bc0382/jobs/752493

duncan-truss: For the app itself we generally set the `TZ` environment variable to "UTC" in the .envrc and other docker files.  I'm not sure if this is picked up differently locally as in CircleCI or if it's more about the system OS level timezone configs being different in CI.

felipe-lee: Hmm, I'll try out changing the location like you said. If that doesn't work, at least truncating to the second seems to work. I had been thinking about slowly changing the time interval that I'm truncating at to be as specific as possible without it failing, so like going to the millisecond next.

swagger-def/internal.yaml
```diff
@@ -4395,6 +4412,95 @@ paths:
           $ref: '#/responses/UnprocessableEntity'
         '500':
           $ref: '#/responses/ServerError'
+  /ppm-shipments/{ppmShipmentId}/submit-ppm-shipment-documentation:
+    parameters:
+      - $ref: 'parameters/ppmShipmentId.yaml'
+    post:
```
duncan-truss, line 4418: Yea the tricky thing here by this being an action endpoint is that we are creating a signed certification with this post request but returning a 200 because the PPM shipment entity is getting updated.  We're deviating a bit here where a 201 should usually be returned from a POST request and we're not really returning the signed certification in the response because it's not really needed.

swagger-def/internal.yaml
```diff
@@ -4395,6 +4412,95 @@ paths:
...
+          $ref: '#/responses/PermissionDenied'
+        '404':
+          $ref: '#/responses/NotFound'
+        '409':
+          $ref: '#/responses/Conflict'
+        '422':
+          $ref: '#/responses/UnprocessableEntity'
+        '500':
+          $ref: '#/responses/ServerError'
+  /ppm-shipments/{ppmShipmentId}/resubmit-ppm-shipment-documentation/{signedCertificationId}:
```
duncan-truss, line 4455: I can definitely see how you said this endpoint is not 100% needed considering the PPM only has one signed certification that we are updating if it exists.

pkg/models/ppm_shipment.go
```diff
@@ -142,3 +143,35 @@ func FetchPPMShipmentFromMTOShipmentID(db *pop.Connection, mtoShipmentID uuid.UU
 	}
 	return &ppmShipment, nil
 }
+
+// GetPPMShipment returns a PPMShipment with associations by ID
+func GetPPMShipment(appCtx appcontext.AppContext, id uuid.UUID) (*PPMShipment, error) {
```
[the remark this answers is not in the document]
duncan-truss: I added a ppm_shipment_fetcher.go in my last PR that you could move this to. The method I added wasn't a generalized one because it was related to the document ownership.

pkg/handlers/internalapi/ppm_shipment.go
```diff
@@ -0,0 +1,276 @@
...
+	handlers.HandlerConfig
+	services.PPMShipmentNewSubmitter
+}
+
+// Handle saves a new customer signature for PPMShipment documentation submission and routes PPM shipment to the
+// service counselor.
+func (h SubmitPPMShipmentDocumentationHandler) Handle(params ppmops.SubmitPPMShipmentDocumentationParams) middleware.Responder {
+	return h.AuditableAppContextFromRequestWithErrors(params.HTTPRequest,
+		func(appCtx appcontext.AppContext) (middleware.Responder, error) {
+			if appCtx.Session() == nil {
```
duncan-truss, line 27: I think the UserAuthMiddleware takes care of the session and userID not being nil.  I think you're right that the milapp check is needed.

https://github.com/transcom/mymove/blob/main/pkg/handlers/authentication/auth.go#L113

pkg/handlers/internalapi/ppm_shipment.go
```diff
@@ -0,0 +1,276 @@
...
+					handlers.BadRequestErrMessage,
+					errDetail,
+					h.GetTraceIDFromRequest(params.HTTPRequest),
+				)
+
+				return ppmops.NewSubmitPPMShipmentDocumentationBadRequest().WithPayload(errPayload), err
+			}
+
+			payload := params.SavePPMShipmentSignedCertificationPayload
+			if payload == nil {
```
duncan-truss, line 55: I think because you marked this param as required it will be caught by Swagger.  For other validations within that payload you could run `params.SavePPMShipmentSignedCertificationPayload.Validate(strfmt.Default)` in your test but you can't call that on the required payload itself.

pkg/services/ppmshipment/ppm_shipment_new_submitter.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+// SubmitNewCustomerCloseOut saves a new customer signature for PPM documentation agreement and routes PPM shipment
+func (p *ppmShipmentNewSubmitter) SubmitNewCustomerCloseOut(appCtx appcontext.AppContext, ppmShipmentID uuid.UUID, signedCertification models.SignedCertification) (*models.PPMShipment, error) {
+	if ppmShipmentID.IsNil() {
+		return nil, apperror.NewBadDataError("PPM ID is required")
+	}
+
+	ppmShipment, err := models.GetPPMShipment(appCtx, ppmShipmentID)
```
duncan-truss, line 35: This is a recurring problem that we also don't check that this PPM shipment belongs to the session user.  Having a fetcher that does that would be an improvement we can incorporate.

pkg/services/ppmshipment/ppm_shipment_new_submitter.go
```diff
@@ -0,0 +1,77 @@
...
+
+	signedCertification.SubmittingUserID = appCtx.Session().UserID
+	signedCertification.MoveID = ppmShipment.Shipment.MoveTaskOrderID
+	signedCertification.PpmID = &ppmShipment.ID
+
+	certType := models.SignedCertificationTypePPMPAYMENT
+	signedCertification.CertificationType = &certType
+
+	txErr := appCtx.NewTransaction(func(txnAppCtx appcontext.AppContext) error {
+		ppmShipment.SignedCertification, err = p.SignedCertificationCreator.CreateSignedCertification(txnAppCtx, signedCertification)
```
duncan-truss, line 49: Huh, it never occurred to me to assign to a return variable that wasn't an err to an existing type like `ppmShipment.SignedCertification`.

pkg/services/ppmshipment/ppm_shipment_router.go
```diff
@@ -0,0 +1,117 @@
...
+	ppmShipment.Status = models.PPMShipmentStatusWaitingOnCustomer
+
+	if ppmShipment.ApprovedAt == nil {
+		ppmShipment.ApprovedAt = models.TimePointer(*ppmShipment.Shipment.ApprovedDate)
+	}
+
+	return nil
+}
+
+// SubmitCloseOutDocumentation sets the PPM shipment to the NEEDS_CLOSE_OUT status
```
duncan-truss, line 97: Comment references the now removed status.

pkg/services/ppmshipment/ppm_shipment_new_submitter.go
```diff
@@ -0,0 +1,77 @@
...
+
+		err = p.PPMShipmentRouter.SubmitCloseOutDocumentation(txnAppCtx, ppmShipment)
+
+		if err != nil {
+			return err
+		}
+
+		verrs, err := txnAppCtx.DB().ValidateAndUpdate(ppmShipment)
+
+		if verrs.HasAny() {
```
duncan-truss, line 63: I noticed we don't have a Validate() method in our PPMShipment models file 🫤 To be extra safe we should check that verrs is not nil before calling if it contains errors.

Should we be calling the service object validator instead in this case?

pkg/handlers/internalapi/ppm_shipment_test.go
felipe-lee: So, I ran out of time with meetings and wrapping up other things, I won't be able to make this change anymore, but if whoever picks it up has time, I'd be curious if this fixes it.

swagger-def/internal.yaml
felipe-lee: Good point, I was mostly following what we did with the move submission here, but yeah I agree, it's a bit of a strange one.

swagger-def/internal.yaml
felipe-lee: Yeah, I mostly ended up splitting it out so that the handlers and service objects were a bit more straightforward, but I could see us using just one endpoint and having the logic in there to do one or the other after querying for the ppm shipment.

pkg/handlers/internalapi/ppm_shipment.go
felipe-lee: Ooh, thanks, I didn't realize we had those in there. That means we can cut down on these checks in several of our handlers haha

pkg/models/ppm_shipment.go
```diff
@@ -142,3 +143,35 @@ func FetchPPMShipmentFromMTOShipmentID(db *pop.Connection, mtoShipmentID uuid.UU
...
+	var ppmShipment PPMShipment
+
+	err := appCtx.DB().Scope(utilities.ExcludeDeletedScope()).
+		EagerPreload(
+			"Shipment",
+			"WeightTickets",
+			"MovingExpenses",
+			"ProgearExpenses",
```
duncan-truss, line 156: We may need to add in the uploads here too because this is coming back without that data.  Not that we will necessarily need access to the uploads after a successful submission as the user will likely log out/re-open the homepage in the interim.  More so to protect someone from using this method accidentally and getting weird results.

pkg/handlers/internalapi/ppm_shipment.go
felipe-lee: Ah ok, cool. In that case we can remove this. Also, thanks for calling out the `.Validate`, I did it for the response payloads, but not the request ones.

pkg/services/ppmshipment/ppm_shipment_new_submitter.go
felipe-lee: Yeah, in general we don't always check that we're fetching data associated with the user. I can add a ticket for this to my list of tickets to create.

pkg/services/ppmshipment/ppm_shipment_new_submitter.go
felipe-lee: Ah, yeah I figured it was nil before so if there's an error it'll keep being nil, and if not, then it'll already be where we need it.

duncan-truss:
I was able to successfully test with Postman and saw the signed certification created in the database and the ppm shipment get updated with the new status and submittedAt timestamp set 🎉 I know we talked about maybe not trusting the date value being sent in, although I guess the client's date could be different than the server's if that matters.

Screen Shot 2022-11-04 at 4 49 05 PM

pkg/services/ppmshipment/ppm_shipment_new_submitter.go
felipe-lee: We don't have to check for nil because `HasAny` does that check:
https://github.com/gobuffalo/validate/blob/cf5b9b109982bffd4d2eaff7febe14128e05a4ac/validate.go#L90-L97

But, good question about whether we should call the service validator since we don't have validation at the model level 😞 We'd need to either export the validation func though or pull in the updater as a dependency. I'm leaning more toward the latter mainly so it covers everything, but not sure if there would be anything special to consider with that or not.

duncan-truss: I added in the call to use the update PPM Shipment validator, it was already in the same package so just needed to expose the checks.  I didn't add any extra rules around status changes or timestamps or signed certifications though.  It's a little awkward because I need to make a copy to add in the changes, since we don't merge in the status field or timestamp like the other partial updates.

bellex3 (approved):
LGTM. Tested API for POST & PUT requests in Postman. Results were correct and working as it should.

PR #9453, opened 2022-10-25
MB-14264: relocate w2 address on about page and integration

ronaktruss:
## Summary

Move the W-2 address input form the final closeout page to the about page. Add tests to ensure that we don't allow any address from being updated.

### Pending/Nice to haves

- [ ] Look into adding a service object for addresses

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

You can also run the go tests locally by running the following:
```
make server_test_setup
make server_test
```
 For general information on how to run go test you can refer to [this](https://transcom.github.io/mymove-docs/docs/backend/testing/run-go-tests).

### Frontend

### Backend

## Screenshots
milmovelocal_3000_ (4)
milmovelocal_3000_ (5)


ronaktruss:
Hey @rileyorr when you get a chance can you take a look at the [Happo diffs](https://happo.io/a/488/p/603/compare/02a35ad093b21c48953dc49254932dff1dcbc458/cac76e6059d04f35e2bad37bd4d77f04a07996e8). The changes that are in right now are the core changes. If you take a look at the happo differences you may see that the Back / Save & Continue buttons are cut off (this is an existing issue). I'm planning on making an additional change to fix this but it will create a lot more happo differences so I wanted to give you the chance to review the actual changes before I add those.

ronaktruss:
> LGTM! I also reviewed the Happo diffs

@rileyorr Thanks for reviewing that! 

I just [updated some of our generic css overrides for storybook](https://github.com/transcom/mymove/pull/9453/files#diff-9a30cef10a543c375f79aed9432a7d09a18e3c6c506f42f7fece52735a401c2a). These new [happo diffs](https://happo.io/a/488/p/603/compare/02a35ad093b21c48953dc49254932dff1dcbc458/a8cd26609fbf6a5c4128da9842f5290d0bc3e774) should all be basically the same except with some pixel shifts. We were adding a 5px padding to the top of the components that would cut off the bottom 5 px of all of the components. I'm happy to pair if there are any questions or concerns.

pkg/handlers/internalapi/internal/payloads/payload_to_model.go
```diff
@@ -27,6 +27,21 @@ func AddressModel(address *internalmessages.Address) *models.Address {
 	}
 }
 
+func AddressModelWithoutID(address *internalmessages.Address) *models.Address {
```
duncan-truss, line 30: Hmm do we need a separate payload to model method here?  I think the existing one supports nil values and isn't the service object making the determination to create a new one or know which record to update regardless of the param if that's your concern?

ronaktruss: So the reason that this was added is to handle the scenario where an id is passed in for say an address that is a duty location. If we didn't have this rule we would technically allow users to update arbitrary addresses if they had the corresponding id. 

We could alternatively add logic to ensure that the id that we are trying to update is associated with the ppmShipment.

duncan-truss: I see your [merge method](https://github.com/transcom/mymove/pull/9453/files#diff-63190e6f0b86f2b79185118041d816fa8cc4554928ee010ad00b40ebd6c7bbd1R97-R99) making sure to update the right address if it exists but I guess you're thinking of the scenario where it doesn't yet and validateAndSave would instead update an arbitrary id if it was provided?

I might have slight preference to have a single payload function and have the service object ignore a passed in id (or clear the value) if there is no existing address on the PPM shipment.

pkg/services/ppmshipment/ppm_shipment_updater_test.go
```diff
@@ -760,4 +760,60 @@ func (suite *PPMShipmentSuite) TestUpdatePPMShipment() {
...
+				State:          "NJ",
+				PostalCode:     "08004",
+			},
+		}
+
+		updatedPPM, err := subtestData.ppmShipmentUpdater.UpdatePPMShipmentWithDefaultCheck(appCtx, &newPPM, originalPPM.ShipmentID)
+
+		suite.NilOrNoVerrs(err)
+
+		suite.Equal(*newPPM.W2Address, *updatedPPM.W2Address)
```
duncan-truss, line 785: Hmm does testify really consider these equal if updatedPPM has a generated ID value and newPPM doesn't?

pkg/handlers/internalapi/internal/payloads/payload_to_model.go
ronaktruss: > I see your [merge method](https://github.com/transcom/mymove/pull/9453/files#diff-63190e6f0b86f2b79185118041d816fa8cc4554928ee010ad00b40ebd6c7bbd1R97-R99) is making sure to update the right address if it exists but I guess you're thinking of the scenario where it doesn't yet and validateAndSave would instead update an arbitrary id if it was provided?

Yeah exactly, see [this test](https://github.com/transcom/mymove/pull/9453/files#diff-0dc4f49ea27f9cbe2b9257753c04679e143009d5db990c8997cd4e0894aca26fR1070-R1096). 

As you mentioned we could also just override the id in the merge function. The current implementation feels a bit more readable (because you see that id isn't used at all) but requires maintaining multiple functions for the same model.

With service objects we might be able to find a happy middle ground because we'll explicitly be calling the create or update methods.

src/components/Customer/PPM/Closeout/AboutForm/AboutForm.module.scss
```diff
@@ -11,4 +11,35 @@
...
+      :global(.grid-row.grid-gap) {
+        div:nth-child(1), div:nth-child(2) {
+          @include at-media('desktop'){
+            width: 50%;
+          }
+          :global(.usa-form-group) {
+            width: 100%;
+          }
+        }
+      }
```
duncan-truss, line 42: Nice job incorporating responsive styles

src/components/Customer/PPM/Closeout/AboutForm/AboutForm.stories.jsx
```diff
@@ -51,6 +51,13 @@ BlankWithDefaultZIPs.args = {
   postalCodeValidator: () => {},
 };
 
+const validW2Address = {
```
duncan-truss, line 54: Did you look into using the test data factory for address here?

pkg/services/ppmshipment/ppm_shipment_updater_test.go
ronaktruss: I think newPPM gets updated with the generated id since its being passed in as a pointer and the modifications are being made to that referenced object.

duncan-truss: Oh that's surprising, maybe check that the values are equal to explicit string values in that case because that feels like we're comparing the same object basically and would it ever fail?

duncan-truss:
@ronaktruss One other minor issue I'm seeing is we are allowing the streetAddress2 to be set to the empty string when saving it in the database.  I would probably prefer that to ultimately be saved as null instead of the blank string since it's an optional field.  I'll leave it up to you on how you want to do that since there are probably a few different ways to go about it.

src/components/Customer/PPM/Closeout/AboutForm/AboutForm.stories.jsx
ronaktruss: Is there an existing one?

ronaktruss: 🤦 I was using the file search and using `factory` so it didn't show up. Since this is for storybook we want consistent values so we don't have random happo diffs from test to test so it might not make sense to use it in this instance.

ronaktruss:
> @ronaktruss One other minor issue I'm seeing is we are allowing the streetAddress2 to be set to the empty string when saving it in the database. I would probably prefer that to ultimately be saved as null instead of the blank string since it's an optional field. I'll leave it up to you on how you want to do that since there are probably a few different ways to go about it.

Yeah I noticed this as well. I think it makes more sense to have the DB store it as null, but didn't want to make this a one off change. Our other usages of the address component also set the DB value to the empty string. Hopefully with the address service object we can refactor the existing usages with the service object and make this change as well.

src/components/Customer/PPM/Closeout/AboutForm/AboutForm.stories.jsx
duncan-truss: Great point, yea maybe either provide overrides or just use it in the jest tests.

src/components/Customer/PPM/PPM.module.scss
```diff
@@ -38,7 +38,7 @@
   .sectionWrapper {
     border-bottom: 1px solid $base-lighter;
 
-    @include at-media('mobile') {
+    @include at-media-max('tablet') {
```
duncan-truss, line 41: I wish we still had mobile happo diffs to make this easier to see, this is used in the booking parts of PPMs too.

duncan-truss (approved):

PR #9472, opened 2022-10-27
[ on-call ] Upgrade PSQL to 12.11 for the application

rogeruiz:
## Summary

This will help run the tests before I deploy this to EXP for further
testing.

We're trying to validate that upgrading RDS to version 12.11 will not affect
application engineering engineers in any way besides having to restart their
local Docker containers to get the new versions.


jesset-truss (approved):
Looks good!

PR #9480, opened 2022-10-28
MB-14164 Get ProGearWeights from service member's order's entitlement.

scottmries:
## Summary

In the office app, service counselors can update pro gear allowances for service members, but these updates were not being reflected in the API.

This ticket uses those values if they exist, otherwise defaulting to the previous functionality, setting them to allowances based on rank.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Log in as a Service Counselor in the office app
2. Select a move (I used `PPMSIT`)
3. Click Edit Allowances
4. Update Pro-gear and Spouse pro-gear values
5. Log in as the corresponding user in the customer app
6. In Redux, check that the updated values are present (`entities.orders.{id}.entitlement`)

### Backend


pkg/handlers/internalapi/service_members.go
```diff
@@ -33,6 +33,15 @@ func payloadForServiceMemberModel(storer storage.FileStorer, serviceMember model
 		weightAllotment = payloadForWeightAllotmentModel(models.GetWeightAllotment(*serviceMember.Rank))
 	}
 
+	if len(serviceMember.Orders) > 0 {
```
duncan-truss, line 36: I might suggest that we leave the weightAllotment alone since it is something distinct.  We can write a Redux selector in the Pro gear component to pull out the entitlement values and use those if they are set or fallback to the weight allotment.  We can test the component under those different conditions too.

bellex3:
Followed the above steps to update the pro-gear weights for self & spouse. Confirming new progear weights are updated in Redux. 
Screen Shot 2022-10-31 at 12 30 04 PM

bellex3 (commented):
Followed the above steps to update progear weights for self and spouse. Confirming new changes are shown in Redux. 
Screen Shot 2022-10-31 at 12 30 04 PM

pkg/handlers/internalapi/service_members.go
ronaktruss: Do you have some context for why/how weightAllotment is distinct? I know most of our FE where we show weight limits is based off of these weightAllotment.  Do we not want these values to reflect the data in the entitlements table? Right now(prior to these changes) they are just reflecting the default values.

scottmries: The last selector in that file might also work, if that's the way to go:
https://github.com/transcom/mymove/blob/db17a1fe5b166d37f3ca5797a5b1daf2f6044991/src/store/entities/selectors.js#L230-L243

duncan-truss: @ronaktruss The weight allotment is supposed to be purely what the GHC contract defines as maximum weight limits based on the service member's rank and dependents.  Our entitlements table represents the office user either lowering or raising that allowance limit at their discretion.

If you create a new service member, when orders get created we also create the entitlement record and those values pro_gear_weight and  pro_gear_weight_spouse are set to zero, while the weight allotment for pro-gear should still be 2000 lbs and 500 lbs.

So if we wanted to start using those limits in say the PPM booking form for pro-gear (instead of the current hard-coded maxes for every rank), we would want to rely on the weight allotment limits and not the initialized entitlement values before an office user has counseled/approved the shipment.  We may want to be able to still display the allotment values other places (like we do for shipment allotments when creating a new shipment or when updating your rank in the profile) without it being affected by what the office user ultimately decides.

I think overloading the same fields with different values makes people confused about what is actually happening too.  Model to payload methods should not be performing that sort of business logic if we can help it and simpler singular responsibility will lead to fewer issues.

duncan-truss: @scottmries We do use that selector in a few places already but that is being used strictly for the allotment and mixing in entitlement values would lead to unexpected results I think.

I see part of the issue is that the internal orders API doesn't define entitlements on the orders model the way that the office does.  I think I would consider adding that entity to the orders response.

ronaktruss: Gotcha, that makes sense to me then. I wasn't aware that weight allotment was meant for that use case.

scottmries: @duncan-truss I'm now attaching the entitlement to the Orders payload, and I added a selector to pull it in for the service member.

I repeat the definition of the Entitlement shape in the internal.yaml, which feels wrong, but I couldn't find another example of a top-level entity with `type: object` that pulled in from a reference to a definition file. Suggestions about how to refactor that would be welcome.

Is the actual use case for the selector built yet? I could import and call it in the ProGearForm or the parent page component, but I'd be defining but not using that variable, I believe.

In any case, the database or default value is now on `entities.orders.{id}.entitlement`.

src/store/entities/selectors.js
```diff
@@ -250,3 +250,18 @@ export const selectEntitlementsForLoggedInUser = createSelector(
...
 );
+
+export const selectEntitlementForServiceMember = createSelector(
+  selectServiceMemberFromLoggedInUser,
+  selectCurrentOrders,
+  (serviceMember, orders) => {
+    let entitlement = {
+      proGear: serviceMember.weight_allotment?.pro_gear_weight,
+      proGearSpouse: serviceMember.weight_allotment?.pro_gear_weight_spouse,
+    };
```
ronaktruss, line 261: Is there a scenario where we use these default values? 

Also should we be defaulting to these values in the first place? I would expect the entitlements object to always just give the info from the entitlements table.

swagger-def/internal.yaml
```diff
@@ -1794,6 +1794,17 @@ definitions:
...
+        type: object
+        properties:
+          proGear:
+            type: integer
+            example: 2000
+            x-nullable: true
+          proGearSpouse:
+            type: integer
+            example: 500
+            x-nullable: true
```
ronaktruss, line 1807: If you want to just define this in this current file you can likely just do the following
```suggestion
        $ref: '#/definitions/Entitlement'
```
You may also need to updated the orders.go file to use `internalmessages.Entitlement` instead of `internalmessages.OrdersEntitlement` once you make this change.

We can also get rid of the Entitlement.yaml file as well. Alternatively, if you want to define a new file you can add the definition there and reference it similar to what we do for ppmShipment([ref](https://github.com/transcom/mymove/blob/master/swagger-def/internal.yaml#L2236-L2237)) and the associated `PPMShipment.yaml` file.

src/store/entities/selectors.js
duncan-truss: Yea I guess it depends on how we want to use this but definitely leaning towards just returning only the entitlement object's values here.  The entitlement will exist once orders are created and the values will be zero for proGear and proGearSpouse.

We could rename the other selector `selectEntitlementForLoggedInUser` -> `selectWeightAllowanceForLoggedInUser` to avoid confusion there too and that one will always return the weight allowance.

If the component wants to apply some logic let it call both selectors and make it explicit.

scottmries:
The entitlement is now available directly on `entities.orders.{id}.entitlement` in Redux. I believe this passes the AC on the ticket; I'm assuming the selector and its tests would part of a ticket for the component that consumes this data? 
@duncan-truss What do you think?

swagger-def/internal.yaml
```diff
@@ -1823,6 +1834,18 @@ definitions:
       - uploaded_orders
       - created_at
       - updated_at
+  # i think this is right:
+  Entitlement:
```
duncan-truss, line 1838: I think to Ronak's comment above you defined this but didn't reference it in the Orders definition entitlement property.

Could we add description fields here saying that these are set by the office user different from the weight entitlement possibly?

swagger-def/internal.yaml
duncan-truss: The [Entitlement.yaml](https://github.com/transcom/mymove/pull/9480/files#diff-054742b2e8d142980849dc48ad9d5ea65acf602c47aa5c507ab7a369dd987ce2) file looks empty so let's delete that before merging.

duncan-truss (approved):
Tested the changes with a service member and see both weight allotment values and entitlement values coming back in the logged_in network request.

PR #9502, opened 2022-11-01
[MB-14168] backend handler and service object for creating and updating progear

pearl-truss:
## Summary

This PR creates the services to `create` and `update` progear. As well as the respective handlers and YAML definitions and endpoints. And includes work done to create services that was originally in https://github.com/transcom/mymove/pull/9509

## Reviewing Code
- Service file changes are primarily under `./pkg/services/progear_weight_ticket`
- YAML changes are in `swagger-def`
- Handler changes are in `./pkg/handlers/internalapi`


pkg/services/progear_weight_ticket/validation_test.go
```diff
@@ -0,0 +1,77 @@
...
+			verrs.Add("ID", "fake error")
+
+			return verrs
+		})
+
+		err := validateProgearWeightTicket(suite.AppContextForTest(), nil, nil, []progearWeightTicketValidator{checkAlwaysReturnValidationErr}...)
+
+		suite.Error(err)
+		suite.IsType(apperror.InvalidInputError{}, err)
+		suite.Contains(err.Error(), "Invalid input found while validating the weight ticket.")
```
NamibiaTorres, line 63: For the error message in these tests can you add `progear` before anywhere where we find the words `weight ticket` that way it's much clearer that this is a progear weight ticket rather than a regular one. This was something I forgot to update before you merged the SO work.

pkg/services/progear_weight_ticket/validation_test.go
```diff
@@ -0,0 +1,77 @@
...
+	})
+
+	suite.Run("Runs validation and returns other errors", func() {
+		checkAlwaysReturnOtherError := progearWeightTicketValidatorFunc(func(_ appcontext.AppContext, _ *models.ProgearWeightTicket, _ *models.ProgearWeightTicket) error {
+			return apperror.NewNotFoundError(uuid.Must(uuid.NewV4()), "Weight ticket not found.")
+		})
```
NamibiaTorres, line 69: `progear weight ticket`

pkg/services/progear_weight_ticket/validation_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+		err := validateProgearWeightTicket(suite.AppContextForTest(), nil, nil, []progearWeightTicketValidator{checkAlwaysReturnOtherError}...)
+
+		suite.Error(err)
+		suite.IsType(apperror.NotFoundError{}, err)
+		suite.Contains(err.Error(), "Weight ticket not found.")
+	})
```
NamibiaTorres, line 76: `progear weight ticket`

pkg/handlers/internalapi/internal/payloads/model_to_payload.go
```diff
@@ -408,3 +408,39 @@ func ProGearWeightTickets(proGearWeightTickets models.ProgearWeightTickets) []*i
 	// TODO: MB-14168 will fill this in. Needed to at least have this for MB-13773, but the goal wasn't to get this working fully until MB-14168 and related tickets are done.
 	return payload
 }
+
```
NamibiaTorres, line 411: Wondering if we still need the TODO or can it be removed?

ronaktruss (commented):
So far this looks good to me. I still need to test a few more things in the morning and take a closer look at the tests.

pkg/services/progear_weight_ticket/progear_weight_ticket_creator.go
```diff
@@ -0,0 +1,69 @@
...
+
+	var progearWeightTicket models.ProgearWeightTicket
+
+	txnErr := appCtx.NewTransaction(func(txnCtx appcontext.AppContext) error {
+
+		document := models.Document{
+			ServiceMemberID: appCtx.Session().ServiceMemberID,
+		}
+		allDocs := models.Documents{document}
+		verrs, err := appCtx.DB().ValidateAndCreate(allDocs)
```
ronaktruss, line 38: Since we only have a singular document we can likely remove the allDocs line and call `ValidateAndCreate` directly on the document created on the previous line. 

allDocs was used for weightTickets so that we could create multiple documents within the same ValidateAndCreate call.

pkg/services/progear_weight_ticket/progear_weight_ticket_updater.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		switch err {
+		case sql.ErrNoRows:
+			return nil, apperror.NewNotFoundError(progearWeightTicketID, "while looking for ProgearWeightTicket")
+		default:
+			return nil, apperror.NewQueryError("ProgearWeightTicket fetch original", err, "")
+		}
+	}
+
+	progearWeightTicket.Document.UserUploads = FilterDeletedValued(progearWeightTicket.Document.UserUploads)
```
ronaktruss, line 110: We can get rid of the `FilteredDeletedValued` function and call in favor of the `FilterDeleted()` method that was added to `UserUploads`. 

We'll likely need to clean this up for weightTickets as well, but that doesn't have to be in this PR. (I might look into making a quick PR for this change)

pkg/services/progear_weight_ticket/rules.go
```diff
@@ -0,0 +1,149 @@
...
+		if newProgearWeightTicket.Description == nil || *newProgearWeightTicket.Description == "" {
+			verrs.Add("Description", "Description must have a value of at least 0")
+		}
+
+		if newProgearWeightTicket.HasWeightTickets == nil {
+			verrs.Add("HasWeightTickets", "HasWeightTickets is required")
+		}
+
+		// ARE WE ALLOWING 0 values now?
+		if newProgearWeightTicket.Weight == nil || *newProgearWeightTicket.Weight < 1 {
```
ronaktruss, line 43: I think this is correct and we should be only allowing a weight of 1 or more. We'll likely have to update the FE during the integration ticket. 

(Question/thought): Should we be ensuring this at the swagger level as well?

pkg/services/progear_weight_ticket/progear_weight_ticket_creator.go
pearl-truss: I do get errors in the creator test when I try to apply this, will have to look into an alternative possibly

pkg/services/progear_weight_ticket/progear_weight_ticket_updater.go
pearl-truss: I made the update here and saw that the weightTicketUpdater seems to have already been updated as well

pkg/services/progear_weight_ticket/rules.go
pearl-truss: Do we need to enforce this in 2 places?

pkg/services/progear_weight_ticket/progear_weight_ticket_updater.go
ronaktruss: Oh nice, Duncan did mentioned he touched some of those when he was working on Felipe's PR.

pkg/services/progear_weight_ticket/rules.go
ronaktruss: It's definitely not "necessary", and it seems like in most cases we have defaulted to a min of 0 on swagger, but it might be worth reviewing this pattern. We don't consistently choose where we are doing validations. Sometimes we rely on swagger and other times on our application code. So we are relying on both either intentionally or not.

pearl-truss: That's a good point that it's in 2 places regardless....And I checked and there's already a min of 0 applied in the swagger def on the progear definitions

pkg/services/progear_weight_ticket/progear_weight_ticket_updater_test.go
```diff
@@ -0,0 +1,343 @@
...
+		updatedProgearWeightTicket, updateErr := updater.UpdateProgearWeightTicket(appCtx, *desiredProgearWeightTicket, etag.GenerateEtag(originalProgearWeightTicket.UpdatedAt))
+
+		suite.Nil(updateErr)
+		suite.Equal(originalProgearWeightTicket.ID, updatedProgearWeightTicket.ID)
+		suite.Equal(originalProgearWeightTicket.DocumentID, updatedProgearWeightTicket.DocumentID)
+		suite.Equal(*desiredProgearWeightTicket.Description, *updatedProgearWeightTicket.Description)
+		suite.Equal(*desiredProgearWeightTicket.Weight, *updatedProgearWeightTicket.Weight)
+		suite.Equal(*desiredProgearWeightTicket.HasWeightTickets, *updatedProgearWeightTicket.HasWeightTickets)
+		suite.Equal(*desiredProgearWeightTicket.BelongsToSelf, *updatedProgearWeightTicket.BelongsToSelf)
+		suite.Equal(2, len(updatedProgearWeightTicket.Document.UserUploads))
```
ronaktruss, line 157: We don't need to change this, but just sharing this because I only recently found out about this:
There is a `suite.Len` function.

pkg/services/progear_weight_ticket/rules.go
```diff
@@ -0,0 +1,148 @@
...
+			return verrs
+		}
+
+		if newProgearWeightTicket.PPMShipmentID.IsNil() {
+			verrs.Add("PPMShipmentID", "PPMShipmentID must exist")
+		}
+
+		if newProgearWeightTicket.Document.ServiceMemberID.IsNil() {
+			verrs.Add("ServiceMemberID", "Document ServiceMemberID must exist")
+		}
```
ronaktruss, line 73: These two if checks are not explicitly covered by our tests within this repo. They might be tangentially tested by the tests in the handler. If we can easily add tests for this, that might be a nice to have.

ronaktruss (approved):
LGTM, I left some minor enhancements comments.

pearl-truss:
> LGTM, I left some comments related to minor enhancements.

Thanks for the review @ronaktruss! I made all of the suggested changes including additional test

swagger-def/internal.yaml
```diff
@@ -4244,6 +4260,85 @@ paths:
...
+        pro-gear weight ticket. Data should be filled in using the patch endpoint.
+      operationId: createProGearWeightTicket
+      tags:
+        - ppm
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters:
+        - $ref: 'parameters/ifMatch.yaml'
```
duncan-truss, line 4279: I'm not sure the POST endpoint would need the If-Match header with the eTag only the PATCH would require it?

swagger-def/internal.yaml
```diff
@@ -4244,6 +4260,85 @@ paths:
...
+      operationId: updateProGearWeightTicket
+      tags:
+        - ppm
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters:
+        - $ref: 'parameters/ifMatch.yaml'
+        - $ref: 'parameters/ppmShipmentId.yaml'
```
duncan-truss, line 4317: This may have been defined already above in the shared endpoint parameters https://github.com/transcom/mymove/pull/9502/files#diff-d75579e46ab58b0cd40af93d565278cc0643cb0a2de4ea57a065ab1da6b1d5e2R4301

swagger-def/internal.yaml
```diff
@@ -4244,6 +4260,85 @@ paths:
...
+      tags:
+        - ppm
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters:
+        - $ref: 'parameters/ifMatch.yaml'
+      responses:
+        '200':
```
duncan-truss, line 4281: The response here should be a `201` created for a POST instead of a `200` for consistency.

pkg/handlers/internalapi/progear_weight_ticket.go
```diff
@@ -0,0 +1,138 @@
...
+type CreateProGearWeightTicketHandler struct {
+	handlers.HandlerConfig
+	progearCreator services.ProgearWeightTicketCreator
+}
+
+// Handle creating a progear weight ticket
+func (h CreateProGearWeightTicketHandler) Handle(params progearops.CreateProGearWeightTicketParams) middleware.Responder {
+	return h.AuditableAppContextFromRequestWithErrors(params.HTTPRequest,
+		func(appCtx appcontext.AppContext) (middleware.Responder, error) {
+			if appCtx.Session() == nil {
```
duncan-truss, line 26: I think not having a session will be caught in the auth middleware but you can keep the check for the IsMilApp below https://github.com/transcom/mymove/blob/fde0969da9ffab37f81897c0f7d18893fa050cc3/pkg/handlers/authentication/auth.go#L121-L125

pkg/services/progear_weight_ticket/rules.go
```diff
@@ -0,0 +1,149 @@
...
+
+		if len(oldProgearWeightTicket.Document.UserUploads) < 1 {
+			verrs.Add("Document", "At least 1 weight ticket is required")
+		}
+
+		if newProgearWeightTicket.Status != nil {
+			if (*newProgearWeightTicket.Status == models.PPMDocumentStatusExcluded || *newProgearWeightTicket.Status == models.PPMDocumentStatusRejected) && (newProgearWeightTicket.Reason == nil || *newProgearWeightTicket.Reason == "") {
+				verrs.Add("Reason", "A reason must be provided when the status is EXCLUDED or REJECTED")
+			}
+		}
```
duncan-truss, line 54: I think we may need an else condition here if status was nil but reason had a value provided.

pearl-truss:
@duncan-truss thanks for the review, all has been addressed

duncan-truss (approved):
You made those changes fast, thank you!

PR #9505, opened 2022-11-01
MB-14427: Add address service object

ronaktruss:
## Summary

This adds a service object for address and utilizes it to modify the W-2 address on a ppmShipment. This also adds logic to prevent the DB from storing empty strings for optional values. I'm not entirely sold on this being implemented on the BE but the benefit is that we wouldn't have to make the change in all of the FE code where we are passing the payload.

## Setup to Run Your Code

See steps in the [original PR](https://github.com/transcom/mymove/pull/9453)


pkg/services/address/address_updater.go
```diff
@@ -0,0 +1,55 @@
...
+		return nil, apperror.NewBadDataError("invalid ID used for address")
+	}
+
+	// verify ETag
+	if etag.GenerateEtag(originalAddress.UpdatedAt) != eTag {
+		return nil, apperror.NewPreconditionFailedError(originalAddress.ID, nil)
+	}
+
+	// tk handle merge?
+
```
ronaktruss, line 35: I'm planning on adding merge functionality that should handle setting things like streetAddress2 to null if we get an empty string.

pkg/services/address/address_updater.go
```diff
@@ -0,0 +1,77 @@
...
+		return nil
+	})
+	if txnErr != nil {
+		return nil, txnErr
+	}
+
+	return &mergedAddress, nil
+}
+
+func mergeAddress(address, originalAddress models.Address) models.Address {
```
ronaktruss, line 57: This function helps prevent empty strings from being set as the value in the DB for optional fields. As mentioned in the description I am not entirely sold on this being the best way to do this. I'm open to other suggestions.

pkg/services/address/address_creator.go
```diff
@@ -0,0 +1,63 @@
...
+		return nil
+	})
+	if txnErr != nil {
+		return nil, txnErr
+	}
+
+	return &transformedAddress, nil
+}
+
+func transformNilValuesForOptionalFields(address models.Address) models.Address {
```
ronaktruss, line 47: This function helps prevent empty strings from being set as the value in the DB for optional fields. As mentioned in the description I am not entirely sold on this being the best way to do this. I'm open to other suggestions.

pkg/services/address/address_updater.go
```diff
@@ -0,0 +1,77 @@
...
+func NewAddressUpdater() services.AddressUpdater {
+	return &addressUpdater{
+		checks: []addressValidator{
+			checkID(),
+		},
+	}
+}
+
+func (f *addressUpdater) UpdateAddress(appCtx appcontext.AppContext, address *models.Address, eTag string) (*models.Address, error) {
+	originalAddress := models.FetchAddressByID(appCtx.DB(), &address.ID)
```
ronaktruss, line 24: To get the eTag in `pkg/services/ppmshipment/ppm_shipment_updater.go` where we call this function we end up fetching the associated address twice.

pearl-truss (approved):
I left some nonblocking comments but it also looks good to go as is

pkg/services/address/rules.go
```diff
@@ -0,0 +1,23 @@
+package address
```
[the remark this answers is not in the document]
ronaktruss: The swagger definition for address covers the basic formatting rules. I expect the rules to become more lax if we begin supporting oconus moves and since this is likely to be shared for all addresses didn't include any explicit rules.

pkg/services/address/address_creator.go
ronaktruss: Not explicitly, but preventing this using `pattern` would mean we would have to update how our UI sends the data in order to comply with the rules. The current implementation allows both not sending the optional fields with an empty string and not sending it all.

ronaktruss:
@mergifyio refresh

PR #9530, opened 2022-11-04
Fix ready to merge label since master has been renamed to main

YanZ777:
## Summary

I noticed that the ready to merge label got broken when the branch got renamed to main. 
Ideally, this should fix the label and I think I should be able to use this PR as a test.


rogeruiz (approved):
lgtm 🌈

Great work on this @YanZ777. thanks.

PR #9531, opened 2022-11-04
MB-13877 factory.BuildUser added as new version of testdatagen.makeUser

shimonacarvalho:
## Summary

This ticket adds a factory package with first function BuildUser intended to be an eventual replacement for testdatagen. 
* Created BuildUser and shared functions
* Created tests for all of those. 
* Replaced all makeUser and makeDefaultUser calls except those in
   * testdatagen
   * testdatagen/scenarios

Reason for not replacing all
* testdatagen make functions will slowly be deprecated as we keep going, so there's no need to replace them. The functions will eventually get deleted. Also there's no good way to "pass down assertions" from the old make functions to the new build functions. 
* testdatagen/scenarios *will* be updated but we intend to break that work down into tickets for our team. 
* Wanted to get the base code in and examples of how to replace so we can split up the work.

Please pay note to:
* Most functionality has been reviewed in a spike. 
* But new additions in this ticket are
   *  the `assignDefaultTypes` code and 
   * the `setupCustomizations` functions. 
* Those simplify usage for devs. Please give extra attention

## Setup to Run Your Code

### Frontend

- No frontend changes

### Backend

- No schema changes

### DIFF
--- a/pkg/factory/factory_test.go
+++ b/pkg/factory/factory_test.go
@@ -1,0 +1,369 @@
+package factory
+
+import (
+	"testing"
+	"time"
+
+	"github.com/go-openapi/swag"
+	"github.com/gofrs/uuid"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/transcom/mymove/pkg/models"
+	"github.com/transcom/mymove/pkg/testingsuite"
+)
+
+type MakerSuite struct {
+	*testingsuite.PopTestSuite
+}
+
+func TestMakerSuite(t *testing.T) {
+
+	ts := &MakerSuite{
+		PopTestSuite: testingsuite.NewPopTestSuite(testingsuite.CurrentPackage(), testingsuite.WithPerTestTransaction()),
+	}
+	suite.Run(t, ts)
+	ts.PopTestSuite.TearDown()
+}
+
+func (suite *MakerSuite) TestElevateCustomization() {
+
+	suite.Run("Customization converted ", func() {
+
+		customEmail := "leospaceman123@example.com"
+		streetAddress := "235 Prospect Valley Road SE"
+
+		customizationList :=
+			[]Customization{
+				{
+					Model: models.User{LoginGovEmail: customEmail},
+					Type:  User,
+				},
+				{
+					Model: models.Address{StreetAddress1: streetAddress},
+					Type:  Addresses.ResidentialAddress,
+				},
+			}
+		tempCustoms := convertCustomizationInList(customizationList, Addresses.ResidentialAddress, Address)
+		// Nothing wrong with customizations
+		tempCustoms, err := validateCustomizations(tempCustoms)
+		suite.NoError(err)
+		// Customization has new type
+		suite.Equal(Address, tempCustoms[1].Type)
+		// Old customization list is unchanged
+		suite.Equal(Addresses.ResidentialAddress, customizationList[1].Type)
+	})
+}
+
+// getTraitActiveArmy is a custom Trait
+func getTraitActiveArmy() []Customization {
+	army := models.AffiliationARMY
+	return []Customization{
+		{
+			Model: models.User{
+				LoginGovEmail:         "trait@army.mil",
+				CurrentAdminSessionID: "my-session-id",
+			},
+			Type: User,
+		},
+		{
+			Model: models.ServiceMember{
+				Affiliation: &army,
+			},
+			Type: ServiceMember,
+		},
+	}
+}
+
+func (suite *MakerSuite) TestMergeCustomization() {
+
+	suite.Run("Customizations and traits merged into result", func() {
+		// Under test:       mergeCustomization, which merges traits and customizations
+		// Set up:           Create a customization without a matching trait, and traits with no customizations
+		// Expected outcome: All should exist in the result list
+		streetAddress := "235 Prospect Valley Road SE"
+		result := mergeCustomization(
+			// Address customization
+			[]Customization{
+				{
+					Model: models.Address{
+						StreetAddress1: streetAddress,
+					},
+					Type: Address,
+				},
+			},
+			// User and ServiceMember customization
+			[]Trait{
+				getTraitActiveArmy,
+			},
+		)
+		suite.Len(result, 3)
+		_, custom := findCustomWithIdx(result, Address)
+		suite.NotNil(custom)
+		address := custom.Model.(models.Address)
+		suite.Equal(streetAddress, address.StreetAddress1)
+
+		// Check that User customization from trait is available
+		_, custom = findCustomWithIdx(result, User)
+		suite.NotNil(custom)
+		user := custom.Model.(models.User)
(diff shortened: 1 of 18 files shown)


pkg/factory/shared.go
```diff
@@ -0,0 +1,387 @@
...
+			if callerCustom != nil {
+				// If a customization has an ID, it means we use that precreated object
+				// Therefore we can't merge a trait with it, as those fields will not get
+				// updated.
+				// While this feels like we should warn or error out, we want to support overriding a
+				// trait with a precreated object so it's not an error.
+				if !hasID(callerCustom.Model) {
+					result := mergeInterfaces(traitCustom.Model, callerCustom.Model)
+					callerCustom.Model = result
+					customs[j] = *callerCustom
```
reggieriser, line 272: **thought**: Does this function modify the `customs` that are passed in as a parameter?  I don't think any appends will be reflected outside of the function, but I wonder if a line like above will end up modifying the original slice parameter's values.  If so, as currently set up, I think I would have expected the `customs` parameter to not be touched, but expect any modifications to be reflected in the return value (sort of like how `append` works).  It may not matter given how you're using this, but just something I wondered about when tracing through the code.

pkg/factory/shared.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+	// Store the validation result
+	controller.isValid = true
+	return customs
+
+}
+
+// uniqueCustomizations ensures there's only one customization per type.
+// Requirement: All customizations should already have a type assigned.
+func uniqueCustomizations(customs []Customization) ([]Customization, error) {
```
reggieriser, line 171: **nitpick:** Is there any reason to return the customization slice here?  I don't think we modify anything, but just check for duplicates, so returning just the error may be sufficient.

pkg/factory/traits.go
```diff
@@ -0,0 +1,41 @@
...
+import "github.com/transcom/mymove/pkg/models"
+
+// GetTraitActiveUser returns a customization to enable active on a user
+func GetTraitActiveUser() []Customization {
+	return []Customization{
+		{
+			Model: models.User{
+				Active: true,
+			},
+			Type: &User,
```
reggieriser, line 12: **question:** Given the new default type assignment, this would be optional to include now, right?

pkg/factory/factory_test.go
```diff
@@ -0,0 +1,498 @@
...
+		{
+			Model: models.ServiceMember{
+				Affiliation: &army,
+			},
+			Type: &ServiceMember,
+		},
+	}
+}
+
+func (suite *MakerSuite) TestMergeCustomization() {
```
reggieriser, line 48: **nitpick:** Should these tests (but not the test suite setup) go in a `shared_test.go` file since it seems to be testing things in `shared.go`?

pkg/factory/user_factory_test.go
```diff
@@ -0,0 +1,67 @@
...
+				Model: models.User{
+					LoginGovEmail: customEmail,
+				},
+			}}, []Trait{
+			GetTraitActiveUser,
+		})
+		suite.Equal(customEmail, user.LoginGovEmail)
+		suite.True(user.Active)
+	})
+
```
reggieriser, line 66: **thought:** Should we have a test for `BuildDefaultUser`?  It's a simple function, but it is a little different than your first test above since it sets the active user trait.

pkg/factory/shared.go
```diff
@@ -0,0 +1,387 @@
...
+	// If it exists and is valid, return, this list has been setup and validated
+	controller := controlCustom.Model.(controlObject)
+	if controller.isValid {
+		return customs
+	}
+
+	// If not valid:
+
+	// Make sure all customs have a proper type
+	setDefaultTypes(customs)
```
reggieriser, line 154: **praise:** Thanks for adding this!  I think it'll be nice to only be required to specify a model when it's one of the special cases.

reggieriser (commented):
This is looking good!🥇  I added a few comments for your consideration, but nothing major jumped out at me.

Wonder if it'd be worth having another example builder for something like `BuildServiceMember` (or an easier example if there is one) that would exercise the creation of related records and/or the custom types like `ResidentialAddress`.  But that may be scope creep for what you had in mind for this PR.

pkg/factory/user_factory.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"github.com/gobuffalo/pop/v6"
+	"github.com/gofrs/uuid"
+
+	"github.com/transcom/mymove/pkg/models"
+	"github.com/transcom/mymove/pkg/testdatagen"
+)
+
+// UserMaker is the base maker function to create a user
```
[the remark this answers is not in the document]
shimonacarvalho: Yup noticed my comments all had the older names. Will fix. 

About the BuildServiceMember, I had that in my spike. But yeah just trying to break up the tickets and also test the theory of iteratively replacing the layers of testdatagen.

pkg/factory/factory_test.go
shimonacarvalho: You're probably right, I just thought shared_test sounds so meaningless and we probably have others in the codebase. But it's better to stick by the pattern.

pkg/factory/shared.go
shimonacarvalho: Yes, in replacing the calls to `makeUser` it did seem nicer to have a similar format.

pkg/factory/user_factory_test.go
shimonacarvalho: Done.

pkg/factory/traits.go
shimonacarvalho: Yes, fixed.

pkg/factory/shared.go
shimonacarvalho: It does modify the `customs` that were passed in. 

It was way more complexity it seemed, to preserve `customs` and construct a brand new slice and watch for all the corner cases of traits without customs, and customs without traits. 

I see what you mean though - someone might create a customization list in a test and intend to use it repeatedly. Hmm if I made a copy though, I'd have to go one level further, the models too will have to be copied. See any gotchas there? 

We could also make people use a trait instead if they want to use a customization repeatedly? 

Thoughts?

reggieriser: The key thing may be to document any parameters that could have changes made to them and how they'd be changed.  Since `setupCustomizations` calls this and has a similar signature, it may need some annotation too.

Since this is an unexported function, consider this just some thoughts and not a blocker.

mlm55 (approved):
lgtm. thanks for all the explanatory comments!

pkg/factory/shared.go
shimonacarvalho: Ok since we don't have a good deep copy library at this point, I'm going to move forward with just documenting that the `customs` will be edited by these functions. And we can consider updating in a later ticket.

reggieriser (approved):

PR #9534, opened 2022-11-04
Revert PR #9390 (upgrade to Pop 6.0.8)

reggieriser:
## Summary

This PR reverts PR #9390, an upgrade to Pop 6.0.8.  The `deploy_stg_migrations` step was failing after merging the original PR, but it's not clear to me that the failure is definitely related to this upgrade.  Nevertheless, wanted to get this revert ready in case we need it.


rogeruiz:
@ahobson we should consider doing something about the ☝🏽 fact that Mergify uses your key rather than a more generic key soon. I'll look into it and report back on Slack somewhere ⌨️

rogeruiz (approved):
lgtm 🌈

@reggieriser This should be okay to merge. I've been following the conversation from Friday and am okay with this change. Thanks for catching this.

rogeruiz:
#9538 This may fix the issue that this revision is trying to solve.

PR #9538, opened 2022-11-07
Manually construct pop URL when using iampg auth

ahobson:
## Summary

After https://github.com/gobuffalo/pop/pull/774 iam password auth no longer works because pop is query escaping the password. Manually construct the pg url when iam auth is enabled

- [ ] Tested in the Loadtest environment (for changes to containers, app startup, or connection to data stores)


pkg/cli/dbconn.go
```diff
@@ -315,6 +314,14 @@ func InitDatabase(v *viper.Viper, creds *credentials.Credentials, logger *zap.Lo
...
 
 		dbConnectionDetails.Password = passHolder
+		// pop now use url.QueryEscape on the password, but that
+		// doesn't work with iampg. If we manually construct the URL,
+		// we can override that behavior
+		s := "postgres://%s:%s@%s:%s/%s?%s"
+		dbConnectionDetails.URL = fmt.Sprintf(s,
+			dbConnectionDetails.User, dbConnectionDetails.Password,
+			dbConnectionDetails.Host, dbConnectionDetails.Port,
+			dbConnectionDetails.Database, dbConnectionDetails.OptionsString(""))
```
ahobson, line 324: If we changed the `passHolder` to something that does not need query escaping, we wouldn't have to do this assignment to `URL`, but that seems like leaving a footgun around, even if we comment on it.

jesset-truss (commented):
@ahobson I think we might be moving quickly on this. I'm concerned this may have a downstream effect on to an issue that I'm currently working with others in the Platform team on an issue with iampostgres.go not being able to exit. Could we please discuss the desired outcome and how this will be testable and ensure no downstream effect?

rogeruiz:
Also, @jesset-truss I believe you're right here with the issue of how this will affect our DMS security solution with ECS. We should be leveraging the AppContext package and the DB Connection library from within our custom ECS solution. We should add this to our system architecture diagrams around our DMS ECS task solution.

ahobson:
> @ahobson I think we might be moving quickly on this. I'm concerned this may have a downstream effect on to an issue that I'm currently working with others in the Platform team on an issue with iampostgres.go not being able to exit. Could we please discuss the desired outcome and how this will be testable and ensure no downstream effect?

Sure, but this isn't changing any behavior that would affect iampostgres.go, it's just about escaping the placeholder password. The reason iampostgres doesn't quit is because the code is using go channels and the code in dbconn.go isn't providing a way to quit even though the iampostgres code provides that capability.

I feel these are unrelated issues, but can work through how to fix the shutdown issue with you if you want

reggieriser:
Talked to @ahobson and decided to cherry-pick my previous Pop 6.0.8 changes from #9390 (which we reverted in #9534) into this PR since we'll need his IAM fixes from this PR, the Pop 6.0.8 upgrade changes, and the fixes I previously made in #9390 for Pop 6.0.8.  We'll test all this together in experimental before we consider merging anything.

reggieriser (approved):
LGTM!  Walked through things with @ahobson and the latest code [gets past all checks](https://app.circleci.com/pipelines/github/transcom/mymove/46988/workflows/460c10ea-28e6-4515-afc2-36c8b7703cc0) on experimental.  Note that this PR has some of my code from #9390, but that had already been vetted in that PR.

PR #9546, opened 2022-11-07
MB-14347 Add closeout office column to moves

PDickeyTruss:
## Summary

This PR adds a field to the Move table to allow association of a closeout office. It also includes test data (in devseed) to create a move with this field populated (Move Locator = `CLSOFF`). 

Note: This story calls out not to name this field anything with the word 'location' in it. I went with `closeoutOffice` since the type of this field is a transportation office. Please let me know if anyone has opinions on this naming, I am happy to rename it if anyone feels strongly about it. 

## Testing/Validation
There are no changes visible in the app UI, this will be more of a code/db change type review. There is test data that can be loaded that will populate this new field in the database (see move w/ locator `CSLOFF`)
Screenshot 2022-11-07 at 2 40 18 PM

## To Verify
- [ ] New migration adds new column as expected
- [ ] Move model is updated to include CloseoutOffice 
- [ ] Test data is present with the new field populated

### DIFF
--- a/migrations/app/schema/20221104180553_add_closeout_office_move_association.up.sql
+++ b/migrations/app/schema/20221104180553_add_closeout_office_move_association.up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE moves ADD COLUMN closeout_office_id uuid REFERENCES transportation_offices(id);
--- a/migrations/app/schema/20221104180553_add_closeout_office_move_association.up.sql
+++ b/migrations/app/schema/20221104180553_add_closeout_office_move_association.up.sql
@@ -1,1 +1,3 @@
 ALTER TABLE moves ADD COLUMN closeout_office_id uuid REFERENCES transportation_offices(id);
+
+COMMENT on COLUMN moves.closeout_office_id IS 'The ID of the assiociated transportation office that is the closeout office for a move.';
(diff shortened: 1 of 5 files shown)


reggieriser (changes requested):
LGTM.  I ran the migration fine and was able to generate test data. 🥳  One thing we should do before merging though is add a database comment for this new field (see `COMMENT ON COLUMN`). 

Also, there is a `Validate` method on this model where we could call `OptionalUUIDIsPresent` on the new `CloseoutOfficeID` field if we wanted (like is done for `ExcessWeightUploadID`).  But it doesn't look like we've been particularly consistent with doing these types of validations on this model, so consider that an FYI and optional.

PDickeyTruss:
Great feedback items @reggieriser, thank you! Both have now been addressed 😁

migrations/app/schema/20221104180553_add_closeout_office_move_association.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE moves ADD COLUMN closeout_office_id uuid REFERENCES transportation_offices(id);
+
+COMMENT on COLUMN moves.closeout_office_id IS 'The ID of the assiociated transportation office that is the closeout office for a move.';
```
reggieriser, line 3: **nitpick**: `assiociated` -> `associated`

reggieriser (approved):
LGTM!  I see the database comment now after running migrations and the extra validation looks good.  Did notice one typo, but going ahead and approving.

migrations/app/schema/20221104180553_add_closeout_office_move_association.up.sql
PDickeyTruss: Thanks, fixed!  🦅 👁️

PR #9566, opened 2022-11-10
MB-13883 turn on the flag to audit race conditions during build and test time

YanZ777:
## Summary

Reenable the checks for the race condition flags. 

Please check that I'm adding the flags in the correct spots / it makes sense to add them into those spots. 

I'm only adding the flag to the tests rather than the build since adding the flag to the build would incur a large memory and performance cost. 

The covermode for the tests had to be changed to atomic since count is not supported when the -race flag is set.
[This article](https://github.com/go-playground/overalls/issues/19) explains a bit of as to why tests can only be run in atomic when the -race flag is set.

[This article](https://go.dev/doc/articles/race_detector) explains more about go race detectors.

## Setup to Run Your Code

💻 You will need to use a terminals to test this locally.

### Additional steps

1. You'll want to run `make server_test`. There should be no errors output from running that command.


Makefile
```diff
@@ -267,7 +267,7 @@ bin/iws: cmd/iws
 PKG_GOSRC := $(shell find pkg -name '*.go')
 
 bin/milmove: $(shell find cmd/milmove -name '*.go') $(PKG_GOSRC)
-	go build -gcflags="$(GOLAND_GC_FLAGS) $(GC_FLAGS)" -asmflags=-trimpath=$(GOPATH) -ldflags "$(LDFLAGS) $(WEBSERVER_LDFLAGS)" -o bin/milmove ./cmd/milmove
+	go build -race -gcflags="$(GOLAND_GC_FLAGS) $(GC_FLAGS)" -asmflags=-trimpath=$(GOPATH) -ldflags "$(LDFLAGS) $(WEBSERVER_LDFLAGS)" -o bin/milmove ./cmd/milmove
```
reggieriser, line 270: **question:** Does this mean that race detection would be compiled into the milmove binary that we end up deploying?  I think it might, but I may not be following everything correctly in the build.

If so, according to [the last paragraph here](https://go.dev/doc/articles/race_detector): `The cost of race detection varies by program, but for a typical program, memory usage may increase by 5-10x and execution time by 2-20x.`  That sounds significant.  Would it instead be sufficient to just compile it just into the binaries that we test against?

YanZ777: Yea, that is a pretty big memory usage jump. Just so that I understand correctly, are you asking if it is possible to compile the code with that flag turned on only during test time?

reggieriser: Note that it's not just memory usage, but execution time too.  What I'm wondering is if we are required to turn this on in production and suffer these significant memory/performance hits overall, or do we feel it's sufficient to just run a `go test -race` locally and in CI to see if that turns up any race condition warnings?  If the latter, then I don't think we need to add the `-race` to the building of the `milmove` executable.  Here's some guidance from the [previously linked article](https://go.dev/doc/articles/race_detector):

> To start, run your tests using the race detector (go test -race). The race detector only finds races that happen at runtime, so it can't find races in code paths that are not executed. If your tests have incomplete coverage, you may find more races by running a binary built with -race under a realistic workload.

ahobson: I agree with @reggieriser that I don't think we want to turn this on for all builds. I think starting with enabling it for our server tests would be a good start. From reading the POAM, I think that would be sufficient

YanZ777: After rereading the Epic, I think just adding it to the server tests would be sufficient. I'll make those changes. Thanks for both your feedback!

scripts/run-server-test
```diff
@@ -65,7 +66,7 @@ fi
 # Add coverage tracker via go cover
 if [[ "${COVERAGE:-}" == "1" ]]; then
   coverpkg=$(echo "${package_list[@]}" | paste -s -d , - | tr '[:blank:]' ',')
-  gotest_args+=(-coverprofile=coverage.out -covermode=count -coverpkg="${coverpkg}")
+  gotest_args+=(-coverprofile=coverage.out -covermode=atomic -coverpkg="${coverpkg}")
```
reggieriser, line 69: **question:** Can you tell me a little more about this change?  I noticed this from the Go blog:

> The default is ‘set’, which we’ve already seen. The atomic setting is needed only when accurate counts are required when running parallel algorithms. It uses atomic operations from the [sync/atomic](https://go.dev/pkg/sync/atomic/) package, which can be quite expensive. For most purposes, though, the count mode works fine and, like the default set mode, is very cheap.

I'm going to try a couple of local (unscientific) tests to see how much of a difference `atomic` seems to make.  Just wondering because we run this in CI on every build I believe.  How important is it for the counts to be 100% accurate in every situation?  I know we want to know if each statement ran (`set` mode), but is a precise count of how many times a statement ran something we use/need?

reggieriser: I ran `time COVERAGE=1 SERVER_REPORT=1 make clean server_test` locally (to try to match what I see in the CircleCI config) and this is what I found:

- Current main (no race, but count mode): 226.14s user 89.62s system 184% cpu 2:51.42 total
- Your branch (with race, but changed to use `count` mode): error `-covermode must be "atomic", not "count", when -race is enabled`
- Your branch (with race, but changed to use `set` mode): error `-covermode must be "atomic", not "set", when -race is enabled`
- Your branch (as it currently is with race/atomic): 985.64s user 259.68s system 422% cpu 4:54.52 total

Oh!  I now see in the PR description that you mention why you had to switch to `atomic`.  Sorry about that!  🤦‍♂️  I was hoping that `set` might be an option, but that doesn't work either.

Unfortunately, it looks like our server test duration will increase quite a bit.

YanZ777: No worries, I added that a bit later, so it is entirely possible that you may have missed it! 
I've linked [this article](https://github.com/go-playground/overalls/issues/19) that explains why the covermode must be atomic when -race flag is set.
And yea, downsides of doing any form of race condition checking. 😞

reggieriser (commented):
@YanZ777 I see four more usages of `go test` in the `Makefile`.  Looks like they're used when running acceptance tests and a webhook client test.  Do we need to add`-race` to those as well?

YanZ777:
> @YanZ777 I see four more usages of `go test` in the `Makefile`. Looks like they're used when running acceptance tests and a webhook client test. Do we need to add`-race` to those as well?

I missed those! Thanks for pointing them out. I'm actually not sure if the -race flag needs to be added to those. I'm trying to figure out when those tests get run / how to trigger them. [The documentation on how to run the acceptance tests](https://transcom.github.io/mymove-docs/docs/backend/testing/run-go-tests#run-acceptance-tests) isn't working for me.

YanZ777:
> @YanZ777 I see four more usages of `go test` in the `Makefile`. Looks like they're used when running acceptance tests and a webhook client test. Do we need to add`-race` to those as well?

After further investigation, I don't think `-race` needs to be added to those.
The `acceptance_test` job only runs when folks turn it on, I've only seen this happen in experimental, and it usually fails. No one that I've asked knows what the tests are running. I will be making a separate Jira ticket / PR to remove those tests completely.
Similarly, with the `webhook_client_test`, those are not being run either and I thus I do not think that the `-race` flag needs to be added to them.

reggieriser (approved):
LGTM!  Thanks for all the follow-up on my questions.🥇

PR #9588, opened 2022-11-14
Allow the server to start even if the db connection is not ready

ahobson:
## Summary
Similar change to 73d13c009e8b025da56730b7cbdd930462574448 but in the server code. This allows the server to come up if the db is not ready

### DIFF
--- a/cmd/generate-payment-request-edi/main.go
+++ b/cmd/generate-payment-request-edi/main.go
@@ -88,8 +88,6 @@ func main() {
 	// DB connection
 	dbConnection, err := cli.InitDatabase(v, nil, logger)
 	if err != nil {
-		// No connection object means that the configuraton failed to validate and we should not startup
-		// A valid connection object that still has an error indicates that the DB is not up and we should not startup
 		logger.Fatal("Connecting to DB", zap.Error(err))
 	}
 
--- a/cmd/milmove/serve.go
+++ b/cmd/milmove/serve.go
@@ -366,9 +366,14 @@ func initializeDB(v *viper.Viper, logger *zap.Logger,
 	}
 
 	// Create a connection to the DB
-	dbConnection, errDbConnection := cli.InitDatabase(v, dbCreds, logger)
-	if errDbConnection != nil {
-		logger.Warn("DB is not ready for connections", zap.Error(errDbConnection))
+	dbConnection, err := cli.InitDatabase(v, dbCreds, logger)
+	if err != nil {
+		logger.Fatal("Invalid DB Configuration", zap.Error(err))
+	}
+
+	err = cli.PingPopConnection(dbConnection, logger)
+	if err != nil {
+		logger.Warn("DB is not ready for connections", zap.Error(err))
 	}
 
 	return dbConnection
--- a/cmd/milmove/serve.go
+++ b/cmd/milmove/serve.go
@@ -373,6 +373,8 @@ func initializeDB(v *viper.Viper, logger *zap.Logger,
 
 	err = cli.PingPopConnection(dbConnection, logger)
 	if err != nil {
+		// if the db is not up yet, the server can still start. This
+		// prevents a failure loop when deploying containers
 		logger.Warn("DB is not ready for connections", zap.Error(err))
 	}
 
--- a/cmd/milmove/serve.go
+++ b/cmd/milmove/serve.go
@@ -368,14 +368,7 @@ func initializeDB(v *viper.Viper, logger *zap.Logger,
 	// Create a connection to the DB
 	dbConnection, errDbConnection := cli.InitDatabase(v, dbCreds, logger)
 	if errDbConnection != nil {
-		if dbConnection == nil {
-			// No connection object means that the configuraton failed to validate and we should kill server startup
-			logger.Fatal("Invalid DB Configuration", zap.Error(errDbConnection))
-		} else {
-			// A valid connection object that still has an error indicates that the DB is not up but we
-			// can proceed (this avoids a failure loop when deploying containers).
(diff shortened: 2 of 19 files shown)


cmd/milmove/serve.go
```diff
@@ -368,14 +368,7 @@ func initializeDB(v *viper.Viper, logger *zap.Logger,
...
 	if errDbConnection != nil {
-		if dbConnection == nil {
-			// No connection object means that the configuraton failed to validate and we should kill server startup
-			logger.Fatal("Invalid DB Configuration", zap.Error(errDbConnection))
-		} else {
-			// A valid connection object that still has an error indicates that the DB is not up but we
-			// can proceed (this avoids a failure loop when deploying containers).
-			logger.Warn("DB is not ready for connections", zap.Error(errDbConnection))
-		}
+		logger.Warn("DB is not ready for connections", zap.Error(errDbConnection))
```
reggieriser, line 371: I think you've found the correct root cause.  But I'm wondering though what happens if we have an invalid DB configuration.  I think it will now just warn, put an error message in the log about the config issue, but presumably the server will keep running whereas it was killed before?  Is that OK?

ahobson: Actually, I think I should split what `InitDatabase` does into two functions: One to initialize the database connection object, which if it errors is fatal and then another function that pings the database. That way the client can choose what to do if the db is not pingable. I'm going to do that and see what it looks like

cmd/milmove/serve.go
```diff
@@ -366,16 +366,14 @@ func initializeDB(v *viper.Viper, logger *zap.Logger,
...
 
 	// Create a connection to the DB
-	dbConnection, errDbConnection := cli.InitDatabase(v, dbCreds, logger)
-	if errDbConnection != nil {
-		if dbConnection == nil {
-			// No connection object means that the configuraton failed to validate and we should kill server startup
-			logger.Fatal("Invalid DB Configuration", zap.Error(errDbConnection))
-		} else {
-			// A valid connection object that still has an error indicates that the DB is not up but we
-			// can proceed (this avoids a failure loop when deploying containers).
```
reggieriser, line 376: We may want to keep this comment to explain why we `Warn` here as opposed to `Fatal` like in other places.

cmd/milmove/serve.go
ahobson: The other alternative I've thought of to what I implemented is having a `cli.InitDatabase` and a `cli.InitDatabaseWithPing`. The former wouldn't try to ping the db, but the latter would and return an error if the ping failed.

Everything except the server would use `InitDataBaseWithPing`.

@reggieriser any thoughts on which one you would prefer?

pkg/cli/dbconn.go
```diff
@@ -388,24 +388,29 @@ func InitDatabase(v *viper.Viper, creds *credentials.Credentials, logger *zap.Lo
...
 	}
 
-	dbWithPinger, ok := connection.Store.(pinger)
+	// Return the open connection
+	return connection, nil
+}
+
+// PingPopConnection pings the database and returns an error if it is
+// not reachable
+func PingPopConnection(c *pop.Connection, logger *zap.Logger) error {
```
reggieriser, line 397: Do we need a test for this now that we've separated it out?  I see a test for `InitDatabase`, although I'm not sure if it's enabled or not -- it seems to look at an environment variable to determine if it should be run.

reggieriser (commented):
I like the idea of separating out the ping part!  I added a few comments.  Also thinking we may want to deploy to experimental once we think we're done with PR changes.

cmd/generate-payment-request-edi/main.go
```diff
@@ -88,8 +88,6 @@ func main() {
 	// DB connection
 	dbConnection, err := cli.InitDatabase(v, nil, logger)
 	if err != nil {
-		// No connection object means that the configuraton failed to validate and we should not startup
-		// A valid connection object that still has an error indicates that the DB is not up and we should not startup
```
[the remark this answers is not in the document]
ahobson: When looking at the code for the command line tools, it sure seems to me that if the database isn't available, the tool will fail almost instantly the first time it tries to do anything with the db. To me, that seems just as good as pinging, but maybe you see it differently?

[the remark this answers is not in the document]
reggieriser: Gotcha -- that seems OK to me.

pkg/cli/dbconn.go
ahobson: Good idea! I've added tests

cmd/milmove/serve.go
ahobson: Good idea!

cmd/milmove/serve.go
reggieriser: I don't feel particularly strongly one way or the other, but I think I lean a little toward how you have it set up now.  Make pinging an explicit, separate function.  That also means you can more explicitly ping in a loop when it's warranted like we do with migrations.  I agree that there's not much point in pinging in the CLIs.

reggieriser (approved):
LGTM -- great job! 🥇   I'm approving, pending the tests passing and a successful deploy to experimental (since we changed code around the database initialization/pinging that bit us the other day).

reggieriser:
I also just kicked off an ephemeral deploy of this branch to make sure that works now ... https://trussworks.slack.com/archives/C01T40WC4G1/p1668529662626229

ahobson:
[Deploying to experimental now](https://app.circleci.com/pipelines/github/transcom/mymove/47241/workflows/e74361f0-673d-4493-b651-e822d9adf30f)

PR #9607, opened 2022-11-16
Filter additional test users from the customer local login

ronaktruss:
## Summary

This removes the users associated with the following emails: `needscloseout@ppm.closeout`, `1needscloseout@ppm.closeout`, `2needcloseout@ppm.closeout`, `needscloseout@ppmHHG.closeout` from the local login for the customer app, because they are used primarily to create data for the office application.

## Setup to Run Your Code

<details>

##### Terminal 1

Update your local dev seed data

```sh
 make db_dev_e2e_populate
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Before
image
### After
You should no longer see those users

### DIFF
--- a/pkg/models/user.go
+++ b/pkg/models/user.go
@@ -218,7 +218,7 @@ func FetchAppUserIdentities(db *pop.Connection, appname auth.Application, limit 
 				sm.middle_name AS sm_middle
 			FROM service_members as sm
 			JOIN users on sm.user_id = users.id
-			WHERE users.login_gov_email != 'first.last@login.gov.test'
+			WHERE users.login_gov_email NOT IN ('first.last@login.gov.test', 'needscloseout@ppm.closeout', '1needscloseout@ppm.closeout', '2needcloseout@ppm.closeout', 'needscloseout@ppmHHG.closeout')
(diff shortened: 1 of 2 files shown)


pkg/models/user.go
```diff
@@ -218,7 +218,7 @@ func FetchAppUserIdentities(db *pop.Connection, appname auth.Application, limit
 				sm.middle_name AS sm_middle
 			FROM service_members as sm
 			JOIN users on sm.user_id = users.id
-			WHERE users.login_gov_email != 'first.last@login.gov.test'
+			WHERE users.login_gov_email NOT IN ('first.last@login.gov.test', 'needscloseout@ppm.closeout', '1needscloseout@ppm.closeout', '2needcloseout@ppm.closeout', 'needscloseout@ppmHHG.closeout')
```
duncan-truss, line 221: If we're filtering them out of the customer local login, we could just change them to use `first.last@login.gov.test` as the user email right in the testdatagen? I don't think we're relying on the email elsewhere in the office side, they can still modify the first and last name and move code.

PR #9623, opened 2022-11-18
[MB-14420] Add a certificate validity check to circleCI workflow

chtakahashi:
The motivation here is to shorten the feedback loop from updating a certificate to knowing whether its valid or not in the CircleCI workflow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1764,7 +1764,9 @@ workflows:
       - pre_deps_golang
 
       - pre_deps_yarn
-
+      
+      - check_tls_certificate_prd
+      
       - check_generated_code:
           requires:
             - pre_deps_golang
@@ -1776,7 +1778,6 @@ workflows:
           filters:
             branches:
               only: main
-
       - pre_test:
           requires:
             - pre_deps_golang
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -729,6 +729,31 @@ jobs:
       - run: scripts/check-generated-code pkg/gen/ $(find . -type d -name "*mocks" -exec echo -n '{} ' \;)
       - announce_failure
 
+  # `check_tls_certificate_env` is used to confirm that the certificate-key pair match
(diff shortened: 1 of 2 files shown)


scripts/check-tls-pair
```diff
@@ -0,0 +1,19 @@
...
+#   check-tls-pair confirms that the TLS cert and key pair are a match.
+#   This will allow us to trigger a failure quickly instead of finding out that the keypair doesn't work 30 minutes into the deploy.
+#
+set -eu -o pipefail
+
+key=$1
+cert=$2
+
+KEY_FILE=$(echo $1 | base64 --decode)
+CERT_FILE=$(echo $2 | base64 --decode)
```
rogeruiz, line 12: Love it. We've gotta base64 decode what we base64 encode. Since these are multi-line variables they have to be base64 encoded according to the docs.

[=> The CircleCI docs on Setting environment variables](https://circleci.com/docs/set-environment-variable/#encoding-multi-line-environment-variables)

.circleci/config.yml
```diff
@@ -1739,7 +1764,9 @@ workflows:
       - pre_deps_golang
 
       - pre_deps_yarn
-
+      
+      - check_tls_certificate_dp3
+
```
rogeruiz, line 1769: I'm so excited for this to be happening so early on in our process!

rogeruiz (approved):
lgtm 🌈

Great work on this @chtakahashi. Love the direction you took on this.

rogeruiz:
Okay this should now be working. But since I'm the last person committing here, I'm not going to review this. Change the code anyway you'd like. 👋🏽

rogeruiz:
At a quick glance, I see two things wrong here:

- the `openssl` commands may not be accurate. But I'm not sure. It's complaining about not having the correct files.
- when this fails, the rest of the tests still run. So we need to add the check everywhere else in the CI config file so the workflow doesn't keep going.

I'll leave those for y'all to figure out.

PR #9624, opened 2022-11-18
[MB-14525] Normalize serialized json timestamp fractional second precision in orders audit log test

duncan-truss:
## Summary

The server_test was failing for the move history test around acknowledging amended orders by the office user.

Screen Shot 2022-11-21 at 10 27 00 AM

The first thing I noticed was that the timestamp was initialized to the zero value of time.Time (0001-01-01 00:00:00), meaning that something wasn't getting parsed properly.  We were ignoring the potential error coming back from time.Parse() and so I began asserting there was no error.  This revealed that often we'd be trying to parse a string with a different precision of fractional seconds from the json changed data.  At first I assumed maybe this was a difference between local and CI environments but would still see the occasional error.

This led me to realize maybe there was something happening in the conversion between database timestamps and the value that got serialized into the audit history changed data column.  Serialization happens using the `row_to_json` Postgres function.  You can test this in your database query console to see how it behaves:

`select row_to_json(orders.*) from orders where id = 'some orders id'`

If you edit one of the orders timestamps (created_at, updated_at, amended_orders_acknowledged_at) to have trailing zeros for the microseconds (`2022-11-15 17:44:54.321000`) you'll see that it gets converted to `2022-11-15T17:44:54.321`.

This means to fix the test we either need to remove the fractional seconds or normalize it to always include 6 decimal places to guarantee parsing in the same format.  I went the normalization route, which is slightly more involved just to fix a test not application code.  I think leaving an artifact of what is happening could help document some of the behavior in the future.  It's worth debating how valuable it is that the timestamp is within a particular duration at all as long as it's set properly and in the format we want (not really?).  I assume we didn't get bit by this elsewhere because `created_at` and `updated_at` timestamps are ignored in the change log usually.

https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-JSON-CREATION-TABLE
https://github.com/transcom/mymove/blob/main/migrations/app/schema/20220201170324_audit_trigger.up.sql

##### Terminal 1

Run the Move History Fetcher server tests

```sh
go test ./pkg/services/move_history -testify.m TestMoveFetcher
```

</details>

### Frontend

### Backend


pkg/services/move_history/move_history_fetcher_test.go
```diff
@@ -688,7 +708,7 @@ func (suite *MoveHistoryServiceSuite) TestMoveFetcherWithFakeData() {
 		moveRouter := moverouter.NewMoveRouter()
 		creator := mtoserviceitem.NewMTOServiceItemCreator(builder, moveRouter)
 
-		reService := testdatagen.MakeReService(suite.DB(), testdatagen.Assertions{
+		reService := testdatagen.FetchOrMakeReService(suite.DB(), testdatagen.Assertions{
```
duncan-truss, line 711: There is a unique constraint for re_service item codes and running test packages directly would error out trying to create a duplicate service item because rows were not truncated and rolled back.

duncan-truss:
Ahh I see `strings.Repeat()` exists if it's worth getting rid of the messier format string stuff

pkg/services/move_history/move_history_fetcher_test.go
```diff
@@ -285,13 +286,32 @@ func (suite *MoveHistoryServiceSuite) TestMoveFetcher() {
...
+				amendedAcknowledgedAt, ok := changedData["amended_orders_acknowledged_at"].(string)
+				if !ok {
+					suite.Fail("casting changedData amendedOrdersAcknowledgedAt to string value failed")
+				} else {
+					// separate the fractional seconds part of the timestamp
+					parts := strings.Split(amendedAcknowledgedAt, ".")
+					if len(parts) > 1 {
+						// create format string to generate the missing zeros if any ("%05s")
+						fillFormat := "%0" + fmt.Sprintf("%ds", 6-len(parts[1]))
+						// combine the existing fractional seconds with the filled zeros ("900000")
```
duncan-truss, line 301: I could replace this with `strings.Repeat()` now that I see that exists if this is too confusing.

kctruss (approved):
Thank you for tackling this, @duncan-truss . That is interesting that we are testing for the timestamp range. That's also super helpful information to know about the `row_to_json` idiosyncrasy.

duncan-truss:
> 

Yea, we've done that with other timestamps in the past to varying degrees of success.  Unless a timestamp is getting overwritten, it is usually enough to assert that it is a non-empty value I would think most of the time.

scottmries (approved):

PR #9633, opened 2022-11-21
MB-14214 customer can enter pro gear details integration

scottmries:
## Summary

This PR integrates the Pro Gear weight tickets form with the backend. Users should be able to:
- Create a new weight ticket from the Review page by clicking "Add Pro-Gear Weight."
- Select the pro-gear owner, add/edit a description, set a weight, and toggle whether they have tickets.
- Upload a delete weight ticket documents.
- Save these changes.

Just a note that some ancillary tests were failing because of the change from using empty/full weights on pro-gear to just using weight, so this PR includes fixes for those, in addition to the primary integration tests.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the Customer site
2. Login as a user who is ready to close out (I use `progearWeightTicket@ppm.approved`)
3. Navigate to the Review page
4. Add and delete uploads, and set and update all other fields
5. See errors for weights entered that exceed the allowances

### DIFF
--- a/pkg/handlers/internalapi/uploads.go
+++ b/pkg/handlers/internalapi/uploads.go
@@ -179,7 +179,7 @@ func (h CreatePPMUploadHandler) Handle(params ppmop.CreatePPMUploadParams) middl
 			documentID := uuid.FromStringOrNil(params.DocumentID.String())
 
 			// Fetch document to ensure user has access to it
-			document, docErr := models.FetchDocument(appCtx.DB(), appCtx.Session(), documentID, false)
+			document, docErr := models.FetchDocument(appCtx.DB(), appCtx.Session(), documentID, true)
 			if docErr != nil {
 				docNotFoundErr := fmt.Errorf("documentId %q was not found for this user", documentID)
 				return ppmop.NewCreatePPMUploadNotFound().WithPayload(payloads.ClientError(handlers.NotFoundMessage, docNotFoundErr.Error(), h.GetTraceIDFromRequest(params.HTTPRequest))), docNotFoundErr
--- a/src/components/Customer/PPM/Closeout/FinalCloseoutForm/FinalCloseoutForm.test.jsx
+++ b/src/components/Customer/PPM/Closeout/FinalCloseoutForm/FinalCloseoutForm.test.jsx
@@ -26,10 +26,7 @@ describe('FinalCloseoutForm component', () => {
 
     const weightTicket = createCompleteWeightTicket({ serviceMemberId }, { emptyWeight: 14000, fullWeight: 18000 });
     const movingExpense = createCompleteMovingExpense({ serviceMemberId }, { amount: 30000 });
-    const proGearWeightTicket = createCompleteProGearWeightTicket(
-      { serviceMemberId },
-      { emptyWeight: 14000, fullWeight: 16000 },
-    );
+    const proGearWeightTicket = createCompleteProGearWeightTicket({ serviceMemberId }, { weight: 1500 });
 
     const mtoShipment = createPPMShipmentWithFinalIncentive({
       ppmShipment: {
@@ -52,7 +49,7 @@ describe('FinalCloseoutForm component', () => {
     const findListItemWithText = prepListSearchForItem(screen.getAllByRole('listitem'));
 
     expect(findListItemWithText('4,000 lbs total net weight')).toBeInTheDocument();
-    expect(findListItemWithText('2,000 lbs of pro-gear')).toBeInTheDocument();
+    expect(findListItemWithText('1,500 lbs of pro-gear')).toBeInTheDocument();
     expect(findListItemWithText('$300.00 in expenses claimed')).toBeInTheDocument();
 
     expect(
@@ -74,10 +71,9 @@ describe('FinalCloseoutForm component', () => {
       createCompleteMovingExpense({ serviceMemberId }, fieldOverrides),
     );
 
-    const proGearWeightTickets = [
-      { emptyWeight: 15000, fullWeight: 15500 },
-      { emptyWeight: 15000, fullWeight: 16000 },
-    ].map((fieldOverrides) => createCompleteProGearWeightTicket({ serviceMemberId }, fieldOverrides));
+    const proGearWeightTickets = [{ weight: 750 }, { weight: 750 }].map((fieldOverrides) =>
+      createCompleteProGearWeightTicket({ serviceMemberId }, fieldOverrides),
+    );
 
     const mtoShipment = createPPMShipmentWithFinalIncentive({
       ppmShipment: {
--- a/src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
+++ b/src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
@@ -61,8 +61,10 @@ const ProGearForm = ({ proGear, setNumber, onSubmit, onBack, onCreateUpload, onU
                 <h2>Set {setNumber}</h2>
                 <FormGroup>
                   <Fieldset>
-                    <legend>Who does this pro-gear belongs to?</legend>
-                    <Hint className={ppmStyles.hint}>You have to separate yours and your spouse&apos;s pro-gear.</Hint>
+                    <label htmlFor="selfProGear" className={classnames('usa-label', styles.descriptionTextField)}>
+                      Who does this pro-gear belong to?
+                      <Hint className={styles.hint}>You have to separate yours and your spouse&apos;s pro-gear.</Hint>
+                    </label>
                     <Field
                       as={Radio}
                       id="ownerOfProGearSelf"
@@ -90,7 +92,7 @@ const ProGearForm = ({ proGear, setNumber, onSubmit, onBack, onCreateUpload, onU
                         label="Brief description of the pro-gear"
                         labelHint={
                           <Hint className={styles.hint}>
-                            Examples of pro-gear includes specialized apparel and government issued equipment. Check the{' '}
+                            Examples of pro-gear include specialized apparel and government&ndash;issued equipment.
                             {jtr} for examples of pro-gear.
                           </Hint>
                         }
--- a/src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
+++ b/src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
@@ -21,14 +21,6 @@ import ppmStyles from 'components/Customer/PPM/PPM.module.scss';
 import SectionWrapper from 'components/Customer/SectionWrapper';
 import MaskedTextField from 'components/form/fields/MaskedTextField/MaskedTextField';
 
-const validationSchema = Yup.object().shape({
-  selfProGear: Yup.bool().required('Required'),
-  proGearDocument: Yup.object().nullable(),
-  proGearWeight: Yup.number().required('Required').min(0, 'Enter a weight 0 lbs or greater'),
-  description: Yup.string().required('Required'),
-  missingWeightTicket: Yup.string(),
-});
-
 const proGearDocumentRef = createRef();
 
 const ProGearForm = ({
@@ -42,6 +34,25 @@ const ProGearForm = ({
   proGearEntitlements,
 }) => {
   const { selfProGear, document, proGearWeight, description, missingWeightTicket } = proGear || {};
+
+  const validationSchema = Yup.object().shape({
+    selfProGear: Yup.bool().required('Required'),
+    proGearDocument: Yup.object().nullable(),
+    proGearWeight: Yup.number()
+      .required('Required')
+      .min(0, 'Enter a weight 0 lbs or greater')
+      .when('selfProGear', (selfProGearField, schema) => {
+        if (selfProGearField === null) {
+          return schema;
+        }
+        if (selfProGearField) {
+          return schema.max(proGearEntitlements.proGear);
+        }
+        return schema.max(proGearEntitlements.proGearSpouse);
+      }),
+    description: Yup.string().required('Required'),
+    missingWeightTicket: Yup.string(),
+  });
 
   let proGearValue;
   if (selfProGear === true) {
@@ -68,14 +79,14 @@ const ProGearForm = ({
   return (
     <Formik initialValues={initialValues} validationSchema={validationSchema} onSubmit={onSubmit}>
       {({ handleSubmit, isValid, isSubmitting, setFieldValue, values, ...formikProps }) => {
-        const entitlement = () => {
+        const getEntitlement = () => {
+          if (values.selfProGear === null) {
+            return null;
+          }
           if (values.selfProGear === 'true') {
-            return formatWeight(proGearEntitlements.proGear);
+            return proGearEntitlements.proGear;
           }
-          if (values.selfProGear === 'false') {
-            return formatWeight(proGearEntitlements.proGearSpouse);
-          }
-          return null;
+          return proGearEntitlements.proGearSpouse;
         };
         return (
           <div className={classnames(ppmStyles.formContainer, styles.ProGearForm)}>
@@ -129,7 +140,11 @@ const ProGearForm = ({
                         defaultValue="0"
                         name="proGearWeight"
                         label="Shipment's pro-gear weight"
-                        labelHint={<Hint className={styles.hint}>Your maximum allowance is {entitlement()}.</Hint>}
+                        labelHint={
+                          <Hint className={styles.hint}>
+                            Your maximum allowance is {formatWeight(getEntitlement())}.
+                          </Hint>
+                        }
                         id="proGearWeight"
                         mask={Number}
                         scale={0} // digits after point, 0 for integers
--- a/src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
+++ b/src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
@@ -1,4 +1,5 @@
 import React, { createRef } from 'react';
+import { connect } from 'react-redux';
 import * as Yup from 'yup';
 import { Field, Formik } from 'formik';
 import { func, number } from 'prop-types';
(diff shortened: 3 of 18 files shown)


src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
```diff
@@ -178,8 +208,15 @@ ProGearForm.propTypes = {
...
 
-export default ProGearForm;
+const mapStateToProps = (state) => {
+  const proGearEntitlements = selectProGearEntitlements(state);
+  return {
+    proGearEntitlements,
+  };
+};
+
+export default connect(mapStateToProps)(ProGearForm);
```
ronaktruss, line 222: (Suggestion): We may want to look into using `useSelector` instead of mapStateToProps. 
[General documentation for useSelector](https://react-redux.js.org/api/hooks#useselector).
Eg. usage in [WeightTicket.jsx](https://github.com/transcom/mymove/blob/main/src/pages/MyMove/PPM/Closeout/WeightTickets/WeightTickets.jsx#L25)

scottmries: Thanks for that, I updated it in the next commit.

rileyorr:
@scottmries great work! I just noticed a couple things. I'm not sure if they're in scope for this ticket, so I've categorized them as what I think may be in scope and out of scope. Let me know if you have any questions!

**Findings that may be in scope for this ticket:**
- When I edit the pro-gear, change the pro-gear weight to ‘0’ and select ‘Save & Continue’ I’m not shown an error to enter a weight greater than 0
   - There’s an error document linked in the epic that should capture this (I can’t currently access it because Confluence/Jira appear to be down). 

----

**Findings that may be out of scope for this ticket:**
- When I add a new pro-gear and select either ‘Me’ or ‘Spouse’, enter a shipment weight less than 2000 (in this case I entered 20), and select ‘Save & Continue’ I’m not shown an error notifying me to upload a weight ticket 
    - This error handling may be outside of this ticket’s scope since it’s related to the file upload component 

- When I add some pro-gear information and return to the home page, the pro-gear data I've entered appears on the 'Review' page. @jenniferlynparsons I think the team had a discussion when I first joined that resulted in the capability to delete items being removed from the MVP. Is there a ticket that captures how we're going to handle that? I tried searching for it but Jira is down for me. I see a potential scenario in which someone adds incorrect pro-gear data, can't remove it, and is forced to submit it. 

Video for reference:

https://user-images.githubusercontent.com/46329147/203412842-d13fb2d6-0c15-4354-8d30-a11128f37fef.mov

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
```diff
@@ -18,80 +21,110 @@ import ppmStyles from 'components/Customer/PPM/PPM.module.scss';
...
+        let maximum;
+        if (belongsToSelfField) {
+          maximum = proGearEntitlements.proGear;
+        } else {
+          maximum = proGearEntitlements.proGearSpouse;
+        }
+        return schema.max(maximum, `Pro gear weight must be less than or equal to ${formatWeight(maximum)}.`);
+      }),
+    description: Yup.string().required('Required'),
+    hasWeightTickets: Yup.string(),
```
ronaktruss, line 50: I might be a little confused about the usage of `hasWeightTickets` and `missingWeightTicket`, but I think we want this to be `missingWeightTicket` since that is the value that we are using in the form. I'm not sure if we should be using string or bool here, but based on the info below I don't think it actually makes a difference atm.

I also just found out that using `.required` effectively just prevents null and undefined (and '' for strings).

ronaktruss (commented):
Running through a few scenarios this looks great! 🎆 

I left some suggestions, but still working my through the tests.

scottmries:
@rileyorr I added a validation to weight so that 0 is invalid and the error is shown. Just to be clear: should the button be disabled when the form is valid, or should the form not trigger validation until the user attempts to save?

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.test.jsx
```diff
@@ -2,58 +2,78 @@ import React from 'react';
...
+    it('selects "Me" radio when belongsToSelf is true', () => {
+      const { container } = render(<ProGearForm {...defaultProps} {...proGearProps} />, { wrapper: MockProviders });
       expect(screen.getByLabelText('Me')).toBeChecked();
       expect(screen.getByLabelText('My spouse')).not.toBeChecked();
       expect(container).toHaveTextContent("You have to separate yours and your spouse's pro-gear.");
     });
     // TODO: Move test to WeightTicketUpload.test.jsx
     // it('populates form when weight ticket is missing', async () => {
-    //   render(<ProGearForm {...defaultProps} {...selfProGearProps} />);
+    //   render(<ProGearForm {...defaultProps} {...belongsToSelfProps} />);
```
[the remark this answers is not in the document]
scottmries: I moved this into a file to test that component and got it to pass.

Do you have a sense of why I'm getting warnings about wrapping tests that affect Formik state in `act(())`? I don't see this anywhere in the codebase, but have a couple errors now.

pkg/handlers/internalapi/uploads.go
```diff
@@ -179,7 +179,7 @@ func (h CreatePPMUploadHandler) Handle(params ppmop.CreatePPMUploadParams) middl
 			documentID := uuid.FromStringOrNil(params.DocumentID.String())
 
 			// Fetch document to ensure user has access to it
-			document, docErr := models.FetchDocument(appCtx.DB(), appCtx.Session(), documentID, false)
+			document, docErr := models.FetchDocument(appCtx.DB(), appCtx.Session(), documentID, true)
```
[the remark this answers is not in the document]
scottmries: @duncan-truss and I spoke about this, and before, if a document had no uploads, this would return a nil document. On the front-end, this meant that, if I had deleted all the uploads, left the page, returned, and tried to upload, the upload endpoint no longer had a document to upload to, and would error. This ensures that we always have a document.

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
```diff
@@ -18,80 +21,110 @@ import ppmStyles from 'components/Customer/PPM/PPM.module.scss';
...
+  const { belongsToSelf, document, weight, description, hasWeightTickets } = proGear || {};
+
+  const proGearEntitlements = useSelector((state) => selectProGearEntitlements(state));
+
+  const validationSchema = Yup.object().shape({
+    belongsToSelf: Yup.bool().required('Required'),
+    document: Yup.object().nullable(),
+    weight: Yup.number()
+      .required('Required')
+      .min(0, 'Enter a weight 0 lbs or greater.')
```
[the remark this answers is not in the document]
scottmries: This is updated.

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
```diff
@@ -18,80 +21,110 @@ import ppmStyles from 'components/Customer/PPM/PPM.module.scss';
...
+  if (belongsToSelf === true) {
     proGearValue = 'true';
   }
-  if (selfProGear === false) {
+  if (belongsToSelf === false) {
     proGearValue = 'false';
   }
 
   const initialValues = {
-    selfProGear: proGearValue,
```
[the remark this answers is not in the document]
scottmries: `belongsToSelf` is nullable and gets initialized as `null`. When a proGear first gets loaded, the input should be unselected, so we need to preserve the distinction between `false` (meaning the spouse owns it) and `null` (unselected).

I technically could use a nested ternary, but the linter prevents it.

src/components/Customer/PPM/Closeout/FinalCloseoutForm/FinalCloseoutForm.test.jsx
```diff
@@ -26,10 +26,7 @@ describe('FinalCloseoutForm component', () => {
 
     const weightTicket = createCompleteWeightTicket({ serviceMemberId }, { emptyWeight: 14000, fullWeight: 18000 });
     const movingExpense = createCompleteMovingExpense({ serviceMemberId }, { amount: 30000 });
-    const proGearWeightTicket = createCompleteProGearWeightTicket(
-      { serviceMemberId },
-      { emptyWeight: 14000, fullWeight: 16000 },
-    );
+    const proGearWeightTicket = createCompleteProGearWeightTicket({ serviceMemberId }, { weight: 1500 });
```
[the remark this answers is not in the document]
scottmries: This is updated.

src/pages/MyMove/PPM/Closeout/Review/Review.jsx
```diff
@@ -64,7 +64,7 @@ const Review = () => {
   const mtoShipment = useSelector((state) => selectMTOShipmentById(state, mtoShipmentId));
 
   const weightTickets = mtoShipment?.ppmShipment?.weightTickets;
-  const proGear = mtoShipment?.ppmShipment?.proGear;
+  const proGear = mtoShipment?.ppmShipment?.proGearWeightTickets;
```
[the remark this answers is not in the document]
scottmries: This is updated.

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
scottmries: Good eye, thanks for pointing this out. I updated it.

bellex3 (commented):
Checked this out locally and verified that a customer is able to add & edit pro-gear weight tickets, upload & delete documents. Validation on weight allowance number works too.

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
```diff
@@ -17,81 +20,103 @@ import formStyles from 'styles/form.module.scss';
...
+      {({ isValid, isSubmitting, handleSubmit, values, ...formikProps }) => {
+        const getEntitlement = () => {
+          return values.belongsToSelf === 'true' ? proGearEntitlements.proGear : proGearEntitlements.proGearSpouse;
+        };
         return (
           <div className={classnames(ppmStyles.formContainer, styles.ProGearForm)}>
             <Form className={classnames(ppmStyles.form, styles.form)}>
               <SectionWrapper className={formStyles.formSection}>
                 <h2>Set {setNumber}</h2>
                 <FormGroup>
```
duncan-truss, line 84: A radio input is required for belongsToSelf but we don't display an error message here if the user presses submit without a selection because it's a group of inputs.  We can add in logic to set the FormGroup error since it's not a per field validation like the other inputs.

https://github.com/transcom/mymove/blob/94cedd2e4f6b24bc2bc463656b510ce6263334ad/src/components/Customer/PPM/Closeout/WeightTicketForm/WeightTicketForm.jsx#L237-L242

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
```diff
@@ -17,81 +20,103 @@ import formStyles from 'styles/form.module.scss';
...
   const initialValues = {
-    selfProGear: proGearValue,
-    proGearDocument: document?.uploads || [],
-    proGearWeight: proGearWeight ? `${proGearWeight}` : '',
+    belongsToSelf: proGearValue,
+    document: document?.uploads || [],
+    weight: weight ? `${weight}` : '',
     description: description ? `${description}` : '',
-    missingWeightTicket: missingWeightTicket ? `${missingWeightTicket}` : '',
+    missingWeightTicket: hasWeightTickets === 'false',
```
duncan-truss, line 64: I think `hasWeightTickets` coming back from the API is a boolean field and not a string, so this is being set to false after selecting I don't have weight tickets and saving the form.
Screen Shot 2022-11-28 at 12 15 17 PM
g">

rileyorr:
> @rileyorr I added a validation to weight so that 0 is invalid and the error is shown. Just to be clear: should the button be disabled when the form is valid, or should the form not trigger validation until the user attempts to save?

@scottmries thank you for adding that validation! The way you currently have it styled where the validation triggers when the focus is no longer on the input field matches the pattern used in the 'Expenses' and 'Weight moved' sections of the closeout flow. So, this looks good to me! 

> A radio input is required for belongsToSelf but we don't display an error message here if the user presses submit without a selection because it's a group of inputs. We can add in logic to set the FormGroup error since it's not a per field validation like the other inputs.

Great catch! I documented the pro-gear error messages in [this Confluence page](https://dp3.atlassian.net/wiki/spaces/MT/pages/1932656799/Error+handling+-+PPM+closeout+flow#Pro-gear), which includes the error message for this use case. Apologies for not linking it as a reference sooner; Confluence was down when I reviewed this PR last week.

scottmries:
@rileyorr and @duncan-truss 

Adding the validation to that form group is straightforward, but it has a strange side effect, which can also be seen when adding a new Expense. Here's how to reproduce this:

- Log in as `progearWeightTicket@ppm.approved` on a demo environment.
- Click "Upload PPM Documents" and "Add Expenses."
- Attempt to submit the new Expense form without selecting a type.
- Note the error: "Required"
- Select a type.
- With the entire form now open, Description and Amount are both shown as invalid, despite never having been touched.

Analogously, the required fields are shown as invalid in the same scenario on new Pro Gear tickets, when I add this error condition. This happens because the entire form, including the hidden fields, attempts to validate when it's submitted. 

We can imperatively set `touched` states on the other fields, etc., to avoid this, but that feels like a code smell to me. The form is never in a submittable state prior to selecting a Pro Gear owner (or an Expense type), so I wonder why in this case the submit button is enabled at all? It's not otherwise enabled when the form is invalid. Would disabling it in this scenario be an acceptable solution?

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
```diff
@@ -1,11 +1,14 @@
 import React, { createRef } from 'react';
+import { useSelector } from 'react-redux';
 import * as Yup from 'yup';
 import { Field, Formik } from 'formik';
 import { func, number } from 'prop-types';
 import { Button, Form, Link, Radio } from '@trussworks/react-uswds';
 import { FormGroup } from '@material-ui/core';
```
duncan-truss, line 7: I noticed this is not the FormGroup component we want either, we want the one from @trussworks/react-uswds.

duncan-truss:
> @rileyorr and @duncan-truss
> 
> Adding the validation to that form group is straightforward, but it has a strange side effect, which can also be seen when adding a new Expense. Here's how to reproduce this:
> 
>     * Log in as `progearWeightTicket@ppm.approved` (or the like) on a demo environment.
> 
>     * Click "Upload PPM Documents" and "Add Expenses."
> 
>     * Attempt to submit the new Expense form without selecting a type.
> 
>     * Note the error: "Required"
> 
>     * Select a type.
> 
>     * With the entire form now open, Description and Amount are both shown as invalid, despite never having been touched.
> 
> 
> Analogously, the required fields are shown as invalid in the same scenario on new Pro Gear tickets, when I add this error condition. This happens because the entire form, including the hidden fields, attempts to validate when it's submitted.
> 
> We can imperatively set `touched` states on the other fields, etc., to avoid this, but that feels like a code smell to me. The form is never in a submittable state prior to selecting a Pro Gear owner (or an Expense type), so I wonder why in this case the submit button is enabled at all? It's not otherwise enabled when the form is invalid. Would disabling it in this scenario be an acceptable solution?



> @rileyorr and @duncan-truss
> 
> Adding the validation to that form group is straightforward, but it has a strange side effect, which can also be seen when adding a new Expense. Here's how to reproduce this:
> 
>     * Log in as `progearWeightTicket@ppm.approved` (or the like) on a demo environment.
> 
>     * Click "Upload PPM Documents" and "Add Expenses."
> 
>     * Attempt to submit the new Expense form without selecting a type.
> 
>     * Note the error: "Required"
> 
>     * Select a type.
> 
>     * With the entire form now open, Description and Amount are both shown as invalid, despite never having been touched.
> 
> 
> Analogously, the required fields are shown as invalid in the same scenario on new Pro Gear tickets, when I add this error condition. This happens because the entire form, including the hidden fields, attempts to validate when it's submitted.
> 
> We can imperatively set `touched` states on the other fields, etc., to avoid this, but that feels like a code smell to me. The form is never in a submittable state prior to selecting a Pro Gear owner (or an Expense type), so I wonder why in this case the submit button is enabled at all? It's not otherwise enabled when the form is invalid. Would disabling it in this scenario be an acceptable solution?

I do see what you're getting at and agree it's a little awkward because we don't reveal the entire form until those first fields are populated.  I think it's still consistent with how we do other forms that don't have conditionally revealed fields, where the user just hasn't scrolled to see the entire form yet. At least for expenses there can be a different form shown for the storage type, the pro-gear forms for self and spouse are the same.  Maybe we'll get some usability around conditionally displayed forms eventually because they aren't used much elsewhere in the app.

Like you said trying to submit the form touches every field and then attempts to validate them as well. I agree with you that we'd be adding a special work around for a use case that I would guess is going to be fairly rare in practice? This behavior was happening even before adding the radio group error.

src/components/Customer/PPM/Closeout/ProGearForm/ProGearForm.jsx
scottmries: Good catch, this is resolved.

duncan-truss:
I found one existing issue with the backend service when deleting pro gear uploads and then returning to the edit screen they were still being shown.

I think we need to assign the filtered uploads back to the model since it returns a copy and doesn't modify the underlying field directly

https://github.com/transcom/mymove/blob/3a24d836677d3315a6fd26ce4a07f261e0266bb6/pkg/services/progear_weight_ticket/progear_weight_ticket_updater.go#L110

This should become
```go
proGearWeightTicket.Document.UserUploads = proGearWeightTicket.Document.UserUploads.FilterDeleted()
```

duncan-truss (approved):
This one had some complexity to it, great job 🎉 The deleted upload change was the last comment I had.

rileyorr (approved):
@scottmries thanks for pointing out that the required error message renders on subsequent fields if the customer attempts to proceed without selecting who the pro-gear belongs to. I agree that it's an awkward interaction. I also think it's a rare enough use case to be alright for MVP, especially since it's happening in multiple areas of the app. I took another look at the UI and Happo diffs. Great work on this ticket!

PR #9635, opened 2022-11-21
[MB-14447] Standardized use of constants and string literals for templates and test

afoon:
## Summary

For move template test, we want to use string literals for:

- action
- eventName
- tableName

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the Office Ap
2. Login as a TOO
3. Make sure move templates do not have any regressions.

### DIFF
--- a/src/constants/MoveHistory/EventTemplates/UpdateBillableWeight/updateBillableWeight.test.jsx
+++ b/src/constants/MoveHistory/EventTemplates/UpdateBillableWeight/updateBillableWeight.test.jsx
@@ -1,18 +1,15 @@
 import { render, screen } from '@testing-library/react';
 
-import o from 'constants/MoveHistory/UIDisplay/Operations';
 import getTemplate from 'constants/MoveHistory/TemplateManager';
 import e from 'constants/MoveHistory/EventTemplates/UpdateBillableWeight/updateBillableWeight';
-import a from 'constants/MoveHistory/Database/Actions';
-import t from 'constants/MoveHistory/Database/Tables';
 
 describe('when given an update billable weights', () => {
   const historyRecord = {
-    action: a.UPDATE,
+    action: 'UPDATE',
     // To-do: Change to max billable weight once available in database.
     changedValues: { authorized_weight: '5800' },
-    eventName: o.updateBillableWeight,
-    tableName: t.entitlements,
+    eventName: 'updateBillableWeight',
+    tableName: 'entitlements',
   };
   it('correctly matches the update billable weights event', () => {
     const result = getTemplate(historyRecord);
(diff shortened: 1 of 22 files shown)


src/constants/MoveHistory/EventTemplates/UpdateBillableWeight/updateBillableWeightRemarksAsTIO.test.jsx
```diff
@@ -1,18 +1,15 @@
...
-import t from 'constants/MoveHistory/Database/Tables';
 import getTemplate from 'constants/MoveHistory/TemplateManager';
 import e from 'constants/MoveHistory/EventTemplates/UpdateBillableWeight/updateBillableWeightRemarksAsTIO';
-import a from 'constants/MoveHistory/Database/Actions';
 
 describe('when given an update billable weight remarks as a TIO', () => {
   const historyRecord = {
-    action: a.UPDATE,
+    action: 'UPDATE',
     // To-do: Change to max billable weight once available in database.
```
ruizajtruss, line 9: Question: Is this comment in reference to the eventName or the changedValues field? It feels like the similar comment made in the JSX file implies it's talking about the changedValues field and no the event name

ruizajtruss (changes requested):
Left one comment on something that seemed worth double checking, other than that everything else looked good! Thanks for doing this, we needed it 🎉

ruizajtruss (approved):
LGTM! 🚀

src/constants/MoveHistory/EventTemplates/UpdateBillableWeight/updateBillableWeight.test.jsx
```diff
@@ -1,18 +1,15 @@
...
 import e from 'constants/MoveHistory/EventTemplates/UpdateBillableWeight/updateBillableWeight';
-import a from 'constants/MoveHistory/Database/Actions';
-import t from 'constants/MoveHistory/Database/Tables';
 
 describe('when given an update billable weights', () => {
   const historyRecord = {
-    action: a.UPDATE,
-    // To-do: Change to max billable weight once available in database.
+    action: 'UPDATE',
+     // To-do: Update changedValue key to max_billable_weight once available in database.
```
ruizajtruss, line 9: Just a heads up, `pre_test` is failing and I think it may be due to prettier not liking this space in the beginning

kctruss (approved):
🎉

YanZ777 (commented):
I'm curious about the motivation behind using string literals for tests. I don't inherently mind, but I do want to know what the motivation is.

afoon:
> 



> I'm curious about the motivation behind using string literals for tests. I don't inherently mind, but I do want to know what the motivation is.

We didn't want to use production constants in our test. It also helps catch mistakes that may involve constants. I caught a casing error in our operation constant when updating event names to string literals.

YanZ777:
> > 
> 
> > I'm curious about the motivation behind using string literals for tests. I don't inherently mind, but I do want to know what the motivation is.
> 
> We didn't want to use production constants in our test. It also helps catch mistakes that may involve constants. I caught a casing error in our operation constant when updating event names to string literals.

That makes sense! Thanks for the explanation. Could you add that blurb to the PR description? 
Also, would you recommend that all FE tests should not be using production constants, or just these ones in particular? (The motivation behind these questions is to see if this would be a good ADR to write up on and amend the documentation.)

PR #9636, opened 2022-11-21
[MB-14535] Remove clear statements and scope to form in about page component test

duncan-truss:
## Summary

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the
2. Login as a
3.

### Frontend

### Backend

### DIFF
--- a/src/components/Customer/PPM/Closeout/AboutForm/AboutForm.jsx
+++ b/src/components/Customer/PPM/Closeout/AboutForm/AboutForm.jsx
@@ -95,7 +95,7 @@ const AboutForm = ({ mtoShipment, onBack, onSubmit, postalCodeValidator }) => {
       {({ isValid, isSubmitting, handleSubmit, values }) => {
         return (
           <div className={classnames(ppmStyles.formContainer, styles.AboutForm)}>
-            <Form className={classnames(formStyles.form, ppmStyles.form, styles.W2Address)}>
+            <Form className={classnames(formStyles.form, ppmStyles.form, styles.W2Address)} data-testid="aboutForm">
               <SectionWrapper className={classnames(ppmStyles.sectionWrapper, formStyles.formSection)}>
                 <h2>Departure date</h2>
                 <DatePickerInput
--- a/src/pages/MyMove/PPM/Closeout/About/About.test.jsx
+++ b/src/pages/MyMove/PPM/Closeout/About/About.test.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, waitFor, screen } from '@testing-library/react';
+import { render, waitFor, screen, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { generatePath } from 'react-router';
 import { v4 } from 'uuid';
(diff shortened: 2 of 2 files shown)


scottmries:
@duncan-truss I looked into this some. I wonder what you think of this solution.
Using `userEvent.paste()` in place of the `.type()` calls in your form filler functions drops the execution time of both tests that use them significantly, allowing the entire suite to run consistently in just under the timeout of 5s (at least locally).
That's one pro, and another is that a user should expect to be able to paste into these fields, so testing the use of `paste()` is not inappropriate.
There are two cons though as far as I can see: 
First, `paste()` is not consistent behavior, since we use `type()` everywhere else, and there might be a reasonable expectation that typing is the more common interaction.
Second, the primary bottleneck in both of the slower tests is the `waitFor` call wrapping the two mocks of `patchMTOShipment`. I haven't gotten to the bottom of why a mock function would not be instantaneous, given which we would expect `waitFor` to run no longer than its 50ms retry time. As it is, those `waitFor`s take between 500 and 1000ms and retry around 4 times.
I'm happy to refactor to move the user interaction tests to Cypress if you don't think this `paste()` solution is viable.

duncan-truss:
> @duncan-truss I looked into this some. I wonder what you think of this solution.
> 
> Using `userEvent.paste()` in place of the `.type()` calls in your form filler functions drops the execution time of both tests that use them significantly, allowing the entire suite to run consistently in just under the timeout of 5s (at least locally). That's one pro, and another is that a user should expect to be able to paste into these fields, so testing the use of `paste()` is not inappropriate.
> 
> There are two cons though as far as I can see:
> 
> First, `paste()` is not consistent behavior, since we use `type()` everywhere else, and there might be a reasonable expectation that typing is the more common interaction.
> 
> Second, the primary bottleneck in both of the slower tests is the `waitFor` call wrapping the two mocks of `patchMTOShipment`. I haven't gotten to the bottom of why a mock function would not be instantaneous, given which we would expect `waitFor` to run no longer than its 50ms retry time. As it is, those `waitFor`s take between 500 and 1000ms and retry around 4 times.
> 
> I'm happy to refactor to move the user interaction tests to Cypress if you don't think this `paste()` solution is viable.

Oh neat, I thought I went looking for something similar but wasn't sure if I saw that in the docs or not. I know that `type()` takes an options object where you can specify the `delay` property but is set to zero by default. One reason paste might be more performant is that the onchange event isn't triggered as frequently and therefore fewer validation callbacks are run.  I know particularly for the postal codes we have async validators not just the validation schema.  Similarly blur events will trigger extra field validations and on form submission Formik calls validate again on all inputs.

I think that the 5s timeout is a per test limit and not a suite limit.  It was just the "calls the patch shipment  with appropriate payload" test that was timing out. On my local machine that test tended to hover around 2s for me, yours may be faster or we're seeing improvements.

You could also try shortening the interval of `waitFor` to something below 50ms https://testing-library.com/docs/dom-testing-library/api-async/#waitfor?

I think what you've added is worth merging and seeing if the success rate improves.  It's last failure was 10 days ago on the insights page so with the caveat that people haven't been merging over the holidays it's not failing egregiously.

duncan-truss:
I was going to approve this but I can't as the author, so feel free to give yourself an approval 😆

scottmries (approved):

PR #9640, opened 2022-11-22
MB-14486 Update react-admin usage of react-router to use latest version

PDickeyTruss:
## Summary

This PR updates the parts of react-router that interact with react-admin to use the latest (v6) version of react-router. This is one of many incremental changes required to upgrade react-admin to the latest version. The app will not compile until the rest of these changes are complete, which of course makes this basically un-testable. I would suggest just doing a code review as the review here and we will have to fully test out functionally when all updates are done and the app will build again. 

Note: This PR is merging into our feature branch that includes the react-admin upgrade and both versions of react-router. It will not work in main without the changes included in this feature branch. 

## Changes in PR
The changes in this PR align with the bullets in the ticket, no additional changes were needed. From the ticket:

- refactor <CustomRoutes> component (`src/scenes/SystemAdmin/CustomRoutes.jsx`)
- use `useNavigate` instead of `useHistory` (src/pages/InvalidPermissions and src/pages/SignIn
- use BroswerRouter component in place of createBrowserHistory (src/scenes/SystemAdmin/Home.jsx)
- update code to set base path as /admin

### DIFF
--- a/src/pages/InvalidPermissions/InvalidPermissions.jsx
+++ b/src/pages/InvalidPermissions/InvalidPermissions.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import qs from 'query-string';
 import { bool, shape, string } from 'prop-types';
 import { Button, ButtonGroup } from '@trussworks/react-uswds';
-import { useHistory } from 'react-router-dom-old';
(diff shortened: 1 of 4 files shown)


mandaem (approved):
LGTM

PR #9641, opened 2022-11-22
MB-14481-Refactor data actions.

dymonmoore:
## Summary

This pull request refactors react-admin data actions from the admin app rest provider. This is an incremental change required to upgrade the admin app. 

The app will not compile until the rest of the changes are complete. This pull request cannot be tested at this time. We recommend a code review for now.

## Changes in Pull Request

- Refactor React Admin Data Actions
-- File: src/scenes/SystemAdmin/shared/rest_provider.js

### DIFF
--- a/src/scenes/SystemAdmin/shared/rest_provider.js
+++ b/src/scenes/SystemAdmin/shared/rest_provider.js
@@ -1,18 +1,7 @@
 import { stringify } from 'query-string';
 import { diff } from 'deep-object-diff';
 import { snakeCase } from 'lodash';
-import {
-  CREATE,
-  DELETE,
-  DELETE_MANY,
-  fetchUtils,
-  GET_LIST,
-  GET_MANY,
-  GET_MANY_REFERENCE,
-  GET_ONE,
-  UPDATE,
-  UPDATE_MANY,
-} from 'react-admin';
+import { fetchUtils } from 'react-admin';
 
 /**
  * Maps react-admin queries to a simple REST API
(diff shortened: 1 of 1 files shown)


src/scenes/SystemAdmin/shared/rest_provider.js
```diff
@@ -1,18 +1,7 @@
...
-  DELETE,
-  DELETE_MANY,
-  fetchUtils,
-  GET_LIST,
-  GET_MANY,
-  GET_MANY_REFERENCE,
-  GET_ONE,
-  UPDATE,
-  UPDATE_MANY,
-} from 'react-admin';
```
PDickeyTruss, line 15: I am a little unsure if we need to remove usage of these. All of these imports seem to still exist the latest version of react-admin. It looks like these are just constants for strings of the same value. We are basically just switching on these types to figure out what type of request is going on and set it up. These existing in the new react admin makes me think we can just leave them as is? 

@leia-sefkin Would love to get your 👀 and thoughts on this as I think you may have a better understanding of what part of the upgrade this ticket is addressing.

PDickeyTruss: I think the 'Data Actions' that were removed in the new react-admin were different that these (though I am pretty clueless about react-admin). The [Data Actions in the upgrade guide](https://marmelab.com/react-admin/doc/4.0/Upgrade.html#no-more-data-actions) that need removed seem to be more like `CRUD_GET_ONE` rather than `GET_ONE`. This, in combination with these values still being present in the new version, makes me think that these are different and that we do not use the type of "Data Actions" that were removed in the upgraded react-admin.

leia-sefkin: @PDickeyTruss @dymonmoore apologies just getting around to reviewing this one. 

I noticed the difference in the upgrade guide as well, but I assumed it might cover the other data actions (not just those that start with `CRUD`). If it's not necessary to make the change that's fine by me to mark this ticket as not-doing. apologies to @dymonmoore for the back and forth on direction 🙇

PDickeyTruss: I am good with closing as not-doing. I think we probably had to do this investigation of whether those CRUD prefixed 'data actions' were the same as the non-prefixed we are using no matter what since the guide wasn't quite clear enough... so this wasn't all for nothing 😜

leia-sefkin: you're a 💎 Patrick - gracias

dymonmoore: @PDickeyTruss  @leia-sefkin  Thanks for the clarification! I'm totally fine with closing this ticket.

PR #9658, opened 2022-11-28
MB-14433-Removed prop injection from page components and deleted the base path prop.

dymonmoore:
## Summary

This pull request is an incremental change apart of the React Admin upgrade. Prop injection was removed from all React Admin Page components (i.e List, Create, Edit) and the basePath prop was removed from the OfficeUserList component. 

The app will not compile until the rest of the upgrade changes are complete. This pull request cannot be tested at this time. For now, all is needed is a **code review.**

Current State of Admin App without Prop Injection due to upgrade being incomplete.
Screen Shot 2022-11-29 at 10 34 25 AM

### DIFF
--- a/src/pages/Admin/AdminUsers/AdminUserCreate.jsx
+++ b/src/pages/Admin/AdminUsers/AdminUserCreate.jsx
@@ -1,8 +1,8 @@
 import React from 'react';
 import { Create, SimpleForm, TextInput, ReferenceInput, AutocompleteInput, required } from 'react-admin';
 
-const AdminUserCreate = (props) => (
-  <Create {...props}>
+const AdminUserCreate = () => (
+  <Create>
     <SimpleForm>
       <TextInput source="email" validate={required()} />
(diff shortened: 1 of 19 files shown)


dymonmoore:
Hi team! I removed prop injection from the React Admin page components and deleted the use of baseProp according to the upgrade guide. There is some additional prop injection in non-page components in the admin app. Should it be removed? Here's examples:  

const OfficeUserEditToolbar = (props) => (
  <Toolbar {...props}>
    <SaveButton />
  </Toolbar>
);

const OfficeUserListFilter = (props) => (
  <Filter {...props}>
    <TextInput source="search" alwaysOn />
  </Filter>
);

PDickeyTruss (approved):
Looks great, approved! 

Re your question about if you should remove the additional prop injection in non-page components in the admin app. I think you were correct to leave those as is, I see them mention 'page components' in that area of the guide. 👍

PR #9660, opened 2022-11-29
MB-14365 Add office endpoint to get weight ticket info

felipe-lee:
## Summary

This pull request adds an office endpoint to retrieve all the weight tickets associated with a PPM shipment.

<details>
<summary>Changes</summary>

- [Add FatalTrue test suite helper func]()
- [Add weight ticket fetcher service object]()
    - This mostly already existed as a function that lived with the Weight Ticket updater service object. Moved this to a fetcher service object and added an auth check.
    - [Removed the old fetching function and swapped usages to use the new fetcher service object]()
- [Add function to weight ticket fetcher to grab all weight tickets associated with a PPM shipment]()
- [Add ghc endpoint definition for retrieving all weight tickets for a PPM shipment]()
- [Add ghc handler that retrieves all weight tickets for a PPM shipment]()

</details>

## Checking Changes

<details>
<summary>💻</summary>

### Verifying By Running Tests

If you'd like to validate purely by running the tests, then you can run

```sh
make server_test
```

### Verifying With Postman

If you'd like to verify using postman, you can follow the [postman setup instructions](https://transcom.github.io/mymove-docs/docs/tools/postman/setting-up-postman) for more detailed instructions, and info on getting your cookies set up.

#### Populate the Dev DB

You only need to do this if you don't have any PPM shipments with weight tickets locally.

```sh
make db_dev_fresh
```

#### Run the Server

```sh
make server_run
```

#### Run Office Client

```sh
make office_client_run
```

#### Making a Request

Endpoint to test: 

`/ppm-shipment/{ppmShipmentID}/weight-tickets`

E.g.

`http://officelocal:3000/ghc/v1/ppm-shipments/7a5e932d-f1f6-435e-9518-3ee33f74bc88/weight-tickets`

You should get back the list of weight tickets (though for the example I linked, there is only one).

</details>

### Backend


pkg/handlers/ghcapi/weight_ticket.go
```diff
@@ -1,16 +1,70 @@
...
+	services.WeightTicketFetcher
+}
+
+// Handle retrieves all weight tickets for a PPM shipment
+func (h GetWeightTicketsHandler) Handle(params weightticketops.GetWeightTicketsParams) middleware.Responder {
+	return h.AuditableAppContextFromRequestWithErrors(params.HTTPRequest,
+		func(appCtx appcontext.AppContext) (middleware.Responder, error) {
+			errInstance := fmt.Sprintf("Instance: %s", h.GetTraceIDFromRequest(params.HTTPRequest))
+
+			errPayload := &ghcmessages.Error{Message: &errInstance}
```
felipe-lee, line 31: Tbh, not sure that this is really needed. It seems like our middleware adds this as a response header. I mainly added it because there were other ghc handlers that similarly add this to some of their error responses. What do y'all think?

pkg/handlers/ghcapi/weight_ticket.go
```diff
@@ -1,16 +1,70 @@
...
+
+			if !appCtx.Session().IsOfficeApp() {
+				return weightticketops.NewGetWeightTicketsForbidden().WithPayload(errPayload), apperror.NewSessionError("Request should come from the office app.")
+			}
+
+			ppmShipmentID := uuid.FromStringOrNil(params.PpmShipmentID.String())
```
felipe-lee, line 37: So, `go-swagger` takes care of making sure this is a valid UUID format and returns a 422 response if it isn't (hence why that's a possible response in the yaml even though I don't explicitly return that in this handler). It doesn't check that it isn't a nil UUID (all zeros), but I kind of think that's fine since that'll just lead to no weight tickets being found since no ppm shipment will have that ID. That does mean though that they would get a 200 response with an empty array though, so I could see an argument being made for checking here explicitly to give back an error instead. What do y'all think?

pkg/services/weight_ticket/weight_ticket_fetcher.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if err != nil {
+		return nil, apperror.NewQueryError("WeightTicket", err, "unable to find WeightTickets")
+	}
+
+	if weightTickets == nil {
+		return weightTickets, nil
+	}
+
+	if appCtx.Session().IsMilApp() && weightTickets[0].EmptyDocument.ServiceMemberID != appCtx.Session().ServiceMemberID {
```
felipe-lee, line 78: I'm kind of making the assumption here that all weight tickets documents would have the same service member associated with them. Do y'all think that's fine, or should I be checking differently (checking them all, checking the orders associated with the ppm shipment, or some other way)?

pkg/handlers/ghcapi/weight_ticket.go
NamibiaTorres: Is there a way to test our middleware first to see if it is needed? I also noticed this in the GHC handlers, but a of them are not following the newer convention for handlers that was implemented months ago.

pkg/handlers/ghcapi/weight_ticket.go
NamibiaTorres: Good question. I think I took this check out of mine  too, but now I'm debating whether to add it back in. I'm ok with leaving it out since it is being handled somewhere.

pkg/services/weight_ticket/weight_ticket_fetcher.go
NamibiaTorres: I made the same assumption in mine. For the tests, I think this is fine. Checking that the orders are associated with the ppm shipment is a nice to have if you have the time to add this to the test.

NamibiaTorres (approved):
Looking through this PR, I feel like you addressed all of the requirements. If there are things you are considering changing, I think it's okay to include this in another PR for next week. Your code looks good!

pkg/services/weight_ticket/weight_ticket_fetcher.go
NamibiaTorres: I think that is reasonable. If we want to go through the more complex route, let's just make that it's own PR.

felipe-lee:
> Looking through this PR, I feel like you addressed all of the requirements. If there are things you are considering changing, I think it's okay to include this in another PR for next week. Your code looks good!

I think based on how the demo went yesterday, I'll keep things as they are. We can come back later like you said if there are other changes we think we need.

PR #9671, opened 2022-11-30
MB-14565: Add final incentive

ronaktruss:
## Summary

This PR adds finalIncentive calculations to the BE and ensures that the final submission/review page displays the latest calculated finalIncentive value.

There are two scenarios when the final incentive changes.
1. When information on the about page changes (specifically the move date, pickup postal code, and destination postal code)
2. When information on the weight tickets change (specifically the calculated weight changes -  ie. changes to either the full or empty weight)

In the first scenario we can call an update the the mtoShipments api and update the ppmShipment with the new values and calculate the final incentive by summing the total weight moved across the existing weight tickets.

In the second scenario we have to update both the weightTicket and the ppmShipment's final incentive. To do this we have added a dependency to the weightTicketUpdater to rely on the ppmShipmentUpdater. So when a weightTicket is modified such that the total weights across all weight tickets would change we also call the ppmShipmentUpdater.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the customer app
1. Login as `ppm.test.user1@example.com` to create some sample ppmShipments with data to generate estimated Incentives
1. Login as `weightTicketPPM@ppm.approved` and recreate those scenarios in by modifying the about page and the weightTickets. Ensure that the calculated final incentive matches those in the above step.

### Frontend

### Backend


pkg/services/weight_ticket/weight_ticket_updater.go
```diff
@@ -118,3 +119,31 @@ func FetchWeightTicketByIDExcludeDeletedUploads(appContext appcontext.AppContext
...
+
+func FetchWeightTicketAndPPMShipment(appContext appcontext.AppContext, weightTicketID uuid.UUID) (*models.WeightTicket, error) {
+	var weightTicket models.WeightTicket
+
+	err := appContext.DB().Scope(utilities.ExcludeDeletedScope()).
+		EagerPreload(
+			"EmptyDocument.UserUploads.Upload",
+			"FullDocument.UserUploads.Upload",
+			"ProofOfTrailerOwnershipDocument.UserUploads.Upload",
+			"PPMShipment.Shipment",
```
pearl-truss, line 131: Curious if this should just be `"PPMShipment"`

pkg/services/ppmshipment/ppm_shipment_updater.go
```diff
@@ -88,6 +88,17 @@ func (f *ppmShipmentUpdater) updatePPMShipment(appCtx appcontext.AppContext, ppm
 			}
 		}
 
+		// TODO: Replace with call to calculate net weight
```
pearl-truss, line 91: Essentially we don't want to call the `FinalIncentive` function when the PPM is being updated without affecting the final incentive. I'm not sure if we want to add a condition for when `FinalIncentive` is called or not to be based on the weight vs calling the `FinalIncentive` function which will return `nil` and setting that as the `FinalIncentive` value?

src/utils/ppmCloseout.jsx
```diff
@@ -37,7 +37,7 @@ export const formatAboutYourPPMItem = (ppmShipment, editPath, editParams) => {
           id: 'advance',
           label: 'Advance:',
           value: ppmShipment.hasReceivedAdvance
-            ? `Yes, $${formatCentsTruncateWhole(ppmShipment.advanceAmountRequested)}`
+            ? `Yes, $${formatCentsTruncateWhole(ppmShipment.advanceAmountReceived)}`
```
ronaktruss, line 40: I believe we were using the wrong value here. Let me know if we should be actually showing the requested amount here over the received amount.

src/pages/MyMove/PPM/Closeout/FinalCloseout/FinalCloseout.jsx
```diff
@@ -21,11 +21,25 @@ const FinalCloseout = () => {
...
+      })
+      .catch(() => {
+        setErrorMessage('Failed to fetch shipment information');
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+  }, [moveId, mtoShipmentId, dispatch]);
+
+  if (!mtoShipment || isLoading) {
```
ronaktruss, line 42: We need to fetch the mtoShipments info since when a weightTicket change causes the finalIncentive amount to be updated redux does not have the correct value. Since we are updating the weightTicket we don't return the finalIncentive for the UI to use.

ronaktruss:
I'm still planning on adding a few more tests, but this should be in a good spot for others to start reviewing.

pkg/services/weight_ticket/weight_ticket_updater.go
```diff
@@ -46,12 +51,31 @@ func (f *weightTicketUpdater) UpdateWeightTicket(appCtx appcontext.AppContext, w
...
 
+	hasTotalWeightChanged := hasTotalWeightChanged(*originalWeightTicket, mergedWeightTicket)
+	var currentPPMShipment models.PPMShipment
+	if hasTotalWeightChanged {
+		ppmShipmentFromDB, issue := ppmshipment.FindPPMShipmentAndWeightTickets(appCtx, originalWeightTicket.PPMShipmentID)
+		if issue != nil {
+			return nil, issue
+		}
+		currentPPMShipment = *ppmShipmentFromDB
+		currentPPMShipment.WeightTickets = models.WeightTickets{mergedWeightTicket}
```
felipe-lee, line 66: **question**: Doesn't this end up replacing all the weight tickets if there is more than 1? Not saving that to the DB, but I'm wondering more about side effects of things that run on this expecting all weight tickets and only getting the currently modified one.

felipe-lee (commented):
I haven't finished reviewing, but left a few questions/comments. I have tested it locally and it seems to be working for the comparisons I made (a few versions of having multiple weight tickets add up to the estimated weight).

pkg/services/ppmshipment/ppm_estimator.go
```diff
@@ -133,13 +154,68 @@ func (f *estimatePPM) estimateIncentive(appCtx appcontext.AppContext, oldPPMShip
...
+
+	if len(ppmShipment.WeightTickets) >= 1 {
+		for _, weightTicket := range ppmShipment.WeightTickets {
+			if weightTicket.ID == updatedWeightTicketID {
+				totalWeight += changedWeight
+			} else if weightTicket.FullWeight != nil && weightTicket.EmptyWeight != nil {
+				totalWeight += *weightTicket.FullWeight - *weightTicket.EmptyWeight
+			}
+		}
+	}
```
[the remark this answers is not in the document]
ronaktruss: Yeah I was actually working on something similar since I think that solution is likely more intuitive and likely more easily understandable because of that. I was hoping to get that implementation done so that folks could compare the two approaches, but just didn't get the time to finish that.

The one caveat is that when we fetch all of the weightTickets within the weightTicketUpdater is that we get the info from the DB without the new changes. This would require us to filter through those weightTickets and replace the correct index of weightTicket with the current weight ticket that is being modified. Additionally this approach would likely do some unnecessary calculations. For example, we are ensuring that the weight is being changed prior to calling the ppmShipmentUpdater but for readability/intuitiveness we would likely be calculating both the old and new weights. Since the number of weightTickets is fairly small, these are negligible in terms of performance so I would be in favor of the more readable and maintainable solution.

src/utils/ppmCloseout.jsx
pearl-truss: It's my understanding that we would want `advanceAmountReceived` here

PR #9682, opened 2022-12-01
[MB-14563] Validate/fix payloads for GHC API

reggieriser:
## Summary

**WIP**

This PR validates (and in some cases, fixes) the response payloads for the GHC API.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the
2. Login as a
3.


swagger-def/ghc.yaml
```diff
@@ -4480,6 +4480,14 @@ definitions:
         type: integer
       queuePaymentRequests:
         $ref: '#/definitions/QueuePaymentRequests'
+  QueuePaymentRequestStatus:
```
reggieriser, line 4483: The problem with validation on the payment request queue handler was that we return a different set of display statuses in its response rather than the statuses in the database.  I thought about moving the display statuses to the front end, but some of these database statuses get rolled up in a way that would make for a fairly big refactor.

pkg/services/move/move_router.go
```diff
@@ -478,7 +478,7 @@ func (router moveRouter) CompleteServiceCounseling(appCtx appcontext.AppContext,
 // ApproveOrRequestApproval routes the move appropriately based on whether or
 // not the TOO has any tasks requiring their attention.
 func (router moveRouter) ApproveOrRequestApproval(appCtx appcontext.AppContext, move models.Move) (*models.Move, error) {
-	err := appCtx.DB().Q().EagerPreload("MTOServiceItems", "Orders", "MTOShipments.SITExtensions").Find(&move, move.ID)
+	err := appCtx.DB().Q().EagerPreload("MTOServiceItems", "Orders.ServiceMember", "Orders.NewDutyLocation.Address", "MTOShipments.SITExtensions").Find(&move, move.ID)
```
reggieriser, line 481: You'll see several places like this where I had to load additional related data because the swagger definition was expecting it to be there.  Ideally, we'd only load data we absolutely need, but that would require more specialized responses since the ones we have are often shared across endpoints where that additional data may be important.  I didn't want to try to tackle that in this PR.

ahobson (approved):
👏 🥳 

This is such nice work! I reviewed the changes and they all made sense to me, but I may not be the most knowledgable about all areas.

PR #9683, opened 2022-12-02
MB-14606 fix prereqs and cac prereqs scripts for nix users

felipe-lee:
## Summary

This updates a few of the paths in the `prereqs` and `cac-prereqs` scripts to not be hard-coded anymore. It is primarily aimed at fixing things for `nix` users, but hopefully some things like the `PKCS11` and `PKCS15` env var setting will also benefit `brew` users if the install path ever changes.

I also fixed the `PKCS11MODULE` var being set in `.envrc` for `nix` users. Tbh, I'm not sure if it needed to be set, since it isn't being set for `brew` users, but since it was being set, I updated it and now we do use it in the `cac-prereqs` script.

## Checking Changes

There are three files that changed. All need a `nix` reviewer, but only the `cac-prereqs` one really needs a `brew` reviewer. Feel free to check them all if you are a `brew` reviewer if you'd like, but your results will be different.

<details>
<summary>💻</summary>

### `.envrc` Check

This one is for `nix` users only, since the line that changed only gets triggered for `nix` users.

First, get the vars set.

```sh
direnv allow
```

Check that the `PKCS11MODULE` env var is set correctly for you.

```sh
test -f $PKCS11MODULE && echo "found the module" || echo "didn't find it"
```

You should hopefully see that it found the module. If not, let me know!

Note that if you are a `brew` user, this variable isn't set. 

### `prereqs` check

This one is also mainly for `nix` users, though it wouldn't (shouldn't...) hurt if others ran it. 

Check that the `prereqs` script works correctly for you.

```sh
prereqs
```

* For `nix` users that means seeing something like this:
    
    ```sh
    Using docker
    Homebrew Bundle complete! 1 Brewfile dependency now installed.
    nix dependencies installed
    OK: all prereqs found

    Golang go version go1.19.3 darwin/amd64 installed
    Node v16.15.0 installed
    ```
 
* I'm not sure what `brew` users would see, but if someone reviewing thinks theirs looks correct, can you either share it or update this description?

### `cac-prereqs` check

This one is for both `nix` and `brew` users!

Check that the `cac-prereqs` script works correctly for you.

```sh
cac-prereqs
```

If all goes well, there should be no output. If not, let me know!

</details>


scripts/cac-prereqs
```diff
@@ -9,40 +9,38 @@ set -eu -o pipefail
...
 
-# Looks like placement of this module can vary by versions. Some even have both. The module's docs don't really have
-# info on this, so we'll just have to check both. It seems though, like if we move in the direction of using the cask
-# version of opensc, that one may only use the /usr/local/lib/opensc-pkcs11.so path.
+# Looks like non-nix placement of this module can vary by versions. Some even have both. The module's docs don't really
+# have info on this, so we'll just have to check both. It seems though, like if we move in the direction of using the
+# cask version of opensc, that one may only use the /usr/local/lib/opensc-pkcs11.so path.
 MODULE=
-if [[ -f /usr/local/lib/opensc-pkcs11.so ]]; then
+if [ ! -r .nix-disable ] && [[ -f "${PKCS11MODULE-}" ]]; then
```
duncan-truss, line 36: Added the default `-` here because without nix I was getting an unbound var error.

scripts/cac-prereqs
```diff
@@ -9,40 +9,38 @@ set -eu -o pipefail
...
+  MODULE="${PKCS11MODULE}"
+elif [[ -f /usr/local/lib/opensc-pkcs11.so ]]; then
   MODULE=/usr/local/lib/opensc-pkcs11.so
 elif [[ -f /usr/local/lib/pkcs11/opensc-pkcs11.so ]]; then
   MODULE=/usr/local/lib/pkcs11/opensc-pkcs11.so
-elif [ ! -r .nix-disable ] && [[ -f /nix/var/nix/profiles/mymove/lib/opensc-pkcs11.so ]]; then
-  MODULE=/nix/var/nix/profiles/mymove/lib/opensc-pkcs11.so
+elif [[ -f /Library/OpenSC/lib/pkcs11/opensc-pkcs11.so ]]; then
```
duncan-truss, line 42: Here is my new path, it looks like a postinstall script is supposed to copy it over to `/usr/local/lib` https://github.com/OpenSC/OpenSC/blob/dd48facd38a9618bbe6ec035453433f5a24c6316/MacOSX/scripts/postinstall#L4

duncan-truss: After forcing a reinstall `brew reinstall --cask opensc` it got copied over to `/usr/local/lib/` as expected.

scripts/cac-prereqs
felipe-lee: Thanks for catching that and fixing it!

scripts/cac-prereqs
felipe-lee: Interesting. Well, good to have the fallback in case it messes up for other people too.

rogeruiz (approved):
lgtm 🌈

> Reviewed as a Nix user.

PR #9689, opened 2022-12-02
[MB-13856] New columns for PPM closeout tab in Services Counselor queue

mboren:
## Summary

The PPM closeout tab has the following new columns:
- Closeout initiated
  - The latest time that any of the PPM shipments associated with a move was submitted.
- Full or partial PPM
  - This comes from the `ppm_status` field on `moves`
- Destination duty location
- PPM closeout location
  - For Army/Air Force moves, this will show the closeout location chosen by the customer (however, customers cannot set this field yet, so it will mostly be blank, that work will be done in [MB-11915](https://dp3.atlassian.net/browse/MB-11915)) 
  - For Navy, Coast Guard, and Marines moves, this will show the "closeout GBLOC" associated with their service: NAVY for Navy, USCG for Coast Guard, and TVCB for Marines.
    - Both sorting and filtering on this column have no effect for NAVY/USCG/TVCB office users because every move they see will have exactly the same value in here.

The filters on PPM closeout location and Destination duty location work the same as the existing filter on Origin duty location: they do a case insensitive search, matching from the beginning of the word. 
So a search for "fort bra" will match "Fort Bragg"
But a search for "bragg" will not.

We may change the way the closeout location filter works in the future: [MB-14594](https://dp3.atlassian.net/browse/MB-14594)

Passing the closeout gblocs through to the closeout location was kind of awkward. I decided just send back kind of a fake transportation office with the GBLOC set as the name. If there's a classier way to do this I'd love to hear about it.
An alternative approach would have been to look up the transportation offices used by the office users in the those GBLOCs. We would still have to go through those somewhere (probably model to payload?) to use the GBLOC instead of the transportation office name though. 

Also I used a bit of a hack to fix the issue with filters persisting when you switch tabs: I just wrapped one of the `TableQueue`s in a `div`. As far as I can tell, this hasn't broken the styling and seems to work fine.

The complexity in `order_fetcher.go` is getting out of hand. I think we should split Services Counselor and TOO queues up. I think at this point the logic is getting hard enough to follow that it's worth duplicating some code. This code originally started as a pretty generic thing that fetched orders, but it's exclusively used for queues and the business logic for the queues has gotten way more complicated than it used to be, so i think it's worth coming up with more specific solutions.

## Setup to Run Your Code
```
make db_dev_e2e_populate
make server_run
```
```
make client_run
```

### Additional steps
- Log in as a Services Counselor user
- Try sorting and filtering the new columns and make sure the results make sense
- Verify that switching between the Counseling and PPM Closeout tabs resets any filters or sorts that were applied
- Verify that values in the new columns match the designs
- For at least one of the closeout GBOCs (TVCB, NAVY, USCG):
  - Login as a services counselor in that gbloc
  - On the PPM Closeout tab, make sure that the new filters and sorts work
  - Note that filtering by the "Closeout location" column will do nothing if you are in a closeout GBLOC

https://user-images.githubusercontent.com/2627844/205415280-8f090b82-d63d-4989-be55-0e5c56ef9e96.mov

In this screen recording, it's a bit hard to test some of the columns because the test data I was using didn't have any moves with the *Full* ppm type, so everything is Partial. And the destination duty location for every move was Fort Gordon.

### Frontend

### Backend


mandaem (commented):
It appears that the PPM closeout location filter isn't working. It returns full results no matter what the search is for (instead of just matching results). here's one example screenshot:
Screenshot 2022-12-06 at 11 27 06 AM


That's the only thing I've noticed, the other filters are working as far I could test, and the right people see the right things: only PPM tab for USCG, NAVY, and TVCB, all for KKFA.

mboren:
> It returns full results no matter what the search is for (instead of just matching results)

This is intentional for office users in closeout GBLOCs, but I meant to do a little more here to make it less confusing and forgot in the weekend shuffle.

It was going to be both hard and sort of pointless to implement the closeout location search for office users in closeout GBOCs because every single move that is visible to them has exactly the same closeout location (which is the same as the office user's GBLOC).

So I just ignored that filter for office users in closeout GBLOCs. I think it would probably make sense to disable it on the frontend as well, to prevent confusion, but I should ask design about that.

mandaem (approved):
LGTM

dinabeigelman (approved):
lgtm!

pkg/services/order/order_fetcher.go
```diff
@@ -171,17 +181,30 @@ func (f orderFetcher) ListOrders(appCtx appcontext.AppContext, officeUserID uuid
...
 		return []models.Move{}, 0, err
 	}
 	// Get the count
 	count := query.Paginator.TotalEntriesSize
 
+	overwriteCloseoutOfficeWithGBLOC := ppmCloseoutGblocs && params.NeedsPPMCloseout != nil && *params.NeedsPPMCloseout
+	closeoutOffice := models.TransportationOffice{
+		Name: officeUserGbloc,
+	}
+
```
leia-sefkin, line 207: this feels like an ok approach to me for handling cases where we need to populate the office name with gbloc. i don't think it's your code that's awkward, just the business constraints.

suggestion: throw in a comment above relating why you need to overwrite for folks who may not have context on why this is necessary

pkg/services/order/order_fetcher.go
```diff
@@ -333,6 +368,36 @@ func requestedMoveDateFilter(requestedMoveDate *string) QueryOption {
 	}
 }
 
+func closeoutInitiatedFilter(closeoutInitiated *time.Time) QueryOption {
+	return func(query *pop.Query) {
+		if closeoutInitiated != nil {
+			// Between is inclusive, so the end date is set to 1 milsecond prior to the next day
```
leia-sefkin, line 374: appreciate the helpful breadcrumb here!

src/pages/Office/ServicesCounselingQueue/ServicesCounselingQueue.jsx
```diff
@@ -144,21 +226,23 @@ const ServicesCounselingQueue = () => {
...
-            defaultCanSort
-            defaultSortedColumns={[{ id: 'submittedAt', desc: false }]}
-            disableMultiSort
-            disableSortBy={false}
-            columns={columns()}
-            title="Moves"
-            handleClick={handleClick}
-            useQueries={useServicesCounselingQueueQueries}
-          />
+          <div>
```
leia-sefkin, line 229: not a hack! this is a totally legit way to get this working.

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -1346,6 +1346,20 @@ func QueueMoves(moves []models.Move) *ghcmessages.QueueMoves {
...
 		}
+		var closeoutLocation string
+		if move.CloseoutOffice != nil {
+			closeoutLocation = move.CloseoutOffice.Name
+		}
+		var closeoutInitiated time.Time
+		for _, shipment := range move.MTOShipments {
+			if shipment.PPMShipment != nil {
+				if shipment.PPMShipment.SubmittedAt != nil {
+					if closeoutInitiated.Before(*shipment.PPMShipment.SubmittedAt) {
```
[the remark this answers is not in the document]
mboren: since it's not a pointer, `closeoutInitiated` should get initialized to the zero value for `time.Time`, which is [January 1, year 1, 00:00:00.000000000 UTC](https://pkg.go.dev/time#Time), which should be before any other date.

hmm i think there might be a bug here though. If there are no PPM shipments with `SubmittedAt` set, it looks like it will save the zero time on the payload, when it should really be nil.

pkg/services/order/order_fetcher.go
mboren: oh yeah there should definitely be a comment here!

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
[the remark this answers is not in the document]
mboren: Turns out `handlers.FmtDateTimePtr` converts a pointer to the zero time to nil, which prevents this from turning into a bug. Convenient, but surprising.
```
// FmtDateTimePtr converts pop type to go-swagger type
func FmtDateTimePtr(dateTime *time.Time) *strfmt.DateTime {
	if dateTime == nil || dateTime.IsZero() {
		return nil
	}
	return (*strfmt.DateTime)(dateTime)
}
```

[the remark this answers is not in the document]
mboren: This won't change the functionality at all, but i'm going to combine some of these `if` statements here to cut down on the level of nesting

PR #9708, opened 2022-12-06
[MB-14176] PPM Closeout flow Cypress tests

duncan-truss:
## Summary

This PR adds Cypress tests that has the customer completing the entire closeout flow from beginning to end.  There's also an end to end test that goes back and edits existing closeout documents.  I had to add in pro-gear tests and expand on some other page specific forms to fill out.

The three main issues I encountered were:
- The About Your PPM form calls the postal code validator when the actual move date datepicker value is typed in instead of selected from the calendar view.  This is similar to the bug Scott fixed for our jest tests, the async validators get run on every onchange event because origin and destination postal code fields have pre-filled values.
- We still have the routing bug for when editing About Your PPM from the review page so I had to complete filling out an extra weight ticket because we can't yet delete review items.
- Our update PPM service was using a fetcher that was missing pro-gear and expenses relations, so when editing About Your PPM we would update Redux with a PPM shipment value missing those associations.  I fixed the fetcher used by the updater to include those relations instead of trying to merge the Redux entities.

##### Terminal 1

```sh
make e2e_test_fresh
```

### Additional steps

1. From the Cypress menu select entireShipmentCloseout to run the desktop and mobile happy path and edits tests.

### Frontend

### Backend


cypress/support/ppmCustomerShared.js
```diff
@@ -80,19 +80,35 @@ export function navigateFromHomePageToReviewPage(isMoveSubmitted = false) {
 }
 
 export function fillOutAboutPage(selectAdvance) {
-  cy.get('input[name="actualMoveDate"]').clear().type('01 Feb 2022').blur();
+  cy.intercept('GET', '**/internal/rate_engine_postal_codes/**?postal_code_type=origin').as('validateOriginZIP');
+  cy.intercept('GET', '**/internal/rate_engine_postal_codes/**?postal_code_type=destination').as('validateDestZIP');
+
+  // editing this field with the keyboard instead of the date picker runs async validators for pre-filled postal codes
+  // this helps debounce the API calls that would be triggered in quick succession
+  cy.get('input[name="actualMoveDate"]').as('actualMoveDate').clear().type('0');
```
duncan-truss, line 88: This avoids successive API calls to validate the pre-filled postal code values with every character change.  We could do something similar with the paste route that Scott did or I saw we could manually set the value with invoke() and trigger a change event.  In practice the user selects the date from the calendar datepicker, which would only cause one onchange event causing the other validators to run.

pkg/services/ppmshipment/ppm_shipment_updater.go
```diff
@@ -43,7 +43,7 @@ func (f *ppmShipmentUpdater) updatePPMShipment(appCtx appcontext.AppContext, ppm
 		return nil, nil
 	}
 
-	oldPPMShipment, err := models.FetchPPMShipmentFromMTOShipmentID(appCtx.DB(), mtoShipmentID)
+	oldPPMShipment, err := FindPPMShipmentByMTOID(appCtx, mtoShipmentID)
```
duncan-truss, line 46: The model fetcher being used was not eager loading the progear and expenses associations so when the Redux state was updated it looked like none existed on the review page when I was doing the edit test.

cypress/support/ppmCustomerShared.js
pearl-truss: I think it would be nice to use the paste method that Scott introduced so there's some consistency

pearl-truss (approved):
LGTM

cypress/support/ppmCustomerShared.js
duncan-truss: Scott did it with testing-library, to do it in Cypress would require something like:

```javascript
cy.get('input[name="actualMoveDate"]').click().invoke('val', '01 Feb 2022').trigger('change').blur()
```

I think the reason I avoided that was because we're manually setting the input value and triggering the on change event ourselves rather than what feels more accurate behavior with type().

If I had more time I would actually try clicking on the date in the calendar but know that requires digging into the 3rd party library DOM.

scottmries: The `.paste()` method is only on `userEvent`s from `testing-library`, unfortunately; Cypress doesn't have it, but yeah,  you could do something along the lines of 

`cy.get('@actualMoveDate').invoke('val', '06 Dec 202').type('2').blur();`

(Oddly, calling `invoke('change')` does not seem to actually set the date picker's value, but `type(...)` does, of course.)

scottmries (approved):

cypress/support/ppmCustomerShared.js
pearl-truss: Ah okay, didn't realize that paste event wasn't in cypress :doh:

> If I had more time I would actually try clicking on the date in the calendar but know that requires digging into the 3rd party library DOM. The date value would also likely change as it displays the current month and that would impact things like the final incentive pricing.

Yeah that doesn't seem like something needed to do in this particular PR. Thanks for the explanations

duncan-truss:
@Mergifyio rebase

PR #9729, opened 2022-12-08
Update script to work locally

shimonacarvalho:
## Summary

The word `make` went missing from an accidental edit, adding it back. It doesn't make the tests all pass, but is better than the script failing due to the missing command. 

## Setup to Run Your Code

This should compile and run, even if the tests don't pass. 
```
make e2e_test_docker
```

### DIFF
--- a/scripts/run-e2e-test-docker
+++ b/scripts/run-e2e-test-docker
@@ -37,7 +37,7 @@ docker run \
        -e GOPATH \
        -e CIRCLECI=1 \
        ${circleci_build_image} \
-       server_build \
+       make server_build \
        bin/generate-test-data \
        bin/prime-api-client \
        bin/rds-ca-2019-root.pem \


reggieriser (approved):
I tried this out locally and eventually got a failure, but it almost certainly has nothing to do with your change.  Your change is definitely needed, so I'm approving.  Thanks for fixing (especially since it was my screw-up)! 🥇 

For future reference, the failure I get is:
```
[FAILED] [22:1208/193421.855952:ERROR:bus.cc(393)] Failed to connect to the bus: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory
```
I'm on Apple Silicon, if that matters.

shimonacarvalho:
Strange. I merged, but this script needs more work for sure. I could at least get all the tests to run on my machine.

PR #9730, opened 2022-12-08
Mb 14489 update fields react admin

dymonmoore:
## Summary
This pull request is an incremental change apart of the React Admin upgrade. Here's what was refactored:

- The record prop is no longer injected to Field components, they useRecordContext() instead. 

- AddLabel props have been removed

The app will not compile until the rest of the upgrade changes are complete. This pull request cannot be tested at this time. For now, all is needed is a code review.


PDickeyTruss (approved):
Looks great!

PR #9734, opened 2022-12-09
[MB-14451] Cleanup evaluation and travel time length minutes fields

mboren:
## Summary

Previous work removed these fields from use, and switched to using timestamps instead, and in this PR I'm just cleaning up the last few spots where they are referenced.
I deleted the columns from the database, removed them from some example data used in tests and storybook, and removed some references in PDF generation code.

We decided that it was not necessary to do a data migration before removing these columns.

## Setup to Run Your Code
Make sure migrations apply
```
make db_dev_migrate
```
And verify that the columns `travel_time_minutes` and `evaluation_length_minutes` are no longer present on the `evaluation_reports`

Also double check the QAEReportTable and QAEViolationsForm components in Storybook just to make sure that they render without errors and nothing looks weird.


dinabeigelman (approved):
approving noting there is no need for a design review bc there's nothing to look at here.

duncan-truss (approved):
DB changes look good, didn't see any other usages of those columns.

PR #9736, opened 2022-12-12
MB-14700: Cypress test + increase readability of Final Incentive

ronaktruss:
## Summary

This adds cypress tests to test the final incentive amount and passes all of the weight tickets to the ppmShipmentUpdater when the weightTicketUpdater is called and the total weights have changed.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the customer app
1. Login as ppm.test.user1@example.com to create some sample ppmShipments with data to generate estimated Incentives
1. Login as weightTicketPPM@ppm.approved and recreate those scenarios in by modifying the about page and the weightTickets. Ensure that the calculated final incentive matches those in the above step.

### Backend


pkg/services/ppmshipment/ppm_estimator.go
```diff
@@ -184,29 +179,26 @@ func (f *estimatePPM) finalIncentive(appCtx appcontext.AppContext, oldPPMShipmen
...
-	var changedWeight unit.Pound
-	if len(newPPMShipment.WeightTickets) == 1 {
-		weightTicket := newPPMShipment.WeightTickets[0]
-		updatedWeightTicketID = weightTicket.ID
-		if weightTicket.FullWeight != nil && weightTicket.EmptyWeight != nil {
-			changedWeight = *weightTicket.FullWeight - *weightTicket.EmptyWeight
-		}
-	}
-
+// if second argument doesn't have weightTickets assume weight is the same (when coming here from ppmShipmentUpdater weightTickets is not populated)
```
felipe-lee, line 182: **issue**: In testing locally, it looks like there are weight tickets when going only through the ppm shipment updater. It looks like the fetcher that the updater uses does grab the weight tickets: https://github.com/transcom/mymove/blob/20249fee02c538156a2d94516a935c0479638f9b/pkg/services/ppmshipment/ppm_shipment_fetcher.go#L36-L47

So, the comment doesn't seem to match what is happening.

Also, should we be copying over the original total weight? What if they delete all their weight tickets? I know we won't let them get to the final incentive page, but that should also mean we have a new total weight of zero here so that the `shouldSkipCalculatingFinalIncentive` func can properly skip the calculation. I guess that raises another question, of should we be setting the final incentive to `nil` if we had calculated it, but then they delete all their weight tickets?

I haven't tested locally exactly what happens when weight tickets are deleted. The page won't let me delete for some reason, but it's the end of my day so I'll look into that tomorrow.

felipe-lee (commented):
Thanks for making these updates! I think this is easier to follow, at least for me. I do have some questions about total weight if weight tickets are deleted, but otherwise, this lgtm.

pkg/services/ppmshipment/ppm_estimator.go
ronaktruss: You're right that comment is no longer accurate. I didn't realize that logic was changed as part of merge in the previous final incentive code change. I'll update that.

I was under the impression that the delete functionality was still pending. Yeah, I think we should be resetting the finalIncentive to nil when we are missing information to calculate it and "skip calculations" only when the data hasn't changed. I'll make some changes accordingly.

felipe-lee: Ooh, that explains why I can't delete 😆 I guess I forgot we hadn't actually implemented that yet.

NamibiaTorres (commented):
Looks like you're close. It looks like all you have to do is increase the test coverage.

pkg/services/ppmshipment/ppm_estimator.go
```diff
@@ -168,45 +167,45 @@ func (f *estimatePPM) finalIncentive(appCtx appcontext.AppContext, oldPPMShipmen
...
+	if !isMissingInfo {
+		skipCalculateFinalIncentive = shouldSkipCalculatingFinalIncentive(newPPMShipment, &oldPPMShipment, originalTotalWeight, newTotalWeight)
+		if !skipCalculateFinalIncentive {
+			finalIncentive, err = f.calculatePrice(appCtx, newPPMShipment, newTotalWeight)
+			if err != nil {
+				return nil, err
+			}
 		}
+	} else {
+		finalIncentive = nil
```
pearl-truss, line 185: Question/thought: is setting `finalIncentive` to nil enough to communicate that info is missing? Or is it worth communicating with logging or even throwing an err?

ronaktruss: There are times where we would expect info to be missing and the final incentive to be set to nil. For example when first going through the closeout phase on the about page a user would provide all of the necessary information except they wouldn't have any weightTickets (ie. have a weight of 0) that would result in the final incentive being set to nil. 

Because we don't necessarily expect all of the information at all times setting finalIncentive to nil is not really a special scenario. Throwing an error is a bit heavy handed as this is the expected scenario at times (another example is when a user removes their final weight ticket). 

That said I think logging might be worthwhile. Atm I don't think logging is particularly useful as we have a fairly linear process for when and how this information is added, but this might change as we work on the office approval/rejections. Happy to be convinced otherwise if you think there is a strong case for adding them.

felipe-lee: **thought**: I agree that not throwing an error is likely best here. I'd be fine with or without logging in this case since it's mainly keeping our data making sense and this scenario could easily be hit, at least, once we implement deleting weight tickets.

felipe-lee (approved):
Thanks for making those tweaks and adding the extra test. This lgtm!

PR #9748, opened 2022-12-15
MB-14756 Make swagger_autorebuild the default

shimonacarvalho:
## Summary

This changes swagger_generate to skip the prompt by default. 
Devs can re-enable the prompt by setting `SWAGGER_AUTOREBUILD=0` locally.

##### Terminal 1

```sh
direnv allow 
touch swagger/prime.yaml 
make swagger_generate
## Should bundle the api files...
./scripts/openapi bundle -o swagger/ ## Bundles the API definition files into a complete specification

touch swagger/prime.yaml 
SWAGGER_AUTOREBUILD=0 make swagger_generate
## Should prompt first before bundling the api files
Unexpected changes found in swagger build files. Code may be overwritten.
Continue with rebuild? [y/N] : 
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -310,7 +310,7 @@ SWAGGER_AUTOREBUILD=1
 endif
 SWAGGER_FILES = $(shell find swagger swagger-def -type f)
 .swagger_build.stamp: $(SWAGGER_FILES)
-ifndef SWAGGER_AUTOREBUILD
+ifeq ($(SWAGGER_AUTOREBUILD),0)
 ifneq ("$(shell find swagger -type f -name '*.yaml' -newer .swagger_build.stamp)","")
 	@echo "Unexpected changes found in swagger build files. Code may be overwritten."
 	@read -p "Continue with rebuild? [y/N] : " ANS && test "$${ANS}" == "y" || (echo "Exiting rebuild."; false)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -310,7 +310,7 @@ SWAGGER_AUTOREBUILD=1
 endif
 SWAGGER_FILES = $(shell find swagger swagger-def -type f)
 .swagger_build.stamp: $(SWAGGER_FILES)
-ifndef SWAGGER_AUTOREBUILD
+ifeq ($(SWAGGER_AUTOREBUILD),0)
```
ahobson, line 313: I'm not sure why you would need to change this? Did you see any problems just changing `.envrc`?

shimonacarvalho: Yes so once I defined it in `.envrc` then `ifndef` is never the case. `SWAGGER_AUTOREBUILD` is always "defined" even if set to 0, and the prompt never triggers.

ahobson: Ok! The Makefile is also invoked by CI (and thus why the variable is set on line 309 of the Makefile) and so I'm trying to ensure I understand if this Makefile change is necessary or just something to try to make things clearer.

shimonacarvalho: Yup since I had to change it, I was worried about any side effects in CI. To my eye, it seems correct. 
You think this is good then?

ahobson (approved):

PR #9747, opened 2022-12-14
MB-14714 Implement BuildEntitlement function

shimonacarvalho:
## Summary

This ticket replaces uses of `testdatagen.MakeEntitlement` with the new `factory.BuildEntitlement`.

## Setup to Run Your Code

1. Run `make server_test` and make sure tests pass

PR #9754, opened 2022-12-15
MB-14538 factory.BuildAddress added as new version of testdatagen.MakeAddress

StuTruss:
## Summary

This ticket adds a factory package with function BuildAddress intended to be an eventual replacement for testdatagen.MakeAddress
* Created BuildAddress and shared functions
* Created tests for all of those. 

## Setup to Run Your Code

### Frontend

- No frontend changes

### Backend

- No schema changes


pkg/factory/address_factory.go
```diff
@@ -0,0 +1,52 @@
...
+
+// BuildAddress creates a single Address and associated service member.
+func BuildAddress(db *pop.Connection, customs []Customization, traits []Trait) models.Address {
+	customs = setupCustomizations(customs, traits)
+
+	// Find address assertion and convert to models address
+	var cAddress models.Address
+	if result := findValidCustomization(customs, Address); result != nil {
+		cAddress = result.Model.(models.Address)
+	}
```
shimonacarvalho, line 19: Need to check LinkOnly and return if true... see the sample here:

https://docs.google.com/document/d/12XMhDQI7di_C6bV-c7tM_1UjqBNnZQJPC-bp0I_MWhg/edit#heading=h.yqt2mj53s8xo

pkg/factory/address_factory.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"github.com/go-openapi/swag"
+	"github.com/gobuffalo/pop/v6"
+
+	"github.com/transcom/mymove/pkg/models"
+	"github.com/transcom/mymove/pkg/testdatagen"
+)
+
+// BuildAddress creates a single Address and associated service member.
+func BuildAddress(db *pop.Connection, customs []Customization, traits []Trait) models.Address {
```
shimonacarvalho, line 12: Does BuildAddress create an "associated service member"?

pkg/factory/address_factory.go
```diff
@@ -0,0 +1,52 @@
...
+	if result := findValidCustomization(customs, Address); result != nil {
+		cAddress = result.Model.(models.Address)
+	}
+
+	// Create default Address
+	address := models.Address{
+		StreetAddress1: "123 Any Street",
+		StreetAddress2: swag.String("P.O. Box 12345"),
+		StreetAddress3: swag.String("c/o Some Person"),
+		City:           "Beverly Hills",
```
shimonacarvalho, line 26: We are looking to switch away from `swag` functions to our own `models.StringPointer` type functions. Exact same usage... `models.StringPointer("my string")`

pkg/factory/address_factory.go
```diff
@@ -0,0 +1,52 @@
...
+
+// BuildDefaultAddress makes an Address with default values
+func BuildDefaultAddress(db *pop.Connection) models.Address {
+	FetchOrBuildTariff400ngZip3(db, nil, nil)
+	return BuildAddress(db, nil, nil)
+}
+
+// BuildStubbedAddress returns a stubbed address without saving it to the DB
+func BuildStubbedAddress(customs []Customization, traits []Trait) models.Address {
+	return BuildAddress(nil, customs, traits)
```
shimonacarvalho, line 51: I'd delete that, we can simply replace the calls with the nil version. 

`BuildStubbedAddress(customs, traits)` is more typing than 
`BuildAddress(nil, customs, traits)` anyway.

pkg/factory/address_factory_test.go
```diff
@@ -0,0 +1,89 @@
...
+	defaultPostalCode := "90210"
+	customAddress1 := "101 This is Awesome Street"
+	suite.Run("Successful creation of default address", func() {
+		// Under test:      BuildAddress
+		// Mocked:          None
+		// Set up:          Create an Address with no customizations or traits
+		// Expected outcome:Address should be created with default values
+
+		address := BuildAddress(suite.DB(), nil, nil)
+		suite.Equal(defaultAddress1, address.StreetAddress1)
```
shimonacarvalho, line 16: Please test every single value of the created model. Easy for some error to slip through and never be caught. 

See here for example: 

https://github.com/transcom/mymove/pull/9747/files#diff-792615c80ad9f6877fddf43e042cfc40530512aa8ad6f5ec3d2e25d0b9839202

pkg/factory/address_factory_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	suite.Run("Successful creation of an address with customization", func() {
+		// Under test:      BuildAddress
+		// Set up:          Create an Address with a customized StreetAddress1 and no trait
+		// Expected outcome:Address should be created with custom street address
+		address := BuildAddress(suite.DB(), []Customization{
+			{
+				Model: models.Address{
+					StreetAddress1: customAddress1,
+				},
```
shimonacarvalho, line 28: Please customize and then test every single value of the created model. 
See here for example: 

https://github.com/transcom/mymove/pull/9747/files#diff-792615c80ad9f6877fddf43e042cfc40530512aa8ad6f5ec3d2e25d0b9839202

pkg/factory/traits.go
```diff
@@ -25,3 +29,54 @@ func GetTraitArmy() []Customization {
 		},
 	}
 }
+
+// GetTraitAddress2 is a sample GetTraitFunc
+func GetTraitAddress2() []Customization {
+	return []Customization{
```
shimonacarvalho, line 35: I think we should move these traits to the address factory file. If a trait is really focused on a particular model and not spread across many models, we can just put it with the factory for that model. 

I know we have this file which implies otherwise but I wanted to put it here as a holder for future multi-model traits. It will get too unwieldy if all the traits end up here with no grouping. 

I'll update the dev documentation.

shimonacarvalho (changes requested):
Hi Stuart - functionality looks good to me, except for the linkOnly check needs to be added. 

Requested a number of changes to testing/organization.

pkg/factory/address_factory_test.go
```diff
@@ -0,0 +1,89 @@
...
+		}, []Trait{
+			GetTraitAddress3,
+		})
+
+		suite.Equal(customAddress1, address.StreetAddress1)
+		suite.Equal("50309", address.PostalCode)
+	})
+
+	suite.Run("Successful creation of stubbed address", func() {
+		// Under test:      BuildAddress
```
shimonacarvalho, line 65: Nice job adding a stubbed test. Please also add a linkOnly test. Should be pretty similar.

pkg/factory/address_factory_test.go
```diff
@@ -0,0 +1,169 @@
...
+func (suite *FactorySuite) TestBuildAddress() {
+	defaultAddress1 := "123 Any Street"
+	defaultAddress2 := models.StringPointer("P.O. Box 12345")
+	defaultAddress3 := models.StringPointer("c/o Some Person")
+	defaultCity := "Beverly Hills"
+	defaultState := "CA"
+	defaultPostalCode := "90210"
+
+	customAddress1 := "101 This is Awesome Street"
+	customAddress2 := models.StringPointer("Unit 2525")
```
shimonacarvalho, line 14: If these variables are only used in a single subtest, move them into the subtest. That way it's easier to see all the code related to a test at once. This section is more for things that are shared across tests, which in general we want to avoid so each test is isolated and doesn't depend on other tests.

pkg/factory/address_factory_test.go
```diff
@@ -0,0 +1,169 @@
...
+		address := BuildAddress(suite.DB(), nil, nil)
+
+		// VALIDATE RESULTS
+		suite.Equal(defaultAddress1, address.StreetAddress1)
+		suite.Equal(defaultAddress2, address.StreetAddress2)
+		suite.Equal(defaultAddress3, address.StreetAddress3)
+		suite.Equal(defaultCity, address.City)
+		suite.Equal(defaultState, address.State)
+		suite.Equal(defaultPostalCode, address.PostalCode)
+		suite.Equal(models.StringPointer("US"), address.Country)
```
shimonacarvalho, line 34: This looks like it's comparing pointers, not values. My guess is it actually does compare the values but it's not clear. To make it more clear, deference both pointers so you know it's comparing values. 

`suite.Equal("US", *address.Country)`

pkg/factory/address_factory_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+		// Count how many addresses are in the DB, no new addresses should have been created
+		count, err := suite.DB().Count(&models.Address{})
+		suite.NoError(err)
+		suite.Equal(precount, count)
+	})
+
+	suite.Run("Successful creation of address with linked address", func() {
+		linkedAddress := BuildAddress(suite.DB(), []Customization{
```
shimonacarvalho, line 142: Not quite what we're looking to test for linkOnly. 

What we want to test is simply that if a BuildAddress is passed a model marked as LinkOnly, no new record is created in the database. So here's an example of how we do that (line 96)

https://github.com/transcom/mymove/pull/9747/files#diff-792615c80ad9f6877fddf43e042cfc40530512aa8ad6f5ec3d2e25d0b9839202R96

shimonacarvalho (changes requested):
Some further changes, thanks for all the good work on this one!

shimonacarvalho (approved):
Looks good!

PR #9757, opened 2022-12-15
MB-14711 API endpoints for transportation office search

PDickeyTruss:
## Summary

This PR adds a new 'Transportation Offices' API endpoint to both the GHC API and the internal API. Currently the end point just returns mock data as the search functionality has not been built yet. 

##  How to test
This PR should be validated by running API requests and reviewing the responses are as expected. This can be done via [postman](https://transcom.github.io/mymove-docs/docs/tools/postman/setting-up-postman), the [swagger UI](http://officelocal:3000/swagger-ui/ghc.html#), or numerous other methods to make a GET request. 

## Verification 
- [ ] The `/transportation-offices` endpoint is present in the GCH API  and returns reasonable mock data when requested. 
- [ ] The `/transportation-offices` endpoint is present in the Internal API and returns reasonable mock data when requested. 

## Screenshots
Screenshot 2022-12-16 at 7 32 50 AM
Screenshot 2022-12-16 at 7 33 18 AM


swagger-def/internal.yaml
```diff
@@ -4172,6 +4176,36 @@ paths:
           description: transportation office not found
         '500':
           description: internal server error
+  /transportation_offices:
```
duncan-truss, line 4179: I think you used hyphenated in the ghc.yaml and underscores here.  I believe the style guide wants us to coalesce on using hyphens (kebab case) even though I know there are a lot of legacy paths in this file. https://transcom.github.io/mymove-docs/docs/api/guides/guide-to-creating-an-endpoint#adding-a-new-entry-into-the-yaml

swagger-def/internal.yaml
```diff
@@ -4172,6 +4176,36 @@ paths:
...
+    get:
+      produces:
+        - application/json
+      summary: Returns the transportation offices matching the search query
+      description: Returns the transportation offices matching the search query
+      operationId: getTransportationOffices
+      tags:
+        - transportation_offices
```
duncan-truss, line 4187: I think the linter complains when these tags aren't first declared in a global block at the top of the file.  I'm surprised we're missing that entirely from the internal.yaml.  Here's the example in the ghc.yaml https://github.com/transcom/mymove/blob/67ac2060efcc8ef6ed63465785e5cefa7d435f01/swagger-def/ghc.yaml#L15

```sh
 lint-yaml-with-spectral swagger/internal.yaml
```

Screen Shot 2022-12-16 at 11 55 31 AM

swagger-def/internal.yaml
```diff
@@ -4172,6 +4176,36 @@ paths:
...
+      parameters:
+        - in: query
+          name: search
+          type: string
+          required: true
+          description: Search string for transportation offices
```
duncan-truss, line 4193: I wonder if we want to match the front-end constraint of requiring the string length to be at least 2 characters long here?  You may need to use a `pattern` to achieve that.  Does that feel too hard-coded to you or maybe it protects us from some long running query searches if we don't vet it elsewhere?

swagger-def/internal.yaml
PDickeyTruss: Good call, I was matching what I saw as the prevalent pattern in each file but if we have in our style guide (and linting) that we want to use hyphens I am all for it! Code updated ✅

swagger-def/internal.yaml
PDickeyTruss: Good eye, I for sure overlooked the global tags block. Super weird that it didn't exist at all in internal. I went ahead and added the full tags block for all tags in internal as adding a global block with just `transportation_offices` felt a bit silly 😁

swagger-def/internal.yaml
PDickeyTruss: Good idea, I see no reason anyone should be searching on less than 2 characters. I was able to fix by adding a `minLength` to the search param.

swagger-def/ghc.yaml
```diff
@@ -2473,7 +2474,21 @@ paths:
...
+              status,
+              requestedMoveDate,
+              submittedAt,
+              originGBLOC,
+              originDutyLocation,
+              destinationDutyLocation,
+              ppmType,
+              closeoutInitiated,
+              closeoutLocation,
+            ]
```
PDickeyTruss, line 2491: This was my editor auto applying formatting per our project rules. I initially avoided including it since it is irrelevant to this PR but it managed to slip in the most recent commit.  Now that it is here, I am inclined to leave it so others don't have to dance around it in the future.

swagger-def/internal.yaml
duncan-truss: Brilliant, I'll take 70 less warnings 🎉

swagger-def/internal.yaml
duncan-truss: Great, I knew about minimum for numeric types but didn't know if swagger 2.0 had anything else to offer.

pkg/handlers/internalapi/tranportation_office.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/transcom/mymove/pkg/handlers/internalapi/internal/payloads"
+	"github.com/transcom/mymove/pkg/services"
+)
+
+type GetTransportationOfficesHandler struct {
+	handlers.HandlerConfig
+	services.TransportationOfficesFetcher
+}
+
+func (h GetTransportationOfficesHandler) Handle(params transportationofficeop.GetTransportationOfficesParams) middleware.Responder {
```
duncan-truss, line 19: For customer API endpoints our middleware doesn't check that the session belongs to the customer app and that the user's session is a customer one like it does for the office and admin. https://github.com/transcom/mymove/blob/79467c7dbf2fa390aa37929db1b8c7f25c9f6f67/pkg/handlers/authentication/auth.go#L113

Do you mind adding something for the 403 status check here and a basic handler test for this and the office endpoint?  I understand the response data will change so no need to do extra work there.

pkg/services/transportation_office/transportation_office_fetcher.go
```diff
@@ -0,0 +1,68 @@
+package pwsviolation
```
leia-sefkin, line 1: package needs updating, should be transportationoffice (i believe_

pkg/services/transportation_office/transportation_office_fetcher.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func NewTransportationOfficesFetcher() services.TransportationOfficesFetcher {
+	return &transportationOfficesFetcher{}
+}
+
+func (o transportationOfficesFetcher) GetTransportationOffices(appCtx appcontext.AppContext) (*models.TransportationOffices, error) {
+
+	// Mock response values, to be replaced with real values in later ticket
+	mockTransportationOffices := models.TransportationOffices{
```
leia-sefkin, line 23: ty for stubbing this out, makes finishing and merging my work in with yours nice and easy

swagger-def/ghc.yaml
leia-sefkin: i'm all for letting linters do their thing

swagger-def/internal.yaml
leia-sefkin: out of curiosity: do we need to worry about capping search length at a certain maximum here?

swagger-def/ghc.yaml
```diff
@@ -2817,6 +2832,37 @@ paths:
...
+      summary: Returns the transportation offices matching the search query
+      description: Returns the transportation offices matching the search query
+      operationId: getTransportationOffices
+      tags:
+        - transportationOffice
+      parameters:
+        - in: query
+          name: search
+          type: string
+          required: true
```
leia-sefkin, line 2848: suggestion: this could be optional, so a blank search returns the entire list of transportation offices

leia-sefkin (approved):
overall LGTM 👍 👍 👍  i had a couple of comments and questions but nothing blocking

pkg/services/transportation_office/transportation_office_fetcher.go
PDickeyTruss: Good spot! 🦅 👁️  Fixed ✅

pkg/handlers/internalapi/tranportation_office.go
PDickeyTruss: Added 403 status checks and handler tests. Thanks for calling this out! 🙇

swagger-def/internal.yaml
PDickeyTruss: 🎉

swagger-def/ghc.yaml
duncan-truss: I know the current component requires at [least 2 characters before trigger an API request](https://github.com/transcom/mymove/blob/e67dc24c8abb5c90cf88b1501d8d77c95f8dc741/src/components/DutyLocationSearchBox/DutyLocationSearchBox.jsx#L42).  Does the Postgres similarity score weighting require a minimum string length to be useful?

pkg/handlers/ghcapi/transportation_offices_test.go
```diff
@@ -0,0 +1,30 @@
...
+	handler := GetTransportationOfficesHandler{
+		HandlerConfig:                suite.HandlerConfig(),
+		TransportationOfficesFetcher: fetcher}
+
+	response := handler.Handle(params)
+	suite.Assertions.IsType(&transportationofficeop.GetTransportationOfficesOK{}, response)
+	responsePayload := response.(*transportationofficeop.GetTransportationOfficesOK)
+
+	// Validate outgoing payload
+	suite.NoError(responsePayload.Payload.Validate(strfmt.Default))
```
duncan-truss, line 29: 👏 Thanks for remembering this.

The params in this test may be invalid though at the swagger level because search is required and has the minLength attribute, I would suggest providing a value for that here.  I think we are only able to call Validate() for body params and not query params in our tests.

duncan-truss (approved):
Looks great 🚢

PR #9763, opened 2022-12-16
[ MB-13151 ] Update the GHC diesel fuel price task to use v2 of the EIA API

mmayo-truss:
Changes include:
- modification to api url in .envrc and fetcher
- Update to data structure to account for changes to the response data in v2
- addition of more validation checks
- updates to existing tests
- creation of validation tests
- additional logging

## Summary

The [EIA fuel APv1](https://www.eia.gov/opendata/documentation/APIv2.1.0.pdf) that we were calling had support discontinued November 2022. This PR switches to v2.1.0. The functionality will remain the same: each week a new price will be added to the db in table `ghc_diesel_fuel_prices`. 

Some notes on the changes: 
-  In v2.1.0 the EIA api endpoint `seriesid` was created for easier backwards compatibility with v1 users
- The date format of the returned periods was changed to YYYY-MM-DD
- The structure of the response data changed from an array of arrays of period and price, to an array of objects with key value pairs of more information
-  Due to this increase in information, more validators were added to ensure that we are recieving and storing the correct fuel price
- The last updated field that was returned in v1 is no longer being returned in v2. Any references to this have been removed. It was only being used for logging purposes and wasn't being stored in the db

##### Terminal 1

Pull this branch down locally and run the following:

`make tasks_save_ghc_fuel_price_data`

### Additional steps

1. The first time you run this it will take around 15 minutes to build the docker image, however after that it will be quick (assuming no changes to the code was made)
2. If you want to rerun the task, you can 
  - modify the price on the newest saved row using a db editor like Postico, the task will then update the price back to the retrieved value from the API call
  - delete the newest row, the task will save a new row again
  - rerun without any db changes and the task will just log that there was nothing to save.


pkg/services/ghcdieselfuelprice/ghc_diesel_fuel_price_validator.go
```diff
@@ -0,0 +1,139 @@
...
+		message: message,
+	}
+}
+
+// Error is the string representation of the InternalServerError
+func (e GHCAPIValidationError) Error() string {
+	return e.message
+}
+
+func (e EIAData) validateEIAData() error {
```
michaelvaldes-truss, line 21: Ooh, this validator is a nice touch!

pkg/services/ghcdieselfuelprice/ghc_diesel_fuel_price_validator.go
```diff
@@ -0,0 +1,139 @@
...
+func (e EIAData) validateUnits() error {
+	if e.ResponseData.FuelData[0].Units != "$/GAL" {
+		return NewGHCAPIValidationError(
+			fmt.Sprintf("Expected Units to be $/GAL, received %s", e.ResponseData.FuelData[0].Units))
+	}
+	return nil
+}
+
+func (e EIAData) validateDateFormat() error {
+	if e.ResponseData.DateFormat != "YYYY-MM-DD" {
```
michaelvaldes-truss, line 119: Should we be validating this field, or using it to parse the date value?

michaelvaldes-truss (approved):
Tested locally running the task and fetching the prices, and it works! The extra validation/testing is really great.

pkg/services/ghcdieselfuelprice/ghc_diesel_fuel_price_validator.go
mmayo-truss: I went back and forth on this. After looking through the EIA documentation, I haven't found a way to specify the date format that is returned. I will look into this a little bit more.

mmayo-truss: @michaelvaldes-truss I updated the validator and the publication date function to reference a map of EIA date formats.

michaelvaldes-truss (approved):
Parsing the dateformat and using that to parse the date is pretty sweet

PR #9771, opened 2022-12-19
Update gin dependency and remove "replace" directive in go.mod

reggieriser:
## Summary

We have a "replace" directive in `go.mod` to work around a problem that originated from `gin` using an old dependency URL for `urfave/cli`.  I noticed that this URL has since been updated in the `gin` source, so this PR updates `gin` and cleans up the `replace` directive from `go.mod`.

As an aside, this is not a dependency that gets updated by dependabot (I presume because it doesn't have explicit releases).

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

Just make sure you can `make server_run` fine locally (this is where `gin` is used for live-reloading).  You could change some go source and the running app should recompile and update in your console.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/alexedwards/scs/v2 v2.5.0
 	github.com/aws/aws-sdk-go v1.44.162
 	github.com/benbjohnson/clock v1.3.0
-	github.com/codegangsta/gin v0.0.0-20171026143024-cafe2ce98974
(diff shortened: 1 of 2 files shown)


ruizajtruss (approved):
Everything compiles fine and seems to be in working order 🎉

PR #9790, opened 2022-12-22
MB-14781 Add additional triggers for tables for DMS

YanZ777:
## Summary

Add additional triggers for tables so that DMS can be better supported.
Triggers for tables were initially only added for move history support, this adds an additional set of triggers for other tables. 

All data that is considered "operational" needs to have triggers added to them so that they can be recorded by the `audit_history` table.

### Tables that have triggers already prior to this PR
- `moves`
- `mto_agents`
- `mto_service_item_customer_contacts`
- `mto_service_item_dimensions`
- `mto_service_items`
- `mto_shipments`
- `orders`
- `payment_requests`
- `proof_of_service_docs`
- `reweighs`
- `service_members`
- `storage_facilities`
- `backup_contacts`
- `user_uploads`

### Tables for which triggers have **NOT** been added
These seemed fairly clear to me that they should not have triggers on them.

- `audit_history`
- `schema_migration`

### Specific review request questions

What are the implications of adding triggers to all these tables from the perspective of the UI?

Do I need to add triggers for the following tables? 
- `archived_access_codes`
- `archived_move_documents`
- `archived_moving_expense_documents`
- `archived_personally_procured_moves`
- `archived_reimbursements`
- `archived_signed_certifications`
- `archived_weight_ticket_set_documents`
- `client_certs`
- `contractors`
- `signed_certifications`
- `tariff400ng_full_pack_rates`
- `tariff400ng_full_unpack_rates`
- `tariff400ng_item_rates`
- `tariff400ng_items`
- `tariff400ng_linehaul_rates`
- `tariff400ng_service_areas`
- `tariff400ng_shorthaul_rates`
- `tariff400ng_zip3s`
- `tariff400ng_zip5_rate_areas`
- `traffic_distribution_lists`
- `webhook_notifications`
- `webhook_subscriptions`

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Perform some actions on various different things. 
2. In whatever database viewer you are using, you should see new entries in the audit history table.

### DIFF
--- a/migrations/app/schema/20221221204441_add_additional_triggers_for_dms.up.sql
+++ b/migrations/app/schema/20221221204441_add_additional_triggers_for_dms.up.sql
@@ -1,0 +1,702 @@
+
+SELECT add_audit_history_table(
+    target_table := 'backup_contacts',
+    audit_rows := BOOLEAN 't',
+    audit_query_text := BOOLEAN 't',
+    ignored_cols := ARRAY[
+        'created_at',
+        'updated_at'
+    ]
+);
+
+
+SELECT add_audit_history_table(
+    target_table := 'customer_support_remarks',
+    audit_rows := BOOLEAN 't',
+    audit_query_text := BOOLEAN 't',
+    ignored_cols := ARRAY[
+        'created_at',
+        'updated_at'
+    ]
+);
+
+SELECT add_audit_history_table(
+    target_table := 'distance_calculations',
+    audit_rows := BOOLEAN 't',
+    audit_query_text := BOOLEAN 't',
+    ignored_cols := ARRAY[
+        'created_at',
+        'updated_at'
+    ]
+);
+
+SELECT add_audit_history_table(
+    target_table := 'documents',
+    audit_rows := BOOLEAN 't',
+    audit_query_text := BOOLEAN 't',
+    ignored_cols := ARRAY[
+        'created_at',
+        'updated_at'
+    ]
+);
+
+SELECT add_audit_history_table(
+    target_table := 'duty_location_names',
+    audit_rows := BOOLEAN 't',
+    audit_query_text := BOOLEAN 't',
+    ignored_cols := ARRAY[
+        'created_at',
+        'updated_at'
+    ]
+);
+
+SELECT add_audit_history_table(
+    target_table := 'duty_locations',
+    audit_rows := BOOLEAN 't',
+    audit_query_text := BOOLEAN 't',
+    ignored_cols := ARRAY[
+        'created_at',
+        'updated_at'
+    ]
+);
+
+SELECT add_audit_history_table(
+    target_table := 'edi_errors',
+    audit_rows := BOOLEAN 't',
+    audit_query_text := BOOLEAN 't',
+    ignored_cols := ARRAY[
+        'created_at',
+        'updated_at'
+    ]
+);
+
+SELECT add_audit_history_table(
+    target_table := 'edi_processings',
+    audit_rows := BOOLEAN 't',
+    audit_query_text := BOOLEAN 't',
+    ignored_cols := ARRAY[
+        'created_at',
+        'updated_at'
+    ]
+);
+
+SELECT add_audit_history_table(
+    target_table := 'edi_processings',
+    audit_rows := BOOLEAN 't',
+    audit_query_text := BOOLEAN 't',
+    ignored_cols := ARRAY[
+        'created_at',
+        'updated_at'
+    ]
+);
+
+SELECT add_audit_history_table(
+    target_table := 'electronic_orders',
+    audit_rows := BOOLEAN 't',
+    audit_query_text := BOOLEAN 't',
+    ignored_cols := ARRAY[
+        'created_at',
+        'updated_at'
+    ]
+);
+
+SELECT add_audit_history_table(
+    target_table := 'electronic_orders_revisions',
+    audit_rows := BOOLEAN 't',
+    audit_query_text := BOOLEAN 't',
+    ignored_cols := ARRAY[
+        'created_at',
+        'updated_at'
+    ]
+);
+
+SELECT add_audit_history_table(
+    target_table := 'entitlements',
+    audit_rows := BOOLEAN 't',
+    audit_query_text := BOOLEAN 't',
+    ignored_cols := ARRAY[
+        'created_at',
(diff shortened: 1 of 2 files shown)


mmayo-truss:
Answer to "What are the implications of adding triggers to all these tables from the perspective of the UI?": 

If you only add the triggers and do nothing else: Nothing. We will store all data associated with all CREATE, UPDATE, and DELETE events for each of the newly added tables on the `audit_history` table . 

However, in order for the data to be returned to the move history table changes would need to be made to the SQL query in `move_history_fetcher.sql`. 

And after that data is returned to the front end, event templates associated with its respective `event_name` would need to be added to the front end code in order to display more than a generic row.

migrations/app/schema/20221221204441_add_additional_triggers_for_dms.up.sql
```diff
@@ -0,0 +1,649 @@
...
+    audit_rows := BOOLEAN 't',
+    audit_query_text := BOOLEAN 't',
+    ignored_cols := ARRAY[
+        'created_at',
+        'updated_at'
+    ]
+);
+
+SELECT add_audit_history_table(
+    target_table := 'fuel_eia_diesel_prices',
```
mmayo-truss, line 122: Note: I do not believe we are currently using this table. It was replaced with `ghc_diesel_fuel_prices` in [this task](https://dp3.atlassian.net/browse/MB-2652). I don't think it hurts anything to keep it here though.

felipe-lee:
We talked a bit at BE/FE check-in and one of the things I mentioned was that the archive tables: 

```
archived_access_codes
archived_move_documents
archived_moving_expense_documents
archived_personally_procured_moves
archived_reimbursements
archived_signed_certifications
archived_weight_ticket_set_documents
```

Are from when we had data in Production (original iteration of PPMs). We archived the original PPM data in https://dp3.atlassian.net/browse/MB-11261 because the way PPMs needed to be implemented changed enough that we wanted tie into the current state of things better (ppms being a shipment type vs a move type). We don't really have a fully ironed out archival process, so there are still open questions about things like, can we extract the data from the tables and get rid of the old tables, should we continue to archived data for tables that have both a current and archived version (e.g. signed certs), etc.

ahobson:
> We don't really have a fully ironed out archival process, so there are still open questions about things like, can we extract the data from the tables and get rid of the old tables, should we continue to archived data for tables that have both a current and archived version (e.g. signed certs), etc.

I mean, that's what the data warehouse is for. I'll confess I think we're taking the wrong approach by using triggers to send deltas for data warehouse because that relies on us never missing a table or never making a mistake with our triggers. Personally, I would suggest we should have an ADR about this decision because we have effectively changed the AC from the prior definition

jesset-truss:
> So my question is: Do we need to display all changes to these tables in the UI, or is just storing this information in the audit_history table enough?

We have no need or requirement to display any of this specific information in the UI and that would not be part of this request. The information would be stored to the audit_history table though so it can be loaded to the the Data Warehouse for further processing, such as seeing fuel price change over time.

jesset-truss:
> We talked a bit at BE/FE check-in and one of the things I mentioned was that the archive tables:
> 
> ```
> archived_access_codes
> archived_move_documents
> archived_moving_expense_documents
> archived_personally_procured_moves
> archived_reimbursements
> archived_signed_certifications
> archived_weight_ticket_set_documents
> ```
> 
> Are from when we had data in Production (original iteration of PPMs). We archived the original PPM data in https://dp3.atlassian.net/browse/MB-11261 because the way PPMs needed to be implemented changed enough that we wanted tie into the current state of things better (ppms being a shipment type vs a move type). We don't really have a fully ironed out archival process, so there are still open questions about things like, can we extract the data from the tables and get rid of the old tables, should we continue to archived data for tables that have both a current and archived version (e.g. signed certs), etc.

For the Data Warehouse objectives this will not matter. If the trigger was on a table prior to archiving it then it would still have written to the audit_history on updates/inserts and would have been logged there with the original table name. I don't see any reason to change the archiving process as a result of Data Warehouse, we can remove the trigger from the archived table, or, if no new data is being written to the table it wouldn't matter anyway since we are only pulling from the audit_history table.

jesset-truss:
> I mean, that's what the data warehouse is for. I'll confess I think we're taking the wrong approach by using triggers to send deltas for data warehouse because that relies on us never missing a table or never making a mistake with our triggers. Personally, I would suggest we should have an ADR about this decision because we have effectively changed the AC from the prior definition

@ahobson - Which prior definition are you referring to? I agree with the comments regarding the ADR and we can do that!

ahobson:
> > I mean, that's what the data warehouse is for. I'll confess I think we're taking the wrong approach by using triggers to send deltas for data warehouse because that relies on us never missing a table or never making a mistake with our triggers. Personally, I would suggest we should have an ADR about this decision because we have effectively changed the AC from the prior definition
> 
> @ahobson - Which prior definition are you referring to? I agree with the comments regarding the ADR and we can do that!

The [Ensuring all of the MilMove data elements are captured](https://dp3.atlassian.net/wiki/spaces/MT/pages/1802797074/OLD+ARCHIVE+Advana+Data+Warehouse+Integration)

@jesset-truss  Data warehousing is an opportunity to have an independent system to capture the entire contents of the database (and changes) without the opportunity for Truss to make a mistake that results in data not being warehoused. By tying the implementation of changes to the H&A feature, it's now on Truss to never, ever do anything that prevents a table from being warehoused. If we do make that mistake, it can't be undone; we can't go back in time. That could literally lead to breaking conditions established by the GAO. That's not a risk I'm willing to take, but it's your head, not mine.

In addition, by tying the implementation of data warehousing (an async process) to H&A, it means tables we add to data warehousing will affect the performance of the UI for H&A. It means when we inevitably need to archive old moves and old history, it affects both the H&A feature and the data warehousing feature. We've tied two independent features together which will make it all the more difficult to make changes to one or the other.

Again, it's not my call, but I see this implementation as significantly increasing risk for MilMove.

eeeady:
Hey Drew - Let’s make sure we’re being thoughtful in our wider communications as we maintain our Truss values.

YanZ777:
Closing since we're no longer using the audit history tables for the data warehouse work.

PR #9795, opened 2022-12-22
MB-13828-progear-office-handler-complete-NT

NamibiaTorres:
## Summary

This PR adds error handling and unhappy path tests for the Progear Weight Ticket Handler

## Setup to Run Your Code

<details>
<summary>💻 You will need to use 1 terminal to test this locally.</summary>

##### Testing terminal

```sh
go test ./pkg/handlers/ghcapi --testify.m TestUpdateProGearWeightTicketHandler -v
```

</details>

### Backend


pkg/handlers/ghcapi/progear_weight_ticket.go
```diff
@@ -22,11 +27,49 @@ func (h UpdateProgearWeightTicketHandler) Handle(params progearops.UpdateProGear
...
+				return progearops.NewUpdateProGearWeightTicketForbidden().WithPayload(errPayload), apperror.NewSessionError("Request should come from the office app.")
+			}
+
 			progearWeightTicket.ID = uuid.FromStringOrNil(params.ProGearWeightTicketID.String())
 
-			updatedProgearWeightTicket, _ := h.progearUpdater.UpdateProgearWeightTicket(appCtx, *progearWeightTicket, params.IfMatch)
+			handleError := func(err error) (middleware.Responder, error) {
+				appCtx.Logger().Error("ghcapi.UpdateWeightTicketHandler", zap.Error(err))
+
+				switch e := err.(type) {
```
pearl-truss, line 44: **Praise:** I really like how consistent this is with the weight tickets handler

pearl-truss (approved):

pkg/handlers/ghcapi/progear_weight_ticket.go
```diff
@@ -22,11 +27,49 @@ func (h UpdateProgearWeightTicketHandler) Handle(params progearops.UpdateProGear
 		func(appCtx appcontext.AppContext) (middleware.Responder, error) {
 			payload := params.UpdateProGearWeightTicket
 
+			errInstance := fmt.Sprintf("Instance: %s", h.GetTraceIDFromRequest(params.HTTPRequest))
+			errPayload := &ghcmessages.Error{Message: &errInstance}
```
felipe-lee, line 31: **suggestion (non-blocking)**: After we'd last talked about this when we did it in past PRs, I'd mentioned I wasn't sure we needed to do this, but needed to test it. I forgot to do it sooner, but I did this week and confirmed I was understanding our middleware correctly. It should be inserting the trace ID as a header on every response, so we don't need to be putting it into the error messages. Tbh though, maybe this should be something we talk about in the BE/FE check in meeting, to see how everyone feels. I personally think it's better to do less work in the BE and let the FE decide if/when we want to show those trace IDs in our messages to users. Since we don't really have a standard around this, I'm fine with this staying or being removed. Eventually we'll want to be consistent, once we have a decision

pkg/handlers/ghcapi/progear_weight_ticket.go
```diff
@@ -22,11 +27,49 @@ func (h UpdateProgearWeightTicketHandler) Handle(params progearops.UpdateProGear
...
+							h.GetTraceIDFromRequest(params.HTTPRequest),
+							e.ValidationErrors,
+						),
+					), err
+				case apperror.PreconditionFailedError:
+					msg := fmt.Sprintf("%v | Instance: %v", handlers.FmtString(err.Error()), h.GetTraceIDFromRequest(params.HTTPRequest))
+					return progearops.NewUpdateProGearWeightTicketPreconditionFailed().WithPayload(
+						&ghcmessages.Error{Message: &msg},
+					), err
+				case apperror.QueryError:
```
felipe-lee, line 61: **suggestion (non-blocking)**: We seem to only do this for some of the ghc handlers, but sometimes we have that `if e.Unwrap` line to give us better logging of query errors: 

https://github.com/transcom/mymove/blob/0880ff52227435a89158a0ae0e7aec1740c761bf/pkg/handlers/ghcapi/weight_ticket.go#L47-L56

So, my vote would be to add it, but it isn't a blocker for this.

felipe-lee (approved):
LGTM! I left two comment of things that could change, but aren't blockers, mainly suggestions, but not ones we have standards on afaik.

pkg/handlers/ghcapi/progear_weight_ticket.go
NamibiaTorres: Hmm, would you be alright with me leaving a comment here so that when that discussion happens then we can decide what the best course of action is?

felipe-lee: Sorry, I totally missed this comment for some reason. I think the comment would have been fine, but it's also ok without it since we have lots of handlers we'd need to update anyway if we decided to remove the trace ID from error handling.

PR #9826, opened 2023-01-03
MB-14803 Uploader with extended params

shimonacarvalho:
## Summary

This is a change to the factory operation - you can now pass in extended params. This will allow us to package up and deliver the params needed for advanced functionality like MakeUpload has where it has a mode where you call a function to actually upload the file, vs just creating the record in the database. 

I called it a "spike" but this is complete enough to check in, and I'd like to merge it in, so please review with that in mind.

We will have to use similar functionality in UserUpload and PrimeUpload perhaps. 

Please check the server tests and take a look at how things are set up.


YanZ777 (approved):
The changes look good to me, but you might want to get someone else to review it too.

StuTruss (approved):

pkg/factory/upload_factory_test.go
```diff
@@ -0,0 +1,118 @@
...
+				},
+			},
+		}, nil)
+		suite.Contains(upload.Filename, uploadFile)
+		suite.Equal(encodeMD5("b1e74a6bc8e52bdf45075927168c4bb0"), upload.Checksum)
+		suite.Equal(int64(37986), upload.Bytes)
+		suite.Equal("image/jpeg", upload.ContentType)
+		suite.Equal(models.UploadTypeUSER, upload.UploadType)
+	})
+}
```
felipe-lee, line 113: **suggestion (non-blocking)**: I know you mentioned this was a bit of a spike and this is blocking other work, so I'd be fine if this is done in a follow-up pr, but I think it would be good to also test some of the failure cases, like a missing appctx.

pkg/factory/shared.go
```diff
@@ -445,3 +453,23 @@ func setBoolPtr(customBoolPtr *bool, defaultBool bool) *bool {
...
+// FixtureOpen opens a file from the testdata dir
+func FixtureOpen(name string) afero.File {
+	fixtureDir := "testdata"
+	cwd, err := os.Getwd()
+	if err != nil {
+		log.Panic(fmt.Errorf("failed to get current directory: %s", err))
+	}
+
+	// if this is called from inside another package remove so we're left with the parent dir
+	cwd = strings.Split(cwd, "pkg")[0]
```
felipe-lee, line 466: **suggestion (non-blocking)**: For some reason this took me a sec to understand 😅 I wonder if it would be good to add an example to the comment above it like "e.g. x -> y"

felipe-lee (approved):
LGTM! I left a few comments and questions, but I don't think any are really blocking.

mlm55 (approved):
lgtm!

pkg/factory/shared.go
shimonacarvalho: Yup done.

pkg/factory/upload_factory.go
```diff
@@ -0,0 +1,114 @@
...
+			file = cUploadParams.File
+		} else {
+			file = FixtureOpen("test.pdf")
+		}
+
+		// Create file upload
+		if cUploadParams.AppContext == nil {
+			log.Panic("If Uploader is provided, AppContext must also be provided.")
+		}
+		upload, verrs, err := cUploadParams.Uploader.CreateUpload(cUploadParams.AppContext, uploader.File{File: file}, uploader.AllowedTypesServiceMember)
```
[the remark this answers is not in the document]
shimonacarvalho: I think if it becomes a thing we could easily pass it in, that's the nice thing about the extended params. But I don't want to solve for future problems in this phase of replacement, as long as we have a good mechanism for solving them later.

pkg/factory/upload_factory_test.go
shimonacarvalho: I was about to and realized I can't test anything that will trigger a logPanic, because that will panic, not return an error. (To my knowledge anyway). 

For a testing utility I think this is okay, it's how testdatagen worked as well. Devs just can't make a passing test if they didn't include the appcontext.

felipe-lee: Testify does have a `Panics` assertion:
https://pkg.go.dev/github.com/stretchr/testify/assert?utm_source=godoc#Panics

But I'm not sure how that is affected by the logging.

I agree some things don't need to be tested, at least if it isn't a core part of our testing utils, which this isn't. So I'm fine with skipping it in this case, just mentioned the `Panics` assertion since I was curious and looked into it.

shimonacarvalho: Oh good to know, thanks! There have been moments where I wish I had a way to test, I'll give it a shot.

felipe-lee (approved):
Thanks for making those updates!

pkg/factory/upload_factory_test.go
shimonacarvalho: Done.

PR #9843, opened 2023-01-05
[ MB-11857 ] Removing move management line item for ppm only moves

mmayo-truss:
## Summary

This PR makes it so that the move management line item is not visible for ppm only moves. 

It also allows for the TOO to approve service items on moves when a shipment is created by the prime and sent to the TOO for approval. (For example: A move is PPM only but then the Prime creates an HHG shipment and sends it to the TOO for approval. This PR allows for the TOO to add the move management service item to the move, since the addition of the HHG shipment qualifies the Prime for payment for move management.)

## Setup to Run Your Code

<details>

##### Terminal 1
Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps
There are a couple of scenarios to account for in this PR:

#### PPM only scenario that is routed to the TOO 

1. Access the office app as a TOO. It is best to use the default KKFA GBLOC.
2. Select a PPM only move. Here are a couple of good move locators: PPMSUB and PPM001
3. You should see the PPM move ready to be approved (you might have to fix orders).
4. Verify that the Move Management checkbox is not visible under the Add service items section. Counseling should be (assuming that the move you selected did not have service counseling applied, If you selected either of the locators given that should be the case).
5. Approve the counseling service item and the PPM shipment.
6. You will be redirected to the Move Task Order tab.
7. Select the Move details tab.
8. Verify that Counseling is listed under the service items.
9. Sign out. (remember the locator you used)
10. Sign in to the Prime simulator in the same GBLOC.
11. Select the move you just approved.
12. Add an HHG shipment to the move.
13. Sign out.
14. Sign back in as a TOO and select that same move.
15. You should see the HHG shipment to be approved.
16. Verify the Move Management checkbox is now visible.
17. Approve the MM service item and the HHG shipment.
18. Go back to the Move details tab and verify that the Move management service item is listed under service items.

#### HHG only scenario

In a similar manner to the details above. 

1. Access an HHG scenario as a TOO (CSLCMP is a locator I found)
2. You should see Move management (CSLCMP has already received counseling, so that won't show up as a checkbox)
3. Approve the shipment and MM service item
4. Log in to the Prime simulator and create an HHG shipment for this move
5. Log back in as a TOO and access this same move
6. you should see the new HHG shipment available to be approved
7. You won't see any service items available, since both counseling and move management have been applied to this move already

### Frontend

### Backend

## Screenshots
A PPM only move that does not have service counseling will go to the TOO for approval. The Move Management will not be available.
image

The following is a scenario of the Prime creating an HHG move and requesting approval from the TOO after a single PPM shipment was approved
image

Now the move management service item is associated with the move: 
image


pkg/models/move.go
```diff
@@ -461,3 +461,18 @@ func (m Move) IsCanceled() *bool {
 	}
 	return swag.Bool(false)
 }
+
+// IsPPMOnly returns true of the only type of shipment associate with the move is "PPM", false otherwise
+func (m Move) IsPPMOnly() bool {
```
michaelvaldes-truss, line 466: **question:** Is this the sort of logic we would want to put directly on the model? I'm torn on whether this should live here or in a service

michaelvaldes-truss (approved):
lgtm, works in testing and changes look good! I left one recommendation that's really more of a question.

pkg/models/move.go
kctruss: I think that would be a good idea in the long term, for sure! Short term, since I'm the babysitter, I'm gonna merge this for Mayo and we can chat about it on Tuesday. I'll put a reminder in slack.

mmayo-truss: After talking about it in the huddle with @michaelvaldes-truss and @kctruss , I think that the IsCancelled method above shows a similar pattern. And we can leverage the IsPPMOnly function in the future to be included in the payload in a similar fashion, simplifying our front end code.

PR #9856, opened 2023-01-06
[ on-call ] Updating logging on add user to context middleware to provide more information

mmayo-truss:
## Summary

Adding more information to a logging statement for future debugging purposes.

### DIFF
--- a/pkg/handlers/authentication/add_user_to_context.go
+++ b/pkg/handlers/authentication/add_user_to_context.go
@@ -3,6 +3,8 @@ package authentication
 import (
 	"net/http"
 
+	"go.uber.org/zap"
+
 	"github.com/transcom/mymove/pkg/appcontext"
 	"github.com/transcom/mymove/pkg/audit"
 	"github.com/transcom/mymove/pkg/services"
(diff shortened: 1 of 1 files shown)


kctruss (approved):
👍

pkg/handlers/authentication/add_user_to_context.go
```diff
@@ -18,7 +20,7 @@ func AddAuditUserToRequestContextMiddleware(appCtx appcontext.AppContext) func(n
 			if newAppCtx.Session() != nil {
 				auditUser, err := user.NewUserFetcher(query.NewQueryBuilder()).FetchUser(newAppCtx, []services.QueryFilter{query.NewQueryFilter("id", "=", newAppCtx.Session().UserID)})
 				if err != nil {
-					newAppCtx.Logger().Info("No user attached to the session")
+					newAppCtx.Logger().Error("Error encountered when fetching user with session UserID.", zap.Error(err))
```
duncan-truss, line 23: I think we can see the SessionId value in the logs already but could you also add the UserID value here either to the message or as `zap.String("userId", newAppCtx.Session().UserID)`.

duncan-truss (approved):

pkg/handlers/authentication/add_user_to_context.go
mmayo-truss: Sounds good. I'll add that.

afoon (approved):
LGTM! Error showed as expected with the userId

PR #9865, opened 2023-01-09
[MB-14546] Service counselor defaults to correct queue based on GBLOC

reggieriser:
## Summary

This PR adjusts the queue routing for the services counselor as follows:
- if the service counselor is in a PPM closeout GBLOC (`TVCB`, `NAVY`, or `USCG`), then they route by default to the `PPM Closeout` tab of the queue
- otherwise, they route to the `Counseling` tab of the queue.

Also, a few additional notes:
- the active tab should get highlighted at all times now; previously you would get routed to the `Counseling` queue but the tab would not visually change.
- I removed a route (`/counseling/queue`) that no longer seemed necessary.  Let me know if you think this will cause any issues.
- after talking with Patrick, there could be some route refactoring we could do to nest the queue routes further down into the `ServicesCounselingQueue` instead of at the `index.jsx` level, but we felt that may be too much for this ticket (plus Patrick had already done some work toward that for the react router upgrade that's still in process).

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

Build the office app locally and log in as a services counselor.  Create one service counselor in the `TVCB` GBLOC, and another in a non-closeout GBLOC like `KKFA`.  Ensure that each gets routed as described above.

I tried to cover the routes with client tests, so ensure that they pass.

### Frontend


mandaem:
_"I removed a route (/counseling/queue) that no longer seemed necessary. Let me know if you think this will cause any issues_"  I brought that route up to Ryan and I think he noted something along the lines that all the queues had 2 paths, the something/else and  the /) but that he didn't know if it was necessary. After that, out of an abundance of caution, I had personally kept it in my PR from earlier. But IIRC Felipe created it so maybe he'd have more context?

mandaem (approved):
I wrote a note about route history. otherwise LGTM/no issues when I test

src/pages/Office/ServicesCounselingQueue/ServicesCounselingQueue.test.jsx
```diff
@@ -271,20 +280,54 @@ describe('ServicesCounselingQueue', () => {
...
+      // fields: user, user description, path to try, is counselor view
+      [serviceCounselorUser, 'counselor', servicesCounselingRoutes.DEFAULT_QUEUE_PATH, true],
+      [serviceCounselorUser, 'counselor', servicesCounselingRoutes.QUEUE_COUNSELING_PATH, true],
+      [serviceCounselorUser, 'counselor', servicesCounselingRoutes.QUEUE_CLOSEOUT_PATH, false],
+      [serviceCounselorUserForCloseout, 'closeout', servicesCounselingRoutes.DEFAULT_QUEUE_PATH, false],
+      [serviceCounselorUserForCloseout, 'closeout', servicesCounselingRoutes.QUEUE_COUNSELING_PATH, true],
+      [serviceCounselorUserForCloseout, 'closeout', servicesCounselingRoutes.QUEUE_CLOSEOUT_PATH, false],
+    ];
+
+    testUsersAndPaths.forEach((testCase) => {
```
YanZ777, line 295: **suggestion:** You can also use [Jest's way of doing table driven tests](https://jestjs.io/docs/api#testeachtablename-fn-timeout) here.

reggieriser:
@duncan-truss had noticed that the GBLOC link at the top right of the screen was not working (or at least not rendering the queue) in some scenarios.  I didn't even realize that was a link! 😞  It turns out that my `useEffect` was not running in those scenarios, so I changed it to run on all renders.  It's pretty lightweight so hopefully not an issue.

reggieriser:
> But IIRC Felipe created it so maybe he'd have more context?

Looking at the history, it turns out that @duncan-truss originally added the route which precedes the queue tabs.  I checked with him and he felt the route was OK to remove.

PDickeyTruss (approved):
All works as expected! Really nice work getting this flushed out while avoiding major routing refactoring. 🚢 

I know we discussed and it is mentioned in this PR but I feel it is worth stating again just to be extra safe. Using a `useEffect` hook to do default page routing is not something we should typically be doing.  Instead, it would be preferred to use the native react-routing functionality to do this (Route matching and re-routing). However, it was clear looking at this work with Reggie that the amount of routing specific refactoring needed to allow the typical react-routing functionality to work for this ticket far exceeds the scope of this ticket.  I am in support of using the `useEffect` hook to reroute in the interim knowing that future react-router refactor work is on the way that will clean up the routing and allow us to get the same default routing functionality without needing any custom hooks/functionality.

dinabeigelman (approved):
Looks good to me!

reggieriser:
@Mergifyio refresh

PR #9866, opened 2023-01-09
MB-11915 customer select closeout office

PDickeyTruss:
## Summary

This PR adds functionality to allow Army and Air Force users to search for, select, and save a closeout office to a move in the internal/customer app. This functionality is included when both creating and editing a PPM shipment.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Validation/Testing
The most straightforward way to test this PR is to create two new users in the internal/customer app, one of branch Army or Air-Force, the other of any other branch.  Continue through the profile and orders setup until you are able to add and update PPM shipments.

Army/Air-Force users should be able to:
- [ ] View the Closeout Office section in the PPM shipment form
- [ ] Search for a Closeout Office
- [ ] Select a Closeout Office
- [ ] Save Closeout Office by submitting the form when all required fields are populated
- [ ] Closeout Office is required to allow submission of the PPM shipment form
- [ ] Update the Closeout Office of a previously created PPM shipment that can still be edited
- [ ] Update Closeout Office on a shipment applies to all shipments for the current move

All other branch users should: 
- [ ] Not see the Closeout Office sections/select
- [ ] Have validation and submission behavior unchanged

#### Army/Air-Force user can search for, select, and save a new closeout office:
https://user-images.githubusercontent.com/62263329/212374192-d4c8230b-ad2e-4c47-afff-4012104febec.mov

#### Navy user does not see closeout office section, validation and submission work as before:
https://user-images.githubusercontent.com/62263329/212374229-ea0d619f-4819-402b-8b45-5296c00d2e03.mov

### DIFF
--- a/cypress/support/ppmCustomerShared.js
+++ b/cypress/support/ppmCustomerShared.js
@@ -218,15 +218,15 @@ export function submitsDateAndLocation() {
 
   cy.get('input[name="destinationPostalCode"]').clear().type('76127');
   cy.get('input[name="expectedDepartureDate"]').clear().type('01 Feb 2022').blur();
-
+  cy.selectDutyLocation('Fort Bragg');
   navigateFromDateAndLocationPageToEstimatedWeightsPage();
 }
 
-export function navigateFromDateAndLocationPageToEstimatedWeightsPage(actionToWaitOn) {
+export function navigateFromDateAndLocationPageToEstimatedWeightsPage(actionsToWaitOn) {
   cy.get('button').contains('Save & Continue').should('be.enabled').click();
 
-  if (actionToWaitOn) {
-    cy.wait(actionToWaitOn);
+  if (actionsToWaitOn && actionsToWaitOn.length > 0) {
+    actionsToWaitOn.forEach((action) => cy.wait(action));
   }
 
   cy.location().should((loc) => {
--- a/cypress/support/ppmCustomerShared.js
+++ b/cypress/support/ppmCustomerShared.js
@@ -228,9 +228,7 @@ export function submitsDateAndLocation() {
 export function navigateFromDateAndLocationPageToEstimatedWeightsPage(actionsToWaitOn) {
   cy.get('button').contains('Save & Continue').should('be.enabled').click();
 
-  if (actionsToWaitOn && actionsToWaitOn.length > 0) {
-    actionsToWaitOn.forEach((action) => cy.wait(action));
-  }
+  if (actionsToWaitOn) cy.wait(actionsToWaitOn);
 
   cy.location().should((loc) => {
     expect(loc.pathname).to.match(/^\/moves\/[^/]+\/shipments\/[^/]+\/estimated-weight/);
--- a/cypress/support/ppmCustomerShared.js
+++ b/cypress/support/ppmCustomerShared.js
@@ -218,7 +218,10 @@ export function submitsDateAndLocation() {
 
   cy.get('input[name="destinationPostalCode"]').clear().type('76127');
   cy.get('input[name="expectedDepartureDate"]').clear().type('01 Feb 2022').blur();
+
+  // Select closeout office
   cy.selectDutyLocation('Fort Bragg');
+
   navigateFromDateAndLocationPageToEstimatedWeightsPage();
 }
 
--- a/pkg/testdatagen/scenario/shared.go
+++ b/pkg/testdatagen/scenario/shared.go
@@ -455,17 +455,20 @@ func createUnSubmittedMoveWithMinimumPPMShipment(appCtx appcontext.AppContext, u
 }
 
 func createUnSubmittedMoveWithPPMShipmentThroughEstimatedWeights(appCtx appcontext.AppContext, userUploader *uploader.UserUploader) {
+	closeoutOfficeID := uuid.FromStringOrNil("4afd7912-5cb5-4a90-a85d-ec72b436380e")
+
 	/*
 	 * A service member with orders and a PPM shipment updated with an estimated weight value and estimated incentive
 	 */
 	moveInfo := moveCreatorInfo{
-		userID:      testdatagen.ConvertUUIDStringToUUID("4512dc8c-c777-444e-b6dc-7971e398f2dc"),
-		email:       "estimated_weights@ppm.unsubmitted",
-		smID:        testdatagen.ConvertUUIDStringToUUID("81b772ab-86ff-4bda-b0fa-21b14dfe14d5"),
-		firstName:   "EstimatedWeights",
-		lastName:    "PPM",
-		moveID:      testdatagen.ConvertUUIDStringToUUID("e89a7018-be76-449a-b99b-e30a09c485dc"),
-		moveLocator: "PPMEWH",
+		userID:           testdatagen.ConvertUUIDStringToUUID("4512dc8c-c777-444e-b6dc-7971e398f2dc"),
+		email:            "estimated_weights@ppm.unsubmitted",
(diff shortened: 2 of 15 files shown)


PDickeyTruss:
PR is functionally ready to go so it should be all set for a design review. I still have a few tests in the works that will be coming soon.

src/components/Customer/PPM/Booking/DateAndLocationForm/DateAndLocationForm.test.jsx
```diff
@@ -137,6 +148,42 @@ describe('DateAndLocationForm component', () => {
...
+      const navyServiceMember = {
+        ...defaultProps.serviceMember,
+        affiliation: SERVICE_MEMBER_AGENCIES.NAVY,
+      };
+
+      render(<DateAndLocationForm {...defaultProps} serviceMember={navyServiceMember} />);
+      expect(screen.queryByText('Closeout Office')).not.toBeInTheDocument();
+      expect(screen.queryByLabelText('Which closeout office should review your PPM?')).not.toBeInTheDocument();
+      expect(screen.queryByText('Start typing a closeout office...')).not.toBeInTheDocument();
+    });
```
reggieriser, line 186: **nitpick**: Might be overkill, but wondering if we could set the above tests up as table-based tests to reduce some duplication and make it easier to test other affiliations if desired.

dinabeigelman (approved):
Everything looks good! We updated the label to say "Which closeout office should review your PPM?" so that it made more sense.

reggieriser (approved):
This looks really good to me.🥇  Had just one small optional nitpick.  Tested everything out locally and saw the closeout office drop-down in the right scenarios and was able to set it, edit it, and also watch closeout office changes show up for all PPMs in the move.  Approving, but might be worth having someone who's strong on the front-end and/or the customer app to review as well.

pkg/testdatagen/scenario/shared.go
```diff
@@ -151,9 +154,6 @@ func createGenericPPMRelatedMove(appCtx appcontext.AppContext, moveInfo moveCrea
 			CloseoutOfficeID: moveInfo.closeoutOfficeID,
 		},
 	}
-	println("##############################################################################################################################")
-	println(moveInfo.closeoutOfficeID)
-	println("##############################################################################################################################")
```
PDickeyTruss, line 156: @dymonmoore  I heard you mention you may have a PR on the way to remove this same dev logging. Just wanted to bring to your attention that I also removed it in this PR, so we may end up with some conflicts depending who merges first 😁

cypress/support/ppmCustomerShared.js
```diff
@@ -219,14 +219,17 @@ export function submitsDateAndLocation() {
...
 }
 
-export function navigateFromDateAndLocationPageToEstimatedWeightsPage(actionToWaitOn) {
+export function navigateFromDateAndLocationPageToEstimatedWeightsPage(actionsToWaitOn) {
   cy.get('button').contains('Save & Continue').should('be.enabled').click();
 
-  if (actionToWaitOn) {
-    cy.wait(actionToWaitOn);
+  if (actionsToWaitOn && actionsToWaitOn.length > 0) {
+    actionsToWaitOn.forEach((action) => cy.wait(action));
```
duncan-truss, line 232: You don't have to change this but it does look like cy.wait can take an array of aliases https://docs.cypress.io/api/commands/wait#Aliases

src/services/internalApi.js
```diff
@@ -223,11 +223,12 @@ export async function getMove(moveId) {
   );
 }
 
-export async function patchMove(move) {
+export async function patchMove(moveId, move, ifMatchETag) {
```
duncan-truss, line 226: Thanks for adding this, I definitely missed adding the If-Match eTag in 🎉

src/components/Customer/PPM/Booking/DateAndLocationForm/DateAndLocationForm.jsx
```diff
@@ -101,8 +104,17 @@ const DateAndLocationForm = ({
...
+    serviceMember.affiliation === SERVICE_MEMBER_AGENCIES.AIR_FORCE;
+  if (showCloseoutOffice) {
+    validationShape.closeoutOffice = Yup.object().required('Required');
+  } else {
+    delete validationShape.closeoutOffice;
+  }
+
   return (
-    <Formik initialValues={initialValues} validationSchema={validationSchema} onSubmit={onSubmit}>
+    <Formik initialValues={initialValues} validationSchema={Yup.object().shape(validationShape)} onSubmit={onSubmit}>
```
duncan-truss, line 117: Huh not sure I knew the shape was immutable, so I'm guessing it couldn't be modified as is?

src/services/internalApi.js
PDickeyTruss: No problem ☺️

cypress/support/ppmCustomerShared.js
PDickeyTruss: Awesome! Updated to pass the array of aliases to wait, great spot! 🙇

src/components/Customer/PPM/Booking/DateAndLocationForm/DateAndLocationForm.jsx
PDickeyTruss: Hmm great question... for some reason I thought it was immutable but now that I read your comment I realize I am not super confident in that. Playing with this now to confirm 😁

src/components/Customer/PPM/Booking/DateAndLocationForm/DateAndLocationForm.test.jsx
PDickeyTruss: Great question! I for sure see what you are getting at. I think that table-based could help reduce duplication in the first two tests here since they have the same assertions (just different service members).  The last test has almost inverse assertions (verifies CO field is not present) which I think makes it tricky to include in a table-based type test with the first two. I lean towards your suggestion that this may be overkill for these tests but think it is a very valid suggestion and was worth considering. Will probably leave as is for now but I could be easily swayed if you feel strongly. 😁

src/components/Customer/PPM/Booking/DateAndLocationForm/DateAndLocationForm.jsx
PDickeyTruss: Update after playing with this: I see it as documented or defined as immutable anywhere but it seems to behave as such. When i try to make updates to the `Yup.object().shape(...) ` after initially defined it  just quietly ignores any updates. However, changing the object it is defined with (As in the PR) seems to do the trick? Admittedly feels kinda strange to me that it would be immutable since it can just be defined with an object that changes based on props... 🤷

src/components/Customer/PPM/Booking/DateAndLocationForm/DateAndLocationForm.test.jsx
reggieriser: No problem -- it's really just a style choice and there's not a right or wrong answer.  I think it was on my mind because I [did something similar](https://github.com/transcom/mymove/blob/e4c14eb37c9f5f7f1f3f9d3e7e5056b7160b2e12/src/pages/Office/ServicesCounselingQueue/ServicesCounselingQueue.test.jsx#L284-L324) (including having inverse assertions) in a recent PR.

BTW, I think I should have called it "table-driven" tests rather than "table-based" tests. 😞

PR #9884, opened 2023-01-12
[MB-11916] services counselor can add and edit closeout office

mboren:
## Summary

Added the ability for Services Counselors to update the closeout office when adding or editing a PPM shipment for Army or Air Force moves.
Closeout office is a move-level field, so if it is updated on one PPM shipment for a move, it will change the closeout office for every one of that move's PPM shipments.

Updating the closeeout office and updating the shipment are separate API requests, so I had to change the form to make two requests on submission. 

The new field was added to the `<ShipmentForm>` component which is used in the following places:
- SC creates a shipment of any type
- SC edits a shipment of any type
- TOO creates/edits a shipment of any type

When `<ShipmentForm>` is used for a PPM, it is a two page form
On the first page, user can update destination/pickup and closeout office, and on the second page, they update advance information.

This means there is a lot of conditional logic in `ShipmentForm` (is it a PPM? is the user a SC or TOO? are we on the advance page? etc).

In particular, when the first page of the form is submitted:
- if the move is not a PPM
  - Save shipment and redirect to move details view
- if the move is a PPM
  - Save the shipment
  - Save the closeout office
  - redirect to the Advance page (which will render the same component)

This is an easy place for bugs to hide!

If there's an error with any of the HTTP requests, I wanted an error to be displayed on the page, and the user should have the option to submit the form again. In prior versions of the form, we were never telling Formik that the form was done submitting, we were just navigating to a new page on success. So if there was a problem, you'd get stuck on your current page with the submit button disabled, because formik would think the submit was still running. I think I've fixed these cases with `setSubmitting(false)`.

It was hard to figure out how to wire up the HTTP requests to happen sequentially and also let the closeout office request be skipped when we don't want it.
I decided to shipment information and closeout office information to the update shipment mutation, and have it decide whether or not to update the closeout office based on the presence of the closeout office field passed into it.
This involved some weird wrapping and unwrapping of things, and took a lot of fiddling to get it work with places that expect promises and stuff.
If you can think of a classier way to do this, I'd love to hear about it!

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Login as a SC
2. Find a Navy, Marines, or Coast Guard move
3. Add a PPM shipment. Verify that you *do not* see the closeout office field
4. Edit a PPM shipment. Verify that you *do not* see the closeout office field
5. Find an Army or Air Force move
6. Add a PPM shipment. Verify that you *do* see the closeout office field
7. Edit a PPM shipment for the same move (can be the same shipment). Verify that you *do* see the closeout office field. The closeout office field should already be filled in with the closeout office you chose when adding the previous shipment. Change it to something different and save your changes.
8. Click edit on any PPM shipment for that same move and verify that the closeout office was updated again.

1. Log in as TOO
2. Try editing some different kinds of shipments and make sure nothing is broken.
3. Note that when you edit and save a PPM shipment as a TOO you will end up on this page:
image
This bug also exists on staging. It happens because TOOs don't have permission to access the Advance portion of the form. I don't have time to fix this in this PR.

### Frontend

### Backend


src/pages/Office/ServicesCounselingAddShipment/ServicesCounselingAddShipment.jsx
```diff
@@ -30,12 +38,33 @@ const ServicesCounselingAddShipment = ({ match }) => {
...
+      if (result.closeoutOffice) {
+        mutateMoveCloseoutOffice({
+          locator: moveCode,
+          ifMatchETag: move.eTag,
+          body: { closeoutOfficeId: result.closeoutOffice.id },
+        }).then(() => {
+          // TODO do query invalidation
+        });
+      }
+      // TODO i think it's OK if this happens before the closeout office mutation runs
```
PDickeyTruss, line 60: I agree with this comment. Since the shipment got updated in the DB to get to this point we should also make sure we invalidate the query so that the FE knows that it has out of date shipments (and can request the up-to-date ones if needed).

src/pages/Office/ServicesCounselingAddShipment/ServicesCounselingAddShipment.jsx
```diff
@@ -30,12 +38,33 @@ const ServicesCounselingAddShipment = ({ match }) => {
...
-      mtoShipments.push(newMTOShipment);
-      queryCache.setQueryData([MTO_SHIPMENTS, newMTOShipment.moveTaskOrderID, false], mtoShipments);
-      queryCache.invalidateQueries([MTO_SHIPMENTS, newMTOShipment.moveTaskOrderID]);
-      return newMTOShipment;
+  const [mutateMoveCloseoutOffice] = useMutation(updateMoveCloseoutOffice, {
+    onSuccess: () => {
+      queryCache.invalidateQueries([MOVES, moveCode]);
+    },
+    onError: () => {
+      // TODO invalidate some query data?
```
PDickeyTruss, line 46: I think if the move update fails you probably just need to display an error, shouldn't need to invalidate any queries since the move didn't actually get updated.

src/components/LocationSearchBox/LocationSearchBox.module.scss
```diff
@@ -16,5 +16,5 @@
 
 .hint {
   @include u-margin-top(0);
-  @include u-margin-bottom(1);
+  @include u-margin-bottom(1 !important);
```
duncan-truss, line 19: We should generally avoid important here and probably add either a global class to that hint className prop that isn't scoped to styles and target it from your form components with more specificity.  Not the biggest concern though.

The more out of scope change would be adding in a hintClassName prop to the LocationSearchBox itself.

src/components/form/fields/CloseoutOfficeInput.jsx
```diff
@@ -0,0 +1,51 @@
+import { useField } from 'formik';
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import './DropdownInput.module.scss';
+import LocationSearchBox from 'components/LocationSearchBox/LocationSearchBox';
+import { SearchTransportationOffices } from 'services/ghcApi';
+
+export const CloseoutOfficeInput = (props) => {
```
duncan-truss, line 9: I didn't think I saw this in Patrick's PR, looks like he made the props change only but thanks for figuring this out.  Definitely something I would have included in the original generalizing the component PR if I had thought of it.  Appreciate the clearer name, although you can always rename the import too if we kept it DutyLocationInput.

src/components/Office/ShipmentForm/ShipmentForm.jsx
```diff
@@ -644,12 +680,19 @@ ShipmentForm.propTypes = {
...
   SACs: AccountingCodesShape,
   userRole: oneOf(officeRoles).isRequired,
   displayDestinationType: bool,
   isAdvancePage: bool,
+  closeoutOffice: TransportationOfficeShape,
+  move: shape({
+    eTag: string,
+    id: string,
+    closeoutOffice: TransportationOfficeShape,
+  }),
```
reggieriser, line 695: **question:** If the `closeoutOffice` is part of the `move`, is there a reason we need to pass `closeoutOffice` in as a prop explicitly here or could it just ride along with the move?

mboren: You're right, I forgot to remove this after adding the move prop

src/pages/Office/ServicesCounselingAddShipment/ServicesCounselingAddShipment.jsx
```diff
@@ -30,12 +38,26 @@ const ServicesCounselingAddShipment = ({ match }) => {
...
+          body: { closeoutOfficeId: result.closeoutOffice.id },
+        }).then(() => {
+          queryCache.invalidateQueries([MOVES, moveCode]);
+        });
+      }
+
+      return result.newShipment;
+    },
+    onError: () => {
+      // TODO
```
mboren, line 60: I will make a follow-on story to handle this error case

src/components/LocationSearchBox/LocationSearchBox.module.scss
mboren: Still trying to find a classier way to do this. As far as I can tell, the margin-top/margin-bottom styles in here never get applied (without !important) because the `Hint` component contains styles that take precedence. There's gotta be a simple way to bump up the specificity here, but I haven't seen it yet.

mboren: I was able to make this work without !important by doubling the class, which seems better than !important, but still not ideal.

rileyorr (approved):
@mboren this looks great! I followed your steps and can confirm I did not see a closeout office when I added and edited a Marine move.

I did see the closeout office when I added and edited a PPM shipment on an Army move. The closeout office also correctly updated when edited the same shipment. 

Visually this looks like it matches the designs and the flow matches the intended experience.

reggieriser (approved):
Reviewed the new changes and they LGTM. 🎉

I did notice one small issue of a failing API call to the `closeout-office` endpoint when I was adding/editing a PPM for a `NAVY` user (see screenshot below).  That user doesn't deal with closeout offices (and nothing about closeout office showed up when adding/editing, which is correct).  But we've got this lingering API call being made that maybe shouldn't be.  I think this is something we can fix later though because it doesn't seem to cause a problem with the core functionality, so going ahead and approving.

Screenshot 2023-01-18 at 4 42 06 PM

PR #9892, opened 2023-01-13
MB-14842 Convert office QAECSR and TXO tests to playwright

ahobson:
## Summary

Convert office/qaecsr and office/txo integration tests to playwright.

## Setup to Run Your Code

<details>

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

##### Terminal 3
```sh
PLAYWRIGHT_MY_URL="http://milmovelocal:3000" \
PLAYWRIGHT_ADMIN_URL="http://adminlocal:3000" \
PLAYWRIGHT_OFFICE_URL="http://officelocal:3000" \
playwright test
```
</details>


playwright/tests/utils/baseTest.js
```diff
@@ -20,8 +21,33 @@ export class BaseTestPage {
...
+    this.testHarness = new TestHarness(request);
+  }
+
+  /**
+   * @param {string} inputData
+   * @param {string} fieldName
+   * @param {string} classSelector
+   * @param {number} nth
+   */
+  async genericSelect(inputData, fieldName, classSelector, nth) {
```
kylehilltruss, line 34: GitHub's UI is saying this is a function that only gets referenced once, we can probably move this into the `selectDutyLocation` function.

playwright/tests/utils/officeTest.js
```diff
@@ -9,10 +9,11 @@ const base = require('@playwright/test');
...
 /**
- * OfficePage
+ * office test fixture for playwright
+ * See https://playwright.dev/docs/test-fixtures
  * @extends BaseTestPage
  */
-class OfficePage extends BaseTestPage {
+export class OfficePage extends BaseTestPage {
   /**
    * Wait for the page to finish loading.
```
kylehilltruss, line 18: One medium-sized thing -- and it can be refactored later, I wouldn't consider it blocking today -- is that I strongly prefer `waitForSomething` methods that are more specific and affirmative, rather than a generic "hey the loading indicator's gone" method like `waitForLoading` below. Examples these can found in the customer tests I wrote (https://github.com/transcom/mymove/blob/main/playwright/tests/utils/customerTest.js#L16-L60). Methods like this won't get caught up by any quick flashes of content that aren't the page we're expecting, but also control for issues like accidentally directing Playwright to an unintended link, which caught me out when I was converting over a couple of the tests.

kylehilltruss (approved):
Looks generally good; I've left some comments that are all minor fixes or non-blocking concepts, which we had intended to clean up later. I know you're working on some issues at the moment, but I'm comfortable merging this into main whenever you think it's ready.

playwright/tests/utils/signIn.js
```diff
@@ -9,96 +14,136 @@ export const ServicesCounselorOfficeUserType = 'Services Counselor office';
...
-       * Sign in as existing customer with devlocal
-       *
-       * @param {string} userId
-       */
-      async existingCustomer(userId) {
-        await signInAsUserWithId(userId);
-      },
-    },
-  };
+export class SignIn {
```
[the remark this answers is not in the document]
ahobson: Ok, that's a good point! Would you like for me to make that change as part of this PR?

playwright/tests/utils/baseTest.js
```diff
@@ -20,8 +21,33 @@ export class BaseTestPage {
...
+    // Click on the first presented option
+    await this.page.locator(classSelector).locator('div[class*="option"]').nth(nth).click();
+  }
+
+  /**
+   * @param {string} dutyLocationName
+   * @param {string} fieldName
+   * @param {number} nth
+   */
+  async selectDutyLocation(dutyLocationName, fieldName, nth = 0) {
```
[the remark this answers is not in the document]
ahobson: I'm pretty sure this was in an existing cypress test, and I'm pretty sure we had reimplemented it in cypress at least 3 times and maybe more. Hopefully we can do better in playwright :)

playwright/tests/utils/officeTest.js
ahobson: Yeah, that's a good point and it would be a nice thing to refactor into the test fixtures so that folks writing tests can clearly assert which page they expect to be on. I felt like that was more than just "converting from cypress" so I left it alone.

PR #9672, opened 2022-11-30
MB-14462: Optionally use ECS Health Check

ahobson:
## Summary

* The health endpoint now returns 500 errors if the health check fails
* Configure db connection idel and max lifetime
* Use a health check command to optionally hit the health endpoint
* If RDS IAM creds are enabled, do not stop generating creds on error

## Setup to Run Your Code

Pretty much all the changes are only relevant in AWS, so local testing will be less than helpful.

Lots of testing in experimental will be required

## Verification Steps for Author

These are to be checked by the author.

- [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
- [ ] Request review from a member of a different team.
- [ ] Have the Jira acceptance criteria been met for this change?

## Verification Steps for Reviewers

These are to be checked by a reviewer.

### Backend

- [ ] Code follows the guidelines for [Logging](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide#logging)

[MB-14462]: https://dp3.atlassian.net/browse/MB-14462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9907, opened 2023-01-17
Enable health server in all dp3 environments

ahobson:
Ensure all DP3 environments run the health server as the health check is also enabled there.


rogeruiz (approved):
lgtm 🌈

This enables the ECS health checks that were introduced in #9672 for the `EXP` environment but not propagated over to the `DEMO` and `LOADTEST` environments.

PR #9912, opened 2023-01-18
[ MB-14462 ] Improve logging when health check fails by writing to server stdout

ahobson:
Switch health checker to log to server stdout to ensure we get logs when the health check fails. Tested in experimental


rogeruiz (approved):
lgtm 🌈

Thanks for this @ahobson this will also help me with #9942 to bring logging to more of the App.

PR #9926, opened 2023-01-20
Add step to CircleCI pre_test to ensure spectral lint errors decrease

ahobson:
## Summary

Run the spectral linter in CI and ensure the number of issues goes down.

## Setup to Run Your Code

<details>

##### Terminal 1

Run the 
```sh
./scripts/ensure-spectral-lint missing-baseline spectral
```

</details>


scripts/ensure-spectral-lint
```diff
@@ -0,0 +1,99 @@
...
+            if "severity" in entry:
+                severity = entry["severity"]
+                severity_name = "unknown"
+                if severity in SEVERITY_NAMES:
+                    severity_name = SEVERITY_NAMES[severity]
+                if severity_name not in total_count:
+                    total_count[severity_name] = 0
+                total_count[severity_name] += 1
+
+    return total_count
```
felipe-lee, line 34: **suggestion (non-blocking)**: This is more of a stylistic change, and to share that `collections` has `Counter`, a `dict` subclass that (imo) simplifies counting a bit. Also de-indented a bit for legibility.

```suggestion


def total_linting_counts(filename):
    """
    parse out the total lint problems from the json
    """
    total_count = {}
    with open(filename) as f:
        findings = json.load(f)

    for entry in findings:
        severity = entry.get("severity")
        severity_name = SEVERITY_NAMES.get(severity, "unknown")

        total_count.update([severity_name])

    return total_count
```

scripts/ensure-spectral-lint
```diff
@@ -0,0 +1,99 @@
...
+
+
+def linting_filename_from_basename(filename):
+    """convert the base filename to a linting result filename"""
+    root, ext = os.path.splitext(os.path.basename(fn))
```
felipe-lee, line 39: **suggestion**: `filename` isn't being used and this is relying on outer scope currently to work.

```suggestion
def linting_filename_from_basename(filename):
    """convert the base filename to a linting result filename"""
    root, ext = os.path.splitext(os.path.basename(filename))
```

felipe-lee (commented):
Thanks for putting this up so quickly! I left a few comments, though really only one I think is a blocker for this to get merged.

scripts/ensure-spectral-lint
felipe-lee: One downside to my change in not checking if `severity` is in `entry` means it's possible we get more `unknown` values, but at least with the ghc findings currently, that didn't seem to be an issue.

scripts/ensure-spectral-lint
```diff
@@ -0,0 +1,99 @@
...
+if len(sys.argv) != 3:
+    print(f"Usage: {sys.argv[0]} baseline_dir new_linting_dir")
+    sys.exit(1)
+
+baseline_dir = sys.argv[1]
+new_dir = sys.argv[2]
+
+os.makedirs(new_dir, exist_ok=True)
+files = []
+for fn in glob.glob("swagger-def/*.yaml"):
```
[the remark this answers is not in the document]
ahobson: Thanks, fixed!

scripts/ensure-spectral-lint
ahobson: Thank you, done!

felipe-lee (approved):
Thanks for making those updates. This lgtm as is, though I do wonder about testing. I was thinking of maybe creating a branch off this one and changing the condition for when the baseline is saved temporarily, then creating branches off that one to test out different conditions. 

So something like:

* Your branch with cache being saved on main only
    * New branch with cache saved on this new branch, then branches off it that test when:
        * A new error is added (should fail)
        * A new warning is added (should fail)
        * No new findings (should pass)
        * New error but different error fixed (should pass)
        * New error but warning fixed (should fail)

There are other iterations, but that feels sufficient to be assured that there likely won't be issues. Do you think that'd be fine? Or could that mess with the cache later when we merge this branch to main?

ahobson:
> Thanks for making those updates. This lgtm as is, though I do wonder about testing. I was thinking of maybe creating a branch off this one and changing the condition for when the baseline is saved temporarily, then creating branches off that one to test out different conditions.

I could change the script to just print if there are problems but not actually fail the build. That way we could inspect the output before turning it into a failure

felipe-lee:
> > Thanks for making those updates. This lgtm as is, though I do wonder about testing. I was thinking of maybe creating a branch off this one and changing the condition for when the baseline is saved temporarily, then creating branches off that one to test out different conditions.
> 
> I could change the script to just print if there are problems but not actually fail the build. That way we could inspect the output before turning it into a failure

Yeah, that sounds good to me!

PR #9942, opened 2023-01-25
[ office-hours, MB-14891, MB-15073 ] Adding better logging when working with certificates for the Prime API

rogeruiz:
## Summary

- [x] Logging the request's SHA256 hash on all requests;
- [ ] ~~Adding the DoD certificate chain to EXP for further testing @jesset-truss~~
- [ ] ~~Adding the update DoD 5.9 CA Package @jesset-truss~~
- [ ] ~~Removing the DoD 5.6 CA Package @jesset-truss~~
- [ ] ~~Adding the DoD ECA Root CA S23C from Identrust @jesset-truss~~
- [x] Adding config to deploy to EXP environment
    - [x] This will be reverted later so this can be merged into the default branch.

@ahobson suggested that we could log the SHA256 that gets generated from the
certificate request to help with troubleshooting our Prime API interactions.
This adds that functionality. My intention is to test this in EXP.

Because of where the logging is being done, this cannot be tested locally in its
current iteration.

### edit: 2023-01-30

Now that this branch isn't being used to further test the Prime API in EXP, I can refocus this work to ensure better logging around certificates. Please take this in mind when reviewing the work. There are no longer any changes related to certificate chains or updating the P7B file that gets added to the application any more. The only changes that should exist in this PR are related to logging of Certificate events in the MyMove application.


pkg/handlers/authentication/client_cert.go
```diff
@@ -49,6 +51,11 @@ func ClientCertMiddleware(appCtx appcontext.AppContext) func(next http.Handler)
 			hash := sha256.Sum256(r.TLS.PeerCertificates[0].Raw)
 			hashString := hex.EncodeToString(hash[:])
 
+			newAppCtx.Logger().Info(
+				"Logging hash from the request",
+				zap.String("requestHash", hashString),
+			)
+
 			clientCert, err := models.FetchClientCert(newAppCtx.DB(), hashString)
 			if err != nil {
```
ahobson, line 60: Actually, what if we logged on each of the conditions here? Then we could log that the client cert was successfully or unsuccessfully authenticated?

rogeruiz: I've taken this suggestion and standardized the two calls to the Logger with a `SHA256_hash_string` to make things easier to find with an `ispresent("SHA256_hash_string")` Log Insight function.

ahobson (dismissed):
🚀

pkg/handlers/authentication/client_cert.go
```diff
@@ -52,10 +62,21 @@ func ClientCertMiddleware(appCtx appcontext.AppContext) func(next http.Handler)
...
 				// This is not a known client certificate at all
-				newAppCtx.Logger().Info("Unknown / unregistered client certificate")
+				newAppCtx.Logger().Error(
```
ahobson, line 65: I don't think we have an agreed upon convention on whether or not this should be logged as an error or not. One point of view is that the app is working correctly if someone tries to log in with the wrong cert, so this isn't an application error. The other POV is this is an http error and should be logged as an error.

I personally don't find this to be an application error and would not log it as such, but this isn't a blocking comment

ahobson (approved):

rogeruiz:
Fixing the server_test failures in here shortly.

rogeruiz:
Sorry for the soft-push. Since this branch was used for experimenting it contains a lot of commits that were later reverted. I'd like to keep the history though. The need for the rebase was because the client test coverage was off by a `0.1` again, so I figured I'd add the latest `main` work below this branch to fix it.

PR #9947, opened 2023-01-26
MB-15062: improve playwright local dev

ahobson:
## Summary

Have playwright default to using servers on port 3000 which means local dev does not need to override environment variables. No need to set e.g. `PLAYWRIGHT_MY_URL` anymore.

Also fix FE bug around date formatting which may improve playwright reliability


rogeruiz (approved):
lgtm 🌈 

Less custom env variables the better. 🥳🥳

kylehilltruss (approved):
yessss thank you, remembering to change the port numbers every time I pull down the latest changes was a pain

PR #9948, opened 2023-01-26
Remove yarn resolutions for babel-loader, ua-parser-js

ahobson:
We don't need these resolutions any more and this is a pre-requisite for upgrading to node18


reggieriser (approved):
LGTM.  Everything compiled fine (I'm using `nodenv` to manage node as opposed to `nix`).  Started up app without any issues as well.

Thanks for doing this PR!

PR #9950, opened 2023-01-26
MB-15042-ppm-document-endpoint

NamibiaTorres:
## Summary

This endpoint adds a new endpoint that fetches an objects containing arrays of weight ticket, progear, and expenses documents.

Changes include:

- Adding a new endpoint called `PPMDocumentsFetcher` (was considering naming it `PPMDocuments`, any thoughts?)
- New definition called ProGearWeightTickets
- New definition called MovingExpenses

## Setup to Run Your Code

<details>

##### Terminal 1

This file updates the yaml so if you can't pull in the autogen files you can run swagger updates

```sh
make swagger_generate
```

</details>


felipe-lee (commented):
Thanks for putting this up so quickly! I left a few comments, but otherwise looks good. I ran into an issue locally trying to run `make server_generate`, which I've tracked down to an issue with the openapi cli tool we use to generate the files in `swagger/`. I don't know if it's related to this PR though (currently running cleanup/build commands), but it feels like it probably isn't. I can update here once I've confirmed.

felipe-lee:
> Thanks for putting this up so quickly! I left a few comments, but otherwise looks good. I ran into an issue locally trying to run `make server_generate`, which I've tracked down to an issue with the openapi cli tool we use to generate the files in `swagger/`. I don't know if it's related to this PR though (currently running cleanup/build commands), but it feels like it probably isn't. I can update here once I've confirmed.

Ok, confirmed the issue was with my local env somehow. So I think other than the suggestions I made, this looks good.

scottmries:
I agree that the endpoint URL and name of the yaml ref are clearer with "fetcher", but otherwise LGTM!

swagger-def/ghc.yaml
```diff
@@ -1228,6 +1228,34 @@ paths:
       summary: Denies a SIT extension
       x-permissions:
         - update.SITExtension
+  /ppm-shipments/{ppmShipmentId}/ppm-documents-fetcher:
```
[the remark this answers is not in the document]
NamibiaTorres: Thanks I was debating about the name (noted in the summary). I'll go ahead and change it.

swagger-def/ghc.yaml
```diff
@@ -1228,6 +1228,39 @@ paths:
       summary: Denies a SIT extension
       x-permissions:
         - update.SITExtension
+  /ppm-shipments/{shipmentID}/ppm-documents:
```
felipe-lee, line 1231: **suggestion (non-blocking)**: I think this should start with `/shipments/` instead of `/ppm-shipments/` so that it isn't confusing which shipment ID the parameter is referring to, but on the other hand, we do call the field that points to the MTO shipment from the PPM shipment `shipmentID`, so it might not be confusing to some. I'll leave it up to you though.

```suggestion
  /shipments/{shipmentID}/ppm-documents:
```

felipe-lee (approved):
Awesome, thanks for making those updates. I left one more comment, but I'm on the fence about it so I'll leave it to you and what you prefer.

swagger-def/ghc.yaml
NamibiaTorres: Omg, I missed this in first fix. Thanks for calling it out!

NamibiaTorres: I ended up making the name change @felipe-lee. I think it makes things much more clear and I am all for being as clear as possible in the codebase.

swagger-def/ghc.yaml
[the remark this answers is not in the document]
felipe-lee: Sorry, I somehow missed that in the summary above, I think I jumped straight into the files to review 🤦🏼

PR #9952, opened 2023-01-26
MB-15044 Rename the file for the GitHub Action Happo

YanZ777:
## Summary

Rename file to align with branch naming scheme.
I was trying to write documentation for this GitHub Action, and the name of the file does not match our current naming schema for how our branches are named. I am renaming the file to reflect that. 

## Testing
I'm not sure how we would test this. I think we might have to merge and see if it works?


rogeruiz (approved):
lgtm 🌈

PR #9964, opened 2023-01-30
Prime integration certificates

chtakahashi:
Following the success of the testing in exp that was removed [here](https://github.com/transcom/mymove/pull/9942/commits/d6238d210adb36967432269a41dcb051af464d8b), trying the same work out in stg.

- p7b file is renamed to `milmove-cert-bundle.p7b` so that it's obvious its custom. Added the in-progress Runbook to the readme for future engineers.


Dockerfile
```diff
@@ -15,7 +15,7 @@ COPY bin/rds-ca-rsa4096-g1.pem /bin/rds-ca-rsa4096-g1.pem
 COPY bin/rds-ca-2019-root.pem /bin/rds-ca-2019-root.pem
 COPY bin/milmove /bin/milmove
 
-COPY config/tls/Certificates_PKCS7_v5.6_DoD.der.p7b /config/tls/Certificates_PKCS7_v5.6_DoD.der.p7b
+COPY config/tls/milmove-cert-bundle.p7b /config/tls/Certificates_PKCS7_v5.6_DoD.der.p7b
```
rogeruiz, line 18:
```suggestion
COPY config/tls/milmove-cert-bundle.p7b /config/tls/milmove-cert-bundle.p7b
```

I don't think we should keep the name the same here for what gets picked up in the application. I know this way of doing things requires more app changes in the env files. But ultimately it is clearer to folks working through the code to see that this is no longer exclusively a DoD certificate bundle.

rogeruiz (commented):
❌ This PR needs changes ...

I'm only commenting on this and not trying to block or approve the work directly.

jesset-truss (approved):
Confirmed ticket to change the file name has been created. This looks good to me.

PR #9966, opened 2023-01-30
MB-15033: convert customer ppm tests to playwright

ahobson:
## Summary

Convert customer ppm tests to playwright

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

##### Terminal 3
```sh
playwright test
```
</details>


playwright/tests/utils/customerTest.js
```diff
@@ -98,6 +98,39 @@ export function useMobileViewport() {
...
+    //
+    base.test.describe(`with ${viewportName} viewport`, async () => {
+      if (isMobile) {
+        useMobileViewport();
+      }
+      await callbackfn({ viewportName, isMobile });
+    });
+  });
+}
+
```
rogeruiz, line 133: This feels a lot clearer to me that we set the viewport string in our environment or the default is `desktop` and `mobile`. I see that the main work for this function has been removed from the rest of the `*.spec.js` files. Seems straightforward enough.

The name of `forEachViewport` also seems appropriate here as `viewport` is something that Playwright uses in their documentation as well. The only thing I can suggest besides `forEachViewport` is something like `forEachDevice` but that's only because we're defaulting to two different devices right now. I think `viewport` is the proper work to use here.

kylehilltruss (approved):
Apologies for the delay in getting this reviewed. Everything looks good, but I'm seeing repeated failures in the Estimated Incentive tests when I'm running them (locally, serially) on my intel-based laptop, which I've DMed you about in Slack. Assuming those are cleared up, I'm happy with merging this in.

PR #9967, opened 2023-01-30
fixes p7b file

chtakahashi:
I believe I forgot to run this step before putting the p7b together:

```
openssl pkcs7 -inform DER -outform PEM -in Certificates_PKCS7_v5.6_DoD.der.p7b -print_certs > certificate_bundle.cer
```


jesset-truss (approved):
LGTM

rogeruiz:
What do y'all think about adding a script for generating locally something like `scripts/generate-custom-p7b-container` written in Bash?

chtakahashi:
@rogeruiz def like the idea!

PR #9971, opened 2023-01-31
Disables Happo for the FileUpload component

kylehilltruss:
## Summary

This pull request introduces a parameter to the configuration for the `FileUpload` component's stories that disables Happo tests for it. 

Stories for this component have been perpetually been triggering false positives in Happo during runs; beyond being annoying to resolve, this can cause issues when trying to use automatic merging tools.


rileyorr (approved):
Makes sense to me to disable Happos for this component to prevent automatic merging issues.

PR #9972, opened 2023-01-31
MB-14805 Add PPM Document fetcher service

felipe-lee:
## Summary

This PR adds a new service object for fetching all of the documents for a `PPMShipment` given an `MTOShipment` ID, excluding deleted documents and deleted uploads. The reason to key off of the `MTOShipment` ID is because the expected caller will only have that and not the `PPMShipment` ID.

I would like to have done this in fewer queries, but I couldn't find a way to retrieve all the different document types and their uploads without hitting the DB even more times. [My first attempt](https://github.com/transcom/mymove/commit/9bec7e8184ee85064e5996ee7f507347c7b93a4f#diff-fac432e1feb91b4504b702771ffce089a09484a026dfafbd5d9b059816d93eda) loaded the PPM shipment with only the document associations, but then had to loop over each document type to load its uploads. My current approach seems a bit better to me, though a bit more verbose. Lmk if you think I should do this differently.

As a side note, I'm testing using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). 

<details>

* [feat: Add service to fetch PPM Shipment documents](https://github.com/transcom/mymove/pull/9972/commits/6ae2d88dbbe601fccc4ecd8b838865223e9d68d7)

</details>

<details>

* [refactor: Add model name to SoftDestroy error messages](https://github.com/transcom/mymove/pull/9972/commits/49dd19ff99ce871fe2b1ddf865eb9ead8211ab19)
    * I was running into some issues with this function when trying to write my tests. I ended 
up not using it because I couldn't for what I needed, but I figured I'd commit this since it made debugging easier.

</details>

## Testing Changes

This is entirely back-end focused, and isn't being called by anything, so the only way to test is by running the tests. If you have a debugger set up, you can step through the service logic as the test runs.

<details>
<summary>💻</summary>

### Server Tests

Run all the server tests with:

```sh
make server_test
```

</details>

### Backend


pkg/models/ppm_shipment.go
```diff
@@ -67,6 +67,14 @@ var AllowedPPMDocumentStatuses = []string{
 	string(PPMDocumentStatusRejected),
 }
 
+// PPMDocuments is a collection of the different PPMShipment documents. This type exists mainly to make it easier to
+// work with the group of documents as a whole when we don't actually retrieve the PPM Shipment itself.
+type PPMDocuments struct {
+	WeightTickets
+	MovingExpenses
+	ProgearExpenses ProgearWeightTickets
```
felipe-lee, line 75: Mainly changed the name here to make it match how it's named on `PPMShipment`

felipe-lee, line 75: Mainly changed the name here to make it match how it’s named on `PPMShipment`

scottmries: praise: Thanks for making this consistent!

scottmries:
LGTM! 
The unit tests are present and clear, and they pass.
I agree that the current approach is preferable -- eager loading an object that can have a variable number of results (a document's uploads) seems clearer than doing so on objects with a constant number (the three document set types) and then looping over their associations. It's easier to understand and black-boxes the more appropriate details.
I appreciate the conventional commits approach, too; it makes the intents of changes clearer and overall this PR more straightforward to review.
Thanks @felipe-lee!

scottmries (approved):

PR #9979, opened 2023-01-31
MB-15109: fight playwright flakes in circleci

ahobson:
## Summary

These changes are almost entirely about CircleCI, so running locally doesn't really provide any value.

Check out the CI runs for the results


kylehilltruss (approved):
This makes sense, thanks for diagnosing the issues with the out-of-order payment requests.

PR #9981, opened 2023-01-31
[MB-14903] backend deletion of ppm closeout line items

mboren:
## Summary
This PR adds internal API endpoints for deleting PPM closeout line items (weight tickets, pro gear weight tickets, and moving expenses).
I chose to add a DELETE to existing routes, rather than make one route to rule them all. This approach is more boilerplatey but I like having each route do just one thing, feels simpler to implement and test.
I had to do a bit of tinkering to make sure that all the relevant stuff respects the soft deletion of each of the line items.

The implementations for deleting moving expenses and pro-gear weight tickets are identical, except for having to use different fetcher methods.
Weight ticket deletion is a little different because it affects the final incentive, so I had to make sure that got recalculated

## Setup to Run Your Code
```
make server_run
```
```
make client_run
```

#### Screen recording of deleting a pro-gear weight ticket

https://user-images.githubusercontent.com/2627844/216469210-a588010f-efe7-4271-8a54-4c1a42de8cb2.mov
#### screen recording of deleting a weight ticket and verifying that the final incentive was updated
https://user-images.githubusercontent.com/2627844/216471794-57bf7453-29ff-4622-b2a6-c3f45ddecf09.mov

#### instructions
- log in as a service member that is ready to upload line items (eg `readyForFinalComplete@ppm.approved`)
- Create line items of each type (if they don't exist already)
- Every time there's a page load, the CSRF token changes, so you'll need to find it using the network tab of your browser's dev tools .
- Use the network tab of your browser's dev tools to snag the CSRF token for the session, the PPM shipment ID, and IDs for any line items you'd like to delete  from the `GET /internal/moves/{moveID}/mto_shipments` request that is made automatically when the ppm shipment review page is loaded. If you like, you can also get all the line item IDs for a particular PPM shipment id with the following SQL query instead of pulling them out of the HTTP response. 
<details>
<summary>Optional SQL query to get line item IDs for a particular PPM shipment</summary>

```sql
SELECT *
FROM (SELECT ppm_shipment_id, 'weight-ticket' AS line_item_type, id AS line_item_id, vehicle_description AS description, deleted_at
      FROM weight_tickets
      UNION
      (SELECT ppm_shipment_id, 'pro-gear-weight-tickets' AS line_item_type, id AS line_item_id, description, deleted_at
       FROM progear_weight_tickets)
      UNION
      (SELECT ppm_shipment_id, 'moving-expenses' AS line_item_type, id AS line_item_id, description, deleted_at
       FROM moving_expenses)) AS t
WHERE ppm_shipment_id = '6d1d9d00-2e5e-4830-a3c1-5c21c951e9c1'
ORDER BY line_item_type;
```

</details>

- use the browser's JS console to make requests to delete line items of each type. The different values for `lineItemType` are: - `"weight-ticket"`
- `"pro-gear-weight-tickets"`
- `"moving-expenses"`
- Note that `weight-ticket` is not plural but the others are!
<details>
<summary>Paste this expression into your browsers console and replace the CSRF token, IDs, and line item type at the bottom with your own</summary>

```js
(function ({csrf, ppmID, lineItemType, lineItemID}) {
if (!(lineItemType === "moving-expenses" || lineItemType === "weight-ticket" || lineItemType === "pro-gear-weight-tickets")) {
  console.error("invalid lineItemType:", lineItemType);
  return;
}
fetch("http://milmovelocal:3000/internal/ppm-shipments/" + ppmID + "/" + lineItemType + "/" + lineItemID, {
  "headers": {
    "accept": "application/json",
    "accept-language": "en-US,en;q=0.9",
    "cache-control": "max-age=0",
    "content-type": "application/json",
    "x-csrf-token": csrf
  },
  "referrer": "",
  "method": "DELETE",
  "mode": "cors",
  "credentials": "include"
}).then((resp) => console.log(resp));
})({
  csrf: "nszpGgOAHvwDNUeRRpwffo4L0ChKCQ+SDaDm4gclcw/X8kzQQwKtkmufnJ9o1vYX+pm46eISmelsUtl3RPFbFA==",
  ppmID:"6d1d9d00-2e5e-4830-a3c1-5c21c951e9c1",
  lineItemType:"weight-tickets",
  lineItemID:"eec6958d-d803-48e4-a63e-c07fcd5c6953"
});
```

</details>
- You should get a 204 response, and when you refresh the page, the item you deleted should no longer appear.
- If you deleted a weight ticket, the final incentive should be updated. You can check the final incentive on the next page by clicking "save & continue". Note that you will only be able to get to this page if there is at least one weight ticket.

Error cases to try:
- You should get an error if you try to delete an already deleted line item
- Get the the IDs for a line item for one service member, and then log in as another service member and try to delete this line item. You should not be able to.

### Backend


pkg/handlers/internalapi/weight_ticket.go
```diff
@@ -141,3 +145,69 @@ func (h UpdateWeightTicketHandler) Handle(params weightticketops.UpdateWeightTic
...
+			}
+
+			// Make sure the service member is not modifying another service member's PPM
+			ppmID := uuid.FromStringOrNil(params.PpmShipmentID.String())
+			var ppmShipment models.PPMShipment
+			err := appCtx.DB().Scope(utilities.ExcludeDeletedScope()).
+				EagerPreload(
+					"Shipment.MoveTaskOrder.Orders",
+				).
+				Find(&ppmShipment, ppmID)
```
duncan-truss, line 178: Thanks for adding in this authorization check.  Am I reading this right where they could still pass in a weightTicketID associated with a different PPM Shipment as long as the ppmShipmentID in the params is their own?  Seems like we would still end up fetching whatever weight ticket they provided in the service as long as they could get past this check.

pkg/services/weight_ticket/weight_ticket_deleter.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func handleSoftDestroyError(err error) error {
+	if err == nil {
+		return nil
+	}
+	switch err.Error() {
+	case "error updating model":
+		return apperror.NewUnprocessableEntityError("while updating model")
+	case "this model does not have deleted_at field":
```
duncan-truss, line 34: Will these error strings match?  It looks like in SoftDestroy they add in the model name to these messages?  It would definitely be easier if we could just compare types but that's too much work to include here since it's used across the system. https://github.com/transcom/mymove/blob/08fff54987346e5b72aa0dfe1b25ef89016c74ae/pkg/db/utilities/utilities.go#L43

pkg/handlers/internalapi/weight_ticket.go
mboren: you're right, I'll add a check for that

pkg/services/weight_ticket/weight_ticket_deleter.go
mboren: I copied this chunk from somewhere else and didnt look at it very closely tbh. I'll dig into this and make sure things match up.

pkg/services/weight_ticket/weight_ticket_deleter.go
```diff
@@ -0,0 +1,94 @@
...
+		newPPM, err := ppmshipment.FindPPMShipmentAndWeightTickets(appCtx, weightTicket.PPMShipmentID)
+		if err != nil {
+			return err
+		}
+
+		finalIncentive, err := d.PPMEstimator.FinalIncentiveWithDefaultChecks(appCtx, *oldPPM, newPPM)
+		if err != nil {
+			return err
+		}
+		newPPM.FinalIncentive = finalIncentive
```
duncan-truss, line 78: Do you think it's worth actually checking to see if this results in a different final incentive and not doing the update?  I'm thinking of spurious events being added to the history and audit log possibly because we'd be touching the `updated_at` field (although that is usually ignored).  Also it might reduce needing to refetch the PPM shipment if the eTag hasn't changed.

duncan-truss (approved):
Great work I think all the functionality is here.

A few non-essential things I might consider:
- The api linter `lint-yaml-with-spectral swagger/internal.yaml` complained about description fields needing to be 128 characters 🤷‍♂️
- Maybe the handler auth datbase queries could be moved to a service fetcher.

pkg/services/weight_ticket/weight_ticket_deleter.go
mboren: Yeah, upon a closer look, I don't see a practical way to handle these errors, so I think I'll just bubble them up. Some error types coming out of softdestroy would make life a lot easier.

mboren:
i haven't come up with a spot for the queries in the handlers that i'm happy with, so i'm waiting until tomorrow to merge this in case i wake up with an inspiration.

PR #9992, opened 2023-02-01
Improved logging around cert chains

shimonacarvalho:
## Summary

This PR improves our cert chain logging to help with debugging upcoming changes to add a prime CA

* I did have to move a staticcheck mitigation comment - it should be ok but I'm not sure if there are any processes around touching any of that. 
* I did move a log from debug → info level that lists the trust chain. Any ATO concerns around that? It's all public certs, not the contents of the cert, but just the subject asn1 encoded.

This should not affect anything other than logging.

### DIFF
--- a/cmd/milmove/serve.go
+++ b/cmd/milmove/serve.go
@@ -386,17 +386,6 @@ func initializeTLSConfig(appCtx appcontext.AppContext, v *viper.Viper) *tls.Conf
 	if certificates == nil || rootCAs == nil || err != nil {
 		appCtx.Logger().Fatal("Failed to initialize DOD certificates", zap.Error(err))
 	}
-	appCtx.Logger().Debug("Server DOD Key Pair Loaded")
-	// RA Summary: staticcheck - SA1019 - Using a deprecated function, variable, constant or field
-	// RA: Linter is flagging: rootCAs.Subjects is deprecated: if s was returned by SystemCertPool, Subjects will not include the system roots.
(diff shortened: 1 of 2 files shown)


kctruss (approved):
🙌

cmd/milmove/serve.go
```diff
@@ -409,6 +398,17 @@ func initializeTLSConfig(appCtx appcontext.AppContext, v *viper.Viper) *tls.Conf
...
+	// RA: Linter is flagging: rootCAs.Subjects is deprecated: if s was returned by SystemCertPool, Subjects will not include the system roots.
+	// RA: Why code valuable: It allows us to log the root CA subjects that are being trusted.
+	// RA: Mitigation: The deprecation notes this is a problem when reading SystemCertPool, but we do not use this here and are building our own cert pool instead.
+	// RA Developer Status: Mitigated
+	// RA Validator Status: Mitigated
+	// RA Validator: leodis.f.scott.civ@mail.mil
+	// RA Modified Severity: CAT III
+	// nolint:staticcheck
+	subjects := rootCAs.Subjects()
+	appCtx.Logger().Info("Trusted CAs", zap.Any("num", len(subjects)), zap.Any("subjects", subjects))
```
rogeruiz, line 411: I don't see any security concerns making this `.Info()` instead of `.Debug()`. The Trusted CAs are all in the `config/tls` directory in this project and this info can be inferred from those certs.

shimonacarvalho: Both great points. Thank you.

PR #9997, opened 2023-02-02
MB-15121: Enable DTOD_USE_MOCK for local dev, needed for playwright

ahobson:
## Summary

This enables DTOD_USE_MOCK for local dev, which we need to run playwright tests against a local dev environment.

See the [slack thread](https://ustcdp3.slack.com/archives/CP6PTUPQF/p1674761263821299) for discussion

I've been running with this setting enabled for weeks in my local dev environment with no problems.


pkg/route/dtod_zip5_distance_mock.go
```diff
@@ -43,6 +43,25 @@ func (m *mockDTODZip5DistanceInfo) DTODZip5Distance(appCtx appcontext.AppContext
...
 
+	// According to https://www.unitedstateszipcodes.org/80901/ this
+	// zip is a PO Box only zip. From the slack thread
+	// https://ustcdp3.slack.com/archives/C0250Q8469K/p1672933844613089
+	// it seems that DTOD will return a 0 distance for PO Box only
+	// zips
+	if pickupZip5 == "80901" || destinationZip5 == "80901" {
+		// in dtod_zip5_distance.go, this is the error if DTOD returns
+		// a distance less than or equal to 0
+		return -1, fmt.Errorf("invalid distance using pickup %s and destination %s", pickupZip5, destinationZip5)
```
reggieriser, line 54: I believe @duncan-truss was saying that DTOD returns a distance of 0 when you use a PO Box zip code, but we're returning a -1 here (looks like `dtod_zip5_distance.go` returns whatever distance came back in the <= 0 scenario, so we might be returning a -1 in the mock when it returns a 0).  That said, I'm not sure it really matters because the caller shouldn't be using the returned distance anyway since a non-nil error is being returned.

ahobson: Ohh, I didn't realize the existing code returns the distance and an error. Ick.

Yes, you are right, the distance returned from the server is actually 0 in both cases. I guess I'll change the mock just to be consistent

ahobson: I hit DTOD directly:

```
2023-02-02T16:56:54.756Z	FATAL	toolbox/main.go:104
DTOD error	{"distance": 0, "error": "invalid distance using pickup 30307 and destination 11111"}
```

reggieriser (commented):
I ran this and tried to create a PPM with the mock DTOD on (default in this PR).  Using 80901 generated an error as expected (although the UI shows an unhelpful "internal server error" alert -- not something to worry about in this PR).  But I'm not able to test 11111 because we apparently block it on the zip code entry screen:
Screenshot 2023-02-02 at 3 15 43 PM

What's interesting (and sort of an aside) is that the above zip code check is relying on the older `Tariff400ng*` tables and the lat/long zip lookup embedded data to tell whether a zip is good or not.  Those haven't been updated in a while and may not be used outside of this check, but I'd have to dig into it more to know for sure.

So, your code seems fine, but not sure if there's an easy way to test a bad zip with the mock.

reggieriser (approved):

PR #9998, opened 2023-02-02
[MB-14811] Disregard rejected weight tickets in new weight calculations 

pearl-truss:
## Summary

This PR updates the `SumWeights` function that we use to calculate the net weight of a PPM's weight tickets. The change excludes rejected weight tickets from the calculations


pkg/services/ppmshipment/ppm_estimator.go
```diff
@@ -193,14 +193,30 @@ func SumWeightTickets(ppmShipment, newPPMShipment models.PPMShipment) (originalT
...
-				originalTotalWeight += *weightTicket.FullWeight - *weightTicket.EmptyWeight
+				if weightTicket.Status != nil {
+					if *weightTicket.Status == models.PPMDocumentStatusRejected {
+						originalTotalWeight += 0
+					} else {
+						originalTotalWeight += *weightTicket.FullWeight - *weightTicket.EmptyWeight
+					}
+				} else {
+					originalTotalWeight += *weightTicket.FullWeight - *weightTicket.EmptyWeight
+				}
```
ronaktruss, line 204: We may be able to reduce one level of if conditions by combining the two if conditions. The logic would look something like:
1. check if status is not nil AND the status is rejected and then add 0 (this should early terminate if the status is nil to prevent the null pointer error)
2. else add the difference

Something similar can be done for the if condition below as well.

pearl-truss: For 1) I'm not sure if we want early terminate since this block is for the old weight ticket and we still want to go through the subsequent logic for the new weight ticket

ronaktruss: Yeah you're right. "Early terminate" was not the correct word,  I meant to capture that in the case that the status was nil this check would fail on the first half of the condition of it being not nil and we wouldn't need to worry about a null pointer issue.

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -719,6 +719,118 @@ func (suite *PPMShipmentSuite) TestPPMEstimator() {
 			suite.Equal(unit.Cents(70064364), *ppmFinal)
 		})
 
+		suite.Run("Final Incentive - Success with disregarding rejected weight tickets", func() {
+			setupPricerData()
+			moveDate := time.Date(2020, time.March, 15, 0, 0, 0, 0, time.UTC)
+			oldPPMShipment := testdatagen.MakeApprovedPPMShipmentWithActualInfo(suite.DB(), testdatagen.Assertions{
```
felipe-lee, line 725: **suggestion (non-blocking)**: This func by default would create a shipment whose closeout docs haven't been submitted. The one I suggest below would be more accurate data-wise in other regards. I know you're overridding some parts of it anyway, but I figured I'd share just so that the rest of the data is more accurate, even if we don't need it. It would also mean you don't have to create the weight ticket yourself in the assertions.

```suggestion
			oldPPMShipment := testdatagen.MakePPMShipmentThatNeedsPaymentApproval(suite.DB(), testdatagen.Assertions{
```

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -719,6 +719,118 @@ func (suite *PPMShipmentSuite) TestPPMEstimator() {
...
+			weightOverride := unit.Pound(19500)
+			rejected := models.PPMDocumentStatusRejected
+			newPPM.WeightTickets = models.WeightTickets{
+				testdatagen.MakeWeightTicket(suite.DB(), testdatagen.Assertions{
+					WeightTicket: models.WeightTicket{
+						Status:     &rejected,
+						FullWeight: &weightOverride,
+					},
+				}),
+			}
```
felipe-lee, line 747: **nit**: could you update the existing weight ticket instead of creating a new one? I know it doesn't really matter much in terms of the test itself because the final incentive check just uses whatever we pass in, but it could be a bit confusing if someone isn't too familiar with that or weight tickets.

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -719,6 +719,118 @@ func (suite *PPMShipmentSuite) TestPPMEstimator() {
...
+			suite.Nil(ppmFinal)
+		})
+
+		suite.Run("Final Incentive - Success updating finalIncentive with rejected weight tickets", func() {
+			setupPricerData()
+			weight := unit.Pound(19500)
+			newWeight := unit.Pound(21500)
+
+			moveDate := time.Date(2020, time.March, 15, 0, 0, 0, 0, time.UTC)
+			oldPPMShipment := testdatagen.MakeApprovedPPMShipmentWithActualInfo(suite.DB(), testdatagen.Assertions{
```
felipe-lee, line 777: **suggestion (non-blocking)**: same as in the test above, suggestion to use a different testdatagen func

```suggestion
			oldPPMShipment := testdatagen.MakePPMShipmentThatNeedsPaymentApproval(suite.DB(), testdatagen.Assertions{
```

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -719,6 +719,118 @@ func (suite *PPMShipmentSuite) TestPPMEstimator() {
...
+					ActualPickupPostalCode:      models.StringPointer("90210"),
+					ActualDestinationPostalCode: models.StringPointer("30813"),
+					ActualMoveDate:              models.TimePointer(moveDate),
+					FinalIncentive:              models.CentPointer(unit.Cents(500000)),
+					Status:                      models.PPMShipmentStatusWaitingOnCustomer,
+					WeightTickets: models.WeightTickets{
+						testdatagen.MakeWeightTicket(suite.DB(), testdatagen.Assertions{
+							WeightTicket: models.WeightTicket{
+								FullWeight: &weight,
+							},
```
felipe-lee, line 788: **suggestion (non-blocking)**: If you do change to using the other testdatagen func, then this could become a top-level assertion. Additionally, could you pass the empty weight as part of the assertions so that this test doesn't then fail if anyone ever changes the default used in `MakeWeightTicket` (or if we change at some point to allow random data gen)

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -719,6 +719,118 @@ func (suite *PPMShipmentSuite) TestPPMEstimator() {
...
+						FullWeight: &newWeight,
+					},
+				}),
+				testdatagen.MakeWeightTicket(suite.DB(), testdatagen.Assertions{
+					WeightTicket: models.WeightTicket{
+						Status:     &approved,
+						FullWeight: &newWeight,
+					},
+				}),
+			}
```
felipe-lee, line 810: **nit**: same nit as in the previous test about updating the existing weight ticket instead of creating a new one. I guess here we are in fact adding one that wasn't there originally, but maybe that should be created with the original assertions above and then updated here to have a new weight and status.

felipe-lee (commented):
This mostly lgtm! I left a few suggestions/nits, but none are blocking. It's mostly suggestions to make the data match a little more closely to what it would when a real shipment runs through this.

pkg/services/ppmshipment/ppm_estimator_test.go
```diff
@@ -719,6 +719,118 @@ func (suite *PPMShipmentSuite) TestPPMEstimator() {
...
+					ActualDestinationPostalCode: models.StringPointer("30813"),
+					ActualMoveDate:              models.TimePointer(moveDate),
+					Status:                      models.PPMShipmentStatusWaitingOnCustomer,
+					WeightTickets: models.WeightTickets{
+						testdatagen.MakeDefaultWeightTicket(suite.DB()),
+					},
+				},
+			})
+
+			newPPM := oldPPMShipment
```
pearl-truss, line 737: I didn't realize structs are mutable in golang. I see that when `newPPM.WeightTickets[0].Status` is updated so is `oldPPM.WeightTickets[0].Status`

pearl-truss:
@felipe-lee I believe everything is addressed now. And I think the test to ensure the final incentive gets updated is enough but if you feel otherwise I can try to add that in tomorrow

felipe-lee (approved):
Awesome, thanks for making those updates. This lgtm!

Then only suggestion (non-blocking) I have is that I liked that you were setting explicit weights in the test before, since that makes the test not depend on the defaults set for weights in the Make functions. It's also not a huge deal to leave it as is though since if those defaults ever do change, the PR that changes them will flag any tests that fail and they can be updated accordingly.

pearl-truss:
> Awesome, thanks for making those updates. This lgtm!
> 
> Then only suggestion (non-blocking) I have is that I liked that you were setting explicit weights in the test before, since that makes the test not depend on the defaults set for weights in the Make functions. It's also not a huge deal to leave it as is though since if those defaults ever do change, the PR that changes them will flag any tests that fail and they can be updated accordingly.

That sounds worthwhile, I just pushed up changes for this

NamibiaTorres:
LGTM too! It looks like the test coverage decreased, you can do a quick type check and add it to the assertions for one of the tests to satisfy the coverage requirement.

PR #10004, opened 2023-02-02
MB-15126: Extend playwright office tests fixtures vs major beforeEach setup

ahobson:
## Summary

Now the only bits in beforeEach are to create the move specific for the tests (and possibly navigate to the move)

[these docs from playwright about fixtures](https://playwright.dev/docs/test-fixtures#with-fixtures) explains more about the approach used.


kylehilltruss (approved):
Changes look good -- thanks for refactoring the `scPage` class to inherit from `officePage` rather than have two different methods of composition

PR #10014, opened 2023-02-03
[MB-15098] Fix documentation for prime certs migration, troubleshooting migration

chtakahashi:
- Removed an unnecessary `&& /` in a comment to avoid confusion in the future
- The script needed to be modified to include variable declarations


rogeruiz (approved):
lgtm 🌈

PR #10015, opened 2023-02-03
[ spike ] Adding PropagateTags to the EcsParameters

rogeruiz:
## Summary

This is something that has come up on the Platform team as to why these
PropagateTags don't seem to stick when we do a `tf plan` on our infrastructure.
The drift here is caused by the Golang code modifying our infrastructure. This
helps mitigate that drift a bit in concerns with CloudWatch Events ECS
parameters.


chtakahashi (approved):
:seems-legit:

PR #10018, opened 2023-02-03
Add PPM Documents handler and tests.

scottmries:
## Summary

This creates a handler for PPM Documents and connects it to the service object. The endpoint is `/shipments/{shipmentID}/ppm-documents`. Tests check that errors are returned when the SO returns errors, and that the happy path returns a 200 and the documents. This approach is largely based on the existing weight ticket handler.

Note that the YAML definition for Moving Expenses now references `MovingExpenses.yaml` rather than `ProGearWeightTickets`, which I think had been in error.

Note too that the endpoint is listed under PPM in the Swagger UI, though we're querying via the MTO shipment; I assume that listing it here with give it a better chance of being found, since it's not used for non-PPM shipments.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

1. Ensure that the endpoint appears at milmovelocal:3000/swagger-ui/ghc.html or the equivalent, under the PPM heading.
2. Ensure that the payload is as expected.
3. Run server tests.

### Backend


pkg/handlers/ghcapi/ppm_document.go
```diff
@@ -0,0 +1,66 @@
...
+			shipmentID := uuid.FromStringOrNil(params.ShipmentID.String())
+
+			ppmDocuments, err := h.PPMDocumentFetcher.GetPPMDocuments(appCtx, shipmentID)
+
+			if err != nil {
+				appCtx.Logger().Error("ghcapi.GetPPMDocumentsHandler error", zap.Error(err))
+
+				switch e := err.(type) {
+				case apperror.ForbiddenError:
+					return ppmdocumentops.NewGetPPMDocumentsForbidden().WithPayload(errPayload), nil
```
felipe-lee, line 46: **suggestion (non-blocking)**: The service object doesn't return a `ForbiddenError` rn, so this case can get removed.

pkg/handlers/ghcapi/ppm_document_test.go
```diff
@@ -0,0 +1,214 @@
...
+		ppmDocuments := models.PPMDocuments{
+			WeightTickets:   ppmShipment.WeightTickets,
+			MovingExpenses:  ppmShipment.MovingExpenses,
+			ProgearExpenses: ppmShipment.ProgearExpenses,
+		}
+
+		ppmDocumentFetcher.On("GetPPMDocuments",
+			mock.AnythingOfType("*appcontext.appContext"),
+			mock.AnythingOfType("uuid.UUID"),
+		).Return(&ppmDocuments, nil)
```
felipe-lee, line 184: **suggestion (non-blocking)**: Since you set up `setUpMockPPMDocumentFetcher`, you can use that here instead of setting up the mock fetcher from scratch again on line 173 and then setting up the call here.

```suggestion
		ppmDocumentFetcher := setUpMockPPMDocumentFetcher(&ppmDocuments, nil)
```

pkg/handlers/ghcapi/ppm_document_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+		handler := setUpHandler(ppmDocumentFetcher)
+
+		response := handler.Handle(params)
+
+		if suite.IsType(&ppmdocumentops.GetPPMDocumentsOK{}, response) {
+			okResponse := response.(*ppmdocumentops.GetPPMDocumentsOK)
+			returnedPPMDocuments := okResponse.Payload
+
+			fmt.Println("about to validate returnedPPMDocuments")
```
felipe-lee, line 194: **todo**: Can you take out this print? 

As a side note, if you use `goland` I'd be happy to pair with you and show you how to use the `goland` debugger which can be very helpful for this type of thing. Here are the general docs on the `goland` debugger https://www.jetbrains.com/help/go/debugging-code.html, which iirc work for the most part as is. The main caveat is that you'll need to open `goland` from a terminal after you've run `direnv allow` so that it has access to the env vars it needs to connect to things like the test db.

pkg/handlers/ghcapi/ppm_document_test.go
```diff
@@ -0,0 +1,214 @@
...
+			for i, returnedMovingExpense := range returnedPPMDocuments.MovingExpenses {
+				suite.Equal(ppmShipment.MovingExpenses[i].ID.String(), returnedMovingExpense.ID.String())
+			}
+
+			for i, returnedProGearWeightTicket := range returnedPPMDocuments.ProGearWeightTickets {
+				suite.Equal(ppmShipment.ProgearExpenses[i].ID.String(), returnedProGearWeightTicket.ID.String())
+			}
+		}
+	})
+}
```
felipe-lee, line 214: **todo**: Can you add an integration test? Since you based this off the weight ticket tests, there's an example there for the integration version as well. It's pretty similar, just want to make sure the flow works fully without mocking out the service object: https://github.com/transcom/mymove/blob/ce7a60fa3faedaac7134ef70dcdef20c6488b76b/pkg/handlers/ghcapi/weight_ticket_test.go#L177-L243

felipe-lee (commented):
Thanks for getting this up so quickly! It mostly lgtm, I just left a few comments for things to do, and some suggestions as well. 

One thing that is missing before this can be considered done is hooking up the new handler you created into the `ghcapi/api.go` file. As an example, here is a weight ticket handler getting set up: https://github.com/transcom/mymove/blob/f4dbab355d6b3992e6f1f1c7545bc14d085068ae/pkg/handlers/ghcapi/api.go#L410-L415

[We have docs on all the steps for adding an API](https://transcom.github.io/mymove-docs/docs/api/guides/guide-to-creating-an-endpoint#steps-to-creating-a-new-handler), but tbh they could use some TLC. They mention the step of adding a handler to an `api.go` file, but don't go into it too much. Just figured I'd share them anyways since they have other tidbits that could be helpful for future endpoints. I think you've covered everything else you would need for this one though.

pkg/handlers/ghcapi/ppm_document.go
pearl-truss: If that's the case, lines 33-35 canalso be removed

pearl-truss (commented):
Pending the comments left by Felipe, this looks good to me

scottmries:
Thanks for the detailed feedback as always, @felipe-lee and @pearl-truss. Each of those changes should be ready in my second commit.

pkg/handlers/ghcapi/ppm_document_test.go
```diff
@@ -0,0 +1,290 @@
...
+					ServiceMember: ppmShipment.Shipment.MoveTaskOrder.Orders.ServiceMember,
+					PPMShipment:   ppmShipment,
+					UserUploader:  userUploader,
+				}),
+			)
+		}
+
+	})
+
+	setUpRequestAndParams := func() ppmdocumentops.GetPPMDocumentsParams {
```
NamibiaTorres, line 69: Praise: I like the very clear naming of this function.

NamibiaTorres (approved):
Really great job! I'm so happy I get to review one of your back end tickets. I see that there is a n error related to playwright. Try merging in the main branch again to see if it goes away. I didn't find the error messages from the server super helpful, but given that this isn't a FE ticket it could be one of those weird flaky tests on the FE. Approving now because you covered all of the necessary BE code and updated per the previous reviewer comments.

pkg/handlers/ghcapi/ppm_document.go
felipe-lee: **note**: The forbidden error in 33-35 is one that the handler is raising if the request isn't from the office so that one is good there. The reason I called this one out here is mainly because this `switch/case` is for handling the errors the service object returns and in this case it doesn't return forbidden errors (the weight ticket handler that this came from used a service object that did return forbidden errors so that's why it was there).

felipe-lee (approved):
Awesome, thanks for making those updates. This lgtm!

felipe-lee:
Oh, I also tested with postman and successfully got the documents for a shipment out!
Used shipment `bf119998-785a-4357-a3f1-5e71ee5bc757`

I minimized two of the sections to be able to show all the doc types.

image

PR #10020, opened 2023-02-04
MB-14513 react-router upgrade

PDickeyTruss:
## Summary
The goal of this PR is to get the fronted applications working on the latest version of react-router (v6). The big pain point here is that there is interdependence between `react-admin` and `react-router` that necessitates upgrading both simultaneously. As a result, this PR includes enough react-admin upgrade work to allow the app to build/run, however, the primary focus of this PR review should be on the react-router upgrade work not the react-admin upgrade work. The react-admin work will be reviewed separately when complete, there are still a few outstanding tasks in [this epic](https://dp3.atlassian.net/browse/MB-14433).

### App testing
The following should work exactly like it does on the main branch:
* Office app for all user roles (SC, TIO, TOO, QA, Prime). 
* Customer app
* Routing in the Admin app. There are still admin app issues related to the admin upgrade the but links/routing in it should be fully flushed out and working. 
* All unit tests should pass
* All e2e tests should pass except for a couple admin tests that fail due to functionality broken in the upgrade. 

**Please pay extra attention to testing any and all links or special routing behavior that you may be aware of (especially if not covered by a e2e test)**. The kind of error that will most likely be present will be links or buttons that incorrectly navigate to either a blank page or a 404 error page rather than the expected page. 

### Code review
This is unfortunately and unavoidably a pretty brutal amount of code to review. I made an effort to add comments to key code changes in this PR to help highlight the most impactful and/or controversial code changes. I would recommend starting by looking though these comments before setting out into the full diffs. The [react-router v6 upgrade guide](https://reactrouter.com/en/main/upgrading/v5#introduction) is a great resource to get a feel for what changed in v6 and was the primary documentation supporting the work done in this PR. 

Note: Be sure to expand all the 'hidden items' in this PR. They include comments and this PR is long enough that github collapses them making it extremely easy to overlook them.

### Packages Upgraded
* `react-router` & `react-router-dom` (v5.1.2 -> v6.4.3)
* `react-admin` (v3.19.11 -> v4.5.3)
* `react-admin-import-csv` (v1.0.33 -> v4.0.0)
####Packages Removed:
* `history`
* `react-router-last-location`

### Other Highlights
- All existing react-router functionality has been refactored and/or replaced by new routing functionality and `connected-react-router` functionality has been fully replaced with native react-router functionality. This means that routing is no longer intertwined Redux and there is no need to pass around routing functions via props. 
- Added a New `withRouting` custom hook to provide backwards compatibility for routing hook functionality to class based components
- New testing utilities to simplify mocking routing when testing components that utilize routing (and other mock providers)
- Test updates and refactors to get existing tests passing and add significantly more coverage to files responsible for routing (index and MoveInfo files). 
- Standardize routing imports (always 'react-router-dom')

## Outstanding work/issues
- [ ] Unit Test Coverage decreased - This one is a bummer. In short, I added a ton of new test coverage around routing but the overall coverage dropped. After some exploring it looks like this coverage drop is due to the combination of a few things:
    1. Untested admin files got a few more lines of code as a part of the upgrade (going from props -> hooks) which resulted in more 'uncovered' lines. 
    2. New routing functionality in office/customer components that does have good test coverage typically required less lines of code for the same functionality, which means the total number of covered lines decreased 
    3. Greatly improved mocking and testing of routing files means less incidental coverage of nested components. 
    The combination of these things resulted in the coverage % dropping even though I would argue we now have more/better test coverage. Going forward we will either need to add/update tests to increase coverage or decide to reset the coverage threshold. 
- [ ] Documentation - I have an outstanding task that I have not yet started to add documentation around new routing and testing patterns. This will be coming soon. 
- [ ] Ensure no functionality was lost by removal of `react-router-last-location` - This package is no longer maintained and does not work with the new version of `react-router`. It was only minimally used in the codebase but I want to double check with someone familiar with it about the impacts of its removal. 

## Setup to Run Your Code
<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>


src/pages/Office/index.jsx
```diff
@@ -162,161 +168,201 @@ export class OfficeApp extends Component {
...
-                    <PrivateRoute
+                    <Route
+                      path="/moves/queue"
+                      end
+                      element={
+                        <PrivateRoute requiredRoles={[roleTypes.TOO]}>
+                          <MoveQueue />
+                        </PrivateRoute>
+                      }
+                    />
```
PDickeyTruss, line 198: We can still use `PrivateRoute` to restrict a route to users with specific roles but it is now inside of the Route's element prop.

src/pages/Office/index.jsx
```diff
@@ -162,161 +168,201 @@ export class OfficeApp extends Component {
...
+                      }
                     />
 
-                    <PrivateRoute
+                    <Route
                       key="txoMoveInfoRoute"
-                      path="/moves/:moveCode"
-                      component={TXOMoveInfo}
-                      requiredRoles={[roleTypes.TOO, roleTypes.TIO, roleTypes.QAE_CSR]}
+                      path="/moves/:moveCode/*"
```
PDickeyTruss, line 342: The trailing `*` on this path is required here to signify that this Route had nested (child) routes. The child routes (in TXOMoveInfo) will be relative to this path.

src/pages/Office/TXOMoveInfo/TXOMoveInfo.jsx
```diff
@@ -86,82 +98,88 @@ const TXOMoveInfo = () => {
...
-              destinationDutyLocationPostalCode={order?.destinationDutyLocation?.address?.postalCode}
-            />
-          </Route>
-
-          <Route path={qaeCSRRoutes.EVALUATION_REPORT_PATH} exact>
-            <Restricted to={permissionTypes.updateEvaluationReport} fallback={<Forbidden />}>
-              <EvaluationReport
+        <Routes>
+          <Route
+            path="details"
```
PDickeyTruss, line 103: Paths in this file are relative to the parent (/moves/:moveCode/*)

src/utils/routing.jsx
```diff
@@ -0,0 +1,18 @@
...
+    const location = useLocation();
+    const navigate = useNavigate();
+    const params = useParams();
+
+    // eslint-disable-next-line react/jsx-props-no-spreading
+    return <Component {...props} router={{ location, navigate, params }} />;
+  }
+
+  return ComponentWithRouterProp;
+}
```
PDickeyTruss, line 16: This is a utility to give class components routing functionality as a prop since they can not use hooks. It simply calls the 3 hooks we commonly use and passes them to as a prop named `router`.

src/containers/Headers/CustomerLoggedInHeader.jsx
```diff
@@ -10,18 +10,15 @@ import { logOut as logOutAction } from 'store/auth/actions';
...
     LogoutUser().then((r) => {
       const redirectURL = r.body;
       if (redirectURL) {
         window.location.href = redirectURL;
       } else {
-        history.push({
-          pathname: '/sign-in',
-          state: { hasLoggedOut: true },
-        });
+        navigate('/sign-in', { state: { hasLoggedOut: true } });
```
PDickeyTruss, line 21: This is a good example of how to navigate with state in react-router v6.

src/App/index.jsx
```diff
@@ -57,14 +57,14 @@ const App = () => {
         <Provider store={store}>
           <PersistGate loading={<LoadingPlaceholder />} persistor={persistor}>
             <AppContext.Provider value={officeContext}>
-              <ConnectedRouter history={history}>
```
PDickeyTruss, line 60: The changes in this file replace the top level setup of connected-react-router with react-router [BrowserRouter](https://reactrouter.com/en/main/router-components/browser-router) components.

src/testUtils.jsx
PDickeyTruss, line 1: There was a time working on this PR after the bulk of the upgrade work was done that I had over 700 failing unit tests due to new/changed react-router functionality and related refactoring. When working though getting these all passing again I found myself needing to extend our test utilities to simplify testing components that interact with react-router and other mock 'Providers/Wrappers' like redux, react-query, and permissions. I made an effort to keep the utils similar to our existing patterns to minimize impact across the app while also adding 'opt in' type of functionally.

Probably the most impactful test util update is that mock routing gets setup with more info to provide a more complete mock routing setup that allows hooks like `useLocation` and `useParams` to work correctly without needing to be mocked in the test file. This means there is no need to do mocking like: 
```
jest.mock('react-router', () => ({
  ...jest.requireActual('react-router'),
  useParams: jest.fn().mockReturnValue({ moveCode: 'LR4T8V' }),
}));
```

The biggest mindset and code changes is that we typically will want to setup up mock routing with two (optional) values: `path` and `params` rather than using an `initialEntries` string. The `initialEntries` string was used as basically the URL you want to render the component at ('/moves/ABC123/details'). Unfortunately, this string alone doesn't give react-router enough info to know what route it is matched on and what sections of string are params vs pathing. Providing `path` ('/moves/:moveCode/details') and a params object (`{moveCode: 'ABC123'}`) does give enough info to create and match on a mock route and parse out params/location values as needed in hooks.

src/testUtils.jsx
```diff
@@ -1,56 +1,88 @@
...
+  ];
+
+  // include any additional routes provided
+  if (routes && routes.length > 0) mockRoutes.push(...routes);
+
+  const router = createMemoryRouter(mockRoutes, {
+    initialEntries: [generatePath(path, params)] || ['/'],
+  });
+
+  return router;
```
PDickeyTruss, line 28: This is a key bit of code. It does all the mock routing setup for all other test utility functions/components. There is more/better inline documentation around these values in functions that are exported from this file but this is the spot where they really get used.

src/testUtils.jsx
```diff
@@ -1,56 +1,88 @@
...
     <ReactQueryWrapper client={client}>
       <PermissionProvider permissions={permissions} currentUserId={currentUserId}>
         <Provider store={mockStore.store}>
-          <ConnectedRouter history={mockHistory}>
-            <Suspense fallback={<LoadingPlaceholder />}>{children}</Suspense>
-          </ConnectedRouter>
+          <RouterProvider router={mockRouter} />
         </Provider>
       </PermissionProvider>
     </ReactQueryWrapper>
```
PDickeyTruss, line 88: HOC to provide all mock providers around the provided children: redux, routing, permissions, and react-query.

src/testUtils.jsx
```diff
@@ -60,32 +92,79 @@ export const MockProviders = ({
...
+ * @param {*} ui - The component to be rendered using react-testing-library
+ * @param {*} options - Routing options used to create a mock router. Prefered options are `path` and `params`. Other options are for extended use cases and supporting existiong patterns.
+ * @returns {*} - The result of the render call from react-testing-library
+ */
+export const renderWithRouter = (ui, options) => {
+  return render(<MockRouterProvider {...options}>{ui}</MockRouterProvider>);
+};
+
+export const renderWithProviders = (ui, options) => {
+  return render(<MockProviders {...options}>{ui}</MockProviders>);
```
PDickeyTruss, line 131: Helper functions to render with routing/providers around a given component.

src/testUtils.jsx
```diff
@@ -60,32 +92,79 @@ export const MockProviders = ({
...
+        .substring(path.lastIndexOf('/') + 1)
+        .replace(/([A-Z])/g, ' $1')
+        .trim();
 
-  return {
-    memoryHistory,
-    mockProviderWithHistory,
-  };
+    return () => <div>{`Mock ${componentName} Component`}</div>;
+  });
```
PDickeyTruss, line 169: Mock the component that is imported from the provided path with a basic div including the component name as text. Used when testing files that setup routes like the index files and MoveInfo files to mock sub components that are out of scope of the current test.

src/pages/Office/ServicesCounselingQueue/ServicesCounselingQueue.jsx
```diff
@@ -178,83 +178,72 @@ const closeoutColumns = (ppmCloseoutGBLOC) => [
...
-
   if (isLoading) return <LoadingPlaceholder />;
   if (isError) return <SomethingWentWrong />;
+  if (!queueType) {
+    return inPPMCloseoutGBLOC ? (
+      <Navigate to={servicesCounselingRoutes.BASE_QUEUE_CLOSEOUT_PATH} />
+    ) : (
+      <Navigate to={servicesCounselingRoutes.BASE_QUEUE_COUNSELING_PATH} />
+    );
+  }
```
PDickeyTruss, line 202: Refactored default queue behavior to ensure correct queue is shown for users gbloc. The crux of being able to do this vs needing a useEffect was proving the queueType as a param by tweaking the parent route path definition in the Office Index file so it could be retrieved via the useParams hook.

src/containers/CustomerPrivateRoute/CustomerPrivateRoute.jsx
PDickeyTruss, line 1: This file was deleted as its functionality (loading state and logged out redirect) was replaced by native react-router functionality in the Customer index.

src/scenes/MyMove/index.jsx
```diff
@@ -100,154 +99,154 @@ export class CustomerApp extends Component {
...
+                    <div className="usa-grid">
+                      <h2>We are experiencing an internal server error</h2>
+                    </div>
+                  }
+                />
+                <Route end path="/invalid-permissions" element={<InvalidPermissions />} />
+                <Route
+                  path="*"
+                  element={(loginIsLoading && <LoadingPlaceholder />) || <Navigate to="/sign-in" replace />}
+                />
```
PDickeyTruss, line 161: This route handles most of the functionality that was in CustomerPrivateRoute (now deleted).

src/scenes/MyMove/index.jsx
```diff
@@ -100,154 +99,154 @@ export class CustomerApp extends Component {
...
-                  <Route render={(routeProps) => <NotFound {...routeProps} handleOnClick={this.props.goBack} />} />
-                </Switch>
-              )}
-            </main>
-            <Footer />
-          </div>
-          <div id="modal-root"></div>
-        </LastLocationProvider>
+                {/* 404 - user logged in but at unknown route */}
+                <Route path="*" element={<NotFound />} />
```
PDickeyTruss, line 243: No need to pass routing and navigation props around. Go back is just `navigate(-1)`and other routing functionality can be obtained via hook 🎉

src/scenes/Office/Queues.jsx
```diff
@@ -1,28 +1,28 @@
...
 import QueueList from './QueueList';
 import QueueTable from './QueueTable';
 import LoadingPlaceholder from '../../shared/LoadingPlaceholder';
 
-export class Queues extends Component {
-  render() {
-    const queueType = this.props.queueType || this.props.match.params.queueType;
+const Queues = (props) => {
+  const params = useParams();
+  const queueType = props.queueType || params.queueType;
```
PDickeyTruss, line 10: This was a trivial class component that received routing params via props.  I went ahead and refactored this to be a functional component so it can use the routing hooks.

src/types/router.js
```diff
@@ -15,13 +8,8 @@ export const LocationShape = shape({
...
+  location: LocationShape,
+  params: shape({}),
 });
-
-// Components that are rendered at a route (<Route />) will receive these props automatically
-export const RouteProps = {
-  match: MatchShape.isRequired,
-  location: LocationShape.isRequired,
-  history: HistoryShape.isRequired,
-};
```
PDickeyTruss, line 27: No need for old routing props and there shapes anymore as all now comes via hook (or `router` prop for class components)

src/scenes/MyMove/index.jsx
PDickeyTruss: Little hard to tell from the dif that got picked up here: This went from `<Route render={(routeProps) => <NotFound {...routeProps} handleOnClick={this.props.goBack} />} />` to `<Route path="*" element={<NotFound />} />` 😎

src/constants/routes.js
```diff
@@ -42,42 +42,72 @@ export const customerRoutes = {
...
-  EVALUATION_VIOLATIONS_PATH: '/moves/:moveCode/evaluation-reports/:reportId/violations',
+  BASE_EVALUATION_REPORTS_PATH: `${BASE_MOVES_PATH}/evaluation-reports`,
+  EVALUATION_REPORTS_PATH: '/evaluation-reports',
+  BASE_EVALUATION_REPORT_PATH: `${BASE_MOVES_PATH}/evaluation-reports/:reportId/*`,
+  EVALUATION_REPORT_PATH: '/evaluation-reports/:reportId',
+  BASE_EVALUATION_VIOLATIONS_PATH: `${BASE_MOVES_PATH}/evaluation-reports/:reportId/violations`,
+  EVALUATION_VIOLATIONS_PATH: '/evaluation-reports/:reportId/violations',
+  BASE_CUSTOMER_SUPPORT_REMARKS_PATH: `${BASE_MOVES_PATH}/customer-support-remarks`,
+  CUSTOMER_SUPPORT_REMARKS_PATH: 'customer-support-remarks',
 };
```
PDickeyTruss, line 111: There was some inconsistency here that could probably be cleaned up in future work. The customer app is using all absolute/non-relative routing (all full paths staring with '/'). Whereas the office app is using nested/relative routing. This means in the office app that we need both a full 'base' path for the routing setup and also need the relative part of the path for in nested routing files (TXOMoveInfo.jsx) and when creating nested links/navigations. 

My solution to this for now is to add a 'Base' and non-base version of the links with the 'Base' version being absolute and the non-base being relative. 

There are valid arguments supporting and rejecting using relative vs absolute routing/pathing. Relative linking allows for much simpler links that don't require building full absolute paths from URL params but run the risk of being bad links if they are reused in a different parent/child nesting context than intended. I don't think it would be too painful to switch all our routing to be consistently absolute or relative but I am leaving that decision and work for future tickets.

src/constants/routes.js
```diff
@@ -42,42 +42,72 @@ export const customerRoutes = {
...
+  BASE_CUSTOMER_INFO_EDIT_PATH: `${BASE_COUNSELING_MOVE_PATH}/customer`,
+  CUSTOMER_INFO_EDIT_PATH: 'customer',
+  BASE_MOVE_VIEW_PATH: `${BASE_COUNSELING_MOVE_PATH}/details`,
+  MOVE_VIEW_PATH: 'details',
+  BASE_ORDERS_EDIT_PATH: `${BASE_COUNSELING_MOVE_PATH}/orders`,
+  ORDERS_EDIT_PATH: 'orders',
+  BASE_SHIPMENT_ADD_PATH: `${BASE_COUNSELING_MOVE_PATH}/new-:shipmentType`,
+  SHIPMENT_ADD_PATH: 'new-:shipmentType',
+  BASE_SHIPMENT_EDIT_PATH: `${BASE_COUNSELING_MOVE_PATH}/shipments/:shipmentId`,
+  SHIPMENT_EDIT_PATH: 'shipments/:shipmentId',
```
PDickeyTruss, line 67: There was some inconsistency here that could probably be cleaned up in future work. The customer app is using all absolute/non-relative routing (all full paths staring with '/'). Whereas the office app is using nested/relative routing. This means in the office app that we need both a full 'base' path for the routing setup and also need the relative part of the path for in nested routing files (TXOMoveInfo.jsx) and when creating nested links/navigations.

My solution to this for now is to add a 'Base' and non-base version of the links with the 'Base' version being absolute and the non-base being relative.

There are valid arguments supporting and rejecting using relative vs absolute routing/pathing. Relative linking allows for much simpler links that don't require building full absolute paths from URL params but run the risk of being bad links if they are reused in a different parent/child nesting context than intended. I don't think it would be too painful to switch all our routing to be consistently absolute or relative but I am leaving that decision and work for future tickets.

src/scenes/Office/QueueList.jsx
```diff
@@ -8,32 +8,44 @@ export default class QueueList extends Component {
...
               data-testid="ppm-payment-requests-queue"
             >
               <span>Payment requested</span>
             </NavLink>
           </li>
           <li className="usa-sidenav__item">
-            <NavLink to="/queues/ppm_completed" activeClassName="usa-current" data-testid="ppm-payment-requests-queue">
+            <NavLink
+              to="/queues/ppm_completed"
+              className={({ isActive }) => (isActive ? 'usa-current' : '')}
```
PDickeyTruss, line 37: The 'activeClassName' prop changed to this in the new version of RR NavLink per: https://reactrouter.com/en/main/upgrading/v5#remove-activeclassname-and-activestyle-props-from-navlink-

src/testUtils.jsx
mandaem: 700+ failing tests, github needs this https://scrapbox.io/emoji/party_try_not_to_cry

src/components/Customer/Review/Summary/Summary.jsx
```diff
@@ -202,17 +203,24 @@ export class Summary extends Component {
...
     const officePhone = currentDutyLocation?.transportation_office?.phone_lines?.[0];
 
     const rootReviewAddressWithMoveId = generatePath(customerRoutes.MOVE_REVIEW_PATH, { moveId });
 
     // isReviewPage being false is the same thing as being in the /edit route
-    const isReviewPage = rootReviewAddressWithMoveId === match.url;
+    const isReviewPage = matchPath(
+      {
+        path: rootReviewAddressWithMoveId,
+        end: true,
```
mboren, line 220: Just a note for other reviewers, not a change request: 

 `end: true` is not explained in the docs for `matchPath` https://reactrouter.com/en/main/utils/match-path
 but it is mentioned in the v5->v6 upgrade doc: https://reactrouter.com/en/main/upgrading/v5#change-the-order-of-arguments-passed-to-matchpath-change-pathpattern-options
it says:
> `true` == pattern should match the entire URL pathname

src/constants/routes.js
```diff
@@ -42,42 +42,72 @@ export const customerRoutes = {
...
 };
 
 export const qaeCSRRoutes = {
   MOVE_SEARCH_PATH: '/qaecsr/search',
-  EVALUATION_REPORTS_PATH: '/moves/:moveCode/evaluation-reports',
-  EVALUATION_REPORT_PATH: '/moves/:moveCode/evaluation-reports/:reportId',
-  EVALUATION_VIOLATIONS_PATH: '/moves/:moveCode/evaluation-reports/:reportId/violations',
+  BASE_EVALUATION_REPORTS_PATH: `${BASE_MOVES_PATH}/evaluation-reports`,
+  EVALUATION_REPORTS_PATH: '/evaluation-reports',
+  BASE_EVALUATION_REPORT_PATH: `${BASE_MOVES_PATH}/evaluation-reports/:reportId/*`,
```
mboren, line 105: is the `*` meant to be here? I don't see it on any of the other routes.

src/shared/WizardPage/Form.jsx
```diff
@@ -153,17 +150,12 @@ WizardFormPage.propTypes = {
   pageKey: PropTypes.string.isRequired,
   valid: PropTypes.bool,
   dirty: PropTypes.bool,
-  push: PropTypes.func,
-  match: PropTypes.object, //from withRouter
+  router: PropTypes.object, //from withRouter
```
duncan-truss, line 153: Could we define this prop type shape if we haven't yet?  I feel like PropTypes.object is usually flagged but maybe because we're in `src/shared` it is ignoring that eslint rule?

src/containers/PrivateRoute.jsx
```diff
@@ -20,32 +18,24 @@ export function userIsAuthorized(userRoles, requiredRoles) {
   return !!userRoles?.find((r) => requiredRoles.indexOf(r) > -1);
 }
 
-const PrivateRoute = (props) => {
-  const { loginIsLoading, userIsLoggedIn, requiredRoles, userRoles, ...routeProps } = props;
+function PrivateRoute({ requiredRoles, userRoles, children }) {
   const userRoleTypes = getRoleTypesFromRoles(userRoles);
-  if (loginIsLoading) return <LoadingPlaceholder />;
```
mboren, line 26: Why were we able to just remove `loginIsLoading`? Is it handled in a parent or child component instead?

I'm pretty sure that this is right, I just don't understand it.

src/pages/MyMove/Orders.test.jsx
```diff
@@ -404,12 +416,16 @@ describe('Orders page', () => {
...
+        </MockRouterProvider>,
+      );
       await waitFor(() => {
         expect(wrapper.find('OrdersInfoForm').prop('ordersTypeOptions')).toEqual([ordersOptions[0]]);
       });
     });
   });
 
-  afterEach(jest.clearAllMocks);
+  afterEach(jest.resetAllMocks);
```
mboren, line 430: is this redundant with the `jest.resetAllMocks` call on line 148?

src/pages/Office/CustomerInfo/CustomerInfo.jsx
dymonmoore, line 1: Do we no longer need the move code in the route path?

src/pages/Office/EditShipmentDetails/EditShipmentDetails.test.jsx
dymonmoore, line 1: What's the difference between render and renderWithProviders?

src/shared/WizardPage/Form.jsx
PDickeyTruss: Great spot! We do have a router shape that I missed using here. Fixed ✅

src/pages/Office/EditShipmentDetails/EditShipmentDetails.test.jsx
PDickeyTruss: `render` is the react-testing-library standard render function to render a component you wish to test. `renderWithProviders` is essentially an enhanced version of this that first wraps the component to be rendered with all our commonly used providers (redux, react-query, routing, and permissions) before rendering it with `render`. Probably the best place to look for more on this its implementation in the test utils file (src/testUtils.jsx).

src/pages/Office/CustomerInfo/CustomerInfo.jsx
PDickeyTruss: Great question! The new routing path is using relative routing while the old one is absolute. The old one used to build up the full path that had the move code as part of the path, something like `/counseling/moves/ABC123/details`. However, the new one takes advantage of relative routing to not need to build up this full path using the move code. It builds a path that is simply `../details`, which means go up one routing level and append `/details` on the end, which gets us to the same place as building out the while path but requires.

src/pages/MyMove/Orders.test.jsx
PDickeyTruss: Yep! This is absolutely redundant! Have removed it, good spot! 🦅 👁️

src/pages/Office/PPM/ReviewDocuments/ReviewDocuments.jsx
dymonmoore, line 1: I'm noticing a pattern of removing MatchShape... Curious, what was it's prior purpose?

PDickeyTruss: We used to pass routing functions/info around via props (a bad smell in v6). One of the most common ones was a `match` prop that we used to use to pull out routing params from the URL. In v6 we no longer need to pass anything around via props (except to class components by use of `withRouter`) because we can use hooks to get at this info/functionality. In the case of `match`, we used to commonly do something like `const { shipmentId, moveCode } = match.params;` to get URL params from the match prop. In v6 we can just use the `useParams` hook to get at this info: `const {shipmentId, moveCode} = useParams();`.

src/constants/routes.js
PDickeyTruss: Very good spot, this is now removed and was indeed unnecessary. The trailing `*` tells react-router that there are nested/child routes that will go underneath the route. This route can have `violations` added on to it but i ended up making this it's own route/path (BASE_EVALUATION_VIOLATIONS_PATH) rather than deal with nesting. I think the `*` was an artifact from when I first refactored and was trying to use nesting. Thanks for calling this one out!

src/containers/PrivateRoute.jsx
PDickeyTruss: Very good question 💯. I tweaked the 'PrivateRoute' component to be more singularly focused, mostly because we can only use pure `Routes` in v6. The PrivateRoute component is now only responsible for restricting access based on user roles. All other functionality it had (like displaying loading placeholder when loginIsLoading) is now taken care of up in the file(s) that define the routes using native react-router functionality. So in the case of Office app, the loading placeholder functionality is taken care of by [lines 179-182 of /src/pages/Office/index.jsx](https://github.com/transcom/mymove/blob/1e95d2e8b184e909a3de89564977b757cdbe93b2/src/pages/Office/index.jsx#L180).  The main driver behind making this change was this part of the upgrade doc: https://reactrouter.com/en/main/upgrading/v5#refactor-custom-routes. 

Here is pretty good doc that I used when doing this specific part of the work that is worth a peek if you have the time/interest: https://gist.github.com/mjackson/d54b40a094277b7afdd6b81f51a0393f

src/containers/Headers/CustomerLoggedInHeader.jsx
duncan-truss: I have said we could probably use state more than we do for some things like success notifications between pages or potentially saying which page was referring you somewhere if it wasn't obvious from the history.  On the other hand it's something that can probably be abused if relied on too heavily.

src/shared/WizardPage/index.jsx
```diff
@@ -102,8 +101,7 @@ WizardPage.propTypes = {
   pageIsValid: PropTypes.bool,
   canMoveNext: PropTypes.bool,
   dirty: PropTypes.bool,
-  push: PropTypes.func,
-  match: PropTypes.object, //from withRouter
+  router: PropTypes.object, //from withRouter
```
duncan-truss, line 104: Sorry found one more generic proptype

src/scenes/MyMove/index.jsx
```diff
@@ -1,8 +1,6 @@
 import React, { Component, lazy } from 'react';
 import PropTypes from 'prop-types';
-import { LastLocationProvider } from 'react-router-last-location';
```
duncan-truss, line 3: Looks like this was only used for legacy PPM things so safe to remove

src/scenes/MyMove/index.jsx
```diff
@@ -100,154 +99,154 @@ export class CustomerApp extends Component {
...
+                <Route end path="/weight-ticket-examples" element={<WeightTicketExamples />} />
+                <Route end path="/trailer-criteria" element={<TrailerCriteria />} />
+                <Route end path="/allowable-expenses" element={<AllowableExpenses />} />
+                <Route end path="/infected-upload" element={<InfectedUpload />} />
+                <Route end path="/processing-upload" element={<ProcessingUpload />} />
+                <Route path="/moves/:moveId/ppm-payment-request-intro" element={<PPMPaymentRequestIntro />} />
+                <Route path="/moves/:moveId/ppm-weight-ticket" element={<WeightTicket />} />
+                <Route path="/moves/:moveId/ppm-expenses-intro" element={<ExpensesLanding />} />
+                <Route path="/moves/:moveId/ppm-expenses" element={<ExpensesUpload />} />
+                <Route path="/moves/:moveId/ppm-payment-review" element={<PaymentReview />} />
```
duncan-truss, line 216: Sorry most of these are obsolete and could have been removed but sometimes that requires fixing a lot of tests too.

src/scenes/MyMove/index.jsx
```diff
@@ -271,6 +271,7 @@ CustomerApp.defaultProps = {
   loadInternalSchema: no_op,
   loadUser: no_op,
   initOnboarding: no_op,
+  loginIsLoading: false,
```
duncan-truss, line 274: Not sure this default has any real consequence as the value is coming from the redux selector but noticed that in the office we do set it to true.

src/scenes/MyMove/index.test.js
```diff
@@ -1,143 +1,229 @@
...
+        <Suspense fallback={<div>Loading...</div>}>
+          <CustomerApp {...minProps} />
+        </Suspense>
+      </Provider>
+    </MemoryRouter>,
+  );
+};
+
+afterEach(() => {
+  cleanup();
```
duncan-truss, line 74: Is this because we've mixed enzyme and testing-library tests that we need to call this cleanup method or some other reason?

src/shared/store.js
```diff
@@ -2,8 +2,6 @@ import { createStore, applyMiddleware } from 'redux';
 import { persistStore, persistReducer } from 'redux-persist';
 import storage from 'redux-persist/lib/storage';
 import { appReducer, adminAppReducer } from 'appReducer';
-import { createBrowserHistory } from 'history';
-import { routerMiddleware } from 'connected-react-router';
```
duncan-truss, line 6: I saw in one [migration guide](https://github.com/supasate/connected-react-router/blob/master/FAQ.md#how-to-migrate-from-v4-to-v5v6﻿﻿) the routerMiddleware is still used.  How were we able to replace it?

src/shared/WizardPage/index.jsx
PDickeyTruss: Good 👁️ , fixed!

src/scenes/MyMove/index.jsx
PDickeyTruss: Very glad to hear that... 😌 I had a hunch that was the case but very much appreciate the confirmation. The package depended on old react-router so it kinda had to go. 😜

src/scenes/MyMove/index.jsx
PDickeyTruss: No worries, pretty easy bulk changes to update all these so not much time/effort spent on it.

src/scenes/MyMove/index.jsx
PDickeyTruss: Good stuff, thank you for looking deep enough to find things like this! I agree this should be consistent even if it doesn't have much consequence. I updated the office app to have this as `loginIsLoading: false` since that seems like the most appropriate default state so align on.

src/scenes/MyMove/index.test.js
PDickeyTruss: This is an artifact of a test file I struggled with 😆. I just removed it. In short, I was getting tests that passed individually but not when I ran them all together and I went way overboard trying to make sure everything was fully cleaned/reset between runs. Turns out [cleanup gets called automatically anyways](https://testing-library.com/docs/react-testing-library/api/#cleanup) so there is no real reason for us to do this.

src/pages/Office/ServicesCounselingMoveInfo/ServicesCounselingMoveInfo.jsx
dymonmoore, line 1: I was wondering why we use replace on the last route here?

src/pages/Office/ServicesCounselingQueue/ServicesCounselingQueue.jsx
dymonmoore: LGTM!

src/pages/Office/index.jsx
dymonmoore, line 1: I'm a bit confused by lines 115 and 116, are they suppose to have the same name?

dymonmoore, line 1: I noticed most of the Route component pass in a end parameter, but some of them do not. What's the difference hear? I tried to peak at documentation, but didn't find much.

dymonmoore (commented):
Overall the code looks good to me. I left a bunch of comments to help me get a better understanding of some things. In my review, here's some frequent changes I found: replacement of useHistory for useNavigate (including their mocks) according to the guide, removal of the initialEntries parameters containing paths from Mock Providers component, Updated render functions to pass routing parameters, updated the <Routes> component to <Switch>, and removed <WithRouter>.

dymonmoore:
Finished app testing. The routing looks good and similar to what's on main. I noticed under the payment requests tab (on this branch and main), there's an error when clicking on the review service items button. 
Screenshot 2023-04-06 at 12 57 53 PM

duncan-truss:
@dymonmoore Could you provide us more information on which move code or user that payment request belongs to locally https://github.com/transcom/mymove/pull/10020#issuecomment-1499361430?  It would also be helpful to capture the error message shown in the dev console and if you see anything suspicious in the network tab as well.

Sometimes the devseed moves have randomly generated move codes and information so helping trace it to a particular seed function can be helpful in understanding if it is missing required data or an actual error in our application's logic.

ahobson:
For another change I was working on, I added some tests to the Admin pages to bump up the client test coverage. Would it be helpful to share those here?

PDickeyTruss:
> Finished app testing. The routing looks good and similar to what's on main. I noticed under the payment requests tab (on this branch and main), there's an error when clicking on the review service items button.

Good find here, this one is actually present in main. I found this same one in my testing and have a bug created for it here: https://dp3.atlassian.net/browse/MB-15520

src/pages/Office/index.jsx
PDickeyTruss: Appreciate the good question and explanation here. To add just a hair more, the reason for the destructuring being a little funky here is that I needed both the `location` and the `pathname` out of the `router` prop. The object getting destructured looks something like:
```
const props = {router: {location: {pathname: 'abc', search: 'a=b', ...}, ...}}
```
So to get both the location and the pathname out of this we could destructure like:
```
    const {router} = props
    const {location} = router
    const {pathname} = location
```
In the file I do essentially the same thing but all at once. This gives me the two vars out of this that we needed later in this file, `location` and `pathname`. 
```
const { router: { location, location: { pathname }}
```

src/pages/Office/ServicesCounselingMoveInfo/ServicesCounselingMoveInfo.jsx
PDickeyTruss: The code I was refactoring here used to be a `Redirect` component that by default uses 'replace' logic. However, in react-router v6 the `Navigate` component uses 'push' logic by default. I had to include the 'replace' prop to get the same functionality that existed previously. I only knew to do this due to the upgrade guide, specifically [this section](https://reactrouter.com/en/main/upgrading/v5#use-usenavigate-instead-of-usehistory). 

Their example:
```
// Change this:
<Redirect to="about" />
<Redirect to="home" push />

// to this:
<Navigate to="about" replace />
<Navigate to="home" />
```

src/pages/Office/index.jsx
PDickeyTruss: In react-router v6 they changed from using `exact` to `end` to mean (almost) the same thing. In these routes you can basically think of 'end' as meaning 'exact'.  As an example, a route with 'end' and path='/abc' will only match when the URL is exactly '/abc'. Whereas if this route did not have the 'end' prop, the URL '/abc/123' would also match.

duncan-truss (approved):
@PDickeyTruss I think you can remove the design review here, not sure there's any ask of design and we could wait until the final PR.

PR #10026, opened 2023-02-06
MB-15157: Change playwright worker default based on arch

ahobson:
## Summary

This changes the defaults for playwright workers because we were seeing flaky tests with the default configuration on local machines.

Run the playwright tests locally and you should see better reliability.


kylehilltruss (approved):
Looks like a good plan. I've been showing folks the `--workers 1` flag and telling them to focus on running smaller numbers of tests when they're actively working on development for something related to integration tests for a faster feedback cycle, though obviously we'll want folks to be able to just run a full `playwright test` before they open a PR. 🚀

PR #10037, opened 2023-02-07
MB-15172: Use the default channel for circleci messages for one stop config

ahobson:
## Summary

Switch to the default channel for CircleCI failure messages

### DIFF
--- a/scripts/circleci-announce-broken-branch
+++ b/scripts/circleci-announce-broken-branch
@@ -17,10 +17,11 @@ if [[ $CIRCLE_JOB == *"deploy"* ]]; then
   color="danger"
 fi
 
+# use the default channel on the webhook so that config is handled in
+# one place
 slack_payload=$(
 cat <<EOM
 {
-    "channel": "#prac-engineering",
     "attachments": [
         {
             "fallback": "$message $CIRCLE_BUILD_URL",


kctruss (approved):
👍

rogeruiz:
I agree with the approval here, but I wonder if configuring this backwards in the future is a better strategy. I personally would like the UI to be the last place folks need to look for the configuration changes. But since this is a small script, I'd say the comment you put here @ahobson is good enough for now. cc: @transcom/truss-platform-team

YanZ777 (approved):
I have some mild trepidation about people forgetting how all this stuff works. That's a bit outside the scope of this PR, so I'm approving.
I do like having the config just be in one place though.

ahobson:
I'll merge it now and we can decide if we want a different idea going forward.

The iam notifier uses the default channel configured in the slack UI, so I felt like I wanted both to work the same way. I also liked the idea that we can change the channel in the slack admin UI without requiring a commit for the config propagate everywhere. But again, it's easy to change our mind.

rogeruiz:
Thanks for doing all this @ahobson and I disagree that changing the name in any UI though is a good idea. It took a while to find the UI you mentioned.

[=> Slack UI where web hooks are updated](https://ustcdp3.slack.com/apps/A0F7XDUAZ-webhooks-entrantes?tab=settings&next_id=0)

On this page there's a deprecation notice for these UI hooks, too.

> Please note, this is a legacy custom integration - an outdated way for teams to integrate with Slack. These integrations lack newer features and they will be deprecated and possibly removed in the future. We do not recommend their use. Instead, we suggest that you check out their replacement: [Slack apps](https://api.slack.com/start).

Now, as I said before, configuring this backwards where the configuration is in code and then pushed up to Slack gives us greater accountability and reliability when it comes to making changes like you did this week. Sadly, there's not really a great ecosystem for updating these things outside of the web UI at first glance. But I think doing a bit more research on this on the Platform team can help us come up with a good overall solution.

PR #10060, opened 2023-02-10
MB-15202: switch to playwright from cypress

ahobson:
## Summary

Completely convert to using playwright. We can revert this PR if necessary.

* Convert the mtls integration tests to run in a similar way to playwright, outside of docker
* Remove the `docker-compose.e2e.yml` and related config since that is now unused.
* Convert `Makefile` targets and scripts to use playwright.

You can try out a run with
```shell
make e2e_test
```


kylehilltruss (approved):
Looks good - lets do it!

PR #10061, opened 2023-02-10
[ MB-14692, MB-14723 ] Updating the Admin API metadata

rogeruiz:
## Summary

This touches the `schemes`, `host`, and `contact` stanzas only. The main reason
for using the email address for Pagerduty is this is what the client is
expecting to use for support escalation. We can modify this email to be a
placeholder `admin@example.com` email as well. I've also made the `host:`
stanza equal to the production URL as Swagger 2.x doesn't allow for multiple
hosts the way OpenAPI 3.x does. Since the main place this API will be used
is in the PRD environment, I don't think this is a problem.

**Note**: The Jira ticket for this work is a bit vague but I wanted to tie it to
the work associated with the AC for the ticket rather than the ticket itself.
Hope this makes sense, please ask me for clarification if it does not and I'll
update this section accordingly.

## Setup to Run Your Code

I ran `make swagger_generate server_generate` on this branch to get all the necessary files checked into the repository. 

As the reviewer, you can run `lint-yaml-with-spectral` from inside this repository and with the `swagger/admin.yaml` and should see that there are only **✖ 128 problems (42 errors, 86 warnings, 0 infos, 0 hints)** which is different from the default branch of **✖ 130 problems (42 errors, 88 warnings, 0 infos, 0 hints)**


swagger-def/admin.yaml
```diff
@@ -1,4 +1,7 @@
+schemes:
+  - https
 swagger: '2.0'
+host: admin.move.mil # OpenAPI 2.x only supports a single host, while OpenAPI 3.x supports multiple server stanzas.
```
felipe-lee, line 4: **question**: Do you happen to know if we have a public doc where the other env hosts are available? Asking because if so, we could link to it in the comment. But if that isn't public, this looks good to me.

felipe-lee (commented):
This lgtm, just have a few questions. Only not approving yet so it doesn't auto-merge before getting those answers.

Also, I pulled the branch down and confirmed that the meta warnings from `main`

```shell
   1:1    warning  oas2-api-schemes               OpenAPI host "schemes" must be present and non-empty array.
   2:6    warning  info-contact                   Info object must have "contact" object.                                 info
```

are no longer coming up in this branch.

swagger-def/admin.yaml
```diff
@@ -7,6 +10,9 @@ info:
   license:
     name: MIT
     url: https://github.com/transcom/mymove/blob/main/LICENSE.md
+  contact: # Contact is set to Pagerduty
+    name: MilMove AppEng
+    email: support@movemil.pagerduty.com
```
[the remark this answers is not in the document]
rogeruiz: > **questions**:  
> 
> 1) Is this email fine to have publicly available?

Yup! It's because PagerDuty has the ability to filter emails based on names and host names as well. Also having a clear support email will be helpful for readers of the API to know who to contact. 

[➡️ PagerDuty email filtering](https://support.pagerduty.com/docs/email-management-filters-and-rules)

> 2) Do we have appeng in pagerduty anymore? I thought we'd moved away from that, though maybe I misunderstood. 

@Ryan-Koch has access as anchor lead. This has been very ambiguous in the past, but it will eventually be clear I assume.

swagger-def/admin.yaml
rogeruiz: > **question**: Do you happen to know if we have a public doc where the other env hosts are available? 

I'm sure there are references to our 5 environments in various places. They are also documented in Confluence. 

> Asking because if so, we could link to it in the comment. But if that isn't public, this looks good to me.

I would recommend we shy away from comments for important / pertinent information like that. We can add this to the description of all of our APIs and have the include happen from Markdown files.

I am tackling descriptions and documentation in a much different PR coming shortly. It will be very verbose as our APIs lack a lot of descriptive or summarized text beyond a single sentence or two.

swagger-def/admin.yaml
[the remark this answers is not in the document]
felipe-lee: Ok, sounds good, thanks!

swagger-def/admin.yaml
felipe-lee: Ah ok, that works for me.

felipe-lee (approved):
Thanks for answering my questions, this lgtm!

PR #10064, opened 2023-02-10
[ MB-14692, MB-14723 ] Reordering the defintions stanza for Admin Swagger

rogeruiz:
## Summary

- [x] Sorting the defintions;
  - We can most likely turn this into an error again if we want. This only
removes a single warning from the checks, but I think it's important to
have these sorted in our `swagger-def` directory as it's now much easier
to read all the definitions and know what we support when code-folding
is enabled in your editor of choice.
- [x] Updating the definitions-sort given stanza;
  - Before this change, this the `defintions-sort` rule was giving false
positives.

## Setup to Run Your Code

I ran make swagger_generate server_generate on this branch to get all the
necessary files checked into the repository.

As the reviewer, you can run lint-yaml-with-spectral from inside this repository
and with the swagger/admin.yaml and should see that there are only **✖ 129
problems (42 errors, 87 warnings, 0 infos, 0 hints)** which is different from
the default branch of **✖ 130 problems (42 errors, 88 warnings, 0 infos, 0
hints)**.


rogeruiz:
Here's a screenshot of the sorted order.

Captura de pantalla 2023-02-10 a la(s) 12 28 31 p m

kctruss (approved):
Thanks for doing this!

PR #10067, opened 2023-02-10
[ MB-14692, MB-14723 ] Swagger Admin updates for responses stanza

rogeruiz:
## Summary

This removes the Spectral parsing error around numbers being the keys in
the responses stanza.

## Setup to Run Your Code

I ran `make swagger_generate server_generate` on this branch to get all the necessary files checked into the repository. 

As a reviewer, if your editor has linting you won't see any warnings for the
file around having numbers as keys in the `reponses` stanza. Example of the
parsing error is as follows:

```
spectral: must be a string scalar rather than number [parser]
```


kctruss (approved):
👍

PR #10077, opened 2023-02-13
MB-15204: playwright typescript linting

ahobson:
## Summary

Add better linting for playwright tests by using jsdoc + typescript. In particularl, this allows us to catch errors when a test does not await a promise. Adding these linting rules uncovered about a hundred missing awaits. Fixing those significantly decreased the flakiness of the playwright tests.

This kind of linting cannot be done by javascript alone because javascript linters cannot preserve async type information across files, only TypeScript can do that.

```
make client_deps
yarn lint
pre-commit run playwright_tsc --all-files
```


playwright/tests/utils/waitForPage.js
```diff
@@ -0,0 +1,111 @@
+// @ts-check
+import * as base from '@playwright/test';
+
+export class WaitForPage {
+  /**
+   * Create a WaitForPage.
+   * Use a class so we can get catch typos or type errors
```
YanZ777, line 7:
```suggestion
   * Use a class so we can catch typos or type errors
```

ahobson: Thank you! Fixed!

YanZ777 (approved):
Let's do this linting thing!

PR #10078, opened 2023-02-13
[ MB-15224 ] Specifying golang version for golangci-lint environment creation

mmayo-truss:
## Summary

This PR specifies that pre-commit uses our golang version that is installed on our system in the golangci-lint precommit yaml file. In the update to 3.0.x golang gets bootstrapped when it is not found on the system and it defaults to the latest version of golang when not specified. [Link to PR
](https://github.com/pre-commit/pre-commit/pull/2651/files?diff=split&w=1#diff-57a5ab580c429c008bdc02dd72a1bd25f92551500d17d6c2b88d0b91a2e51a9cR87)

For developers that utilize `asdf` to manage their golang versions, `pre-commit` does not find golang on their `PATH` because `pre-commit` purposely excludes any paths that are located in a `shims` directory.  ([Here is a link to a conversation when pre-commit implemented that choice](https://github.com/pre-commit/pre-commit/pull/1668#issuecomment-858626286)).

At the time of the creation of this PR we were on golang 1.19.3 and golangci-lint version 1.50.x. When pre-commit went to bootstrap golang for the pre-commit environment it pulled in golang version 1.20.x which is not compatible with our golangci-lint version. This mismatch in golang versions was causing a huge increase in memory usage. Here is what it was using on my machine.
image

Now that we have upgraded our golang version to 1.20 and upgraded the corresponding golangci-lint version, we won't see this memory problem. However, pre-commit version 3.x will still pull in the most up-to-date version of golang into the pre-commit environments unless we specify to use the golang version found on our system.

As noted in [the pre-commit documents](https://pre-commit.com/#golang), pre-commit now supports `language_version` with golang as of version 3.0.0. 

## Setup to Run Your Code

- Open up Activity Monitor and look at the memory tab
- run `pre-commit run -a golangci-lint`
- If you get an error about no GOPATH that means you haven't run the golangci-lint hook since the update of pre-commit to 3.0.x. You will need to run `pre-commit clean` to rebuild your pre-commit environment cache. The next time you run, it will rebuild each environment (Which will take a few minutes)
- Your run should not take very long and should not use much memory. (It looks like it will use about 2ish GBs of memory)


.pre-commit-config.yaml
```diff
@@ -53,6 +53,7 @@ repos:
     rev: v1.50.1
     hooks:
       - id: golangci-lint
+        language_version: 1.19.3
```
reggieriser, line 56: The [pre-commit docs](https://pre-commit.com/#golang) say: `language: golang also now supports language_version`.  Does that mean we should put a `language: golang` here too?  I see `language` on a few other hooks.

mmayo-truss: that is a good idea.

reggieriser (commented):
Thanks for tackling this!  Looks like we may have to upgrade everybody (and our version in CI) to pre-commit 3.x for this because this is what I get now with pre-commit 2.x when doing a `pre-commit install-hooks`:

```
An unexpected error has occurred: AssertionError: for now, pre-commit requires system-installed golang -- you selected `language_version: 1.19.3`
```

reggieriser:
I upgraded from 2.x to pre-commit 3.0.4.  The hooks installed fine and a `pre-commit run -a` ran without problems.  Didn't see crazy memory usage on `golangci-lint` and it finished in a reasonable amount of time.  So I think you've landed on the fix (good sleuthing!) but we may not be able to have both pre-commit 2 and 3 working at the same time.

mmayo-truss:
Before the pre_test check can pass, [this PR](https://github.com/transcom/circleci-docker/pull/375) needs to be merged to get our Circle CI docker image on pre-commit 3.0.4

felipe-lee:
**todo**: We pin the `pre-commit` version for `nix` users so that will also need to be update

https://github.com/transcom/mymove/blob/66683736d3c013b3339bc1c787f71e729bd098f9/nix/default.nix#L63-L71

mmayo-truss:
> **todo**: We pin the `pre-commit` version for `nix` users so that will also need to be updated
> 
> https://github.com/transcom/mymove/blob/66683736d3c013b3339bc1c787f71e729bd098f9/nix/default.nix#L63-L71
> 
> You can use https://ahobson.github.io/nix-package-search to find a package version

Thanks for this info @felipe-lee

mmayo-truss:
@felipe-lee It looks like nix does not currently have a package for pre-commit v3.0.4. There [is a PR here for it](https://github.com/NixOS/nixpkgs/pull/215429) that is in the process of getting reviewed. I will watch out for this and then reopen when this is available.

.pre-commit-config.yaml
```diff
@@ -1,3 +1,9 @@
+default_language_version:
+  golang: system
+
+# In order to pass the golang default version flag, pre-commit needs to be at least at version 3.0.0
+minimum_pre_commit_version: 3.0.0
```
ahobson, line 5: Any reason this shouldn't be `3.1.0` since that is the version we want?

mmayo-truss: @ahobson I was considering that. I went with this since it is the minimum version you would need to allow for the default language version flag to work for golang. I can switch it though.

afoon:
Removed myself and tagged @rogeruiz as the Moose reviewer after he volunteered at sync since an IS3 review is required.

rogeruiz (approved):
lgtm 🌈

.pre-commit-config.yaml
```diff
@@ -1,3 +1,9 @@
+default_language_version:
+  golang: system
```
ahobson, line 2: It seems like ideally, we'd also use the system node version, right? Is there any reason we shouldn't specify that, too?

mmayo-truss: @ahobson I can't think of a good reason why we wouldn't, however when I tested this locally, pre-commit would hang when trying to build the markdownlint-cli environment. I tried it twice with `node: system` added to the default language version flags, and both times it hung up.

nix/default.nix
```diff
@@ -60,15 +60,13 @@ buildEnv {
...
-        rev = "cd8d1784506a7c7eb0796772b73437e0b82fad57";
-      })
-      { }).pre-commit
+    (import (builtins.fetchGit {
+          # Descriptive name to make the store path easier to identify
+          name = "pre-commit-3.1.0";
+          url = "https://github.com/NixOS/nixpkgs/";
+          ref = "refs/heads/nixpkgs-unstable";
+          rev = "796b4a3c1d903c4b9270cd2548fe46f524eeb886";
+        }) {}).pre-commit
```
reggieriser, line 69: *nitpick:* The indentation and spacing is a little different here than the other `import` blocks.

reggieriser (approved):
I'm on `asdf` and this worked well for me.  I initially ran it when I had pre-commit 2.x and it correctly told me that I needed to update to at least 3.1 first.  After I grabbed 3.1.1 via `brew`, I did a `pre-commit clean && pre-commit install && pre-commit install-hooks` followed by a `pre-commit run -a` and all that ran fine and with normal duration/memory.

Approving pending the merging of https://github.com/transcom/circleci-docker/pull/383 and the final updating of hashes here.

ahobson:
We should remind `nix` users they need to run `pre-commit install` to re-install the git hooks so the new version is used

.pre-commit-config.yaml
ahobson: @mmayo-truss Huh. I just tried it with nix and it all works fine. I deeply distrust asdf and brew

PR #10079, opened 2023-02-13
MB-15226: Fix CircleCI playwright runner to respect path

ahobson:
## Summary

In circleci, we split up the playwright tests by admin, office, and my using a path parameter, but I made a thinko and left it off the command invocation so all runs in CI run all tests.

This fix only applies to CircleCI


rogeruiz (approved):
lgtm 🌈

kylehilltruss (approved):

PR #10080, opened 2023-02-13
MB-15227: fix playwright prime simulator

ahobson:
## Summary

Fix the playwright prime simulator tests in CI. They pass locally

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -681,6 +681,8 @@ commands:
             export HTTP_ORDERS_SERVER_NAME=orderslocal
             export HTTP_PRIME_SERVER_NAME=primelocal
             export ENVIRONMENT=test
+            # disable for speed, playwright tests can fail otherwise
+            export DB_DEBUG=false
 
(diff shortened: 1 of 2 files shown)


rogeruiz (approved):
lgtm 🌈

.circleci/config.yml
```diff
@@ -681,6 +681,8 @@ commands:
             export HTTP_ORDERS_SERVER_NAME=orderslocal
             export HTTP_PRIME_SERVER_NAME=primelocal
             export ENVIRONMENT=test
+            # disable for speed?
```
kylehilltruss, line 684: We should probably clarify this comment, or at least remove question tense ambiguity, for the sake of someone trying to understand it a year or so from now.

kylehilltruss (approved):
Change functionally looks good, aside from the comment below

.circleci/config.yml
ahobson: Thank you, PR updated

PR #10100, opened 2023-02-16
[ MB-14692, MB-14723 ] Removing the word Payload from Admin API Swagger definition

rogeruiz:
## Summary

- [x] Removing the word Payload from definitions
- [x] Adding the generated Swagger and Go files
- [x] Updating the handler and services;

## Setup to Run Your Code

Checkout this branch and run `make server_test` and see the tests passing.


felipe-lee (approved):
LGTM, thanks for fixing all of these!

PR #10102, opened 2023-02-16
[MB-15104] Update PPM Status after Reviewing Documents

pearl-truss:
## Summary

There is a new service: `SubmitReviewedDocuments` that updates the status of the ppmShipment when PPM documents are reviewed. 

It calls the new method `SubmitReviewedDocuments` that is a part of the ppm shipment router. This method checks if 
there is at least one rejected ppm document it will set the status to WAITING_ON_CUSTOMER or PAYMENT_APPROVED respectively.

## Open Question(s)
- Prior to changing the PPM shipment status to either `WAITING_ON_CUSTOMER` or `PAYMENT_APPROVED` is there a status the PPM should be in in order to transition to either of those statuses without throwing an error?


NamibiaTorres (commented):
I wanted to answer your open question regarding statuses. There are 2 potential statuses that the ppm_shipment  _status can be set to. `'NEEDS_PAYMENT_APPROVAL'` and `'DRAFT'` seem like appropriate options for that limbo state.

pkg/services/ppmshipment/ppm_shipment_router_test.go
```diff
@@ -395,3 +395,98 @@ func (suite *PPMShipmentSuite) TestSubmitCloseOutDocumentation() {
...
+			},
+		})
+		movingExpense := testdatagen.MakeDefaultMovingExpense(suite.DB())
+		weightTicket := testdatagen.MakeDefaultWeightTicket(suite.DB())
+
+		ppmShipmentRouter := setUpPPMShipmentRouter(mtoShipmentRouterMethodToMock, nil)
+		docs := models.PPMDocuments{
+			WeightTickets:   models.WeightTickets{weightTicket},
+			MovingExpenses:  models.MovingExpenses{movingExpense},
+			ProgearExpenses: models.ProgearWeightTickets{progear},
```
NamibiaTorres, line 441: **Suggestion**: Instead of ProgearExpenses I think we should rename this ProgearWeightTickets because that's the term we use in the services and handler.

pkg/services/ppmshipment/ppm_shipment_review_documents.go
```diff
@@ -0,0 +1,83 @@
...
+	ppmShipment, err := FindPPMShipment(appCtx, ppmShipmentID)
+
+	if err != nil {
+		return nil, err
+	}
+
+	ppmDocuments, err := p.PPMDocumentFetcher.GetPPMDocuments(appCtx, ppmShipment.ShipmentID)
+	if err != nil {
+		return nil, err
+	}
```
felipe-lee, line 45: **suggestion (blocking)**: I think we can skip this call because the `FindPPMShipment` function above [fetches all the associations for a PPM shipment, including getting all the documents](https://github.com/transcom/mymove/blob/8c945f0e73af9ab9ea07ef943e388ca4cba4c193/pkg/services/ppmshipment/ppm_shipment_fetcher.go#L71-L93). This would also simplify the signature of the `SubmitReviewedDocuments` function on the PPM shipment router since it wouldn't need to accept the PPM documents separately from the `ppmShipment` itself. 

Let me know if I'm missing something and you think we should keep it the way it is.

pkg/services/ppmshipment/ppm_shipment_review_documents_test.go
```diff
@@ -0,0 +1,76 @@
...
+			nonexistentPPMShipmentID,
+		)
+
+		if suite.Error(err) {
+			suite.Nil(updatedPPMShipment)
+
+			suite.IsType(apperror.NotFoundError{}, err)
+			suite.Contains(err.Error(), "not found while looking for PPMShipment")
+		}
+	})
```
felipe-lee, line 75: **suggestion**: Can you add test cases for when:

* the router func (`PPMShipmentRouter.SubmitReviewedDocuments`) returns an error
* routing is successful

pkg/services/ppmshipment/ppm_shipment_router.go
```diff
@@ -115,3 +115,41 @@ func (p *ppmShipmentRouter) SubmitCloseOutDocumentation(_ appcontext.AppContext,
 
 	return nil
 }
+
+// SubmitReviewedDocuments sets the PPM shipment status to the PAYMENT_APPROVED if all docs approved otherwise WAITING_ON_CUSTOMER
+func (p *ppmShipmentRouter) SubmitReviewedDocuments(_ appcontext.AppContext, ppmShipment *models.PPMShipment, ppmDocuments models.PPMDocuments) error {
+	// IS there a status the PPM should be in before proceeding
```
felipe-lee, line 121: **thought**: I think the only status that would be valid as of right now would be `NEEDS_PAYMENT_APPROVAL`. Something we'll need clarification on in the future is how we get back from `WAITING_ON_CUSTOMER` since that is going to be handled outside the system, but I think for now we can make this function work on the assumption that shipments should be in the `NEEDS_PAYMENT_APPROVAL` status when a SC is reviewing the PPM docs.

pkg/services/ppmshipment/ppm_shipment_router_test.go
felipe-lee: **thought**: I definitely agree we should have uniform naming across things, but this discrepancy pre-dates this PR. I think a ticket for cleaning this up would be good because the problem is that this isn't the only object that doesn't have consistent naming for pro gear stuff. I think there are like 3 or 4 variations on the name 😞

pkg/services/ppmshipment/ppm_shipment_router.go
```diff
@@ -115,3 +115,41 @@ func (p *ppmShipmentRouter) SubmitCloseOutDocumentation(_ appcontext.AppContext,
...
+		}
+	}
+
+	if len(ppmDocuments.MovingExpenses) >= 1 {
+		for _, movingExpenses := range ppmDocuments.MovingExpenses {
+			if movingExpenses.Status != nil && *movingExpenses.Status == models.PPMDocumentStatusRejected {
+				numOfRejectedDocs++
+			}
+		}
+	}
```
felipe-lee, line 146: **suggestion (non-blocking)**: I think these are fine considering we likely won't have many document sets for a given shipment, but I think it would be a bit more efficient if we broke out of the for loops as soon as we find a single one that has been rejected. The var could be a `bool` and that could also be used to skip the other loops too if we've already found a rejected one because there is no need to loop through the other document sets if we know a previous set was rejected.

felipe-lee (commented):
Hey, this is looking good. I left a few suggestions and comments, but I think this is pretty close to ready.

pkg/services/ppmshipment/ppm_shipment_router.go
pearl-truss: makes sense and changes have been pushed

pkg/services/ppmshipment/ppm_shipment_router_test.go
NamibiaTorres: Thanks Pearl! That's great and someone can pick it up during Sprint 0 since it cleans up older code.

NamibiaTorres (approved):
It looks like you've addressed all comments. Thank you!

pkg/services/ppmshipment/ppm_shipment_router.go
```diff
@@ -117,35 +117,43 @@ func (p *ppmShipmentRouter) SubmitCloseOutDocumentation(_ appcontext.AppContext,
...
+			),
+		)
+	}
+
+	hasRejectedDocuments := false
+	if len(ppmShipment.WeightTickets) >= 1 {
+		for _, weightTicket := range ppmShipment.WeightTickets {
 			if weightTicket.Status != nil && *weightTicket.Status == models.PPMDocumentStatusRejected {
-				numOfRejectedDocs++
+				hasRejectedDocuments = true
```
felipe-lee, line 135: **suggestion (non-blocking)**: One other thing that could be done after setting this to `true` is using the `break` statement, but again, not a huge gain in this case, especially since we do skip entire loops now.

https://go.dev/ref/spec#Break_statements

felipe-lee (approved):
Awesome, thank you for making those updates! I left one more suggestion, but this lgtm!

pkg/services/ppmshipment/ppm_shipment_router.go
pearl-truss: thanks for link, I made this change and will merge when CI clears

PR #10103, opened 2023-02-16
[GHC Prime Conference] :memo: Clear up confusing documentation

kctruss:
This came up at the GHC Prime onsite. There is a confusing line under updateMTOPostCounselingInformation that indicates the endpoint should be used to update PPMs. ~~This is wrong. Removing to stop the prime from attempting to use this endpoint incorrectly.~~ Added clarifying language to enable the prime to use this endpoint as intended without trying to guess at how it should work.


ronaktruss:
It looks like this end point could be used for updating ppm shipment information based on [this line](https://github.com/transcom/mymove/blob/main/pkg/services/move_task_order/move_task_order_updater.go#L342). If that's not the intention we should probably remove this logic in addition to removing this documentation.

kctruss:
> It looks like this end point could be used for updating ppm shipment information based on [this line](https://github.com/transcom/mymove/blob/main/pkg/services/move_task_order/move_task_order_updater.go#L342). If that's not the intention we should probably remove this logic in addition to removing this documentation.

This needs to be understood and documented in greater detail. For the short term, we've been asked to just remove this line because as of right now the Prime keeps trying to use this endpoint to update PPMs and they are getting an error about services counseling. I'm pulling this line out now for the fast win, and Jules and I will follow up on a task to understand and update with more correct information.

swagger-def/prime.yaml
```diff
@@ -157,8 +157,6 @@ paths:
         ### Functionality
         This endpoint **updates** the MoveTaskOrder after the Prime has completed Counseling.
```
ronaktruss, line 159: Nit: Can we also get rid of this line?

ronaktruss (approved):
Approving under the assumption there will a ticket/task created to track the work of looking into this API a bit more. Once those are created can you link the tickets in slack?

kctruss:
Update: 
Lint errors + PTO = more time to investigate the actual intention of the endpoint in question. Rather than remove the line about PPMs, I've called out a bit more expected behavior based on the existing code and experience. This should meet the request of the client to clear up confusion the prime is having around how to use this endpoint.

@ronaktruss when you get time can you re-review these changes?

swagger-def/prime.yaml
```diff
@@ -157,8 +157,8 @@ paths:
...
 
-        PPM related information is updated here. Most other fields will be found on the specific MTOShipment and updated using [updateMTOShipment](#operation/updateMTOShipment).
-
+        PPM related information may also be passed in with the counseling information here. **NOTE:** PPM information cannot be updated from this endpoint
+        after counseling information is updated. Most other fields will be found on the specific MTOShipment and updated using [updateMTOShipment](#operation/updateMTOShipment).
```
ronaktruss, line 161: I just looked over the code again and it looks like we are basically just taking in the mtoID and updating the associated move and subsequent child objects based on that. 

Would this be clearer if we just indicate that we don't expect anything in the body of this request?

kctruss: 🤦 Yes. This is ridiculous. I'll get that change pushed. 

So much for "Just make this change real quick" 🙄

kctruss: @ronaktruss let me know if the latest change reads right to you.

swagger-def/prime.yaml
```diff
@@ -155,10 +155,9 @@ paths:
       summary: updateMTOPostCounselingInformation
       description: |
         ### Functionality
-        This endpoint **updates** the MoveTaskOrder after the Prime has completed Counseling.
-
-        PPM related information may also be passed in with the counseling information here. **NOTE:** PPM information cannot be updated from this endpoint
-        after counseling information is updated. Most other fields will be found on the specific MTOShipment and updated using [updateMTOShipment](#operation/updateMTOShipment).
+        This endpoint **updates** the MoveTaskOrder to indicate that the Prime has completed Counseling.
+        This update uses the moveTaskOrderID provided in the path, updates the move status and marks child elements of the move to indicate the update.
```
ronaktruss, line 159: Nit: Personally, `marks child elements of the move to indicate the update` accurately describes what happens, but is vague enough that I would have a few more follow up questions if I was reading this as a consumer of the API. I don't have suggestions for this, but I think should clear up the main confusion about ppm information.

ronaktruss (approved):
Left a nit that's purely based on my opinion. It's not a blocker.

PR #10111, opened 2023-02-17
[ MB-14692, MB-14723 ] Going from snakecase to kebobcase

rogeruiz:
## Summary

This involves going from underscores to dashes, or snakecase to kebobcase, in
our API paths. 

## Setup to Run Your Code

Run `lint-yaml-with-spectral swagger/admin.yaml` and compare it to the last
successful run on the default branch. The number of errors/warnings should be
going down.


kctruss (approved):
🌠

PR #10113, opened 2023-02-17
[MB-15113] Creates a script to automatically create p7b files

chtakahashi:
Somehow broke my previous branch. trying again

Usage:
```sh
./scripts/generate-p7b-file <OUTPUT/FILEPATH>.p7b cert1.pem cert2.der.p7b cert3.cer cert4.crt
```

## Background
our previous usage of p7b files was to just use the `config/tls/Certificates_PKCS7_v5.9_DoD.der.p7b`. When allowing prime API access, we needed to add additional root/intermediate certificates to this file. The purpose of this script is to prevent user error packaging the script together, as a faulty p7b file can break the deploy pipeline.


jesset-truss (approved):
Nice work!

PR #10121, opened 2023-02-21
[MB-15290] Refactor Playwright tests for customer role

kylehilltruss:
## Summary

During the effort to migrate our integration tests from Cypress to Playwright, we used a semi-automated process to translate much of the existing code that was written for the former so that it could quickly work in the latter. That directly translated code contains some issues that we want to address in a refactor so we can align our code with Playwright best practices:

* It makes heavy use of page.locator instead of page.getBy selectors, which can make the code more fragile if the underlying HTML changes,
* It doesn’t use affirmative verification of page state and readiness, which introduces a higher potential for future race conditions and test flakiness,
* It contains some integration tests which have been manually skipped due to persistent flakiness when run in Cypress, and some tests which verify functionality that no longer exists in the application.

This pull request refactors any translated code in the `playwright/tests/my/mymove` directory, in order to bring it into alignment with Playwright best practices.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the client locally.

```sh
make client_run
```

##### Terminal 2

Start the server locally.

```sh
make server_run
```

##### Terminal 3

Run the playwright tests modified by this pull request.

```sh
npx playwright run playwright/tests/my/mymove
```

</details>


playwright/tests/my/mymove/orders.spec.js
```diff
@@ -1,42 +1,23 @@
...
-
-  await page.locator('input[name="issue_date"]').click();
-  await page.locator('input[name="issue_date"]').type('6/2/2018');
-  await page.locator('input[name="issue_date"]').blur();
+  await customerPage.waitForPage.ordersDetails();
 
-  await page.locator('input[name="report_by_date"]').type('8/9/2018');
-  await page.locator('input[name="report_by_date"]').blur();
+  // Fill in orders details
+  await page.getByTestId('dropdown').selectOption('PERMANENT_CHANGE_OF_STATION');
```
ahobson, line 16: Honest question: should we prefer `.selectOption({label: 'foo'})`? I wonder if using the browser visible option would be more in line with our testing philosophy.

playwright/tests/my/mymove/orders.spec.js
```diff
@@ -48,24 +29,24 @@ test('orders entry will accept orders information', async ({ page, customerPage
   //
   await page.locator('div:has(label:has-text("Are dependents")) >> div.usa-radio').getByText('No').click();
 
-  // Choosing same current and destination duty location should block you from progressing and give an error
+  // Verify that a user can't use the same duty location
   await customerPage.selectDutyLocation('Yuma AFB', 'new_duty_location');
```
ahobson, line 33: I'd love to find a better way than using the class selector here, but I'm going to vote that is out of scope for your PR

playwright/tests/utils/baseTest.js
```diff
@@ -71,10 +71,10 @@ export class BaseTestPage {
    */
   async uploadFileViaFilepond(locator, relativeFilePath) {
     const filePath = path.join('playwright/fixtures', relativeFilePath);
-    const chooser = locator.getByText('choose from folder');
-    await expect(chooser).toBeVisible();
+    const actionLink = locator.locator('.filepond--label-action');
+    await expect(actionLink).toBeVisible();
     const fileChooserPromise = this.page.waitForEvent('filechooser');
-    await chooser.click();
+    await actionLink.click();
```
ahobson, line 77: Can you help me understand this change? I was trying to avoid using `locator`

ahobson (approved):
This is awesome, awesome work

playwright/tests/utils/baseTest.js
kylehilltruss: Yeah -- annoyingly, the text that was previously being selected isn't always the text in the blue action link for our filepond instances. (The Add and Edit Orders pages both use different text for that link, for example.) The rest of this function makes sense to generalize and re-use, though, which we're doing in the Orders test.

Since that link text also doesn't seem to have an ARIA role associated with it, which would open up a `getByRole`, it seemed easiest in this case to find the action link via its class with `locator`. That class is associated with the third-party component, so it feels to me it like it should be more stable.

playwright/tests/my/mymove/orders.spec.js
kylehilltruss: Oof, I didn't even look at the internals here, I just migrated over what you'd done in the first pass for this line. I definitely agree that using a getBy selector would be better, but yeah, it's out-of-scope at the moment.

playwright/tests/my/mymove/orders.spec.js
kylehilltruss: That's a better idea -- I just copied over what `codegen` showed me for this one.

playwright/tests/utils/baseTest.js
ahobson: That totally makes sense! How would you feel about adding something like that as a comment to that helper so other folks follow along (and understand it's not a pattern to follow) ?

PR #10128, opened 2023-02-22
MB-15285: Make iampostgrest_test more robust to race conditions

ahobson:
## Summary

The iampostgres module uses goroutines and so we have to be very careful when writing tests to deal with the unpredictability of which code will run first.

## Setup to Run Your Code

<details>

##### Terminal 1

Run the test locally.

```sh
repeat 20 go test -count 1 ./pkg/iampostgres/...
```

</details>

### DIFF
--- a/pkg/iampostgres/iampostgres_test.go
+++ b/pkg/iampostgres/iampostgres_test.go
@@ -50,6 +50,8 @@ func TestGetCurrentPassword(t *testing.T) {
 	assert := assert.New(t)
 
 	rdsu := RDSUTest{}
+	// We need to put a blank token first to avoid race conditions.
+	// See more below
 	rdsu.passes = append(rdsu.passes, "")
 	rdsu.passes = append(rdsu.passes, "abc")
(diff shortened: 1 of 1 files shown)


reggieriser (approved):
I ran the test 500 times and it never failed. 🎉  I'm approving, but do you mind adding some context as to why adding an empty string password to the front of the `passes` slice helps to fix the issue?  I think I follow, but want to make sure given the async code here.

ahobson:
> I ran the test 500 times and it never failed. 🎉 I'm approving, but do you mind adding some context as to why adding an empty string password to the front of the `passes` slice helps to fix the issue? I think I follow, but want to make sure given the async code here.

Ok, I added some comments; please let me know if they are confusing or don't help explain the problem. Otherwise I'll plan on merging later today

reggieriser:
> Ok, I added some comments; please let me know if they are confusing or don't help explain the problem. Otherwise I'll plan on merging later today

Comments look great and make sense!  Thanks for adding them and tackling this flaky test.

PR #10136, opened 2023-02-23
[ MB-14692, MB-14723 ] Updating consumes and produces all path stanzas

rogeruiz:
## [Jira ticket](https://dp3.atlassian.net/browse/MB-14723) for this change

Originally, I thought that the global consumes and produces stanzas
would be enough. Some of our APIs though don't consume JSON as they
consume form data. So even though it's very repetitive, it leads to much
better documentation.


kctruss (approved):
🌈

PR #10062, opened 2023-02-10
[ MB-14692, MB-14723 ] Duplicating tags at the root stanza

rogeruiz:
## Summary

This fixes the `operation-tag-defined` error.

## Setup to Run Your Code

I ran `make swagger_generate server_generate` on this branch to get all the necessary files checked into the repository. 

As the reviewer, you can run `lint-yaml-with-spectral` from inside this repository and with the `swagger/admin.yaml` and should see that there are only **✖ 104 problems (42 errors, 62 warnings, 0 infos, 0 hints)** which is different from the default branch of **✖ 130 problems (42 errors, 88 warnings, 0 infos, 0 hints)**

PR #10146, opened 2023-02-24
[ MB-14692, MB-14723 ] Swagger Admin tags refactored

rogeruiz:
## Summary

- [x] Renaming tag for "Admin users"
- [x] Renaming tag for "Electronic orders"
- [x] Renaming tag for "Moves"
- [x] Renaming tag for "Notifications"
- [x] Renaming tag for "Transportation offices"
- [x] Renaming tag for "Office users"
- [x] Renaming tag for "Organizations"
- [x] Renaming tag for "TSPPs"
- [x] Renaming tag for "Uploads"
- [x] Renaming tag for "Users"
- [x] Renaming tag for "Webhook subscriptions"

This work was done in #10062 but did not quite work on the API page for Docusaurus due to the tags having snake-case versus kabob-case because of what their names were.


rogeruiz:
Keeping this in draft for now as it's a lot of mechnical work to rename things in our `pkg/handlers/adminapi` files. I'll have to also check the changes across the `src/` files as well.

PR #10148, opened 2023-02-24
[ on-call ] Sorting service items to fix dev seed flakiness

mmayo-truss:
## Summary

This PR provides a fix for the intermittent error that occurs when trying to populate dev seed data. 
Here is a link to the error in CircleCI: https://app.circleci.com/pipelines/github/transcom/mymove/50340/workflows/5d636e0f-d0fd-4ee9-8593-f85132245cba/jobs/829841

This error also occurs on the local machine as you can see in this screenshot:
image

The error occurs when the DDFSIT service item appears before the DDASIT service item in the Payment Service Items array. This causes the DDFSIT service item to already be attached to the MTO shipment when doing a validation check on the DDASIT service item, which throws the error. 

The fix involves reshuffling the order in which the service items were approved and then sorting them by approval date. 

## Setup to Run Your Code
run `make db_dev_e2e_populate` and  `make db_dev_fresh`

verify that the move with move code `PARAMS` is functioning as you would expect.


pkg/payment_request/service_param_value_lookups/number_days_sit_lookup.go
```diff
@@ -161,7 +161,7 @@ func calculateNumberSITAdditionalDays(paymentServiceItem models.PaymentServiceIt
 	return int(days), nil
 }
 
-func fetchAndVerifyMTOShipmentSITDates(mtoShipmentSITPaymentServiceItems models.PaymentServiceItems, mtoServiceItem models.MTOServiceItem) (time.Time, time.Time, error) {
+func fetchAndVerifyMTOShipmentSITDates(appCtx appcontext.AppContext, mtoShipmentSITPaymentServiceItems models.PaymentServiceItems, mtoServiceItem models.MTOServiceItem) (time.Time, time.Time, error) {
```
felipe-lee, line 164: **praise**: Always great to pass this more places.

felipe-lee (approved):
LGTM! I saw you run this several times locally without error and everything seems good to me. Hopefully if this does pop up again, we'll have the helpful logs you've added for debugging.

PR #10150, opened 2023-02-24
MB-15283:deleted shipments are not considered for the TOO queue

ronaktruss:
## Summary

This make it so that the number of shipments is correct and when filtering for a specific date deleted shipments are ignored. (eg. If a shipment with a requested move date of 15 Mar 2020 is deleted if you search using that date it will still appear in the search. With this change you should no longer see that happening).

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Create a new user
2. Create some shipments with varying information. Before submitting delete that some of the shipments
3. Ensure that we are not seeing information corresponding to the deleted shipment.

Alternatively you can just manually add a deleted_at timestamp to a shipment.

You should compare and contrast this behavior with the current behavior to ensure this is working as expected.


kctruss (approved):
Looks great :shipit:

PR #10171, opened 2023-02-28
[ on-call ] Removing old environment constants

rogeruiz:
## Summary

There were references to `prod` and `staging` environments which were 4 years
old in the codebase. I removed all references to these variables and ran the
server tests locally to ensure these old variables were not being used anymore.
The new names for production and staging are `prd` and `stg` respectively.


ahobson (approved):
Thank you! Excellent idea!

rogeruiz:
Not having this patch caused a failed deploy in #10168

PR #10182, opened 2023-03-01
MB-14619 factory.BuildServiceMember added as new version of testdatagen.MakeServiceMember

StuTruss:
## Summary

This ticket adds a factory package with function BuildServiceMember intended to be an eventual replacement for testdatagen.MakeServiceMember
* Created BuildServiceMember and shared functions
* Created tests for BuildServiceMember. 
* Replaces instances in code where `MakeServiceMember` and `MakeDefaultServiceMember` exists with `BuildServiceMember`

## Setup to Run Your Code

### Frontend

- No frontend changes

### Backend

- No schema changes

### DIFF
--- a/pkg/factory/service_member_factory.go
+++ b/pkg/factory/service_member_factory.go
@@ -24,11 +24,29 @@ func BuildServiceMember(db *pop.Connection, customs []Customization, traits []Tr
 		}
 	}
 
+	// Find/create the ResidentialAddress
+	tempResAddressCustoms := customs
+	result := findValidCustomization(customs, Addresses.ResidentialAddress)
+	if result != nil {
+		tempResAddressCustoms = convertCustomizationInList(tempResAddressCustoms, Addresses.ResidentialAddress, Address)
+	}
+
+	resAddress := BuildAddress(db, tempResAddressCustoms, nil)
+
+	// Find/create the BackupMailingAddress
+	tempBackupAddressCustoms := customs
+	backupAddressResult := findValidCustomization(customs, Addresses.BackupMailingAddress)
+	if backupAddressResult != nil {
+		tempBackupAddressCustoms = convertCustomizationInList(tempBackupAddressCustoms, Addresses.BackupMailingAddress, Address)
+	}
+
+	backupAddress := BuildAddress(db, tempBackupAddressCustoms, nil)
+
 	// Find/create the user model
 	user := BuildUser(db, customs, traits)
 
-	// Find/create the address model
-	currentAddress := BuildAddress(db, customs, traits)
+	// Find/create the dutyLocation model
+	dutyLocation := BuildDutyLocation(db, customs, traits)
 
 	email := "leo_spaceman_sm@example.com"
 	agency := models.AffiliationARMY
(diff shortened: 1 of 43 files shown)


pkg/factory/service_member_factory_test.go
```diff
@@ -0,0 +1,154 @@
...
+		// CALL FUNCTION UNDER TEST
+		serviceMember := BuildServiceMember(suite.DB(), nil, nil)
+
+		// VALIDATE RESULTS
+		suite.Equal(defaultServiceMember.FirstName, serviceMember.FirstName)
+		suite.Equal(defaultServiceMember.LastName, serviceMember.LastName)
+		suite.Equal(defaultServiceMember.Rank, serviceMember.Rank)
+		suite.Equal(defaultServiceMember.PersonalEmail, serviceMember.PersonalEmail)
+		suite.Equal(defaultServiceMember.Affiliation, serviceMember.Affiliation)
+
```
shimonacarvalho, line 47: I think you missed telephone here?

pkg/factory/service_member_factory_test.go
```diff
@@ -0,0 +1,154 @@
...
+		// Set up:          Create a Service Member and pass custom fields
+		// Expected outcome:serviceMember should be created with custom fields
+		// SETUP
+		customServiceMember := models.ServiceMember{
+			FirstName:          models.StringPointer("Gregory"),
+			LastName:           models.StringPointer("Van der Heide"),
+			Telephone:          models.StringPointer("999-999-9999"),
+			SecondaryTelephone: models.StringPointer("123-555-9999"),
+			PersonalEmail:      models.StringPointer("peyton@example.com"),
+		}
```
shimonacarvalho, line 67: Try to customize all the things - edipi, affiliation, rank etc... because we want to make sure nothing slips customization.

pkg/factory/service_member_factory_test.go
```diff
@@ -0,0 +1,154 @@
...
+		suite.Equal(customServiceMember.Telephone, serviceMember.Telephone)
+		suite.Equal(customServiceMember.SecondaryTelephone, serviceMember.SecondaryTelephone)
+		suite.Equal(customServiceMember.PersonalEmail, serviceMember.PersonalEmail)
+
+		// Check that address was customized
+		suite.Equal(customAddress.StreetAddress1, serviceMember.ResidentialAddress.StreetAddress1)
+
+		// Check that user was customized
+		suite.Equal(customUser.LoginGovEmail, serviceMember.User.LoginGovEmail)
+	})
```
shimonacarvalho, line 96: Nice job checking the associated objects.

shimonacarvalho:
Looks good, I'll give it another peek after you do the replacements.

pkg/handlers/internalapi/mto_shipment_test.go
```diff
@@ -1634,8 +1634,30 @@ func (suite *HandlerSuite) TestDeleteShipmentHandler() {
 	})
 
 	suite.Run("Returns 403 when servicemember ID doesn't match shipment", func() {
-		sm1 := testdatagen.MakeStubbedServiceMember(suite.DB())
-		sm2 := testdatagen.MakeStubbedServiceMember(suite.DB())
+		sm1 := factory.BuildServiceMember(nil, []factory.Customization{
```
shimonacarvalho, line 1637: Hmm this seems like quite a few lines of code compared to the original. 
This could be a trait (the ids part)? Just make a single trait with both IDs set and use that?

pkg/factory/service_member_factory.go
```diff
@@ -0,0 +1,77 @@
...
+
+	if db != nil {
+		mustCreate(db, &serviceMember)
+	}
+
+	return serviceMember
+}
+
+// GetTraitStubbedServiceMember is a sample GetTraitFunc
+func GetTraitStubbedServiceMember() []Customization {
```
shimonacarvalho, line 64: I wouldn't call it "StubbedServiceMember" because selecting this trait doesn't make it stubbed or not. So it could be confusing. Maybe "ServiceMemberSetIDs" would work. That's more accurate. Also the comment description should be updated.

shimonacarvalho:
There are a bunch of hits in `pkg/testdatagen/scenario` . We do want to update those. After you do, make sure to run the yarn playwright tests to ensure nothing was broken. There are a lot, if needed consider breaking out that ticket. Decide with the team if so.

pkg/factory/service_member_factory.go
```diff
@@ -0,0 +1,100 @@
...
+	if db != nil {
+		mustCreate(db, &serviceMember)
+	}
+
+	return serviceMember
+}
+
+// ServiceMemberSetIDs is a sample GetTraitFunc
+// that sets ids for both ServiceMember and User models
+func ServiceMemberSetIDs() []Customization {
```
ahobson, line 87: I think I remember a guideline in the Backend Factory Design doc that all traits should start with `GetTrait` for discoverability. Do you agree? If so, should this be changed?

pkg/factory/shared.go
```diff
@@ -346,7 +348,7 @@ func findValidCustomization(customs []Customization, customType CustomType) *Cus
 	}
 
 	// Else check that the customization is valid
-	if err := checkNestedModels(*custom); err != nil {
+	if err := checkNestedModels(*custom); err != nil && !custom.LinkOnly {
 		log.Panic(fmt.Errorf("Errors encountered in customization for %s: %w", *custom.Type, err))
```
ahobson, line 352: If `custom.LinkOnly` is true but `err` is nil, then the error message in the panic won't really explain what is going on, right? Doesn't that mean the check for `custom.LinkOnly` should be separate so we can supply an error message specific to it?

pkg/factory/service_member_factory.go
StuTruss: you're correct here, I'll change this now

pkg/factory/shared.go
StuTruss: It's my understanding that the log message won't run unless `err` is not nil and not a `LinkOnly` customization. having said that I do believe I should make the check for the `LinkOnly` separate because there are multiple errors to check for in the `checkNestedModels` and we only want to deal with certain ones with the `custom.LinkOnly` check

ahobson (approved):
🚀

PR #10180, opened 2023-03-01
Revert "[MB-15297] Upgrade to go 1.20.1"

reggieriser:
Reverts transcom/mymove#10169

Trying to see if it fixes our current CircleCI problems in the `deploy_stg_app` and `deploy_stg_app_client` steps.

PR #10169, opened 2023-02-28
[MB-15297] Upgrade to go 1.20.1

reggieriser:
## Summary

This PR updates the project from Go 1.19.3 to Go 1.20.1 (which includes security fixes and some possible options related to FIPS 140-2 compliance).

Go 1.20.x version history:
https://golang.org/doc/devel/release.html#go1.20

Go 1.19.x version history:
https://golang.org/doc/devel/release.html#go1.19

Here's the [corresponding PR from circleci-docker](https://github.com/transcom/circleci-docker/pull/381) for the docker image hash referenced in this PR.  Please verify that it seems correct and that all relevant references have been updated.

~~**Note:** After this PR has been initially vetted, I will merge in the circleci-docker PR and update this PR with the new hash.~~ DONE!

I generally followed [these upgrade instructions](https://transcom.github.io/mymove-docs/docs/backend/guides/how-to/upgrade-go-version).

Some issues that were discovered by the new golangci-lint were already fixed in [this earlier merged PR](https://github.com/transcom/mymove/pull/10034).

## Setup to Run Your Code

<details>
<summary>💻 You will need to use three separate terminals to test this locally.</summary>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

If you're on `asdf`:
- `asdf install`
- `asdf global golang 1.20.1`

If you're on `nix`:
- run `./nix/update.sh`

Just verify that all tests still pass and that your normal workflow seems to be working fine.  Let me know if you're using `asdf` or `nix` -- I'd like to have at least one reviewer from each.

## Verification Steps for Author

These are to be checked by the author.

- [ ] Tested in the Experimental environment (for changes to containers, app startup, or connection to data stores)
- [ ] Request review from a member of a different team.
- [ ] Have the Jira acceptance criteria been met for this change?

## Verification Steps for Reviewers

These are to be checked by a reviewer.

### Backend

- [ ] Code follows the guidelines for [Logging](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide#logging)
- [ ] The requirements listed in [Querying the Database Safely](https://transcom.github.io/mymove-docs/docs/dev/contributing/backend/Backend-Programming-Guide/#querying-the-database-safely) have been satisfied.

PR #10185, opened 2023-03-02
[MB-15297] Update to go 1.20.1 again but with old base image

reggieriser:
We had previously [tried to update to go 1.20.1](https://github.com/transcom/mymove/pull/10169) but ran into failures in a couple of deployment steps in CircleCI.  See this [Slack thread](https://ustcdp3.slack.com/archives/CP4U2NKRT/p1677706828317889) for lots of context.  So we [reverted](https://github.com/transcom/mymove/pull/10180) to get things flowing again.

After further investigation, it appears that the go upgrade may not have been the root cause.  Instead we were pulling in a new `circleci-docker` base image that was merged last week.  That image had a new version of Ubuntu among other things.  For some reason, it seems to cause issues with the deployment steps, based on some testing in experimental (details in the above Slack thread).  So I rolled back to the previous base image, but kept the Go 1.20.1 upgrade in place, and am using that hash in this PR.  That way, we can still move forward with Go 1.20.1 while we reassess the base image.

Note that once I get approvals from `asdf` and `nix` users on this PR, I plan to update the hashes once more to the `main` hash from the `circleci-docker` repo after the [revert PR](https://github.com/transcom/circleci-docker/pull/385) there is merged.


.circleci/config.yml
```diff
@@ -810,7 +810,7 @@ jobs:
           name: Install go-swagger
           command: make bin/swagger
       - save_cache:
-          key: go-mod-sources-v4-{{ checksum "go.sum" }}-{{ checksum ".go-version" }}
+          key: go-mod-sources-v5-{{ checksum "go.sum" }}-{{ checksum ".go-version" }}
```
reggieriser, line 813: Note that I did have to reset the go-mod-sources cache key in CircleCI to work around [some glibc library problems](https://app.circleci.com/pipelines/github/transcom/mymove/50554/workflows/f304acc8-bc1e-4e3a-87bb-1179ee8aef9c/jobs/834097) running the swagger binary via pre-commit, likely due to the cached version being compiled with the newer version of Ubuntu that was in the newer base image we end up deploying yesterday.

NamibiaTorres (approved):

pearl-truss (approved):

felipe-lee (approved):
Lgtm (as a `nix` user)! I ran `make server_test` and `make server_run` (with `make client_run`) and things seem to work without issues.

rogeruiz (approved):
lgtm 🌈

PR #10187, opened 2023-03-02
[ MB-15049 ] Modifying the query to include weight tickets on the ppmShipment object

mmayo-truss:
## Summary

Modifying the query to include the weight tickets on the ppmShipment object instead of returning a separate array of Documents.


afoon (approved):
worked as expected in the huddle

michaelvaldes-truss (approved):
lgtm!

PR #10204, opened 2023-03-06
MB-8032 Add min value validation to prime estimated and actual weights in shipment create/update

PDickeyTruss:
## Summary

This PR sets the minimum value for both the prime estimated and actual weight fields in the create/update MTO shipment endpoints. The result of this is that negative values in the weigh fields cause a validation error and prevent updating/creation, while positive values pass validation and update/create successfully.

## Outstanding issue
~~It looks like all works as expected except when the weight value is set to zero. This seems to somehow get past the minimum value validation but is ignored in the update/create.  Seems like a bug with the `minimum` functionality when it's value is zero?~~
This should be fixed now! 

## Testing
Use postman (or something similar) to make a requests to the prime API.
We want to test both the create and update endpoints.
- Use `listMoves (GET /moves)` to find the ID for a move that is available to the prime
- Use `createMTOShipment (POST /mto-shipments)` to add a shipment to that move. Use a few different estimated weight values, and make sure that any weights that are <1 are rejected.
<details>
<summary>createMTOShipment Payload</summary>

```
{
    "moveTaskOrderID": "8fdf2e13-65c6-4e89-973f-5ae7156b1944",
    "primeEstimatedWeight": 0,
    "requestedPickupDate": "1965-04-01",
    "pickupAddress": {
        "streetAddress1": "123 Main Ave",
        "city": "Anytown",
        "state": "VA",
        "postalCode": "90210",
        "streetAddress2": "Apartment 9000",
        "streetAddress3": "Montmârtre",
        "eTag": "laborum exe",
        "country": "USA"
    },
    "shipmentType": "HHG",
    "customerRemarks": "handle with care",
    "diversion": false,
    "pointOfContact": "laboris sint fugiat"
}
```

</details>

- Grab the ID and eTag of a shipment out of the response to one of your createMTOShipment requests.
- Use that information to update that shipment with `updateMTOShipment (PATCH /mto-shipments/<mtoShipmentID>)`. Put the etag value in the If-Match header. Every time you run this request, you have to get the new etag from the shipment from the response, and stick it in your If-Match header the next time you run it. Use a few different estimated weight values, and make sure that any weights that are <1 are rejected.
<details>
<summary>updateMTOShipment Payload</summary>

```
{
    "scheduledPickupDate": "2023-03-14",
    "primeEstimatedWeight": 0
}
```

</details>

### Positive weight value succeeds
https://user-images.githubusercontent.com/62263329/223155467-223cae2f-7ae3-4e6e-978e-24fa04e61d1f.mov

### Negative weight value fails with validation error
https://user-images.githubusercontent.com/62263329/223155517-f5c2c652-df40-4ea1-961d-8f7c6d1e5498.mov

### [BAD] Zero value should fail minimum validation but is instead ignored 😢 (this should be fixed now, but please double check when you're testing)
https://user-images.githubusercontent.com/62263329/223156335-b0b4a231-2e07-4c17-b184-a78f7225ba23.mov


pkg/gen/primemessages/update_m_t_o_shipment.go
```diff
@@ -256,6 +266,30 @@ func (m *UpdateMTOShipment) validatePpmShipment(formats strfmt.Registry) error {
 	return nil
 }
 
+func (m *UpdateMTOShipment) validatePrimeActualWeight(formats strfmt.Registry) error {
+	if swag.IsZero(m.PrimeActualWeight) { // not required
+		return nil
+	}
```
mboren, line 272: i think this is our problem with zero, swagger is interpreting zero as an omitted value, rather than the value zero.

mboren: I changed the primeEstimatedWeight fields to be nullable, and I think this is fixed now.

duncan-truss:
Do you mind also updating the model validate method to match what we are now enforcing in the API:

https://github.com/transcom/mymove/blob/d17fad39af5bcd5743f9f11bee3fa212ca219623/pkg/models/mto_shipments.go#L172-L177

I think it's probably ok to leave the prime simulator forms as is the way they are, they user will get an error message even if we aren't doing thorough form validation and the input mask.

mboren:
Can do. I also realized I only fixed the zero check for the primeEstimatedWeight field, i'll make sure the primeActualWeight field is treated consistently.

duncan-truss (approved):
I tested create and update in the prime simulator and working as I'd expect 🚀

PDickeyTruss:
This looks great @mboren! The `0` weight values now work as expected, thank you for getting this one across he finish line! 🏁

mboren:
@Mergifyio update

PR #10224, opened 2023-03-08
[MB-15313] Update to pop 6.1.1 (including related dependencies/code fixes)

reggieriser:
## Summary

This PR updates `pop` to 6.1.1.  I also ran a `go get -u` to update related dependencies because the baseline 6.1.1 includes a couple of dependencies whose version has been retracted by their authors.

We were having failures in the dependabot PRs for [pop](https://github.com/transcom/mymove/pull/9916) and [flect](https://github.com/transcom/mymove/pull/10153) that required us to start using a `TableName` override for another 18 or so tables (flect changed some of its pluralization rules -- see [this Slack thread](https://ustcdp3.slack.com/archives/CP4979J0G/p1678132232182209) for details).  We already had a good many overrides, so I decided to just add `TableName` to every model to be explicit/consistent everywhere and not rely on a pluralization routine that seems to change and hard to make 100% accurate.  I also cleaned up some unused boilerplate methods (like `validateCreate` and `validateUpdate`) while I was going through the models.

The `AuditHistory` model was a special case because its struct already had a `TableName` field in it.  Since Go won't allow a method name on a struct that already has a field by the same name, I changed the struct field's name to `Table` instead of `TableName` (if there's a better name, glad to adjust).  I only changed the struct field and left the swagger field and the database field alone to avoid a bigger change.  We can adjust those too if someone feels strongly about that.

The `ghc-pricing-parser` had to be updated as well.  It contains some code that calculates the table name from the model struct name, similar to what pop does.  Previously, it didn't look at `TableName` which was OK since it operates on a small subset of rate engine tables, but those particular model structs require a `TableName` now due to the new pluralization rules.  So, it's been changed to take `TableName` into consideration.

Note that there is a `model-vet` utility that has [an existing bug](https://dp3.atlassian.net/browse/MB-1994) around its lack of `TableName` support.  That one looked tougher to fix since it deals with an AST, plus I don't think it's been used in a long time, so I've left that one as is.

~~I plan to deploy this on experimental to make sure everything seems OK.  I seem to recall we had a deploy issue the last time we did a pop upgrade.~~  DONE: Experimental deploy seemed to work fine.  Also did an ephemeral deploy and it worked as well.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Storybook locally.

```sh
make storybook
```

##### Terminal 2

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

Just make sure tests pass and you don't see any issues just spot checking the app.

### Backend


pkg/models/audit_history.go
```diff
@@ -50,7 +50,9 @@ type AuditHistory struct {
 	ActionTstampClk time.Time `json:"action_tstamp_clk" db:"action_tstamp_clk"`
 }
 
-// AuditHistories is not required by pop and may be deleted
+// Not defining a TableName here yet because the AuditHistory struct already has a TableName field,
+// but you can't have both a field and a method with the same name for a struct.
```
reggieriser, line 54: Given that Go won't let you have a struct with a field and method of the same name, I didn't think it was worth changing the struct field here just to force the TableName override.  The struct field change wouldn't be too bad, but we might want to change the database field too for consistency which adds more complexity.  But if folks feel strongly, I can go down that path.

pkg/models/edi_processing.go
```diff
@@ -34,11 +39,6 @@ func (e *EDIProcessing) Validate(tx *pop.Connection) (*validate.Errors, error) {
 	), nil
 }
 
-// TableName overrides the table name used by Pop.
-func (e *EDIProcessing) TableName() string {
```
reggieriser, line 38: Note that [the Pop docs](https://gobuffalo.io/documentation/database/models/#using-a-custom-table-name) recommend that you use a value receiver over a pointer receiver for the `TableName` method.  I did end up running into problems on some models when I initially used pointer receivers.  I think this is due to the fact that it's calling the method via an interface, and an interface can't call a pointer receiver (or something like that).

pkg/services/dbtools/table_from_slice_creator.go
```diff
@@ -47,7 +48,14 @@ func (c tableFromSliceCreator) CreateTableFromSlice(appCtx appcontext.AppContext
...
 	}
 	builder.WriteString("TABLE ")
-	tableName := name.Tableize(elementType.Name())
+
+	var tableName string
+	if modelTable, ok := reflect.New(elementType).Interface().(pop.TableNameAble); ok {
+		tableName = modelTable.TableName()
+	} else {
+		tableName = name.Tableize(elementType.Name())
+	}
```
reggieriser, line 57: This code is used by the ghc-pricing-parser and needs to now by `TableName` aware (see notes in PR description).

ahobson (approved):
I'll admit my eyes glazed over a bit by the end, but it all looks good to me.

As always, I admire your thoroughness

mandaem (approved):
checked table names and poked around office and SM apps not seeing any new issues. LGTM

PR #10226, opened 2023-03-08
MB-15284:  Fix issue where office app was setting dates to the zero value

ronaktruss:
## Summary

Resolve issue where we were populated the expected dates with the zero date (01-01-0001). This explicitly resolves the issue for NTS-release shipments. There potentially a similar bug for PPM shipments and with this change we can confirm if the existing shipments that are not showing data in the requested move date column are due to bad data or something else.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the customer app 
2. Login as `estimated_weights@ppm.unsubmitted`
3. Delete the exisitng PPM shipment and add an NTS-release shipment and submit the move
4. Then login as an TOO user and find that move. Update the NTS-release shipment information.
5. Return back to the queue. You should see that the date associated with this shipment is not blank anymore.


pkg/handlers/ghcapi/mto_shipment.go
```diff
@@ -55,12 +55,11 @@ func (h ListMTOShipmentsHandler) Handle(params mtoshipmentops.ListMTOShipmentsPa
 			if err != nil {
 				return handleError(err)
 			}
-			mtoShipments := models.MTOShipments(shipments)
 
 			shipmentSITStatuses := h.CalculateShipmentsSITStatuses(appCtx, shipments)
 
 			sitStatusPayload := payloads.SITStatuses(shipmentSITStatuses)
-			payload := payloads.MTOShipments(h.FileStorer(), &mtoShipments, sitStatusPayload)
+			payload := payloads.MTOShipments(h.FileStorer(), (*models.MTOShipments)(&shipments), sitStatusPayload)
```
kctruss, line 62: This is not a go syntax I'm familiar with, do you have documentation you can point me towards to learn more?‏

kctruss (approved):
Just had one question for my own edification, as I'm still trying to grok all the Go idiosyncrasies, but otherwise this fix looks good

pkg/handlers/ghcapi/mto_shipment.go
ronaktruss: I haven't found any specific docs that are all that helpful. In general this is casting (this [link](https://www.codingninjas.com/codestudio/library/go-type-casting) seems okay). So normally if we just wanted to cast shipments it would look something like `models.MTOShipments(shipments)`. This looks particularly odd because we are casting to a pointer which is why we are wrapping the type as well in parenthesis.

kctruss: Ah, okay! It was the extra parenthesis that definitely threw me. Thanks for this!

PR #10234, opened 2023-03-09
feat (make): add targets for deploying/reverting deploys to non-ATO envs

felipe-lee:
## Summary

This came up as a nice-to-have improvement while testing out deployment failures. [First pass](https://github.com/transcom/mymove/pull/10183/commits/8abb9ececf9db10c28b0c912230b907baa886a1b) at this had used `:` as a separator. I switched it to `#` because I was trying to use a non-capturing group, which didn't end up working, but at that point I didn't see a reason to switch back. Lmk if you think I should switch back to it though. The original version also was only for the `exp` env, but this version should work for the three non-ATO envs we can deploy to.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1150,4 +1150,39 @@ anti_virus: ## Scan repo with anti-virus service
 # ----- END ANTI VIRUS TARGETS -----
 #
 
+#
+# ----- START NON-ATO DEPLOYMENT TARGETS -----
+#
+
+.PHONY: prepare_for_deploy
+prepare_for_deploy:  ## Replace placeholders in config to deploy to a non-ATO env. Requires DEPLOY_ENV to be set to exp, loadtest, or demo.
+ifeq ($(DEPLOY_ENV), exp)
+	@echo "Preparing for deploy to experimental"
+else ifeq ($(DEPLOY_ENV), loadtest)
+	@echo "Preparing for deploy to loadtest"
+else ifeq ($(DEPLOY_ENV), demo)
+	@echo "Preparing for deploy to demo"
+else
+	$(error DEPLOY_ENV must be exp, loadtest, or demo)
+endif
+	sed -E -i '' "s#(&dp3-branch) placeholder_branch_name#\1 $(GIT_BRANCH)#" .circleci/config.yml
+	sed -E -i '' "s#(&integration-ignore-branch) placeholder_branch_name#\1 $(GIT_BRANCH)#" .circleci/config.yml
+	sed -E -i '' "s#(&integration-mtls-ignore-branch) placeholder_branch_name#\1 $(GIT_BRANCH)#" .circleci/config.yml
+	sed -E -i '' "s#(&client-ignore-branch) placeholder_branch_name#\1 $(GIT_BRANCH)#" .circleci/config.yml
+	sed -E -i '' "s#(&server-ignore-branch) placeholder_branch_name#\1 $(GIT_BRANCH)#" .circleci/config.yml
+	sed -E -i '' "s#(&dp3-env) placeholder_env#\1 $(DEPLOY_ENV)#" .circleci/config.yml
+
+.PHONY: restore_from_deploy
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1150,4 +1150,39 @@ anti_virus: ## Scan repo with anti-virus service
...
+	@echo "Preparing for deploy to demo"
+else
+	$(error DEPLOY_ENV must be exp, loadtest, or demo)
+endif
+	sed -E -i '' "s#(&dp3-branch) placeholder_branch_name#\1 $(GIT_BRANCH)#" .circleci/config.yml
+	sed -E -i '' "s#(&integration-ignore-branch) placeholder_branch_name#\1 $(GIT_BRANCH)#" .circleci/config.yml
+	sed -E -i '' "s#(&integration-mtls-ignore-branch) placeholder_branch_name#\1 $(GIT_BRANCH)#" .circleci/config.yml
+	sed -E -i '' "s#(&client-ignore-branch) placeholder_branch_name#\1 $(GIT_BRANCH)#" .circleci/config.yml
+	sed -E -i '' "s#(&server-ignore-branch) placeholder_branch_name#\1 $(GIT_BRANCH)#" .circleci/config.yml
+	sed -E -i '' "s#(&dp3-env) placeholder_env#\1 $(DEPLOY_ENV)#" .circleci/config.yml
```
rogeruiz, line 1173:
```suggestion
	sed -E -i '' "s#(&dp3-env) placeholder_env#\1 $(DEPLOY_ENV)#" .circleci/config.yml
	@git diff .circleci/config.yml
	@echo "Please make sure to commit the changes in .circleci/config.yml in order to have CircleCI deploy $(GIT_BRANCH) to the Non-ATO $(DEPLOY_ENV) environment."
```

rogeruiz: Let's add some more helpful output when these `sed` commands run so folks who don't read `sed` can understand what just happened and what their responsibilities are for deploying.

rogeruiz (changes requested):
❌ This PR needs changes ...

Good work on updating this to support multiple environments, @felipe-lee. I have some suggestions for making this more ergonomic for everyone on the project.

rogeruiz:
Once this is merged in, let's update the Confluence page which has folks manually doing these steps in their editors.

[=> Confluence page](https://dp3.atlassian.net/wiki/spaces/MT/pages/1470922798/How+to+deploy+to+Non-ATO+environments+Experimental+or+Demo+or+Loadtest+using+CircleCI#How-to-do-it) 🔒

Makefile
felipe-lee: @rogeruiz, I agree with most of the changes requested, but I feel like the git diff bit shouldn't be part of it. Everyone on the project uses git differently and some may not be aware of how to even get out of the pager. We could make it not use the pager though, and just output it to stdout like this:

```shell
@git --no-pager diff .circleci/config.yml
```

What do you think?

rogeruiz: Great idea, let's have the `--no-pager`. I think this is a good balance. My original intention was to show the person using this what changed rather than just a bunch of `sed` commands.

rogeruiz (approved):
lgtm 🌈

felipe-lee:
> Once this is merged in, let's update the Confluence page which has folks manually doing these steps in their editors.
> 
> [=> Confluence page](https://dp3.atlassian.net/wiki/spaces/MT/pages/1470922798/How+to+deploy+to+Non-ATO+environments+Experimental+or+Demo+or+Loadtest+using+CircleCI#How-to-do-it) 🔒

@rogeruiz I made updates to [=> Confluence page](https://dp3.atlassian.net/wiki/spaces/MT/pages/1470922798/How+to+deploy+to+Non-ATO+environments+Experimental+or+Demo+or+Loadtest+using+CircleCI#How-to-do-it) 🔒. Lmk what you think, or feel free to make updates if you think something should be worded differently.

PR #10246, opened 2023-03-11
Discovery/Spike: Get gotenberg running locally - Do not merge, this isn't meant to be final implementation code

felipe-lee:
## Summary

This adds a temporary CLI command to:

- [x] convert a PPM Shipment's document uploads to PDFs
- [x] merge document upload PDFs into a single PDF
- [x] upload merged PDF and save info in DB
- [x] can run with s3 locally [instructions for using s3 locally](https://dp3.atlassian.net/wiki/spaces/MT/pages/1470955567/How+to+test+storing+data+in+S3+locally)🔒

The idea is that main code that the command uses would likely get moved to service objects, which could then be called from wherever.

Note: This command can only run locally at the moment, and would take some work (not sure how much) to be able to run in a deployed environment.

Note: This doesn't actually handle the full payment packet, it only handles PPM shipment document uploads, so it isn't handling the SSW or the orders (which are needed for the AOA and the payment packet). 

## Setup to Run Your Code

<details>
<summary>💻</summary>

### Gotenberg

Get `gotenberg` running locally

```sh
docker run --rm -p 2000:3000 gotenberg/gotenberg:7 gotenberg --chromium-disable-routes=true --webhook-disable=true
```

### PDF conversion command

This runs the new cli command which converts the two uploads it has from pngs to pdfs, merges them, uploads the merged PDF, downloads the newly uploaded PDF, and saves it to your `mymove/tmp` directory with the name format `payment-packet-<timestamp string>.pdf`.

```sh
go run ./cmd/convert-uploads-to-pdf -p defb263e-bf01-4c67-85f5-b64ab54fd4fe
```

An alternative PPM shipment ID you can use is `08ab7a25-ef97-4134-bbb5-5be0e0de4734`, which contains two weight tickets. The first has the same [test pdf]() as both the empty and full document uploads, but the second weight ticket has the [test png]() as the empty document upload and the [constructed weight sheet]() for the full document upload.

</details>


migrations/app/schema/20230314011931_add_fields_to_ppm_shipments_for_aoa_and_payment_packets.up.sql
felipe-lee, line 1: To be honest, not really sure where to put these. I just put them on `ppm_shipments` to get the work done for now, but we can definitely discuss what makes sense for the real implementation.

pkg/testdatagen/make_weight_ticket.go
```diff
@@ -82,6 +82,30 @@ func MakeDefaultWeightTicket(db *pop.Connection) models.WeightTicket {
 	return MakeWeightTicket(db, Assertions{})
 }
 
+// MakeWeightTicketWithConstructedWeight creates a single WeightTicket and associated relationships with weights and documents
```
felipe-lee, line 85: This is somewhat hacky tbh, but it gets what we need without over-complicating the testdatagen code in the `shared.go` file.

pkg/testdatagen/scenario/shared.go
felipe-lee, line 1: Generally added `UserUploader: userUploader` to a few assertions because without it, files that are "uploaded" in testdatagen functions aren't viewable when you run the app.

scottmries (commented):
This LGTM as a preliminary, @felipe-lee! I successfully generated PDFs for both of the PPMs listed, which was really fast. The included PDFs are pixel-perfect, and the tables look good.

suggestion: The only thing I would note is that the test PNG loses a bit of fidelity in the generated PDF; it seems to be pinned about halfway down the page, and I think might overrun the page boundary some? Considering that PNGs will be user-generated and thus likely vary widely in the scale and quality of content, I wonder if maximally scaling them, pinning them as close to the page origin as possible, and rotating them to portrait orientation (if need be) would be good, to give them the best chance at being legible? I don't think this needs to happen here, I just thought I'd point it out.

cmd/convert-uploads-to-pdf/main.go
```diff
@@ -0,0 +1,715 @@
...
+
+	res, clientErr := http.DefaultClient.Do(req)
+
+	if clientErr != nil {
+		appCtx.Logger().Error("Failed to convert file to PDF", zap.Error(clientErr))
+
+		return nil, clientErr
+	}
+
+	if res.StatusCode != http.StatusOK {
```
pearl-truss, line 482: **Question:** since this is ultimately called sequentially going through a group of files, would it be helpful to include `fileName` or `id` in the error message?

pkg/uploader/allowed_file_types.go
```diff
@@ -1,26 +1,45 @@
...
+	FileTypePDF = "application/pdf"
+	// FileTypeText is the content type for text files
+	FileTypeText = "text/plain"
+	// FileTypeTextUTF8 is the content type for text files with UTF-8 encoding
+	FileTypeTextUTF8 = "text/plain; charset=utf-8"
+	// FileTypeExcel is the content type for Excel files
+	FileTypeExcel = "application/vnd.ms-excel"
+	// FileTypeExcelXLSX is the content type for Excel xlsx files
+	FileTypeExcelXLSX = "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
+)
```
pearl-truss, line 18: Nice improvement!

pearl-truss (commented):
Great job Felipe! I left a few non pressing questions

felipe-lee:
> This LGTM as a preliminary, @felipe-lee! I successfully generated PDFs for both of the PPMs listed, which was really fast. The included PDFs are pixel-perfect, and the tables look good.

🎉 

> suggestion: The only thing I would note is that the test PNG loses a bit of fidelity in the generated PDF; it seems to be pinned about halfway down the page, and I think might overrun the page boundary some? Considering that PNGs will be user-generated and thus likely vary widely in the scale and quality of content, I wonder if maximally scaling them, pinning them as close to the page origin as possible, and rotating them to portrait orientation (if need be) would be good, to give them the best chance at being legible? I don't think this needs to happen here, I just thought I'd point it out.

Ah, yeah, that was something that I was wondering about. There is [a function in the original upload conversion code](https://github.com/transcom/mymove/blob/ab14a31c4e666ffda8197160ed21552da54edce7/pkg/paperwork/generator.go#L228-L274) (I think it was proof-of-concept code because it wasn't wired up in the app, only in a command like this code) that tries to reduce the unused space and rotates the image, but the main reason I didn't pull that in is because we don't analyze the contents of the image, so if we just rotate based on size, they might end up with text that is sideways. Maybe that doesn't matter though as long as it's maximally legible.

cmd/convert-uploads-to-pdf/main.go
```diff
@@ -0,0 +1,715 @@
...
+		// we'll close the downloaded file if we've converted it because we're not returning it, but if we didn't
+		// convert it (and thus didn't get to this part), we don't close it because we're returning it and the caller
+		// will need to close it.
+		if err := download.Close(); err != nil {
+			appCtx.Logger().Error("Failed to close download stream", zap.Error(err))
+		}
+
+		// Not setting up closing of outputPDF file since we're returning it. Caller will need to close it.
+
+		if conversionErr != nil {
```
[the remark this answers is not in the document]
felipe-lee: Good point, we probably should close the stream if we are erroring out, I'll add that.

cmd/convert-uploads-to-pdf/main.go
```diff
@@ -0,0 +1,715 @@
...
+	}
+
+	return res.Body, nil
+}
+
+// saveMergedPDF uploads the merged PDF to storage and saves the relevant DB info.
+func saveMergedPDF(appCtx appcontext.AppContext, userUploader *uploader.UserUploader, ppmShipment *models.PPMShipment, mergedPDF io.ReadCloser) error {
+	txnErr := appCtx.NewTransaction(func(txnAppCtx appcontext.AppContext) error {
+		// We'll need to decide if we want to load this like this, or if we'll have a custom loader for PPMShipment that
+		// includes the ServiceMember already.
```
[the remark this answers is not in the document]
felipe-lee: Good question, I meant having another fetcher that loads exactly the data needed for this work. Our ppm shipment fetcher already has a few versions of fetching ppm shipments: https://github.com/transcom/mymove/blob/main/pkg/services/ppmshipment/ppm_shipment_fetcher.go

Although I think it might be worthwhile to make a more [generic fetcher like the mto shipment one](https://github.com/transcom/mymove/blob/ab14a31c4e666ffda8197160ed21552da54edce7/pkg/services/mto_shipment/mto_shipment_fetcher.go#L117-L137) so the caller can pass in the desired associations instead of each use case having its own version of the fetch.

cmd/convert-uploads-to-pdf/main.go
```diff
@@ -0,0 +1,715 @@
...
+			txnAppCtx.Logger().Error("Failed to prepare file for upload", zap.Error(prepErr))
+
+			return prepErr
+		}
+
+		newUpload, uploadVerrs, uploadErr := userUploader.CreateUserUploadForDocument(
+			txnAppCtx,
+			ppmShipment.PaymentPacketID,
+			// Do we have a system user ID? This is meant to be the uploader ID, but it'll be the system, not a specific
+			// user.
```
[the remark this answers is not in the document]
felipe-lee: That feels like it could be ok, though the actual thing we save once we fully implement this will include the SSW which we generate ourselves, so I'm not sure if that changes things. This might be a question we ask USTC in terms of auditing purposes. I feel like it should be fine in that we will know that any upload that is an AOA or payment packet is one generated by the system, not the customer, but USTC may disagree.

cmd/convert-uploads-to-pdf/main.go
felipe-lee: Yeah, that's a good point. In general it might be helpful to pass around an object that contains data about the file (like id, name, type, and the stream itself) instead of the bare stream. That way we could more easily use info as needed for things like this.

cmd/convert-uploads-to-pdf/main.go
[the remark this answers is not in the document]
pearl-truss: Thanks for the links! I see what you're saying now and the generic fetcher makes the most sense to me here since there's multiple use cases

felipe-lee:
After demoing at the internal demo, some good feedback was brought up:

@mmayo-truss asked: I was thinking like a devious person and I wondered what would happen if you tried to generate a pdf where someone put data in the last row and column in the excel sheet. I think it is row 1048576 and column XFD.

I replied saying that we haven't tested that, but we should. It currently doesn't output all the empty cells so maybe it won't be an issue, but it could. It could also affect processing time, but we're already going to process this async to help with general pdf gen taking a while.

@rogeruiz : it might be good to demo this or have tests for this with different kinds of PDFs and images that are more recognizable. but this looks awesome!

I replied saying my plan was as we clean up the code and get it into proper places in the codebase rather than all in the command, we’ll be able to test it with any documents we want to upload. 

@duncan-truss brought up a few things:

* That’s really encouraging the spreadsheet rendered so cleanly.  Do we care about PDF and image orientation before combining?
    * This is also something Scott brought up above https://github.com/transcom/mymove/pull/10246#pullrequestreview-1341700172. We should probably spend some time looking into that.
* I know our max per document size is ~25MB, do we know of any memory or resource caps to be aware of? I see reference to our S3 files needing to be smaller than 250 MB for the anti-virus to run.
    * The proof of concept code doesn’t take file size into account, but it likely is something we’ll need to test more thoroughly to see if we are running into issues. For reference, the file I generated for the demo is 342 KB, so we’re well under the limit right now, but it also doesn’t have all the things in it yet.

I hope I captured all the feedback, feel free to correct me if I misstated anything!

This PR won't address any of this feedback directly as it was a proof of concept and we're working on cleaning up individual bits of the code now, but capturing the feedback here because it came up in response to this being demoed.

felipe-lee:
Closing since most of the code has made it into the repo now and the remaining bits may or may not be needed, depending on how we implement things in the end. Merging main into this branch would be a bit of a pain since the branches diverged for many of these changes. If needed again, it would be easier to take the command-specific logic in this PR and add it to a new branch off of main, using the now integrated portions of the code for interacting with gotenberg and the DB. This PR will still be available for reference as needed, just closed.

PR #10251, opened 2023-03-13
MB-15451: refactor pop transaction tests

ahobson:
## Summary

This PR refactors how we set up the transactional db server tests.

The first commit only moves all code related to transactions from `pop_suite.go` to `pop_suite_txn.go` and does not change any behavior.

The second commit is the big one.  It makes some bigger changes

* No need for a custom store. Once we register go-txdb, we can let the
  usual pop process connect using the custom driver
* Use a manual savepoint for preloaded data instead of trying to use
  sqlx.Beginx. We can do that because we know all connections will be
  inside a transaction thanks to go-txdb.
* Ensure that if a test process is reused for multiple packages, it
  uses the same database since go-txdb registers a db URL once

## Setup to Run Your Code

<details>

##### Terminal 1

Run the server tests

```sh
make server_test
```

</details>


mlm55 (approved):
Thanks for walking though this with us!

PR #10282, opened 2023-03-16
MB-15344 add and implement factory.BuildExtendedServiceMember

StuTruss:
## Summary

This ticket adds a function `BuildExtendedServiceMember` intended to be an eventual replacement for `testdatagen.MakeExtendedServiceMember`
* Created `BuildExtendedServiceMember` function
* Created tests for `BuildExtendedServiceMember`. 
* Replaces instances in code where `MakeExtendedServiceMember` exists with `BuildExtendedServiceMember`

## Setup to Run Your Code

### Frontend

- No frontend changes

### Backend

- No schema changes


pkg/testdatagen/scenario/shared.go
```diff
@@ -111,7 +111,24 @@ func createGenericPPMRelatedMove(appCtx appcontext.AppContext, moveInfo MoveCrea
...
+			},
+		},
+		{
+			Model: models.User{
+				LoginGovUUID:  models.UUIDPointer(uuid.Must(uuid.NewV4())),
+				LoginGovEmail: moveInfo.Email,
+				Active:        true,
+			},
+		},
+	}, nil)
```
ahobson, line 131: It seems like `smAssertions` are no longer used, so we could remove them, right?

Should we consider trying to populate the service member and user customizations from the passed in `assertions` ?

pkg/factory/service_member_factory.go
```diff
@@ -82,6 +82,43 @@ func BuildServiceMember(db *pop.Connection, customs []Customization, traits []Tr
...
+	}
+
+	serviceMember := BuildServiceMember(db, customs, traits)
+
+	backupContact := BuildBackupContact(db, []Customization{
+		{
+			Model:    serviceMember,
+			LinkOnly: true,
+		},
+	}, nil)
```
ahobson, line 111: Huh.

I think we want a way to customize the backup contact when creating an extended service member.

I noticed `BuildBackupContact` calls `BuildServiceMember` with all customizations and traits. Maybe `BuildExtendedServiceMember` should just call `BuildBackupContact`, get the serviceMember from it, and then return it after appending to `BackupContacts`.

Thoughts?

pkg/factory/service_member_factory.go
```diff
@@ -82,6 +82,43 @@ func BuildServiceMember(db *pop.Connection, customs []Customization, traits []Tr
...
+	// Find ServiceMember customization and extract the custom ServiceMember
+	var cServiceMember models.ServiceMember
+	if result := findValidCustomization(customs, ServiceMember); result != nil {
+		cServiceMember = result.Model.(models.ServiceMember)
+		if result.LinkOnly {
+			return cServiceMember
+		}
+	}
+
+	serviceMember := BuildServiceMember(db, customs, traits)
```
ahobson, line 104: Looks like the existing code has [different defaults for an extended service member](https://github.com/transcom/mymove/blob/main/pkg/testdatagen/make_service_member.go#L113-L121),  in particular the `EmailIsPreferred` setting

pkg/factory/service_member_factory.go
```diff
@@ -79,9 +123,57 @@ func BuildServiceMember(db *pop.Connection, customs []Customization, traits []Tr
 		mustCreate(db, &serviceMember)
 	}
 
+	if buildType == serviceMemberBuildExtended {
```
mlm55, line 126: question: since this is different from our typical pattern (where creating the model is the last thing in the function), would it be worth commenting why it has to be done this way?

pkg/factory/service_member_factory.go
```diff
@@ -8,11 +8,31 @@ import (
...
 
-// BuildServiceMember creates a single ServiceMember
-// Also creates, if not provided:
-// - Residential Address of the ServiceMember
-// - User
-func BuildServiceMember(db *pop.Connection, customs []Customization, traits []Trait) models.ServiceMember {
+type serviceMemberBuildType byte
+
+const (
+	serviceMemberBuildBasic serviceMemberBuildType = iota
```
mlm55, line 14: praise: i like how you did this. I hadn't heard of iota before!

pkg/testdatagen/scenario/shared.go
```diff
@@ -297,16 +303,21 @@ func createMoveWithPPMAndHHG(appCtx appcontext.AppContext, userUploader *uploade
...
+				FirstName:     models.StringPointer("Submitted"),
+				LastName:      models.StringPointer("Ppmhhg"),
+				Edipi:         models.StringPointer("6833908165"),
+				PersonalEmail: models.StringPointer(email),
+			},
 		},
-	})
+		{
+			Model:    user,
+			LinkOnly: true,
```
mlm55, line 318: this is out of scope for this ticket, but i wonder if we should start combining some of these build calls at some point. Like, instead of calling `factory.BuildUser` and then linking it to `smWithCombo`, we could just add the user customizations here. Might be marginally faster since `BuildExtendedServiceMember` ends up calling `BuildUser` anyway

mlm55 (approved):

mlm55 (approved):
had a couple comments, but nothing blocking. Great job dealing with all the complexity!

pkg/factory/service_member_factory.go
ahobson: Excellent idea!

PR #10286, opened 2023-03-17
[ office-hours ] Updating the browsers list database

rogeruiz:
## Summary

> I forgot to commit and push this up on the 15th of March 2023 during my office
> hours. So I'm creating this PR now.

When running `make client_test`, I noticed that our browsers lists
database was out of date. I ran it using `npx update-browserslist-db@latest` as suggested by the warning message.

Read more about why updating regularly is important:

- https://github.com/browserslist/update-db#why-you-need-to-call-it-regularly

### Additional steps

1. Run `make client_test` and see no warnings about browsers list DB being out
   of date.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -7577,20 +7577,10 @@ caniuse-api@^3.0.0:
     lodash.memoize "^4.1.2"
     lodash.uniq "^4.5.0"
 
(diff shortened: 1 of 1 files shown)


mmayo-truss (approved):
LGTM. The documentation on why we should be updating this frequently was helpful and informative.

PR #10287, opened 2023-03-17
[MB-15459] Update office hint text for missing weight tickets

pearl-truss:
## Summary

This PR updates the UI for when the customer doesn't have a weight ticket for empty or full weight.

**NOTE:**
- There is a font size difference for the upload help text: "Upload your completed constructed weight spreadsheet". In the design this was slightly larger than the upload help text for the empty weight field. Since both use the same component, I wanted to double check if a font size override for one instance was still desired?
- There is a color difference in the blue info badge here than what's in the design. I used the default Alert component which has a color difference than the design. Since I didn't see the design color for this info Alert used anywhere else I wanted to double check if the default color is okay to use for consistency?

## Setup to Run Your Code

Start the Storybook locally.

```sh
make storybook
```

Navigate to the Weight Ticket Form component

## Screenshots
Screenshot 2023-03-17 at 10 10 43 AM
Screenshot 2023-03-17 at 10 18 38 AM
Screenshot 2023-03-17 at 10 18 53 AM


rileyorr (commented):
Great job on this @pearl-truss! The only thing I noticed was that the constructed weight spreadsheet file uploader should only accept XLS or XLSX. Right now the copy says it accepts PDF, JPG, or PNG only (screenshot attached below). 

The hint text should be updated to: 'XLS or XLSX only. Maximum file size of 25 MB.' Here's a link to [the art board with the correct copy](https://www.figma.com/file/tRkjtoWjT8727XH5m9k7mK/Milmove-Customer-App?node-id=4615%3A25296&t=VTYYm6ATqJKhRjby-1) for reference. Let me know if you have any questions!

Screenshot 2023-03-20 at 11 54 06 AM

rileyorr (approved):
> There is a font size difference for the upload help text: "Upload your completed constructed weight spreadsheet". In the design this was slightly larger than the upload help text for the empty weight field. Since both use the same component, I wanted to double check if a font size override for one instance was still desired?

Great callout! I just checked and it looks like something with the way the components are set up in Figma. The larger label text is for the desktop version and the smaller label text is for mobile. For the purposes of what we've built we **shouldn't** override for this instance. Thank you for asking!

> There is a color difference in the blue info badge here than what's in the design. I used the default Alert component which has a color difference than the design. Since I didn't see the design color for this info Alert used anywhere else I wanted to double check if the default color is okay to use for consistency?

Using the default alert component was a good call! This looks like another Figma component inconsistency. The design team has an epic capturing these items so I'll add it if it isn't already documented. Thanks for pointing this out! 

@kylehilltruss this LGTM!

bellex3 (approved):
checked storybook after making the constructed weight copy changes. LGTM

PR #10289, opened 2023-03-17
[ MB-15134 ] Adding documentation to date formatter

rogeruiz:
## Summary

According to the Moment documentation, we should be providing a date
format to parse the string. This fixes that, but also adds more
documentation around this function and the new constant.


mmayo-truss (approved):
This looks good and is definitely more informative. I think adopting more usage of JS docs in our code going forward will be beneficial to us across the project.

PR #10306, opened 2023-03-21
MB-15529 playwright wait for navigation

ahobson:
## Summary

The playwright team has [deprecated waitForNavigation](https://playwright.dev/docs/api/class-page#page-wait-for-navigation).

This PR replaces it with `locator.waitFor` when possible, and `page.waitForURL` otherwise

Also do some flakiness fixes found along the way

## Setup to Run Your Code

<details>

##### Terminal 1

Run the e2e tests

```sh
make e2e_test
```

</details>


scripts/run-e2e-test
kylehilltruss, line 1: Just calling out that there's changes in this file as well that seem kinda unrelated to the others, but likely fall under the column header of "flakiness fixes," so if they were intentional that's fine.

kylehilltruss (approved):
Andrew and I [discussed](https://ustcdp3.slack.com/archives/C04GP2BRQBX/p1679409491315649) whether or not the `page.getBySomething().waitFor` model of affirmative content was a good choice or not, and ultimately decided that it makes more cognitive sense to use it rather than `expect(page.getBySomething()).toBeVisible`. I'll go in to update methods I've written that use that model soon.

scripts/run-e2e-test
ahobson: Yeah! They are! I found them when working locally and figured it was nice to get them in. Thank you for asking

PR #10330, opened 2023-03-24
[ MB-12170 ] Edit PPM net weight query invalidation clarification

mmayo-truss:
## Summary

Removed the reference to useMemo for the weight calculation allows for the weight calculation to get refreshed when only the Documents query gets invalidated. This way we don't have to invalidate the MTO shipments query.

I also removed the addition of the full weight to the initial values of the form since it is unnecessary for the validation check.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the office app
2. Login as a services counselor
3. Access a PPM closeout move with excess weight (Move code XSWT01 is a good one)
4. click the review documents button at the bottom of the ppm shipment card
5. click on the edit net weight button 
image
6. edit the  net weight to be greater or equal to the full weight
7. you should see an error message
image
8. Edit the net weight to make the move not overweight and click save
9. You should not see the warning on the net weight and the weights should have been refreshed in the calculations.
image


michaelvaldes-truss (approved):
lgtm, and the solution makes sense!

PR #10362, opened 2023-03-29
[ MB-14981 ] Add orders type to getMTO endpoint in the prime api

mmayo-truss:
## Summary

This PR adds ordersType to the Prime API getMoveTaskOrder endpoint. The definition for ordersType was repeated in both the internal and ghc swagger definition files. Instead of copying it one more time, I moved the definition into its own file.

One thing to consider: I noticed that, in the struct definition for the order model, [OrdersType is typed as the internal messages OrdersType generated by go-swagger](https://github.com/transcom/mymove/blob/e15877069712a3ef2afba279bcf345dd07a568e0/pkg/models/order.go#L43). This is different from how we define other similar fields where we declare a type in the model file. (ex. [ShimentType](https://github.com/transcom/mymove/blob/25b77e9c2995e79794ae45f893f6047db6322621/pkg/models/mto_shipments.go#L135)) I decided to leave this like it is, as it seems like it is working fine, but thought it would be worth pointing out.

You can see how the ordersType is [added to the documentation here](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/transcom/mymove/mmayo/MB-14981/add-orders-type-to-getMTO-api/swagger/prime.yaml#tag/moveTaskOrder/operation/getMoveTaskOrder).

## Setup to Run Your Code

<details>

##### Terminal 1

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Hit the getMoveTaskOrder endpoint through Postman OR view the response from the Prime Simulator 
2. You should see ordersType in the orders. 

image

### Frontend

### Backend


swagger-def/prime.yaml
```diff
@@ -1054,7 +1054,7 @@ definitions:
         description: >
           The estimated weight of this shipment, determined by the movers during the pre-move survey.
           This value **can only be updated once.** If there was an issue with estimating the weight and a mistake was made,
-          the Prime contracter will need to contact the TOO to change it.
+          the Prime contractor will need to contact the TOO to change it.
```
kctruss, line 1057: **praise:** ‏Thanks for catching this typo!

swagger-def/internal.yaml
```diff
@@ -1755,17 +1755,6 @@ definitions:
...
       CANCELED: Canceled
-  OrdersType:
-    type: string
-    title: Orders type
-    enum:
-      - PERMANENT_CHANGE_OF_STATION
-      - RETIREMENT
-      - SEPARATION
-    x-display-value:
-      PERMANENT_CHANGE_OF_STATION: Permanent Change Of Station (PCS)
```
ronaktruss, line 1766: Do we know what x-display-value does / how it's used?

Aside/additional details:
We have a drop down on the orders page that uses the `Permanent Change Of Station (PCS)` value, but that is using the value from [here](https://github.com/transcom/mymove/blob/main/src/constants/orders.js#L8). 

Page that this is being used on:
image

ronaktruss: Sorry the above may not be very clear. Here's some context:
I was under the impression that the x-display-values were being used to either display the values or do some sort of validation. Based on that rationale I thought the above page would have caused some errors. It doesn't and the field is actually being populated through some other constants.

kctruss: Just to confirm @ronaktruss - since the `x-display-value` is not being used to populate the UI dropdown, this change does not need design approval for the removal of `(PCS)`, is that right?

kctruss (approved):
This looks good to me. I'm curious about the history of `x-display-value` as well, but I think we can look into that separately from this PR.

swagger-def/internal.yaml
ronaktruss: Yeah, that is my understanding.

Additional context:
We do have a [language guide](https://dp3.atlassian.net/wiki/spaces/MT/pages/1653112899/MilMove+Copy+Style+Guide#%5CuD83C%5CuDDF5-P-S) that indicates that on the customer app most users are familiar with `PCS` acronym rather than the elongated form.

kctruss: Gotcha. So it might be worthwhile to return that parenthetical to the display value, then? What are your thoughts @mmayo-truss ?

mmayo-truss: @ronaktruss Looking into it more and I am not entirely sure. It looks like it could have at one point been [something like this](https://redocly.com/docs/api-reference-docs/specification-extensions/x-enum-descriptions/) that will display in our redocly docs. However I tried the x-enumdescriptions and that did not work with our version of redocly. It looks like I could use the description to do something similar like we do with shipmentType. 
image

PR #10374, opened 2023-03-31
[MB-14962] gbloc output for prime

kctruss:
## Summary

Adding the GBLOC to the Orders response payload in the Prime API 

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access the office app
2. Login as a prime user
3. Select a move and open the network tab of the developer console
4. Verify that the response object includes `OriginDutyLocationGBLOC` with the correct GBLOC

## Screenshots
Screenshot 2023-03-31 at 1 03 02 PM

### DIFF
--- a/pkg/handlers/primeapi/payloads/model_to_payload.go
+++ b/pkg/handlers/primeapi/payloads/model_to_payload.go
@@ -140,6 +140,7 @@ func Order(order *models.Order) *primemessages.Order {
 		Entitlement:             Entitlement(order.Entitlement),
 		ID:                      strfmt.UUID(order.ID.String()),
 		OriginDutyLocation:      originDutyLocation,
+		OriginDutyLocationGBLOC: swag.StringValue(order.OriginDutyLocationGBLOC),
 		OrderNumber:             order.OrdersNumber,
 		LinesOfAccounting:       order.TAC,
 		Rank:                    order.Grade,
--- a/pkg/handlers/primeapi/payloads/model_to_payload.go
+++ b/pkg/handlers/primeapi/payloads/model_to_payload.go
@@ -149,6 +149,10 @@ func Order(order *models.Order) *primemessages.Order {
 		OrdersType:              primemessages.OrdersType(order.OrdersType),
 	}
 
+	if payload.Customer.Branch == "Marines" {
+		payload.OriginDutyLocationGBLOC = "USMC"
+	}
+
 	return &payload
 }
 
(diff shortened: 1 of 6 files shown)


pkg/handlers/primeapi/payloads/model_to_payload.go
```diff
@@ -148,6 +149,10 @@ func Order(order *models.Order) *primemessages.Order {
 		OrdersType:              primemessages.OrdersType(order.OrdersType),
 	}
 
+	if payload.Customer.Branch == "Marines" {
```
mmayo-truss, line 152: Question: we don't store the correct GBLOC for Marines?

kctruss: We do not, as the services counselor routing is based on the duty location, but we want the Prime to see "USMC"

mmayo-truss: Follow up Question: Would it be better to store "USMC" when we create the orders and then send exactly what is on the db? I am not sure how this field will be used in the future, but any time we use this field going forward I assume we will have to remember to make this same check for Marine-ness. https://github.com/transcom/mymove/blob/4ec8ecec1a2b3562979d68cd6241295b279d7fed/pkg/handlers/internalapi/orders.go#L145-L153

kctruss: That's a fair question, and I also want to rephrase my previous answer. 
The GBLOC we store on the orders table is *always* the GBLOC associated with the postal code of the origin duty station, regardless of branch.

As it currently stands, MTOs for Marines override both the OriginDutyLocationGBLOC and the MoveGBLOC with "USMC". Hence this logic being all over the place. 

Based on previous discussions with Dan, we need to carry forward this USMC logic for what we send to the prime and what we send in the EDI (see Namibia's current ticket). Without disrupting the rest of the routing logic further up in the service layer, which would require a significant LOE for refactor, I believe following the existing pattern makes the most sense within our current Milestone 1 directive.

mmayo-truss: Makes sense to me. Thank you for that context.

mmayo-truss (approved):
LGTM! I was able to verify that the prime api getMoveTaskOrder endpoint is returning the origin gbloc as expected. 🏅 🌈 
image

PR #10381, opened 2023-04-03
MB-15084: Add submitted date for TOO

ronaktruss:
## Summary

This adds a new column (date submitted) for the TOO queue and adds the associated filtering and sorting functionality. We were able to leverage existing fields for this, but to account for the scenario where a prime adds a service item which leads to the move returning to the TOO queue a new column was added. While we can technically derive that information using the `mto_service_items` table this new column simplifies this behavior and makes it explicit what date should be used.

### Needs/Areas to focus on

- Verify that this approach is sound. (Ie. Is the column necessary?) I think that its possible to derive the date by looking at the mto_service_items but having an explicit field felt easier to manage since its on the move level.

- Double check the column name. Is there a better name? I had some trouble coming up with a name that isn't overly long but also captures the function of the column.

- I haven't updated existing test data to include data that populates this information. If you think this is something that should be done within this PR please let me know.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

There are three scenarios that determine what date we are showing. The date shown should indicate when the move entered/reentered the TOO queue.
1. If a duty location does not provide service counseling and the move is directed directly to the TOO
2. If the duty location does provide service counseling and the move moves from the SC queue to the TOO queue.
3. When the move is available to the prime and a new service item is requested which makes the move appear in the TOO queue.

## Screenshots
image


kctruss:
@ronaktruss Can you touch on some of the tradeoffs between deriving this date value vs. storing it in the DB? I'm on the fence about this approach and would love to hear more about your process reaching this decision. 

Specifically:
- Is this approach extensible, should additional workflows be added to trigger the return to this status?
- If we do not backfill this column for existing moves in this status, what do those moves look like in the queue?
- Is there a noticeable performance difference between deriving the value vs. storing it?

As far as column naming, I would suggest something closer to `approvals_requested_at` since the time stamp is specific for that status.

ronaktruss:
Yeah that naming sounds more appropriate.

To answer your questions and provide some context:

General context:
Right now to derive this value I think we would have to look at the mto_service_items table and for a specific move id find the earliest created_at date across all service items that are in the submitted status. Its unclear to me if this covers all of the scenarios we have for service items (See the answer to question 2 for an example of when not having an explicit field might be detrimental). I had some earlier drafts of the description indicating that it may not cover all scenarios but didn't want to preemptively sway the conversation. 

> Is this approach extensible, should additional workflows be added to trigger the return to this status?

I think the current approach is more extensible as it allows us to fully control when we are actually updating the value that we are concerned with rather than relying on the circumstantial data we have. This change is occurring on the move router so we should be able to control this logic anytime we expect to make any changes to the move status. Without this column we are limited to scenario where we can derive information based on the status, created_at, or updated_at fields in the mto_service_items table. While this may be enough, this does appear to be less flexible and less consistent with our current approach. 

> If we do not backfill this column for existing moves in this status, what do those moves look like in the queue?

If we add this new column we wouldn't have a way of backfilling this data. Those moves would likely have a previous date in them (either the move level submitted_at time or the service_counseling_approved_at time). If we try to derive this value from existing tables we rely on the status of the service items so once a move is approved we would loose that context and we would see the same issue for moves that are in the status of `move approved` (on the UI). We may be able to handle this logic by leveraging the status and updated_at time stamps, but this would require some deeper understanding of the possible scenarios.
While this was a consideration it was a bit lower since we already conceded to this being the reality for many of the existing moves when we were adding the previous column. 

> Is there a noticeable performance difference between deriving the value vs. storing it?

I don't think there is a noticeable performance difference, but would put the edge on storing the value since we are already touching the moves table when we are introducing this new logic. Deriving this value would mean some additional join queries which would theoretically have a larger impact.

The main reasons for advocating for this approach are flexibility, simplicity in terms of implementation, and ease of understanding.

**Flexibility**:  This is explained in the response to you first question.
**Simplicity in terms of implementation**: While we are adding a column we are then only doing simple logic with that value, rather than potentially having to apply different logic based on the different statuses of the move and the service items.
**Ease of understanding**: This makes it so the logic is in the service object and should we need to update it we know that we will need to adjust the service object to ensure that the field is updated when it should be rather than hide the logic within sql queries/ pop and a mix of application code.

pkg/services/order/order_fetcher_test.go
```diff
@@ -251,6 +251,61 @@ func (suite *OrderServiceSuite) TestListOrders() {
 
 	})
 
+	suite.Run("returns moves filtered appeared in TOO at", func() {
+		// Under test: ListOrders
+		// Expected outcome: Only the one move with the right date should be returned
```
kctruss, line 256: **praise:** I really appreciate this commenting convention. It's not something I've noticed us doing regularly, and I'd encourage you to bring it up in a future appeng huddle or FE/BE sync, because I think it would be a really beneficial habit for anyone writing Go tests to add. ‏

pkg/services/order/order_fetcher.go
```diff
@@ -465,6 +481,7 @@ func sortOrder(sort *string, order *string, ppmCloseoutGblocs bool) QueryOption
 		"locator":                 "moves.locator",
 		"status":                  "moves.status",
 		"submittedAt":             "moves.submitted_at",
+		"appearedInTooAt":         "GREATEST(moves.submitted_at, moves.service_counseling_completed_at, moves.sent_back_to_too_at)",
```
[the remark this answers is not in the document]
ronaktruss: Yeah, that's my understanding. At least one of these values should be populated if the move is in the TOO queue. 

There is some test data that doesn't abide by this, but this may get resolved in some of the upcoming refactors to leverage factories instead of testdatagen functions for moves.

pkg/services/order/order_fetcher.go
```diff
@@ -17,6 +17,8 @@ import (
 	"github.com/transcom/mymove/pkg/services"
 )
 
+const RFC3339Micro = "2006-01-02T15:04:05.999999Z07:00"
```
kctruss, line 20: **suggestion:** I appreciated getting the context behind why this additional time zone logic was required. Would you mind adding a comment in the code to that effect so we don't lose that context to the Zoom ether?‏

ronaktruss:
Oh one other tidbit is that the submitted_at in moves is using timestamp and not timestamptz which theoretically means that comparing the times when using `GREATEST()` might be problematic (since we are comparing timestamps with timestamps with time zones)but since we set the default timezone of the db to be UTC consistently we should be good.

kctruss:
> Yeah that naming sounds more appropriate.
> 
> To answer your questions and provide some context:
> [. . .]

Thank you for this additional context. Given the tradeoffs, I do agree with you that storing this value makes the most sense and introduces the least risk short and long term.

mmayo-truss:
> Oh one other tidbit is that the submitted_at in moves is using timestamp and not timestamptz which theoretically means that comparing the times when using `GREATEST()` might be problematic (since we are comparing timestamps with timestamps with time zones)but since we set the default timezone of the db to be UTC consistently we should be good.

This is a good callout @ronaktruss

pkg/handlers/ghcapi/internal/payloads/model_to_payload.go
```diff
@@ -1440,7 +1442,17 @@ func QueueMoves(moves []models.Move) *ghcmessages.QueueMoves {
...
+	possibleValues := [3]*time.Time{move.SubmittedAt, move.ServiceCounselingCompletedAt, move.ApprovalsRequestedAt}
+	for _, time := range possibleValues {
+		if time != nil && (latestOccurance == nil || time.After(*latestOccurance)) {
+			latestOccurance = time
+		}
+	}
+	return latestOccurance
+}
+
+func findEarliestDateForRequestedMoveDate(shipment models.MTOShipment) (earliestDate *time.Time) {
```
mmayo-truss, line 1455: Praise: I appreciate this more specific function name.

mmayo-truss (approved):
This looks good @ronaktruss. I appreciate the thoroughness of your PR description and comment responses as well. It was helpful to provide extra context.

🏅 🌈

pkg/services/order/order_fetcher.go
```diff
@@ -17,6 +17,8 @@ import (
 	"github.com/transcom/mymove/pkg/services"
 )
 
+// Since timestamps in a postgres DB are stored with at the microsecond precision, we want to ensure that we are checking all timestamps up until that point to prevent moves from not showing up
```
mmayo-truss, line 20: Thank you for this extra context @ronaktruss

pkg/services/order/order_fetcher.go
```diff
@@ -369,6 +372,19 @@ func submittedAtFilter(submittedAt *time.Time) QueryOption {
...
 }
+
+func appearedInTOOAtFilter(appearedInTOOAt *time.Time) QueryOption {
+	return func(query *pop.Query) {
+		if appearedInTOOAt != nil {
+			start := appearedInTOOAt.Format(RFC3339Micro)
+			// Between is inclusive, so the end date is set to 1 microsecond prior to the next day
+			appearedInTOOAtEnd := appearedInTOOAt.AddDate(0, 0, 1).Add(-1 * time.Microsecond)
+			end := appearedInTOOAtEnd.Format(RFC3339Micro)
+			query.Where("(moves.submitted_at between ? AND ? OR moves.service_counseling_completed_at between ? AND ? OR moves.approvals_requested_at between ? AND ?)", start, end, start, end, start, end)
```
kctruss, line 385: **suggestion:** ‏This is purely a matter of preference, but I find complex comparisons like this easier to read with additional parentheses e.g.:

```
query.Where("((moves.submitted_at between ? AND ?) OR (moves.service_counseling_completed_at between ? AND ?) OR (moves.approvals_requested_at between ? AND ?))", start, end, start, end, start, end)
```

migrations/app/migrations_manifest.txt
```diff
@@ -793,3 +793,4 @@
 20230314155457_rogeruiz_cac_update.up.sql
 20230324203740_add_secondary_address_radio_states.up.sql
 20230404195728_katyc_cac_access.up.sql
+20230405164241_add-date-sent-for-service-item-approvals-to-moves.up.sql
```
kctruss, line 796: **nitpick:** ‏I was just corrected for this yesterday - migration naming convention is snake_case

kctruss (approved):
I just have a few additional comments that I would not consider to be blockers. Running through a handful of test cases worked exactly as I'd expect, and the migration ran as expected.

migrations/app/migrations_manifest.txt
ronaktruss: Nice catch

PR #10407, opened 2023-04-06
MB-15217 Update page title on navigation

scottmries:
## Summary

This PR adds a `useTitle` hook and `PageTitle` component that sets the page title and announces it using an `aria-live` element. This strategy is suggested [here](https://hidde.blog/accessible-page-titles-in-a-single-page-app/#heading-4).

The title is generated from a base title dependent on which application the component is in, and a subtitle based on the path. This subtitle can also be manually overridden by calling `useTitle(override)` at the component level, which this PR does in cases where manual titles were already being set.

The `PageTitle` component is included in all three applications at the App level, which gives us application-wide coverage.

Note that in some instances, generated page titles may include IDs or alphanumeric params from the path (e.g. the Move review page). This PR does not address these instances specifically.

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 2

Start the UI locally.

```sh
make server_run
```

</details>

### Additional steps

1. Access any page on the site
2. Note the title
3. Note that this title is different from others on the site
4. Note that a screenreader announces page changes


src/shared/Statements/AccessibilityStatement.test.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import AccessibilityStatement from './AccessibilityStatement';
+
+describe('Accessibility Statement page', () => {
+  it('has the correct title', () => {
+    render(<AccessibilityStatement></AccessibilityStatement>);
```
YanZ777, line 8:
```suggestion
    render(<AccessibilityStatement />);
```

src/shared/Statements/PrivacyAndPolicyStatement.test.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import PrivacyPolicy from './PrivacyAndPolicyStatement';
+
+describe('Privacy Policy page', () => {
+  it('has the correct title', () => {
+    render(<PrivacyPolicy></PrivacyPolicy>);
```
YanZ777, line 8:
```suggestion
    render(<PrivacyPolicy />);
```

src/scenes/SystemAdmin/index.jsx
```diff
@@ -33,7 +31,7 @@ class AdminWrapper extends Component {
           {/* no auth */}
           <Route path="/sign-in" component={SignIn} />
           <Route path="/invalid-permissions" component={InvalidPermissions} />
-          <Route path="/" component={defaultComponent} />)
```
YanZ777, line 36: **comment:** That's weird that there was just a random parentheses floating there.

YanZ777 (commented):
Thanks for picking this up! I have a few comments regarding organization and whatnot, but the code itself mostly looks good.

src/shared/utils.js
```diff
@@ -133,6 +142,7 @@ export function openLinkInNewWindow(url, windowName, window, relativeSize) {
     .open(
       url,
       windowName,
+      // eslint-disable-next-line prettier/prettier
```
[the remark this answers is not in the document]
YanZ777: I figured out why this isn't flagged, but that problem is a bit out of the scope of this PR. I'd still like to avoid adding  `eslint-disable-next-line`s though, which it looks like you managed to not need 🎉

src/hooks/custom.js
```diff
@@ -77,3 +82,54 @@ export const useCalculatedEstimatedWeight = (mtoShipments) => {
...
+  }
+  return baseTitle;
+}
+
+export function generatePageTitle(string) {
+  const baseTitle = getBasePageTitle();
+  return baseTitle + (string ? ` - ${string}` : '');
+}
+
+export function useTitle(string) {
```
rogeruiz, line 129:
```suggestion

/**
 * @func useTitle
 * @desc This function generates a subtitle using the pathname from the React Router DOM useLocation function unless a string is passed into the function and assigns the value to `subtitle`. It then calls useEffect to update the document's title attribute using the `generatePageTitle` function with the internally created subtitle.
 * @param {string} [string] - A string value to be used for the title instead of using the URL path.
 */
export function useTitle(string) {
```

src/hooks/custom.js
```diff
@@ -77,3 +82,54 @@ export const useCalculatedEstimatedWeight = (mtoShipments) => {
...
+  if (isMilmoveSite) {
+    baseTitle = MILMOVE_BASE_PAGE_TITLE;
+  }
+  if (isOfficeSite) {
+    baseTitle = OFFICE_BASE_PAGE_TITLE;
+  }
+  return baseTitle;
+}
+
+export function generatePageTitle(string) {
```
rogeruiz, line 124:
```suggestion

/**
 * @func generatePageTitle
 * @desc A function that generates the page title using a provided string appended to the output of `getBasePageTitle`.
 * @param {string} [string] - A string that is appended to the base title for a page.
 * @returns {string} A generated page title.
 */
export function generatePageTitle(string) {
```

src/hooks/custom.js
```diff
@@ -77,3 +82,54 @@ export const useCalculatedEstimatedWeight = (mtoShipments) => {
     return calculateEstimatedWeight(mtoShipments);
   }, [mtoShipments]);
 };
+
+/**
+ * This function a page subtitle from the path,
+ * by splitting the path at slashes, dashes and underscores, capitalizing, and join wish spaces and dashes.
```
rogeruiz, line 88:
```suggestion
 * by splitting the path at slashes, dashes and underscores, capitalizing, and join with spaces and dashes.
```

src/hooks/custom.js
```diff
@@ -77,3 +82,54 @@ export const useCalculatedEstimatedWeight = (mtoShipments) => {
...
+        segment
+          .split(/[-_]/)
+          .map((word) => capitalize(word))
+          .join(' '),
+      )
+      .join(' - ')
+  );
+}
+
+export function getBasePageTitle() {
```
rogeruiz, line 110:
```suggestion

/**
 * @func getBasePageTitle
 * @desc Creates a string using the variables with the `_BASE_PAGE_TITLE` suffix defined in and exported from `src/shared/constants`.
 * @returns {string} baseTitle - A base title which may or may not be blank.
 */
export function getBasePageTitle() {
```

rogeruiz (commented):
Adding some JSDoc comments and fixing some typos. Good work on the ones you've added already @scottmries.

rileyorr:
Great work on this @scottmries! Quick question: should I be testing this on the ephemeral deploy? I just poked around different pages and it might be something I'm doing wrong while navigating with VoiceOver but I'm not hearing the title announcement when I navigate between pages unless I tab to it. Additionally, I noticed that when navigating to edit a shipment the page's title contains many additional characters.

I took a screen recording for reference: 

https://user-images.githubusercontent.com/46329147/231217073-5b9201a6-55ff-46f7-8813-906d5402c797.mov

Let me know if you have any questions! I'm also happy to jump on a call to pair.

src/hooks/custom.js
scottmries: Thanks for these suggestions. I brought them in and updated them where appropriate.

src/components/TitleAnnouncer/TitleAnnouncer.module.scss
```diff
@@ -0,0 +1,5 @@
+.TitleAnnouncer {
+  position: absolute;
+  left: -999px;
+  top: -999px;
+}
```
YanZ777, line 5: **suggestion:** Can we rename this class to something like ".HiddenTItleAnnouncer" so that it's clear by the name that it should be hidden? 

**context:** This article also provides another alternative css to hide elements. https://css-tricks.com/places-its-tempting-to-use-display-none-but-dont/

scottmries: I applied your suggestion, and thanks for the context!

scottmries:
@rileyorr I refactored this after an update to route handling, and the title should be announcing across the entire app now.

rileyorr (approved):
Great job on this @scottmries! The titles are announcing across the entire app now 🎉 Overall the pages are announcing as expected, except on `/service-member/dod-info` the screen reader attempts to pronounce DOD as a word instead of an acronym ('dawd' vs 'dee oh dee'). 

However, I know this ticket has been open for a while and the page titles are announcing so I'm good with approving this if you'd prefer spinning up a separate ticket to resolve that use case in the future. Happy to provide an additional review either way.

scottmries:
Thanks, @rileyorr! Good point about the DOD Info page; we also have issues wherever there are IDs in the path, since those get read out literally.

I wonder if a good strategy would be to do a pass over the site and make a list of manual title overrides, and knock them all out in one go. That could be part of this work, but like you said, I'd just as soon merge this and follow up on those later on.

PR #10421, opened 2023-04-10
[ MB-15334 ]Adding SITDestinationFinalAddress to getMoveTastOrder endpoint

mmayo-truss:
## Summary

Added SITDestinationFinalAddress to the eager preload in the move task order fetcher in order to ensure that it shows up in the getMoveTaskOrder response payload.

## Setup to Run Your Code

<details>
<summary>💻 You will need to use one terminal to test this locally.</summary>

##### Terminal 1

Start the Go server locally.

```sh
make server_run
```

</details>

### Additional steps

1. Using Postman request a payload from the getMoveTaskOrder endpoint for a move with a service item that has a DDFSIT service item that has a SITDestinationFinalAddress. In the dev local data there is only one and the locator is `PARAMS`
2. Make sure that the DDFSIT service item displays the  SITDestinationFinalAddress in the payload. 
3. It will also now show up for DDASIT and DDDSIT service items as well.

## Screenshots
Before:
image

After: 
image


pkg/handlers/primeapi/move_task_order_test.go
```diff
@@ -237,6 +238,64 @@ func (suite *HandlerSuite) TestGetMoveTaskOrder() {
...
+		if suite.Len(movePayload.MtoServiceItems(), 1) {
+			serviceItem := movePayload.MtoServiceItems()[0]
+
+			// Take the service item and marshal it into json
+			raw, err := json.Marshal(serviceItem)
+			suite.NoError(err)
+
+			// Take that raw json and unmarshal it into a MTOServiceItemDestSIT
+			ddfSITSI := primemessages.MTOServiceItemDestSIT{}
+			err = ddfSITSI.UnmarshalJSON(raw)
```
ronaktruss, line 289: (nit/suggestion): I assume SI refers to service item. It took me a few seconds to figure out what this variable was doing. Can we use the full word?

pkg/handlers/primeapi/move_task_order_test.go
```diff
@@ -274,7 +333,7 @@ func (suite *HandlerSuite) TestGetMoveTaskOrder() {
 		movePayload := moveResponse.Payload
 
 		// Validate outgoing payload
-		suite.NoError(movePayload.Validate(strfmt.Default))
+		// suite.NoError(movePayload.Validate(strfmt.Default))
```
ronaktruss, line 336: I think this should be reverted?

ronaktruss (approved):
Thanks for adding that test. I left some feedback, but generally looks good.

Given the context, marshaling seems okay.

(Not work to be done in this PR) It might be worth thinking about how we want to structure future tests for some of the missing fields. Do we want singular tests for each of the fields. Can we have a single test that checks all of the fields? Does it make more sense to have the tests at the payload to model conversion and at the fetcher level.

pkg/handlers/primeapi/move_task_order_test.go
mmayo-truss: Thanks for that catch!

pkg/handlers/primeapi/move_task_order_test.go
mmayo-truss: Sure thing. I think I made it more clear with my most recent commit.

mmayo-truss:
> Thanks for adding that test. I left some feedback, but generally looks good.
> 
> Given the context, marshaling seems okay.
> 
> (Not work to be done in this PR) It might be worth thinking about how we want to structure future tests for some of the missing fields. Do we want singular tests for each of the fields. Can we have a single test that checks all of the fields? Does it make more sense to have the tests at the payload to model conversion and at the fetcher level.

@ronaktruss I went ahead and added another test at the fetcher level. I wanted to give you a chance to look it over before I merged.

pkg/services/move_task_order/move_task_order_fetcher_test.go
```diff
@@ -66,6 +66,40 @@ func (suite *MoveTaskOrderServiceSuite) TestMoveTaskOrderFetcher() {
...
+			},
+			Move: expectedMTO,
+			ReService: models.ReService{
+				Code: models.ReServiceCodeDDFSIT, // DDFSIT - Domestic destination 1st day SIT
+			},
+		})
+
+		actualMTO, err := mtoFetcher.FetchMoveTaskOrder(suite.AppContextForTest(), &searchParams)
+		suite.NoError(err)
+		// Should get two shipments back since we didn't set searchParams to exclude external ones.
```
ronaktruss, line 93: Is this a remnant of a previous test? Is this applicable here?

ronaktruss (approved):
The additional test looks good. I added a remark to one of the comments.

PR #10436, opened 2023-04-11
MB-15707 Fix linter errors in shared and scenes

YanZ777:
## Summary

Some of the rules are considered bypassable.
https://github.com/transcom/mymove/blob/main/eslint-plugin-ato/no-unapproved-annotation.js#L24-L71

Additionally, some of the code has changed enough that the eslint disable is no longer applicable. 

## Steps to reproduce
You'll need to modify the eslintrc files inside of the `shared` and `scenes` directories. Match what is in #10437 

Run `yarn run lint:production`. No errors should appear.

## Verification Steps for Author

These are to be checked by the author.

- [x] Request review from a member of a different team.
- [x] Have the Jira acceptance criteria been met for this change?

PR #10437, opened 2023-04-11
MB-15704 turn on linters in shared and scenes

YanZ777:
## Summary

This PR enables the custom ato linter rules to run inside of `shared` and `scenes`. 
I also added a check for `security/detect-non-literal-fs-filename` since we had a line that disabled that rule in #10436. 

## Testing this

Run `yarn run lint:production`. 
No errors should appear since this branch is based off of `yan-MB-15707-fix-errors-found-by-custom-linters-in-shared-and-scenes`, #10436.
That branch fixes all the errors that would appear.


src/scenes/.eslintrc
```diff
@@ -9,10 +9,14 @@
     "no-only-tests/no-only-tests": "error",
     "no-case-declarations": "error",
     "no-debugger": "warn",
+    "no-unused-vars": "warn",
```
felipe-lee, line 12: **question**: I'm curious why this should be "warn" instead of "error" ([the default if we used `eslint:recommended`](https://eslint.org/docs/latest/rules/no-unused-vars#options)). Though I do see we have it set to warn at the top level, so maybe nvm.

felipe-lee (approved):
LGTM, I have a question but it isn't a blocker. Thanks for turning these on!

src/scenes/.eslintrc
YanZ777: I decided to match what we set at the top level.

LeDeep (approved):

PR #10447, opened 2023-04-12
MB-15653 Update prime name and contract number

PDickeyTruss:
## Summary

This PR adds a new secure migration to update the Prime contractor name and contract number in the database. The secure migration file has already been uploaded to prod and staging s3 buckets. The scrubbed dev equivalent has been uploaded to demo, loadtest, and exp. It is also included in this PR per the documentation on [creating a secure migration](https://transcom.github.io/mymove-docs/docs/backend/setup/database-migrations/#creating-secure-migrations).  

There were also two queries in the backend that had hardcoded references to the old development contract number (`HTC111-11-1-1111`). Both of these were updated to instead query for the Prime contractor using the `type` field of the contractor table (`type="Prime"`). 

There is an additional schema (not secure) migration included in this PR to enforce the contractors type field is unique. 

## Setup to Run Your Code

<details>

##### Terminal 1

Start the UI locally.

```sh
make client_run
```

##### Terminal 3

Start the Go server locally.

```sh
make server_run
```

</details>

### Migration
The scrubbed development migration is equivalent to the secure migration with secure data (name/contract number) replaced with test data. The development migration can be tested locally by running `make db_dev_migrate`.  The secure migration can be tested locally by running `make run_prd_migrations`. 

### Code changes
The two code changes are a bit tricky to validate since the contractor queries are part of larger queries and the contractor info does not get included response data very frequently. Below is how I ended up testing that both updated queries successful and returning the 'Prime' contractor. 

#### Validating `moves.go` change
The changed query fetches the prime contractor so that its ID can be included when creating a new move. To test this you will need to create a new move using either the browser (customer app) or with a tool like [postman](https://transcom.github.io/mymove-docs/docs/tools/postman/setting-up-postman). Using a browser, watch the network tab for a POST request ending in `internal/orders` as you create a new move and add orders to it in the customer app. If this request is successful it implies the contractor query returned a contractor without any errors. You can track down the contractor that was returned using the moveCode and ContractorID (on to the move) if you would like to validate that the contractor was the 'Prime' contractor (`type` field of `contractors` table). If you would prefer to use postman or similar the (local) endpoint you will need is the POST request to `http://milmovelocal:3000/internal/orders` with a valid body. 

#### Validating `contractor.go` + `upload.go`
The query changes in contractor.go and upload.go update the contractor associated with a prime user upload to be the 'Prime' contractor rather than the (now removed) 'Test Prime' contractor. This can be tested in the prime simulator by uploading a new document to a move and verifying there were no errors. You can dig into the contractor that got associated to ensure it was the 'Prime' by finding the ContractID on your upload in the `prime_uploads` table in the database (sort by created_date). 

## Links
* [Setting Up Postman](https://transcom.github.io/mymove-docs/docs/tools/postman/setting-up-postman)
* [Secure Migrations Documentation](https://transcom.github.io/mymove-docs/docs/backend/setup/database-migrations/#creating-secure-migrations)


dymonmoore:
He @PDickeyTruss, the code looks good to me. I was just wondering how you made the migration "secure"?

dymonmoore (commented):
He @PDickeyTruss, the code looks good to me. I was just wondering how you made the migration "secure"?

PDickeyTruss:
> He @PDickeyTruss, the code looks good to me. I was just wondering how you made the migration secure?

This was actually my first secure migration, details of the process are here: [Secure Migrations Documentation](https://transcom.github.io/mymove-docs/docs/backend/setup/database-migrations/#creating-secure-migrations). In short, the migration w/ secure data is uploaded to an s3 bucket rather than checked into git. When running `make run_prd_migrations` secure migrations are grabbed from the s3 bucket rather than the dev ones.

pkg/models/contractor.go
```diff
@@ -36,16 +36,13 @@ func (c *Contractor) Validate(tx *pop.Connection) (*validate.Errors, error) {
 	), nil
 }
 
-// FetchGHCPrimeTestContractor returns a test contractor for dev
-func FetchGHCPrimeTestContractor(db *pop.Connection) (*Contractor, error) {
+// FetchGHCPrimeContractor returns the prime contractor
+func FetchGHCPrimeContractor(db *pop.Connection) (*Contractor, error) {
 	var contractor Contractor
-	err := db.Q().Where("contract_number='HTC111-11-1-1111'").First(&contractor)
```
reggieriser, line 42: It doesn't appear that the `name` field is unique per the database schema (the `contract_number` is, though).  In fact, when I ran `db_dev_e2e_populate`, I see two `Prime` entries (the second is named `Default contractor name for test`).  It looks like we're just telling Pop to give us the `First` which would be non-deterministic since we don't include any type of ordering on the query.

If the assumption is that we never have more than one contractor of a given type, we should probably enforce that with a database constraint, although that may mean some rework of some test stuff since it appears we do add an additional contractor there.  If we can have multiple `Prime` entries, then we need some way to guarantee we get the "right" one back each time.

pkg/models/contractor.go
```diff
@@ -36,16 +36,13 @@ func (c *Contractor) Validate(tx *pop.Connection) (*validate.Errors, error) {
...
+	err := db.Q().Where("type='Prime'").First(&contractor)
```
reggieriser, line 42: It doesn't appear that the `name` field is unique per the database schema (the `contract_number` is, though).  In fact, when I ran `db_dev_e2e_populate`, I see two `Prime` entries (the second is named `Default contractor name for test`).  It looks like we're just telling Pop to give us the `First` which would be non-deterministic since we don't include any type of ordering on the query.

If the assumption is that we never have more than one contractor of a given type, we should probably enforce that with a database constraint, although that may mean some rework of some test stuff since it appears we do add an additional contractor there.  If we can have multiple `Prime` entries, then we need some way to guarantee we get the "right" one back each time.

PDickeyTruss: Ah yeah, I see what you mean, bummer. It looks like the dev seed data is adding an additional Prime contractor after the migrations have already run.  My instinct is we should probably not have the tests creating contractors since they are set up via migration. Test should then use the contractors in the database the migrations created. Does that seem right? I suspect that won't be to painful to change code-wise. 🤞 

Also, I agree it would be good to add a uniqueness constraint if we are going to use the `type` field as if it is unique (`first()`).

migrations/app/schema/20230413164702_add_uniqueness_constraint_to_contractor_type.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE contractors
+  ADD CONSTRAINT unique_type UNIQUE (type);
```
PDickeyTruss, line 2: Constrain contractors.type to be unique

pkg/testdatagen/make_contractor.go
```diff
@@ -35,6 +35,14 @@ func MakeContractor(db *pop.Connection, assertions Assertions) models.Contractor
...
 	}
 
+	// Don't create multiple contractors of the same type
+	err = db.Q().Where(`type=$1`, contractor.Type).First(&contractor)
+	if err != nil && err != sql.ErrNoRows {
+		log.Panic(err)
+	} else if err == nil {
+		return contractor
+	}
+
```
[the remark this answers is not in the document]
reggieriser: I think this is OK, particularly now that we're adding a constraint that only one contractor of a given type can be present.

migrations/app/schema/20230413164702_add_uniqueness_constraint_to_contractor_type.up.sql
reggieriser: **nitpick:** We may want to prefix the constraint name with the table name (we appear to have that for the other constraints on this table like `contractors_contract_number_key`).  I can't remember if constraint names have to be unique across the table or the entire database, but having the prefix can't hurt and will be consistent.

migrations/app/migrations_manifest.txt
```diff
@@ -795,3 +795,5 @@
 20230403135359_add_fields_to_ppm_shipments_for_aoa_and_payment_packets.up.sql
 20230404195728_katyc_cac_access.up.sql
 20230405164241_add_date_sent_for_service_item_approvals_to_moves.up.sql
+20230412175236_update_prime_name_and_contract_number.up.sql
+20230413164702_add_uniqueness_constraint_to_contractor_type.up.sql
```
reggieriser, line 799: For some reason, when I try to download this secure migration like so:
```
download-secure-migration 20230413164702_add_uniqueness_constraint_to_contractor_type.up.sql
```
I get:
```
Downloading from: loadtest
fatal error: An error occurred (404) when calling the HeadObject operation: Key "secure-migrations/20230413164702_add_uniqueness_constraint_to_contractor_type.up.sql" does not exist
```
I don't have a problem doing the same with the `20230412175236_update_prime_name_and_contract_number.up.sql` file (and can see that we have different migrations now for stg/prd vs loadtest/demo/exp).  I even tried to download from each environment individually to see if there was one that was causing a problem, but they all failed in the same way.

However, when I run `make run_prd_migrations`, it seems to apply all migrations, including this one, and I see the constraint in the database schema. 😕 

Does this work for you?  Wondering if I'm doing something wrong.

PDickeyTruss: Ah yeah, I see what you are saying. I think that is expected since only the `update_prime_name_and_contract_number` is a secure migration. The`add_uniqueness_constraint_to_contractor_type` migration is just a normal schema migration, so it is not uploaded to s3. I will add something to the PR description similar to the below to make this a bit more clear since there are two migrations of different types in one PR. 

Migrations:
* `update_prime_name_and_contract_number` - Secure migration. Sensitave data version uploaded to prod and staging. Non-sensitive test version of the migration uploaded to demo, loadtest, and exp (as well as being checked into git). 
* `add_uniqueness_constraint_to_contractor_type` - Normal schema migration, not currently uploaded to s3.

migrations/app/schema/20230413164702_add_uniqueness_constraint_to_contractor_type.up.sql
PDickeyTruss: Updated, I for sure agree including the table name is a good idea. Thanks for spotting this! 💯

migrations/app/migrations_manifest.txt
reggieriser: Oh yeah, of course that makes sense now.  I don't know why that didn't cross my mind yesterday.

reggieriser (approved):
LGTM! 🎉

PR #10451, opened 2023-04-13
MB-15488 Add gotenberg-related service objects

felipe-lee:
## Summary

This was part of the work done in #10246 and is being broken out into its own PR so we can move forward with cleaned up and mergeable parts of the code (since the other PR is a proof-of-concept and shouldn't be merged).

One of the "Other" commits that replaces some hard-coded content types changes a lot of files so it might be easier to review that commit separately from the rest. In general I try to make it so that each commit can be reviewed in order from start to finish to ease reviewing the PR as a whole.

<details>

- [x] [feat (service): add helper func to convert a file to PDF](https://github.com/transcom/mymove/pull/10451/commits/899c1afb59387bf39053cdd151a2db132f49df50)
- [x] [feat (services): add UserUploadToPDFConverter service object](https://github.com/transcom/mymove/pull/10451/commits/f5a3294f809e550af74b05f6b0e798230a106d3d)
- [x] [feat (services): add service object to merge PDF files](https://github.com/transcom/mymove/pull/10451/commits/e8cadcf5deb04403e3f6b038cea334ed316eac23)

</details>

<details>

- [x] [feat (uploader): make file type strings be constants to ease re-use](https://github.com/transcom/mymove/pull/10451/commits/22c7c10a3ee81abf8deb89269c30425664b0264a)
- [x] [feat (storagetest): update fake s3 storage to also fail fetching](https://github.com/transcom/mymove/pull/10451/commits/acb4b13792e380bbc5865c3b11b2e328ee9b72d3)

</details>

## Testing Changes

<details>
<summary>💻</summary>

### Run Server Tests

```sh
make server_test
```

</details>

### Backend


pkg/services/paperwork.go
```diff
@@ -35,3 +38,33 @@ type FormTemplate struct {
...
+//go:generate mockery --name UserUploadToPDFConverter
+type UserUploadToPDFConverter interface {
+	ConvertUserUploadsToPDF(appCtx appcontext.AppContext, userUploads models.UserUploads) ([]*FileInfo, error)
+}
+
+// PDFMerger merges PDFs
+//
+//go:generate mockery --name PDFMerger
+type PDFMerger interface {
+	MergePDFs(appCtx appcontext.AppContext, pdfsToMerge []io.ReadCloser) (io.ReadCloser, error)
```
felipe-lee, line 69: I ended up not passing a slice of `FileInfo` structs here because we'll also be passing in the PDF that gets generated for the SSW, which is not an upload so it feels like it makes less sense to force them into the `FlieInfo` struct. Alternatively, the `FileInfo` struct could be made a little more generic in that we could add a file name to the struct (currently not there because we can pull the file name from the `userUpload` model). Then we could have an alternative constructor for when we don't have an original stream. Or since the stream can be `nil`, just have an expanded constructor that takes all the fields, but any other than file name can be `nil`. What do y'all think makes more sense? The downside I see to not passing the `FlieInfo` struct in any way to the PDFMerger is that the error messages are a bit generic in that they can only say the position in the slice of the one that has an error, instead of being able to say the real file name. Same with how we structure the request, though I think that is a little less important.

pearl-truss: If I'm following I think not using `FileInto` here is okay. I get that the error message will be more generic but knowing the position in the slice might be enough. But that's just my opinion, definitely curious what other folks think

pkg/services/paperwork/pdf_merger.go
```diff
@@ -0,0 +1,137 @@
...
+// pdfMerger is the concrete struct implementing the PDFMerger interface
+type pdfMerger struct{}
+
+// NewPDFMerger creates a new pdfMerger struct
+func NewPDFMerger() services.PDFMerger {
+	return &pdfMerger{}
+}
+
+// MergePDFs merges PDFs into a single PDF
+func (p *pdfMerger) MergePDFs(appCtx appcontext.AppContext, pdfsToMerge []io.ReadCloser) (io.ReadCloser, error) {
```
pearl-truss, line 26: **Praise:** thanks! I see a lot of earlier feedback incorporated in here and the other services!

pkg/services/paperwork.go
felipe-lee: @pearl-truss I edited my original comment to add some code examples of what I meant. Lmk if they make sense, if I should expand on anything, or if you have alternative ideas

pearl-truss: @felipe-lee sorry for just responding to this and thanks for additional context! I still think it makes the most sense to keep the approach to not pass in the `FileInto` slice. Seems to add additional complication without adding much imo

pkg/services/paperwork/paperwork_service_test.go
```diff
@@ -20,3 +24,30 @@ func TestPaperworkServiceSuite(t *testing.T) {
 	suite.Run(t, ts)
 	ts.PopTestSuite.TearDown()
 }
+
+// setUpMockGotenbergServer sets up a mock Gotenberg server and sets the corresponding env vars to make it easier to
```
scottmries, line 28: suggestion (non-blocking): I think you're missing a verb here. Did you mean "...easier to _test_ functions that..."?

praise: Regardless, this docstring is very helpful.

scottmries (approved):
LGTM! Thanks for this Felipe; your comments and your demos of this code have both been really hopeful in making this easier to review.
I left one comment, but it's not blocking.

pkg/services/paperwork/paperwork_service_test.go
felipe-lee: Nice catch, I'll push up a fix

